PR #1904, opened 2020-09-24
Fix part of #1861: [RunAllTests] Update Bazel CI workflow to run all Oppia tests

BenHenning:
Fix part of #1861.

At a high-level, this PR introduces running all of Oppia's Robolectric tests in GitHub Actions via Bazel now that Bazel has shown sufficient maturity while being used to continually build the app binary. In particular:
1. The tests take much longer to run since Bazel's builds are more complicated than on Gradle. This has been partially mitigated by improving parallelization (e.g. running each test as its own action), among other techniques described below. However, even with these optimizations Gradle will still outperform Bazel in the long-term when it comes to running tests. This is a cost we need to accept as a team.
2. One optimization done here is leveraging Bazel's remote caching functionality. However, to ensure the cache remains properly locked down, only branches created directly in this repo can leverage caching functionality. Consequently, forks from contributors will take longer to run through the actions. Despite caching, the overall performance benefits aren't as great as expected: only about 20% of the actions are being restored during build time. This is due to poor reproducibility in the data-binding part of the Android resources build pipeline and will require further investigation & fixes to address. The remote caching functionality is locked down using git-secret. Currently only I have access to the secrets, plus the fake robot account used by the workflow runner.
3. The new actions actually introduce computations for affected targets. This means only the tests relevant to the underlying changes will be run, except if on ``develop`` (e.g. for continuous builds) in which case all tests will run, or if the PR title includes ``[RunAllTests]``. While this is a nice addition, the real benefits won't be realized for a while: the lack of modularization means small changes (e.g. to just 1 app module file) will result in all tests tied to that module to need to re-run, and any changes to Bazel/BUILD files will trigger a re-run of the entire test suite since the Bazel files transitively expand to the whole set of BUILD files due to a circular dependency between domain & app. Changes to WORKSPACE will always trigger a full test re-run for safety.
4. This PR introduces automatic workflow cancellation. This is even more important now that we're running so many actions at once.
5. GitHub will try to run as many actions at once as it can. I've seen times when it only runs 5 simultaneously, and other times where it runs nearly 50. This depends on current active resources across the organization, so we'll need to keep an eye on runtime to see how good/bad it is for contributors.
6. This PR also fixes a few build errors that have crept up in the last few weeks/months.
9. Per (3) this PR is running all of the tests to demonstrate that they work as expected (though given that a Bazel file was changed in this PR, it would be likely that all of the tests would have been run, anyway).
[...]

PR #2693, opened 2021-02-12
[RunAllTests] Use faster Bazel setup with 4.0 LTS

BenHenning:
This is a follow-up from #1904.

This is meant to speed up Bazel in CI by using a much leaner repository containing custom Bazel setup (just the Android tools from https://github.com/oppia/oppia-bazel-tools versus cloning the much larger oppia/bazel repository). This is expected to shave up to 30 seconds off each test action requiring Bazel (depending on how long the Bazel setup action itself takes).


BenHenning:
Ah nice. This is much faster: compute affected targets took ~18s instead of the ~1.5 minutes from before. There will still be a slight delay from cloning the custom tools repo, but it should still be a nice improvement over what the previous setup was doing.

jcqli (approved):
LGTM, waiting on tests to finish

BenHenning:
Thanks! Looks like another JVM crash in CI. :( Restarting failing tests & enabling auto-merge.

PR #2656, opened 2021-02-08
Fix #2634: Fix talkback bug on activity transition to topic screen

Sparsh1212:
## Explanation

Fixes #2634: The talkback previously used to read "Oppia" on transition to the topic info screen.  In this PR, this bug is fixed and it now reads "Topic: `topic_name`" instead of "Oppia".

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #2694, opened 2021-02-12
Reorganize GitHub Actions workflows

BenHenning:
This PR makes a variety of organizational changes to the workflow files:

1. It moves linter checks to the static analysis workflow (which seems like a generally better place for them).
2. It moves the Bazel build job to its own, dedicated workflow (which will also be able to contain future build tests--these can be useful to get early checks for more expensive tests like e2e without waiting for the whole suite to run; while they don't verify correctness, they can catch cases when the test would definitely fail to run)
3. It renames the main workflow to correspond specifically to Gradle tests
4. It cleans up some of the workflow documentation

This is part of a broader effort to clean up our workflows & improve restartability if something fails (since GitHub only supports restarting whole workflows vs. specific jobs).


BenHenning:
Note that I will switch the required check to the new one before merging this & after it's approved to fix the missing Lint Tests check expected above.

anandwana001:
Nit conflicts.

anandwana001 (changes requested):
I think we should rename this `name: CI Lint and Tests` as no more linting from `main.yml`

BenHenning:
Agreed @anandwana001. I'll rename it to 'Gradle Tests' since that's all that's left, but this has to happen right before I merge since I need to go & update the required checks in the repository settings at the same time. You'll need to approve the PR before I can make that change (otherwise it will fail the required checks). If you're okay with that, I can make the change first otherwise I suggest approving so that I can update settings & the PR at the time of merging.

anandwana001 (approved):
Sounds Good

BenHenning:
Ah, actually the Bazel binary build is still in this workflow. I'll create another PR to move that out before updating this one.

Edit: hmm actually, I'm going to go ahead and make all the workflow changes I think are worthwhile even if it breaks the tests in this PR. I think it actually does make more sense to get a sign-off on the final state.

BenHenning:
Thanks @anandwana001. Given this PR has changed & evolved a bunch, PTAL again.

As before, I'll update required checks after approval (but I did go ahead with the final names so that I can get your thoughts). Given that we require everyone to keep their branches up-to-date right now, the new required checks will be naturally incorporated into PRs over the next couple weeks (& immediately for new PRs).

.github/workflows/build_tests.yml
```diff
@@ -0,0 +1,84 @@
+# Contains jobs corresponding to build tests (such as verifying that certain targets build, including the app itself).
+
+name: Build Tests
```
anandwana001, line 3: Can we say it `Bazel Build Test`, as we will be running the tests on Bazel or is it just Bazel Building in this file?

.github/workflows/static_checks.yml
```diff
@@ -20,3 +20,56 @@ jobs:
         with:
           checks: "duppatterns,files,syntax"
           experimental_checks: "notowned"
+
```
anandwana001, line 23: Can we add this above

```
# This workflow has the following jobs:
# Check Codeowners
# linters: Java, Kotlin, Protobuf, Bazel
```

.github/workflows/build_tests.yml
anandwana001: ah, we are running Bazel test in `.github/workflows/unit_tests.yml`, so I guess we can rename this file which gives a sense that this file is trying to build apk using bazel.

.github/workflows/unit_tests.yml
```diff
@@ -1,3 +1,5 @@
+# Contains jobs corresponding to unit tests. Integration (e.g. emulator or screenshot testing) or End-to-End tests should not be triggered here.
+
 name: Unit Tests (Robolectric - Bazel)
```
anandwana001, line 3: Can we name this file `.github/workflows/bazel_unit_tests.yml`

anandwana001 (changes requested):
Nit points @BenHenning

.github/workflows/build_tests.yml
BenHenning: The idea for this file was that it hold all build tests (e.g. tests that are pass/fail based on whether a build succeeds rather than the runtime). This may be useful as a location to add future tests where building is much faster than running to get an early-failure indication for those tests without waiting for the full runtime to complete (such as potentially end-to-end tests).

.github/workflows/static_checks.yml
BenHenning: Can you clarify the purpose or benefit of doing this? I know we did it earlier, but it seems fairly easy to determine from a glance or from GitHub itself which jobs are run, and this list easily becomes out-of-date.

.github/workflows/unit_tests.yml
BenHenning: Right now the name is indicating Robolectric & Bazel, but once we remove Gradle this is really just a place to hold unit tests (which is actually the intent). Bazel is a bit irrelevant in that way.

I'm fine with renaming 'main' to be 'gradle_checks' or similar, though, if that would help clarify. WDYT?

BenHenning:
Thanks @anandwana001. I replied to all--PTAL.

.github/workflows/static_checks.yml
anandwana001: Yes, this seems like require updation all the time, let's skip this.

anandwana001 (approved):
LGTM, naming sounds good.

BenHenning:
Thanks! Will sync with develop after #2656 is merged, then prepare this for merging.

Issue #2691, opened 2021-02-11
Verify Bazel CI test changes are working as expected


BenHenning:
From #1904:

- [x] Verify that the workflows behave correctly on forks where caching has to be disabled
- [x] Verify that the automatic at-head builds on develop work correctly with the new workflow changes
- [x] Verify that manual workflow runs work correctly with the changes
- [ ] (Once 1-3 are completed) Set the new 'Run test' workflow as required
- [ ] (After 4) Verify that the requirement doesn't break when creating a PR that doesn't run any affected tests (thereby not running the required workflow)
- [ ] Send out a communication explaining the new tests, what people can expect, how to investigate issues, and who to contact
- [ ] Fix whatever is going on with this: https://github.com/oppia/oppia-android/pull/2671/checks?check_run_id=1908497857 (affected targets issues)
- [x] Fix hanging compute affected targets actions when Bazel files themselves are changed (#2648)

BenHenning:
Running on develop: https://github.com/oppia/oppia-android/actions/runs/559499045 (strange JVM crash failure, but tests otherwise look operational).

BenHenning:
Manual run kicked off: https://github.com/oppia/oppia-android/actions/runs/560109387. Seems to be working, but I'll check in just to make sure before checking the above.

BenHenning:
For the fork test, will need to wait for a new PR to be created to check.

BenHenning:
Ugh, looks like the check might be failing when no targets are computed: https://github.com/oppia/oppia-android/actions/runs/560112155.

BenHenning:
I think we can conclude that manually triggering the workflows does work, and that the no affected targets check is NOT working. I'm fairly confident once the underlying issue is fixed that the conclusion task should pass & thus work when its required, so I may preemptively check that once #2696 is verified as working & merged in.

BenHenning:
Another issue: https://github.com/oppia/oppia-android/pull/2695.

BenHenning:
Confirming forks fail at least due to the key management piece: https://github.com/oppia/oppia-android/pull/2690. There's probably a check missing here.

BenHenning:
Per https://github.com/oppia/oppia-android/pull/2690/checks?check_run_id=1885145971 it looks like the fork check is failing unexpectedly. Those actions weren't supposed to run which is why they're failing.

BenHenning:
[...]

PR #2696, opened 2021-02-12
Fix affected target computation breaking CI workflows when no targets are affected

BenHenning:
See https://github.com/oppia/oppia-android/issues/2691#issuecomment-777995766 & https://github.com/oppia/oppia-android/actions/runs/560112155 for context. It seems when there are no affected targets an empty line will be printed which breaks something between the jobs. I think this will fix it.

Conveniently, this PR should trigger the "no affected targets" workflow & can self-verify itself.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,6 @@ on:
 jobs:
   linters:
     name: Lint Tests
-    if: "false"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -66,7 +65,6 @@ jobs:
 
   robolectric_tests:
     name: Robolectric Tests (Non-App Modules)
-    if: "false"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


BenHenning:
(Note that this is effectively in 'draft' mode but can't actually be marked as draft since we need to run CI for investigation here).

BenHenning:
https://github.community/t/empty-matrix-fails-workflow/128736/2 seems like a viable fix.

BenHenning:
Now that #2710 is merged restarting these workflows to see if I fixed the issue.

BenHenning:
Yay, looks like this solution works. However, it does cause the check test results action to fail. :( I was partly expecting that to happen. I guess we need to also need to condition the steps of that task or something. Need to think on it a bit more.

BenHenning:
Awesome, it's passing now. Reenabling other tests & sending this out for review.

anandwana001 (approved):
LGTM, Thanks @BenHenning

PR #2697, opened 2021-02-12
[RunAllTests] Consolidate testing to ensure all tests are covered in Bazel CI

BenHenning:
This PR updates our test declarations such that all tests under each module are always included. In all, this effort has introduced 17 test targets that were previously missing.

Note that this is intentionally a stop-gap solution since we're planning to move all tests over to individual targets (at which point we will need another check in place to ensure tests are being properly added). To prepare for this work, each module BUILD file now has a MIGRATED_TESTS list that will automatically filter out migrated tests from module-level test generation.

This PR includes fixes for ExceptionsControllerTest (similar to the fixes for LogWorkerUploadTest in #1904) since stack traces differ at runtime between Gradle & Bazel. Also, this PR forces all app module tests to be resource-compatible which introduces an extra genrule for each one. This does affect the build time slightly, but it's not substantially noticeable unless large numbers of tests are being built/run at one time.


BenHenning:
Forcing all tests to run in this PR to illustrate the new tests being covered. I can't rely on compute_affected_targets since there seem to be some issues with it at the moment (see https://github.com/oppia/oppia-android/issues/2691#issuecomment-778042891).

app/src/test/java/DataBinderMapperImpl.java
```diff
@@ -14,3 +14,4 @@
     addMapper(new org.oppia.android.app.test.DataBinderMapperImpl());
   }
 }
+
```
rt4914: Remove this extra blank line.

rt4914 (approved):
Nit change suggested. You can resolve this comment directly.

app/src/test/java/DataBinderMapperImpl.java
BenHenning: Oops; removed. Thanks!

fsharpasharp (approved):
Great, this will be really helpful for the migration to make sure the tests don't break.

BenHenning:
Updated the PR to include some reworking of Bazel doc comments to be more normalized/consistent.

vinitamurthi (approved):

BenHenning:
@anandwana001 PTAL--we should try to get this in soon since it's reintroducing tests that were previously missed in Bazel builds.

anandwana001 (approved):
LGTM, missed it earlier.

PR #2707, opened 2021-02-12
Fix #2706: Use a more correct pattern for checking forks

### LINKED ISSUE #2706: CI checks are blocked due to Bazel build check failing on forks
See https://github.com/oppia/oppia-android/issues/2691#issuecomment-778337053 for context. This was caused by #1904, and I'm going do 1 attempt at fix-forwarding it for timeliness. If that doesn't work, I'll be reverting the original change.

BenHenning:
Fixes #2706.

This attempts to fix the CI issue by using a better way to check if the run is actually running on a fork. The original solution might have broken, or maybe it was always wrong. The new one is sourced from https://github.community/t/how-to-detect-a-pull-request-from-a-fork/18363.

**NOTE TO REVIEWERS**: I am doing this as a fix-forward because we can verify full correctness from the CI run on this PR itself. If you prefer for me to revert the original PR per the reversion policy, I will do that instead. Please let me know.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -177,7 +177,7 @@ jobs:
       # This also uses a different directory to install git-secret to avoid requiring root access
       # when running the git secret command.
       - name: Install git-secret (non-fork only)
-        if: github.repository == 'oppia/oppia-android'
+        if: ${{ github.event.pull_request.head.repo.full_name == 'oppia/oppia-android' }}
         shell: bash
         run: |
           cd $HOME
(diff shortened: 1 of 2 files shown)


BenHenning:
Note that it looks like this _is_ fixing the issue since the Bazel build step is correctly using the 'fork' paths now instead of the non-fork ones. Note that the non-required Bazel tests are still failing due to other issues that I'm working on--you can ignore those.

rt4914 (approved):
Approving this considering this solution is from Github Community Guideline and is marked as "Solved" there.

rt4914:
> Note that it looks like this _is_ fixing the issue since the Bazel build step is correctly using the 'fork' paths now instead of the non-fork ones. Note that the non-required Bazel tests are still failing due to other issues that I'm working on--you can ignore those.

Ack.

srikanthkb (approved):
LGTM.

BenHenning:
Thanks!

PR #2320, opened 2020-12-26
Fix #936 & #2297 : Oppia Android Promoted Stories App Layer.

veena14cs:
## Explanation

This PR is the app layer implementation of promoted stories and coming soon topics on the home screen. This PR fixes issues #936 & #2297 .

**Design Document**
https://docs.google.com/document/d/1o7hOz1rP-hJaKn1f5x6LV6ms0bzhyCpsTifJQY5crCc/edit?ts=5f9fc81b#heading=h.ovlu0iazlug6

Mock Links
- https://xd.adobe.com/view/ee9e607b-dbd6-4372-48dc-b687d32af3da-98af/screen/1086e5df-d445-46a5-949e-5a7d02d5045a/specs/
- https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/67bfcbac-7b8e-491b-89c3-38937b37e5c2/
- https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/b3019177-fe86-459e-8a63-8109c53005f5/specs/
- https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/114521bd-bd0f-426f-83f9-3fa8180e3c89/

## Checklist

- [ ] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [ ] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [ ] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [ ] The PR is made from a branch that's **not** called "develop".
- [ ] The PR is made from a branch that is up-to-date with "develop".
- [ ] The PR's branch is based on "develop" and not on any other branch.
- [ ] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #2272, opened 2020-12-16
Oppia Android Promoted Stories Domain Layer

veena14cs:
## Explanation

This PR shows a list of promoted stories on the home screen. This implemenation is Domain layer part and App layer implementation is in PR #2320 
The list(referred to as promoted stories list in code)  can show one of the following:

1. Recently Played Stories
2. Last Played Stories
3. Stories for you
4. Recommended Stories
5. Coming Soon Topics
6. Or no list.

**The number of cases for Stories for you** 
- While onboarding new users a list of topics is shown on the HomeScreen , based on this selection it promotes the first chapter from the first story from this particular topic.
- Based on the selection of the topic, the first chapter from next topic will be recommended.
- For e.g.  If user started first chapter from "Fractions" but not finished, the app promotes ongoing topic "Fraction" and the next topic "Ratios" under Recommended Stories. 

**The number of cases for Recommended stories** 
- If the user has completed first story from "Fractions" , and there are no more ongoing stories, then it promotes only "Ratios" under Recommended Stories.

**The number of cases for Recently-played/Last-played stories** 
- Considering there are 2 promoted stories "Fractions" and "Ratios" under Recommended Stories.
- If user starts first chapter from "Ratios"  but not finished, it promotes "Ratios" & "Fraction" topics,  changes the header from Recommended stories to Recently-played stories
- If the above topics are in progress from more than a week, header of promoted list is changed to Last-played stories.
- If the user finishes first story from "Fraction", It will have only one topic "Ratios" under Recently-played stories.
- Now, if the user has finished first story from "Ratios", It promotes next topic's first story and changes header from Recently-played/Last-played to Recommended Stories.

**Coming Soon topics**
- User has finished playing at least one story from all the topics, header of promoted list is changed to **Coming Soon** topics and a list of coming soon topics is shown.

**Design Document**
https://docs.google.com/document/d/1o7hOz1rP-hJaKn1f5x6LV6ms0bzhyCpsTifJQY5crCc/edit?ts=5f9fc81b#heading=h.ovlu0iazlug6

**Note**:
This PR is only focusing on domain layer. As there are high-end changes and app layer CI check are failing. This errors are fixed in PR #2320 .

## Checklist

- [ ] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
[...]

PR #2708, opened 2021-02-12
Fix initial recommendation

veena14cs:
## Explanation
This PR is the fix for #2272 and #2320. The initial recommendation on home screen, when user has no progress is fixed in this PR. New user with no guidance can get recommended topics to learn.

## Checklist


BenHenning:
Ah, this is going to require #2320 to get merged in first. It's far too large for me to review on its own.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -783,9 +808,9 @@ class HomeActivityTest {
   fun testHomeActivity_firstTestTopic_topicSummary_topicNameIsCorrect() {
     launch<HomeActivity>(createHomeActivityIntent(internalProfileId1)).use {
       testCoroutineDispatchers.runCurrent()
-      scrollToPosition(position = 2)
+      scrollToPosition(position = 3)
```
BenHenning, line 811: Here and elsewhere: why do these positions need to be updated?

BenHenning (commented):
Thanks @veena14cs. I reviewed the tests but not the controller yet. Had one comment.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
veena14cs: Initially when there are no topic's in progress, recommended stories are shown by default. So the all topics list item position is moved down by 1position.

veena14cs: Ditto else where.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -429,6 +429,24 @@ class TopicListController @Inject constructor(
     }
   }
 
+  /*
+  * # Dependency graph:
+  *
+  *      Fractions
+  *         |
+  *        |
+  *       v
```
rt4914: Should these be shifted to represent an arrow?

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -429,6 +429,24 @@ class TopicListController @Inject constructor(
     }
   }
 
+  /*
+  * # Dependency graph:
```
rt4914: Why do we need this `#` here and below too?

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -245,72 +243,74 @@ class TopicListController @Inject constructor(
     topicProgressList: List<TopicProgress>,
     completionTimeFilter: (Long) -> Boolean
   ): List<PromotedStory> {
+    if (topicProgressList.isNotEmpty()) {
+      val playedPromotedStoryList = mutableListOf<PromotedStory>()
+      val sortedTopicProgressList =
+        topicProgressList.sortedByDescending {
+          val topicProgressStories = it.storyProgressMap.values
+          val topicProgressChapters = topicProgressStories.flatMap { it.chapterProgressMap.values }
```
rt4914: In this and above lines `it` can be confusing because of nesting.
Suggest using `topicProgress` and `storyProgress` respectively.

rt4914 (approved):
Left high level comments else LGTM. I have tested this by running the app and it works correctly. Haven't been able to run the tests on my local device but overall looks good. Will need to defer to @BenHenning on full verification of tests.

@BenHenning You can resolve my comments if needed.

Also, approving this PR considering it looks good and also to not keep it blocked.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: Done,

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: I tried shifting it but it doesn't make perfect arrow.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: Added it to highlight the title. Found it in one of the comments in our code in `StatePlayerRecyclerViewAssembler` for Flowchart.

rt4914: Okay

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
rt4914: I think if the initial space is consistent it should work.

rt4914: @veena14cs PTAL

veena14cs: I tried giving consistent spacing, but didn't workout.

rt4914: @BenHenning Any specific reason why it might not be getting aligned in line?

veena14cs: @rt4914 I have updated code please check. Also there is difference in how it shows up in Github and in Android Studio. I am attaching screenshot for reference.
Screenshot (13)

rt4914: Looks correct. Thanks. It should look correct in android studio.

rt4914: For me it looks like this:
Screenshot 2021-02-17 at 11 01 43 AM

veena14cs: Thanks! my Android studio is weird :).

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -429,6 +431,24 @@ class TopicListController @Inject constructor(
     }
   }
 
+  /*
+  * # Dependency graph:
```
BenHenning: This is a test-only dependency graph--I don't think we should be adding it as a comment to production code. If you want to use it as an example for the explanation below, that's fine but it should be reordered & contextualized correctly.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -245,72 +243,76 @@ class TopicListController @Inject constructor(
     topicProgressList: List<TopicProgress>,
     completionTimeFilter: (Long) -> Boolean
   ): List<PromotedStory> {
+    if (topicProgressList.isNotEmpty()) {
```
BenHenning: Why is this check needed? Won't an empty list still be returned even without it?

I'm also slightly confused about the logic here. It seems exactly the same except one change: before we didn't set the promoted story list at all if there wasn't topic progress, but now we set an empty list. How does that result in the correct recommendations being shown?

BenHenning (changes requested):
Thanks @veena14cs! I had one suggestion for potentially simplifying the logic, and another looking for clarity on the solution since I think I'm missing something. PTAL.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: Before I had this check in `computePromotedActivityList()` if it is empty it was returning `promotedActivityListBuilder` with empty list. But as we need to initially recommend some topics without any progress I moved this check to `computePlayedStories()`. Here we doing some computation on topic's that are ongoing so we need to check if it not empty.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: I was thinking, this will be helpful to understand the logic so added it over here.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: I thought if I don't add a check it would result in crash. I cross checked and if I don't add check it works fine. So removed this check and updated code.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: By reordering you mean?

BenHenning: It should be something like:

```kotlin
/*
 * Explanation for logic: ...
 *
 * Here's an example topic graph to illustrate:
 *  ...
 *
 * In this example, when topic X is finished, topic Y will be recommended.
 */
```

I think this will flow a bit better.

Also, we shouldn't be using markdown here since it's not a KDoc comment. If you intend it to be a KDoc comment, you should change it to use ``/**`` and have the first bit be a summary fragment for the method. After that you can explain the specific logic & use the example (though this honestly seems better as a block comment rather than a KDoc). Finally, I don't recommend using headers anyway since they can be obnoxiously large depending on the markdown renderer, but that's mainly my opinion. :)

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
BenHenning: The new implementation seems resilient to crashing since it's always operating on the original list (which is empty).

What I don't understand is: why does empty topic progress list result in the first two topics of the graph being suggested? That's how it's supposed to work (per computeSuggestedStories), but why does a now-empty promoted story list trigger computeSuggestedStories to return the correct thing?

veena14cs: As per earlier documentation, unless user has any ongoing topic it wasn't suggesting any topics, which in this case gives recommendation for selected topics. Our previous logic was all correct except that, we were not entering into computation logic of suggestion, if there are no topics in progress. Hope this clears your doubt.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -440,11 +445,7 @@ class TopicListController @Inject constructor(
   *     \                           /
   *       -----> Test topic 1 <----
   *
-  * # Logic for recommendation system
-  *
-  * We always recommend the next topic that all dependencies are completed for. If a topic with
-  * prerequisites is completed out-of-order (e.g. test topic 1 above) then we assume fractions is already done.
-  * In the same way, finishing test topic 2 means there's nothing else to recommend.
+  * In this example, when topic X is finished, topic Y will be recommended.
```
BenHenning: Please fill in 'X' and 'Y'. That was an example from my comment--we should actually be specific here otherwise the example isn't very helpful to readers.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
BenHenning: Ah, I see where I was making the mistake in reading the code. This does make complete sense now--thank you. :)

BenHenning:
Thanks @veena14cs! Just had one comment. PTAL & then I think this will be ready for merging.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
veena14cs: Done. Thanks.

veena14cs:
> Thanks @veena14cs! Just had one comment. PTAL & then I think this will be ready for merging.

@BenHenning  done thanks!

BenHenning (approved):
Thanks @veena14cs! LGTM.

PR #2724, opened 2021-02-14
Added tracking for hints and solutions in training session

TheRealJessicaLi:
## Description

Added tracking for viewing hints and solutions during a training session, as per design doc [here](https://docs.google.com/document/d/1ISuqvKs6b0xdE0-aEIHXYG7oDbCmDBcgUia0TiR9GpQ/edit?ts=5fb0a916#). Every time the user views a hint or solution for a question, the corresponding `QuestionSessionMetrics` instance for that question is updated to reflect the state change.

## Checklist


BenHenning:
Hi @TheRealJessicaLi. Couple of things:
1. I'm going to defer to @vinitamurthi to take a first pass on these PRs for now (since I'm a bit overloaded--I need to load-balance where I can)
2. Thinking about this a bit more, another approach you might be able to take with this PR is to first introduce the API that adds the ability to analyze metrics & add tests for that (which would more or less verify that no metrics are being computed yet), then add this PR after to verify that the metrics are now being computed. This is sort of similar to [red-green refactoring of TDD](https://www.codecademy.com/articles/tdd-red-green-refactor) (sort of), and it avoids the awkward gap of introducing code without testing. Given I advised the current approach over chat I won't explicitly push back on your PR, but I wanted to propose this as a potentially better alternative that's a bit better from a "technical discipline" perspective.

TheRealJessicaLi:
> Hi @TheRealJessicaLi. Couple of things:
> 
> 1. I'm going to defer to @vinitamurthi to take a first pass on these PRs for now (since I'm a bit overloaded--I need to load-balance where I can)
> 2. Thinking about this a bit more, another approach you might be able to take with this PR is to first introduce the API that adds the ability to analyze metrics & add tests for that (which would more or less verify that no metrics are being computed yet), then add this PR after to verify that the metrics are now being computed. This is sort of similar to [red-green refactoring of TDD](https://www.codecademy.com/articles/tdd-red-green-refactor) (sort of), and it avoids the awkward gap of introducing code without testing. Given I advised the current approach over chat I won't explicitly push back on your PR, but I wanted to propose this as a potentially better alternative that's a bit better from a "technical discipline" perspective.

No worries, thanks for taking the time to review my PRs so far! 

That's a good idea, but I think I would need this tracking implementation to run any tests for the API regardless; even if I were to write the calculation code first, I couldn't test it until these tracking methods are added, since the API is using the tracked interaction state under the hood. There also aren't that many more changes needed for tracking - after this PR, I only need 1 additional PR to track the answers submitted. @vinitamurthi What do you think?

vinitamurthi:
> > Hi @TheRealJessicaLi. Couple of things:
> > 
> > 1. I'm going to defer to @vinitamurthi to take a first pass on these PRs for now (since I'm a bit overloaded--I need to load-balance where I can)
> > 2. Thinking about this a bit more, another approach you might be able to take with this PR is to first introduce the API that adds the ability to analyze metrics & add tests for that (which would more or less verify that no metrics are being computed yet), then add this PR after to verify that the metrics are now being computed. This is sort of similar to [red-green refactoring of TDD](https://www.codecademy.com/articles/tdd-red-green-refactor) (sort of), and it avoids the awkward gap of introducing code without testing. Given I advised the current approach over chat I won't explicitly push back on your PR, but I wanted to propose this as a potentially better alternative that's a bit better from a "technical discipline" perspective.
> 
> No worries, thanks for taking the time to review my PRs so far!
> 
> That's a good idea, but I think I would need this tracking implementation to run any tests for the API regardless; even if I were to write the calculation code first, I couldn't test it until these tracking methods are added, since the API is using the tracked interaction state under the hood. There also aren't that many more changes needed for tracking - after this PR, I only need 1 additional PR to track the answers submitted. @vinitamurthi What do you think?

As discussed, I think we should go ahead with the tracking first, followed by the calculation of the score.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -201,6 +201,7 @@ class QuestionAssessmentProgressController @Inject constructor(
           // in an 'always showing hint' situation. This can specifically happen if hint throws an
           // exception.
           progress.advancePlayStageTo(TrainStage.VIEWING_STATE)
+          progress.hintViewed()
```
vinitamurthi: I am wondering if we should put this in the try block rather than the finally block. Reason being that we should penalize the user only if they actually were able to view the hint. If the hint fails to be revealed for whatever reason, we probably shouldn't penalize them. What do you think?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -234,6 +235,7 @@ class QuestionAssessmentProgressController @Inject constructor(
           // in an 'always showing solution' situation. This can specifically happen if solution
           // throws an exception.
           progress.advancePlayStageTo(TrainStage.VIEWING_STATE)
+          progress.solutionViewed()
```
vinitamurthi: Ditto here

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
TheRealJessicaLi: Yeah that makes total sense, thanks for pointing that out! I just double checked this, I assumed (incorrectly) that the finally block only executes if you don't get an exception.

TheRealJessicaLi: Fixed, thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
TheRealJessicaLi: Fixed!

vinitamurthi (approved):

PR #2725, opened 2021-02-15
Fix #2542: Update the ktlinter check script to check for committed files only

### LINKED ISSUE #2542: Improvement of pre-push lint checks
When working on changes on local repository, we usually add debug statements, make temporary changes in the code. Sometimes, we make changes in files which are not going to be committed or pushed to the remote repository. Even though, we do not include these files in a git commit, when we actually try to push changes in files (to be pushed to a remote branch), our pre-push git hook, performs `ktlint` check on all of the files (including temporary files/un-committed files). 

We are not allowed to push our changes until we fix the lint errors and warnings in the un-committed files, which is not required (as they are never going to be published) and could get time consuming for major changes. 

If possible, improve the pre-push git hooks scripts to check **ONLY** for the files that are **included** in the commit(s) being pushed to remote branch. This way, we wouldn't have to clean-up the code in other files before publishing the changes.

srikanthkb:
## Explanation

Fixes #2542 
* Use git diff-filters to run ktlinter on committed files only
* This will prevent checks on unstaged files, which is not necessary

## Checklist


rt4914:
Unassigning myself as I am not the code-owner. If you still need me take a look please assign back. Thanks.

scripts/ktlint_lint_check.sh
```diff
@@ -17,7 +17,7 @@ fi
 committed_files=`git diff --name-only HEAD^ HEAD -- '***.kt'`
 
 if [ -z "$committed_files" ]; then
-    echo No kotlin files have been committed, skipping ktlinter.
+    echo "No kotlin files have been committed, skipping ktlinter."
```
anandwana001: "skipping file formatting checks on kotlin file" how this sounds?

anandwana001 (changes requested):

scripts/ktlint_lint_check.sh
srikanthkb: Sounds more informative, I will update a commit. 
Also, for the situation when a user runs the `ktlint_lint_check.sh` before committing a file, no checks will be performed (as these changes are not committed). Should we also add a statement saying "Please commit the files that you want to check..." or something like that, which could prompt the user to commit and then run the script again ?

srikanthkb: > "skipping file formatting checks on kotlin file" how this sounds?

Done.

scripts/ktlint_lint_check.sh
```diff
@@ -14,7 +14,15 @@ else
     jar_file_path="$github_actions_path/oppia-android-tools/ktlint"
 fi
 
-java -jar $jar_file_path --android app/src/**/*.kt data/src/**/*.kt domain/src/**/*.kt testing/src/**/*.kt utility/src/**/*.kt
+committed_files=`git diff --name-only HEAD^ HEAD -- '***.kt'`
```
BenHenning, line 17: A few questions:
1. Should we only be performing this check when running pre-push hooks? I think people may also use this to test their files prior to committing them.
2. Does this work both for files from several commits ago & files staged for commit?

BenHenning (commented):
Thanks @Srikanth-Kb! This is a nice idea, though I have a few questions. PTAL.

scripts/ktlint_lint_check.sh
srikanthkb: 1. Yes. It is ideal to perform these checks only while running pre-push hooks. With the current change, users might not be able to check their `.kt` files before committing them. However, we can add an option **-c** to `ktlint_lint_check.sh`, and we can use this option in `pre-push.sh` while calling `ktlint_lint_check.sh` script to check only for committed files. 
When no arguments are specified, then we can run the ktlinter for all modules in our project.
What do you think ?

2. The command checks for files with changes between the current commit(HEAD) and the immediate previous commit(HEAD^). So, this does not work for files from several commits ago. 
And, this does not list the files staged for commit. It only lists out the files committed already.

anandwana001: For point 1, we need to think of
1. Complexity for the contributor to understand these steps.
2. How important is this and what percentage of people will run the script just to check the file formatting?

BenHenning: Thoughts:
1. Can we instead make the argument just for the pre-push hook? The default behavior should just lint the whole codebsae.
2. We should be designing this to compare against all files committed relative to develop so that it can work in a pre-push context

srikanthkb: @anandwana001 
1. For users, I think it will not make a difference. Only in the pre-push script, we can pass an option '-c' to make sure to check committed files only (which the users need not worry about). For regular users, when they run `ktlint_lint_check.sh` (without any additional arguments), it will proceed to check all the files in project (which is same as the existing behavior)
2. This could help us while working on projects that involve changes in a lot of files (which are not staged for commit). However, I think contributors might not always run the `ktlint_lint_check.sh` script separately, as they are notified of any lint errors by the `pre-push` script.

srikanthkb: @BenHenning 
1. Yes, this is what I meant to convey here https://github.com/oppia/oppia-android/pull/2725#discussion_r577342454. The default behavior will not be affected in this case.
2. I am not sure if this is achievable using git filters and commands. I will investigate more on this to check if it is possible to make this work in a pre-push context.

anandwana001:
@Srikanth-Kb any update on this?

srikanthkb:
Closing this PR, as this approach works in a pre-commit context (not in pre-push context). 
For a pre-push script, more investigation is required on a different approach to achieve the objective.

Issue #2684, opened 2021-02-10
Finish community profile


BenHenning:
This came up in a discussion within #2682. Our [community page](https://github.com/oppia/oppia-android/community) is currently missing the following (checkboxes indicate whether it's been fixed):
- [ ] README (this is intermittently missing)
- [ ] Code of conduct (see https://github.com/oppia/oppia-android/pull/2770#issuecomment-784829036 for plan)
- [ ] Contributing guidelines
- [ ] Pull request template
- [x] Accepting content reports

The problem is that we actually have the [first four](https://github.com/oppia/oppia-android/tree/develop/.github) in our .github folder. This seems to be a GitHub bug, so I'm going to follow up with a report. It seems the code of conduct [must exactly match](https://github.community/t/code-of-conduct-is-present-in-repo-but-insights-community-tab-doesnt-detect-it/1534/7) the text of the corresponding covenant selected, and ours is very slightly modified. I might look into reverting these extra changes, but first we need to understand why GitHub seems to be looking in the wrong place for these files.

We should probably enable content reports, so going ahead and turning that on.

This is generally a low-priority effort since it's mainly a means to demonstrate a certain community quality bar (which we seem to meet given that we do have all of the things suggested minus the last one).

BenHenning:
https://github.community/t/community-profile-items-not-recognized/160414 is tracking the issue. It seems GitHub staff is aware of the issue & is looking into it.

anandwana001:
As we have issue template folder and template inside it, can we have a similar folder for pull request too, it might catch in community profile!!
https://github.com/oppia/oppia-android/tree/develop/.github/ISSUE_TEMPLATE

BenHenning:
Ah interesting. Per https://github.blog/2018-01-25-multiple-issue-and-pull-request-templates/ it is possible to create multiple PR templates. Seems worthwhile to try that, at least. Will send out a PR soon.

BenHenning:
@2765 did not solve the issue. :( I suspect, then, that the PR template flake is also due to the other bugs GitHub seems to have around community profile. We'll just need to wait and see if their fixes address our issues.

anandwana001:
https://github.com/oppia/oppia-android/community

Community Page detects all of the things except Pull Request Template, I guess reverting our earlier PR template folder moving might fix this. Looks like this detection working is correcting from the GitHub side.

anandwana001:
@BenHenning  All seems done in the community profile. good to close.

[...]

PR #2770, opened 2021-02-23
Fix part of #2684: Update CODE_OF_CONDUCT.md

BenHenning:
Fix part of #2684.

This is updating the code of conduct to exactly match the template GitHub expects so that it shows up as completed on the community profile. See https://github.community/t/code-of-conduct-is-present-in-repo-but-insights-community-tab-doesnt-detect-it/1534/7 for context.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,9 +1,9 @@
-# Code of Conduct
+# Contributor Covenant Code of Conduct
 
 ## Our Pledge
 
-In the interest of fostering an open and welcoming environment where everyone gets a chance to learn, we as
-contributors and maintainers pledge to making participation in our project, Oppia and
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
 our community a harassment-free experience for everyone, regardless of age, body
 size, disability, ethnicity, sex characteristics, gender identity and expression,
 level of experience, education, socio-economic status, nationality, personal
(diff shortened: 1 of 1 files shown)


BenHenning:
@seanlip sending this your way to get your thoughts since it seems you were the one that customized the code of conduct in the first place. It seems that we are losing some of our tenets here--how strongly do you feel about them? We apparently must follow the exact template in order for GitHub for the community profile. It's a small thing that I don't feel strongly about, but it seems sort of nice to have all the green checkboxes complete (one of which requires making this change).

Please do not consider this urgent.

seanlip (changes requested):
Not worth fixing in this way, IMO. I suggest sending an email to GitHub support to let them know, similar to what was suggested in the linked thread.

For what it's worth, the oppia/oppia repo used to have all the boxes ticked, but I just checked and the ones for README, CODE_OF_CONDUCT and CONTRIBUTING are now unticked. I'm not sure what disqualifies our README from being a README. I think this is more just an issue on GitHub's end, and we shouldn't be changing our docs just to fix that.

seanlip:
Actually, scratch the first paragraph, I misread the thread and only looked at the last few comments.

That said, my thoughts in the second paragraph remain unchanged. How do you feel about it?

seanlip:
OK, successive visits to the oppia/oppia community profile show the following (without me changing anything):

Screenshot from 2021-02-23 20-27-32

Screenshot from 2021-02-23 20-27-43

An earlier visit also only showed the README, code of conduct and contributing being empty, with "pull request template" being ticked. I haven't changed anything between successive visits.

So my opinion is that their community profile page is seriously buggy and we shouldn't put too much stock into it. (For the record, oppia/oppia _does_ have a pull request template.)

BenHenning:
@seanlip this is a known issue: https://github.community/t/community-profile-items-not-recognized/160414 for most of the items, but per https://github.community/t/code-of-conduct-is-present-in-repo-but-insights-community-tab-doesnt-detect-it/1534/2 it seems like GitHub is specifically designed to do an exact match on the code of conduct. Did Oppia web have all the boxes checked after December 2018 (the date of that post)? If so then I'm inclined to believe this is the first issue & am happy to close this, otherwise if we're unsure then I'd expect it to be happy after updating the code of conduct to match.

Also happy with waiting until the first issue is fixed before revisiting this. Any preference?

seanlip:
Pretty sure we did have all the boxes checked the last time I looked at it, but I think that was around end-2018 so it might have been before Dec 2018.

But it's more the principle of the thing, tbh. I don't want to change our docs to satisfy an arbitrary checklist, especially one that is kind of buggy. That seems like gamification working in a negative way. Unless that checklist has actual impact somewhere?

BenHenning:
Hmm, I am unsure how important fulfilling that list is. I couldn't find a satisfactory answer when I was digging into it, so I'm trying it more as a "project health" measure. Unfortunately, the way it appears is it almost seems like we don't have a code of conduct at all which seems wrong to me. I agree with the principle you're mentioning here. I'll close this for now & revisit once the other issue is fixed. If this is still not showing up correctly, I'll raise it with GitHub staff.

PR #2774, opened 2021-02-23
Fix caching workflow check so that develop-based runs can utilize caching

BenHenning:
I noticed this when looking at recent actions runs of the Bazel tests. See: https://github.com/oppia/oppia-android/actions/runs/593329365 (in particular https://github.com/oppia/oppia-android/runs/1963398637?check_suite_focus=true). The tests on ``develop`` are running without caching, but they can & should to improve performance. It seems this is due to the fork conditional check being updated to specifically assume a PR event whereas these are push events.

This PR fixes the check so that if it's not on a fork OR not a push event on develop, then run without caching. This can probably be improved in the future (e.g. to utilize caching on manual dispatch workflows, but to some extent that's not desired since we probably want to run a more "pure" build in such cases).

Reference for the branch-based check: https://github.community/t/github-actions-branch-conditional/16057/5.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -78,7 +78,7 @@ jobs:
           echo build --android_databinding_use_androidx >> $HOME/.bazelrc
       # See explanation in bazel_build_app for how this is installed.
       - name: Install git-secret (non-fork only)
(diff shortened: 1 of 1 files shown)


jcqli (approved):
LGTM

BenHenning:
Thanks!

PR #2668, opened 2021-02-09
Fix #2643: Added BUILD.bazel in accessibility package

FareesHussain:
## Explanation

Fixes #2643

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

Issue #1720, opened 2020-08-26
Add support for Dagger Hilt [Blocked: #59]


BenHenning:
**Is your feature request related to a problem? Please describe.**
The team is currently limited when creating test-only dependencies: either we need to blur the lines between modules by adding dependencies to the test-only testing module, or we need to include test-only dependencies in the production build of the app due to Dagger & manifest requirements.

**Describe the solution you'd like**
Dagger Hilt https://dagger.dev/hilt/ solves this & other problems (like simplifying test application code in tests) by automatically generating components based on what's included in the build graph. This pairs extremely nicely with Bazel, so this issue can be considered blocking on #59.

**Describe alternatives you've considered**
The alternatives would be:
1. Not making this improvement & living with the current situation
2. Rolling out our own code gen
3. Investigate an alternative to Dagger that solves the underlying component issue & doesn't introduce runtime reflection requirements

(2) & (3) would require a potentially enormous amount of work, and (1) doesn't solve the underlying problem.

adhiamboperes:
Finally got around to prototyping this. TLDR; Hilt does work with our versions of Gradle and Bazel.

## Purpose
Create means for reusing dagger modules in tests

## Resources:
https://medium.com/androiddevelopers/dagger-navigation-support-in-android-studio-49aa5d149ec9 
https://developer.android.com/codelabs/android-dagger-to-hilt#0 

## Todo
- Verify that Hilt is supported on our project.
    - Hilt is supported in Android Studio 4.0 or higher. We use Arctic Fox and Bumblebee, which are newer than 4.0.
    - Hilt works with our versions of Bazel(6.5.0) and Gradle(6.8), AGP 4.2.2.
- How to add Hilt to the project: keep Dagger and Hilt working side by side during the migration by keeping the app functional while migrating each Dagger component to Hilt.
   - Add Hilt dependencies https://developer.android.com/training/dependency-injection/hilt-android#setup 
   -  Add Hilt to Bazel
## Migration strategy
- Application or @Singleton component first, and later migrate activities and fragments.
    -   ApplicationComponent first
    -   Annotate all modules with @InstallIn(start with Application level modules and app module modules)
    -   Remove all @Component and @Subcomponent interfaces
    -   After the migration, all Application/Activity/Fragment/View/Service/BroadcastReceiver classes should be annotated with @AndroidEntryPoint and any code instantiating or propagating components should also be removed.
    -   Migrate Test files to Hilt

## Compatibility issues during migration

[...]

PR #2775, opened 2021-02-23
Fix #2643: Introduce canonical Bazel library structure & migrate util accessibility package

### LINKED ISSUE #2643: Create a BUILD file for accessibility/
This task is part of the build system migration from Gradle to Bazel. A prerequesite before working on this issue is that you are able to build Oppia using Bazel. Check out [Oppia Bazel guide](https://github.com/oppia/oppia-android/wiki/Oppia-Bazel-Guide) for Bazel terminology and more of what it does.

Create a BUILD.bazel file in 

`src/main/java/org/oppia/android/util/accessibility/`

This package does not depend on any other packages inside of Oppia itself. Create a library for each file in the package, if possible, and add the dependency artifacts that are required based on the imports in the file.

## Create the libraries

I'm using the filename interchangebly with the library here.

`AndroidAccessibilityManager.kt` should be a private library since it is only used by `AccessibilityModule.kt`.

`CustomAccessibilityManager.kt` is referenced by `AccessibilityModule.kt`, `FakeAccessibilityModule.kt` and externally. It should be a public library and be a deps dependency of the two first libraries.

## Confirm that the package builds

When you've added the libraries, confirm that you can build the package by using Bazel to build the libraries. If you're in the  directory in your terminal, run

`bazel build :all` 

to build all your libraries.

## Replace srcs dependencies in the module BUILD files

Add the required deps dependencies to the module's BUILD.bazel file.

Since `FakeAccessibilityModule.kt` was also used in some tests, we will have to find the corresponding libraries there as well to replace the srcs dependency with the new deps dependency.

`CustomAccessibilityManager.kt` is used in the app module. Look for the libraries that have `//utility` in its deps dependencies and check if it needs the new library you created as a dep dependency.

## Confirm that the app builds

Now you can confirm that the package works by building the app successfuly.

BenHenning:
Fix #2643.

This is a replacement/continuation of #2668.

This PR does a few things:

1. It introduces a canonical structure for laying out packages with both production & test dependencies (see https://github.com/oppia/oppia-android/pull/2668/files#r577336935) using the accessibility package under the utility module as an example (& as part of fixing #2643).
2. It introduces new top-level package groups in the root BUILD.bazel to simplify visibility management & to provide us with fine tune control (see https://docs.bazel.build/versions/master/be/functions.html#package_group).
3. It refactors the codebase to more consistently switch between prod & test modules for the accessibility checker to ensure dependencies are properly hooked up.
4. It does some other minor cleanups in the new package.
5. It introduces a way to exclude migrated packages (using a glob).

One thing to note: long-term it's expected that modules will only be referenced in test dependencies & the root app binary targets. However, at the moment production modules need to be included in the main ``app`` library due to being tightly coupled with the Dagger component classes. This will continue to be the case until #1720 is finished (since then build graph inclusion of a module predicates whether it's included in the generated component).


BenHenning:
This replaces #2668.

BenHenning:
NB: I am expecting some Bazel failures, so they will need to be fixed in a follow-up commit once the run finishes.

rt4914 (approved):
Approving for files for which I am the code-owner. Thanks.

utility/BUILD.bazel
```diff
@@ -7,10 +7,15 @@ load("@dagger//:workspace_defs.bzl", "dagger_rules")
 load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_android_library")
 load("//utility:utility_test.bzl", "utility_test")
 
+MIGRATED_PROD_FILES = glob(["src/main/java/org/oppia/android/util/android/**/*.kt"])
```
FareesHussain: We need to exclude the files inside [accessibility](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/accessibility)

I guess you need to add a kdoc here similar to app/BUILD.bazel.

```suggestion
MIGRATED_PROD_FILES = glob(["src/main/java/org/oppia/android/util/accessibility/**/*.kt"])
```

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityCheckerImpl.kt
```diff
@@ -0,0 +1,18 @@
...
+class AccessibilityCheckerImpl @Inject constructor(
+  private val context: Context
+) : AccessibilityChecker {
+  private val accessibilityManager by lazy {
+    context.getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager
+  }
+
+  override fun isScreenReaderEnabled(): Boolean {
+    return accessibilityManager.isEnabled
+  }
```
FareesHussain, line 17: I guess you've missed adding the Kdocs for these two functions.

BUILD.bazel
```diff
@@ -1,4 +1,50 @@
 # TODO(#1532): Rename file to 'BUILD' post-Gradle.
+
+# Corresponds to being accessible to all Oppia targets. This should be used for production APIs &
+# modules that may be used both in production targets and in tests.
+package_group(
+    name = "oppia_api_visibility",
+    packages = [
+        "//...",
```
FareesHussain, line 8: Isn't this similar to `oppia_binary_visibility` below?

if not what packages does each cover, when do we need to use `oppia_api_visibility` & `oppia_binary_visibility`

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
```diff
@@ -0,0 +1,66 @@
...
+    ],
+    visibility = ["//:oppia_api_visibility"],
+)
+
+kt_android_library(
+    name = "impl",
+    srcs = [
+        ":AccessibilityCheckerImpl.kt",
+    ],
+    visibility = ["//visibility:private"],
```
FareesHussain: Do we need to specify this as it is private by default?

FareesHussain (changes requested):
@BenHenning Added some suggestions and comments, PTAL.

utility/BUILD.bazel
BenHenning: Ah. I'm not sure how I missed this--thanks! Will update it locally a bit later and make sure everything's working as expected.

BUILD.bazel
BenHenning: Hmm I tried making this clear in the documentation, but maybe it could use a bit more work. We'll generally be using API visibility unless something should only be accessible to the binary. I split them out to make the visibility groups themselves more explicit & to provide more control.

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
BenHenning: Nope--good point. I'll remove this when I address the other code-based changes.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityCheckerImpl.kt
BenHenning: I see just one function: ``isScreenReaderEnabled``. Am I maybe missing something?

Also, per https://developer.android.com/kotlin/style-guide#exception_overrides KDocs aren't needed on overrides. In general, we only add them in places where they add value, and method KDocs should always be describing the API introduced by a method. This fits better in the interface or superclass as that's outlining the API implementations need to follow.

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
BenHenning: Done.

utility/BUILD.bazel
BenHenning: Done.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityCheckerImpl.kt
FareesHussain: Got it

FareesHussain (approved):
LGTM, thanks.

anandwana001 (approved):
Approving for files for which I am the code-owner. Thanks.

BUILD.bazel
```diff
@@ -1,4 +1,50 @@
 # TODO(#1532): Rename file to 'BUILD' post-Gradle.
```
fsharpasharp, line 1: Will we be replacing all instances of //visibility:public with one of these package groups?

fsharpasharp (approved):

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+    srcs = [
+        ":AccessibilityChecker.kt",
+    ],
+    visibility = ["//:oppia_api_visibility"],
+)
+
+kt_android_library(
+    name = "impl",
+    srcs = [
+        ":AccessibilityCheckerImpl.kt",
```
FareesHussain: Why use `:<file_name>` here 
https://github.com/oppia/oppia-android/blob/72bc8970ad2aa843109ffc5023e71bbc6181f5ea/app/src/main/java/org/oppia/android/app/viewmodel/BUILD.bazel#L21

Here I've used only file name

app/BUILD.bazel
```diff
@@ -533,6 +532,7 @@ kt_android_library(
         "//third_party:androidx_databinding_databinding-common",
         "//third_party:androidx_databinding_databinding-runtime",
         "//third_party:circularimageview_circular_image_view",
+        "//utility/src/main/java/org/oppia/android/util/accessibility",
```
FareesHussain, line 535: I've missed this earlier.

Why are we adding the deps in app/BUILD.bazel 
and why not utility/BUILD.bazel

BenHenning: Because this is where the dependency is needed. It's not used by anything in utility, and to some extent, that's expected since utility is a collection of utility libraries to be used elsewhere in the codebase.

BUILD.bazel
BenHenning: No, I don't think so. We may want to replace all instances with _a_ package group, but the ones here are specifically corresponding to the types of libraries introduced in this PR.

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
BenHenning: Mostly auto-complete. I believe they're equivalent, but it's less code-smell to omit the ':'. Will do that in a follow-up commit.

app/BUILD.bazel
FareesHussain: Does that mean if there are any packages, which are directly used in the app module then we can directly add this in the app/BUILD.bazel?

but I think it would be better if we can pass it through utility/BUILD.bazel

BenHenning: Yes, that's how it should be. We're planning to get rid of all the top-level BUILD files--libraries should always depend as directly on the thing they need as possible. Occasionally we may introduce libraries that export parts of their implementation (which can be useful when versioning APIs), but that's something that would come after everything is modularized.

Happy to chat about this further, and I'm sure @fsharpasharp can provide some insight here, too.

utility/src/main/java/org/oppia/android/util/accessibility/BUILD.bazel
BenHenning: Done.

BenHenning:
I think all of your comments are now addressed @FareesHussain. PTAL & resolve if they are satisfactorily addressed.

FareesHussain (approved):
LGTM, thanks.

BenHenning:
Thanks! Looking at failures.

BenHenning:
Hmm, Bazel failures seem more of the incorrect affected targets computation. Gonna see if I can repro locally & actually fix the script.

Gradle failure seems like a weird kapt flake I haven't seen before.

BenHenning:
Re-running tests to unblock down-stream work. Would like to try and investigate that Bazel issue, though. It's been eluding me...

BenHenning:
Aha, it appears the Gradle failures are real. Have some fixes to do.

Issue #2743, opened 2021-02-18
[A11y] Add labels for different screens for Talkback reader


rt4914:
When `Talkback` is on, whenever we click on any option which leads to new screen the reader will automatically read the `label` associated with that screen first. If in case label in not present then it will read app name i.e. `Oppia`.

We need correct labels for following screens:
- [ ] [Onboarding Flow]() → `Onboarding Flow`
- [x] [Profile Chooser Screen](https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/18e69208-8528-424a-a368-b2c5d530d235/specs/) → `Profile selection page`
- [ ] [Admin Add Pin](https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/c3809dea-7afa-4905-9a93-ea4273459a68/specs/) → `Authorize to add profiles`
- [ ] [Admin Auth (when clicked on Add Profiles)](https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/392e9d89-db04-4475-9281-d0b24e7519f5/specs/) → `Authorize to add profiles`
- [ ] [Admin Auth (when clicked on Admin Controls)]() → `Authorize to access Administrator Controls`
- [ ] [Pin Verification]() → `PIN verification`
- [ ] [Add Profile](https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/176b4831-2c35-46b3-bf7f-7ee95469aa8e/specs/) →`Add Profile`
- [x] [Topic Page]() → `Topic Page`
- [ ] [Story Screen / Chapter List]() → `Chapter List`
- [ ] [Revision Card]() → `Skill Page`
- [ ] [Question Player]() → `Practice Mode` 
- [x] [Concept Card]() → `Concept Card` 
- [ ] [Exploration Player](https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/screen/40d6cf5f-f673-42c8-af15-035f711ef6b9/specs/) → `Exploration Player` 
- [x] [Hints & Solution Screen]() → `Hints and Solution` 
- [ ] [Profile Progress]() → `Profile`
- [ ] [Topics in Progress]() → `Topics in Progress` 
- [ ] [Stories Completed]() →`Stories Completed`
- [ ] [Options Screen]() → `Options`
[...]

PR #2765, opened 2021-02-23
Maybe fix part of #2684: Move PR template to subdirectory

BenHenning:
Maybe fix part of #2684 by moving the PR template to its own subdirectory to have parity with issue templates (per @anandwana001's suggestion [here](https://github.com/oppia/oppia-android/issues/2684#issuecomment-777400643)).

PR #2781, opened 2021-02-25
Fixes part of #2743: Label for Recently Played Activity A11y

rt4914:
Fixes part of #2743 

I checked the app and found that the title inside the `Recently Played Activity` is fixed and therefore just creating this PR.

Should we make this header dynamic similar to the text that we have in `Home Screen` to the side of `View All`. The possible options for the header are: `Stories for you` and `Recently played stories`. 

Reference: https://docs.google.com/document/d/1o7hOz1rP-hJaKn1f5x6LV6ms0bzhyCpsTifJQY5crCc/edit?disco=AAAALj4Y6X8

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -41,6 +41,7 @@
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.home.recentlyplayed.RecentlyPlayedActivity"
+      android:label="@string/recently_played_activity_label"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.mydownloads.MyDownloadsActivity"
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -429,4 +429,5 @@
   <string name="coming_soon">Coming Soon</string>
   <string name="recommended_stories">Recommended Stories</string>
   <string name="stories_for_you">Stories For You</string>
+  <string name="recently_played_activity_label">Recently played stories</string>
 </resources>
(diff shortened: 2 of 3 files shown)


rt4914:
The pull request template is not getting used by default now. cc: @BenHenning

app/src/main/res/values/strings.xml
```diff
@@ -429,4 +429,5 @@
   <string name="coming_soon">Coming Soon</string>
   <string name="recommended_stories">Recommended Stories</string>
   <string name="stories_for_you">Stories For You</string>
+  <string name="recently_played_activity_label">Recently played stories</string>
```
BenHenning, line 432: Will we be using ``title`` instead of ``label`` per the discussion from the team meeting this week?

app/src/main/AndroidManifest.xml
```diff
@@ -41,6 +41,7 @@
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.home.recentlyplayed.RecentlyPlayedActivity"
+      android:label="@string/recently_played_activity_label"
```
BenHenning, line 44: Should we also be using this for the toolbar to ensure the two are kept in sync?

BenHenning (commented):
Thanks @rt4914. Left some thoughts.

BenHenning:
Thanks for confirming @rt4914. Was waiting for someone to test it out. Will roll back #2765.

rt4914:
@BenHenning I will update the PR meanwhile please have a look at the PR description and reply on the document.

BenHenning:
Responded to the comment. Assigning back to you until this is ready for review.

PR #2572, opened 2021-01-28
Fix #2277, #682: Introduce utility for framework agnostic activity rotation

BenHenning:
Fix #2277
Fix #682

This PR replaces our existing action for managing configuration changes with a new utility that can perform proper activity rotations with both configuration changes & activity recreation for both Robolectric & Espresso tests.

Some specifics:
- This uses RuntimeEnvironment.setQualifiers() to alter the orientation + an activity recreation (via ActivityScenario) to trigger the destroy/recreate flows
- This introduces a new injector/provider to allow us to use a very nice ActivityScenario extension syntax (versus needing to inject the rotator utility & pass the scenario into the methods)

PR #2783, opened 2021-02-25
Fixes #2785 & Fixes part of #2743: A11y activity label added

### LINKED ISSUE #2785: Optimise test case and use ActivityScenarioRule in AdminControls and AppVersion
Optimise test case and use ActivityScenarioRule in AdministratorControlsActivity and AppVersionActivityTest

Advantages of using `ActivityScenarioRule` is that we don't need to launch the activity in each test case and thus reducing one level of nesting as well as reducing the code.

rt4914:
Fixes #2785 

While adding the new test cases Ben mentioned that we are using `activityTestRule` as well as `ActivityScenario` for test cases. So it would be nice if somehow we can use only of such thing and keep consistency. I tried using the `ActivityScenarioRule` and it works correctly in `AdministratorControlsActivity` and `AppVersionActivity`

Fixes part of #2743 

This PR adds label for `AdministratorControlsActivity` and `AppVersionActivity`.

Along with that there are some minor changes in this PR to improve the code.

## Output

https://user-images.githubusercontent.com/9396084/109181484-f8156300-77b1-11eb-9f47-b9b46a2c74d6.mp4

## Note for reviewers:
In general while working on #2743 I will need to add test cases across almost all activities in `app` module and therefore I am thinking of optimising and make more consistent test cases. Just majorly this and upcoming PRs will be focused on three things:
- Add label to activity and test case.
- Use `ActivityScenarioRule` in place of `activityTestRule` and `ActivityScenario`.
- General test case optimisation and consistency.

@BenHenning and @anandwana001 WDYT about the use of `ActivityScenarioRule` usage?

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,11 +19,11 @@
     <meta-data android:name="expiration_date" android:value="2020-09-01" />
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
-      android:label="@string/administrator_controls_activity_label"
+      android:label="@string/administrator_controls_activity_title"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.administratorcontrols.appversion.AppVersionActivity"
-      android:label="@string/app_version_activity_label"
+      android:label="@string/app_version_activity_title"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.completedstorylist.CompletedStoryListActivity"
@@ -86,7 +86,7 @@
       android:windowSoftInputMode="adjustResize" />
     <activity
       android:name=".app.profile.ProfileChooserActivity"
-      android:label="@string/profile_chooser_activity_label"
+      android:label="@string/profile_chooser_activity_title"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.profileprogress.ProfilePictureActivity"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,9 +19,11 @@
     <meta-data android:name="expiration_date" android:value="2020-09-01" />
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
+      android:label="@string/administrator_controls_activity_label"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.administratorcontrols.appversion.AppVersionActivity"
+      android:label="@string/app_version_activity_label"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.completedstorylist.CompletedStoryListActivity"
--- a/app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
@@ -11,8 +11,6 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.core.widget.NestedScrollView
 import androidx.drawerlayout.widget.DrawerLayout
 import androidx.recyclerview.widget.RecyclerView
-import androidx.test.core.app.ActivityScenario
-import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.PerformException
@@ -34,8 +32,8 @@ import androidx.test.espresso.matcher.ViewMatchers.withContentDescription
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.espresso.util.HumanReadables
+import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
-import androidx.test.rule.ActivityTestRule
 import com.google.common.truth.Truth.assertThat
 import dagger.Component
 import org.hamcrest.Matchers
@@ -107,11 +105,8 @@ import javax.inject.Singleton
 class AdministratorControlsActivityTest {
 
   @get:Rule
-  val activityTestRule: ActivityTestRule<AdministratorControlsActivity> = ActivityTestRule(
-    AdministratorControlsActivity::class.java,
-    /* initialTouchMode= */ true,
-    /* launchActivity= */ false
-  )
+  val activityScenarioRule: ActivityScenarioRule<AdministratorControlsActivity> =
+    ActivityScenarioRule(createAdministratorControlsActivityIntent(profileId = 0))
 
   @Inject
   lateinit var profileTestHelper: ProfileTestHelper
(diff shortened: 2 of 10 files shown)


app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -103,6 +106,13 @@ import javax.inject.Singleton
 )
 class AdministratorControlsActivityTest {
 
+  @get:Rule
+  val activityTestRule: ActivityTestRule<AdministratorControlsActivity> = ActivityTestRule(
```
rt4914: Try using `ActivityScenarioRule`

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
```diff
@@ -85,6 +82,11 @@ import javax.inject.Singleton
 )
 class AppVersionActivityTest {
 
+  @get:Rule
+  val activityTestRule: ActivityTestRule<AppVersionActivity> = ActivityTestRule(
+    AppVersionActivity::class.java, /* initialTouchMode= */ true, /* launchActivity= */ false
+  )
+
   @Inject
   lateinit var context: Context
```
rt4914, line 90: Looks better.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
```diff
@@ -85,6 +82,10 @@ import javax.inject.Singleton
 )
 class AppVersionActivityTest {
 
+  @get:Rule
+  val activityScenarioRule: ActivityScenarioRule<AppVersionActivity> =
+    ActivityScenarioRule(createAppVersionActivityIntent())
```
anandwana001, line 87: launch function internally using ActivityScenario and calling ` scenario.launchInternal();`. 

Currently, we use ActivityScenario and this implementation is of ActivityScenarioRule, need to check the difference as I guess there is a tiny difference between them that,
ActivityScenario doesn't clean up the state after test ends but ActivityScenarioRule does it. 

Also, in the `ActivityScenarioRule`, we cannot do any kind of setup before launching the activity, that's something we can do with `ActivityScenario`. 

So if wee move with `ActivityScenario` and don't use `.use{}`, then we need something like

```kotlin
@After
fun cleanup() {
    scenario.close()
}
```

Need some expert points here @BenHenning

anandwana001 (changes requested):

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
BenHenning: We talked about this during the team meeting today @anandwana001. I suggest looking over the team notes; there were some bits in there that were recorded from the discussion I think.

@rt4914 I'm curious about one thing: does scenario rule let us handle configuration changes correctly? I was under the impression that the scenario gets recreated when the activity does (since scenarios treat DESTROYED as a terminal state). Can you confirm?

app/src/main/AndroidManifest.xml
```diff
@@ -84,7 +86,7 @@
       android:windowSoftInputMode="adjustResize" />
     <activity
       android:name=".app.profile.ProfileChooserActivity"
-      android:label="@string/profile_chooser_activity_label"
+      android:label="@string/profile_chooser_activity_title"
```
BenHenning, line 89: Similar to earlier PR: here & elsewhere, should these be synced to each activity's title?

BenHenning (commented):
Thanks @rt4914. Left some thoughts.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
rt4914: I have made two new commits.
Commit https://github.com/oppia/oppia-android/pull/2783/commits/9d77057fed5b97e5cd99cb7c8d223d76c8305cdd shows that config change can easily be done with `ActivityScenarioRule`.

Commit https://github.com/oppia/oppia-android/pull/2783/commits/d9fd2989c7e4e420d303db1aa31bd35d1a7c8446 shows that the code which we need before starting the activity also works correctly with `ActivityScenarioRule`.

Overall there is an **advantage** of using `ActivityScenarioRule`: 
(a.) It can replace both `ActivityTestRule` and `ActivityScenario` and can provide consistency to all tests.
(b.) ActivityScenario does't clean up device state automatically and may leave the activity keep running after the test finishes. Call close() in your test to clean up the state or use try-with-resources statement. This is optional but highly recommended to improve the stability of your tests. Also, consider using ActivityScenarioRule. [Reference](https://developer.android.com/reference/androidx/test/core/app/ActivityScenario)

**Limitation** with `ActivityScenarioRule`:
**It does not allow us to change intent inside tests which we normally does for various test cases.**

**Solution** is to use `LazeActivityScenarioRule` which can allow us to change the intent for each test case. [Reference](https://medium.com/stepstone-tech/better-tests-with-androidxs-activityscenario-in-kotlin-part-1-6a6376b713ea)

**Open Questions**
If we use `LazeActivityScenarioRule` can this create some other type of issue?
WDYT we should do finally?

Also, I highly recommend reading [this blog](https://medium.com/stepstone-tech/better-tests-with-androidxs-activityscenario-in-kotlin-part-1-6a6376b713ea) as it summaries the difference between `ActivityTestRule`, `ActivityScenario` and `ActivityScenarioRule` nicely.

rt4914:
@anandwana001 and @BenHenning PTAL to my reply above.

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
```diff
@@ -185,23 +191,17 @@ class RecentlyPlayedFragmentTest {
       profileId,
       timestampOlderThanOneWeek = true
     )
-    ActivityScenario.launch<RecentlyPlayedActivity>(
```
anandwana001, line 188: aren't we removing other such launches in this class?
looks like the activity launches two times.

anandwana001 (changes requested):
Screenshot 2021-03-01 at 09 41 38
Screenshot 2021-03-01 at 09 40 32
Screenshot 2021-03-01 at 09 39 22
Screenshot 2021-03-01 at 09 36 07

@rt4914 putting espresso test result for reference. 
Nit comment left

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
BenHenning: Wouldn't that be wrong? I'd expect Espresso/Robolectric tests to only ever verify a single activity.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
BenHenning: Please keep in mind we can't just copy LazyActivityScenarioRule--we would need to create our own rule that effectively does the same thing as ActivityScenarioRule but lets us customize the intent. I think the need here is pretty clear, but we need to define specifically what the desired API is. I suspect defaulting the intent is actually desirable for most situations, and having per-test overriding seems reasonable. However, it's not straightforward to implement that without accidentally opening multiple activities. I think we would need to start a draft PR introducing such a utility & iterate on the design a bit there to figure out something that would work for all cases. I do not recommend referencing the one from the blog--its existence introduces the idea that we can try to create on our own, but we need to be more careful about overly depending on external code that's not pre-packaged into a library.

Also, I'm not sure my original question regarding configuration changes was answered. See #2572 for specifics here--the perform orientation change doesn't actually force configuration changes in all environments (e.g. Robolectric). To answer this question, I think we need to check a test that passes in #2572 for both Espresso & Robolectric with the new utility, but also using ActivityScenarioRule.

BenHenning (commented):
Thanks @rt4914. Left some thoughts.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
rt4914: @BenHenning Considering that this requires more research work now, can I pass in to @anandwana001 and focus back on A11y PRs. WDYT?

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
rt4914: We should update all tests. But this was just for example for one test.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
BenHenning: Yup, sgtm @rt4914.

rt4914: Thanks.

rt4914:
Closing this as it has been filed as good-first-issue #3062

PR #2786, opened 2021-02-26
Revert "Maybe fix part of #2684: Move PR template to subdirectory"

BenHenning:
Reverts oppia/oppia-android#2765

See https://github.com/oppia/oppia-android/pull/2781#issuecomment-785813188. It seems this directory structure is not actually used correctly by GitHub. It appears that per https://github.community/t/multiple-pull-request-template/874/10 the old way we had it is the only way to set the default template, so that solution is actually just wrong.


BenHenning:
@FareesHussain you were definitely right in https://github.com/oppia/oppia-android/pull/2765#issuecomment-784395825 to want to test this. :) I don't think GitHub's documentation clearly defined subdirectory behavior, and I missed this.

FareesHussain (approved):
LGTM, thanks.

BenHenning:
@seanlip can you PTAL at this since it's currently preventing our PR template from being used in new PRs?

seanlip (approved):

BenHenning:
Thanks!

PR #2525, opened 2021-01-20
Fixes Part of #2326: Changes in Views inside Practice Tab [A11y] 

Mehul-Bisht:
## Explanation

Fixes part of #2326: Changes in Views inside Practice Tab [A11y]
Fixed issue with checkbox accessibility for all layouts

## Portrait Mode
| Before | After |
:-------------------------:|:-------------------------:
portrait_before | portrait

## Landscape Mode
| Status |
:-------------------------:
## Before changes
land_before 
## After Changes
land_top_after
land_bottom

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

Issue #2326, opened 2020-12-26
Changes in Views inside Practice Tab [A11y]


Arjupta:
**Describe the bug**

Views inside the Practice Tab are not fully compatible with some users from the Accessibility point of view.
![Screenshot_2020-12-17-11-27-30-51_80a9469f53bfe75dbc6e8989be28e2f0](https://user-images.githubusercontent.com/64064110/103155329-e705a100-47c4-11eb-9af2-2bc829d22c79.jpg)
Issue 1 - The contrast ratio of the topic practice start button is not correct in the inactive state.
Issue 2 - The subtopic checkbox has a height of 24dp which is less than the standard one (48dp)

**To Reproduce**
Steps to reproduce the behaviour:
1. Start the Accessibility Scanner App on your device
2. Open the Oppia Android App and start Fraction Exploration
3. Switch to the Practice Tab
3. Scan the screenshots with the Accessibility Scanner
4. You will see the tests that are not passed by that screen

**Expected behaviour**
All the tests must be passed by scanning the screenshots for both Tablet and Mobile Devices in every orientation.

**Screenshots**
|Issue 1|Issue 2|
|:-:|:-:|
|![Screenshot_2020-12-17-11-27-02-40_80a9469f53bfe75dbc6e8989be28e2f0](https://user-images.githubusercontent.com/64064110/103155332-e836ce00-47c4-11eb-9150-f18329072e55.jpg)|![Screenshot_2020-12-17-11-26-53-15_80a9469f53bfe75dbc6e8989be28e2f0](https://user-images.githubusercontent.com/64064110/103155360-31871d80-47c5-11eb-89c9-6647beae3a6d.jpg)

**Device**
- Real Device - RealMe3
- SDK version - 29

**Additional context**
The Screenshots are descriptive enough for understanding the issues but still, if there is a need of clarification then you can comment here. Multiple Pull Requests can also be made for solving these issues.

Certain Views does not have the correct contrast ratios (Inactive state) as mentioned in the screenshots. To solve this, one must find a way to make the views disabled instead of being in non-clickable state only

Reference - [https://github.com/oppia/oppia-android/wiki/Working-on-UI#accessibility-scanner](https://github.com/oppia/oppia-android/wiki/Working-on-UI#accessibility-scanner)

Mehul-Bisht:
I was able to fix the issue with the checkboxes, Can I be assigned for this issue?

FareesHussain:
[...]

PR #2797, opened 2021-03-01
Fixes part of #2326: Changes in Views inside Practice Tab [A11y]

Mehul-Bisht:
## Explanation

Fixes part of #2326: Changes in Views inside Practice Tab [A11y] for Tablet layouts
Fixed issue with checkbox accessibility for all tablet layouts, I had already fixed the issue with mobile layouts in earlier PR here: #2525 

## Portrait Mode
| Before |
:-------------------------:
practice_tab_portrait_before
| After |
practice_tab_portrait_after

## Landscape Mode
| Before |
:-------------------------:
practice_tab_landscape_before
| After |
practice_tab_landscape_after


Mehul-Bisht:
@ARJUPTA @rt4914 kindly check the changes

rt4914:
@Mehul-Bisht Has this change been suggestion or discussed with Chantel?

From what I see this PR introduces 2 dp change which means UI will slightly change which we general don't want.
If this has not been confirmed with Chantel, then i suggest either look for alternate solution of discuss this with Chantel whether this change is allowed or not.

Mehul-Bisht:
> @Mehul-Bisht Has this change been suggestion or discussed with Chantel?

No it was not discussed with her. @mschanteltc Kindly tell if I should be adding 2 dp padding to the checkboxes in tablet layout to remove the accessibility issue mentioned in the second issue (Issue 2) here: #2326

mschanteltc:
> > @Mehul-Bisht Has this change been suggestion or discussed with Chantel?
> 
> No it was not discussed with her. @mschanteltc Kindly tell if I should be adding 2 dp padding to the checkboxes in tablet layout to remove the accessibility issue mentioned in the second issue (Issue 2) here: #2326

The item's height should be a multiple of 4. The accessibility checker used here recommends a size of at least 48 dp, although the UI [guidelines](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) I suggest we follow recommends that touch targets should be at least 44 CSS pixels. When you convert the item size from dp to px, does this meet this latter requirement?

Mehul-Bisht:
>I suggest we follow recommends that touch targets should be at least 44 CSS pixels. When you convert the item size from dp to px, does this meet this latter requirement?

@mschanteltc Although I am not sure about this, but I found this conversion [here](https://graphicdesign.stackexchange.com/a/55107) 
image

As per this, 48 dp equals 28.8 px

Mehul-Bisht:
And [this answer](https://stackoverflow.com/a/31715400/13343177) states that dp and px can be used interchangeably. Chantel if you are aware about the exact conversion, let me know.

Mehul-Bisht:
@mschanteltc Any updates regarding this?

mschanteltc:
> @mschanteltc Any updates regarding this?

I still say let's only use 44 dp for the checkbox and ignore the 48 dp minimum requirement. As long as it is following the W3 [guidelines](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html), this should suffice. Also is it possible we can center horizontally center the touch target with the checkbox and skill name?

Mehul-Bisht:
>I still say let's only use 44 dp for the checkbox and ignore the 48 dp minimum requirement. As long as it is following the W3 guidelines

@mschanteltc the initial value of the height was 46 dp, so should I change it to 44 dp then? Also, the UI will shift if we make it horizontally centred, should I still be considering it?

mschanteltc:
> > I still say let's only use 44 dp for the checkbox and ignore the 48 dp minimum requirement. As long as it is following the W3 guidelines
> 
> @mschanteltc the initial value of the height was 46 dp, so should I change it to 44 dp then? Also, the UI will shift if we make it horizontally centred, should I still be considering it?

Yes change to 44dp, since I assume changing it to 48dp means that the checkboxes would be more spaced out from each other. 

Can you show me screenshots of how the page looks like when the checkboxes are top-aligned with the touch target and horizontally-centered with the touch target? We can choose from there.

Mehul-Bisht:
This is the screenshot after changing to 44dp 

| Reverted to 44 dp | 
:----------------|
practice_44dp

And since the text is not a separate part than the tap target (the checkbox has text within it), so this is how it will look if I change the width of the checkbox to "wrap_content" instead of "match_parent" 

| checkbox width | 
:----------------|
practice_wrap

Mehul-Bisht:
> > > I still say let's only use 44 dp for the checkbox and ignore the 48 dp minimum requirement. As long as it is following the W3 guidelines
> > 
> > 
> > @mschanteltc the initial value of the height was 46 dp, so should I change it to 44 dp then? Also, the UI will shift if we make it horizontally centred, should I still be considering it?
> 
> Yes change to 44dp, since I assume changing it to 48dp means that the checkboxes would be more spaced out from each other.
> 
> Can you show me screenshots of how the page looks like when the checkboxes are top-aligned with the touch target and horizontally-centered with the touch target? We can choose from there.

@mschanteltc  PTAL

mschanteltc:
> > > > I still say let's only use 44 dp for the checkbox and ignore the 48 dp minimum requirement. As long as it is following the W3 guidelines
> > > 
> > > 
> > > @mschanteltc the initial value of the height was 46 dp, so should I change it to 44 dp then? Also, the UI will shift if we make it horizontally centred, should I still be considering it?
> > 
> > 
> > Yes change to 44dp, since I assume changing it to 48dp means that the checkboxes would be more spaced out from each other.
> > Can you show me screenshots of how the page looks like when the checkboxes are top-aligned with the touch target and horizontally-centered with the touch target? We can choose from there.
> 
> @mschanteltc PTAL

Sorry I believe I used the wrong terminology. I think I meant that the checkbox and skill name should be vertically centered in respect to the touch target, like in the one below:
TP (Practice, Unchecked)

Mehul-Bisht:
By Shifting half of the padding above the checkbox i was able to achieve this

practice_tab_Center

Mehul-Bisht:
@mschanteltc kindly check the changes and let me know if there is still some issue.

mschanteltc:
> By Shifting half of the padding above the checkbox i was able to achieve this
> 
> practice_tab_Center

LGTM!

Mehul-Bisht:
@rt4914 @ARJUPTA Please review this PR once so that it is ready to merge

rt4914 (changes requested):
@Mehul-Bisht There are two issues with solution.
1. The box and the text is actually not correctly aligned. If you run the app in mobile you will notice hat text and the box are correctly aligned.
2. Also, the box should always be on the top-right corner.




## Mobile Output

Mehul-Bisht:
@rt4914 the reason why this is happening is due to the gravity of checkbox and shifting half of the padding above the checkbox. In mobile layouts the gravity is set to `top` but for tablet layouts I had changed it to `center_vertical` as suggested by Chantel to make the checkbox and the text to be vertically centred. When I revert this change the layout for both mobile and tab layout behaves similar as shown:

## Tablet Layout
| After Change |
:---------------:
practice_tab_gravity

## Mobile Layout
| No Change |
:---------------:
Screenshot_1614864624

anandwana001:
@rt4914 PTAL

rt4914:
@Mehul-Bisht I tried to solve this locally but haven't found any solution yet.

PR #2800, opened 2021-03-02
Fix #2791: Create feedback reporting model layer

### LINKED ISSUE #2791: Create Android model layer objects for feedback reporting
Implement part of #53 
Implements the model layer objects needed for Android feedback reporting.
This includes the proto objects as well as Retrofit data objects and service class that sends HTTP requests.

Design doc: [User Feedback Reporting](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

jcqli:
Fix #2791
This PR creates the Android model layer protobufs for user feedback reporting.

[Design Doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit#)


model/BUILD.bazel
```diff
@@ -206,6 +219,7 @@ android_library(
     visibility = ["//visibility:public"],
     exports = [
         ":arguments_java_proto_lite",
+        ":feedback_reporting_java_proto_lite",
```
BenHenning: Prefer not exporting it--we actually want to add these as direct dependencies wherever they're used rather than having a single library to depend on.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+
+package model;
+
+import "profile.proto";
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Top-level proto to store feedback reports locally when there is no network connection to upload
+// them to the remote service
```
BenHenning: Nit: end doc comment with punctuation for consistency.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+message FeedbackReportingDatabase {
+  repeated FeedbackReport reports_list = 1;
+}
+
+// Corresponds to an entire feedback report with user reported information and collected app,
+// device, and build information.
+message FeedbackReport {
+  // The date the report was created.
```
BenHenning: Is this a unix timestamp? Please be specific in the documentation & name. Perhaps ``creation_timestamp_ms`` might be clearer.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  int64 report_date_ms = 1;
+  FeedbackReportingForm feedback_form = 2;
+  FeedbackReportingDeviceInfo device_info = 3;
+  FeedbackReportingBuildInfo build_info = 4;
+  FeedbackReportingAppInfo app_info = 5;
+}
+
+// Information supplied by the user in the feedback report.
+message FeedbackReportingForm {
```
BenHenning: Would ``UserSuppliedInfo`` maybe be clearer? It's a bit odd to call this 'form' since it's not actually a form, just the information collected from one. To some extent, 'form' is a UI-specific detail regarding where this information originates.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+// A suggestion feedback report.
+message Suggestion {
+  enum SuggestionCategory {
+    CATEGORY_UNSPECIFIED = 0;
+    FEATURE_SUGGESTION = 1;
+    LANGUAGE_SUGGESTION = 2;
+    OTHER_SUGGESTION = 3;
```
BenHenning: Please explain each of these (like you would enum constants in code). Ditto for other enums.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+// Information supplied by the user in the feedback report.
+message FeedbackReportingForm {
+  oneof report_type {
+    Suggestion suggestion = 1;
+    Issue issue = 2;
+    Crash crash = 3;
+  }
+}
+
+// A suggestion feedback report.
```
BenHenning: Or perhaps: ``User-provided suggestion for something that could or should change in the app``. Consider that:
1. Readers likely have much less context than you (which will be true for as long as this code exists)
2. We want to establish 'API contracts' rather than just explain what the current behavior is since this code can & will evolve over time, and we should try to make it clear where certain changes should go

Ditto for other proto messages here.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+message Suggestion {
+  enum SuggestionCategory {
+    CATEGORY_UNSPECIFIED = 0;
+    FEATURE_SUGGESTION = 1;
+    LANGUAGE_SUGGESTION = 2;
+    OTHER_SUGGESTION = 3;
+  }
+  SuggestionCategory suggestion_category = 1;
```
BenHenning, line 60: Although this is slightly self-documenting, please do make sure that here & elsewhere all fields are documented.  Even something like: ``The user-selected category corresponding to contextualize this suggestion.`` is sufficient.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  oneof report_type {
+    Suggestion suggestion = 1;
+    Issue issue = 2;
+    Crash crash = 3;
+  }
+}
+
+// A suggestion feedback report.
+message Suggestion {
+  enum SuggestionCategory {
```
BenHenning, line 49: Please add a comment for this & all other enums/constants. In general, everything declared in a proto file should be documented since all messages/enums are part of the public API being generated.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+    LANGUAGE_TEXT_ISSUE = 4;
+    TOPICS_ISSUE = 5;
+    PROFILE_ISSUE = 6;
+    OTHER_ISSUE = 7;
+  }
+  IssueCategory issue_category = 1;
+  // List of checkbox items selected by user in the feedback report.
+  repeated string issues_list = 2;
+  // The user's text input if they select "other" for any option
+  string other_user_input = 3;
```
BenHenning: Might it make sense to model these as oneof's against the category, instead? E.g. we could embed the exact checkbox options & the 'other' string directly in the structure exactly when it's appropriate. That ensures internal consistency with the proto & may simplify future changes to the proto that keep it interoperable.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  CrashLocation crash_location = 1;
+  // The user's text input for explaining what happened.
+  string crash_explanation = 2;
+}
+
+// Information collected about the user's device.
+message FeedbackReportingDeviceInfo {
+  string package_name = 1;
+  string package_code = 2;
+  string country_locale = 3;
```
BenHenning, line 285: Is this an ISO-3166 country code? We should be specific so that users of the proto understand both how to populate & use values for this field.

Ditto for ``language_locale`` & ``device_language``.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  }
+  CrashLocation crash_location = 1;
+  // The user's text input for explaining what happened.
+  string crash_explanation = 2;
+}
+
+// Information collected about the user's device.
+message FeedbackReportingDeviceInfo {
+  string package_name = 1;
+  string package_code = 2;
```
BenHenning: ``package_version_code`` may be more specific; just 'code' is a bit ambiguous.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+
+// Information collected about the device's build.
+message FeedbackReportingBuildInfo {
+  enum NetworkType {
+    NETWORK_TYPE_UNSPECIFIED = 0;
+    WIFI = 1;
+    CELLULAR = 2;
+  }
+  string device_model = 1;
+  string sdk_version = 2;
```
BenHenning: Should this be an int?

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  string device_brand = 3;
+  string build_fingerprint = 4;
+  string phone_type = 5;
```
BenHenning: What does this correspond to? Can it be an enum or is it an OEM-provided value?

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+  NetworkType network_type = 6;
+  string device_language = 7;
```
BenHenning: How do this, language, & country locales differ?

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Information collected about the app's build.
+message FeedbackReportingAppInfo {
+  enum EntryPoint {
+    ENTRY_POINT_UNSPECIFIED = 0;
+    MAIN_MENU = 2;
```
BenHenning: Suggest ``NAVIGATION_DRAWER`` to be consistent with other names for the menu.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,136 @@
...
+    // No stories have been started in this topic.
+    NOT_STARTED = 1;
+    // The topic has been started but not completed by the player.
+    STARTED_NOT_COMPLETED = 2;
+    // The topic has been completed by the player.
+    COMPLETED = 3;
+  }
+
+  EntryPoint entry_point = 1;
+  // Maps topic ID to the completion state of topic, based on state of stories in topics
```
BenHenning: Suggest also mentioning a bit about why this is here (e.g. to capture the learner's saved lesson state to have additional context).

That being said, maybe 'context' is a bit clearer than 'info' for the protos here. WDYT?

model/src/main/proto/onboarding.proto
```diff
@@ -27,6 +27,7 @@ message AppStartupState {
   // Contains the user's startup state upon opening the app. This may change from app open to app
   // open, but should remain constant throughout the lifetime of a single process.
   StartupMode startup_mode = 1;
+  bool is_from_crash = 2;
```
BenHenning, line 31: Please add documentation comment.

BenHenning (changes requested):
Thanks @jcqli! Overall, I think the new proto structure makes a lot of sense and seems good. I'm being rather nitpicky with my comments about improving naming/documentation/etc. PTAL.

model/src/main/proto/feedback_reporting.proto
jcqli: Could you clarify what you mean by this? Do you mean that the category would be a `oneof` instead of an `enum` and within each category's `oneof` we can list the specific checkbox items?

model/src/main/proto/feedback_reporting.proto
jcqli: Ah this would make sense -- I was thinking about the SDK string but the int would be a more readable.

model/src/main/proto/feedback_reporting.proto
jcqli: I was imagining [this network type](https://developer.android.com/reference/android/telephony/TelephonyDisplayInfo#getNetworkType()) but now I'm not sure this will will provide anything substantial since we don't need the phone network for anything. What are your thoughts?

model/src/main/proto/feedback_reporting.proto
jcqli: Country locale, language locale, and device language can all differ, e.g. French-Canada might have language locale = French, country locale = Canada, and device language = something completely different. The combination of these could help us understand the user better and then better contextualize the report. WDYT?

model/src/main/proto/feedback_reporting.proto
jcqli: Ack, I like the using context because that is the goal of collecting all this information -- to gain understanding of the context in which the report is filed in. Will change the names

model/BUILD.bazel
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done, changed to `UserSuppliedFeedback` let me know what you think.

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done, ack!

model/src/main/proto/feedback_reporting.proto
jcqli: Done, listed enums directly above the field they're used in to make it explicit where they're used and what they represent.

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Yes, done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done, also added `REVISION_CARD` since feedback reporting can be shown there

model/src/main/proto/onboarding.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
BenHenning: Yeah, something like:

```proto
message Issue {
  oneof issue_category {
    LessonQuestionIssue lesson_question_issue = 1;
    LanguageIssue language_issue = 2.
    ....
    OtherIssue other_issue = 3;
  }
}

message LessonQuestionIssue {
  enum IssueOption {
    OPTION_UNSPECIFIED = 0;
    LESSON_TOO_LONG = 1;
    ...
    CANNOT_SEE_HINTS = 5;
    ....
  }
  repeated IssueOption user_selected_options = 1;
}

...

message OtherIssue {
  string open_user_input = 1;
}
```

What do you think about that approach? One nice benefit is it avoids using strings for typing, and would allow for directly translation to backend-storable strings for each checkbox (which is something we need to make sure of--we can't tie actual frontend strings to the options selected otherwise translations or string changes will create inconsistencies in reports over time).

This also has the nice effect of essentially exactly matching the structure the user themselves is controlling, and makes impossible weird situations like seeing user-filled input for a non-other issue impossible to even occur structurally.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Information collected about the user device's systems.
+message FeedbackReportingSystemContext {
+  // The package name for the specific app version on the device.
+  string package_name = 1;
+  // The package code for the specific app version non the device.
+  string package_version_code = 2;
+  // The ISO-3166 code for the country locale that the feedback report is sent from.
+  string country_locale = 3;
```
BenHenning, line 285: How will we get this vs. language_locale? Per https://stackoverflow.com/a/3660245 my understanding is that Android only exposes the phone language.

model/src/main/proto/feedback_reporting.proto
BenHenning: That makes sense, but similar to my question above: can you point to the specific Android APIs used for each so that I can better understand how we're retrieving these values?

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,199 @@
...
+    WIFI = 1;
+    // Indicates that device is connected to a cellular network.
+    CELLULAR = 2;
+  }
+  NetworkType network_type = 6;
+  // The ISO-3166 code for the language the device is set to.
+  string device_language = 7;
+}
+
+// Information collected about the specific Oppia Android app.
```
BenHenning:
```suggestion
// Information collected about the current version of the Oppia Android app.
```

Might be a bit more clear.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,199 @@
...
+  map<string, TopicProgressStatus> topic_progress = 2;
+  // The text size the app is set to.
+  ReadingTextSize text_size = 3;
+  // The text language the app is set to.
+  AppLanguage text_language = 4;
+  // The audio language the app is set to.
+  AudioLanguage audio_language = 5;
+  // The app's specific device settings.
+  DeviceSettings device_settings = 6;
+  // If the account sending the report is an admin account.
```
BenHenning:
```suggestion
  // Whether the account sending the report is the admin account.
```

Something about the current wording confused me.

model/src/main/proto/feedback_reporting.proto
BenHenning: Much clearer--thanks!

model/src/main/proto/feedback_reporting.proto
```diff
@@ -0,0 +1,199 @@
...
+  // The user's text input for explaining what happened.
+  string crash_explanation = 2;
+}
+
+// Information collected about the user device's systems.
+message FeedbackReportingSystemContext {
+  // The package name for the specific app version on the device.
+  string package_name = 1;
+  // The package code for the specific app version non the device.
+  string package_version_code = 2;
```
BenHenning: Ah missed this before: this is also an int.

model/src/main/proto/feedback_reporting.proto
BenHenning: Ah. In that case, this should probably be enum. However, I agree that it's not yet clear how this could be useful. Omitting it seems reasonable to start--we can always add it in the future.

model/src/main/proto/feedback_reporting.proto
BenHenning: Ah, what about for questions, too? I'm wondering if we should just focus on explorations & navigation drawer for now. It's easy to add other entry points in the future. What's the main advantage we get for adding it to the revision card toolbar?

BenHenning (commented):
Thanks @jcqli! Had a few follow-up thoughts.

model/src/main/proto/feedback_reporting.proto
jcqli: Ah I see -- I'll remove it

model/src/main/proto/feedback_reporting.proto
jcqli: I wasn't including the question player since there isn't a three-dot menu in practice player's navbar, but it is available in the revision card (mentioned in [this thread](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?pli=1&disco=AAAALqa4y7k)). If the question player is going to have the menu included in it soon then I agree that focusing on just the exploration player and hamburger menu would be more feasible; but if not then I think it's more sensible to include it in the revision card for homogeneity since it's odd to have the two three-dot menu have different selection options.

model/src/main/proto/feedback_reporting.proto
jcqli: Ah I see what you mean -- I like this but I'm wondering how to incorporate "other" as an IssueOption if we decide to include it as a checkbox option. I received design feedback that it would be best to include a text-box / other option within the non-other categories.

I want to use the enum types for checkbox options either way, but I don't know if we can avoid including a text input for all types.

model/src/main/proto/feedback_reporting.proto
jcqli: [Locale](https://developer.android.com/reference/java/util/Locale.html) can give the [country](https://developer.android.com/reference/java/util/Locale.html#getISOLanguages()) and [language](https://developer.android.com/reference/java/util/Locale.html#getISO3Language()) codes as Strings (not all 3166 codes though, will make that change in the comments). It might be feasible to use an enum for the 3166 codes but that's a lot of codes to enumerate and since the API provides the string it seems reasonable to just pull directly form that return value.

jcqli: Ah I see your below comment regarding this. It looks the locale only gives information about the device's locale not the location locale so this device_language would be the same as language_locale. Will omit one of them.

model/src/main/proto/feedback_reporting.proto
jcqli: Ah darn I didn't realize that. It looks like to get location information will require more device access that I don't think we want to ask for right now (i.e. user has to consent to sharing location data).

I think it's reasonable to omit the country for now, WDYT?

jcqli: Ah darn, I didn't realize that locale was not from location -- I'll keep language_locale and country_locale (these might be different, thought I'm curious how country_locale is determined if not by location), and remove device_language since language_locale is giving the device's language.

model/src/main/proto/feedback_reporting.proto
jcqli: Done, enum types added with a `user_input` in all of them

model/src/main/proto/feedback_reporting.proto
jcqli: Done, removed

model/src/main/proto/feedback_reporting.proto
jcqli: Removed device_language since it's the same as language_locale, keeping country_locale as well though it doesn't appear that this is the device's physical location (this requires location services that we probably don't want to be asking users for, not sure if it'll help right now).

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
-  // The user's text input if they select "other" for any option
-  string other_user_input = 3;
+}
+
+// Represents a user-selected category to indicate there is an issue with a lesson or the questions.
+message LessonQuestionIssue {
+  // The specific issue option that a user can choose to include in their report.
+  enum IssueOption {
+    // An unspecified option that represents an "other" issue option in this category.
+    OTHER_UNSPECIFIED = 0;
```
BenHenning: This seems inconsistent. We use 'OTHER' as a distinct option elsewhere, but not here. I also think it's conventional to call the first enum the name of the enum + unspecified, e.g. 'ISSUE_OPTION_UNSPECIFIED' for consistency broadly across proto enums.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
+  oneof language_issue_category {
+    GeneralLanguageIssue general_language_issue = 1;
+    AudioLanguageIssue audio_language_issue = 2;
+    TextLanguageIssue text_language_issue = 3;
+  }
+}
+
+// Represents the user-selected category for an issue with the app's language without specifying a
+// sub-category of text or audio language. This issue does not have any user-selected options as
+// user's must select a sub-category to provide specific issues.
```
BenHenning:
```suggestion
// users must select a sub-category to provide specific issues.
```

Typo.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
+  enum IssueOption {
+    // An unspecified option that represents an "other" issue option in this category.
+    OTHER_UNSPECIFIED = 0;
+    // Represents the issue option that the text language is too big.
+    TOO_BIG = 1;
+    // Represents the issue option that the text language is too small.
+    TOO_SMALL = 2;
+    // Represents the issue option that the user cannot understand the text language.
+    CANNOT_UNDERSTAND = 3;
+    // Represents the issue option that the language the user expects for the text is incorrect.
```
BenHenning, line 178: Should this actually be TRANSLATION_WRONG? Trying to understand what this means, and that option seems important to include here.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
+    CANNOT_FIND_LANGUAGE = 5;
+  }
+  repeated IssueOption user_selected_options = 1;
+  // The user's text input if they selected "other" as an issue option; this can only be non-empty
+  // if the OTHER_UNSPECIFIED option is selected.
+  string other_user_input = 2;
+}
+
+// Represents the user-selected category for an issue with the topics available to the to the user.
+message TopicsIssue{
```
BenHenning: Typo: space before ``{``

Ditto elsewhere.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
+    // An unspecified option that represents an "other" issue option in this category.
+    OTHER_UNSPECIFIED = 0;
+    // Represents the issue option that the user cannot see all the topics they want to see.
+    CANNOT_SEE_ALL = 1;
+    // Represents the issue option that the user cannot find a desired topic.
+    CANNOT_FIND_TOPIC = 2;
+    // Represents the issue option that the user cannot find a story they accessed earlier.
+    CANNOT_FIND_EARLIER_STORY = 3;
+    // Represents the issue option that the user is unsure of what to do next in the app.
+    DO_NOT_KNOW_NEXT = 4;
```
BenHenning: Maybe: DO_NOT_KNOW_NEXT_STEP?

model/src/main/proto/feedback_reporting.proto
```diff
@@ -65,31 +65,167 @@ message Suggestion {
...
+    // An unspecified option that represents an "other" issue option in this category.
+    OTHER_UNSPECIFIED = 0;
+    // Represents the issue option that the user cannot add a profile.
+    CANNOT_ADD = 1;
+    // Represents the issue option that the user cannot remove a profile.
+    CANNOT_DELETE = 2;
+    // Represents the issue option that the user cannot add a new admin account.
+    CANNOT_ADD_NEW_ADMIN = 3;
+    // Represents the issue option that the user cannot change their profile image.
+    CANNOT_CHANGE_IMAGE = 4;
```
BenHenning: What about an option for changing their name?

model/src/main/proto/feedback_reporting.proto
BenHenning: Ah I see. Should we be using https://developer.android.com/reference/java/util/Locale#getISO3Country() for retrieving country/language, instead? That would then provide us with the 3-digit code, instead.

Also, I think the comments should be clearer about what country/language mean. Is this something you could quickly test out locally or research possible options to get an idea?

model/src/main/proto/feedback_reporting.proto
BenHenning: Makes sense--thanks!

model/src/main/proto/feedback_reporting.proto
BenHenning: Got it, thanks. Continuing the discussion in the thread below.

model/src/main/proto/feedback_reporting.proto
BenHenning: Hmm, the 'other' case for the sub-categories seems to have the same issue. I wonder, what if we also did something like:

```proto
message TopicsIssue {
  oneof issue_type {
    CheckboxSelections checkbox_selections = 1;
    string open_prompt = 2;
  }
  message CheckboxSelections {
    enum CheckboxOptions {
      ...
    }
    repeated CheckboxOptions selections = 1;
  }
}
```

That provides the same level of enforced exclusivity. It does tie us a bit to only selecting checkboxes & not providing any additional prompts, but we could always change this structure in the future & migrate it. WDYT?

BenHenning (commented):
Thanks @jcqli! Just a few more follow-up thoughts.

model/src/main/proto/feedback_reporting.proto
jcqli: My initial instinct was that this is too similar to the having an `OtherIssue` at a higher level but I see how this allows for more specificity within the category. Does that mean the user can only choose "other" _or_ checkboxes, but not both? If so, this seems like we're trying to change the feature to accommodate technically which I don't think is beneficial; I think it would be better to leave `other_user_input` and be extra careful that its only available when the user selects the "other_issue" enum.

model/src/main/proto/feedback_reporting.proto
jcqli: Yup, planning to use ISO3 codes for both language and country. Based on [this SO post](https://stackoverflow.com/questions/14389349/android-get-current-locale-not-default) the locale can be set by the user but does represent the country and language that they choose to use on the device, although not location.

Updated comment to hopefully make this clearer.

model/src/main/proto/feedback_reporting.proto
jcqli: Ack, Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: I think CANNOT_UNDERSTAND would be different -- it might be that the phrasing used in the translation is understandable (e.g. a younger student may have a smaller vocabulary range than an older student). Added TRANSLATION_WRONG as a separate enum type.

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done

model/src/main/proto/feedback_reporting.proto
jcqli: Done, nice catch

model/src/main/proto/feedback_reporting.proto
BenHenning: Seems sensible. Thanks!

model/src/main/proto/feedback_reporting.proto
```diff
@@ -256,12 +272,14 @@ message Crash {
 // Information collected about the user device's systems.
 message FeedbackReportingSystemContext {
   // The package name for the specific app version on the device.
-  string package_name = 1;
+  string package_version_name = 1;
   // The package code for the specific app version on the device.
   int32 package_version_code = 2;
-  // The ISO-639 code for the device country locale.
+  // The ISO-639 code for the country locale, as determined by the system settings. This
```
BenHenning: I think your second sentence here & below is just repeating itself. Can you use an actual example for both to provide clarity? I think that's actually what's missing here.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -288,7 +306,7 @@ message FeedbackReportingDeviceContext {
 }
 
 // Information collected about the current version of the Oppia Android app.
-message FeedbackReportingAppContext {
+message FeedbackReportingOppiaContext {
```
BenHenning: I actually thought 'app' was clearer--'Oppia' could mean any number of things: the app, the web frontend, the backend. the name of the tutor, the team name, the organization, etc.

model/src/main/proto/feedback_reporting.proto
BenHenning: Ah, this was me misunderstanding. I thought the options were exclusive--I didn't realize users could input both open text & the checkboxes. In that case, the current design does seem correct. Thanks!

model/src/main/proto/feedback_reporting.proto
BenHenning: Thanks for clarifying in the comment. Left one follow-up to further clarify the context of the comments.

BenHenning (changes requested):
Thanks @jcqli! The PR LGTM, but I do have a couple of new comments. I'd like to take one more quick glance at the PR once addressed.

model/src/main/proto/feedback_reporting.proto
jcqli: Ack, done

model/src/main/proto/feedback_reporting.proto
jcqli: Done, is this clearer?

jcqli:
@BenHenning PTAL -- FYI a utility module check has been failing, so may need to look at that if it still occurs. Also added TYPO as an issue option that could be frequent.

model/src/main/proto/feedback_reporting.proto
```diff
@@ -275,11 +277,12 @@ message FeedbackReportingSystemContext {
   string package_version_name = 1;
   // The package code for the specific app version on the device.
   int32 package_version_code = 2;
-  // The ISO-639 code for the country locale, as determined by the system settings. This
-  // corresponds to the country locale as set by the device settings.
+  // The ISO-639 country code for the country locale set by the user (otherwise determined by the
+  // environment JVM). EX: A locale of Canada versus French Canada suggests different contexts for
```
BenHenning: Does 'different contexts' mean 'different locales' in this case? Please clarify in this comment if so.

model/src/main/proto/feedback_reporting.proto
BenHenning: Yes--definitely. Left one new comment below.

BenHenning (commented):
Thanks @jcqli! Just had one comment. PTAL & address, then I think this can be merged.

BenHenning:
Hmm @jcqli that looks like a potential flake. Going to file an issue & send a revert request to see if we can dig into this.

model/src/main/proto/feedback_reporting.proto
jcqli: I meant to convey that these are two different locales and suggests two different physical contexts that a user might be in -- rephrased to be more explicit about this with a more in-depth example.

BenHenning (approved):
Thanks @jcqli! LGTM.

BenHenning:
I think submission is blocked here. @jcqli you may need to sync this will develop after #2871 is merged & try re-merging it at that time.

PR #2802, opened 2021-03-02
Added answer tracking for practice session

TheRealJessicaLi:
## Description
Added tracking for the answers submitted by a user during a practice session (as per design doc for [score and mastery calculations](https://docs.google.com/document/d/1ISuqvKs6b0xdE0-aEIHXYG7oDbCmDBcgUia0TiR9GpQ/edit?usp=sharing) feature) in `QuestionAssessmentProgress`.

### DIFF
--- a/domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
+++ b/domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
@@ -35,23 +35,29 @@ internal class QuestionAssessmentProgress {
     }
   }
 
-  /** Called every time the user views a hint to track how many hints the user viewed per question this session.  */
-  internal fun hintViewed() {
+  /**
+   * Tracks how many hints the user viewed per question this session. This is expected to be called
+   * for every hint viewed.
+   */
+  internal fun trackHintViewed() {
     val currentQuestionIndex = getCurrentQuestionIndex()
     questionSessionMetrics[currentQuestionIndex].numberOfHintsUsed++
   }
 
-  /** Called every time the user views a solution to track which question solutions the user viewed this session.  */
-  internal fun solutionViewed() {
+  /**
+   * Tracks which question solutions the user viewed this session. This is expected to be called for
+   * every solution viewed.
+   */
+  internal fun trackSolutionViewed() {
     val currentQuestionIndex = getCurrentQuestionIndex()
     questionSessionMetrics[currentQuestionIndex].didViewSolution = true
   }
 
   /**
-   * Called every time the user submits an answer to track how many answers the user submitted per question, along with
-   * any misconceptions.
+   * Tracks how many answers the user submits for each question, along with any misconceptions. This
+   * is expected to be called for every answer submitted.
    */
-  internal fun answerSubmitted(taggedSkillMisconceptionId: String?) {
+  internal fun trackAnswerSubmitted(taggedSkillMisconceptionId: String?) {
     val currentQuestionIndex = getCurrentQuestionIndex()
     questionSessionMetrics[currentQuestionIndex].numberOfAnswersSubmitted++
     if (taggedSkillMisconceptionId != null) {
--- a/domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
+++ b/domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
@@ -47,6 +47,22 @@ internal class QuestionAssessmentProgress {
     questionSessionMetrics[currentQuestionIndex].didViewSolution = true
   }
 
+  /**
+   * Called every time the user submits an answer to track how many answers the user submitted per question, along with
+   * any misconceptions.
+   */
+  internal fun answerSubmitted(taggedSkillMisconceptionId: String?) {
+    val currentQuestionIndex = getCurrentQuestionIndex()
+    questionSessionMetrics[currentQuestionIndex].numberOfAnswersSubmitted++
+    if (taggedSkillMisconceptionId != null) {
+      // Misconceptions are formatted as <skill ID>-<misconception ID> from the AnswerClassificationController,
+      // but we only need the related skill ID for our purposes
+      questionSessionMetrics[currentQuestionIndex].taggedSkillMisconceptionIds.add(
+        taggedSkillMisconceptionId.split("-")[0]
+      )
+    }
+  }
+
   /** Processes when the current question card has just been completed. */
   internal fun completeCurrentQuestion() {
     isTopQuestionCompleted = true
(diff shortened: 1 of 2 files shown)


vinitamurthi (approved):

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -141,6 +142,16 @@ class QuestionAssessmentProgressController @Inject constructor(
...
             progress.stateList.computeAnswerOutcomeForResult(classificationResult.outcome)
           progress.stateDeck.submitAnswer(answer, answeredQuestionOutcome.feedback)
+
+          // Track the number of answers the user submitted, including any misconceptions
+          val misconception =
+            if (classificationResult is ClassificationResult.OutcomeWithMisconception) {
+              classificationResult.taggedSkillMisconceptionId
+            } else {
+              null
+            }
```
vinitamurthi: @BenHenning I am not sure if there's a shorter way of writing this in Kotlin..Do you have any idea?

vinitamurthi (approved):

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -47,6 +47,22 @@ internal class QuestionAssessmentProgress {
     questionSessionMetrics[currentQuestionIndex].didViewSolution = true
   }
 
+  /**
+   * Called every time the user submits an answer to track how many answers the user submitted per question, along with
+   * any misconceptions.
```
BenHenning: Method docs should always start with the action being taken, then why. E.g.:

```suggestion
   * Tracks how many answers the user submits for each question, along with any misconceptions. This
   * is expected to be called for every answer submitted.
```

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -47,6 +47,22 @@ internal class QuestionAssessmentProgress {
...
+   */
+  internal fun answerSubmitted(taggedSkillMisconceptionId: String?) {
+    val currentQuestionIndex = getCurrentQuestionIndex()
+    questionSessionMetrics[currentQuestionIndex].numberOfAnswersSubmitted++
+    if (taggedSkillMisconceptionId != null) {
+      // Misconceptions are formatted as <skill ID>-<misconception ID> from the AnswerClassificationController,
```
BenHenning, line 64: Is this something we did, or is it formatted that way per the exploration structure? If the former, we ought to use a data class to extract the values rather than string formatting (e.g. to avoid the [stringly typed antipattern](https://wiki.c2.com/?StringlyTyped)).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
BenHenning:
```suggestion
          val misconception = if (classificationResult is OutcomeWithMisconception) {
            classificationResult.taggedSkillMisconceptionId
          } else null
```

Shortest I think we can get it without structural changes, and this requires importing OutcomeWithMisconception directly.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -47,6 +47,22 @@ internal class QuestionAssessmentProgress {
     questionSessionMetrics[currentQuestionIndex].didViewSolution = true
   }
 
+  /**
+   * Called every time the user submits an answer to track how many answers the user submitted per question, along with
+   * any misconceptions.
+   */
+  internal fun answerSubmitted(taggedSkillMisconceptionId: String?) {
```
BenHenning: Method names should be present-tense active verbs to describe something that's happening or changing, e.g. ``trackSubmittedAnswer``. The object for which this method is being called is taking some sort of action--we should explicitly call out what that is. The fact that it happens when an answer submitted is a callsite detaill, not an implementation detail for this class. We might, for example, want to track submitted answers at other points in the future not just when the answer is submitted. A clearer name will provide less confusion in that situation.

BenHenning (commented):
Thanks @TheRealJessicaLi! Left some thoughts.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
TheRealJessicaLi: I believe this is formatted per the exploration structure - I checked the web repo and [this](https://github.com/oppia/oppia/blob/3fcc934d20cd756117a56b361315739fc687606d/core/templates/pages/exploration-editor-page/editor-tab/templates/modal-templates/add-answer-group-modal.controller.ts#L60) is how they're creating AnswerGroups. If you look at how they manage the mastery/score calculation logic, [this](https://github.com/oppia/oppia/blob/98c62ff192d75a0ffdaea24ec6559cc4be5a932b/core/templates/components/question-directives/question-player/question-player.directive.ts#L408) is exactly how they do it as well.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
TheRealJessicaLi: I wasn't able to find a cleaner way to do this either, but I modified it because this is better. Thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
TheRealJessicaLi: Fixed the documentation for this method as well as the previous tracking methods I added. Thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
TheRealJessicaLi: That's a really good point, thanks for pointing it out. Fixed!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
BenHenning: I think this is an area for potential improvement, then. Could we potentially replace the 'string' in our proto structure with a structured alternative the represents the constituent pieces? For example, we could have a structured alternative if we used something like:

```proto
message Misconception {
  string skill_id = 1;
  string misconception_id = 2;
}
```

It will require us to do this parsing when we retrieve the JSON object and convert it to proto (e.g. in StateRetriever), but that's already a location that's responsible for parsing so it's not unexpected, and it facilitates leveraging a structured approach elsewhere in the codebase which can help minimize error. WDYT?

FWIW, I suggest following up with a code change to implement the above rather than pulling it in here, but I think it would be a nice improvement to avoid the antipattern being introduced.

BenHenning (approved):
Thanks @TheRealJessicaLi! LGTM minus the stringly typed antipattern, but not blocking on that. See my comment for potential improvement. I suggest filing an issue to track that, and if you'd like, feel free to go ahead and make the change in a subsequent PR.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
TheRealJessicaLi: Yeah I think that's definitely a better approach, I'll go ahead and make an issue for that. Thanks!

PR #2830, opened 2021-03-02
Fix #2803: Fixed title in Recenty played activity.

### LINKED ISSUE #2803: Dynamic for RecentlyPlayedActivity
In home screen the text above the carousel is dynamic and when clicked on `View All` it opens `RecentlyPlayedActivity` in which currently the title is constant. Instead the title should be dynamic similar to text in home screen.

The title in `RecentlyPlayedActivity` can show following texts:
1. Stories for you
2. Recently played stories

In case of `Recommended stories` and `Upcoming topics` we do not show `View All` button and therefore this activity won't appear.

veena14cs:
This PR fixes #2803 , the toolbar title in Recently-Played Activity. It matches with the Home screen crousel, 
1) When there is combination of recently-played topics and recommended topics the title shows `Stories For You`
2) When there are no recommendations it displays `Recently Played Activity`

**Screenshot**
Screenshot_20210302_160113
Screenshot_20210302_160138

**Design Doc**
https://docs.google.com/document/d/1o7hOz1rP-hJaKn1f5x6LV6ms0bzhyCpsTifJQY5crCc/edit?ts=5f9fc81b#


app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -75,67 +75,70 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
-          itemList.add(recentSectionTitleViewModel)
-          for (promotedStory in it.promotedStoryList.recentlyPlayedStoryList) {
-            val ongoingStoryViewModel =
-              OngoingStoryViewModel(
-                promotedStory,
-                entityType,
-                fragment as OngoingStoryClickListener
-              )
-            itemList.add(ongoingStoryViewModel)
+      {
```
rt4914, line 78: This is giving me error.
Screenshot 2021-03-02 at 8 54 53 PM

Replacing `{` with `Observer<PromotedActivityList> ` does work correctly.

How does this work for you?

rt4914 (changes requested):
@BenHenning The solution does work but considering that we will need to change the title by passing an enum to the activity so that we can set the label and therefore it will work correctly for A11y users, do you think this solution is still needed? I think the enum solution should work fine and that would give the same results.

WDYT?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
BenHenning: @rt4914 are you using a version of Android Studio with Kotlin 1.3? There may be inconsistencies if your AS isn't using the same version of Kotlin used in the build.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -75,67 +75,70 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
+        with(it.promotedStoryList) {
```
BenHenning: I think using ``with`` in this way is actually making it harder to understand what's going on.

Plus, this method is starting to get long/hard to understand. Can we break it into smaller helper methods/pieces to make the constituent pieces clearer?

BenHenning (commented):
Thanks @veena14cs. Left some preliminary thoughts.

BenHenning:
@rt4914 is the enum solution done already, or is it something we can do within this PR? Also, how do we ensure there aren't inconsistencies between the host & the destination activity? I suppose that's probably difficult to actually cause, but do we not care if there's a small inconsistency in the title?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
veena14cs: I have optimized code PTAL.

veena14cs:
> @rt4914 is the enum solution done already, or is it something we can do within this PR? Also, how do we ensure there aren't inconsistencies between the host & the destination activity? I suppose that's probably difficult to actually cause, but do we not care if there's a small inconsistency in the title?

With enum, we also need to check that Recently-played activity is coming from ProfileProgress Activity too. @rt4914 let me know what solution could be better as we can use it for section title too. Now the toolbar title and section title is using the same implementation.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
veena14cs: May be inconsistency with kotlin versions. As it was in Android Studio suggestion which I removed it.

rt4914:
@BenHenning and @veena14cs 
I enum solution is not done yet. But as per Veena's last comment there is one issue with enum solution, i.e., it works only for `Home Screen` `View All` button and not for `Profile Progress` `View All` button.

My suggestion:
1. We should go ahead with this PR so that it can take care of header in both cases: (a.) `View All` in `Home Screen` and (b.) `View All` in `Profile Progress`
2. We introduce enum (in separate PR) which will be used when going from `Home` to `Recently Played`. In this case the talkback will be able to output either `Stories for you` or `Recently played stories`. Also in this case the `Talkback` output will be consistent with the `Toolbar Title`
3. In case when we are going from `Profile Progress` to `Recently Played` the talkback will be able to output either `Stories for you`  (as default label). In this case `Talkback` output can be inconsistent with the `Toolbar Title`.

@BenHenning WDYT? Also can you think of some better solution for this? One more alternative can be to keep `Stories for you` as fixed label for `Talkback`.

BenHenning:
I think going ahead with this solution seems reasonable. I think this may be an exceptional case where we need to default the accessibility label & overwrite it to match the toolbar title. We'll still need to pass in the label via proto arguments, and in some odd cases there may be disparity but the label will be eventually consistent (which will only affect cases when the user causes the label to be re-announced, e.g. via the system tasks/recents menu).

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -140,6 +99,55 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
+  }
+
+  private fun getOngoingStoryViewModel(promotedStory: PromotedStory): RecentlyPlayedItemViewModel {
+    return OngoingStoryViewModel(
+      promotedStory,
+      entityType,
+      fragment as OngoingStoryClickListener
+    )
+  }
+  private fun addOlderStoryListSection(olderPlayedStoryList: List<PromotedStory>) {
```
BenHenning, line 125: Newline above the function for consistency.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -140,6 +99,55 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
+      )
+    itemList.add(olderSectionTitleViewModel)
+    for (promotedStory in olderPlayedStoryList) {
+      val ongoingStoryViewModel = getOngoingStoryViewModel(promotedStory)
+      itemList.add(ongoingStoryViewModel)
+    }
+  }
+
+  private fun addRecommendedStoryListSection(suggestedStoryList: List<PromotedStory>) {
+    binding.recentlyPlayedToolbar.title = activity.getString(R.string.stories_for_you)
```
BenHenning: Suggest updating the toolbar outside this method (e.g. in the subscribe method above) for all the cases to keep it consistent.

Also, don't we need to set this for each of the different cases?

BenHenning (changes requested):
Thanks @veena14cs! Left a couple of comments.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
rt4914: Yes, it was because of kotlin version. I updated it and it was fixed. Thanks.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
rt4914: We need this for `addRecentlyPlayedStoryListSection ` function too.

rt4914 (changes requested):
Is it possible to simply create a function similar to `getHeader ` present in `PromotedStoryListViewModel` ?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
veena14cs: Done for all the cases.

veena14cs:
> Is it possible to simply create a function similar to `getHeader ` present in `PromotedStoryListViewModel` ?

Will need to add observers then similar to subscribeToOngoingStoryList().

BenHenning:
Thanks @veena14cs. Functionally, this LGTM.

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
```diff
@@ -237,6 +237,30 @@ class RecentlyPlayedFragmentTest {
     }
   }
 
+  @Test
+  fun testRecentlyPlayedTestActivity_fractionsPlayed_storiesForYouToolbarTitleIsDisplayed() {
```
BenHenning, line 241: Do we have tests for the other toolbar cases? If not, please add them.

BenHenning (approved):
Just one comment on testing coverage, other PR LGTM.

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
veena14cs: Yes, it is present. I will rename other testcase to keep it consistent with this.

veena14cs: Done.

rt4914 (approved):
LGTM, thanks.

app/src/sharedTest/java/org/oppia/android/app/home/RecentlyPlayedFragmentTest.kt
rt4914: Ack.

Issue #53, opened 2019-08-02
App user feedback reporting


BenHenning:
It's important that users have the ability to report issues they observe in the app. This should include both free-form text and optionally a screenshot. System logs and other useful debug information should be included, and these reports should be stored securely in Oppia's backend with an easy way to view them to investigate the issues.

BenHenning:
This is blocked on #8 and #22 for placement in the UI. This feature requires a design document and approval from an Oppia data owner due to the sensitivity of the data being stored in these feedback reports.

BenHenning:
The prototype does not require user feedback reporting since it will only be available to controlled and supervised audiences.

BenHenning:
NB: This is only blocked on #22 for the UI integration. We still need to figure out how we're even going to support user feedback reporting in the app.

BenHenning:
@eraser95 sending this your way.

anandwana001:
updating PR title as this issue is no more blocked on #22

jcqli:
**Design Doc**
[User Feedback Reporting Design Doc
](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit)

**Milestones**
Tasks will be broken down into individual PRs as follows:

**Android Repo**
1. Create Android model layer objects #2791
2. Create Android data layer objects & service
3. Create Android domain controller & implement logic for creating protobufs #2792 
4. Implement Android domain layer upload worker
5. Implement app layer for non-crash entry points and entry fragment
6. Implement app layer for main form fragment
7. Implement app layer for agreement fragment
8. Implement domain controller logic for data storage (local & remote) to connect app layer and domain layer
9. Implement domain layer startup listener & controller for crash reporting
10. Implement app layer for crash reports

**Web Repo**
1. Create backend storage models
2. Create backend domain objects
3. Create backend controllers
4. Create backend web services that processes incoming reports
5. Create cron jobs to delete & aggregate reports
6. Create frontend & backend services that update reports, & frontend components that displays dashboard & filters
7. Create frontend aggregated data view

PR #2838, opened 2021-03-03
Fix #2792: Create data layer for feedback reporting

### LINKED ISSUE #2792: Create Android data layer for feedback reporting
Fix part of #53 
Implements the data layer layer needed for feedback reporting (data class objects & HTTP service)

Design doc: [User Feedback Reporting](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

jcqli:
## Explanation
Fix #2792 -- creates data layer Moshi objects used for feedback reporting and implements the singleton HTTP service that posts requests to backend storage.
Creates a mock service and fake JSON for testing.

Feature tracker: #53

[Design Doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

## Checklist


jcqli:
@BenHenning PTAL

I also considered adding to the [NetworkInterceptor](https://github.com/oppia/oppia-android/blob/develop/data/src/main/java/org/oppia/android/data/backends/gae/NetworkInterceptor.kt#L20) to handle updating the API key but decided that that was outside the scope of these changes / this PR.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -10,6 +10,10 @@ import retrofit2.converter.moshi.MoshiConverterFactory
 import javax.inject.Qualifier
 import javax.inject.Singleton
 
+/** API key to validate messages sent over the network. */
```
BenHenning: Or perhaps: ``Corresponds to an API key specific to Oppia Android to allow the backend to verify the origin of requests sent from the app.`` or similar.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -60,4 +64,12 @@ class NetworkModule {
   fun provideClassroomService(@OppiaRetrofit retrofit: Retrofit): ClassroomService {
     return retrofit.create(ClassroomService::class.java)
   }
+
+  /**
+   * Provides the API key to validate messages received and messages being sent. A stronger, secret
+   * key will be used when putting app out in production.
+   */
```
BenHenning: NB: KDocs aren't needed for providers since they aren't actually called by human-written code. Keeping this for the added context is fine though.

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
```diff
@@ -0,0 +1,14 @@
+package org.oppia.android.data.backends.gae.api
+
+import org.oppia.android.app.model.FeedbackReport
+import retrofit2.Call
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+/** Service that posts feedback reports to remote storage. */
+
```
BenHenning: Remove extra newline.

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
```diff
@@ -0,0 +1,14 @@
...
+interface FeedbackReportingService {
+  @POST("feedbackreporting/incomingreport")
+  // TODO(#76): Update return payload for handling storage failures once retry policy is defined.
+  fun postFeedbackReport(@Body report: FeedbackReport): Call<Unit>
```
BenHenning: Shouldn't we be posting GaeFeedbackReport here, not the proto?

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Data class for the feedback report sent by the Android app to remote storage.
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeFeedbackReport @Inject constructor(
+
+  @NetworkApiKey @Json(name = "api_key") val api_key: String,
+  @Json(name = "report_creation_timestamp") val report_creation_timestamp_ms: Double?,
+  @Json(name = "user_feedback") val user_feedback: GaeUserSuppliedFeedback?,
```
BenHenning:
```suggestion
  @Json(name = "user_feedback") val userFeedback: GaeUserSuppliedFeedback?,
```

Fields should follow naming conventions. Ditto elsewhere.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.squareup.moshi.JsonClass
+import org.oppia.android.data.backends.gae.NetworkApiKey
+import javax.inject.Inject
+
+/**
+ * Data class for the feedback report sent by the Android app to remote storage.
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeFeedbackReport @Inject constructor(
```
BenHenning: This should not be an inject constructor--it's being constructed by us & Moshi, not Dagger.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,22 @@
...
+
+  @NetworkApiKey @Json(name = "api_key") val api_key: String,
```
BenHenning: The qualifier shouldn't be used in this way--we need to inject the key wherever we create instances of this class & assign it at that point.

BenHenning (commented):
Thanks @jcqli! Took an initial pass on the first few files. Left some general comments. Further, are the GAE objects updated to match the changes in the protos, and are they representing the structure we need to consume in the backend? I'm assuming we'll be using strings for the checkboxes rather than enums at this point--is that correct?

jcqli:
> Thanks @jcqli! Took an initial pass on the first few files. Left some general comments. Further, are the GAE objects updated to match the changes in the protos, and are they representing the structure we need to consume in the backend? I'm assuming we'll be using strings for the checkboxes rather than enums at this point--is that correct?

Yes, the checkboxes will be a list of strings and the GAE objects are updated as the protos are updated. When you meant "representative' of the structure, do they need to be exact matches? I presume we can parse the objects as needed before storing them in the models.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Ack, got it!

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
jcqli: Yeah, I was conflating how these services worked :)

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Ah, done

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Got it, removed

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -10,6 +10,13 @@ import retrofit2.converter.moshi.MoshiConverterFactory
 import javax.inject.Qualifier
 import javax.inject.Singleton
 
+/**
+ * Corresponds to an API key specific to Oppia Android for the backend to verify the origin of
+ * requests sent from the app.
+ */
+@Qualifier
+annotation class NetworkApiKey
```
BenHenning: This should go into a separate file (in general, prefer to keep unique top-level classes in their own files).

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.squareup.moshi.JsonClass
+
+/**
+ * Data class for the feedback report sent by the Android app to remote storage.
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeFeedbackReport constructor(
+
+  @Json(name = "api_key") val apiKey: String,
```
BenHenning: Here & elsewhere: please add brief KDoc comments explaining what these are. You'll probably do this in many places throughout the stack, but each place may have someone encounter the code that's has no idea how your feature work, so it's important to make sure each piece can "stand on its own" for whatever purpose that component is trying to accomplish.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Data class for feedback reporting app info represented in the backend domain model
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeFeedbackReportingAppContext(
+
+  @Json(name = "entry_point") val entryPoint: String?,
+  // Maps a string representing the topic progress to the number of topics with that status.
```
BenHenning: Actually, this confuses me. Shouldn't topic progress just be a list of completed explorations? I think that's all we need actually.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeUserSuppliedFeedback.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Data class for feedback reporting form model
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeUserSuppliedFeedback(
+
+  @Json(name = "report_type") val reportType: String?,
+  @Json(name = "category") val category: String?,
```
BenHenning: Is a subcategory needed to contextualize the feedback list?

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeUserSuppliedFeedback.kt
```diff
@@ -0,0 +1,18 @@
...
+  @Json(name = "feedback_list") val feedbackList: List<String>?,
+  @Json(name = "user_input") val userInput: String?
```
BenHenning: Maybe: ``open_form_user_input`` might be a bit more explicit since all of this is technically user input. WDYT?

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,20 @@
...
+  @Json(name = "report_creation_timestamp_ms") val reportCreationTimestampMs: Double?,
```
BenHenning: Why not a ``Long?``?

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,20 @@
...
+  @Json(name = "user_supplied_feedback") val userSuppliedFeedback: GaeUserSuppliedFeedback?,
```
BenHenning: Here & elsewhere: why are these properties optional (e.g. nullable)? Suggest actually making the optional ones nullable and everything else should be required.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Data class for the feedback report sent by the Android app to remote storage.
+ * TODO(#2801): Link backend domain model
+ */
+@JsonClass(generateAdapter = true)
+data class GaeFeedbackReport constructor(
```
BenHenning: Strike 'constructor'--it's not needed since there's nothing being modified about it (e.g. visibility or annotations).

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -60,4 +67,12 @@ class NetworkModule {
...
     return retrofit.create(ClassroomService::class.java)
   }
+
+  /**
+   * Provides the API key to validate messages received and messages being sent. A stronger, secret
+   * key will be used when putting app out in production.
+   */
+  @Provides
+  @NetworkApiKey
+  fun provideNetworkApiKey(): String = ""
```
BenHenning: Are you planning to hook this up in this PR or in a later one? If a later one, I suspect moving these changes to that PR. If in this one, I think that code is currently missing.

BenHenning (changes requested):
Thanks @jcqli! Left another batch of comments. Please send back once addressed & this PR is updated with the changes from #2800. Will take one more full pass, but I suspect it will be close to LGTM by that point.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeUserSuppliedFeedback.kt
jcqli: Most categories don't have a subcategory. Suggestion / Issue / Crash is a report type whereas category would be specific to each type (e.g. suggestion type and new feature / new language / other category). The only subcategory is issue > language > audio / text, but I was thinking that could be considered a category (e.g. audio_language, text_language, general_language if they don't specify subcategory).

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Ah I presumed that this was a nuance of data classes since all other GAE models I saw had nullable fields, will change that.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Do you mean setting the actual value for the API key string or calling this provider in the GAE model that's sent in the message? If the former, I had thought that the key would be overridden in prod anyways so the value here wouldn't matter.

Also, moving this to a later PR since the API key will be added / used in the domain layer.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Ack, I think this would go in the domain layer for the controller to inject when it sends the report over the network so I'll include this the domain layer PR

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
jcqli: My hesitation is that this might get really long as more topics are added, but I'd agree that this would be more useful.
Updated

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeUserSuppliedFeedback.kt
jcqli: Done, changed to `open_text_user_input` so it's explicit that it's a text response.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Ah, I think I had originally changed this to a double in the proto so I was matching that, but keeping the int64 and a Long does make more sense

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -69,10 +63,13 @@ class NetworkModule {
   }
 
   /**
-   * Provides the API key to validate messages received and messages being sent. A stronger, secret
-   * key will be used when putting app out in production.
+   * Provides the Feedback Reporting service implementation.
+   * @param retrofit the Retrofit object used to instantiate the service
+   * @return the Feedback Reporting service implementation
```
BenHenning: Dagger providers don't need KDocs since their public API are consumed only by generated code.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
```diff
@@ -10,16 +10,25 @@ import com.squareup.moshi.JsonClass
...
-  @Json(name = "audio_lang") val audioLang: String?,
-  @Json(name = "download_and_update_only_on_wifi") val downloadAndUpdateOnlyOnWifi: Boolean?,
-  @Json(name = "automatically_update_topics") val automaticallyUpdateTopics: Boolean?,
-  @Json(name = "is_admin") val isAdmin: Boolean?,
-  @Json(name = "event_logs") val eventLogs: String?,
-  @Json(name = "logcat_logs") val logcatLogs: String?
+  /** Corresponds to a location in the app that a user can access feedback reporting. */
+  @Json(name = "entry_point") val entryPoint: String,
+  /** Corresponds to the list of topics that the user has completed. */
+  @Json(name = "topic_progress") val topicProgress: List<String?>,
```
BenHenning: ``List<String>`` I think is what you want here.

BenHenning: Does the proto also need to be fixed for this, or was that done already?

BenHenning (changes requested):
Thanks @jcqli! Left a few more comments.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
BenHenning: Followed up offline--network API key will be implemented via an HTTP header in a later PR.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeUserSuppliedFeedback.kt
BenHenning: Thanks for clarifying.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
BenHenning: It shouldn't be too long (e.g. <1000). If that scales poorly, we can address it later.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
BenHenning: Sounds good--thanks.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
jcqli: Ah I guess that makes sense -- I was thinking a user with no completed stories would have an empty list but that's still `List<String>`

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done, changed to comment

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingAppContext.kt
jcqli: Thought it was.. fixed now :)

model/src/main/proto/feedback_reporting.proto
```diff
@@ -326,20 +326,8 @@ message FeedbackReportingAppContext {
...
-    NOT_STARTED = 1;
-    // The topic has been started but not completed by the player.
-    STARTED_NOT_COMPLETED = 2;
-    // The topic has been completed by the player.
-    COMPLETED = 3;
-  }
-  // Maps topic ID to the completion state of topic, based on state of stories in topics. Helps
-  // contextualize the user's app state and understand the scenario the report is filed during.
-  map<string, TopicProgressStatus> topic_progress = 2;
+  // Lists the topics that have been completed by the user.
```
BenHenning: Or perhaps: ``A list of exploration IDs that have been completed by the user.``

Suggest specifically naming this 'completed_exploration_ids' & ditto elsewhere for this field to be explicit. That should lead to less potential confusion over what this field actually contains/what "topic progress" means.

BenHenning (approved):
Thanks @jcqli! LGTM. Just had one comment to bring consistency to the finished exploration IDs.

model/src/main/proto/feedback_reporting.proto
jcqli: Done

PR #2845, opened 2021-03-06
Partially fix #2844: force Gradle tests to run on JDK 9 in CI

BenHenning:
Fixes part of #2844.

This aims to fix the OpenJDK flakes happening for Gradle tests (per #2844) by forcing the CI environment to use JDK 9 instead of JDK 8 for runtime, since it seems that JDK 8 may have an issue where the JVM can crash when managing many forks (we updated Gradle tests to use forks a while back). However, making the CI environment properly use JDK 9 for Gradle involved a few steps:
- Adding additional steps to add environment diagnostics to ensure that Gradle is actually using JDK 9 (since it seems to use JDK 8 as configured by settings despite 9 being setup). Unfortunately, setup-java doesn't yet seem to support this directly: https://github.com/actions/setup-java/issues/92, and it's important to verify that Gradle itself is using the correct version.
- Note that moving to JDK 9+ requires a compile-time only dependency for some of the annotations needed by Dagger: https://github.com/google/dagger/issues/1449#issuecomment-588277905. It's not clear to me why this wasn't an issue in the JDK 8 world.
- Similarly, JDK 9+ deprecate and eventually remove (JDK 11+) some Java EE libraries that Gradle itself seems to depend on. More compile-only dependencies resolve this issue: https://stackoverflow.com/a/43574427.

Some things to note:
- The build still sets source/target compatibility versions to Java 8 since we don't actually want to use Java 9+ dependencies, we just want to be able to use JDK 9+ to build and run the tests
- I haven't verified this locally to still work with JDK 8, but I'm not expecting there to be issues since the changes are mostly just adding redundancy
- I removed the workaround for the kapt bug since per https://github.com/google/dagger/issues/1449#issuecomment-595238637 this should be fixed in the version of Kotlin that we're now using (though the other fixes above regarding the compile-time dependencies were discovered shortly after & are necessary to make everything work as expected)

We will not actually be able to verify if the issue is fixed until we see no new instances of the crash crop up over the next few weeks. Given that the issue seems to only reproduce occasionally when running the whole batch of tests, it's difficult to make the issue commonplace enough to gain certainty it's been fixed. Given the low volume of the flake, though, waiting a few weeks to ensure it's gone should be reasonable.

Issue #2844, opened 2021-03-06
Flaky OpenJDK crashes (Gradle & Bazel)


BenHenning:
There's no clear reproduction of this issue, but we're occasionally seeing SIGSEGV failures when running tests:
- https://github.com/oppia/oppia-android/runs/2031235243?check_suite_focus=true
- https://github.com/oppia/oppia-android/pull/2830/checks?check_run_id=2037294748 from #2830 

& others. While the above are Gradle-specific, I believe we've seen this with Bazel as well. Here's one failure result from Gradle:

<details>

```
2021-03-04T13:31:07.6689399Z #
2021-03-04T13:31:07.6690431Z # A fatal error has been detected by the Java Runtime Environment:
2021-03-04T13:31:07.6691221Z #
2021-03-04T13:31:07.6692228Z #  SIGSEGV (0xb) at pc=0x00007efe2d2a4261, pid=9982, tid=0x00007efdffaa2700
2021-03-04T13:31:07.6692888Z #
2021-03-04T13:31:07.6706440Z # JRE version: OpenJDK Runtime Environment (8.0_282-b08) (build 1.8.0_282-b08)
2021-03-04T13:31:07.6707614Z # Java VM: OpenJDK 64-Bit Server VM (25.282-b08 mixed mode linux-amd64 compressed oops)
2021-03-04T13:31:07.6708383Z # Problematic frame:
2021-03-04T13:31:07.6709280Z # V  [libjvm.so+0x92a261]  Node::uncast() const+0x1
2021-03-04T13:31:07.6709697Z #
2021-03-04T13:31:07.6710636Z # Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
2021-03-04T13:31:07.6711317Z #
2021-03-04T13:31:07.6711743Z # An error report file with more information is saved as:
2021-03-04T13:31:07.6712727Z # /home/runner/work/oppia-android/oppia-android/app/hs_err_pid9982.log[thread 139628459652864 also had an error]
2021-03-04T13:31:07.6713236Z 
2021-03-04T13:33:07.6689128Z # [ timer expired, abort... ]
2021-03-04T13:36:00.1688295Z 
2021-03-04T13:36:00.1690914Z org.oppia.android.app.player.state.StateFragmentLocalTest > testStateFragment_nextState_viewFourHints_submitWrongAnswer_noNewHintIsAvailable SKIPPED
[...]

Issue #1861, opened 2020-09-18
Optimize CI runs with Bazel [Blocked: #59]


BenHenning:
There are a few things we can do to improve CI runs once Bazel becomes the primary build system for the project, in particular:

1. We could have a multi-stage workflow, where the first set of jobs runs linters, creates the emulator image & related dependencies for #1844, and uses Bazel to build the APK and all tests. The outputs from these jobs would be uploaded as artifacts. The second stage would download these artifacts and run the tests in several parallel jobs (likely Robolectric for 1 job and N jobs for the emulator tests).

2. We could create a feedback loop by re-downloading the Bazel artifacts in stage 1 for subsequent pushes for a new copy of the codebase. If done correctly, this could result in an incremental Bazel build with only the latest changes. The dependencies and emulator artifacts don't need to be recreated. This would result in stage 1 becoming very quick for subsequent commits, leading to a significantly shorter overall test time.

3. For push-based workflow initiations, we should use Bazel to determine exactly which test targets are actually affected by the changes in the PR and only run those tests (we can still build everything--this is just an optimization for stage 2 so that it runs faster).

The result of this work, if possible, would be a bit slower initial commit workflow run but significantly faster subsequent runs up to the PR submission. develop & release branch pushes should still run the entire test suite from scratch.

Some caveats to consider:
1. Timestamp changes between CI machine images could cause Bazel to rebuild everything--some investigation will be needed into Bazel build stability

2. We use homebrew currently to install ffmpeg & its dependencies for capturing videos for emulator tests--there's not a clear way to upload/download Homebrew artifacts without just reinstalling them (or if it will actually be a performance improvement)

3. We need to handle the case of many quick, successive commits (especially after the initial PR commit and before that commit's slower preparation workflows have completed)

BenHenning:
@vinitamurthi any thoughts?

vinitamurthi:
Overall this looks good to me. I think the emulator image specifically will be cached by the GitHub cache action itself. Not so sure about whether we can build Bazel incrementally with the way the cache works today.

vinitamurthi:
I also like the idea of step 3 but we'd have to make a script to do that as far as I understand. We could even use that script in the pre-push hooks and run tests just for the affected files

BenHenning:
Thanks. To be clear, I'm assuming we need to use artifacts as a cache to achieve the goals of this issue, rather than using the existing cache since that doesn't seem to share things across job instances.

[...]

PR #2846, opened 2021-03-06
[RunAllTests] Fix part of #2844: work around Bazel-specific CI issue that's causing an OpenJDK crash/flake

BenHenning:
Fix part of #2844.

This attempts to fix part of #2844 by normalizing test compiling/running on JDK 9 (for parity with Gradle tests--see #2845), and by working around what appears to be a resource constraint issue Bazel is hitting. For the latter, see https://github.com/bazelbuild/bazel/issues/3236#issuecomment-587752609. All configuration changes in this PR are done via changes to the CI environment's ``~/.bazelrc`` file, so nothing here should affect local builds.

Note that configuring Java versions in Bazel is complicated. In particular, at any time there can be 5 different versions of Java at play:
1. The version of Java used to build sources (defaults to a remote JDK for build reproducibility). Not sure what the default is here, but it seems to be 11.
2. The version of Java used to run tests/binaries (defaults to the local environment's JDK as determined by $JAVA_HOME).
3. The version of Java used to build tools used by Bazel. Defaults to a remote JDK 11.
4. The version of Java used to run tools used during the build. Defaults to local JDK 11 (if available--not sure what happens if it's not available; it might use the Java version in (5)).
5. The version of Java used to run Bazel itself. This defaults to an embedded JDK 11.

Now, from what I can tell, I think we only need to configure (1) since (2) should already be using the correct version of Java (JDK 9). We do want the majority of the build to be on JDK 9 since that's what we're running on, but this isn't *super* important. See also https://docs.bazel.build/versions/master/bazel-and-java.html#configuring-the-jdk & https://github.com/bazelbuild/bazel/issues/5594 for more details on how this Java resolution works.

Finally, similar to #2845 this cannot be observed as "correct" until we observe the flake itself disappearing over the next few weeks.


.github/workflows/build_tests.yml
```diff
@@ -19,18 +19,39 @@ jobs:
...
+          $JAVA_HOME/bin/java -version
+
+      - name: Set up Bazel
+        uses: abhinavsingh/setup-bazel@v3
+        with:
+          version: 4.0.0
+
+      - name: Configure Bazel to use JDK 9 for building
+        run: |
+          echo "build --java_language_version=9" >> $HOME/.bazelrc
```
BenHenning, line 44:
```suggestion
      - name: Configure Bazel to use JDK 9 for building
        run: |
          echo "build --java_language_version=9" >> $HOME/.bazelrc

      - name: Configure Bazel to use specific sandbox tmpfs
        run: |
          echo "build --enable_platform_specific_config" >> $HOME/.bazelrc
          echo "build:linux --sandbox_tmpfs_path=/tmp" >> $HOME/.bazelrc
```

.github/workflows/unit_tests.yml
```diff
@@ -62,21 +62,39 @@ jobs:
...
+
       - name: Set up Bazel
         uses: abhinavsingh/setup-bazel@v3
         with:
           version: 4.0.0
+          
+      - name: Configure Bazel to use JDK 9 for building
+        run: |
+          echo "build --java_language_version=9" >> $HOME/.bazelrc
+
```
BenHenning:
```suggestion

      - name: Configure Bazel to use specific sandbox tmpfs
        run: |
          echo "build --enable_platform_specific_config" >> $HOME/.bazelrc
          echo "build:linux --sandbox_tmpfs_path=/tmp" >> $HOME/.bazelrc

```

vinitamurthi:
Is this still required?

BenHenning:
@vinitamurthi this is expected to help, but it may not be sufficient to fully fix the Bazel-caused flakes.

vinitamurthi (approved):

anandwana001 (approved):
LGTM

BenHenning:
Thanks all! Enabling auto-merge after resolving conflicts since some of the changes here may help with work I'm doing in #1861 to make Bazel CI runs a bit faster.

PR #2847, opened 2021-03-07
Fix part of #2691: Add diagnostics to investigate issues with compute affected targets

BenHenning:
Fix part of #2691.

Add extra diagnostic output to investigate issues with the compute affected targets workflow.

See https://github.com/oppia/oppia-android/issues/2691#issuecomment-792150231 for specifics on the issue being investigated here.


BenHenning:
Restart CI to run all tests to demonstrate the new behavior.

BenHenning:
Ah. Oops--the updated workflow specifically doesn't execute when ``[RunAllTests]`` is present. :)

BenHenning:
Better diagnostics were added in #3025 making this unnecessary for now.

PR #2865, opened 2021-03-09
Enable Accessibility Checks in test cases [WIP]

rt4914:
## Explanation

This PR aims as create a demo of how we can enable Accessibility Checks on in either of roboelectric or espresso test cases.

In this PR the solution works for Espresso only. The current test fails with below mentioned error which is correct and was done so as to prove that the implementation works correctly.

`
com.google.android.apps.common.testing.accessibility.framework.integrations.AccessibilityViewCheckException: There was 1 accessibility error:
AppCompatButton{id=2131296812, res-name=profile_rename_save_button, visibility=VISIBLE, width=242, height=132, has-focus=false, has-focusable=true, has-window-focus=false, is-clickable=true, is-enabled=true, is-focused=false, is-focusable=true, is-layout-requested=false, is-selected=false, layout-params=androidx.constraintlayout.widget.ConstraintLayout$LayoutParams@64bf52a, tag=null, root-is-layout-requested=false, has-input-connection=false, x=761.0, y=321.0, text=save, input-type=0, ime-target=false, has-links=false}: TextView does not have required contrast of 3.000000. Actual contrast is 1.000000
`

If we run this PR on Roboelectric then test cases do fail giving error related to Accessibility but it is not clear.
`java.lang.NoClassDefFoundError: Could not initialize class com.google.android.apps.common.testing.accessibility.framework.SpeakableTextPresentViewCheck`

## Problem 1
The [documentation](https://developer.android.com/training/testing/espresso/accessibility-checking#enable-checks) mentions following import: `import androidx.test.espresso.accessibility.AccessibilityChecks`
which requires following import too `'androidx.test.espresso:espresso-accessibility:3.1.0'`

But when I try to run the app with this import
`Unresolved reference: AccessibilityChecks`.

Reference for this problem were following: [documentation](https://developer.android.com/training/testing/espresso/accessibility-checking#enable-checks) and [medium blog](https://medium.com/@artour.bakiev/accessibility-testing-for-android-apps-how-to-start-ce7c09628743)

To solve problem one I had to use the deprecated import as per this suggestion: https://stackoverflow.com/a/51240061

## Problem 2
I tried to enable the `AccessibilityChecks` for Roboelectric by adding the following annotation to the test case `@AccessibilityChecks` but in this case all the tests pass which is incorrect for following activity.

## Summary
I haven't found any working solution for to enable A11y checks for Roboelectric. The solution that worked for Espresso uses a deprecated import and not the import which is mentioned on the developer documentation.

Any suggestion on above problems and how to approach them?

## Checklist

- [ ] The PR does not contain any unnecessary auto-generated code from Android Studio.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -150,6 +150,7 @@ dependencies {
   )
   androidTestImplementation(
       'androidx.test:core:1.2.0',
+      'androidx.test.espresso:espresso-accessibility:3.1.0',
       'androidx.test.espresso:espresso-contrib:3.1.0',
       'androidx.test.espresso:espresso-core:3.2.0',
       'androidx.test.espresso:espresso-intents:3.1.0',
--- a/app/src/main/res/layout/profile_rename_activity.xml
+++ b/app/src/main/res/layout/profile_rename_activity.xml
@@ -85,7 +85,7 @@
           android:layout_height="wrap_content"
           android:layout_marginTop="16dp"
           android:layout_marginEnd="28dp"
-          android:background="@{viewModel.inputName.length > 0 ? @drawable/state_button_primary_background :@drawable/start_button_transparent_background}"
+          android:background="@color/white"
           android:clickable="@{viewModel.inputName.length > 0}"
           android:enabled="@{viewModel.inputName.length > 0}"
           android:text="@string/profile_rename_save"
--- a/app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileRenameActivityTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileRenameActivityTest.kt
@@ -8,11 +8,11 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.accessibility.AccessibilityChecks
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.action.ViewActions.closeSoftKeyboard
 import androidx.test.espresso.action.ViewActions.pressImeActionButton
 import androidx.test.espresso.assertion.ViewAssertions.matches
-import androidx.test.espresso.contrib.AccessibilityChecks
 import androidx.test.espresso.intent.Intents
 import androidx.test.espresso.intent.Intents.intended
 import androidx.test.espresso.intent.matcher.IntentMatchers.hasComponent
--- a/app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileRenameActivityTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileRenameActivityTest.kt
@@ -12,6 +12,7 @@ import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.action.ViewActions.closeSoftKeyboard
 import androidx.test.espresso.action.ViewActions.pressImeActionButton
 import androidx.test.espresso.assertion.ViewAssertions.matches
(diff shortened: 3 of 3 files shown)


BenHenning:
@rt4914 from your description, I'm slightly confused. You mentioned that Robolectric tests do fail as expected, but then you mention they don't work. Is it that Robolectric tests work correctly with the AndroidX AccessibilityChecks, but Espresso doesn't because they fail to build with them?

BenHenning:
Also per https://github.com/robolectric/robolectric/issues/5642#issuecomment-689866428 it may be the case that Robolectric just won't properly support accessibility test use cases, so we may need to lean on Espresso for this purpose.

BenHenning:
I'm unsure why the AndroidX dep issue is happening--maybe dig into the Maven artifact to make sure the class is actually there?

rt4914:
> @rt4914 from your description, I'm slightly confused. You mentioned that Robolectric tests do fail as expected, but then you mention they don't work. Is it that Robolectric tests work correctly with the AndroidX AccessibilityChecks, but Espresso doesn't because they fail to build with them?

@BenHenning Yes it does work but we are using `Espresso` library `A11yChecks` there. So basically `A11yChecks` from `Espresso` works for both `Espresso` as well as `Roboelectric`  but gives descriptive error for `Espresso`

rt4914:
> Also per [robolectric/robolectric#5642 (comment)](https://github.com/robolectric/robolectric/issues/5642#issuecomment-689866428) it may be the case that Robolectric just won't properly support accessibility test use cases, so we may need to lean on Espresso for this purpose.

Correct. Even the developer guide mentions the same: https://developer.android.com/guide/topics/ui/accessibility/testing#robolectric [Check blue box]

rt4914:
Overall Update: The solution mentioned on the `developer.android.com` doc does work now for Espresso and it does give errors correctly.

@BenHenning Does this approach looks good for we need any changes?

app/build.gradle
```diff
@@ -150,6 +150,7 @@ dependencies {
   )
   androidTestImplementation(
       'androidx.test:core:1.2.0',
+      'androidx.test.espresso:espresso-accessibility:3.1.0',
```
BenHenning, line 153: Will this be needed for the Robolectric version, as well?

BenHenning (commented):
Thanks @rt4914. I think the approach looks good, though my main suggested edit would be to move the ``AccessibilityChecks.enable()`` bit to be part of OppiaTestRule since we want to normalize on using that rule everywhere, anyway, and it makes sense to always enable accessibility checks (even for domain layer changes, though it just won't actually produce any failures in those cases).

anandwana001 (commented):
https://github.com/oppia/oppia-android/blob/497263cad1a82a3ef35810196631509c919410ab/testing/src/main/java/org/oppia/android/testing/OppiaTestRule.kt#L18

I guess, we can enable the `AccessibilityChecks` inside the `apply`.
```kotlin
AccessibilityChecks.enable()
```

Now, whenever we need to have this in any test file, we can put the rule

```kotlin
  @get:Rule
  val oppiaTestRule = OppiaTestRule()
```

Learning - https://ncorti.com/blog/junit-rules

rt4914:
> https://github.com/oppia/oppia-android/blob/497263cad1a82a3ef35810196631509c919410ab/testing/src/main/java/org/oppia/android/testing/OppiaTestRule.kt#L18
> 
> I guess, we can enable the `AccessibilityChecks` inside the `apply`.
> 
> ```kotlin
> AccessibilityChecks.enable()
> ```
> 
> Now, whenever we need to have this in any test file, we can put the rule
> 
> ```kotlin
>   @get:Rule
>   val oppiaTestRule = OppiaTestRule()
> ```
> 
> Learning - https://ncorti.com/blog/junit-rules

Thanks @anandwana001 I will check this and update you.

rt4914:
Duplicate https://github.com/oppia/oppia-android/pull/3087

PR #2874, opened 2021-03-10
Fix #2826: Label and test added for ProfileResetPinActivity

### LINKED ISSUE #2826: [A11y] Add label for ProfileResetActivity
## Task
Introduce label for `ProfileResetActivity` as `Reset PIN`.

## General Explanation

### Example

4shutosh:
## Explanation

- Fixes Issue #2826 
- added label and test for the activity `ProfileResetPinActivity`

## Checklist

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -102,7 +102,8 @@
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.settings.profile.ProfileResetPinActivity"
-      android:theme="@style/OppiaThemeWithoutActionBar" />
+      android:theme="@style/OppiaThemeWithoutActionBar"
+      android:label="@string/profile_reset_pin_activity_label"/>
     <activity
       android:name=".app.splash.SplashActivity"
       android:screenOrientation="portrait"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -103,7 +103,7 @@
     <activity
       android:name=".app.settings.profile.ProfileResetPinActivity"
       android:theme="@style/OppiaThemeWithoutActionBar"
-      android:label="@string/profile_reset_pin_activity_label"/>
+      android:title="@string/profile_reset_pin_activity_title" />
     <activity
       android:name=".app.splash.SplashActivity"
       android:screenOrientation="portrait"
--- a/app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
+++ b/app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
@@ -31,7 +31,7 @@ class ProfileResetPinActivityPresenter @Inject constructor(
   }
 
   fun handleOnCreate() {
-    activity.title = activity.getString(R.string.profile_reset_pin_title)
+    activity.title = activity.getString(R.string.profile_reset_pin_activity_title)
     activity.supportActionBar?.setDisplayHomeAsUpEnabled(true)
     activity.supportActionBar?.setHomeAsUpIndicator(R.drawable.ic_arrow_back_white_24dp)
 
--- a/app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
+++ b/app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
@@ -31,7 +31,6 @@ class ProfileResetPinActivityPresenter @Inject constructor(
   }
 
   fun handleOnCreate() {
-    activity.title = activity.getString(R.string.profile_reset_pin_activity_title)
     activity.supportActionBar?.setDisplayHomeAsUpEnabled(true)
     activity.supportActionBar?.setHomeAsUpIndicator(R.drawable.ic_arrow_back_white_24dp)
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -300,8 +300,7 @@
   <string name="profile_rename_label">New Name</string>
   <string name="profile_rename_save">save</string>
   <!-- ProfileResetPinActivity -->
-  <string name="profile_reset_pin_activity_label">Reset PIN</string>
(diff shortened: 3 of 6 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -102,7 +102,8 @@
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.settings.profile.ProfileResetPinActivity"
-      android:theme="@style/OppiaThemeWithoutActionBar" />
+      android:theme="@style/OppiaThemeWithoutActionBar"
+      android:label="@string/profile_reset_pin_activity_label"/>
```
rt4914: Instead of `label` use suffix `title`

Also, add space before `/>`

app/src/main/res/values/strings.xml
```diff
@@ -300,6 +300,7 @@
   <string name="profile_rename_label">New Name</string>
   <string name="profile_rename_save">save</string>
   <!-- ProfileResetPinActivity -->
+  <string name="profile_reset_pin_activity_label">Reset PIN</string>
   <string name="profile_reset_pin_title">Reset PIN</string>
```
rt4914: Rename `profile_reset_pin_activity_label` to `profile_reset_pin_activity_title` and also remove `profile_reset_pin_title ` and use `profile_reset_pin_activity_label` whereever the string was getting used.

app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityTest.kt
```diff
@@ -90,6 +94,11 @@ import javax.inject.Singleton
 )
 class ProfileResetPinActivityTest {
 
+  @get:Rule
+  var activityTestRule: ActivityTestRule<ProfileResetPinActivity> = ActivityTestRule(
```
rt4914: Use `val` instead of `var`

rt4914 (changes requested):
Update your PR title to mention issue number.
`Fix #2826 : Label and test added for ProfileResetPinActivity`

app/src/main/AndroidManifest.xml
4shutosh: Okay,
Can you please explain the difference between the usage of label and title?
Also, there are some activities which are using `label` instead of `title` in a similar manner (the reference PR itself has `android:label` ), should we change that as well introducing an issue?
Why are we referencing it as a label in the issue title itself, when we are supposed to introduce a title?
Thank you for your help.

rt4914: Actually your answer itself address the renaming issue upto an extent.
So basically we want to make sure that naming of strings is consistent. We started with labels initially but `label` is used for hints inside edit text too. and therefore we want to shift to `title`.

Don't change for other activities right now. Just fix for your PR others will be fixed in separate PRs.

rt4914:
@4shutosh Unable to see any commit which shows that you have done the suggestions. Once it is ready to comment with my user name.

4shutosh:
Done @rt4914, Please review.

app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityTest.kt
```diff
@@ -120,6 +131,30 @@ class ProfileResetPinActivityTest {
     ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
   }
 
+  @Test
+  fun testProfileResetPinActivity_hasCorrectActivityLabel() {
+    activityTestRule.launchActivity(
+      createProfileResetPinActivityIntent(
+        internalProfileId,
```
rt4914, line 138: Just one nit. Add named argument.

`internalProfileId = internalProfileId,`

rt4914 (approved):
LGTM, thanks.

app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
```diff
@@ -31,7 +31,7 @@ class ProfileResetPinActivityPresenter @Inject constructor(
   }
 
   fun handleOnCreate() {
-    activity.title = activity.getString(R.string.profile_reset_pin_title)
+    activity.title = activity.getString(R.string.profile_reset_pin_activity_title)
```
BenHenning: @rt4914 is this still needed if it's being set in XML? It seems unnecessary. @4shutosh I think you can remove this.

app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityTest.kt
BenHenning: This actually shouldn't be needed, I think. Named arguments generally provide benefit when contextualizing literals, but in this case 'internalProfileId' is already giving us all the context that's needed.

BenHenning (commented):
Thanks @4shutosh! Just had a couple of comments.

4shutosh:
Is there anything else needed @rt4914? Thanks

app/src/main/java/org/oppia/android/app/settings/profile/ProfileResetPinActivityPresenter.kt
```diff
@@ -31,7 +31,6 @@ class ProfileResetPinActivityPresenter @Inject constructor(
   }
 
   fun handleOnCreate() {
-    activity.title = activity.getString(R.string.profile_reset_pin_title)
```
rt4914, line 34: I checked the app again and it seems that we cannot remove this. This is the reason why your tests are failing. Add this back and tests should pass.

rt4914 (changes requested):
@4shutosh Suggested change.

rt4914:
@4shutosh You will need to create a new PR on a new branch.

I will need to close this PR as you have done a **force-push**. Please never do a force push as it tampers the commit history.
It is mentioned in the wiki too. PTAL: https://github.com/oppia/oppia-android/wiki#address-review-comments-until-all-reviewers-give-lgtm-looks-good-to-me

4shutosh:
Thank you @rt4914, will do the required.

PR #2878, opened 2021-03-10
Fix #2876: Create domain controller for feedback reporting

### LINKED ISSUE #2876: Create feedback reporting management controller
Fix part of #53
Create the controller that will manage storing and uploading the feedback reports, stubbing portions that require connections between the app layer and the domain service

[Design doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

jcqli:
## Explanation
Fix #2876
- Creates the domain controller that collects the feedback report information and either stores it on local disk or pushes it to remote storage. Defines functions to store, fetch, and remove reports from local storage, and send reports to the remote service.
- Adds a new data class object to represent the entry point.
- Add a language code extension to switch between app language, audio language, and a generic language code. No tests were added for the extensions as it seemed trivial to check hard-coded string mappings.
- Add an annotation class for the current report schema version and tests to validate the schema version and expected resort data.
- Add a 'platform' field to GaeFeedbackReport so backend schema validators can properly generalize to both report types.

Feature tracker: #53
[Design doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

## Checklist


domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,289 @@
...
+  @Qualifier
+  annotation class OppiaRetrofit
+
+  // TODO(#89): Move this to a common test application component.
+  @Module
+  class TestNetworkModule {
+    @Provides
+    @Singleton
+    fun provideFeedbackReportingService(
+      @OppiaRetrofit retrofit: Retrofit
```
BenHenning: I think your compiler error is due to there not being any provider for this value (e.g. nothing is providing an OppiaRetrofit-qualified Retrofit object).

BenHenning:
@jcqli spent some time looking into the failures. The test via Bazel was experiencing a runtime failure due to not having access to the data test assets (for API mocks), and the test via Gradle was experiencing a build error due to the data module not running the Dagger compiler. The following gist should fix both: https://gist.github.com/BenHenning/da7d1eeb28360619c56ecc91367d4197/revisions. Confirming all non-app tests pass locally for me now on this branch via Bazel, and domain tests pass via Gradle.

Also, note that your test ``testController_submitFeedbackReport_withNetwork_sendsNetworkRequest_isSuccessful`` takes almost 10 seconds to run for me--that seems quite long. It's also not Robolectric arrangement since the first test took the expected ~8 seconds to start up. It may be worth looking into why this test in particular is taking so long.

jcqli:
> Also, note that your test `testController_submitFeedbackReport_withNetwork_sendsNetworkRequest_isSuccessful` takes almost 10 seconds to run for me--that seems quite long. It's also not Robolectric arrangement since the first test took the expected ~8 seconds to start up. It may be worth looking into why this test in particular is taking so long.

@BenHenning  I did a bit of guess and check and I think it's because of the timeout specified -- when I shorten the timeout amount the test runs faster, and when I increase it slows down accordingly. Might warrant some more research, though since I'm not sure if this is sound reasoning.

jcqli:
Hm I can't reproduce the bazel build error locally but it doesn't seem like a flake. I tried `bazel clean` and then rebuilding it but it looks good on mine, @BenHenning could you try reproducing this?

EDIT: I guess this might be a maven / network issue?

BenHenning:
Sorry, been really busy with non-work items this week. Will need to take a look later in the week.

BenHenning:
> > Also, note that your test `testController_submitFeedbackReport_withNetwork_sendsNetworkRequest_isSuccessful` takes almost 10 seconds to run for me--that seems quite long. It's also not Robolectric arrangement since the first test took the expected ~8 seconds to start up. It may be worth looking into why this test in particular is taking so long.
> 
> @BenHenning I did a bit of guess and check and I think it's because of the timeout specified -- when I shorten the timeout amount the test runs faster, and when I increase it slows down accordingly. Might warrant some more research, though since I'm not sure if this is sound reasoning.

I wouldn't expect the timeout to make a difference--that is a bit odd.



> Hm I can't reproduce the bazel build error locally but it doesn't seem like a flake. I tried `bazel clean` and then rebuilding it but it looks good on mine, @BenHenning could you try reproducing this?
> 
> EDIT: I guess this might be a maven / network issue?

@jcqli I think you're running into #3168. Syncing to the latest develop should fix CI for you.

BenHenning:
@jcqli does this PR have unrelated changes in it from earlier PRs? Some of the data module changes seem extra--could you sync to the latest develop and send back?

jcqli:
> I wouldn't expect the timeout to make a difference--that is a bit odd.

Now I can't reproduce this behavior, the test `testController_submitFeedbackReport_withNetwork_successfullySendsRequestToServer` takes about 4-5 seconds regardless of timeout on Gradle, is there a way to see individual run-times with Bazel?

> @jcqli does this PR have unrelated changes in it from earlier PRs? Some of the data module changes seem extra--could you sync to the latest develop and send back?

The data module Bazel file changes are largely from the earlier gist that resolved the build errors, but the GAE models were more updates to match the field names of the backend models -- let me know if they should be separated into a different PR.

jcqli:
There are linter errors with the bazel files that I'll need to comb through but assigning for a review since it's not blocking anything.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
```diff
@@ -10,8 +10,10 @@ data class GaeFeedbackReport(
 
   /** The current version of the feedback report used. */
   @Json(name = "android_report_info_schema_version") val schemaVersion: Int,
-  /** The date and time in ms that the report was created. */
+  /** The date and time in sec that the report was submitted. */
   @Json(name = "report_submission_timestamp_sec") val reportSubmissionTimestampSec: Int,
```
rt4914, line 16: Should this be `Long` instead of `Int`?

rt4914 (approved):
Checked all `Gae*` files w.r.t json file and everything looks correct. Approving for code-owner files.

model/BUILD.bazel
```diff
@@ -227,6 +228,8 @@ android_library(
     visibility = ["//visibility:public"],
     exports = [
         ":arguments_java_proto_lite",
+        ":profile_java_proto_lite",
```
fsharpasharp: Don't add these, since we're removing the model library.

model/BUILD.bazel
```diff
@@ -165,6 +165,7 @@ format_import_proto_library(
 
 java_lite_proto_library(
     name = "feedback_reporting_java_proto_lite",
+    visibility = ["//visibility:public"],
```
fsharpasharp: Use //:oppia_api_visibility, if possible

utility/src/main/java/org/oppia/android/util/extensions/BUILD.bazel
```diff
@@ -26,3 +26,15 @@ kt_android_library(
...
 )
+
+kt_android_library(
+    name = "language_code_extensions",
+    srcs = [
+        "LanguageCodeExtensions.kt",
+    ],
+    visibility = ["//:oppia_api_visibility"],
+    deps = [
+        "//model",
```
fsharpasharp: Don't use the //model dependency. Add the underlying dependencies directly.

data/BUILD.bazel
```diff
@@ -6,6 +6,16 @@ This library provides data to the rest of the application.
...
 
+android_library(
+    name = "data_test_assets",
+    testonly = True,
+    assets = glob(["src/test/assets/**"]),
+    assets_dir = "src/test/assets/",
+    visibility = ["//:oppia_testing_visibility"],
+    custom_package = "org.oppia.android.data",
+    manifest = "src/test/AndroidManifest.xml",
+)
```
anandwana001, line 17: Bazel Lint Fix

```Bazel
android_library(
    name = "data_test_assets",
    testonly = True,
    assets = glob(["src/test/assets/**"]),
    assets_dir = "src/test/assets/",
    custom_package = "org.oppia.android.data",
    manifest = "src/test/AndroidManifest.xml",
    visibility = ["//:oppia_testing_visibility"],
)

```

domain/BUILD.bazel
```diff
@@ -72,12 +72,16 @@ kt_android_library(
     visibility = ["//visibility:public"],
     deps = [
         ":dagger",
+        "//data/src/main/java/org/oppia/android/data/backends/gae/api",
+        "//data/src/main/java/org/oppia/android/data/backends/gae/model:model",
```
anandwana001: Bazel lint is failing for this and the Buildifier Auto formatter is formating it like this
`"//data/src/main/java/org/oppia/android/data/backends/gae/model",`

Do we need this `"//data/src/main/java/org/oppia/android/data/backends/gae/model:model",` as it is or we can convert it to `"//data/src/main/java/org/oppia/android/data/backends/gae/model",` ?

If not, we need to check into the Buildifier to prevent such a case from failing.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,445 @@
...
+    val report = reportStoreResultCaptor.value.getOrThrow().getReports(0)
+    assertThat(report.reportSubmissionTimestampSec).isEqualTo(LATER_TIMESTAMP)
+    assertThat(report.userSuppliedInfo.suggestion.suggestionCategory)
+      .isEqualTo(SuggestionCategory.LANGUAGE_SUGGESTION)
+    assertThat(report.userSuppliedInfo.suggestion.userSubmittedSuggestion)
+      .isEqualTo(languageSuggestionText)
+  }
+
+  @Test
+  fun testController_submitMultipleFeedbackReports_withoutNetwork_savesAllReportsToStore() {
```
anandwana001, line 317: Should we add 1 more test for `submitMultipleFeedbackReports_withNetwork`?

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,445 @@
...
+      ApplicationInfoBuilder.newBuilder()
+        .setPackageName(context.packageName)
+        .build()
+    val packageInfo =
+      PackageInfoBuilder.newBuilder()
+        .setPackageName(context.packageName)
+        .setApplicationInfo(applicationInfo)
+        .build()
+    packageInfo.versionName = "1.0"
+    packageInfo.versionCode = 1
```
anandwana001: Should we try `packageInfo.longVersionCode = 1` as `versionCode` is deprecated. 
Check `PackageInfoCompat` class.

anandwana001 (changes requested):
looked at the failed static check and test file. Left few thoughts.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReport.kt
jcqli: This is defined as an int since the backend timestamp for seconds uses an int rather than a double so I kept the consistency. The long => int conversion could happen later, though, (i.e. this could remain a Long and then we perform the conversion in the backend controller) -- would that be preferred? I don't see if there's any benefit to keeping this a Long.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
jcqli: Ack -- I initially left this case out since it's the same as `testController_submitFeedbackReport_withNetwork_doesNotSaveReportToStore` multiple times but this might catch something.
Test added.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
jcqli: Done, updated this to be a `Long`. Fixed the proto and extension class as well.

jcqli:
I'm no longer able to run tests on Gradle -- is anyone else able to repro this? It looks to build fine on both Gradle and Bazel, but I can't run the tests in Gradle. Cleaning and rebuilding didn't help.
Screen Shot 2021-05-17 at 9 14 31 AM

domain/BUILD.bazel
jcqli: Ah thanks for catching this! Done, changed to `//data/src/main/java/org/oppia/android/data/backends/gae/model`

data/BUILD.bazel
jcqli: Done

utility/src/main/java/org/oppia/android/util/extensions/BUILD.bazel
jcqli: Done

model/BUILD.bazel
jcqli: Done

model/BUILD.bazel
jcqli: Done

rt4914:
@jcqli I checked on latest develop and I was able to run the app layer and domain layer tests locally on gradle.
Screenshot 2021-05-18 at 5 19 27 PM

rt4914:
@jcqli I rechecked for robolectric tests and it was failing. It is fixed in latest develop now. Reference: #3196

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,466 @@
...
+      unit = testCoroutineDispatcher.DEFAULT_TIMEOUT_UNIT
+    )
+    assertThat(request).isNull()
+  }
+
+  @Test
+  fun testController_submitFeedbackReport_withNetwork_doesNotSaveReportToStore() {
+    // Enqueue a responses so that the MockWebServer knows when the request is successfully sent.
+    mockWebServer.enqueue(MockResponse().setBody("{}"))
+    networkConnectionUtil.setCurrentConnectionStatus(LOCAL)
```
anandwana001, line 199: Should we add test cases for `CELLULAR` type connections?

anandwana001, line 219: Should we add test cases for `CELLULAR` type connections?

anandwana001 (changes requested):

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
jcqli: Done

model/BUILD.bazel
```diff
@@ -227,6 +228,7 @@ android_library(
     visibility = ["//visibility:public"],
     exports = [
         ":arguments_java_proto_lite",
+        ":question_java_proto_lite",
```
fsharpasharp: Remove this line as well.

anandwana001 (approved):
Test files LGTM

model/BUILD.bazel
jcqli: Done

fsharpasharp (approved):
BUILD.bazel files look good to me.

vinitamurthi (approved):

utility/src/main/java/org/oppia/android/util/extensions/ContextExtensions.kt
```diff
@@ -10,8 +10,13 @@ fun Context.getVersionName(): String {
 }
 
 /** Returns the current app version code. */
-fun Context.getVersionCode(): Int {
-  return this.packageManager.getPackageInfo(this.packageName, /* flags= */ 0).versionCode
+fun Context.getLongVersionCode(): Long {
```
BenHenning: Do we really need to use a long? The version code is fixed for a particular binary, and I suspect it's very unlikely for us to exceed 2 billion builds of the app anytime soon (especially considering we don't auto-increment the version code yet).

Though, if the new API only provides a long then maybe just call this 'getVersionCode' since the type distinction isn't that important outside the package manager.

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
```diff
@@ -7,7 +7,7 @@ import retrofit2.http.POST
 
 /** Service that posts feedback reports to remote storage. */
 interface FeedbackReportingService {
-  @POST("android_feedback_reporting/incoming_report")
+  @POST("appfeedbackreporthandler/incoming_android_report")
```
BenHenning, line 10: Just to confirm: are this + the model objects up-to-date with the latest version of the backend handler/were there any changes to that before that haven't yet been reflected here?

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -98,6 +98,7 @@ class RemoteAuthNetworkInterceptorTest {
 
   @Test
   fun testNetworkInterceptor_withIncorrectHeaders_setsCorrectHeaders() {
+    mockWebServer.enqueue(MockResponse().setBody("{}"))
```
BenHenning, line 101: For context, why this change?

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+import javax.inject.Singleton
+
+private const val FEEDBACK_REPORT_MANAGEMENT_CONTROLLER_TAG =
+  "Feedback Report Management Controller"
+private const val FEEDBACK_REPORTS_DATABASE_NAME = "feedback_reports_database"
+private const val PLATFORM_ANDROID = "android"
+
+/** Controller for uploading feedback reports to remote storage or saving them on disk. */
+@Singleton
+class FeedbackReportManagementController @Inject constructor(
```
BenHenning, line 48: For this & its test, could you define a separate kt_android_library & test target at the package level for each? For the most part, we're trying to move toward package-specific build packages now for modules outside of app/ to avoid continuing to move the modularization target for the Bazel migration.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+  }
+
+  /**
+   * Gets any feedback reports saved locally on the device. This is called by a
+   * [FeedbackReportUploadWorker] when it detects a network reconnection to send all local reports
+   * to remote storage.
+   *
+   * @return a list of feedback reports to upload to remote storage
+   */
+  suspend fun getCachedReportsList(): MutableList<FeedbackReport> {
```
BenHenning: We probably shouldn't return something that's mutable, especially in a suspend context since it can make data races pretty hard to avoid. Can we just return a standard List here, instead?

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+  private fun createGaeUserSuppliedFeedback(
+    userSuppliedFeedback: UserSuppliedFeedback
+  ): GaeUserSuppliedFeedback {
+    val reportType = userSuppliedFeedback.reportTypeCase
+    when (reportType) {
+      ReportTypeCase.SUGGESTION -> {
+        val suggestion = userSuppliedFeedback.suggestion
+        return GaeUserSuppliedFeedback(
+          reportType = reportType.name.toLowerCase(),
+          category = suggestion.suggestionCategory.name.toLowerCase(),
```
BenHenning: We ought to not convert enums to strings like this. I actually suggest hardcoding them or using constants to ensure the exact correct string that matches the backend is used (otherwise, renaming the enum will unexpectedly break this flow).

Also, we should pass in a Locale for toLowerCase(). Generally, it should be the user's locale if the string is user-supplied, or otherwise the appropriate locale for the string if it's not user-supplied.

Ditto elsewhere (for both).

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+        val crash = userSuppliedFeedback.crash
+        return GaeUserSuppliedFeedback(
+          reportType = reportType.name.toLowerCase(),
+          category = crash.crashLocation.name.toLowerCase(),
+          userFeedbackSelectedItems = null,
+          userFeedbackOtherTextInput = crash.crashExplanation
+        )
+      }
+      else -> throw IllegalArgumentException(
+        "Encountered unexpected feedback report type: ${userSuppliedFeedback.reportTypeCase.name}"
```
BenHenning:
```suggestion
        "Encountered unexpected feedback report type: ${userSuppliedFeedback.reportTypeCase}"
```

Optional: the enum-to-string conversion should keep this error message the way you intend it to be.

Ditto elsewhere.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+    )
+  }
+
+  // Helper function that construct a Moshi data class object for the device's system information.
+  private fun getSystemContext(
+    systemContext: FeedbackReportingSystemContext
+  ): GaeFeedbackReportingSystemContext {
+    return GaeFeedbackReportingSystemContext(
+      packageVersionName = systemContext.packageVersionName,
+      packageVersionCode = systemContext.packageVersionCode,
```
BenHenning, line 260: This one doesn't use the newly updated longVersionCode method?

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+        val revisionCard = appContext.revisionCard
+        topicId = revisionCard.topicId
+        subtopicId = revisionCard.subtopicId
+      }
+      EntryPointCase.NAVIGATION_DRAWER -> {
+        // If entry point is not an exploration player or revision card, leave story values as null.
+      }
+      EntryPointCase.CRASH_DIALOG -> {
+        // If entry point is not an exploration player or revision card, leave story values as null.
+      }
```
BenHenning:
```suggestion
      EntryPointCase.NAVIGATION_DRAWER, EntryPointCase.CRASH_DIALOG -> {
        // If entry point is not an exploration player or revision card, leave story values as null.
      }
```

Can be simplified.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+      topicId = topicId,
+      storyId = storyId,
+      explorationId = explorationId,
+      subtopicId = subtopicId
+    )
+  }
+
+  // Helper function to retrieve the logcat logs from the device.
+  private fun getLogcatLogs(): List<String> {
+    val logcatReader = consoleLogger.getLogReader()
```
BenHenning, line 334: We should make sure that this gets closed.

That being said, you can probably simplify this entire method & ensure the reader gets closed:

```kotlin
return consoleLogger.getLogReader().readLines()
```

I suspect the IOException isn't needed. If it comes up that the file doesn't exist, consider changing the API for ConsoleLogger to return the list of lines & return an empty list if the file doesn't exist (which can be checked on the File object directly rather than trying to open & read a BufferedReader).

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+    analyticsController.getEventLogStore().toLiveData()
+  }
+
+  private val eventsLogListLiveData = Transformations.map(oppiaEventLogsLiveData) {
+    it.getOrDefault(OppiaEventLogs.getDefaultInstance())
+  }
+
+  private val eventsLogList = Transformations.map(eventsLogListLiveData) { oppiaEventLogs ->
+    oppiaEventLogs.eventLogList.map { it.toString() }
+  }.value
```
BenHenning, line 71: Does this actually work? I wouldn't expect it to since DataProvider-converted-LiveData should never provide a value unless it has an observer.

In general, we never want to do this. LiveData should only be used by the frontend to safely observe async data channels. Instead, we should return a Deferred or DataProvider from AnalyticsController so that it can be observed or transformed in this class. I suspect that, since you want the value of the operation you'll probably need to use a suspend function to retrieve the result directly from the DataProvider. To do that, you'll need to inject a background dispatcher and use it to launch a new async task to provide you with a suspend context.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -617,6 +617,7 @@ exempted_file_path: "utility/src/main/java/org/oppia/android/util/data/DataProvi
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/data/DataProvidersInjector.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/data/DataProvidersInjectorProvider.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/extensions/ContextExtensions.kt"
+exempted_file_path: "utility/src/main/java/org/oppia/android/util/extensions/LanguageCodeExtensions.kt"
```
BenHenning, line 667: Could we add tests for this file instead of exempting it?

utility/src/main/java/org/oppia/android/util/extensions/LanguageCodeExtensions.kt
```diff
@@ -0,0 +1,60 @@
+package org.oppia.android.util.extensions
```
BenHenning, line 1: This file is introducing more duplication against the enum definitions in profile.proto & the mapping defined in OptionControlsViewModel.kt. If we're going to define one mapping, can we do it in one place (ideally here)?

utility/src/main/java/org/oppia/android/util/extensions/LanguageCodeExtensions.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import org.oppia.android.app.model.AppLanguage
+import org.oppia.android.app.model.AudioLanguage
+
+// Extension functions for language codes that convert between the different types of language codes
+// and their String representations.
+
+enum class LanguageCode {
```
BenHenning, line 9: Please add KDocs for all public classes/functions/members in this file.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+import java.io.File
+import javax.inject.Inject
+import javax.inject.Singleton
+
+// Timestamp in sec for 3/2/21, 12:38pm GMT.
+private const val EARLIER_TIMESTAMP = 1614688684
```
BenHenning, line 74: Nit: for this & next one, please add unit to the field names.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+
+  private val earlierCrashReport = FeedbackReport.newBuilder()
+    .setReportSubmissionTimestampSec(EARLIER_TIMESTAMP)
+    .setUserSuppliedInfo(userSuppliedCrashFeedback)
+    .setAppContext(appContext)
+    .build()
+
+  @Before
+  fun setUp() {
+    networkConnectionUtil = NetworkConnectionUtil(ApplicationProvider.getApplicationContext())
```
BenHenning: This should probably be injected, instead.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+    setUpTestApplicationComponent()
+    setUpApplicationForContext()
+    setUpFakeLogcatFile()
+    MockitoAnnotations.initMocks(this)
```
BenHenning, line 150: Can we use the MockitoJUnit rule instead of manually initing mocks?

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+    // format used in the request body of reports sent in network requests.
+    val feedbackReportUnformattedJson = ApiMockLoader.getFakeJson(
+      "feedback_reporting_unformatted.json"
+    )
+
+    val request = mockWebServer.takeRequest(
+      timeout = testCoroutineDispatcher.DEFAULT_TIMEOUT_SECONDS,
+      unit = testCoroutineDispatcher.DEFAULT_TIMEOUT_UNIT
+    )
+    request?.let {
```
BenHenning, line 173: Can we make this a bit stricter by asserting that the request is non-null? Otherwise, if the test goes wrong & this returns null then I think it will still pass (incorrectly).

Ditto for other tests.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+    feedbackReportManagementController.submitFeedbackReport(laterSuggestionReport)
+
+    val reportsStore = feedbackReportManagementController.getFeedbackReportStore().toLiveData()
+    reportsStore.observeForever(mockReportsStoreObserver)
+    testCoroutineDispatchers.advanceUntilIdle()
+
+    verify(mockReportsStoreObserver, atLeastOnce()).onChanged(reportStoreResultCaptor.capture())
+
+    val reportsList = reportStoreResultCaptor.value.getOrThrow().reportsList
+    assertThat(reportsList.size).isEqualTo(2)
```
BenHenning, line 329:
```suggestion
    assertThat(reportsList).hasSize(2)
```

Bit more idiomatic.

domain/src/test/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementControllerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+    assertThat(reportsList.get(0)).isEqualTo(earlierCrashReport)
+    assertThat(reportsList.get(1)).isEqualTo(laterSuggestionReport)
+  }
+
+  @Test
+  fun testController_removeCachedReport_noLongerInStore() {
```
BenHenning, line 335: Other potential test scenario: calling removeFirstCachedReport() without anything being present.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+  fun getFeedbackReportStore(): DataProvider<FeedbackReportingDatabase> {
+    return feedbackReportDataStore
+  }
+
+  /**
+   * Sends an individual [FeedbackReport] to remote storage.
+   *
+   * @param report a [FeedbackReport] to upload
+   */
+  fun uploadFeedbackReport(report: FeedbackReport) {
```
BenHenning, line 112: Should this be private? It seems like we'd always want reports to go through submitFeedbackReport().

If not (particularly because we want to log things through the worker), then we should probably have a few targeted tests that verify this always forces a send regardless of internet connectivity (or put another way: the caller is responsible for calling this in the correct circumstances).

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+  private fun storeFeedbackReport(report: FeedbackReport) {
+    feedbackReportDataStore.storeDataAsync(updateInMemoryCache = true) { feedbackReportDatabase ->
+      return@storeDataAsync feedbackReportDatabase.toBuilder().addReports(report).build()
+    }.invokeOnCompletion {
+      it?.let {
+        consoleLogger.e(
+          FEEDBACK_REPORT_MANAGEMENT_CONTROLLER_TAG,
+          "Failed to store feedback report",
+          it
+        )
```
BenHenning, line 152: Are any of these failure cases covered in tests? Consider that the cache file could become corrupted during operations which would result in some of these error cases.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+      }
+      else -> throw IllegalArgumentException(
+        "Encountered unexpected feedback report type: ${userSuppliedFeedback.reportTypeCase.name}"
+      )
+    }
+  }
+
+  // Helper function that construct a Moshi data class object for the information provided by the user,
+  // based on the type of Issue report.
+  private fun createGaeUserSuppliedFeedbackForIssue(
```
BenHenning, line 193: Just to check: for each of these helper functions, are all of the ``when`` cases covered via tests?

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
```diff
@@ -0,0 +1,346 @@
...
+      entryPointName = appContext.entryPointCase.name.toLowerCase(),
+      topicId = topicId,
+      storyId = storyId,
+      explorationId = explorationId,
+      subtopicId = subtopicId
+    )
+  }
+
+  // Helper function to retrieve the logcat logs from the device.
+  private fun getLogcatLogs(): List<String> {
```
BenHenning, line 333: I suggest testing multiple logcat lines (unless I'm mistaken, I think your current test data only covers a single line which might miss some of the logic being employed here).

BenHenning (changes requested):
Thanks so much for the PR, and your patience in my very delayed review @jcqli! I've taken a full pass and left some comments--PTAL and let me know if you have any questions or would like clarity on any of my suggestions.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: For parity with the other test case and reading clarity -- it's setting up the test / networking environment at the beginning rather than in the middle.

data/src/main/java/org/oppia/android/data/backends/gae/api/FeedbackReportingService.kt
jcqli: Yep ([backend PR](https://github.com/oppia/oppia/pull/12367/files#diff-f264a7b7fd28a5d22210dd5ea1d533cc79f2d650db099daeffa4608f393d4e9cR55))

utility/src/main/java/org/oppia/android/util/extensions/ContextExtensions.kt
jcqli: `versionCode` is deprecated so we only have the long version (I presume we don't want to use old APIs) -- changed the function name.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: Done

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: Done, using constants.

Re: locale, would it just be US English? To be readable for maintainers, this wouldn't be user-supplied but it is being determined from the device or app.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: Done

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: This is getting the version code from a proto, we would use the extension function to create the proto when the report is originally created.

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: Done

domain/src/main/java/org/oppia/android/domain/feedbackreporting/FeedbackReportManagementController.kt
jcqli: Done

BenHenning:
@jcqli do you have an idea on when you might be able to revive this PR?

jcqli:
@BenHenning I'll get some changes and reviewed again next week! Thanks for the patience.

BenHenning:
No problem, and sounds good. Thanks!

PR #2890, opened 2021-03-11
Fix #2711: Improve workflow cancellation support in CI

### LINKED ISSUE #2711: Workflow canceller doesn't seem to work for dynamic jobs
We kick-off >130 jobs for Bazel tests due to sharding, but it seems that subsequent pushes to a branch triggering the workflow canceller does not cancel previous jobs if those jobs are started due to a dynamic matrix strategy. This is especially troubling since the number of jobs (130) will quickly overwhelm GitHub Actions resources even with #2710 checked in.

BenHenning:
Fix #2711.

Per https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks functionality was recently added (https://github.com/styfle/cancel-workflow-action/pull/55) to:
1. Allow cancellation based on workflows starting (regardless of where they start)
2. Remove the need for the access token, allowing proper fork support (I actually thought this wasn't an issue, but maybe we didn't realize it before)

Both should substantially reduce actions time since subsequent commits should properly cancel everything previous run, both for forks & non-forks. I'm expecting this to fix #2711 but I will need to verify in this PR.

### DIFF
--- a/.github/workflows/workflow_canceller.yml
+++ b/.github/workflows/workflow_canceller.yml
@@ -9,20 +9,17 @@ name: Automatic Workflow Canceller
 # hash of the tip of the branch to ensure it doesn't cancel previous workflows that aren't related
 # to the branch being evaluated.
 on:
-  workflow_dispatch:
-  pull_request:
-  push:
-    branches:
-      # Push events on develop branch
-      - develop
+  workflow_run:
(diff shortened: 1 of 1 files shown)


BenHenning:
Hmm it seems like the canceller is not running (all cancelled workflows in the last 2 commits were manual). Will need to dig into why.

PR is probably going to be put on hold for a few days (until I get a brief moment to be able to get back to it).

PR #2892, opened 2021-03-12
Fix #2891: Refactored AnswerGroup misconception to be object instead of string

### LINKED ISSUE #2891: Update AnswerGroup proto to parse misconception ID into object rather than string
In our current AnswerGroup proto structure, we have a string field called 'tagged_skill_misconception_id'. This is because the misconception ID is formatted such that it is '<skill ID>-<misconception ID>'. This introduces the [stringly typed antipattern](https://wiki.c2.com/?StringlyTyped), so we want to instead store the misconception ID as an object that more clearly indicates its constituent parts, such as:
``
message Misconception {
  string skill_id = 1;
  string misconception_id = 2;
}
``

This improvement will require refactoring inside the `StateRetriever` class (which parses JSON objects that we receive), as well as any other places which are currently using misconception ID.

TheRealJessicaLi:
Refactored AnswerGroup proto structure to use a Misconception proto object to store the misconception and skill IDs as separate fields instead of concatenating them into one string, as per discussion [here](https://github.com/oppia/oppia-android/pull/2802#discussion_r592018202). Fixes #2891.

## Checklist


data/src/main/java/org/oppia/android/data/backends/gae/model/GaeMisconception.kt
```diff
@@ -0,0 +1,15 @@
+package org.oppia.android.data.backends.gae.model
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Data class for Misconception model
+ */
```
BenHenning:
```suggestion
/** Data class for misconceptions. */
```

Can be made a bit more concise (both in language & formatting).

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeAnswerGroup.kt
```diff
@@ -10,7 +10,7 @@ import com.squareup.moshi.JsonClass
 @JsonClass(generateAdapter = true)
 data class GaeAnswerGroup(
 
-  @Json(name = "tagged_skill_misconception_id") val taggedSkillMisconceptionId: String?,
+  @Json(name = "tagged_skill_misconception_id") val taggedSkillMisconceptionId: GaeMisconception?,
```
BenHenning: Is this change correct? I'm assuming the backend will still continue to send a string. We would need to refactor the backend itself to send a structured object; I think keeping the structure local to the Android client for now is sufficient.

domain/src/main/java/org/oppia/android/domain/classify/AnswerClassificationController.kt
```diff
@@ -59,12 +59,13 @@ class AnswerClassificationController @Inject constructor(
         try {
           if (ruleClassifier.matches(answer, ruleSpec.inputMap)) {
             // Explicit classification matched.
-            return if (answerGroup.taggedSkillMisconceptionId.isEmpty()) {
+            return if (answerGroup.taggedSkillMisconceptionId.skillId.isNullOrEmpty()) {
```
BenHenning: Maybe a bit clearer:

```suggestion
            return if (answerGroup.hasTaggedSkillMisconception()) {
```

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -57,15 +57,11 @@ internal class QuestionAssessmentProgress {
...
+  internal fun trackAnswerSubmitted(taggedSkillId: String?) {
     val currentQuestionIndex = getCurrentQuestionIndex()
     questionSessionMetrics[currentQuestionIndex].numberOfAnswersSubmitted++
-    if (taggedSkillMisconceptionId != null) {
-      // Misconceptions are formatted as <skill ID>-<misconception ID> from the AnswerClassificationController,
-      // but we only need the related skill ID for our purposes
-      questionSessionMetrics[currentQuestionIndex].taggedSkillMisconceptionIds.add(
-        taggedSkillMisconceptionId.split("-")[0]
-      )
+    if (taggedSkillId != null) {
```
BenHenning, line 63: Can this actually ever be null?

domain/src/main/java/org/oppia/android/domain/question/QuestionSessionMetrics.kt
```diff
@@ -15,9 +15,9 @@ internal data class QuestionSessionMetrics(val question: Question) {
   var didViewSolution: Boolean = false
 
   /**
-   * Tracks the skill misconception IDs corresponding to the user's wrong answers for this question.
-   * A skill misconception ID is a skill ID which identifies which skill the user needs to work on,
-   * based on which misconception the user had when they submitted a wrong answer for a question.
+   * Tracks the skill IDs corresponding to the misconceptions of the user's wrong answers for this
+   * question. These skill IDs identify which skills the user needs to work on, based on which
+   * misconceptions the user had when they submitted wrong answers for this question.
```
BenHenning, line 20: Nice KDoc--this is easy to understand (and easier than the previous version, I think).

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -171,7 +172,18 @@ class StateRetriever @Inject constructor(
           answerGroupJson.optJSONObject("rule_types_to_inputs"), interactionId
         )
       )
-      .build()
+
+    if (answerGroupJson.optJSONObject("tagged_skill_misconception_id") != null) {
```
BenHenning: Suggest factoring the retrieved JSON object into a local val to avoid having to re-grab it. I think that will simplify the implementation here some.

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -171,7 +172,18 @@ class StateRetriever @Inject constructor(
...
+      answerGroup.setTaggedSkillMisconceptionId(
+        Misconception.newBuilder().setSkillId(
+          answerGroupJson.getJSONObject("tagged_skill_misconception_id").toString().split("-")[0]
+        )
+          .setMisconceptionId(
+            answerGroupJson.getJSONObject("tagged_skill_misconception_id").toString().split("-")[1]
+          )
```
BenHenning: A new pattern we're starting to use more broadly for protos is ``apply``. I suggest refactoring this whole method to use that to avoid needing to store the separately. Here's an example of what that would look like for ``Misconception``:

```kotlin
Misconception.newBuilder.apply {
  skillId = <skill_id>
  misconceptionId = <misconception_id>
}.build()
```

Note that for repeated values you'll need to still use 'addAll', but otherwise this helps make things cleaner.

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -162,7 +163,7 @@ class StateRetriever @Inject constructor(
     answerGroupJson: JSONObject,
     interactionId: String
   ): AnswerGroup {
-    return AnswerGroup.newBuilder()
+    var answerGroup = AnswerGroup.newBuilder()
```
BenHenning: A nit: this should be ``val`` since it's not changing references (always prefer ``val`` and only use ``var`` when needed). ``answerGroupBuilder`` would also be clearer and less ambiguous, but I think we can avoid this variable all together--see my comment below.

model/src/main/proto/exploration.proto
```diff
@@ -119,12 +119,18 @@ message CustomSchemaValue {
 // Structure for a single answer group
 // Maps from: data/src/main/java/org/oppia/android/android/data/backends/gae/model/GaeAnswerGroup.kt
 message AnswerGroup {
-  string tagged_skill_misconception_id = 1;
+  Misconception tagged_skill_misconception_id = 1;
```
BenHenning: This isn't really an ID anymore. Maybe ``tagged_skill_misconception`` or just ``tagged_misconception``?

model/src/main/proto/exploration.proto
```diff
@@ -119,12 +119,18 @@ message CustomSchemaValue {
...
   Outcome outcome = 2;
   repeated RuleSpec rule_specs = 3;
   repeated TrainingData training_data = 4;
 }
 
+// Structure for a single misconception
```
```suggestion
// Structure for a single misconception.
```

model/src/main/proto/exploration.proto
```diff
@@ -119,12 +119,18 @@ message CustomSchemaValue {
...
+message Misconception {
+  string skill_id = 1;
+  string misconception_id = 2;
```
BenHenning: These ought to have comments, too. Could we contextualize what they are, and maybe point to a backend code reference to provide additional context?

BenHenning (changes requested):
Thanks @TheRealJessicaLi! Overall I think this looks good, just had some nits. PTAL.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
TheRealJessicaLi: Yes, we're conditionally passing in an argument to this method, so if there isn't a relevant misconception for the wrong answer, the argument defaults to null.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeMisconception.kt
TheRealJessicaLi: Fixed!

TheRealJessicaLi: Good point, I ended up deleting this file though because I didn't realize this was specifying the structure of the JSON we receive from the backend.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeAnswerGroup.kt
TheRealJessicaLi: Fixed, thanks for pointing that out! I misunderstood and thought this was client side as well.

domain/src/main/java/org/oppia/android/domain/classify/AnswerClassificationController.kt
TheRealJessicaLi: Fixed!

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
TheRealJessicaLi: Fixed!

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
TheRealJessicaLi: Updated!

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
TheRealJessicaLi: Updated - I ended up not using a val.

model/src/main/proto/exploration.proto
TheRealJessicaLi: Fixed!

model/src/main/proto/exploration.proto
TheRealJessicaLi: Fixed!

model/src/main/proto/exploration.proto
TheRealJessicaLi: Added documentation, thanks!

.idea/misc.xml
```diff
@@ -6,4 +6,4 @@
   <component name="ProjectType">
     <option name="id" value="Android" />
   </component>
-</project>
+</project>
```
BenHenning: Please  revert--this seems like an unintentional change (also, note that we always have EOF line endings on our files though Android Studio will not add them by default).

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -163,27 +163,22 @@ class StateRetriever @Inject constructor(
...
-            answerGroupJson.getJSONObject("tagged_skill_misconception_id").toString().split("-")[1]
-          )
-      )
-    }
-    return answerGroup.build()
+      val misconceptionJson = answerGroupJson.optJSONObject("tagged_skill_misconception_id")
+      if (misconceptionJson != null) {
+        taggedSkillMisconception =
+          Misconception.newBuilder().apply {
+            skillId = misconceptionJson.toString().split("-")[0]
```
BenHenning: Nit: suggest moving the results of the ``split`` to a var to avoid having to do it twice.

BenHenning (approved):
Thanks @TheRealJessicaLi! Just a couple of comments, otherwise LGTM.

.idea/misc.xml
TheRealJessicaLi: Whoops, my bad - fixed!

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
TheRealJessicaLi: Fixed!

BenHenning:
Looks like the CI pass ran into the JCenter server issues from yesterday. Restarting the whole test suite.

BenHenning (approved):
Thanks @TheRealJessicaLi!

BenHenning (changes requested):
@TheRealJessicaLi it seems there are compiler errors:

```
e: /home/runner/work/oppia-android/oppia-android/domain/src/test/java/org/oppia/android/domain/classify/AnswerClassificationControllerTest.kt: (239, 12): Unresolved reference: setTaggedSkillMisconceptionId
e: /home/runner/work/oppia-android/oppia-android/domain/src/test/java/org/oppia/android/domain/classify/AnswerClassificationControllerTest.kt: (257, 10): Unresolved reference: taggedSkillMisconceptionId
```

Just to check: did you ensure the affected tests were building & running locally, and that the app built? It's always good to double check locally before pushing code since that helps make the process go by faster.

TheRealJessicaLi:
> @TheRealJessicaLi it seems there are compiler errors:
> 
> ```
> e: /home/runner/work/oppia-android/oppia-android/domain/src/test/java/org/oppia/android/domain/classify/AnswerClassificationControllerTest.kt: (239, 12): Unresolved reference: setTaggedSkillMisconceptionId
> e: /home/runner/work/oppia-android/oppia-android/domain/src/test/java/org/oppia/android/domain/classify/AnswerClassificationControllerTest.kt: (257, 10): Unresolved reference: taggedSkillMisconceptionId
> ```
> 
> Just to check: did you ensure the affected tests were building & running locally, and that the app built? It's always good to double check locally before pushing code since that helps make the process go by faster.

Shoot I must have missed those tests; I only checked if the app built. Will fix those, thanks!

TheRealJessicaLi:
@BenHenning Just fixed those tests along with a typo I made. Thanks!

BenHenning:
I think the Gradle failure is another example of #2844.

BenHenning (approved):
Thanks @TheRealJessicaLi! LGTM.

BenHenning:
Weird. GitHub is wrong here: @vinitamurthi shouldn't need to approve this PR since our codeowners overlap here, and I approved it. Strange.

BenHenning:
Hmm not sure what's going on here. I don't think I've seen it behave this way before, but @vinitamurthi I guess you also need to approve this.

vinitamurthi (approved):
Thanks!

vinitamurthi:
Its still showing this even though I reviewed

vinitamurthi:
Oh nevermind, it merged as well!

Issue #2642, opened 2021-02-05
Fix data module tests & add them to CI


BenHenning:
At some point, we deleted the TestMessage proto that was used specifically in tests without updating PersistentCacheStoreTest. That means that test is currently failing, and needs to be fixed. This also illustrates the need to add our data module tests to our CI workflow. This issue is tracking:
1. Running all [data](https://github.com/oppia/oppia-android/tree/develop/data) module tests locally & noting which ones are failing
2. Fix all of the failing tests (e.g. [PersistentCacheStoreTest](https://github.com/oppia/oppia-android/blob/develop/data/src/test/java/org/oppia/android/data/persistence/PersistentCacheStoreTest.kt) & any others that are failing)
3. Update our [CI workflow](https://github.com/oppia/oppia-android/blob/develop/.github/workflows/main.yml#L48) to also run the data module tests

rawlin:
Hi @BenHenning, could you assign me this issue

Akshat-gour:
@BenHenning 
@anandwana001 
can you assign me this issue

BenHenning:
@rawlin are you currently working on this?

BenHenning:
Reopening due to #2871. @prayutsu when creating a new PR that reintroduces the tests, please make sure to verify that the tests aren't flaky & introduce proof (e.g. a screenshot of running them ~100 times without a failure). You will need to dig into what the underlying cause is as the flake is a blocker for introducing data module tests into CI.

BenHenning:
Also removing the 'good second issue' label since this has become harder now.

FareesHussain:
Reopening as this was accidentally closed with https://github.com/oppia/oppia-android/pull/2871.

prayutsu:
> Reopening due to #2871. @prayutsu when creating a new PR that reintroduces the tests, please make sure to verify that the tests aren't flaky & introduce proof (e.g. a screenshot of running them ~100 times without a failure). You will need to dig into what the underlying cause is as the flake is a blocker for introducing data module tests into CI.

Actually, I ran all the tests 3 times and then took the screenshots, so I assumed they were not flaky but will look into this today and make a fresh PR to avoid the flakiness.

PR #2907, opened 2021-03-14
[RunAllTests] Fix #2900: Add network interceptor to request headers

### LINKED ISSUE #2900: Implement network interceptor to add to outgoing request header
Fix part of #53
Implement a network interceptor to catch outgoing requests and add a header with the API key, package version, and package name, and version code for receiving client to authenticate the request 

[Design doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

jcqli:
## Explanation
Fix part of #53
Fix #2900
Implements a new network interceptor that adds request headers to outgoing requests as part of feedback reporting project, although will be applied to all outgoing messages. Renames original `NetworkInterceptor` => `JsonPrefixNetworkInterceptor` to create separate interceptors with independent roles.

[Design doc](https://docs.google.com/document/d/1_RTzad6u4m-ZFn0VTWNddp4bc0fCYrMk-ABIVYVJ7gU/edit?usp=sharing)

## Checklist


anandwana001:
@jcqli Setting the script will help you prevent kotlin lint failures. - https://github.com/oppia/oppia-android/wiki#installation

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        .setPackageName("test_package_name")
+        .build()
+    val packageInfo =
+      PackageInfoBuilder.newBuilder()
+        .setPackageName("test_package_name")
+        .setApplicationInfo(applicationInfo)
+        .build()
+    packageInfo.versionName = "1.0"
+    packageInfo.versionCode = 1
+    packageManager.installPackage(packageInfo)
```
jcqli: This doesn't seem to set the versionName or versionCode -- it throws a NullPointerException later on in the test for the versionName when the RemoteAuthNetworkInterceptor tries to set the headers ([line 44)](https://github.com/oppia/oppia-android/pull/2907/files#diff-c443970e72f2a145c0321816f33ec349b2cdf2054e8bc3b2f0338cb742ef2cf2R44). Any tips?

BenHenning: I think NPE can happen in Context.getVersionName() if there's no PackageInfo for the current package name. You should be able to debug in this test what the package name is for Context and whether it's matching the PackageInfo you're creating (which is using package name 'test_package_name').

utility/src/main/java/org/oppia/android/util/extensions/BUILD.bazel
```diff
@@ -16,7 +16,7 @@ kt_android_library(
 )
 
 kt_android_library(
-    name = "context_extensinos",
+    name = "context_extensions",
```
BenHenning, line 19: Nice catch.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -17,7 +17,8 @@ import javax.inject.Singleton
  * Sample resource: https://github.com/gahfy/Feed-Me/tree/unitTests
  */
 @Module
-class NetworkModule {
+class
+NetworkModule {
```
BenHenning: Should be on the same line.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -31,7 +32,8 @@ class NetworkModule {
   @Singleton
   fun provideRetrofitInstance(): Retrofit {
     val client = OkHttpClient.Builder()
-    client.addInterceptor(NetworkInterceptor())
+    client.addInterceptor(JsonPrefixNetworkInterceptor())
+      .addInterceptor(RemoteAuthNetworkInterceptor())
```
BenHenning: These should be parameter injected, not constructed, otherwise their injections won't be ready.

You can do that by adding these as parameters to the provider method.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -68,4 +70,11 @@ class NetworkModule {
   fun provideFeedbackReportingService(@OppiaRetrofit retrofit: Retrofit): FeedbackReportingService {
     return retrofit.create(FeedbackReportingService::class.java)
   }
+
+  // Provides the API key to use in authenticating remote messages sent or received. This will be
+  // replaced with a stronger, secret key will in production.
+  @Provides
+  @Singleton
```
BenHenning: You should remove this--immutable strings don't need to be scoped since they can be trivially re-retrieved.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -68,4 +70,11 @@ class NetworkModule {
   fun provideFeedbackReportingService(@OppiaRetrofit retrofit: Retrofit): FeedbackReportingService {
     return retrofit.create(FeedbackReportingService::class.java)
   }
+
+  // Provides the API key to use in authenticating remote messages sent or received. This will be
+  // replaced with a stronger, secret key will in production.
```
BenHenning:
```suggestion
  // replaced with a secret key in production.
```

data/src/main/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptor.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Interceptor on top of Retrofit to modify outgoing requests for authenticating messages.
+ *
+ * The Interceptor adds header parameters to outgoing messages.
+ */
+@Singleton
```
BenHenning: This class is stateless so it shouldn't be scoped.

data/src/main/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptor.kt
```diff
@@ -0,0 +1,48 @@
...
+    return chain.proceed(request)
+  }
+
+  /**
+   * This function accepts the outgoing request and adds header values to it before sending.
+   *
+   * @param request the request being sent over the network
+   * @return the exact same request as the parameter with added headers
+   */
+  fun addAuthHeaders(request: Request): Request {
```
BenHenning: Can this be private? It seems to only be needed for the class implementation.

data/src/test/java/org/oppia/android/data/backends/gae/api/ClassroomServiceTest.kt
```diff
@@ -27,7 +27,7 @@ class ClassroomServiceTest {
   @Before
   fun setUp() {
     val client = OkHttpClient.Builder()
-    client.addInterceptor(NetworkInterceptor())
+    client.addInterceptor(JsonPrefixNetworkInterceptor())
```
BenHenning, line 30: Please also inject this everywhere it's used in tests, too.

Though, it might be helpful to wait for #2642 if we can since I think that's largely simplifying data test setup.

utility/src/main/java/org/oppia/android/util/extensions/ContextExtensions.kt
```diff
@@ -9,6 +9,11 @@ fun Context.getVersionName(): String {
   return this.packageManager.getPackageInfo(this.packageName, /* flags=*/ 0).versionName
 }
 
+/** Returns the current app version code. */
+fun Context.getVersionCode(): Int {
+  return this.packageManager.getPackageInfo(this.packageName, /* flags=*/ 0).versionCode
```
BenHenning:
```suggestion
  return this.packageManager.getPackageInfo(this.packageName, /* flags= */ 0).versionCode
```

Maybe fix the one above, too, for an opportunistic fix.

BenHenning (changes requested):
Thanks @jcqli! Left some thoughts. Not sure if you actually want to block on #2916 (might be better not to give the time sensitivity here), but wanted to mention it as an option.

data/src/main/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptor.kt
jcqli: Is there a way to declare this as test-accessible? The network interceptor tests use this to verify the correct headers are added.

jcqli: I made this an internal fun so that the associated test could call on it and verify that headers are added correctly. I think another way to test this and keep the function private would be calling `intercept()` on a chain and verifying the response's associated request matched the interceptor's changes, but this is a bit more indirect (and probably more difficult) and I'm not sure if it's an accurate representation of the request anymore after it goes through the request. Using an internal function seems to be the most beneficial, WDYT?

data/src/main/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptor.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done, I saw this as well but it didn't show up on my AS until now. How odd

utility/src/main/java/org/oppia/android/util/extensions/ContextExtensions.kt
jcqli: Done

data/src/test/java/org/oppia/android/data/backends/gae/api/ClassroomServiceTest.kt
jcqli: Is this interceptor needed for the mock retrofit? All the service tests pass when removing it, maybe that's an indicator that our mock retrofit is not set up properly.

I ask since injecting the interceptors means we then need to declare TestApplications for all the tests that use the interceptors, is that right?

jcqli:
> Thanks @jcqli! Left some thoughts. Not sure if you actually want to block on #2916 (might be better not to give the time sensitivity here), but wanted to mention it as an option.

I think I'd rather have this merged and not blocked on #2916 but would need to make some changes so that its' not blocked by #2878, let me know if that would be better (I think it would make more sense, but requires some more work on my end).

Unrelated, I'm not sure why the linter is giving errors on the third_party/versions.bzl file (from 2878 changes). I think it's about how I added the MockWebServer to dependency but this can be resolved in the other PR, and also depends on if we block on 2878. Let me know your thoughts.

data/src/test/java/org/oppia/android/data/backends/gae/api/ClassroomServiceTest.kt
jcqli: Based on our discussion offline I'm going to leave the interceptor here so that it doesn't interfere with #2642 since it seems that the interceptors will be added there.

third_party/versions.bzl
```diff
@@ -85,6 +85,7 @@ MAVEN_TEST_DEPENDENCY_VERSIONS = {
     "androidx.work:work-testing": "2.4.0",
     "com.github.bumptech.glide:mocks": "4.11.0",
     "com.google.truth:truth": "0.43",
+    "com.squareup.okhttp3:mockwebserver":"4.1.0",
     "com.squareup.retrofit2:retrofit-mock": "2.5.0",
```
prayutsu, line 89: I think lint is failing because there's a space missing before the version (after the colon).

jcqli:
To properly test the network interceptor without exposing private functions (e.g. actually testing the full network request was successful), we used MockWebServer to verify the request call. Turns out, MockWebServer is incompatible with MockRetrofit so the test here does not use the mock services but rather the full service on the mocked server.

/cc @prayutsu Not sure if this will affect your data module PR but might be relevant

data/src/test/java/org/oppia/android/data/backends/gae/JsonPrefixNetworkInterceptorTest.kt
```diff
@@ -109,7 +109,7 @@ class JsonPrefixNetworkInterceptorTest {
       fun build(): TestApplicationComponent
     }
 
-    fun inject(networkInterceptorTest: JsonPrefixNetworkInterceptorTest)
-    fun inject(networkInterceptor: JsonPrefixNetworkInterceptor)
+    fun inject(jsonPrefixNetworkInterceptorTest: JsonPrefixNetworkInterceptorTest)
+    fun inject(jsonPrefixNetworkInterceptor: JsonPrefixNetworkInterceptor)
```
BenHenning: This should be removed--only the test needs an injector generated.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+
+  private val topicName = "Topic1"
+
+  private val mockWebServer = MockWebServer()
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+    setUpApplicationForContext()
+    setUpRetrofit()
```
BenHenning, line 73: I suggest moving your test helpers to also be defined in the same order they're called here (& after the tests since we generally follow an "appendix" style organization for helpers rather than "footnote" style).

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+    mockWebServer.enqueue(MockResponse().setBody("{}"))
+    val call = topicService.getTopicByName(topicName)
+    val serviceRequest = call.request()
+    assertThat(serviceRequest.header("api_key")).isNull()
+    assertThat(serviceRequest.header("app_package_name")).isNull()
+    assertThat(serviceRequest.header("app_version_name")).isNull()
+    assertThat(serviceRequest.header("app_version_code")).isNull()
+
+    call.execute()
+    val interceptedRequest = mockWebServer.takeRequest(timeout = 8, unit = TimeUnit.SECONDS)
```
BenHenning: Suggest going with 10 seconds for parity with TestCoroutineDispatcher (in fact you could use the fields from there for the time & unit if you'd prefer).

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+    verifyRequestHeaders(interceptedRequest!!.headers)
```
BenHenning:
```suggestion
    verifyRequestHeaders(interceptedRequest?.headers)
```

Maybe use nullable calls, instead. The assertion should still fail if the individual retrievals are null, though also recommending a non-null assertion for the request headers in verifyRequestHeaders to make failures a bit more explicit.

'!!' isn't quite as ideal since it will throw a NPE rather than AssertionError. It'll still be clear once someone digs into the failure, but test frameworks generally expect AssertionErrors and may provide slightly better error messaging in those cases. Plus, it's just nice to never use '!!' :)

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+
+  private fun verifyRequestHeaders(headers: Headers) {
+    assertThat(headers.get("api_key")).isEqualTo("test_api_key")
+    assertThat(headers.get("app_package_name")).isEqualTo(context.packageName)
+    assertThat(headers.get("app_version_name")).isEqualTo("1.0")
+    assertThat(headers.get("app_version_code")).isEqualTo("1")
+  }
+
+  private fun setUpRetrofit() {
+    mockWebServer.dispatcher
```
BenHenning: This isn't doing anything--I suggest removing it.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+
+  lateinit var topicService: TopicService
+
+  private val testVersionName = "1.0"
+
+  private val testVersionCode = 1
+
+  private val topicName = "Topic1"
+
+  private val mockWebServer = MockWebServer()
```
BenHenning: Should this be initialized in setUpRetrofit, instead?

Also, does it need to be torn down in an ``@After`` function?

BenHenning: Should this be initialized in setUpRetrofit, instead?

Also, does it need to be torn down in an ``@After`` function?

data/src/main/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptor.kt
BenHenning: It looks like this is resolved now.

data/src/test/java/org/oppia/android/data/backends/gae/api/ClassroomServiceTest.kt
BenHenning: Moved this comment to #2916 via https://github.com/oppia/oppia-android/pull/2916/files#r610890818.

third_party/versions.bzl
BenHenning: Nice catch.

BenHenning (changes requested):
Thanks @jcqli. Not sure why my previous review was submitted automatically, but either way--just have a few comments to address but otherwise I think the PR looks good.

BenHenning:
@jcqli you may also want to assign the other code owners so that they can take a look.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -0,0 +1,200 @@
...
+    mockWebServer.enqueue(MockResponse().setBody("{}"))
+    val call = topicService.getTopicByName(topicName)
+
+    val serviceRequest = topicService.getTopicByName(topicName).request()
+    val request = serviceRequest.newBuilder()
+      .addHeader("api_key", "wrong_api_key")
+      .addHeader("app_package_name", "wrong_package_name")
+      .addHeader("app_version_name", "wrong_version_name")
+      .addHeader("app_version_code", "wrong_version_code")
+      .build()
```
jcqli: @BenHenning I realized that this test doesn't actually use the request with the incorrectly set headers. I don't believe MockWebServer enables us to enqueue a specific request, nor do I know how to pass a request directly into the service (unless we want to mock the service differently). Do you have any suggestions on how we could test this case?

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Ack, done

data/src/test/java/org/oppia/android/data/backends/gae/JsonPrefixNetworkInterceptorTest.kt
jcqli: Done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Ack, done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Ah oops thanks! Done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Yup, done

jcqli: Done

jcqli:
> @jcqli you may also want to assign the other code owners so that they can take a look.

Ack, done!

data/src/main/java/org/oppia/android/data/backends/gae/NetworkApiKey.kt
```diff
@@ -0,0 +1,7 @@
+package org.oppia.android.data.backends.gae
+
+import javax.inject.Qualifier
+
+/** Qualifier for the API key to use when validating incoming and outgoing requests */
```
rt4914: Nit: `.` at the end.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -29,9 +29,13 @@ class NetworkModule {
...
   @Provides
   @Singleton
-  fun provideRetrofitInstance(): Retrofit {
+  fun provideRetrofitInstance(
+    jsonPrefixNetworkInterceptor: JsonPrefixNetworkInterceptor,
+    remoteAuthNetworkInterceptor: RemoteAuthNetworkInterceptor
+  ): Retrofit {
     val client = OkHttpClient.Builder()
-    client.addInterceptor(NetworkInterceptor())
+    client.addInterceptor(jsonPrefixNetworkInterceptor)
```
rt4914: Suggest changing this code to either one of these
```
    val clientBuilder = OkHttpClient.Builder()
      .addInterceptor(jsonPrefixNetworkInterceptor)
      .addInterceptor(remoteAuthNetworkInterceptor)
```

or 

```
    val client = OkHttpClient.Builder()
      .addInterceptor(jsonPrefixNetworkInterceptor)
      .addInterceptor(remoteAuthNetworkInterceptor)
      .build()
```

Because currently your object `client` is actually a `clientBuilder`

data/src/test/java/org/oppia/android/data/backends/gae/api/FeedbackReportingServiceTest.kt
```diff
@@ -29,7 +28,6 @@ class FeedbackReportingServiceTest {
   @Before
   fun setUp() {
     val client = OkHttpClient.Builder()
-    client.addInterceptor(NetworkInterceptor())
```
rt4914, line 32: Does `FeedbackReportingService` do not need `JsonPrefixNetworkInterceptor ` ?

rt4914 (approved):
LGTM, there are three comments. Two you can resolve and other one you can answer and I will check it. Thanks.

anandwana001 (approved):
test files LGTM

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -92,51 +89,82 @@ class RemoteAuthNetworkInterceptorTest {
...
     call.execute()
-    val interceptedRequest = mockWebServer.takeRequest(timeout = 8, unit = TimeUnit.SECONDS)
-    verifyRequestHeaders(interceptedRequest!!.headers)
+    val interceptedRequest = mockWebServer.takeRequest(
+      timeout = testCoroutineDispatcher.DEFAULT_TIMEOUT_SECONDS,
+      unit = testCoroutineDispatcher.DEFAULT_TIMEOUT_UNIT
+    )
+    interceptedRequest?.let {
+      verifyRequestHeaders(it.headers)
+    }
```
BenHenning: I was thinking more like: verifyRequestedHeaders(it?.headers). The null check should work automatically inside verify since you'll also need to do the safe-call operator & Truth will fail if the resulting expression evaluates to null.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
```diff
@@ -92,51 +89,82 @@ class RemoteAuthNetworkInterceptorTest {
...
   @Test
-  fun testNetworkInterceptor_withHeaders_setsCorrectHeaders() {
-    mockWebServer.enqueue(MockResponse().setBody("{}"))
-    val call = topicService.getTopicByName(topicName)
-
-    val serviceRequest = topicService.getTopicByName(topicName).request()
-    val request = serviceRequest.newBuilder()
+  fun testNetworkInterceptor_withIncorrectHeaders_setsCorrectHeaders() {
+    val request = Request.Builder()
+      .url(mockWebServer.url("/"))
```
BenHenning, line 103: Maybe explain briefly via comment why we can't use Retrofit in this test case.

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
BenHenning: Per offline discussion I think we landed on using Okhttp directly (& per the latest code)--can this be resolved now?

BenHenning (commented):
Thanks @jcqli! Generally LGTM, but can you re-run the Bazel tests? I think you're running into one of the known CI issues where compute_affected_tests hangs. You'll need to add '[RunAllTests]' to the PR title & then re-run the Bazel tests workflow.

data/src/test/java/org/oppia/android/data/backends/gae/api/FeedbackReportingServiceTest.kt
jcqli: The service currently doesn't receive anything in the response, so I don't believe so since the interceptor only handles responses.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
jcqli: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkApiKey.kt
jcqli: Done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Ah yup, thanks for following-up on this

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Done

data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt
jcqli: Done

BenHenning (commented):
Thanks @jcqli! Looks like you need to update TestDispatcherModule & RobolectricModule imports in your new tests--the builds for those tests are failing on both Bazel & Gradle. In general, it's good practice to re-run key tests locally (such as new/changed suites) after a merge just to double-check integrations post-sync.

PR LGTM otherwise.

data/src/test/java/org/oppia/android/data/backends/gae/api/FeedbackReportingServiceTest.kt
rt4914: Ack. Thanks.

jcqli:
> Thanks @jcqli! Looks like you need to update TestDispatcherModule & RobolectricModule imports in your new tests--the builds for those tests are failing on both Bazel & Gradle. In general, it's good practice to re-run key tests locally (such as new/changed suites) after a merge just to double-check integrations post-sync.

@BenHenning Thanks for taking a look!

BenHenning (changes requested):
Thanks @jcqli. ``RemoteAuthNetworkInterceptorTest`` appears broken on Bazel. Please fix & send back. (I highly encourage you verify that test builds & passes locally before sending back since the CI run will take ~2 hours to run).

jcqli:
> Thanks @jcqli. `RemoteAuthNetworkInterceptorTest` appears broken on Bazel. Please fix & send back. (I highly encourage you verify that test builds & passes locally before sending back since the CI run will take ~2 hours to run).

Aw man, I missed the data module -- thanks for the pointer, it should be good now!

BenHenning (approved):
Looks like everything is passing now--LGTM. Thanks @jcqli!

Issue #2032, opened 2020-10-22
LogUploadWorkerTests sometimes fail on github actions.


rt4914:
Issue Reference: https://github.com/oppia/oppia-android/pull/1983#issuecomment-712896420

Possible solution reference: https://developer.android.com/topic/libraries/architecture/workmanager/advanced/coroutineworker

<img width="1253" alt="Screenshot 2020-10-20 at 7 58 26 PM" src="https://user-images.githubusercontent.com/9396084/96904445-c2cfd000-14b4-11eb-98b0-314fae5813e0.png">

BenHenning:
New bug report:

**Describe the bug**
LogUploadWorkerTest is sometimes flaky when run with Bazel on CI. Per #2642, we observed this failure:

```
==================== Test output for //domain:src/test/java/org/oppia/android/domain/oppialogger/loguploader/LogUploadWorkerTest:
JUnit4 Test Runner
.Called loadFromPath(/system/framework/framework-res.apk, true); mode=binary sdk=28
.System.logW: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.java.lang.Throwable: Explicit termination method 'close' not called
E
Time: 6.538
There was 1 failure:
1) testWorker_logException_withoutNetwork_enqueueRequest_verifySuccess(org.oppia.android.domain.oppialogger.loguploader.LogUploadWorkerTest)
expected: SUCCEEDED
but was : RUNNING
	at org.oppia.android.domain.oppialogger.loguploader.LogUploadWorkerTest.testWorker_logException_withoutNetwork_enqueueRequest_verifySuccess(LogUploadWorkerTest.kt:167)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

FAILURES!!!
Tests run: 2,  Failures: 1


BazelTestRunner exiting with a return value of 1
JVM shutdown hooks (if any) will run now.
The JVM will exit once they complete.
```

**To Reproduce**
Steps to reproduce the behavior:
1. Check out the repository
[...]

Issue #1675, opened 2020-08-19
Add NetworkModule to ApplicationModule and tests once migrated off of Moshi


miaboloix:
Bazel doesn't support Moshi and, at the moment, Oppia Android's data module uses Moshi but is not functional. Files that were depending on NetworkModule were failing to build with Bazel because of this. So, NetworkModule has been removed from any files previously depending on it. Once data module is migrated off of Moshi and becomes fully functional with Bazel - these references will need to be re-populated.

BenHenning:
Taking this. I was able to get Moshi to work in a separate Kotlin project that's going to download backend JSON structures & encode them in proto for the app. We either will outright remove Moshi, or update it to work based on that project (haven't decided yet).

BenHenning:
Since this is blocking some of @jcqli's work, going to go ahead and patch the repository to properly hook into Moshi. Going to think about how to actually verify this is working, too (e.g. by introducing additional testing).

PR #2916, opened 2021-03-15
Fix #2642: Fix all Data module Tests and add them to run on Github Actions CI check

### LINKED ISSUE #2642: Fix data module tests & add them to CI

prayutsu:
## Explanation

  
Fixes #2642:
All the flaky data module tests are fixed and added to run on CI.

## Checklist

- [x]  The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

<details>

<summary> Screenshots of all passing tests (Each test file is run 100 times) </summary>

**QuestionPlayerServiceTest**
Screenshot from 2021-04-01 10-19-44

**FeedbackReportingServiceTest**
Screenshot from 2021-04-01 10-15-49

**SubtopicServiceTest**
Screenshot from 2021-04-01 10-32-42

**TopicServiceTest**
Screenshot from 2021-04-01 10-28-37

**StoryServiceTest**
Screenshot from 2021-04-01 10-24-39

**ExplorationServiceTest**
Screenshot from 2021-04-01 10-11-46

**ConecptCardServiceTest**
Screenshot from 2021-04-01 10-07-55

**ClassroomServiceTest**
Screenshot from 2021-04-01 10-00-07

**NetworkInterceptorTest**
Screenshot from 2021-04-01 10-41-08

**PersisitentCacheStoreTest**
Screenshot from 2021-04-01 10-42-50

**JsonNetwrokInterceptortest**
Screenshot from 2021-04-27 13-37-50

**RemoteAuthNetwrokInterceptortest**
Screenshot from 2021-04-27 13-38-30

</details>

FeedbackReportingServiceTest was always failing.
Screenshot from 2021-03-15 16-34-55

Fixed FeedbackReportingServiceTest - 
Screenshot from 2021-04-01 10-15-49


prayutsu:
@BenHenning @jcqli PTAL. All test cases of the data module are now fixed except for the `MockFeedbackReportingTest` which has been recently added in #2838, it is always failing regardless of setting failurePercent of the NetworkBehavior to 0 (please also refer to another screenshot in the PR description).

Screenshot from 2021-03-15 16-40-55


@BenHenning I have reformatted `main.yml`, but not sure if it is required or not.

prayutsu:
Screenshot from 2021-03-15 17-37-41
I added the missing JSON attributes and corrected some names to fix this test. However, I am not sure if the corresponding values added are best for this test. Please add your suggestions if you have any.

data/src/test/assets/api_mocks/feedback_reporting.json
```diff
@@ -2,13 +2,25 @@
   "api_key": "",
```
jcqli: This can be removed

data/src/test/assets/api_mocks/feedback_reporting.json
```diff
@@ -19,18 +31,17 @@
     "network_type": "wifi",
     "device_lang": "EN"
```
jcqli: This and the `phone_type` above can both be removed too

jcqli (changes requested):
All other fields look good, thanks for fixing this!

data/src/test/assets/api_mocks/feedback_reporting.json
prayutsu: Done.

data/src/test/assets/api_mocks/feedback_reporting.json
prayutsu: Done.

jcqli (approved):
Thanks, LGTM :)

data/src/test/assets/api_mocks/feedback_reporting.json
prayutsu: Resolving on behalf of @jcqli  as the change has been approved.

data/src/test/assets/api_mocks/feedback_reporting.json
prayutsu: Resolving on behalf of @jcqli  as the change has been approved.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingSystemContext.kt
```diff
@@ -11,7 +11,7 @@ import com.squareup.moshi.JsonClass
 data class GaeFeedbackReportingSystemContext(
 
   /** The package version name for the user's instance of the app. */
-  @Json(name = "version_name") val packageVersionName: String,
+  @Json(name = "package_version_name") val packageVersionName: String,
```
BenHenning: @jcqli can you confirm what the correct JSON key should be in this case? At first glance this doesn't seem correct--the key name may not match the property name, but I also see that the feedback reporting below has some inconsistencies.

data/src/test/java/org/oppia/android/data/backends/test/MockClassroomTest.kt
```diff
@@ -30,13 +30,14 @@ class MockClassroomTest {
...
 
-    retrofit = retrofit2.Retrofit.Builder()
+    retrofit = Retrofit.Builder()
       .baseUrl(NetworkSettings.getBaseUrl())
       .addConverterFactory(MoshiConverterFactory.create())
       .client(client.build())
       .build()
 
     val behavior = NetworkBehavior.create()
+    behavior.setFailurePercent(0)
```
BenHenning: What does this do?

BenHenning (commented):
Thanks @prayutsu! I'm not sure I follow how this fixes the flaky tests, so I'm hoping to better understand that.

@jcqli please also take a look. It seems that the newly introduced feedback reporting test might have been failing prior to this PR. Was it passing for you before your PR was merged?

BenHenning:
Ah, just saw the comments above.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingSystemContext.kt
jcqli: Ah this needs to be changed to `platform_version` -- thanks for catching this

prayutsu:
@BenHenning 
This was the previous behavior of all the test cases - 
Screenshot from 2021-03-16 11-11-53

They were failing only 4-5 time out of 100 times and the cause of failure was - 
```java
retrofit2.mock.MockRetrofitIOException: Failure triggered by MockRetrofit's NetworkBehavior
```

And, here is some part of the declaration of the `NetworkBehavior` class - 
```java
/**
 * A simple emulation of the behavior of network calls.
 * <p>
 * This class models three properties of a network:
 * <ul>
 * <li>Delay – the time it takes before a response is received (successful or otherwise).</li>
 * <li>Variance – the amount of fluctuation of the delay to be faster or slower.</li>
 * <li>Failure - the percentage of operations which fail (such as {@link IOException}).</li>
 * </ul>
 * Behavior can be applied to a Retrofit interface with {@link MockRetrofit}. Behavior can also
 * be applied elsewhere using {@link #calculateDelay(TimeUnit)} and {@link #calculateIsFailure()}.
 * <p>
 * By default, instances of this class will use a 2 second delay with 40% variance. Failures
 * will occur 3% of the time. HTTP errors will occur 0% of the time.
 */
public final class NetworkBehavior {
  private static final int DEFAULT_DELAY_MS = 2000; // Network calls will take 2 seconds.
  private static final int DEFAULT_VARIANCE_PERCENT = 40; // Network delay varies by ±40%.
  private static final int DEFAULT_FAILURE_PERCENT = 3; // 3% of network calls will fail.
  private static final int DEFAULT_ERROR_PERCENT = 0; // 0% of network calls will return errors.
  
  
  
  ....// Some more code 
```
By default, `NetworkBehavior` is supposed to fail almost 3 out of 100 times and which is almost the result we are getting in our tests, to ensure that it never fails, I set the `failurePercent` to 0.

Also here is some more reference to the documentation of `setFailurePercent()` method - 
```java
  /** Set the percentage of calls to {@link #calculateIsFailure()} that return {@code true}. */
  public void setFailurePercent(int failurePercent) {
    checkPercentageValidity(failurePercent, "Failure percentage must be between 0 and 100.");
    this.failurePercent = failurePercent;
  }
  
  /**
   * Randomly determine whether this call should result in a network failure in accordance with
   * configured behavior. When true, {@link #failureException()} should be thrown.
   */
  public boolean calculateIsFailure() {
    return random.nextInt(100) < failurePercent;
  }
  
  private static void checkPercentageValidity(int percentage, String message) {
    if (percentage < 0 || percentage > 100) {
      throw new IllegalArgumentException(message);
    }
  }
```
So, first, we check if the failure percent is between 0 and 100 (default is 3). And if it between 3 and 100 we use it in the method `calculateIsFailure()` which is used in the `BehaviorCall` class.
(Have a look at the onRun() method below in the BehaviorCall class where we require NetworkBehavior)

```java

/** @BenHenning Please go below near onRun() method, as we are not interested in this. */

final class BehaviorCall<T> implements Call<T> {
  final NetworkBehavior behavior;
  final ExecutorService backgroundExecutor;
  final Call<T> delegate;

  private volatile Future<?> task;
  volatile boolean canceled;
  @GuardedBy("this")
  private boolean executed;

  BehaviorCall(NetworkBehavior behavior, ExecutorService backgroundExecutor, Call<T> delegate) {
    this.behavior = behavior;
    this.backgroundExecutor = backgroundExecutor;
    this.delegate = delegate;
  }

  @SuppressWarnings("CloneDoesntCallSuperClone") // We are a final type & this saves clearing state.
  @Override public Call<T> clone() {
    return new BehaviorCall<>(behavior, backgroundExecutor, delegate.clone());
  }

  @Override public Request request() {
    return delegate.request();
  }

  @SuppressWarnings("ConstantConditions") // Guarding public API nullability.
  @Override public void enqueue(final Callback<T> callback) {
    if (callback == null) throw new NullPointerException("callback == null");

    synchronized (this) {
      if (executed) throw new IllegalStateException("Already executed");
      executed = true;
    }
    task = backgroundExecutor.submit(new Runnable() {
      boolean delaySleep() {
        long sleepMs = behavior.calculateDelay(MILLISECONDS);
        if (sleepMs > 0) {
          try {
            Thread.sleep(sleepMs);
          } catch (InterruptedException e) {
            callback.onFailure(BehaviorCall.this, new IOException("canceled"));
            return false;
          }
        }
        return true;
      }
      
      
      /** @BenHenning We are interested here. */
      
      @Override public void run() {
        if (canceled) {
          callback.onFailure(BehaviorCall.this, new IOException("canceled"));
        } else if (behavior.calculateIsFailure()) {
          if (delaySleep()) {
            callback.onFailure(BehaviorCall.this, behavior.failureException());
          }
        } else if (behavior.calculateIsError()) {
          if (delaySleep()) {
            //noinspection unchecked An error response has no body.
            callback.onResponse(BehaviorCall.this, (Response<T>) behavior.createErrorResponse());
          }
        } else {
          delegate.enqueue(new Callback<T>() {
            @Override public void onResponse(Call<T> call, Response<T> response) {
              if (delaySleep()) {
                callback.onResponse(call, response);
              }
            }

            @Override public void onFailure(Call<T> call, Throwable t) {
              if (delaySleep()) {
                callback.onFailure(call, t);
              }
            }
          });
        }
      }
    });
  }
```

So, if the `calculateIsFailure()` returns true, we trigger a failure. That's why the tests were only failing 4-5 times.

data/src/test/java/org/oppia/android/data/backends/test/MockClassroomTest.kt
prayutsu: Please refer to this [comment](https://github.com/oppia/oppia-android/pull/2916#issuecomment-799980130) below.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaeFeedbackReportingSystemContext.kt
prayutsu: @jcqli Should we append "name" to "platform_version" or "platform_version" is sufficient? And also what should be the correct corresponding value of this JSON key in the `feedback_reporting.json`?

jcqli: `platform_version` is great with something like "1.0-example_version" as the JSON value is fine.
Thanks for catching this!

prayutsu: Done. Thanks, @jcqli .

data/src/test/assets/api_mocks/feedback_reporting.json
```diff
@@ -30,7 +30,7 @@
   },
   "report_creation_timestamp_ms": 1610519337000,
   "system_context": {
-    "package_version_name": "example.package.name",
+    "platform_version": "1.0-example_version",
```
BenHenning: Suggest using: 0.1-alpha-abcdef1234

That more closely matches the app version format.

BenHenning (commented):
Ah, that's an excellent find @prayutsu! Might it be possible for us to introduce a new test module (e.g. in testing/ package) that creates & provides the mock retrofit object and can be shared among all of these tests? It'd be nice to set this configuration in one place so that we don't actually hit it in the future when we add more services.

@jcqli I think is working on something similar in a related PR, so it might make sense to collaborate to determine which PR should have that module in it.

jcqli:
> Ah, that's an excellent find @prayutsu! Might it be possible for us to introduce a new test module (e.g. in testing/ package) that creates & provides the mock retrofit object and can be shared among all of these tests? It'd be nice to set this configuration in one place so that we don't actually hit it in the future when we add more services.
> 
> @jcqli I think is working on something similar in a related PR, so it might make sense to collaborate to determine which PR should have that module in it.

For clarification, we would also want to move the mock services to the new module, right? So then both the retrofit mock and the services can be shared with each other

data/src/test/assets/api_mocks/feedback_reporting.json
prayutsu: Done.

data/src/test/java/org/oppia/android/data/backends/test/MockSubtopicTest.kt
```diff
@@ -27,19 +23,7 @@ class MockSubtopicTest {
...
-      .baseUrl(NetworkSettings.getBaseUrl())
-      .addConverterFactory(MoshiConverterFactory.create())
-      .client(client.build())
-      .build()
-
-    val behavior = NetworkBehavior.create()
-    mockRetrofit = MockRetrofit.Builder(retrofit)
-      .networkBehavior(behavior)
-      .build()
+    mockRetrofit = MockRetrofitHelper().createMockRetrofit()
```
prayutsu: @BenHenning Initially, I thought of passing retrofit as an argument to the method `createMockRetrofit()` but as it isn't used anywhere else, I chose to move the whole thing.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
```diff
@@ -0,0 +1,37 @@
...
+import javax.inject.Inject
+
+/**
+ * This helper allows us to create customized [MockRetrofit] instances.
+ */
+class MockRetrofitHelper @Inject constructor() {
+
+  /**
+   * This method returns a MockRetrofit object with [NetworkBehavior] that has 0 failurePercent.
+   */
```
prayutsu: Writing good KDocs is probably the toughest thing, please suggest some good KDocs for both.

prayutsu:
Looks like I am missing some bazel dependency that should be added.

data/BUILD.bazel
```diff
@@ -15,6 +15,7 @@ kt_android_library(
     deps = [
         ":dagger",
         "//model",
+        "//testing",
```
BenHenning: This isn't correct--we shouldn't add test-only dependencies to a production target. This will leak test-only components in the production app. Instead, ``//testing`` should only be used for test targets.

The build bazel binary CI target would fail from adding this if not for the cycle failure being mentioned.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.oppia.android.data.backends.gae.NetworkInterceptor
+import org.oppia.android.data.backends.gae.NetworkSettings
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+import retrofit2.mock.MockRetrofit
+import retrofit2.mock.NetworkBehavior
+import javax.inject.Inject
+
+/**
+ * This helper allows us to create customized [MockRetrofit] instances.
```
BenHenning: ``Helper used to create specifically configured [MockRetrofit] instances.`` may be clearer.

Also, this can be a one-line KDoc since it's short enough to fit. Ditto for other KDocs.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+class MockRetrofitHelper @Inject constructor() {
```
BenHenning: We're constructing this elsewhere rather than injecting it. Could we inject it, instead?

That being said, could we maybe make this a test module & make ``MockRetrofit`` injectable? That might help reduce potential user error in the future since it forces the one ``MockRetrofit`` instance in the Dagger graph to be a correctly configured one.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
BenHenning: What about:

```
Returns a [MockRetrofit] object that can be used to test Retrofit in different situations.

Note that the returned object specifically configures Retrofit to have zero network failures to avoid unwanted test flakes. Tests may override this behavior if they wish to test uncertain network conditions.
```

(will need line wrapping)

Is this clearer & what you wanted to say here?

BenHenning (commented):
Thanks @prayutsu! Sorry for the delayed review. I like how much the new approach simplifies things, but I see a few issues. PTAL at my comment.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
prayutsu: Yes, this is more clear and I wanted to say that I am really bad at writing Documentation comments.

data/BUILD.bazel
prayutsu: Removed it.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
prayutsu: Done.

prayutsu:
@BenHenning @anandwana001 PTAL, made the suggested changes.
But tests are failing with Bazel, looks like I need to add something in BUILD.bzl

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitHelper.kt
prayutsu: Removed it as now we have `MockRetrofitModule`.

prayutsu: Removed it as now we have `MockRetrofitModule`.

prayutsu: Removed it as now we have `MockRetrofitModule`.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitModule.kt
```diff
@@ -1,23 +1,26 @@
...
+ * wish to test uncertain network conditions.
  */
-class MockRetrofitHelper @Inject constructor() {
-
-  /**
-   * This method returns a MockRetrofit object with [NetworkBehavior] that has 0 failurePercent.
-   */
-  fun createMockRetrofit(): MockRetrofit {
+@Module
+class MockRetrofitModule {
```
BenHenning: Suggest: ``RetrofitTestModule`` for consistency with other test modules that we have.

BenHenning (changes requested):
Thanks @prayutsu! Just had one comment, otherwise I think this LGTM.

BenHenning:
@prayutsu I suspect that you need to add Retrofit dependencies to the top-level testing library since the new module requires them. Unfortunately, we don't yet have any examples of what these dependencies need to be. I'll have a PR out soon that fixes Moshi & turns on Retrofit building in the app--that should include the needed dependencies.

testing/src/main/java/org/oppia/android/testing/network/MockRetrofitModule.kt
prayutsu: Done.

prayutsu:
> @prayutsu I suspect that you need to add Retrofit dependencies to the top-level testing library since the new module requires them. Unfortunately, we don't yet have any examples of what these dependencies need to be. I'll have a PR out soon that fixes Moshi & turns on Retrofit building in the app--that should include the needed dependencies.

Sounds good to me.

BenHenning:
Thanks @prayutsu. The PR LGTM, but as mentioned earlier it's blocking on #1675.

BenHenning:
@prayutsu it looks like this can now be updated since #3025 is merged.

prayutsu:
@BenHenning PTAL, the tests still seem broken on Bazel, can we now figure out which dependencies need to be added in the testing module that may fix this issue?

testing/BUILD.bazel
```diff
@@ -25,7 +25,11 @@ kt_android_library(
         "//third_party:androidx_test_ext_junit",
         "//third_party:androidx_test_runner",
         "//third_party:com_google_truth_truth",
+        "//third_party:com_squareup_retrofit2_converter-moshi",
+        "//third_party:com_squareup_retrofit2_retrofit",
+        "//third_party:com_squareup_retrofit2_retrofit-mock",
         "//third_party:junit_junit",
+        "//third_party:moshi",
```
prayutsu, line 48: Added these four dependencies here, and tests started passing on Bazel now.

data/src/test/java/org/oppia/android/data/backends/gae/api/TopicServiceTest.kt
```diff
@@ -1,4 +1,4 @@
-package org.oppia.android.data.backends.gae.api
+package org.oppia.android.data.backends.test
```
BenHenning: This package change doesn't seem correct. The package should be the same as the service being tested (similar to other test suites in the codebase). Ditto for the other service tests.

testing/BUILD.bazel
BenHenning: That seems correct since RetrofitTestModule requires all of these. However, could we maybe inject the existing Retrofit coming from NetworkModule & add it to a new MockRetrofit provided by the new module? We aren't actually replacing the existing Retrofit binding, so it seems like that should "just work" & it will reduce code/dependency duplication here & in the test module.

BenHenning (commented):
Thanks @prayutsu! Just had a few comments--PTAL.

BenHenning:
Also, I think you might be missing a test dependency on the testing module for Bazel data tests @prayutsu.

testing/BUILD.bazel
prayutsu: @BenHenning I didn't get this completely, don't we need a Retrofit object to initialize the Mockretrofit object in the RetrofitTestModule? 
Are you referring RetrofitTestModule as the new module here?

data/src/test/java/org/oppia/android/data/backends/gae/api/TopicServiceTest.kt
prayutsu: Done.

testing/BUILD.bazel
BenHenning: I'm thinking that we could replace RetrofitTestModule with:

```kotlin
@Module
class RetrofitTestModule {
  @Provides
  @Singleton
  fun provideMockRetrofit(@OppiaRetrofit retrofit: Retrofit): MockRetrofit {
    val behavior = NetworkBehavior.create()
    behavior.setFailurePercent(0)

    return MockRetrofit.Builder(retrofit).apply {
      networkBehavior(behavior)
    }.build()
  }
}
```

And then we rely on NetworkModule to create & provide Retrofit. This simplifies dependencies and avoids needing to duplicate the Retrofit arrangement in both places (which is even more important if we change it--we want to keep both in sync).

anandwana001 (changes requested):
Thanks @prayutsu 
my code owners `*test.kt` files look good to me. Let me know once the tests on GitHub checks are passing.

BenHenning (commented):
Thanks @prayutsu. It seems the data tests are not properly pulling in RobolectricTestModule, but I'm not actually sure why. From what I can tell from the PR code, everything *seems* correct. This will require some digging locally, I think.

testing/BUILD.bazel
prayutsu: @BenHenning I tried this but this is showing error - 
Screenshot from 2021-04-03 23-28-26

I also changed the test file and it now looks like - 
```kotlin
@LooperMode(LooperMode.Mode.PAUSED)
class ClassroomServiceTest {

  @Inject
  lateinit var retrofit: Retrofit

  @Inject
  lateinit var mockRetrofit: MockRetrofit

  @Before
  fun setUp() {
  }

  @Test
  fun testClassroomService_usingFakeJson_deserializationSuccessful() {
    val delegate = mockRetrofit.create(ClassroomService::class.java)
    val mockClassroomService = MockClassroomService(delegate)

    val classroom = mockClassroomService.getClassroom("Math")
    val classroomResponse = classroom.execute()

    assertThat(classroomResponse.isSuccessful).isTrue()
    assertThat(classroomResponse.body()!!.topicSummaryDicts?.get(0)?.name).isEqualTo("Math")
  }

    DaggerClassroomServiceTest_TestApplicationComponent
      .builder()
      .setApplication(ApplicationProvider.getApplicationContext()).build().inject(this)
  }

  // TODO(#89): Move this to a common test application component.
  @Singleton
  @Component(modules = [NetworkModule::class, RetrofitTestModule::class])
  interface TestApplicationComponent {
    @Component.Builder
    interface Builder {
      @BindsInstance
      fun setApplication(application: Application): Builder

      fun build(): TestApplicationComponent
    }

    fun inject(test: ClassroomServiceTest)
  }
}
```

Is the above code correct?

prayutsu: @BenHenning I tried this but this is showing error - 
Screenshot from 2021-04-03 23-28-26

prayutsu:
I re-requested the review by mistake.

BenHenning:
@prayutsu does this mean a follow-up review isn't needed yet? I believe there is some local investigation work needed on your part, plus the previous unaddressed comment around the Retrofit Dagger binding.

testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt
```diff
@@ -0,0 +1,40 @@
...
+ * network failures to avoid unwanted test flakes. Tests may override this behavior if they
+ * wish to test uncertain network conditions.
+ */
+@Module
+class RetrofitTestModule {
+  @Provides
+  @Singleton
+  fun provideMockRetrofit(): MockRetrofit {
+    val client = OkHttpClient.Builder()
+    client.addInterceptor(NetworkInterceptor())
```
BenHenning: This should be parameter injected instead of re-constructed.

prayutsu:
@BenHenning I made some changes as per your suggestions, but I am getting this error - 
Screenshot from 2021-04-17 01-19-50

BenHenning:
@prayutsu I think you need to qualify your injection since the provider is qualified. For a similar situation, see: https://github.com/oppia/oppia-android/blob/91c88b96946da0b854340dd201c9632e4b4a5716/data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt#L52. I think in this case you need to do something like:

```kotlin
@field[Inject OppiaRetrofit] lateinit var retrofit: Retrofit
```

That being said, do you even need to inject Retrofit in those tests? The injection doesn't seem to be used.

prayutsu:
> @prayutsu I think you need to qualify your injection since the provider is qualified. For a similar situation, see:
> 
> https://github.com/oppia/oppia-android/blob/91c88b96946da0b854340dd201c9632e4b4a5716/data/src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest.kt#L52
> 
> . I think in this case you need to do something like:
> ```kotlin
> @field[Inject OppiaRetrofit] lateinit var retrofit: Retrofit
> ```
> 
> That being said, do you even need to inject Retrofit in those tests? The injection doesn't seem to be used.

@BenHenning 
Initially, I didn't inject Retrofit in the test files and I was getting this error - 
Screenshot from 2021-04-17 12-54-38

So, I thought maybe I need to first inject Retrofit in the test files so that it can be provided to the MockRetrofit (but this is wrong, as we are injecting Retrofit as a parameter in the provideMockRetroft() method).
I have pushed the code that I tried initially, but I am unable to understand this error -- we already have @Provides annotation added to the provideRetrofitInstance() method in the NetworkModule.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -20,7 +20,7 @@ import javax.inject.Singleton
 class NetworkModule {
 
   @Qualifier
-  private annotation class OppiaRetrofit
+  annotation class OppiaRetrofit
```
BenHenning: Prefer moving this to its own file rather than importing the module elsewhere--modules should only be imported when defining components.

BenHenning (approved):
Thanks @prayutsu. The latest changes are more along what I was thinking. I think your missing binding error in TopicServiceTest (& likely others) is coming from the fact that NetworkModule isn't being included in the component list (keep in mind the full transitive list of modules must be included in the application component's modules list). Adding NetworkModule to each test's component list should fulfill the needs of RetrofitTestModule. Does this help?

Also, just adding a reminder to make sure to ensure all previous open comment threads are also addressed since this PR seems close to wrapping up.

testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt
prayutsu: Done.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
prayutsu: Moved to a separate file.

prayutsu:
> Thanks @prayutsu. The latest changes are more along what I was thinking. I think your missing binding error in TopicServiceTest (& likely others) is coming from the fact that NetworkModule isn't being included in the component list (keep in mind the full transitive list of modules must be included in the application component's modules list). Adding NetworkModule to each test's component list should fulfill the needs of RetrofitTestModule. Does this help?
> 
> Also, just adding a reminder to make sure to ensure all previous open comment threads are also addressed since this PR seems close to wrapping up.

Thanks for the reminder, I have now replied to all the previous open comments.

But I am still getting an error (a little different from the previous one though) while trying running the tests- 
Screenshot from 2021-04-20 14-29-05

It says context can not be provided without @Provides annotation, but I am unable to understand it. Can you please help?

Context is being injected in the RemoteAuthNetworkInterceptor which is one of the dependencies to provide the Retrofit instance. I tried making some changes in the RemoteAuthNetworkInterceptor - 
```kotlin
@Inject
lateinit var context: Context
```
 to 

```kotlin
@field:[Inject Context]
lateinit var context: Context
```
But I think this can only be done for the annotation classes that contain @Qualifier annotation.

BenHenning:
@prayutsu you'll need to add a ``Context`` provider to a test module, like this: https://github.com/oppia/oppia-android/blob/f9eb3cdcba174fdd78690af33ffe3c7a8b547a10/utility/src/test/java/org/oppia/android/util/system/OppiaDateTimeFormatterTest.kt#L73.

prayutsu:
Thanks, @BenHenning, I tried what you suggested and it threw another error -

Screenshot from 2021-04-21 08-59-02

Changes that I made -
```kotlin
 // TODO(#89): Move this to a common test application component.
  @Module
  class TestModule {
    @Provides
    @Singleton
    fun provideContext(application: Application): Context {
      return application
    }
  }

  // TODO(#89): Move this to a common test application component.
  @Singleton
  @Component(modules = [TestModule::class, NetworkModule::class, RetrofitTestModule::class])
  interface TestApplicationComponent {
    @Component.Builder
    interface Builder {
      @BindsInstance
      fun setApplication(application: Application): Builder

      fun build(): TestApplicationComponent
    }

    fun inject(test: ClassroomServiceTest)
  }
}
```

data/src/main/java/org/oppia/android/data/backends/gae/OppiaRetrofit.kt
```diff
@@ -0,0 +1,6 @@
+package org.oppia.android.data.backends.gae
+
+import javax.inject.Qualifier
+
+@Qualifier
+annotation class OppiaRetrofit
```
BenHenning, line 7: Please add documentation for this to explain what the Dagger-bound object tied to this qualifier represents, its type, and which component it's in (see other qualifier KDocs for reference).

BenHenning (commented):
Thanks @prayutsu. Left one more comment.

Regarding the error, can you try pushing the TestModule fixes for all the tests? It'll help if I can see the up-to-date Gradle output from CI.

Note that there are also Bazel BUILD file updates needed per the CI failures. PTAL and resolve if you can, or follow up if you're unsure how to proceed and I can provide some guidance.

BenHenning (changes requested):
Not sure why my status is approved--it should be request changes.

data/src/main/java/org/oppia/android/data/backends/gae/OppiaRetrofit.kt
prayutsu: Done.

prayutsu:
> Thanks @prayutsu. Left one more comment.
> 
> Regarding the error, can you try pushing the TestModule fixes for all the tests? It'll help if I can see the up-to-date Gradle output from CI.
> 
> Note that there are also Bazel BUILD file updates needed per the CI failures. PTAL and resolve if you can, or follow up if you're unsure how to proceed and I can provide some guidance.

@BenHenning I have pushed the changes and I think we need to add the new file `OppiaRetroftit.kt` to some BUILD.bazel file but not sure which one (I am aware that where to add new ViewModels, BindingAdapters but not sure about the Annotation classes)

BenHenning:

BenHenning (commented):
Thanks @prayutsu. To fix your build error I believe you'll need to add OppiaRetrofit to https://github.com/oppia/oppia-android/blob/develop/data/src/main/java/org/oppia/android/data/backends/gae/BUILD.bazel#L34 (preferably as its own library) & add it as a dep to all libraries/tests that refer to it (for tests this likely means just adding the new library to that test's TEST_DEPS list).

Feel free to follow up with me over Hangouts if you're unsure.

FYI @fsharpasharp this is the first time I've seen a new file run into a Bazel issue since we're starting to fragment things. We may need to add some documentation to help contributors troubleshoot these situations.

prayutsu:
@BenHenning Bazel issue seems to be fixed now, but I am still getting some Dagger Binding error - 
Screenshot from 2021-04-24 11-13-12
Screenshot from 2021-04-24 11-14-31

BenHenning:
@prayutsu it looks like there's another Bazel issue: the new oppia_retrofit annotation library will need to have [testing visibility](https://github.com/oppia/oppia-android/blob/develop/BUILD.bazel#L28) so that it can be used by test targets.

Regarding the Dagger error, does adding a dependency on dagger-compiler in the main kapt block for the data module work? Specifically, here: https://github.com/oppia/oppia-android/blob/f9eb3cdcba174fdd78690af33ffe3c7a8b547a10/data/build.gradle#L101

prayutsu:
Thanks @BenHenning, finally all the errors are resolved and all the tests are passing now.

prayutsu:
Also updated screenshots in PR description.

prayutsu:
@BenHenning It looks like tests are fixed on Gradle only and they are still failing on Bazel. I have tried making a separate library for RetrofitTestModule.

data/build.gradle
```diff
@@ -98,7 +98,10 @@ dependencies {
   )
   androidTestImplementation('androidx.test:runner:1.2.0',
       'androidx.test.espresso:espresso-core:3.2.0')
-  kapt("com.squareup.moshi:moshi-kotlin-codegen:1.11.0")
+  kapt(
+      "com.squareup.moshi:moshi-kotlin-codegen:1.11.0",
+      "com.google.dagger:dagger-compiler:2.24"
```
BenHenning: Nit: the deps should be in the other order for lexicographical sorting.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
```diff
@@ -41,4 +41,18 @@ kt_android_library(
     ],
 )
 
+kt_android_library(
+    name = "retrofit_module",
+    srcs = [
+        "RetrofitModule.kt",
```
BenHenning: Is this correct? This doesn't seem to be corresponding to a real module unless I'm missing something. Is this supposed to be ``RetrofitTestModule.kt``?

BenHenning (commented):
@prayutsu I think I spotted the Bazel issue--PTAL. Also, are you building/testing locally with Bazel? If not, I highly suggest to since waiting on CI can be really time consuming.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
prayutsu: Thanks Ben for pointing this out. 

I listed the following deps for this library -
```starlark
kt_android_library(
    name = "retrofit_module",
    srcs = [
        "RetrofitTestModule.kt",
    ],
    visibility = ["//:oppia_testing_visibility"],
    deps = [
        ":dagger",
        ":prod_module",
        "//third_party:com_squareup_retrofit2_converter-moshi",
    ],
)
```

But I am getting this error - 
Screenshot from 2021-04-28 11-56-19

Then I tried removing the prod_module dependency - 
```starlark
kt_android_library(
    name = "retrofit_module",
    srcs = [
        "RetrofitTestModule.kt",
    ],
    visibility = ["//:oppia_testing_visibility"],
    deps = [
        ":dagger",
        "//third_party:com_squareup_retrofit2_converter-moshi",
    ],
)
```

And got this error-  
```
ERROR: /home/prayutsu/StudioProjects/oppia-android/testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel:44:19: KotlinCompile //testing/src/main/java/org/oppia/android/testing/network:retrofit_module_kt { kt: 1, java: 0, srcjars: 0 } for k8 failed: (Exit 1): builder failed: error executing command bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder ... (remaining 1 argument(s) skipped)
error: /tmp/working9325588898131114075/_kotlinc/testing_src_main_java_org_oppia_android_testing_network-retrofit_module_kt_jvm/temp/stubs/org/oppia/android/testing/network/RetrofitTestModule.java:18: error: incompatible types: NonExistentClass cannot be converted to Annotation
    @error.NonExistentClass()
          ^
testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:5:26: error: unresolved reference: data
import org.oppia.android.data.backends.gae.OppiaRetrofit
                         ^
testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:7:18: error: unresolved reference: mock
import retrofit2.mock.MockRetrofit
                 ^
testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:8:18: error: unresolved reference: mock
import retrofit2.mock.NetworkBehavior
                 ^
testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:20:28: error: unresolved reference: OppiaRetrofit
  fun provideMockRetrofit(@OppiaRetrofit retrofit: Retrofit): MockRetrofit {
                           ^
testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:20:63: error: unresolved reference: MockRetrofit
  fun provideMockRetrofit(@OppiaRetrofit retrofit: Retrofit): MockRetrofit {
                                                              ^
Compilation failure: compile phase failed:
error: /tmp/working9325588898131114075/_kotlinc/testing_src_main_java_org_oppia_android_testing_network-retrofit_module_kt_jvm/temp/stubs/org/oppia/android/testing/network/RetrofitTestModule.java:18: error: incompatible types: NonExistentClass cannot be converted to Annotation
    @error.NonExistentClass()
          ^
/home/prayutsu/.cache/bazel/_bazel_prayutsu/ddf5f2ba1270603694fd5d626215d411/execroot/__main__/testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:5:26: error: unresolved reference: data
import org.oppia.android.data.backends.gae.OppiaRetrofit
                         ^
/home/prayutsu/.cache/bazel/_bazel_prayutsu/ddf5f2ba1270603694fd5d626215d411/execroot/__main__/testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:7:18: error: unresolved reference: mock
import retrofit2.mock.MockRetrofit
                 ^
/home/prayutsu/.cache/bazel/_bazel_prayutsu/ddf5f2ba1270603694fd5d626215d411/execroot/__main__/testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:8:18: error: unresolved reference: mock
import retrofit2.mock.NetworkBehavior
                 ^
/home/prayutsu/.cache/bazel/_bazel_prayutsu/ddf5f2ba1270603694fd5d626215d411/execroot/__main__/testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:20:28: error: unresolved reference: OppiaRetrofit
  fun provideMockRetrofit(@OppiaRetrofit retrofit: Retrofit): MockRetrofit {
                           ^
/home/prayutsu/.cache/bazel/_bazel_prayutsu/ddf5f2ba1270603694fd5d626215d411/execroot/__main__/testing/src/main/java/org/oppia/android/testing/network/RetrofitTestModule.kt:20:63: error: unresolved reference: MockRetrofit
  fun provideMockRetrofit(@OppiaRetrofit retrofit: Retrofit): MockRetrofit {
                                                              ^
INFO: Elapsed time: 5.146s, Critical Path: 4.82s
INFO: 11 processes: 8 internal, 3 worker.
FAILED: Build did NOT complete successfully
//data:src/test/java/org/oppia/android/data/backends/gae/JsonPrefixNetworkInterceptorTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/RemoteAuthNetworkInterceptorTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/ClassroomServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/ConceptCardServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/ExplorationServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/FeedbackReportingServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/QuestionPlayerServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/StoryServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/SubtopicServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/backends/gae/api/TopicServiceTest NO STATUS
//data:src/test/java/org/oppia/android/data/persistence/PersistentCacheStoreTest NO STATUS

FAILED: Build did NOT complete successfully

```
I am unable to fix this, I also tried changing the visibility of the libraries. Can you please help here?

data/build.gradle
prayutsu: Done.  I have also replaced the double quotes with single quotes to keep consisitency.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
prayutsu: I have fixed all the build errors but the tests are failing locally for me on Bazel - 
```starlark
 Test output for //data:src/test/java/org/oppia/android/data/backends/gae/api/QuestionPlayerServiceTest:
JUnit4 Test Runner
.E
Time: 1.088
There was 1 failure:
1) testQuestionPlayerService_usingFakeJson_deserializationSuccessful(org.oppia.android.data.backends.gae.api.QuestionPlayerServiceTest)
java.lang.NoClassDefFoundError: android/util/TypedValue
        at java.base/java.lang.Class.getDeclaredFields0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredFields(Class.java:3220)
        at java.base/java.lang.Class.getDeclaredField(Class.java:2567)
        at org.robolectric.shadows.ShadowAssetManager$_AssetManager28_$$Reflector0.setSystemApkAssetsSet(Unknown Source)
        at org.robolectric.shadows.ShadowArscAssetManager9.reset(ShadowArscAssetManager9.java:233)
        at org.robolectric.Shadows.reset(Shadows.java:2304)
        at org.robolectric.android.internal.AndroidTestEnvironment.resetState(AndroidTestEnvironment.java:518)
        at org.robolectric.RobolectricTestRunner.lambda$finallyAfterTest$0(RobolectricTestRunner.java:349)
        at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:75)
        at org.robolectric.RobolectricTestRunner.finallyAfterTest(RobolectricTestRunner.java:347)
        at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:272)
        at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
        at java.base/java.lang.Thread.run(Thread.java:832)
Caused by: java.lang.ClassNotFoundException: couldn't load android.util.TypedValue
        at org.robolectric.internal.bytecode.SandboxClassLoader.maybeInstrumentClass(SandboxClassLoader.java:147)
        at org.robolectric.internal.bytecode.SandboxClassLoader.lambda$loadClass$0(SandboxClassLoader.java:115)
        at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:53)
        at org.robolectric.internal.bytecode.SandboxClassLoader.loadClass(SandboxClassLoader.java:115)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        ... 16 more
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 59
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:195)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:176)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:162)
        at org.robolectric.internal.bytecode.ClassNodeProvider.createClassNode(ClassNodeProvider.java:24)
        at org.robolectric.internal.bytecode.ClassNodeProvider.getClassNode(ClassNodeProvider.java:16)
        at org.robolectric.internal.bytecode.InstrumentingClassWriter.typeInfo(InstrumentingClassWriter.java:115)
        at org.robolectric.internal.bytecode.InstrumentingClassWriter.getCommonSuperClass(InstrumentingClassWriter.java:36)
        at org.objectweb.asm.SymbolTable.addMergedType(SymbolTable.java:1200)
        at org.objectweb.asm.Frame.merge(Frame.java:1299)
        at org.objectweb.asm.Frame.merge(Frame.java:1244)
        at org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1610)
        at org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1546)
        at org.objectweb.asm.MethodVisitor.visitMaxs(MethodVisitor.java:764)
        at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:767)
        at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:647)
        at org.objectweb.asm.tree.ClassNode.accept(ClassNode.java:408)
        at org.robolectric.internal.bytecode.ClassInstrumentor.instrumentToBytes(ClassInstrumentor.java:80)
        at org.robolectric.internal.bytecode.SandboxClassLoader.lambda$maybeInstrumentClass$2(SandboxClassLoader.java:139)
        at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:53)
        at org.robolectric.internal.bytecode.SandboxClassLoader.maybeInstrumentClass(SandboxClassLoader.java:138)
        ... 20 more

FAILURES!!!
Tests run: 1,  Failures: 1
```

Looks like there's some problem with Robolectric version - https://github.com/robolectric/robolectric/issues/5082

prayutsu:
@anandwana001 @BenHenning Data module's tests are failing locally for me on Bazel but they are passing in CI.
Can you please verify if they are passing locally for you?

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
```diff
@@ -41,4 +41,19 @@ kt_android_library(
     ],
 )
 
+kt_android_library(
+    name = "retrofit_module",
+    testonly = True,
```
prayutsu, line 46: Should I also change the name of the library to `retrofit_test_module`?

BenHenning: 'test_module' is sufficient, I think (that's what we conventionally do elsewhere).

data/BUILD.bazel
```diff
@@ -15,6 +15,7 @@ TEST_DEPS = [
     "//data/src/main/java/org/oppia/android/data/persistence:cache_store",
     "//model:test_models",
     "//testing",
+    "//testing/src/main/java/org/oppia/android/testing/network:network",
```
BenHenning: Just ``//testing/src/main/java/org/oppia/android/testing/network`` is sufficient since the target name is the same as the package name.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
BenHenning: Ah. Do you have JDK 8 installed locally? Bazel uses an embedded JDK 11 for actions & building, but for running it uses whatever version of JDK you have locally (we generally recommend Java 8, though CI uses 9 I believe). This seems to be a class version incompatibility, so installing JDK 8 & setting it as your default may fix this issue.

If you don't want to set it as your default, you can set JAVA_HOME temporarily before using ``bazel test`` and it should use that version. There might be a way to configure this via bazelrc, as well, but I'm not sure.

BenHenning (commented):
@prayutsu I can't check right now if they are working locally for me, but I suspect they will if they're passing on CI.

The failures are either the known compute affected tests flake tracked in #2691 & the other failure is #2032. No failures seem tied to this PR.

Otherwise, the PR LGTM. Just have 1 nit + 1 existing comment to address before approving.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
prayutsu: Ack.

data/BUILD.bazel
prayutsu: Done.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
prayutsu: I think I earlier had JDK 8 installed but I recently removed it, but I will configure it again to run bazel tests locally whenever required.
Thanks @BenHenning.

prayutsu:
@anandwana001 PTAL, all the checks are passing now.

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
BenHenning: No problem. Just to check though, did that fix the issue @prayutsu?

BenHenning (approved):
Thanks @prayutsu! The PR underwent a bunch of iterations, but I think it landed in a really nice place. LGTM!

Also as an aside: please don't mark any conversations in PRs resolved unless you authored them (it makes it harder for the person creating the conversation to make sure the changes were addressed).

testing/src/main/java/org/oppia/android/testing/network/BUILD.bazel
prayutsu: Yup, it fixed the issue :)

prayutsu:
> Thanks @prayutsu! The PR underwent a bunch of iterations, but I think it landed in a really nice place. LGTM!
> 
> Also as an aside: please don't mark any conversations in PRs resolved unless you authored them (it makes it harder for the person creating the conversation to make sure the changes were addressed).

@BenHenning Sure, I think I only resolved the conversations initiated by me, maybe I would have accidentally resolved some conversation authored by you, will take care of in the future.

anandwana001 (approved):
all test files LGTM

prayutsu:
@BenHenning PR is ready to merge.

BenHenning:
Ah @prayutsu I think this was my mistake. You're right--the comment of mine I saw disappear was because it was a conversation thread you started. It looks like you did follow the process entirely--thanks! :)

Really happy to see this getting merged. Merging it now.

PR #2932, opened 2021-03-17
Fixes #2427: added configChanges flag to AdministratorControlsActivity in AndroidManifest

### LINKED ISSUE #2427: Logout Dialog disappears on orientation change
**Describe the bug**
In `Adminstrator Controls` screen logout dialog on orientation change disappears 

**To Reproduce**
1. Go to 'Administrator Controls'
2. Click on 'Log Out'
3. Rotate to Landscape
4. Dialog disappears

**Expected behavior**
Alert dialog should persist on orientation change. Also add test for the same in respective file.

**Video**

sakshampruthi:
## Explanation
Fixes #2427 
Since the Administrator Controls Activity doesn't need to update resources during orientation change so I have used configChanges flag for the activity in android manifest. So putting that in the configuration file avoids the system to destroy the AdministratorControlsActivity. Instead, it invokes the onConfigurationChanged(Configuration) method

## Checklist

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,7 +19,8 @@
     <meta-data android:name="expiration_date" android:value="2020-09-01" />
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
-      android:theme="@style/OppiaThemeWithoutActionBar" />
+      android:theme="@style/OppiaThemeWithoutActionBar"
+      android:configChanges="orientation|screenSize"/>
     <activity
       android:name=".app.administratorcontrols.appversion.AppVersionActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,7 +20,7 @@
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
       android:theme="@style/OppiaThemeWithoutActionBar"
-      android:configChanges="orientation|screenSize"/>
+      android:configChanges="orientation|screenSize" />
     <activity
       android:name=".app.administratorcontrols.appversion.AppVersionActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,8 +19,7 @@
     <meta-data android:name="expiration_date" android:value="2020-09-01" />
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
-      android:theme="@style/OppiaThemeWithoutActionBar"
-      android:configChanges="orientation|screenSize" />
+      android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity
       android:name=".app.administratorcontrols.appversion.AppVersionActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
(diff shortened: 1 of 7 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -19,7 +19,8 @@
     <meta-data android:name="expiration_date" android:value="2020-09-01" />
     <activity
       android:name=".app.administratorcontrols.AdministratorControlsActivity"
-      android:theme="@style/OppiaThemeWithoutActionBar" />
+      android:theme="@style/OppiaThemeWithoutActionBar"
+      android:configChanges="orientation|screenSize"/>
```
FareesHussain: Nit fix
```suggestion
      android:configChanges="orientation|screenSize" />
```

FareesHussain (changes requested):

BenHenning (changes requested):
Thanks @sakshampruthi! I don't think this is how we want to solve this problem. The change effectively disables configuration changes for AdministratorControlsActivity by indicating to the system that it will handle configuration changes directly. While that does mitigate the issue for orientation changes & screen size changes, it doesn't solve other cases when a configuration change or activity reconstruction will happen (such as low memory).

The correct way to solve this issue is to use Android's saved instance state bundles. One approach may be to store a value in the bundle to indicate that the dialog should be open, and upon the activity being created automatically open the dialog if that value is present in the bundle. See https://developer.android.com/topic/libraries/architecture/saving-states for more information.

sakshampruthi:
@BenHenning That was initially my approach. But the dialog is being called directly from `administrator_controls_account_actions_view.xml` and the dialog is in `AdministratorControlsAccountActionsViewModel.kt`. So my approach was to save the state in activity using the state bundle and then calling the logout method after checking the state bundle. The problem I am facing is how should I connect  `AdministratorControlsAccountActionsViewModel.kt` and  `AdministratorControlsFragment.kt`. `AdministratorControlsAccountActionsViewModel.kt` requires 2 arguments to be passed to create its reference and   `AdministratorControlsActivity` do not have either of those variables.

ArpitShukIa:
@sakshampruthi As I mentioned [here](https://github.com/oppia/oppia-android/issues/2427#issuecomment-800043536), you need to go down the chain.
- `AdministratorControlsActivity` has a reference to `AdministratorControlsActivityPresenter`.
- `AdministratorControlsActivityPresenter` has a function `getAdministratorControlsFragment()` which gives u a reference to `AdministratorControlsFragment`.
- `AdministratorControlsFragment` has a reference to `AdministratorControlsFragmentPresenter`.
- `AdministratorControlsFragmentPresenter` has a reference to `AdministratorControlsViewModel`.
- `AdministratorControlsViewModel` is creating an object of `AdministratorControlsAccountActionsViewModel` [here](https://github.com/oppia/oppia-android/blob/63a754ebfaccc5f65d12a50d9a8041dd30a9b05a/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt#L86).
- `AdministratorControlsAccountActionsViewModel` is responsible for creating the alert dialog.

This is definitely a long chain, but with the current app architecture, this seems like the only possible way to me.

sakshampruthi:
@ArpitShukIa I know how to communicate between activities using startActivitiesdforresult() but I am not sure how to proceed with these classes. Maybe you can help me out with this?

ArpitShukIa:
@sakshampruthi 
- AdministratorControlsAccountActionsViewModel: Create a public boolean `isDialogVisible`. Set this to `true` in `onLogOutClicked()`. Add an onDismissListener to alert dialog and set this boolean to `false` there.
- AdministratorControlsViewModel: Create a global reference to `AdministratorControlsAccountActionsViewModel` using `AdministratorControlsAccountActionsViewModel(fragment, IntentFactoryShim)` and add this to `itemViewModelList`
<br>[Alternative to this can be to create that boolean here and pass an additional lambda to AdministratorControlsAccountActionsViewModel to change this boolean]
<br>Create two functions here, one which returns the value of `isDialogVisible` to the caller and another function which calls `onLogOutClicked()`. Use the global reference for both these functions.

Now continue creating such functions for every member in the chain.
Then finally in `AdministratorControlsActivity's` `onSaveInstanceState` fetch this boolean value from the presenter and save it in the bundle. And in `onCreate()` check for this value, if present call the corresponding function in presenter to display the dialog again.

rt4914:
@sakshampruthi I will need to close this PR because of **force-push**.
We should not do a force push as it tampers the commit history and makes it hard to review and revisit code.

It is also mentioned in this point: https://github.com/oppia/oppia-android/wiki#address-review-comments-until-all-reviewers-give-lgtm-looks-good-to-me

sakshampruthi:
@rt4914 Can I open a new pull request with no force-pushed code?

PR #2933, opened 2021-03-17
Fix #2131: Hide progress bar when keyboard is visible

### LINKED ISSUE #2131: Progress Bar is present when keyboard is active
Only display the Progress Bar when the keyboard is inactive. By removing the progress bar while keyboard is active, the learner can view more content and the "Submit" button is not cropped.

**How to reproduce?**
1. Open the app.
2. Click on `Fractions` topic.
3. Click on `Practice` tab.
4. Select a couple of skills.
5. Click Start.

Note: The questions in `Practice Mode` are randomised and therefore whenever you start the session you might encounter different set of questions. 

**Update**:
We will have to keep the progress bar even when keyboard is visible but we need to optimise on this UI so that it looks a bit more correct (a.) add bottom-padding to submit button (b.) use gradient background for the progress bar so that the button is visible with opacity and does not cut abruptly.

ArpitShukIa:
## Explanation
Fixes #2131
 I have added a `GlobalLayoutListener` which is invoked when the global layout state within the view tree changes. Then I calculated the difference `diff` between the current fragment height and the height of its root's view. This difference when greater than 200dp (approx.) implies that the keyboard is visible.

https://user-images.githubusercontent.com/51705072/111506910-1f879c00-8770-11eb-959c-e287c39ae1d3.mp4

## Checklist


ArpitShukIa:
@rt4914 PTAL.

anandwana001 (changes requested):
Thanks @ArpitShukIa 
Could you please also write test cases where we can test this fix? The test should pass on both espresso and robolectric.
Also, test it on landscape/tablet mode too. 

Reference - https://github.com/oppia/oppia-android/wiki/Oppia-Android-Testing

prayutsu:
@anandwana001 I think we'll also need to verify this on tablets and small-size screen phones.

anandwana001:
@prayutsu Yes, small devices too. Added the same in above comment 
> Also, test it on landscape/tablet mode too.

ArpitShukIa:
@anandwana001 @prayutsu 
I have written a test for this fix.
```
  @Test
  fun testQuestionPlayer_openThenCloseKeyboard_progressBarHidesWhenKeyboardIsVisible() {
    launchForSkillList(TEXT_INPUT_SKILL).use {
      onView(withId(R.id.question_progress_bar)).check(matches(isCompletelyDisplayed()))
      onView(withId(R.id.text_input_interaction_view)).perform(typeText(" "))
      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
      Espresso.closeSoftKeyboard()
      onView(withId(R.id.question_progress_bar)).check(matches(isCompletelyDisplayed()))
    }
  }
```
- Should I break it down into two small tests?
- It passes when I run it as an Instrumented Test but fails when run as an Android JUnit test. Can you please tell what might be the reason for this?
Also I had a doubt regarding Roboelectric vs Expresso. Is it that **Roboelectric ≡ JUnit test** and **Expresso ≡ Instrumentation test**?

For mobile landscape mode this issue is not applicable as keyboard always opens in full screen.

https://user-images.githubusercontent.com/51705072/111581570-7c6e6b00-87df-11eb-8cd5-19cb78cf3122.mp4

For the small screen phones In portrait mode and tablets, I will verify this and send corresponding screen recordings (and make changes, if required).

For small devices, can I use a 5" Pixel phone?

prayutsu:
@ArpitShukIa Yes, you'll need to breakdown the test as you can only `check` one assertion in a single test case.

However, the flow should remain same. For example, in first test case you test whether the progressbar is visible or not, in second you test after opening the keyboard the progresbar hides or not, in third you test after closing the keyboard the progress bar is again visible or not.

ArpitShukIa:
@prayutsu Can u please specify the breakdown?
Can i do it this way?
- Test1 : Open keyboard, assert that progress bar is hidden.
- Test2: Simpley assert that progress bar is shown. (Because keyboard is hidden by default)

ArpitShukIa:
@prayutsu Okay 👍

prayutsu:
For small screen size, you can use emulator 3.3" WQVGA or similar to it.

anandwana001:
@prayutsu @ArpitShukIa Currently, we are not targetting 3.3" WQVGA this much small screens. You can test it up to 5".

ArpitShukIa:
@anandwana001 @prayutsu 
Pixel 5":

https://user-images.githubusercontent.com/51705072/111604375-4808a800-87fb-11eb-91e1-43e9a80cec59.mov

Tablet Portrait:

https://user-images.githubusercontent.com/51705072/111604438-58208780-87fb-11eb-810a-20248aebf7ba.mp4

Tablet Landscape:

https://user-images.githubusercontent.com/51705072/111604525-6bcbee00-87fb-11eb-9315-0185912e9852.mp4

<br><br><br>
Also, plz take a look at the error while running tests: (You can skip the section between 00:50 and 01:55)

https://user-images.githubusercontent.com/51705072/111604978-e563dc00-87fb-11eb-8961-54e90b63d5c8.mov

Can u plz tell me where I went wrong?

prayutsu:
Can you also show your exact test case?

prayutsu:
Hmm, maybe try `withEffectiveVisibility(Visibility.INVISIBLE)` instead of `not(isDisplayed())`.

ArpitShukIa:
> Can you also show your exact test case?

The complete test case can be seen at 00:00 time frame in the video.

> Hmm, maybe try `withEffectiveVisibility(Visibility.INVISIBLE)` instead of `not(isDisplayed())`.

I tried this but even this is not working. At 2:20 in the video, it says `Got: "ProgressBar{...,visibility=VISIBLE,...}`

prayutsu:
@ArpitShukIa Suggest trying to delay by using `testCoroutineDispatchers.runCurrent()` before the assertion as Robolectric maybe be checking the assertion even before it actually gets invisible.

ArpitShukIa:
@prayutsu @anandwana001
Tried `testCoroutineDispatchers.runCurrent()` as well. Roboelectric test is always failing.
Is it possible that there might be something wrong with my device setup?

prayutsu:
@ArpitShukIa If there would have been some error in the setup then other tests would haven't get passed. Try typing a longer string in the test (this is just to guess the behavior of the test case on robolectric).

ArpitShukIa:
@prayutsu Typed a string of 1000 characters, no success.
Should I push the changes? Can u review the code on your device?? I might have made some silly mistake that is causing trouble.

prayutsu:
Okay, please push the changes then.

ArpitShukIa:
@prayutsu PTAL.

prayutsu:
@ArpitShukIa I am getting the same results as well. I tried - 

1. Adding `testCoroutineDispatchers.runCurrent()`
2. Using `withEffectiveVisibility()`
3. Typing some text after the click to give some delay
4. Using `appendText()` instead of `typeText()`

But none of them worked, this will require more investigation.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
     }
   }
 
+  @Test
+  fun textQuestionPlayer_startQuestionPlayer_progressBarIsVisible() {
```
anandwana001:
```kotlin
fun textQuestionPlayer_startQuestionPlayer_progressBarIsDisplayed()
```

In other tests we have `IsCompletelyVisible` because we are using `isCompletelyDisplayed()` function. 
We can update those function too with `isCompletelyDisplayed`. This will keep a consistency of using `isCompletelyDisplayed()` and `isDisplayed()` wherever needed.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
...
+    launchForSkillList(TEXT_INPUT_SKILL).use {
+      onView(withId(R.id.question_progress_bar)).check(matches(isCompletelyDisplayed()))
+    }
+  }
+
+  @Test
+  fun textQuestionPlayer_openKeyboard_progressBarIsGone() {
```
anandwana001:
```kotlin
fun textQuestionPlayer_openKeyboard_progressBarIsNotDisplayed()
```

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
...
+    launchForSkillList(TEXT_INPUT_SKILL).use {
+      onView(withId(R.id.text_input_interaction_view)).perform(click())
+      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
+    }
+  }
+
+  @Test
+  fun textQuestionPlayer_openThenCloseKeyboard_progressBarIsVisible() {
```
anandwana001:
```kotlin
fun textQuestionPlayer_openThenCloseKeyboard_progressBarIsDisplayed()
```

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -10,6 +10,7 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.RecyclerView
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ApplicationProvider
+import androidx.test.espresso.Espresso
```
anandwana001: `import androidx.test.espresso.Espresso.closeSoftKeyboard`

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
...
+      onView(withId(R.id.text_input_interaction_view)).perform(click())
+      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
+    }
+  }
+
+  @Test
+  fun textQuestionPlayer_openThenCloseKeyboard_progressBarIsVisible() {
+    launchForSkillList(TEXT_INPUT_SKILL).use {
+      onView(withId(R.id.text_input_interaction_view)).perform(click())
+      Espresso.closeSoftKeyboard()
```
anandwana001: `closeSoftKeyboard` we can import function, and skip importing the complete Espresso.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
     }
   }
 
+  @Test
+  fun textQuestionPlayer_startQuestionPlayer_progressBarIsVisible() {
+    launchForSkillList(TEXT_INPUT_SKILL).use {
```
anandwana001:
```suggestion
    launchForSkillList(SKILL_ID_LIST).use {
```

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
...
+      onView(withId(R.id.question_progress_bar)).check(matches(isCompletelyDisplayed()))
+    }
+  }
+
+  @Test
+  fun textQuestionPlayer_openKeyboard_progressBarIsGone() {
+    launchForSkillList(TEXT_INPUT_SKILL).use {
```
anandwana001:
```suggestion
    launchForSkillList(SKILL_ID_LIST).use {
```

anandwana001: need 2 more steps to go on input

```kotlin
@Test
  fun textQuestionPlayer_openKeyboard_progressBarIsNotDisplayed() {
    launchForSkillList(SKILL_ID_LIST).use {

      selectMultipleChoiceOption(optionPosition = 2)
      onView(withId(R.id.continue_navigation_button)).perform(click())

      onView(withId(R.id.text_input_interaction_view)).perform(click())
      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
    }
  }
```

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +297,30 @@ class QuestionPlayerActivityTest {
...
+      onView(withId(R.id.text_input_interaction_view)).perform(click())
+      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
+    }
+  }
+
+  @Test
+  fun textQuestionPlayer_openThenCloseKeyboard_progressBarIsVisible() {
+    launchForSkillList(TEXT_INPUT_SKILL).use {
```
anandwana001:
```suggestion
    launchForSkillList(SKILL_ID_LIST).use {
```

anandwana001: Similar from above comment

```kotlin
selectMultipleChoiceOption(optionPosition = 2)
onView(withId(R.id.continue_navigation_button)).perform(click())
```

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -99,6 +102,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 private val SKILL_ID_LIST = listOf(FRACTIONS_SKILL_ID_0)
+private val TEXT_INPUT_SKILL = listOf(TEST_SKILL_ID_0)
```
anandwana001: we can use `SKILL_ID_LIST`, not required `TEXT_INPUT_SKILL`.

anandwana001 (changes requested):
Thanks @ArpitShukIa 
Left nit suggestions. 
Also, please add espresso and robolectric test result screenshot in the PR description. 

@rt4914 Is it good if we don't use `TEXT_INPUT_SKILL` and keep the testing solid on major data files like `FRACTIONS_SKILL_ID_0`?

ArpitShukIa:
@anandwana001 I have implemented all the requested changes in the latest commit. PTAL.

> Also, please add espresso and robolectric test result screenshot in the PR description.

Tests are passing in expresso but one test `textQuestionPlayer_openKeyboard_progressBarIsNotDisplayed` is always failing with Roboelectric. Plz have a look at the last video in [this](https://github.com/oppia/oppia-android/pull/2933#issuecomment-801780747) comment.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -293,6 +295,38 @@ class QuestionPlayerActivityTest {
...
+  }
+
+  @Test
+  fun textQuestionPlayer_openKeyboard_progressBarIsNotDisplayed() {
+    launchForSkillList(SKILL_ID_LIST).use {
+      selectMultipleChoiceOption(optionPosition = 2)
+      onView(withId(R.id.continue_navigation_button)).perform(click())
+
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.text_input_interaction_view)).perform(click())
```
anandwana001, line 343: try putting a syn

```kotlin
  @Test
  fun textQuestionPlayer_openKeyboard_progressBarIsNotDisplayed() {
    launchForSkillList(SKILL_ID_LIST).use {
      selectMultipleChoiceOption(optionPosition = 2)
      onView(withId(R.id.continue_navigation_button)).perform(click())

      onView(withId(R.id.text_input_interaction_view)).perform(click())

      onView(withId(R.id.question_progress_bar)).check(matches(not(isDisplayed())))
    }
  }
```

anandwana001 (changes requested):
Let's see if syncing solved our failing test.

ArpitShukIa:
@anandwana001 Tried it. Still failing :(

anandwana001:
The possible reason for this failure is the time delay. Robolectric is checking visibility before the keyboard gets open. Will look into this today later. TestCoroutineDispatchers should solve this.

rt4914:
@ArpitShukIa I will review this PR at later stage meanwhile you can fix this as per other reviewers comment.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
anandwana001: @FareesHussain Could you try this one this `testCoroutineDispatchers` and using it's different functions? If we can bring delay while the keyboard get's opened up.

FareesHussain: @anandwana001 It looks like keyboard is not supported in `Robolectric`

[Reference](http://robolectric.org/device-configuration/#:~:text=Keyboard%20availability)

FareesHussain: Can we ignore this for Robolectric?

anandwana001: Yes, but we are not interacting with keyboard which is basically the `Keyboard availability` means I think, what we need here is just to wait till the keyboard get's open, so it seems like we need a bit of stable delay. 
Could you check if we can do this with any of the functions from `testCoroutineDispatchers` ?

ArpitShukIa: @anandwana001 I have already played a lot with `testCoroutineDispatchers` and `Thread.sleep()`. Delay doesn't seem to be a solution. What I doubt is that Roboelectric is not even able to open the keyboard. I have found some other similar hardware related issues in the code such as in `testProfileEdit_configChange_startWithUserProfile_clickDelete_checkOpensDeletionDialog` test where Roboelectic is not able to change the configuration to landscape (the test passed even when I removed the `delete profile` button from the landscape layout XML)
There seems to be some problem with Roboelectric itself.

anandwana001:
@ArpitShukIa The conversation should only be resolved by the one who is starting it or the reviewer only.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
anandwana001: We have ongoing PR for fixing orientation change in robolectric, but I guess for now it's good to have the test written, we cannot confirm it will pass but it's good to have and when orientation change thing worked out, GitHub action will catch if anything fails.

ArpitShukIa:
@anandwana001 Okay, does this mean that I should never `unassign me`?

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
FareesHussain: What I think here is when using robolectric it doesn't toggle the android keyboard

FareesHussain: Or the android keyboard is never displayed in Robolectric

anandwana001: Adding this for reference - https://github.com/robolectric/robolectric/pull/4231

I think yes, we can mark it espresso-only test for now. @ArpitShukIa Please add a comment on a test about why we are keeping this test as espresso only. Also, add the link for reference `https://github.com/robolectric/robolectric/pull/4231`

anandwana001 (changes requested):
@ArpitShukIa PTAL https://github.com/oppia/oppia-android/pull/2933#discussion_r603793589

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
ArpitShukIa: @anandwana001 Should I mark `textQuestionPlayer_openThenCloseKeyboard_progressBarIsDisplayed` as well to run only on Expresso. The test passes on Roboelectric just because the keyboard is never opened (and hence never closed).

anandwana001: Yes, this seems like. a false passing. 
You can mark them to run on espresso only. 
`@RunOn(TestPlatform.ESPRESSO)`

ArpitShukIa: Done

ArpitShukIa:
@anandwana001 PTAL.

anandwana001 (approved):
LGTM, Thanks @ArpitShukIa 
Tests looks good.

rt4914 (approved):
LGTM but adding @BenHenning for review. Mainly for his thoughts on `QuestionPlayerFragmentPresenter` file.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
```diff
@@ -400,6 +401,21 @@ class QuestionPlayerFragmentPresenter @Inject constructor(
     )
   }
 
+  private val keyboardVisibilityListener = ViewTreeObserver.OnGlobalLayoutListener {
+    val diff = binding.root.rootView.height - binding.root.height
+    val threshold = 200 * activity.resources.displayMetrics.density // represents 200 dp
```
BenHenning: We should use Android's conversion mechanism for this (i.e. TypedValue).

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
```diff
@@ -400,6 +401,21 @@ class QuestionPlayerFragmentPresenter @Inject constructor(
...
+    val isKeyboardVisible = diff > threshold
+    questionViewModel.isKeyboardVisible.set(isKeyboardVisible)
```
BenHenning: This seems like a really hacky way to check if the keyboard is open. It's making a lot of assumptions on how large the keyboard is (which is a customizable and OEM-specific property). I think the discussion in https://groups.google.com/g/android-platform/c/FyjybyM0wGA is insightful--it's better practice to build a solution that doesn't directly depend on the keyboard itself. I think this is a case where Android is intentionally designed to help avoid developers inadvertently introducing OEM-specific bugs.

In this case, could we change how we display the progress bar such that it wraps under the keyboard, but stays on top of the recycler view when scrolling? That seems to me to be the more correct solution rather than trying to tweak the UI behavior on the keyboard itself.

@rt4914 in case you have any ideas on how to do the above.

BenHenning (changes requested):
Thanks @ArpitShukIa. I'm a bit concerned with this solution--please see my comment.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
ArpitShukIa: @BenHenning @rt4914 
> I think this is a case where Android is intentionally designed to help avoid developers inadvertently introducing OEM-specific bugs.

Actually, Android Team is working on solutions to provide developers with more control over IME options. In fact, Android 11 has a new API `WindowInsets` dedicated to this. The API provides a function to check if the keyboard is visible or not. (https://youtu.be/acC7SR1EXsI?list=PLWz5rJ2EKKc-Z8NeBXJkf1bzUVhx3fvh4&t=299). I couldn't use this because of supported SDK version constraints.

> In this case, could we change how we display the progress bar such that it wraps under the keyboard, but stays on top of the recycler view when scrolling? 

The keyboard doesn't come over the fragment, it pushes the fragment up. So, I doubt if we can do something like hiding/moving the progress bar "behind" the keyboard.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
ArpitShukIa: Okay, will fix this.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
rt4914: @BenHenning I do agree with your point of view. 
@ArpitShukIa What can be the best solution for the issue #2131  without doing all this calculations?

ArpitShukIa: @BenHenning @rt4914
I am not sure what can be the best solution for this.
We don't have a lot of options while dealing with soft-keyboard. This solution has a lot of upvotes in [stackoverflow](https://stackoverflow.com/questions/2150078/how-to-check-visibility-of-software-keyboard-in-android).

One modification we can do is something like 
`val isKeyboardVisible = binding.root.rootView.height > 1.5 * binding.root.height` 
But yeah, here also we will be assuming that if more than 1/3rd of the screen is covered, it might be due to soft keyboard.
I don't think there is any straightforward way to check keyboard's visibility (before API 30)

I also tried using `android:windowSoftInputMode="adjustPan"` for the activity so that the keyboard comes above the activity and doesn't push it up. The progress bar was hiding correctly behind the keyboard but there was a problem that some part of ScrollView was also hiding behind it and couldn't be scrolled to view the bottom contents.

I can't think of any other way to approach the problem. Please suggest if you have any ideas.

anandwana001:
@ArpitShukIa any update on this?

ArpitShukIa:
@anandwana001 I explained my part in the last comment. I couldn't find any other way to solve the issue without checking for the keyboard visibility.
Please tell me if u have some ideas on how the issue can be fixed.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
BenHenning: Sorry to be clear: I'm suggesting we try to solve this problem a different way that does not require knowing whether the keyboard is open. Could we explore changing the UI itself to address the issue first? It may require a part UI redesign depending on what options we have, but that might be fine. I'm worried about the stability of this solution on different devices--it will be quite difficult for us to know if something goes wrong since the user won't necessary understand that it's broken to report it to us. Given that, we should try to find a way to solve this through the UI structure itself first before trying to factor in the size of the keyboard.

ArpitShukIa: I can try modifying the UI (something like positioning the progress bar on top) but I think we should take some inputs from the UI Designer first regarding any UI change.

ArpitShukIa: @mschanteltc Can you suggest some UI changes to fix this issue #2131?

mschanteltc: The progress bar should always be fixed to the bottom, so when the keyboard is active, it should cover the progress bar.

ArpitShukIa: @BenHenning @rt4914 Need some help here. Can you tell exactly what UI changes need to be done?

BenHenning: @mschanteltc we might need to change the UI experience in another way to make this work. We'll follow up if that's the case.

@rt4914 we should discuss options further here. I'm wondering if there's some way to tweak the UI such that the progress bar stays above other elements but doesn't get pushed up by the keyboard.

rt4914: @BenHenning Ack. But I am actually not sure what would be the best solution here. I will try to use this PR to see if I can do something.

BenHenning: From talking to @rt4914 today, we were discussing of two options:
- Experimenting with switching this layout to adjustPan instead of adjustResize. This might help circumvent the issue, but will worsen the user experience.
- Adding [configuration handling support](https://developer.android.com/guide/topics/resources/runtime-changes#HandlingTheChange) (e.g. via ``keyboardHidden``) to this activity for keyboard changes & then look at the Configuration class for [whether the software keyboard is open](https://developer.android.com/reference/android/content/res/Configuration#keyboardHidden) (& condition the visibility of the progress bar on that)

@ArpitShukIa it might be worth trying (2) and see if that's a viable option here.

ArpitShukIa: @BenHenning @rt4914 
(2) doesn't seem to be working. I added `android:configChanges="keyboardHidden"` in the manifest for QuestionPlayerActivity and overrode onConfigurationChanged() in the activity. But that function is not being triggered when the visibility of softKeyboard changes.

I think a soft keyboard doesn't trigger activity restart on its appearance or disappearance. Looks like the flag just tracks the "availability" of soft and hard keyboard and not the "visibility" (in case of soft keyboard). (The name `keyboardHidden` seems to be a little misleading)

image

Did I miss something here?

BenHenning: Hmm no I think you're right. How does the experience look when the activity is switched to adjustPan mode @ArpitShukIa?

We also might be able to use https://developer.android.com/jetpack/androidx/releases/core#1.5.0-alpha02. There's a new API that lets us check the visibility of the onscreen keyboard in hopefully a more robust way than directly checking its size (since per the stackoverflow page is easy to mess up). However, that API will also require us to use adjustPan so we should check that experience first.

If possible, could you update this PR with the existing solution commented out & adjustPan enabled? The CI checks automatically build an APK that we can download locally & check quickly to see the experience.

ArpitShukIa: @BenHenning I have updated the PR to use adjustPan. PTAL.
The behaviour is the same as I described in [this](https://github.com/oppia/oppia-android/pull/2933#discussion_r606179734) comment.

Also, as per the [docs](https://developer.android.com/guide/topics/manifest/activity-element#wsoft) for adjustPan,
> This is generally less desirable than resizing, because the user may need to close the soft keyboard to get at and interact with obscured parts of the window.

https://user-images.githubusercontent.com/51705072/116185662-ae101580-a73f-11eb-99f3-df3eab2c556a.mp4

BenHenning: @rt4914 & @mschanteltc do either of you have thoughts on the most recent video provided by @ArpitShukIa? Either we need to redesign the question player to work better with the keyboard, or we might need to live with adjustPan since it's a limitation of Android.

Also, in case it wasn't clear in the video, in some cases the user will not be able to access the submit button because of the keyboard being on top of it. We *might* be able to make this a bit better by detecting using the library above when the keyboard is open and dynamically add more space to offset it (@rt4914 & @ArpitShukIa any thoughts on this?).

BenHenning: @rt4914 & @mschanteltc do either of you have thoughts on the most recent video provided by @ArpitShukIa? Either we need to redesign the question player to work better with the keyboard, or we might need to live with adjustPan since it's a limitation of Android.

Also, in case it wasn't clear in the video, in some cases the user will not be able to access the submit button because of the keyboard being on top of it. We *might* be able to make this a bit better by detecting using the library above when the keyboard is open and dynamically add more space to offset it (@rt4914 & @ArpitShukIa any thoughts on this?).

ArpitShukIa: > We might be able to make this a bit better by detecting using the library above when the keyboard is open

I tried using the API but am unable to find any function which can be used for API 19 and 20 (as in our case).
The two ways that I tried are:
1. Using `ViewCompat.setOnApplyWindowInsetsListener(View, OnApplyWindowInsetsListener)`
```kotlin
ViewCompat.setOnApplyWindowInsetsListener(binding.root.rootView) { _, insets ->
      val isKeyboardVisible = insets.isVisible(Type.ime())
      val keyboardHeight = insets.getInsets(Type.ime()).bottom
      insets
}
```
But this function works only for API>=21
image
<br>
2. Using `ViewCompat.getRootWindowInsets(View)` inside `ViewTreeObserver.OnGlobalLayoutListener.onGlobalLayout()` callback
```kotlin
val insets = ViewCompat.getRootWindowInsets(binding.root)!!
val isKeyboardVisible = insets.isVisible(Type.ime())
val keyboardHeight = insets.getInsets(Type.ime()).bottom
```
But this function is of no use for API<23
image

As per the android docs,
> Some caveats about the ime() type, it works very reliably on API 23+ when your Activity is using the adjustResize window soft input mode. If you’re instead using the adjustPan mode, it should work reliably back to API 14.

It says that the API **should** work back to API 14 but I couldn't find the right functions which work below 21.
@rt4914 @BenHenning Any ideas on what functions to use?
<br><br>
@BenHenning Besides this, I had a small doubt. 
- If we are able to make this WindowInsets library work for lower APIs, or 
- if we come up with some UI changes so that it doesn't require checking keyboard's visibility or height.

that would probably be the best solution (second obviously being better).
I just wanted to ask what was actually wrong with the original solution. You said that,
>This seems like a really hacky way to check if the keyboard is open. It's making a lot of assumptions on how large the keyboard is (which is a customizable and OEM-specific property).

I couldn't understand what you meant by **a lot of assumptions**. The only assumption made was for the **minimum** height of a softKeyboard (plus that of system bars). The solution didn't depend on the actual height of keyboard. The exact height is definitely **a customizable and OEM-specific property** but I don't think there is any keyboard which occupies less than 150dp of height. Also, if something is obstructing more than 150dp of device screen, what else can it be other than the keyboard?

mschanteltc: Is it possible to add "extra bottom padding" to the lesson when keyboard is active? When the keyboard is not active, then the "bottom padding" resets to default. Ideally we would animate the transition so the change in the lesson content's position isn't so jarring.

BenHenning: @mschanteltc I don't think so--the problem is that we cannot reliably know if the keyboard is open. I think that Android was designed, in part, to support UIs that don't depend on keyboard-specific behavior.

@ArpitShukIa one issue is that the previous solution was using a 200dp threshold, and even within the stackoverflow discussion linked above there were multiple comments that had subtly different solutions, and others where people mentioned the solution doesn't work on certain devices with different smallest-width or density configurations. That, plus the fact that AndroidX can't properly support this functionality in older API versions with adjustResize I think are fairly strong indicators that trying to detect the keyboard status in this way is fundamentally fragile and likely to result in device-specific bugs (which we really want to try to avoid). Given how large the device ecosystem is for Android (10k's of devices) it's really hard for us to confidently design a size-based solution without the system aiding us, and it seems like the Android team intentionally designed the system to not do that. Does this help clarify?

Regarding with AndroidX API to use, I'm not sure. I found that solution after a search, and you may need to dig more into it to figure out the correct approach for our target API version. Either way, I think it requires us to use adjustPan so we definitely need to decide on whether we should consider a UI redesign or to use adjustPan (or something in-between).

rt4914:
Adding this for better visibility.

The conversation is taking place here  https://github.com/oppia/oppia-android/pull/2933#discussion_r622657683

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
mschanteltc: Is it possible to shift the latest feedback to the top of the page? Regardless of whether or not the keyboard is active, shifting the latest feedback to the top could allow the user to access the input container, though there still might be cases where the feedback and keyboard are large enough to hide it from view.

BenHenning: We should be able to auto-scroll to the feedback, though we may need extra padding to ensure that the content is never underneath the keyboard.

@mschanteltc is your suggestion meant to be for adjustResize or adjustPan for the keyboard?

BenHenning:
Deassigning myself for now since I'll have limited availability over the next two weeks. I believe @rt4914 is looking into this in more detail now.

rt4914:
As per the discussion between me, @ArpitShukIa and @mschanteltc 
we will have to keep the progress bar even when keyboard is visible but we need to optimise on this UI so that it looks a bit more correct (a.) add bottom-padding to submit button (b.) use gradient background for the progress bar so that the button is visible with opacity and does not cut abruptly.

Thanks a lot @ArpitShukIa 
Closing this PR and I will be working on the revised solution.

Issue #2971, opened 2021-03-23
Maven dependencies can pull in extra versions of dependencies already in Bazel build graph


BenHenning:
See https://github.com/bazelbuild/rules_jvm_external/issues/540 for specifics. It's possible for our current build graph to include multiple versions of the same runtime dependencies, which in turn can result in issues like our protos appearing with errors in Android Studio due to being cross-referenced with an earlier protobuf-java dependency but still building on the command line & syncing.

Short of the flag being suggested above, another way to fix this is to move all of our Maven dependencies over to from-source builds or http archive wrappers with exports on other http archive wrappers than export their dependencies. The latter will be harder to build reliably since we won't know about missing dependencies until runtime, and the former will require a substantially longer project build time. It's actually unclear which we should choose (though maybe one option is to have a custom & simpler Maven import rule that effectively uses http_archive to pull in the binary, parses the dependency's POM file to get its dependencies, and then automatically pulls in third-party alternatives for those dependencies).

BenHenning:
Note that I found a way to work around my specific issue: I traced the protobuf dependency back to a Google accessibility library that espresso-contrib depends on. That library updated its protobuf dependency a few years ago, but espresso-contrib only recently updated to the version of the accessibility library with newer protobuf. I just moved us to 3.4.0-alpha05 to get this new dependency (only on Bazel), but if this introduces problems we can always try to instead create an explicit dependency on the newer accessibility library since then maven_install should theoretically pick the higher version in its graph rather than the one that espresso-contrib is asking for.

Sandesh282:
## Techincal Breakdown

`rules_jvm_external` resolves transitive Maven dependencies independently from manually declared `http_archive` deps. This can result in two versions of the same library (e.g. `protobuf-java` 3.x from `http_archive` + 2.x pulled in transitively by another Maven dep) living on the same classpath simultaneously, causing runtime conflicts and Android Studio sync errors.

- **Part 1: Audit the current dependency graph for duplicates**

  **What to do:**
  - a. Run `bazel query 'deps(//app:oppia)' --output=build | grep "maven_jar\|http_archive\|maven_install"` to list all resolved dependency sources.
  - b. Cross-reference `third_party/maven_install.json` against `WORKSPACE` http_archive declarations to identify any library appearing in both (e.g. `com.google.protobuf:protobuf-java`).

- **Part 2: Pin conflicting transitive dependencies in `maven_install`**

  **What to do:**
[...]

PR #2977, opened 2021-03-23
Upgrade Espresso versions to work around build issues

BenHenning:
See #2971 for specifics. This PR is working around a local build inconsistency in Bazel that is causing multiple versions of protobuf to be pulled in. I'm not keen on bringing us several Espresso lib versions ahead, and particularly on a pre-release build, unfortunately the most recent major version still has the same issue (indirectly pulling in an old protobuf version). This should  be satisfactory in the meantime, but as we move tests over to Espresso we will need to be aware that there is a discrepancy between Gradle & Bazel.

Gradle tests are unaffected.


anandwana001 (approved):
Sounds good to me

BenHenning:
NB: This is actually causing issues trying to run the Bazel tests since the new Espresso libraries are making use of Android R's new <queries> tag support in AndroidManifest, and we don't yet target R.

BenHenning:
So I think we'll prefer to pull in the correct dependency version rather than upgrading Espresso. Given this is a localized issue, closing this PR in favor of manually fixing the issue locally for now, and since the solution actually worsens the situation for downstream PRs.

Issue #2940, opened 2021-03-19
Locked lesson thumbnails don't blur when loaded as SVGs


BenHenning:
It seems that when loading real SVG thumbnails, the lock icon appears correct for chapters that can't yet be played, but the blur doesn't work (presumably because the blur transformation is a BitmapTransformation).

BenHenning:
Related: https://github.com/bumptech/glide/issues/2631

Issue #59, opened 2019-08-16
Migrate from Gradle to Bazel


BenHenning:
This isn't an immediate priority, but Gradle has many pitfalls which makes development more challenging:
- There's no control over which resources/manifests are included in the final binary beyond production/test dependencies on modules (which makes creating test modules/libraries more challenging)
- Replacing component implementations (e.g. for test mocks/fakes) is non-trivial with Gradle
- Gradle builds can be a bit slow
- Gradle changes definitions often, so workarounds to any of these problems are likely to break quickly

Bazel allows for more granular control over how dependencies are manged, and in a structured way that will have long-lived support. Bazel includes native support for Android Studio (https://docs.bazel.build/versions/master/bazel-and-android.html) and can support Maven dependencies just like Gradle (https://docs.bazel.build/versions/master/generate-workspace.html).

BenHenning:
NB: This is currently targeting proof of concept, but it might get pushed back to the prototype milestone.

BenHenning:
This is blocked on #4.

BenHenning:
One other benefit: we can structure the project more granularly by directory rather than thinking about module-level separation, or needing to fit a specific structure enforced by Gradle.

BenHenning:
Another benefit: Bazel allows you to force libraries/targets to be test only, making it impossible to accidentally include them in user-facing binaries.

BenHenning:
Another benefit is easier integration with Java proto lite without requiring the hacky proto module workaround that we're currently using. This will be especially nice since it lets us keep the proto files we use close to the code that uses it, rather than having to keep all protos together in their own proto-only module.

BenHenning:
Although we should try to get this done soon, it's not a strict requirement for proof-of-concept and we're a bit time constrained for that milestone. Let's work on this on a best-effort basis in the prototype milestone.

BenHenning:
NB: keeping this as an important priority for the prototype.

BenHenning:
@droidizer is looking into what an integration with Bazel would look like.

Sowmya: let me know if you'd like me to provide more guidance on how Bazel projects are laid out, and how we might be able to translate our existing project over. Initially, I suggest considering:
1. Gradle modules don't need to have parity with Bazel structures (and probably shouldn't--Bazel does a better job at representing packages as libraries)
2. Each directory should have its own BUILD file (and subsequently comprise one or more libraries)
3. The top-level BUILD file should include the android_binary targets for the Oppia application

[...]

PR #2978, opened 2021-03-23
[RunAllTests] Fix part of #2787: Add support for proto lesson loading from local assets

### LINKED ISSUE #2787: Update app to support newest lesson structures
The app needs to be updated to support the latest exploration structures, and re-tested to ensure it properly supports:
- Place values
- Multiplication
- Fractions
- Ratios

BenHenning:
Fix part of #2787

This adds initial support for loading the new lessons needed for alpha. In particular, to better prepare the codebase for post-alpha beta work, this PR is introducing support for directly loading proto assets rather than JSON (bypassing the entire JSON loading pipeline). This requires punching a few holes into various retrievers/controllers (gated by a new qualified boolean: LoadLessonProtosFromAssets) to load proto rather than JSON. The resulting proto code is a bit simpler than the JSON versions since typing becomes much simpler.

In order to ensure the new pathways are (mostly) working in tests, this PR enables proto caching in some tests. However doing this isn't straightforward: we don't want to check actual binary proto files into the codebase since they are binary files and will  unnecessarily bloat our repository. Textproto is a more reasonable alternative since it's plaintext (and actually a bit more compact than JSON), but complicated to use: proto lite does not support loading via TextFormat. Rather than introducing a dependency on JsonFormat (to avoid JSON altogether), I decided instead to introduce a custom Starlark rule in Bazel to, at build time, convert textproto files to binary files & include those at build time. This works quite well: for both the apk itself & tests we can load binary pb files as though we're including them directly. Also, for this reason is why the PR includes textproto conversions of our development assets.

However, the above approach only works for Bazel. To mitigate this, the few test suites that were updated to verify the fixes conditionally enable proto loading based on whether the environment is Bazel. This is a bit hacky, but it seems like the best solution in the interim until we can entirely move off of Gradle per #59.

Other considerations for this PR:
- This removes dependence on test story 1 entirely because it includes structural elements that are incompatible with the pipeline used to convert to text/binary proto. This seems fine since we don't actually need these interactions for testing or development.
- This removes support for confirmed unclassified answers since they aren't used at all, and probably won't be in Android.
- This PR does not introduce support for the up-to-date interactions that will be shipped with the alpha. That will happen in a later PR (mainly #2985).
- The textproto files introduced in this PR were generated using an in-progress script that's also being used to download the assets that will be used for alpha. Follow up with me if you want to see the script.
- Enabling proto caching & running the APK locally will yield to images not loading. This is partly due to #2979 (in that they can't load without that fix), but mostly due to the fact that all of our dev assets short-circuit thumbnails (which is why we never discovered #2940 until working on alpha)


domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -396,12 +426,25 @@ class TopicController @Inject constructor(
...
-      .setDiskSizeBytes(computeTopicSizeBytes(getAssetFileNameList(topicId)))
+      .setDiskSizeBytes(computeTopicSizeBytes(getJsonAssetFileNameList(topicId)).toLong())
       .addAllSubtopic(subtopicList)
       .setTopicPlayAvailability(topicPlayAvailability)
       .build()
   }
 
+  private fun loadSubtopic(topicId: String, subtopicId: Int): Subtopic {
+    val subtopicRecord = assetRepository.loadProtoFromLocalAssets(
+      assetName = "${topicId}_$subtopicId",
```
vinitamurthi, line 437: I'm wondering if there's a better way to sent this request rather than the assetname which is a slightly vague variable name since it looks like it's picking up assets for different ids. Maybe asset type (as an enum), asset Id (which is called asset name here) , and base message?

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -123,6 +130,23 @@ class TopicListController @Inject constructor(
   }
 
   private fun createTopicList(): TopicList {
+    return if (loadLessonProtosFromAssets) {
+      val topicIdList =
+        assetRepository.loadProtoFromLocalAssets(
+          assetName = "topics",
```
vinitamurthi, line 135: Here this kind of breaks the paradigm for assetName compared to the fields that are being used above, so I think it may be better to keep assetType, assetId, and baseMessage

vinitamurthi (changes requested):

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
BenHenning: Not sure I follow--what advantage does an enum give us here? The main typing mechanism is the base message type, and introducing an enum may require us to keep the types & enum in-sync which will add complexity.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
BenHenning: It's not a type though, it's the actual file name checked into the APK. Calling it 'assetType' would be misleading. Perhaps 'assetBaseFIilename' would be the most clear, but it's a bit lengthy. I also don't understand how you're distinguishing between 'type' and 'ID'. This is the topic list so it doesn't have an ID.

It could be named better but I'm following existing filenames to keep things a bit simpler. We'll be removing all of this code in coming months, anyway, in favor of downloading & caching completely dynamically.

BenHenning:
NB: I will change this & downstream PRs to use '[RunAllTests]' to force test runs since these PRs appear to have issues triggering the 'compute affected tests' flow (probably because of changing BUILD files--that sometimes causes the script to hang and I haven't yet figured out why).

BenHenning:
Thanks @vinitamurthi. FYI that this is ready for full review--PTAL.

@anandwana001 & @rt4914 PTAL as codeowners.

rt4914 (approved):
LGTM for code-owner files.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
vinitamurthi: My suggestion was because we are setting the asset name to be of a different format based on the asset types (i.e. subtopics have a different format of asset name (being topicId_subtopicID) , where as other assets have a different format for their asset names. But it looks like this is because of the way the file names are created, and given that this code may be unnecessary in a few months I guess its okay.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
vinitamurthi: Same comment as above..given that this code will be dead soon I think its okay

vinitamurthi (approved):

domain/domain_assets.bzl
```diff
@@ -0,0 +1,213 @@
+"""
+Macros for preparing & creating assets to include in the domain module.
+"""
+
+load("@rules_proto//proto:defs.bzl", "ProtoInfo")
+
+def _extract_proto_sources(deps):
+    """
+    Returns the list of proto source files that make up the specified list of proto dependencies.
```
anandwana001, line 9: we can use the `#` for comments, nit fix, not blocking

anandwana001 (approved):
LGTM, nit change but can be ignored

domain/domain_assets.bzl
BenHenning: @anandwana001 I don't think we should be. See https://skydoc.bazel.build/docs/writing.html for examples on writing Skydocs. They use multi-line strings, not ``#``. I think we should be treating ``#`` as line comments and ``"""`` as documentation (which also matches Oppia web).

I did notice I'm not formatting on the correct line, but we can fix that in a later PR by making all documentation consistent.

BenHenning:
ClassroomService Bazel failure will be fixed by #2916.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
BenHenning: Yeah--I don't think we should over-index on this since the long-term solution should utilize more of a database-type structure for managing files. Thanks for bringing this up, though!

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
BenHenning: Ack, thanks.

BenHenning:
Thanks all for the review! Merging this now to keep the PR chain moving forward.

Issue #2942, opened 2021-03-19
LaTeX SVGs support


BenHenning:
Some of the alpha lessons require LaTeX. Note that this issue is tracking introducing support for rendering LaTeX SVGs, not rendering LaTeX itself.

PR #2980, opened 2021-03-23
Fix #2134: Fix concept card naming/linking

### LINKED ISSUE #2134: Concept Card Name Broken
A Concept Card link in Lesson Player was presented incorrectly. This occurred in the topic **Place Value**, story **Jaime's Adventures in the Arcade**, in **Chapter 3: Comparing Numbers**. On other occasions where I triggered a Concept Card, this did not occur. It might be because the Concept Card is empty. [[Video](https://drive.google.com/file/d/1fFQUzUh9dTlPwUOlREag-OGh5JmWTrvq/view?usp=sharing)]

Screenshot_20201120-005220_Oppia

BenHenning:
Fix #2134 

This fixes an issue with HtmlParser where we weren't correctly unescaping the concept card name/link. Note that this inconsistency may have arisen from the shift to using CustomHtmlContentHandler which partially parses the JSON-escaped custom attribute values (which turns ``&amp;&quot;`` to just ``&quot;``).

This PR also consolidates all custom tags into using the CustomHtmlContentHandler to avoid global unescaping (e.g. replacing all ``&amp;&quot;`` with empty string) since that's incorrect and breaks encoded JSON objects (one of which is used for math tags).

There are also some minor refactors/changes occurring here:
- The concept card tag handler is being split into its own file
- A new handler is being introduced to handle image handling to address the inconsistency above
- Tests are added for both the new handlers & for HtmlParserTest (there are already higher-level tests for concept cards in StateFragmentTest)
- The Mockito helper was moved from app module to testing so that it can be used in the new utility tests
- This PR is preparing for more complex downstream image loading (see #2942) by introducing an ImageRetriever interface for CustomHtmlContentHandler that allows custom tag handlers to specify how images should be styled (only the current 'block' format is supported; another will be added in a later PR for in-line images)
- There are fixes added to HtmlParserTest to make it more correct: it could previously interact with UI objects off the main thread (this was discovered when trying to run the updated tests with Espresso). Some helper methods were added to perform operations on the main thread with issues.


utility/src/main/java/org/oppia/android/util/parser/ConceptCardTagHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+    val text = attributes.getJsonStringValue("text-with-value")
+    if (skillId != null && text != null) {
+      val spannableBuilder = SpannableStringBuilder(text)
+      spannableBuilder.setSpan(
+        object : ClickableSpan() {
+          override fun onClick(view: View) {
+            customOppiaTagActionListener?.onConceptCardLinkClicked(view, skillId)
+          }
+        },
+        0, text.length, Spannable.SPAN_INCLUSIVE_EXCLUSIVE
```
rt4914: Suggest adding inline comment
`/* start= */ 0, text.length, Spannable.SPAN_INCLUSIVE_EXCLUSIVE`

utility/src/main/java/org/oppia/android/util/parser/ImageTagHandler.kt
```diff
@@ -0,0 +1,48 @@
+package org.oppia.android.util.parser
+
+import android.text.Editable
+import android.text.Spannable
+import android.text.style.ImageSpan
+import org.oppia.android.util.logging.ConsoleLogger
+import org.xml.sax.Attributes
+
+/** The custom tag corresponding to [MathTagHandler]. */
```
rt4914: `MathTagHandler ` file does not exists.

rt4914 (commented):
Overall implementation looks good. Thanks.

utility/src/main/java/org/oppia/android/util/parser/ImageTagHandler.kt
BenHenning: Thanks--fixed.

utility/src/main/java/org/oppia/android/util/parser/ConceptCardTagHandler.kt
BenHenning: Nice catch--fixed.

BenHenning:
@rt4914 ready for review now.

BenHenning:
PTAL @anandwana001 as well since some of your codeowners files are affected.

anandwana001 (approved):
Test files LGTM

rt4914 (approved):
LGTM, Thanks.

BenHenning:
I plan to take a chance on Bazel for this & the next few PRs to try and get them in a bit faster. It's not ideal but I'm fairly confident the PRs won't introduce issues since Bazel tests were passing for them before.

BenHenning:
Also, thanks for the reviews!

PR #2981, opened 2021-03-23
Fix #2942: Add support for LaTeX math tags via SVG rendering

### LINKED ISSUE #2942: LaTeX SVGs support

BenHenning:
Fix #2942

Introduces support for LaTeX math expressions. Oppia's custom tag for math expressions includes an embedded JSON object (as a markup attribute) that includes both the LaTeX expression & a SVG filename that can be rendered in place of the raw LaTeX. This PR does the latter: it hooks up LaTeX-based images to our existing ``TextView``s by leveraging our Glide pipeline. No new third-party dependencies are needed as part of this change since we are not actually rendering the LaTeX.

While the app does have existing support for rendering SVGs, making them work in the context of math expressions involved a lot of additional changes:
- A new image loading pathway for in-line images needed to be added. ``UrlImageParser`` previously presumed all images were to be laid out on their own (optionally centered) rather than being within a line of text.
- The SVG library that we used wasn't sizing math SVGs correctly. Since they rely on ex for measurement (similar to em except they're horizontal units), they need to know details about the text environment to be sized correctly. The SVG library would use a default ``TextPaint`` for this purpose rather than the one actually used by the parent ``TextView``. Changes needed to be made in the library to support this, so we now use a [forked version](https://github.com/oppia/androidsvg) of the library.
- The AndroidSVG library also needed to be updated to retrieve the vertical alignment for the text to ensure it's properly centered on the line. This required some awkward extra bound value handling in the rendering pipeline.
- This is the first time we've incorporated a custom dependency fork in our Gradle pipeline which means we are now utilizing jitpack as a Maven repository. This isn't ideal since jitpack can take a long time to assemble the dependency, so we may see intermittent build flakiness ~1 time per week. Hopefully the CI environment helps mitigate for this with its frequent runs.
- A new ``MathTagHandler`` + tests was added to support Oppia's custom math tag

Other changes & details:
- ``UrlImageParser`` was awkwardly using a ``BitmapDrawable`` when it didn't need to, so I introduced a ``ProxyDrawable`` and generally cleaned up that entire class.
- Only basic testing was added for ``UrlImageParser``. While I'd like to add more tests, it's not yet obvious how to without using actual Glide and that's not obvious since we need to fake the networking layer. More thought is needed here (maybe we can leverage a OkHttp mock web server to fake an image file server?)
- Some refactoring was done in the image loading pipeline:
  - Introduction of ``ScalableVectorGraphic`` class to invert our dependency on AndroidSVG (in case we ever chose to move off of it in the future)
  - A new ``SvgPictureDrawable`` with two subclasses: ``BlockPictureDrawable`` and ``TextPictureDrawable`` to support both block and in-line SVG rendering. Note that 2 drawable types are needed to simplify arranging the drawable in ``UrlImageParser``.
- A bug was fixed in ``RepositoryGlideModule``: when loading local SVG drawables it would actually use AndroidSVG instead of Glide's pipeline for local drawables. This PR fixes this by introducing a ``RequestOptions`` option for specifically enabling our SVG loading pipeline when we want it to be used (not for local drawable cases).
- Blurring doesn't work with SVGs currently. That will be fixed in #2984
- ``TestGlideImageLoader`` was improved to track specific images being loaded which has led to some additional testing coverage via new tests in ``HtmlParserTest``


BenHenning:
Unfortunately, I need to close & reopen this PR to properly restart the Bazel tests (manually restarting doesn't seem to use the ``[RunAllTests]`` in the title). However, I'll be merging an upstream PR so I'll just wait for that to come down through the branches to trigger a re-run.

BenHenning:
PTAL @anandwana001 & @rt4914 this is ready for review now.

anandwana001 (approved):
Test files LGTM

rt4914 (approved):
Approving for Code-owner files.

BenHenning:
Thanks for the reviews!

PR #2983, opened 2021-03-23
Fix #2943: Ensure all thumbnails are consistently opaque

### LINKED ISSUE #2943: Thumbnails don't properly ignore the alpha channel
If we receive a color that includes a 0 alpha channel, the color appears as transparent due to app-level processing of the RGB value not including a filled-in alpha channel. The proto properly defines that it only contains RGB values, but the app is using it in cases where Android is expecting ARGB values. This results in all current thumbnails appearing with white backgrounds when using real lesson data.

BenHenning:
Fix #2943

The lesson thumbnail proto defines that the color provided from the domain layer only contains 3 color channels: RGB. We sometimes include alpha, but per the proto it's the responsibility of the UI layer to ensure alphas are properly defined. We do this in some places, but not all. This fixes the other places where we aren't properly filling in the alpha channel.

Regarding tests, some of the changes are a bit harder to test since new test suites need to be added (filed #3098 & #3099 to track these). It's also reasonably fine to live without these tests for a bit because:
- Verifying backgrounds in Robolectric is much less effective than screenshot tests (though for the sake of this specific regression that's not as relevant, but we should focus on visual correctness longer-term via screenshot tests)
- There's little chance of detecting regressions in those tests since this only happens with proto loading and current test suites are using JSON), though this is partly mitigated in some of the testing added (see below)

A new test was added for ``WalkthroughTopicListFragmentTest`` using a new ``GenericVIewMatchers`` class that includes an Espresso matcher for opaque backgrounds. This actually works reasonably well since the test was updated to load the proto versions of lessons (see #2978 for a description on how this mechanism works). I verified that without some of the alpha adding code this test fails. Unfortunately, loading proto assets broke the thumbnail image test with no obvious mitigation, so some TODOs were added for those.

Finally, this introduces a new Espresso package in the testing directory to contain Espresso matchers/utilities. I moved the two relevant classes from the top-level testing module directory to this new package to clean things up a bit.


rt4914 (approved):
LGTM, thanks.

BenHenning:
@rt4914 & @anandwana001 this is now ready for review. PTAL.

Note @rt4914 that I re-requested your review since this PR has grown substantially since your last review.

anandwana001 (approved):
test files LGTM

rt4914 (approved):
LGTM, thanks.

BenHenning:
Bazel failure is due to flaky data module tests (known issue--unrelated to this PR).

BenHenning:
Thanks for the reviews!

Issue #1815, opened 2020-09-05
Introduce screenshot-testing framework


MohamedMedhat1998:
## Screenshot-testing
Screenshot testing is a kind of testing that is used to detect unintentional visual changes or regressions in the UI.   
This can be done by comparing two screenshots together, one is the reference (the golden) screenshot, and the other is the screenshot of the app after some code change.  

## Sub-Tasks
The screenshot testing process can be divided into the following tasks:  
1. Record
2. Verify
3. Update

### Record
It is the process of capturing the golden/reference screenshots so that they can be used in later verifications.

### Verify
The actual testing in which a new screenshot is taken to be tested, then it is compared with the reference screenshot that was taken during the “record” process. If the two screenshots are identical (they are pixel-perfect), the test passes, if not, the test fails.

### Update
It is the process of updating the current golden screenshots. It is used when we intentionally change the UI and want to take a new screenshot of that UI so that it becomes the new reference.

BenHenning:
I have thoughts on this finally! This is an issue that I consider on occasion every few months to try and figure out how this could be made to work. There's generally two main problems that need to be solved:
1. How to set up tests to actually perform screenshot testing.
2. How to manage golden images in the repository.

For (1), my best thoughts here is that we do something like the following:
- Introduce a new Kotlin JVM (i.e. not Android) library that can perform a simple diff between two screenshots (pixel-based matching seems fine). In the future we could consider features like fuzzy matching or subregion matching, though I'm hoping to preempt the latter and to hope the former is never needed.
- Introduce a ScreenDiff utility that can be called in an Espresso test (Robolectric may be possible in the future, but that involves pretty difficult additional work to hook up Robolectric rendering). This utility essentially:
  - Takes a screenshot of a view by ID.
  - Loads the golden image from the local repository.
    - We'll need to come up with a reasonable naming schema for goldens and golden paths, but I suspect we'll probably use the fully qualified test suite and test name.
    - Note that we'll need to include the golden images as part of the test build (either as resource or data files). 
  - Uses the library above (which would be pulled into the Android-based ScreenDiff utility library) to diff the screenshot with the golden, and fail if there's a difference.
  - **Important**: The ScreenDiff utility will need to save the failing image in the test's output directory.
- Write a Kotlin script that can extract the newly generated goldens and replace all of the existing goldens accordingly. This is the script developers will use when they want the goldens to update.

[...]

Issue #1523, opened 2020-07-25
Allow tests to orchestrate Glide to test image loading in testing environments


aggarwalpulkit596:
**Is your feature request related to a problem? Please describe.**
Currently, there is no way to test of loading an image using Glide with the help of caching the image locally. Tests of images shouldn't depend on the internet and should use cache but for ImageInteraction we can't use Roboelectric neither we can use the caching setup with Gradle.

**Describe alternatives you've considered**
In #1492 we've used a timeout of 5 seconds but that can be flaky depending upon the type of internet connectivity also that is an antipattern for these types of tests (see https://developer.android.com/training/testing/espresso/idling-resource#identify-when-needed)

seanlip:
Removing @aggarwalpulkit596 due to inactivity.

PR #2984, opened 2021-03-23
Fix #2940: Fix chapter lesson blurring for SVG thumbnails

### LINKED ISSUE #2940: Locked lesson thumbnails don't blur when loaded as SVGs

BenHenning:
Fix #2940

Adds support for blurring SVG images. The main issue is that our previous transformation is a ``BitmapTransformation``, and those don't apply to SVGs. This actually took me some time to figure out (https://github.com/bumptech/glide/issues/2631#issuecomment-803228591 was one of my braindumps). Glide does have support for attaching custom transformations to custom types (such as our ``ScalableVectorGraphic`` class). Beyond that, I couldn't find a good way to blur the SVG, so I added a pathway to prerender the SVG to a bitmap and then blur that. Not only does this consolidate our blurring to one routine (which adds consistency), it also is likely more performant to render bitmaps than SVGs. I might add bitmap pre-rendering for all SVGs since sometimes the picture rendering fails, and bitmap rendering seems more reliable.

Testing is tricky here. We ideally need Glide testing to verify upstream that the drawable is created & arranged correctly (tracked by #1523), and we need screenshot testing to verify visual correctness especially for blurring (#1815). I'm not clear on what else I can test--if anyone has ideas, please let me know.

In order to test this locally, you'll need to download the alpha assets locally. This is a bit tricky to do, so please follow up with me if you want to verify locally. Otherwise, you can verify this is correct via the alpha test binary after all of the alpha PRs are merged in.

One side change: I updated the bitmap transformation to correctly use a message digest/equals/hashCode as recommended by Glide documentation: https://bumptech.github.io/glide/doc/transformations.html#required-methods. From my understanding, the previous implementation is unlikely to run into any issues but the changes make the transformation more correct per documentation.

Screenshot of blurring with SVG:

image


rt4914:
@BenHenning For this can you mention the steps to load the alpha content so that I can check on that.

BenHenning:
A bug I found while testing with production alpha: sometimes images won't load in the list (& will either white out or just be the filled color). This seems to only affect the story list, and sometimes they appear after scrolling & waiting and other times they don't (even with app restart). It seems like something is failing when converting to bitmap & it never recovers.

rt4914:
@BenHenning Does this PR still needs a review?

BenHenning:
Yup @rt4914, but it looks like no codeowners reviews are necessary. I'll ask @viktoriias to review since she introduced the original blurring code, and you since you're reviewing #2981 and that's introducing most of the SVG stuff used here.

BenHenning:
@rt4914 & @viktoriias PTAL.

utility/src/main/java/org/oppia/android/util/parser/BitmapBlurrer.kt
```diff
@@ -6,52 +6,48 @@ import android.renderscript.Allocation
...
- * The following [tutorial](https://futurestud.io/tutorials/glide-custom-transformation)
- * was used as a reference, as well as this [article](https://stackoverflow.com/a/23119957).
+ * The radius of the blur, a float between 0 and 25. It defines the value of the standard deviation
+ * to the Gaussian function.
  */
-class BlurTransformation(private val context: Context) : BitmapTransformation() {
+private const val BLUR_RADIUS = 20f
 
+/** Utility used to blur [Bitmap]s. */
+internal class BitmapBlurrer(private val context: Context) {
```
rt4914, line 17: Though this class name makes sense to me but reading it sounds a bit awkward.

rt4914 (approved):
LGTM.

viktoriias:
LGTM

viktoriias (approved):

utility/src/main/java/org/oppia/android/util/parser/BitmapBlurrer.kt
BenHenning: Might be because 'blurrer' isn't a word. :) Any suggestions for alternative names?

rt4914: `BitmapBlurEffect` , `BitmapBlurImpl` ?

BenHenning: 'Impl' doesn't really fit since it's not implementing an interface/API. 'BlurEffect' is a bit off, too, since 'effect' actually means something in Android and this isn't that (https://developer.android.com/reference/android/media/effect/Effect).

I think I'll keep this now (for time), but if you have additional thoughts I'm happy to discuss.

BenHenning:
Looks like the Gradle failure is due to the uncommon LogUploadWorkerTest flake being tracked by #2032.

BenHenning:
Thanks for the reviews!

Note also that the code reviewer request above isn't right--it was a temporary request due to the base branch changing without the branch itself being updated to be properly based on develop post the squash merge of #2983.

BenHenning:
Hmm. CI isn't triggering for some reason. Closing & reopening PR to see if that triggers it.

BenHenning:
Note that I do plan on letting the full CI checks run on develop now that this & the previous 4 PRs have been merged.

Issue #2976, opened 2021-03-23
Add support for loading questions from protos


BenHenning:
This is the tracking issue for supporting downloading & loading questions in the Android app. This is primarily dependent on the content pipeline creating sufficient questions to justify bringing them into the app.

BenHenning:
Alternatively, we could just add support for loading the current known question structure in the interim as a proof-of-concept for loading questions with protos. That seems better than the alternative which will likely lead to questions being disabled and potentially regressing once the JSON loading pipeline is deprecated.

BenHenning:
This will need to be done as part of addressing #5571 since the app effectively doesn't support JSON anymore for production builds (so we can't test questions in alpha without using protos).

Issue #352, opened 2019-11-15
Change concept card to show Worked Examples


rt4914:
Currently, Concept Card does not show `Worked Examples`, the domain layer code has already been written, but for now, it is not getting used.

## How to show concept cards
**Approach 1**
- Go to Meaning of Equal parts Exploration
- State 1-> Option D
- State 2-> 3/2 answer
- You will notice concept-card link there.

<p>
<img src="https://user-images.githubusercontent.com/25057618/103724911-75ff8e00-4ffb-11eb-8ce6-fa04c0dd6229.png" width=250  />
<img src="https://user-images.githubusercontent.com/25057618/103724924-7ac44200-4ffb-11eb-9272-793f5628f26f.png" width=250  />
</p>

**Approach 2**
- In `AndroidManifest` file make `ConceptCardTestActivity` as launcher activity.
- Run the app.
- Click on buttons to see different types of concept cards.

**Expected Final Output UI**
[Mobile, Portrait](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/043cd31c-abb3-4ab2-bf5e-9ec443c0e073/)
[Mobile, Landscape](https://xd.adobe.com/view/ee9e607b-dbd6-4372-48dc-b687d32af3da-98af/screen/20bcd85e-23c3-4f60-b51c-19aac5a908ab/)
[Tablet, Portrait](https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/203c578c-57da-461a-9d4d-60fed26f9b64/)
[Tablet, Landscape](https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/883dee89-a1cb-4932-9c7d-ab7fb585238c/specs/)

- [ ] The button should display `View Example` if there is only `1` worked example and it should display `View Examples` for more than 1. Button should be gone in case of `0` worked examples.

viktoriias:
@rt4914 Is there a link to the design/mocks?

rt4914:
> @rt4914 Is there a link to the design/mocks?

No actually. but for now we can just implement low-fidelity without worrying about the exact design and then work on the perfect UI. @viktoriias

viktoriias:
@rt4914 I would like to work on this, but I'm not sure what `Worked Examples` are. Which screen would they appear in?

rt4914:
[...]

PR #2985, opened 2021-03-23
[RunAllTests] Fix #2944,  #2787, #298, #3007: Update interactions to support new localization structures

### LINKED ISSUE #298: Update demo explorations to the latest schema & prohibit older ones from being loaded
We have run into some issues where state schema discrepancy results in inconsistent issues when trying to parse the JSON. We should avoid these by upgrading all explorations to the latest schema structures, and fixing each of our loaders to a specific schema version to ensure consistency.

### LINKED ISSUE #2787: Update app to support newest lesson structures

### LINKED ISSUE #2944: New localization structures for item selection & multiple choice aren't supported by the app
New localization support for item selection & multiple choice interactions aren't supported by the app. We'll need to add support both for these answer types, and will need to update existing lessons.

### LINKED ISSUE #3007: Update ExplorationProgressControllerTest to support protos
#2978 is adding support for loading text protos in place of JSON, and this is going to be enabled in ExplorationDataControllerTest, TopicControllerTest, and TopicListControllerTest. As part of this work, old explorations that couldn't easily be migrated were dropped (one of which is test_exp_id_0 that ExplorationProgressControllerTest heavily depends on). Since the progress controller is primarily interested in verifying an end-to-end playing experience & ensuring all interactions are working, the prototype exploration should be a suitable replacement: test_exp_id_2.

This issue is tracking migrating ExplorationProgressControllerTest over to using test_exp_id_2 & then enabling proto tests in Bazel. Note that the main test work here can be done entirely Gradle & verified in Bazel to make things easier.

BenHenning:
## Explanation
Fix #2944
Fix #2787
Fix #298 (the prohibit bit isn't happening, but JSON loading will eventually be going away so that seems fine)
Fix #3007

This is the primary PR for bringing updated lesson structure support (mainly for explorations/questions) into the app. There were a few migrations in Oppia web since the last time we updated the app's structure (September), and most of them corresponded to the continued effort to bring answer localization support to the platform. At a high-level, the following structural changes affected the app (directly & indirectly):
- The answer & input types changed for item selection to correspond to a set of content IDs/a content ID (respectively) rather than actual options
- Customization args for item selection & multiple choice brought content IDs into them via SubtitledHtml
- DragAndDropSortInput's answer & input types changed for item selection to correspond to a list of set of content IDs rather than actual options
- TextInput's input parameters type has changed to be a list of strings, though I think currently creators can still only specify one
- TextInput's CaseSensitiveEquals rule was removed in favor of always using a case-insensitive check (i.e. the current 'Equals' rule). It seems all of TextInput's rules are now case-insensitive checks, and a few other providers needed to be updated in the codebase to match this.
- Worked examples in concept cards now have both explanations and questions, though the app is only parsing the explanation fields (and isn't yet using them--#352 is tracking adding support for questions and explanations long-term)

Handling all of the above core changes required substantially changing both structural & testing aspects of the codebase. Some additional changes were also needed due to asset & conversion limitations (I'll mention those in more detail below). For production behavior changes, most of the changes just needed to happen within the UI view models for interactions, and in the classifier provider classes in the domain layer. Nearly all other changes fall into the 'testing' category.

To ensure the app is behaving correctly, I updated our script that we're using to download the alpha assets to instead parse the old JSON files and convert them both to new JSON and textproto. However, I noticed that we were inconsistently handling different schema versions. Rather than trying to support multiple schemas versions, I removed the old lessons. This result in a lot of cleanups (e.g. some parsing code pathways could be cleaned up), but also required in a lot of test asset file changes (which are auto-generated from manual changes to the original assets) & a *lot* of test changes. To provide some context, I made the following changes to the assets to try and simplify things:
- test_exp_id_0, ratio_input_exploration, test_exp_id_1, test_exp_id_3 were all removed (which required rewriting much of StateFragmentTest and ExplorationProgressControllerTest, among others)
- test_exp_id_2 includes some additional answer groups to ensure all of RatioInput's rule types are being included (for tests)
- test_exp_id_2 was updated to include hints & a solution to migrate some existing domain tests that relied on a now-removed test exploration
- Manual fixes were needed in questions.json for past fields that were using camel case instead of snake case, and for missing fields/inconsistent structures. This led to some minor test fixes in QuestionAssessmentProgressControllerTest.

Regarding questions, while this does bring in updated questions for JSON and non-empty questions for proto (via a new QuestionListRecord proto), it does not actually bring proto question support yet (which is tracked by #2976). I spent about an hour trying to get this to work, and unfortunately the main issue is that the textproto->pb Bazel Starlark rule seems to not like the question proto since it's the only formatted proto that depends on another formatted proto (explorations). I didn't have more time to dig into it, though I suspect it's fixable.

Besides the above, other noteworthy changes here:
1. This fixes #3007 by enabling proto support in that test when run on Bazel. This has also been enabled for other tests in the PR chain to try and ensure different aspects of the proto pipeline are being covered in tests.
2. I fixed a lot of small style/line length/formatting/consistency issues throughout the codebase as I noticed them (particularly noteworthy is the StateRetriever class which had a bunch of areas for improvement). That unintentionally increases the length of the PR--sorry about that.

Note also that the changes seem to be passing on Espresso (or, at least, no new failures are observed): https://github.com/oppia/oppia-android/pull/2985#issuecomment-826030955.


domain/src/main/java/org/oppia/android/domain/classify/rules/itemselectioninput/ItemSelectionInputEqualsRuleClassifierProvider.kt
```diff
@@ -1,32 +1,35 @@
...
   }
 
-  override fun matches(answer: StringList, input: StringList): Boolean {
-    return answer.htmlList.toSet() == input.htmlList.toSet()
-  }
+  override fun matches(
+    answer: SetOfTranslatableHtmlContentIds,
+    input: SetOfTranslatableHtmlContentIds
+  ): Boolean = answer.getContentIdSet() == input.getContentIdSet()
 }
```
vinitamurthi, line 35: More for my understanding, does this handle mismatched ordering as well?

domain/src/main/java/org/oppia/android/domain/classify/rules/textinput/TextInputStartsWithRuleClassifierProvider.kt
```diff
@@ -1,32 +1,36 @@
...
       "x",
       this
     )
   }
 
-  override fun matches(answer: String, input: String): Boolean {
-    return answer.normalizeWhitespace().startsWith(input.normalizeWhitespace())
+  override fun matches(answer: String, input: TranslatableSetOfNormalizedString): Boolean {
+    val normalizedAnswer = answer.normalizeWhitespace()
+    return input.normalizedStringsList.any { normalizedAnswer.startsWith(it.normalizeWhitespace()) }
```
vinitamurthi: Does this handle case inequality?

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -28,18 +32,25 @@ class DragAndDropSortInteractionViewModel(
...
+      subtitledHtml.contentId to subtitledHtml.html
+    }
+
+  private val _choiceItems: MutableList<DragDropInteractionContentViewModel> =
+    computeChoiceItems(contentIdHtmlMap, choiceSubtitledHtmls, this)
+
+  val choiceItems: List<DragDropInteractionContentViewModel> = _choiceItems
 
-  var isAnswerAvailable = ObservableField<Boolean>(false)
+  private var isAnswerAvailable = ObservableField<Boolean>(false)
```
rt4914: Though this is not part of your PR but lets make it `val`

rt4914 (changes requested):
@BenHenning Are you planning to update json files in this PR as currently if it gets merged it will break the current develop.

BenHenning:
So a brief update on this: this PR is going to be *much* larger than I originally planned. The core structural changes that occurred in web that we're trying to bring in are just for multiple choice, item selection, drag & drop, and text input. Each of these were updated to support localized answers which requires a small structural change. However, these proliferate majorly throughout the codebase:
- Due to the size of the test lessons, I needed to create a script to migrate the old json format to the new one
- As part of the migration, a few issues came up that needed to be addressed (e.g. bugs in the migrator, and needs to add additional functionality to the existing test lessons)
- To simplify the migration, some old lessons were dropped in support (such as the welcome to Oppia lesson), partly due to the fact that the app was actually supporting multiple schema versions. This normalizes that.
- Removing some of the lessons required porting tests over to different lessons
- The new structural changes to answer types required pretty far reaching changes to existing tests
- The JSON loading pipeline & some of the surrounding support code need/benefited from upgrades

The main issue I ran into was questions. I thought we could get away with disabling questions for now (e.g. #3113), but I don't think that is going to work anymore: too many tests rely on questions functionality, and disabling so many regression tests for even a few months is likely to introduce a lot of issues when we try to re-enable it later. Because of this, I think I'm actually going to need to migrate our existing test questions even though the production version of the app (e.g. alpha MR2) is not going to support questions at the current time. This is going to take me a few more hours before I can get this PR into review.

domain/src/main/java/org/oppia/android/domain/classify/rules/itemselectioninput/ItemSelectionInputEqualsRuleClassifierProvider.kt
BenHenning: Yep since ``getContentIdSet()`` returns an actual ``Set``, and ``Set`` equality doesn't rely on ordering: https://docs.oracle.com/javase/8/docs/api/java/util/Set.html#equals-java.lang.Object-.

domain/src/main/java/org/oppia/android/domain/classify/rules/textinput/TextInputStartsWithRuleClassifierProvider.kt
BenHenning: Nice catch--this is a gap I missed. Either we implemented this wrong originally, or as part of getting rid of CaseSensitiveEquals the other TextInput rules were updated to be case-insensitive. Fixed it in the latest batch of changes that should be pushed momentarily.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
BenHenning: Done. Changes should be available to review soon.

BenHenning:
Thanks for the initial review. FYI @rt4914 & @vinitamurthi that this is now ready for a full review.

@rt4914 I did mean to originally update the JSON structures when I sent this out, I just hadn't gotten to it yet. It turns out that led to a bunch of additional work I hadn't anticipated being as large as it ended up being. :)

Also,  please note that I haven't yet run the app tests via Espresso. I will do that tomorrow and fix any issues that arise. Please don't block on that detail unless you have specific concerns (in which case, please let me know).

vinitamurthi (approved):
LGTM!

rt4914 (approved):
LGTM (haven't run the espresso tests)

BenHenning:
Thanks for the reviews! It seems StateFragmentTest & ConceptCardFragmentTest are the only shared tests I changed in this PR, so I'm only testing those with Espresso. I'm using Robolectric passing as a rough proxy for other indirectly affected shared tests since running the whole suite is a bit too time consuming, and likely already has existing failures.

- Existing failures in StateFragmentTest (observed on develop): all loadImageRegion tests
- Observed failures in StateFragmentTest from this PR: none (all pass)
- Existing failures in ConceptCardFragmentTest (observed on develop):  all loadImageRegion tests
- Observed failures in ConceptCardFragmentTest from this PR: none (all pass)

It appears no new tests are failing for Espresso tests.

BenHenning:
Given no open comments, passing CI, and no new detected local Espresso failures, merging this.

PR #2986, opened 2021-03-23
Fix #2952: Add support for disabling practice tab

### LINKED ISSUE #2952: Need support for disabling the practice tab
Since there are very few questions available in production, we need to be able to disable the practice tab so that users don't run into a broken or confusing experience. This will be used in the next alpha release.

BenHenning:
## Explanation
Fix #2952

Introduces support for disabling the practice tab in the topic fragment. This is done via a new Dagger boolean qualified by ``EnablePracticeTab``. In order to easily disable the tab, ``TopicTab`` has been updated to actually compute/manage both tab positions and count based on whether the practice tab should be enabled.

Some other minor refactoring was done in ``TopicTab`` to try and consolidate tab-specific resources & strings into a single location which avoids extra ``when`` statements elsewhere.

I verified that TopicFragmentTest is also passing locally when run with Espresso on a Nexus 5X.


rt4914:
@BenHenning Does this PR still needs a review?

anandwana001:
I will review this today.

anandwana001 (approved):
LGTM, Thanks @BenHenning

anandwana001:
https://github.com/oppia/oppia-android/pull/3072 blocked on this PR as #2986 has a lot of change around `topicTab` enum and TopicInfo which also requires here in this PR. 
@rt4914 Could you review this pr so we can merge it asap.

app/src/main/java/org/oppia/android/app/topic/revisioncard/PracticeTabModule.kt
```diff
@@ -0,0 +1,12 @@
+package org.oppia.android.app.topic
+
+import dagger.Module
+import dagger.Provides
+
+/** Provides dependencies corresponding to the practice tab. */
+@Module
+class PracticeTabModule {
```
rt4914: I think by mistake this file shifted to `revisioncard` package.

rt4914 (approved):
LGTM, one nit change and rest is blocked on tests.

BenHenning:
Ack thanks for the notice of blocking @anandwana001. I'm working to get all upstream work this depends on in. Nearly everything before this PR is now in review (only #2985 isn't yet).

app/src/main/java/org/oppia/android/app/topic/revisioncard/PracticeTabModule.kt
BenHenning: Nice catch. Fixed (will be visible in the next push to this branch).

BenHenning:
Thanks @anandwana001 & @rt4914 for the reviews.

@rt4914 given some new modules were migrated, your earlier comment, and the new test to verify the tab properly hides, PTAL again when you get a chance.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -214,8 +225,39 @@ class TopicFragmentTest {
...
+              R.id.topic_tabs_container
+            )
+          )
+        )
+      )
+    }
+  }
+
+  @Test
+  fun testTopicFragment_practiceTabDisabled_practiceTopicTabIsNotDisplayedInTabLayout() {
```
rt4914, line 247: I think there should be tests to check that on configuration-change the tab remains hidden/visible.

rt4914 (approved):
@BenHenning Overall LGTM. Suggested to add one test case. 
Not blocking this PR on that test suggestion.

BenHenning:
Only Bazel failure was RevisionCardActivityLocalTest which seems to be due to a network flake (a Maven dependency failed to download). I think we can safely ignore this and avoid needing to re-run Bazel tests before merging this PR in.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
BenHenning: Nice idea. Done.

BenHenning:
Thanks for the review! I added the suggest test change & verified that TopicFragmentTest is passing locally with Espresso.

Note I'll probably cancel the Bazel tests per my comment above & verifying tests passing locally. I'd prefer to get the whole chain in quickly since I'll be verifying that Bazel is passing at the end (after #3116 is merged).

PR #2992, opened 2021-03-23
Update compute_affected_tests.sh

BenHenning:
Update affected targets script to work on versions of git older than 2.22. See https://stackoverflow.com/a/6245587 for specifics.

I ran into this locally--I'm using git 2.7.4 and can't seem to update it, so having the more compatible approach seems ideal.


anandwana001 (approved):
LGTM

PR #3014, opened 2021-03-25
Enable D8 dex merging

BenHenning:
## Explanation
This enables D8 dex merging in Bazel builds since it seems the compatibility version doesn't seem to work with the latest SDK tools (31.0.1).

https://github.com/bazelbuild/bazel/issues/10287#issuecomment-584822816 has a bit more context on why these particular flags are necessary.

Note that the binary building successfully on CI should be a sufficient check to ensure nothing has regressed. I've verified locally that the build runs.


jcqli (approved):
LGTM

BenHenning:
Thanks!

PR #3017, opened 2021-03-25
Fix #2953: Add support for concept cards in revision cards

### LINKED ISSUE #2953: Concept cards don't open from revision cards
**Describe the bug**
Concept cards don't seem to open when attempted from concept cards.

**To Reproduce**
1. Open new fractions lesson in the alpha build
2. Navigate to revision tab
3. Open 'Comparing Fractions'
4. Tap ``convert them into fractions with the same denominator Open Concept Card``
5. Observe that nothing happens

**Expected behavior**
Concept cards should open when tapped from any place.

**Additional context**
This issue might be alpha-branch specific, and I didn't check if concept cards are working in explorations or questions (though questions will be disabled for alpha MR2).

BenHenning:
## Explanation
Fix #2953.

Adds support for opening concept cards from revision cards. This follows the traditional boilerplate for integration & testing for concept cards.

Note that this PR updates the second subtopic of the fractions topic to include an arbitrary link to a concept card to make it easier to access them, and to verify the integration. This, in turn, required a slight change to two existing tests that, for unknown reasons, were specifically verifying that no rich-text was included in the subtopic (which is now no longer true with the addition of the concept card).

Note also that RevisionCardFragmentTest was updated to conditionally use proto assets when running in Bazel and JSON when running in Gradle (to ensure the textproto changes are also tests, and that the revision card loading pipeline for protos is at least partially correct via tests).

The new tests do not validate the exit flow since that requires interacting with the concept card's toolbar, and there's not a clear way on how to do this in a Robolectric-friendly way. This unfortunate reality is also reflected for exploration & question concept card tests.

Note that RevisionCardFragmentTest has been verified as passing locally via Espresso.


BenHenning:
TODO: need to add tests & update some JSON lessons to support triggering this locally.

BenHenning:
This is now ready for review. @rt4914 PTAL.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -74,4 +88,10 @@ class RevisionCardFragmentPresenter @Inject constructor(
       oppiaLogger.createRevisionCardContext(topicId, subTopicId)
     )
   }
+
+  override fun onConceptCardLinkClicked(view: View, skillId: String) {
+    ConceptCardFragment
+      .newInstance(skillId)
+      .showNow(fragment.childFragmentManager, CONCEPT_CARD_DIALOG_FRAGMENT_TAG)
```
rt4914, line 95: The const `CONCEPT_CARD_DIALOG_FRAGMENT_TAG` is defined in `StatePlayerRecyclerViewAssembler`.
Considering that now this fragment is getting used outside of `player` too. I think we should shift this to `ConceptCardFragment`

rt4914 (approved):
Once suggestion otherwise LGTM

BenHenning:
WalkthroughFinalFragmentTest failed due to a Maven flake--I'll make sure it passes locally. TopicControllerTest seems to be legitimately failing since it's failing both on Bazel & Gradle. I will address it & verify it's passing before merging this.

Otherwise, I'll be prioritizing the non-Bazel tests per https://github.com/oppia/oppia-android/pull/2986#issuecomment-826033128.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
BenHenning: Nice catch, and good suggestion. Moved to the fragment's companion object & tested affected tests locally to ensure nothing broke.

BenHenning:
Ah. It appears that TopicControllerTest was just failing due to the concept card description text being updated. The local fix was pretty easy (use ``contains`` instead of ``isEqualTo`` since exact equality doesn't seem particularly important for this test).

Confirming this & other failing/affected tests from addressing the comment above are passing:

image

BenHenning:
I also verified that RevisionCardFragmentTest is passing locally via Espresso.

BenHenning:
Gradle test failed with an unrelated LogUploadWorkerTest flake.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
rt4914: Sounds good.

PR #3023, opened 2021-03-29
Revert "Enable D8 dex merging"

BenHenning:
Reverts oppia/oppia-android#3014

It turns out that my issue was actually due to upgrading my build tools version past 31.0 which we don't currently support. While switching to R8 is preferable, it doesn't currently work with desugared classes. This seems like a Bazel issue that needs to be addressed, but we can mitigate it for now by keeping D8 compat and ensuring we don't use newer Android build tools.


jcqli (approved):
Huh interesting, this LGTM

BenHenning:
Thanks!

PR #3035, opened 2021-04-01
[RunAllTests] Fix part of #1861: Enable GitHub Actions caching for Bazel CI workflows

BenHenning:
Fix part of #1861.

## Explanation
This introduces local Bazel caching support in GitHub Actions by leveraging GA's cache via [actions/cache](https://github.com/actions/cache). This approach is largely based on the solutions outlined [here](https://github.com/actions/cache/issues/239#issuecomment-606950711) and [here](https://github.com/actions/cache/issues/109#issuecomment-558771281).

This solution works because Bazel's hermetic & incremental design facilitates strong cache sharing even for completely different builds. The cache will continue to rapidly grow unbounded to maximize caching potential. This behaves similarly to the remote caching functionality we have support for via GCS except it uses GitHub's free resources, and works for forks.

One drawback is that GitHub Actions only allows 5GB of cache storage per app. Due to the unbounded growth factor, we have to forcibly clear the cache otherwise the cache will fail to upload and we'll be stuck using an increasingly out-of-date one. In cases where we detect the cache needs to be cleared, the all actions detecting that will run as if there isn't any cache (e.g. how they run now), and then jobs that run after should be faster since they'll benefit somewhat from the cache.

Since we aren't caching anything else substantial (only Gradle, and that will be going away in the future), this PR establishes a 4.5GB threshold (uncompressed) before forcibly clearing the cache. At current estimates, a full local build requires about a ~2GB uncompressed cache (~900MB compressed), so I'm guessing the 4.5GB threshold will lead to 2-3 force cache resets per day. Given the ~50% compression ratio we're seeing, we might be able to raise this up to 8GB-9GB before actually running into the 5GB limit, but we can tune this if we're seeing issues. The actions provide diagnostic output to track cache size & when the cache needs to be reset,

See the following analysis table for an idea on how actions times seem to improve due to this change:
| Action          | Caching             | Appx. Runtime          |
|-----------------|---------------------|------------------------|
| Build binary    | None                | ~14-15 mins            |
| Build binary    | Full                | ~1-2 mins*             |
| Full test suite | None                | ~9-12 mins (each test) |
| Full test suite | Cached built binary | ~6-9 mins (each test)                    |
| Full test suite | All tests**           | ~2-3 mins (each test)                    |

We saw the full test suite in the last case run in just under 2 hours (a 4x-8x improvement over the previously observed 8-16 hour runtime for the full test suite).

*I noticed one run where back-to-back builds triggered a later build to be a complete rebuild of the app despite having an up-to-date cache. Not quite sure why that happened; maybe something to do with timestamps overlapping since the two builds were happening simultaneously. It's a bit disappointing that the caching didn't seem to help much for the latter build, though it didn complete in under 12 minutes. We'll need to keep an eye out for similar cases in the future.

**Per the issue mentioned above, running the tests in parallel doesn't actually fully cache the results. Also, the cache selection seems a bit random at the moment (though it uses a LRU cache eviction policy), so there's not any guarantee yet that the caches will get better over time.

Finally, I wanted to note a few things:
- Per #1861 we still need to fix some of the caching issues with Android-specific Bazel actions; this will probably substantially improve incremental build performance
- This change alone may not be sufficient to facilitate desharding the Bazel tests, but I suspect a few more build time improvements (e.g. maybe the caching fix above, and our modularization work) will probably let us run 5-10 tests per shard instead of 1 which should result in substantially faster test runs
- There's still a lot more we can do to speed up Bazel CI beyond the above (e.g. only running tests for affected test changes between deltas, moving Android libraries to be kt_jvm_libraries, and more)
- Moving to a proper remote cache that's run through GitHub Actions (this was pitched as a potential option back in 2019, but no traction has been made quite yet)

This is meant to make the Bazel tests in CI a bit more tolerable, but probably won't make them completely pleasant quite y et.

### DIFF
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -45,8 +45,10 @@ jobs:
         id: cache
         with:
           path: ${{ env.CACHE_DIRECTORY }}
-          key: ${{ runner.os }}-${{ env.CACHE_DIRECTORY }}-bazel-${{ github.sha }}
+          key: ${{ runner.os }}-${{ env.CACHE_DIRECTORY }}-bazel-binary-${{ github.sha }}
           restore-keys: |
+            ${{ runner.os }}-${{ env.CACHE_DIRECTORY }}-bazel-binary-
+            ${{ runner.os }}-${{ env.CACHE_DIRECTORY }}-bazel-tests-
             ${{ runner.os }}-${{ env.CACHE_DIRECTORY }}-bazel-
 
       # This check is needed to ensure that Bazel's unbounded cache growth doesn't result in a
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -19,7 +19,7 @@ jobs:
         os: [ubuntu-18.04]
     env:
       ENABLE_CACHING: false
-      CACHE_DIRECTORY: $HOME/.bazel_cache
+      CACHE_DIRECTORY: ~/.bazel_cache
     steps:
       - uses: actions/checkout@v2
 
@@ -86,7 +86,11 @@ jobs:
 
       - name: Configure Bazel to use a local cache
         run: |
-          echo "build --disk_cache=$BAZEL_CACHE_DIR" >> $HOME/.bazelrc
+          # See https://stackoverflow.com/a/27485157 for reference.
+          EXPANDED_BAZEL_CACHE_PATH="${BAZEL_CACHE_DIR/#\~/$HOME}"
+          echo "Using $EXPANDED_BAZEL_CACHE_PATH as Bazel's cache path"
+          echo "build --disk_cache=$EXPANDED_BAZEL_CACHE_PATH" >> $HOME/.bazelrc
+        shell: bash
         env:
           BAZEL_CACHE_DIR: ${{ env.CACHE_DIRECTORY }}
 
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -154,5 +154,5 @@ jobs:
 
       - uses: actions/upload-artifact@v2
         with:
-          name: oppia-bazel-apk
+          name: oppia-bazel.apk
           path: /home/runner/work/oppia-android/oppia-android/oppia.apk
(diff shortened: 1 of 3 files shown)


BenHenning:
Hmm, one likelihood is running lots of tests in parallel will cause them to step over each other when uploading the cache (which may hurt repeat-running that same test). Running the tests in batches or many steps will help with this, but I suspect randomly running the test over longer periods of time (e.g. multiple PRs or days of runs) should provide an "eventual consistency" in cache hits.

BenHenning:
Hmm this can probably be done better by introducing target-specific caches that can be restored from first before restoring from a random cache. I expect that will substantially improve incremental performance at the expense of massively overloading our cache quota (e.g. each PR will probably individually exceed the quota). A midway might be to just qualify 'test' and 'binary' subcaches for now and see how that performs.

BenHenning:
Okay, the latest commit introduces priority-based bucketing for caching:
- Builds will use the test cache if there's no build-specific cache available
- Tests will use their own bucketed cache (based on 'app', 'util', 'testing', 'data', and 'domain'), then any test cache if theirs is unavailable, then the build cache
- Both will use any cache available if all others are missing (which will let them benefit from the previous cache created in earlier CI runs)

This will almost certainly increase the amount of cache evictions that GitHub does, but it should also improve cross-workflow performance within a single PR since the random cache selection doesn't seem to work as well.

BenHenning:
The whole test suite completed in just under 2 hours with almost no prior test results in the cache. :) That's a direct 4-8x improvement over the previous 8-16 hour runtime. Going to run it again as a true incremental build & re-enable the other tests to see if it runs any faster.

BenHenning:
Regarding the failures, ``LogUploadWorkerTest`` seems to be an unrelated flake: #2032. ``QuestionPlayerServiceTest`` is a known data test flake that should be fixed by #2916.

Looks like the second test run took 2 hours and 15 minutes. :)

BenHenning:
@anandwana001 PTAL. This is actually rather important since it should substantially improve Bazel CI performance for contributors, giving them faster feedback when they break Bazel tests & the build.

anandwana001 (approved):
LGTM, Thanks @BenHenning

BenHenning:
Thanks @anandwana001! Merging this now so that we can see if it helps with incoming PRs.

PR #3025, opened 2021-03-30
[RunAllTests] Fix #3018, #1675, #2931, #1600, and part of #2642: Enable data module in Bazel

BenHenning:
Fix #3018
Fix #1675
Fix #2931
Fix #1600
Fix part of #2642

## Explanation
This PR:
1. Enables the data module in Bazel by re-enabling Moshi (avoiding needing #1600 altogether--we can continue using Moshi in Bazel) using a similar solution as #2979
2. This adds the data module's tests to the Bazel BUILD graph (which required pulling one fix from #2916). Note that without #2916 these tests will be flaky, but given that they are not required yet this should be fine.
3. This fixes all of the Bazel tests being broken that was caused by adding the Retrofit interfaces & GAE classes as dependencies in Bazel without Moshi being enabled
4. It removes TODOs on #1675 (other than ApplicationModule, no existing tests actually needed a new dependency on NetworkModule so this was a trivial fix)
5. This introduces initial Bazel modularization in GAE for #2931
6. This adds '.aswb' to .gitignore for folks that are using Android Studio with Bazel for development (mainly me at this point)
7. Adds some debugging lines to compute_affected_tests.sh since it seems to keep hanging on my PRs when changing Bazel files; hopefully this will help us better dig into why. It needs to go into a PR where the issue has been observed since otherwise a separate PR that makes the change won't help reveal potential issues.
8. Updates to Robolectric 4.4 to work around the issue described here: https://github.com/oppia/oppia-android/pull/3025#issuecomment-810796701. This upgrade only affects Bazel builds, and conveniently lets us also migrate off of our custom fork of robolectric-bazel and use the shared one, instead.

For those interested, the underlying fix was explicitly declaring a Bazel ``java_plugin`` that points to Moshi's annotation processor. Using a combination of _direct_ dependencies & ``export_plugins`` from ``java_library`` we can explicitly trigger the annotation processor to run for certain libraries (I suspect that Dagger does something similar under the surface, but I haven't dug into it). In the Gradle world ``kapt`` seems to do this automatically for us (or maybe Gradle does). Note that this does *not* use kapt, but it's not actually needed since Moshi is a Java annotation processor.

PR #3036, opened 2021-04-01
Fix #2804 : Added label for OnboardingActivity

### LINKED ISSUE #2804: [A11y] Add label for OnboardingActivity
## Task
Introduce label for `OnboardingActivity` as `Onboarding Flow`.

## General Explanation

### Example

Arnold2381:
## Explanation
Added a label for Onboarding Activity in manifest and a test for the same.
Fixes #2804

## Checklist


app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
```diff
@@ -15,7 +15,6 @@ class OnboardingActivity : InjectableAppCompatActivity(), RouteToProfileListList
   companion object {
     fun createOnboardingActivity(context: Context): Intent {
       val intent = Intent(context, OnboardingActivity::class.java)
-      intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
```
MaskedCarrot, line 18: Correct me if I am wrong but I think this should not be removed.

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingActivityTest.kt
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  private fun createOnboardingActivity(): Intent {
+    return OnboardingActivity.createOnboardingActivity(
+      ApplicationProvider.getApplicationContext()
+    )
+  }
+
+  // TODO(#59): Figure out a way to reuse modules instead of needing to re-declare them.
+  // TODO(#1675): Add NetworkModule once data module is migrated off of Moshi.
```
MaskedCarrot: remove `TODO(#1675)` as per #3025

MaskedCarrot (changes requested):
@Arnold2381 PTAL at the following.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
Arnold2381: Ummmm yeah! I guess removing this will make a new instance of this activity again and again!
SO this should be not removed True! but if I do not remove this the test is failing, so I got to think of doing something else.
I think creating a new createOnboardingActivity_Test for label test will do the needful.
Any inputs to this? @MaskedCarrot

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingActivityTest.kt
Arnold2381: > remove `TODO(#1675)` as per #3025
@MaskedCarrot  I didn't get this tbh what do I actually need to do?

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
MaskedCarrot: i replacing this line with ` .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_NEW_TASK)` will be the solution to this problem.
adding @rt4914 here to verify that this is the correct way of doing this.

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingActivityTest.kt
MaskedCarrot: @Arnold2381 remove the second TODO, that is `TODO(#1675)`.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
Arnold2381: Yea ig too this can be a solution, will try this, also check if it works and commit once @rt4914 confirms.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
```diff
@@ -15,7 +15,7 @@ class OnboardingActivity : InjectableAppCompatActivity(), RouteToProfileListList
...
+      intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
MaskedCarrot: sorry i was not clear in my last comment. As the android docs mention [here](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_CLEAR_TOP) 

`FLAG_ACTIVITY_CLEAR_TOP` should be used in conjunction with `FLAG_ACTIVITY_NEW_TASK`. so I suggest changing this line to `.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_NEW_TASK)`

MaskedCarrot (changes requested):
@Arnold2381 PTAL

MaskedCarrot (approved):
LGTM, thanks

rt4914 (approved):
LGTM, thanks.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
BenHenning: Hmm this is confusing to me. @rt4914 why do we need to clear top for this activity in the first place? SplashActivity should be calling finish() so there shouldn't be anything to clear.

Similarly, starting a new task in the backstack doesn't seem correct here for the same reason.

app/src/main/res/values/strings.xml
```diff
@@ -54,6 +54,7 @@
   <string name="featured_questions">Featured Questions</string>
   <string name="frequently_asked_questions">Frequently Asked Questions</string>
   <string name="faq_activity_title">FAQs (Frequently Asked Questions)</string>
+  <string name="onboarding_activity_title">Onboarding Flow</string>
```
BenHenning: Not sure about this--it doesn't seem very user friendly @rt4914. 'Onboarding' isn't really a term we share with users, it's more of a development term. What about 'Oppia introduction' or something similar?

BenHenning (changes requested):
Thanks @Arnold2381! Had a couple of thoughts left as comments.

app/src/main/res/values/strings.xml
rt4914: `Oppia introduction` seems more correct.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingActivity.kt
rt4914: Actually even I was confused on this part. I am not sure why this was introduced in the first place. For current implementation I think we can remove it  completely.

Arnold2381: Should I remove this completely? addFlags

rt4914: Yes.

BenHenning (approved):
This looks good--thanks @Arnold2381!

PR #3040, opened 2021-04-04
Added score calculations 

TheRealJessicaLi:
## Explanation
Added score calculations (with testing) for practice sessions. Mastery calculations to be implemented in a separate PR.

## Checklist


BenHenning:
@TheRealJessicaLi looking at the CI results, it seems there might be some issues with building & running the domain questions tests. Do they build & run for you locally?

Also, I'll need to take a closer look at this tomorrow or Wendesday--apologies for the delay.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,96 @@
...
+      scorePerSkillMapping[skillId] = MutableFractionGrade(0.0, 0.0)
+    }
+  }
+
+  /** Calculate the user's overall score and score per skill for this practice session. */
+  private fun calculateScores() {
+    for (questionMetric in questionSessionMetrics) {
+      val totalHintsPenalty = questionMetric.numberOfHintsUsed * viewHintPenalty
+      val totalWrongAnswerPenalty =
+        (questionMetric.numberOfAnswersSubmitted - 1) * wrongAnswerPenalty
```
vinitamurthi: Do we need to ensure that this isn’t a negative value (I.e. numberOfAnswersSubmitted has to be greater than 0?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,96 @@
...
+      var questionScore = 0.0
+      if (!questionMetric.didViewSolution) {
+        questionScore =
+          maxOf(maxScorePerQuestion - totalHintsPenalty - totalWrongAnswerPenalty, questionScore)
+      }
+      this.totalScore.numerator += questionScore
+      for (linkedSkillId in questionMetric.question.linkedSkillIdsList) {
+        if (!scorePerSkillMapping.containsKey(linkedSkillId)) continue
+        scorePerSkillMapping[linkedSkillId]!!.numerator += questionScore
```
vinitamurthi: There could be cases where the score per skill mapping has a numerator and a denominator as 0. Are we handling that case anywhere? Ideally it’s best to remove those entries lower in the stack so that when we actually divide the numerator and denominator we do not need to handle that scenario

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,96 @@
...
+        scorePerSkillMapping[linkedSkillId]!!.denominator += 1
+      }
+    }
+  }
+
+  /** Compute the overall score as well as the score and mastery per skill. */
+  internal fun computeAll(): UserAssessmentPerformance {
+    calculateScores()
+    // TODO: set up mastery calculations
```
vinitamurthi: Can you create an issue for this and add the issue number here? I.e. TODO(#issuenum)....

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    var finalScorePerSkillMapping = mutableMapOf<String, FractionGrade>()
+    for (score in scorePerSkillMapping) {
+      finalScorePerSkillMapping[score.key] = FractionGrade.newBuilder().apply {
+        numerator = score.value.numerator
+        denominator = score.value.denominator
+      }.build()
+    }
+
+    var finalMasteryPerSkillMapping = mapOf<String, Double>()
```
vinitamurthi: Add todo to compute this as well

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
```diff
@@ -1019,6 +1028,368 @@ class QuestionAssessmentProgressControllerTest {
...
+      denominator = 3.0
+    }.build()
+
+    assertThat(userAssessmentPerformance.totalFractionScore).isEqualTo(grade)
+    assertThat(userAssessmentPerformance.fractionScorePerSkillMappingCount).isEqualTo(1)
+    assertThat(userAssessmentPerformance.getFractionScorePerSkillMappingOrThrow(TEST_SKILL_ID_2))
+      .isEqualTo(grade)
+  }
+
+  @Test
```
vinitamurthi, line 1121: Can you add a test where hints are viewed for multiple questions? So one hint and then the right answer , and repeat this for two questions 
Another one is to test two hints for a single question 
Otherwise I think the tests capture most use cases

vinitamurthi (changes requested):

BenHenning:
Will take a pass after the latest changes are addressed.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: I looked into QuestionAssessmentProgressController and it seems like we're only marking a question as completed once the user actually submits an answer for that question. Since we're not calculating scores until all the questions are completed, I don't think numberOfAnswersSubmitted would ever be 0.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Addressed! Thanks for catching that

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Done!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Done!

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
TheRealJessicaLi: Added tests

TheRealJessicaLi:
> Will take a pass after the latest changes are addressed.

Just addressed Vinita's comments!

BenHenning:
Sorry, will need to look at this tomorrow (but maybe Monday).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.oppia.android.app.model.FractionGrade
+import org.oppia.android.app.model.UserAssessmentPerformance
+import java.math.BigDecimal
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Private mutable class that stores a grade as a fraction.
+ */
+private class MutableFractionGrade(var numerator: BigDecimal, var denominator: BigDecimal)
```
BenHenning: At a high-level, why are we using BigDecimal? It seems like a pretty heavy weight solution, and I don't see the need here for perfect precision. What are we trying to mitigate with it?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,106 @@
...
+        scorePerSkillMapping[linkedSkillId]!!.numerator += questionScore
+        scorePerSkillMapping[linkedSkillId]!!.denominator += 1.toBigDecimal()
+      }
+    }
+  }
+
+  /** Compute the overall score as well as the score and mastery per skill. */
+  internal fun computeAll(): UserAssessmentPerformance {
+    calculateScores()
+    // TODO(#3067): create mastery calculations method
```
BenHenning: Nit (for consistency):

```suggestion
    // TODO(#3067): Create mastery calculations method.
```

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
```diff
@@ -21,4 +30,16 @@ class QuestionModule {
...
   @QuestionTrainingSeed
   fun provideQuestionTrainingSeed(oppiaClock: OppiaClock): Long = oppiaClock.getCurrentTimeMs()
+
+  @Provides
+  @ViewHintPenalty
+  fun provideViewHintPenalty(): Double = 0.1
+
+  @Provides
+  @WrongAnswerPenalty
+  fun provideWrongAnswerPenalty(): Double = 0.1
```
BenHenning: For this & above: it feels a bit like the unit is missing here. Maybe 'WrongAnswerScorePenalty' would be clearer?

model/src/main/proto/question.proto
```diff
@@ -82,14 +82,20 @@ message AnsweredQuestionOutcome {
   bool is_correct_answer = 2;
 }
 
+// A user's grade for a practice session represented as a fraction.
+message FractionGrade {
```
BenHenning, line 86: This is a bit confusing since fractions can't be a real value divided by a real value. For what reason do we want to keep a numerator and denominator here?

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
```diff
@@ -21,4 +30,16 @@ class QuestionModule {
   @Provides
   @QuestionTrainingSeed
   fun provideQuestionTrainingSeed(oppiaClock: OppiaClock): Long = oppiaClock.getCurrentTimeMs()
+
+  @Provides
+  @ViewHintPenalty
+  fun provideViewHintPenalty(): Double = 0.1
```
BenHenning: Out of curiosity, do these need to be represented as doubles? We could alternatively employ a point system wherein max score is 10 points and penalties are 1 point each. That might let us keep integers everywhere along the computation path which would offer perfect precision, too, until the final division. Would that work?

BenHenning (dismissed):
Thanks @TheRealJessicaLi! Sorry for the super long delay here. Took a first pass (mostly just considering some of the high-level decisions before doing a closer review).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,106 @@
...
+    for (score in scorePerSkillMapping) {
+      if (score.value.denominator != 0.toBigDecimal()) { // Exclude entries with denominator of 0
+        finalScorePerSkillMapping[score.key] = FractionGrade.newBuilder().apply {
+          numerator = score.value.numerator.toDouble()
+          denominator = score.value.denominator.toDouble()
+        }.build()
+      }
+    }
+
+    // TODO(#3067): set up finalMasteryPerSkillMapping
```
vinitamurthi:
```suggestion
    // TODO(#3067): Set up finalMasteryPerSkillMapping
```

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
vinitamurthi: That should work and it makes sense. I believe we went with a decimal system for web and so thats why we translated the same thing here

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
vinitamurthi: Maybe I missed this..won't we still have cases where both the numerator and denominator can be 0?

vinitamurthi (changes requested):

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: This was to avoid floating point precision errors when doing computations, but I just saw your comment about using integers so I'll try that instead

model/src/main/proto/question.proto
TheRealJessicaLi: I believe Chantel wants to display the score for each skill as a fraction (so we need both numerator and denominator), and Oppia web is calculating scores as real values instead of integers, so I did the same for consistency. Do we want to multiply everything by a factor of 10 so we have integers instead?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    // Set up the return values
+    var finalScore = FractionGrade.newBuilder().apply {
+      numerator = totalScore.numerator.toDouble()
+      denominator = totalScore.denominator.toDouble()
+    }.build()
+
+    var finalScorePerSkillMapping = mutableMapOf<String, FractionGrade>()
+    for (score in scorePerSkillMapping) {
+      if (score.value.denominator != 0.toBigDecimal()) { // Exclude entries with denominator of 0
```
TheRealJessicaLi, line 89: @vinitamurthi We're removing entries with 0 numerator/denominator in the computeAll() method here

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: We're removing any skills with a 0 numerator/denominator in the computeAll() method

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
TheRealJessicaLi: I fixed the implementation to use integers in the score calculation step under the hood (by multiplying everything by a factor of 10), but I'm returning doubles as the result (by dividing everything by 10 again) because Chantel said we may be displaying the scores in fraction form and I wanted to make sure the scores match the decimal system used for web in that case. Let me know what you think!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Done

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+private class MutableFractionGrade(var numerator: Int, var denominator: Int)
+
+/**
+ * Private class that computes the state of the user's performance at the end of a practice session.
+ * This class can exist across multiple training session instances, but calling code is responsible
+ * for ensuring it is properly reset.
+ */
+@Singleton
+internal
+class QuestionAssessmentCalculation @Inject constructor(
```
BenHenning: Nit: should be on the same line

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Initialize member fields. */
+  internal fun initialize(
+    skillIdList: List<String>,
+    questionSessionMetrics: List<QuestionSessionMetrics>
+  ) {
+    this.skillIdList = skillIdList
+    this.questionSessionMetrics = questionSessionMetrics
+    this.totalScore =
+      MutableFractionGrade(0, 0)
```
BenHenning: Nit: this can (& should) fit on one line.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+internal
+class QuestionAssessmentCalculation @Inject constructor(
+  @ViewHintScorePenalty private val viewHintPenalty: Int,
+  @WrongAnswerScorePenalty private val wrongAnswerPenalty: Int,
+  @MaxScorePerQuestion private val maxScorePerQuestion: Int
+) {
+  private lateinit var skillIdList: List<String>
+  private lateinit var questionSessionMetrics: List<QuestionSessionMetrics>
+  private lateinit var totalScore: MutableFractionGrade
+  private lateinit var scorePerSkillMapping: MutableMap<String, MutableFractionGrade>
```
BenHenning: I'm wondering how much of this class's state management could be simplified if it were a recreated object rather than being part of the Dagger graph. Consider instead creating a Factory class that's injected and that creates new instances of the calculation class. Then, you can pass in these values as part of the class's constructor and make them vals to have better immutability (which is a general design plus in the Java/Kotlin world). For a reference, take a look at UrlImageParser and PersistentCacheStore--both of those utilize injectable factories to create the underlying class.

I believe this also lets you get rid of the ``@Singleton`` qualifier, too, since no state needs to be stored in the Dagger graph then.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+
+  /** Compute the overall score as well as the score and mastery per skill. */
+  internal fun computeAll(): UserAssessmentPerformance {
+    calculateScores()
+    // TODO(#3067): Create mastery calculations method
+
+    // Set up the return values
+    var finalScore = FractionGrade.newBuilder().apply {
```
BenHenning: Here & elsewhere: prefer to use val as much as you can. We should almost never need vars. They're harder to make correct since their references can change at any time. For cases where we do need to use mutable classes (like a proto builder or a mutableList/Map), prefer to isolate it just to a single scope & store immutable copies that are shared across a class or across threads.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+      numerator = totalScore.numerator.toDouble() / 10
```
BenHenning: Why divide by 10? Is it because we try to have 10 questions in a practice session? Note that this might not always be the case (e.g. we might change that requirement, make it dynamic in the future, or have fewer than 10 due to insufficient questions to select from when generating the practice session--the last one is a potential issue that can happen today).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,105 @@
...
+            maxScorePerQuestion - totalHintsPenalty - totalWrongAnswerPenalty,
+            questionScore
+          )
+      }
+      this.totalScore.numerator += questionScore
+      this.totalScore.denominator += maxScorePerQuestion
+      for (linkedSkillId in questionMetric.question.linkedSkillIdsList) {
+        if (!scorePerSkillMapping.containsKey(linkedSkillId)) continue
+        scorePerSkillMapping[linkedSkillId]!!.numerator += questionScore
+        scorePerSkillMapping[linkedSkillId]!!.denominator += maxScorePerQuestion
```
BenHenning: This can be made safer by using a scoping function & avoiding the '!!':

```suggestion
        scorePerSkillMapping[linkedSkillId]?.let { mutableGrade ->
          mutableGrade.numerator += questionScore
          mutableGrade.denominator += maxScorePerQuestion
        }
```

Note that the current solution technically has a race condition since controllers can be called by multiple threads & it's assuming nothing changes between the containsKey and the retrieval (which is actually probably true if you initialize the map in the class's constructor per my factory comment above, but it's still good to address this part as well).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
BenHenning: Might it be worth filtering the metrics on those that have at least 1 submitted answer? That might be simple with Kotlin:

```for (questionMetric in questionSessionMetrics.filter { it.numberOfAnswersSubmitted > 0 }) {```

File: domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt [hunk 19]
```diff
@@ -0,0 +1,105 @@
...
+import javax.inject.Singleton
+
+/**
+ * Private mutable class that stores a grade as a fraction.
+ */
+private class MutableFractionGrade(var numerator: Int, var denominator: Int)
+
+/**
+ * Private class that computes the state of the user's performance at the end of a practice session.
+ * This class can exist across multiple training session instances, but calling code is responsible
```
BenHenning: This might change if you decide to use a factory per my comment below.

However, it might also be worth clarifying here that it's the responsibility of the caller to synchronize access to this class (that is, this class makes no effort to ensure thread-safe access). I think that's already consistent with how you're using it, and it provides a nice simplification to the implementation (since there's a lot of mutable state in the current class that would need to be synchronized).

model/src/main/proto/question.proto
BenHenning: I think that's a worthwhile simplification. I see the current implementation and I find it much easier to follow without the floating point bits.

That being said, a few thoughts:
1. If you're multiplying everything by '10', we should make that '10' a Dagger constant since it's now a magic value (e.g. something like 'max points per question' or reuse the existing constant)
2. 'numerator' and 'denominator' are still really unclear. Could we just use the same terms that are being used elsewhere? E.g. 'pointsReceived' & 'totalPointsAvailable', respectively?

BenHenning (commented):
Thanks @TheRealJessicaLi! Sorry, I missed the follow comments earlier. Leaving another batch of comments here. I want to take one more pass on the calculation class after you address these to see if I have any ways to further simplify it (I think there is some potential for that) before reviewing the full PR.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: I think my comment may have gotten lost, but I believe Chantel wants to display the score for each skill as a fraction, and Oppia web is calculating scores as fractions represented by real value numerator/denominators. I took the approach we had previously discussed and we're now using integers to represent the score numerators/denominators and penalties internally to avoid floating point error - I did this by multiplying the total score and penalties by a factor of 10. To stay consistent with Oppia web if Chantel chooses to display the numerator/denominator of each score, I'm dividing everything by a factor of 10 before it is returned.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: I don't think we'd want to filter those out because we would still want to include those questions in the total score. As an alternative, I could make sure the computed `totalWrongAnswerPenalty` is non-negative?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed! Thanks for the suggestion, I think this is definitely better

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Added

model/src/main/proto/question.proto
TheRealJessicaLi: Done, thanks!

TheRealJessicaLi:
@BenHenning @Vinita I addressed all the comments so far, but it seems like the QuestionAssessmentProgressControllerTest suite is failing on develop, and I'm unable to re-test my changes against the tests I wrote. Any chance you guys are aware of potential breaking changes that got merged into develop recently?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
BenHenning: Clarifying a point here: I don't think "this class can exist across..." is necessarily true anymore (since this class's state is no longer a singleton). If this situation does occur it's because of the calling code, not this class. I suggest removing it to avoid potential confusion.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+    fun create(
+      skillIdList: List<String>,
+      questionSessionMetrics: List<QuestionSessionMetrics>
+    ): QuestionAssessmentCalculation {
+      // Set up the data structures needed for computing the user's scores
+      val totalScore = MutableFractionGrade(0, 0)
+      val scorePerSkillMapping = mutableMapOf<String, MutableFractionGrade>()
+      for (skillId in skillIdList) {
+        scorePerSkillMapping[skillId] = MutableFractionGrade(0, 0)
+      }
```
BenHenning: This could be initialized in the class via ``by lazy`` constructs (which is a bit more idiomatic--I find it a bit concerning to be passing in mutable state since we ideally want to keep class's public APIs immutable where we can).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+import javax.inject.Inject
+import kotlin.math.max
+
+/**
+ * Private class that computes the state of the user's performance at the end of a practice session.
+ * This class can exist across multiple training session instances, but calling code is responsible
+ * for ensuring it is properly reset. This class is not thread-safe, so it is the calling code's
+ * responsibility to synchronize access to this class.
+ */
+internal class QuestionAssessmentCalculation constructor(
```
BenHenning:
```suggestion
internal class QuestionAssessmentCalculation private constructor(
```

Since this is only called by the factory class (and should be).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+          it.pointsReceived += questionScore
+          it.totalPointsAvailable += maxScorePerQuestion
+        }
+      }
+    }
+  }
+
+  /** Compute the overall score as well as the score and mastery per skill. */
+  internal fun computeAll(): UserAssessmentPerformance {
+    calculateScores()
```
BenHenning: A nit: I personally prefer listing public API methods first, and then helper methods in the order of appearance after the last public API function. This helps set consistent expectations that the top members of a class are that class's public API.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+      totalFractionScore = finalScore
+      putAllFractionScorePerSkillMapping(finalScorePerSkillMapping)
+      putAllMasteryPerSkillMapping(finalMasteryPerSkillMapping)
+    }.build()
+  }
+
+  /**
+   * Mutable class that stores a grade as a fraction.
+   */
+  class MutableFractionGrade(
```
BenHenning: Can this class be marked as internal?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+    return UserAssessmentPerformance.newBuilder().apply {
+      totalFractionScore = finalScore
+      putAllFractionScorePerSkillMapping(finalScorePerSkillMapping)
+      putAllMasteryPerSkillMapping(finalMasteryPerSkillMapping)
+    }.build()
+  }
+
+  /**
+   * Mutable class that stores a grade as a fraction.
+   */
```
BenHenning:
```suggestion
  /** Mutable class that stores a grade as a fraction. */
```

Nit: prefer fitting KDocs on one line if they fit (for compactness), though this is up to preference. :)

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    val finalScorePerSkillMapping = mutableMapOf<String, FractionGrade>()
+    for (score in scorePerSkillMapping) {
+      if (score.value.totalPointsAvailable != 0) { // Exclude entries with denominator of 0
+        finalScorePerSkillMapping[score.key] = FractionGrade.newBuilder().apply {
+          pointsReceived = score.value.pointsReceived.toDouble() / maxScorePerQuestion
+          totalPointsAvailable = score.value.totalPointsAvailable.toDouble() / maxScorePerQuestion
+        }.build()
+      }
+    }
```
BenHenning:
```suggestion
    val finalScorePerSkillMapping = skillPerScoreMapping.filter { (_, fractionGrade) ->
      fractionGrade.totalPointsAvailable != 0
    }.mapValues { (_, fractionGrade) ->
      FractionGrade.newBuilder().apply {
        pointsReceived = fractionGrade.pointsReceived.toDouble() / maxScorePerQuestion
        totalPointsAvailable = fractionGrade.totalPointsAvailable.toDouble() / maxScorePerQuestion
      }.build()
    }
```

This is more Kotlin-idiomatic, and avoids using mutable state at all. It can also be made more efficient if you use sequences, instead, but the iterable version is fine.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    // TODO(#3067): Set up finalMasteryPerSkillMapping
```
BenHenning: Maybe move TODO down to the newBuilder line below and remove the extra map? The default map state in proto3 is empty.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
BenHenning: Similarly, this can be done in a more Kotlin-idiomatic way (to avoid the mutable map):

```kotlin
skillIdList.associateWith { MutableFractionGrade() }
```

(I also suggest updating the constructor for MutableFractionGrade to default params to 0 to simplify things since it doesn't seem like we actually need to ever initialize with non-zero values).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,118 @@
...
+internal class QuestionAssessmentCalculation constructor(
+  private val viewHintPenalty: Int,
+  private val wrongAnswerPenalty: Int,
+  private val maxScorePerQuestion: Int,
+  private val questionSessionMetrics: List<QuestionSessionMetrics>,
+  private val totalScore: MutableFractionGrade,
+  private val scorePerSkillMapping: MutableMap<String, MutableFractionGrade>
+) {
+  /** Calculate the user's overall score and score per skill for this practice session. */
+  private fun calculateScores() {
```
BenHenning: Thinking about this more, how important is it to save the side effects of this function? I'm actually wondering if we can just compute the latest state without using mutable state like MutableFractionGrade since we only appear to be computing things at the end of the session.

You may notice a theme with my feedback: any places we can reduce mutable state will improve the safety of the system, and often reduce the complexity of the code.

BenHenning (commented):
Thanks @TheRealJessicaLi! Took another pass on QuestionAssessmentCalculation.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: I think avoiding mutable state is a really good point and something I'll definitely consider more! Thanks for all the feedback. However, I'm not sure there is a good way to do this without MutableFractionGrade. We're computing all the scores from the questionSessionMetrics list given to us - each item in the list holds the metrics for one question. Different questions can have overlapping tagged skills, so we need to keep track of the mutable scores for each skill as we process all of questionSessionMetrics. 

If we wanted to directly compute the latest state for each skill and store it in the proto object, I think we would have to filter the questions by their linked skills for every skill we're testing in the session and then do the calculations that are currently in calculateScores(). I think this would be less efficient and intuitive than just processing the metrics for each question once and copying the values into the final proto object, but maybe there is a better approach?

Let me know what you think!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Removed, thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Good point, I'll consider this next time too, thanks! Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Thanks for the suggestion! I'll keep this in mind for next time as well

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
BenHenning: I think there are generally two expected ways to approach this problem (unless I'm missing some fundamental requirements):
1. Each time a question is answered, determine what needs to be tracked & track that. Results are a filtered aggregate/combination of this data.
2. Each time a question is answered, we record that it was answered. When results are needed, we calculate everything in that moment.

You're right that mutable state is needed now, but I think the choices allow flexibility in storing the mutable state either at the stats level (option 1) or at the event level (option 2). Option 1 is more performant when aggregating since we do some work ahead of time, and option 2 is more performant when recording. We should pick whichever option reflects typical access patterns, and which simplifies the implementation.

I think the current approach is more of option 2 in aggregation (everything is happening in compute) but using the storage of option 1. I don't quite see why we need to store MutableFractionGrade when it's only being changed/used in the compute function. It seems a bit like it's memoization, but unless compute is being called a lot that probably isn't needed.

Now, stepping back a moment. Given that the class no longer is storing state in Dagger (& thus has nothing to reset across sessions--we can just create a new one), and it's intended that computeAll only be computed at the end of the session, this additional caching seems a lot less necessary. Avoiding it actually results in a lot of simplifications, I think. One alternative that follows this pattern might be something like (which also conveniently avoids mutable state):

```kotlin
fun computePerformance(questionSessionMetrics: List<QuestionSessionMetrics>): UserAssessmentPerformance {
  val allQuestionScores: List<Int> = questionSessionMetrics.map { it.computeQuestionScore() }
  val totalPointsScored = allQuestionScores.sum()
  val totalAvailablePoints = allQuestionScores.size * maxScorePerQuestion

  // Note to Jessica: this also avoids needing to pass in the list to the class's constructor since we no longer need to initialize the list ahead of time. I think that results in a simpler API even though it comes at the expense of calling computeQuestionScore multiple times (thought that *could* be memoized if we found it to have performance issues).
  val scoresPerSkill: Map<String, List<Int>> = questionSessionMetrics.flatMap { questionMetric ->
    // Convert to List<Pair<String, QuestionSessionMetrics>>>
    questionMetric.linkedSkillIdsList.map { skillId -> skillId to questionMetric}
  }.groupBy(
    // Covert to Map<String, List<QuestionSessionMetrics>> for each linked skill ID.
    { (skillId, _) -> skillId },
    valueTransform = { (_, questionMetric) -> questionMetric }
  ).mapValues { (_, questionMetrics) ->
    // Compute the question score for each question, converting type to Map<String, List<Int>>
    questionMetrics.map { it.computeQuestionScore() }
  }

  // Assemble & return the proto, plus any other needed bookkeeping. Use the same sum()/size() * max to compute the total scores/available for each question metric.
}

private fun UserAssessmentPerformance.computeQuestionScore(): Int = if (!didViewSolution) {
  val hintsPenalty = numberOfHintsUsed * viewHintPenalty
  val wrongAnswerPenalty = getAdjustedWrongAnswerCount() * wrongAnswerPenalty
  (maxScorePerQuestion - hintsPenalty - wrongAnswerPenalty).coerceAtLeast(0)
} else 0

private fun UserAssessmentPerformance.getAdjustedWrongAnswerCount(): Int =
  (UserAssessmentPerformance - 1).coerceAtLeast(0)
```

(this is not complete & could be further split up if desired). https://stackoverflow.com/a/43614691 was a big help for the grouping part since I couldn't quite figure out how to map against the linked skill IDs.

WDYT?

BenHenning (commented):
I spent a bit longer on trying to simplify the calculation class that I intended. LMK what you think about the suggestion and then I will take a pass on the other parts of the PR.

TheRealJessicaLi:
> I spent a bit longer on trying to simplify the calculation class that I intended. LMK what you think about the suggestion and then I will take a pass on the other parts of the PR.

I took a look at the suggestions you made and I agree that it makes things a lot cleaner and safer - I've updated QuestionAssessmentCalculation to reflect the changes. Thanks for taking the time to flesh out the pattern! I'll try to use something similar when I'm implementing mastery calculations to avoid mutable state as much as possible.

BenHenning:
Sorry, been really busy with non-work items this week. Will need to take a look later in the week.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,99 @@
...
+      // TODO(#3067): Set up finalMasteryPerSkillMapping
+    }.build()
+  }
+
+  private fun computeTotalScore(): FractionGrade =
+    FractionGrade.newBuilder().apply {
+      val allQuestionScores: List<Int> = questionSessionMetrics.map { it.computeQuestionScore() }
+      pointsReceived = allQuestionScores.sum().toDouble() / internalScoreMultiplyFactor
+      totalPointsAvailable =
+        allQuestionScores.size.toDouble() * maxScorePerQuestion / internalScoreMultiplyFactor
```
vinitamurthi, line 34: Do we need to divide both the numerator and denominator of the FractionGrade by the factor? Wont the fraction stay equivalent regardless?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,99 @@
...
+      questionMetrics.map { it.computeQuestionScore() }
+    }.filter { (_, scores) ->
+      // Eliminate any skills with no questions this session.
+      !scores.isNullOrEmpty()
+    }.mapValues { (_, scores) ->
+      // Convert to Map<String, FractionGrade> for each linked skill ID.
+      FractionGrade.newBuilder().apply {
+        pointsReceived = scores.sum().toDouble() / internalScoreMultiplyFactor
+        totalPointsAvailable =
+          scores.size.toDouble() * maxScorePerQuestion / internalScoreMultiplyFactor
```
vinitamurthi, line 58: Same comment here

vinitamurthi (approved):
LGTM!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
BenHenning: It might also be simpler to just document the multiplier in the other constants rather than factoring it in with math (since it's unlikely we need to tweak that without changing the other constants).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
```diff
@@ -0,0 +1,99 @@
...
+
+  private fun computeTotalScore(): FractionGrade =
+    FractionGrade.newBuilder().apply {
+      val allQuestionScores: List<Int> = questionSessionMetrics.map { it.computeQuestionScore() }
+      pointsReceived = allQuestionScores.sum().toDouble() / internalScoreMultiplyFactor
+      totalPointsAvailable =
+        allQuestionScores.size.toDouble() * maxScorePerQuestion / internalScoreMultiplyFactor
+    }.build()
+
+  private fun QuestionSessionMetrics.computeQuestionScore(): Int = if (!didViewSolution) {
```
BenHenning: Nit: suggest moving these below since they're separate from the private helpers of this class (they sort of belong to both this class & QuestionSessionMetrics as extension functions).

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -56,6 +57,7 @@ class QuestionAssessmentProgressController @Inject constructor(
 
   private val progress = QuestionAssessmentProgress()
   private val progressLock = ReentrantLock()
+  @Inject internal lateinit var scoreCalculatorFactory: QuestionAssessmentCalculation.Factory
```
BenHenning, line 60: Does this need to be internal or can it be private? Prefer minimizing access to values when possible.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -319,6 +321,31 @@ class QuestionAssessmentProgressController @Inject constructor(
...
+   *
+   * This method should only be called at the end of a practice session, after all the questions
+   * have been completed.
+   */
+  fun calculateScores(skillIdList: List<String>): DataProvider<UserAssessmentPerformance> =
+    progressLock.withLock {
+      return dataProviders.createInMemoryDataProviderAsync(
+        "user_assessment_performance"
+      ) {
+        (this::retrieveUserAssessmentPerformanceAsync)(skillIdList)
```
BenHenning: Why can't this just be ``retrieveUserAssessmentPerformanceAsync(skillIdList)``?

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
```diff
@@ -11,6 +11,18 @@ annotation class QuestionCountPerTrainingSession
 @Qualifier
 annotation class QuestionTrainingSeed
 
+@Qualifier
+annotation class ViewHintScorePenalty
```
BenHenning, line 18: Each of these qualifiers should have documentation to explain what they correspond to/how to use them. E.g.:

```kotlin
/** Qualifier corresponding to the penalty users receive for each hint viewed in a practice session (an application-level injectable int). */
@Qualifier annotation class ViewHintScorePenalty
```

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
```diff
@@ -994,6 +1003,515 @@ class QuestionAssessmentProgressControllerTest {
...
+    verify(
+      mockCurrentQuestionLiveDataObserver,
+      atLeastOnce()
+    ).onChanged(currentQuestionResultCaptor.capture())
+    val ephemeralQuestion = currentQuestionResultCaptor.value.getOrThrow()
+    assertThat(ephemeralQuestion.ephemeralState.pendingState.wrongAnswerCount)
+      .isEqualTo(1)
+    val hintAndSolution = ephemeralQuestion.ephemeralState.state.interaction.solution
+    assertThat(hintAndSolution.correctAnswer.correctAnswer)
+      .isEqualTo("3.0")
```
BenHenning: Will these fit on the same line? If so, prefer that (in general, prefer to fill up as much line space as possible before line wrapping).

Ditto elsewhere.

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
```diff
@@ -994,6 +1003,515 @@ class QuestionAssessmentProgressControllerTest {
     assertThat(updatedState.ephemeralState.state.interaction.solution.solutionIsRevealed).isTrue()
   }
 
+  @Test
+  fun testRevealedSolution_forWrongAnswer_returnScore2OutOf3() {
```
BenHenning, line 1007: Here & below: I suggest factoring high-level actions & their sanity checks into different helper methods to only keep the high-level details relevant to the behavior being tested in the method body. For example, with refactoring this test could instead be:

```kotlin
fun testRevealedSolution_forWrongAnswer_returnScore2OutOf3() {
  // Arrange
  setUpTestApplicationWithSeed(questionSeed = 0)
  subscribeToCurrentQuestionToAllowSessionToLoad()
  startTrainingSession(TEST_SKILL_ID_LIST_2)
  submitWrongAnswerForQuestion2() // NOTE TO JESSICA: this could also perform verification as a sanity check
  submitSolutionIsRevealed()
  submitCorrectAnswerForQuestion2()

  // Act
  val grade = getExpectedSuccessfulGrade()

  // Assert
  assertThat(grade.numerator).isEqualTo(2)
  assertThat(grade.denominator).isEqualTo(3)
}
```

Notice how this can help focus the test body just to the details that the reader needs to see & hide the boilerplate to make it work. This is an easier test to verify, and reduces a lot of the overall code that we need to maintain for all of the tests being added.

I suggest making this or a similar clean up for all new tests here.

model/src/main/proto/question.proto
BenHenning: Following up on this--can these values be ints now that everything is being scaled? You could always perform the final decimal calculation in the UI and keep things ints at the domain layer to simplify management.

BenHenning (commented):
Thanks @TheRealJessicaLi. Took a full pass on the PR & have a few comments. Note that I'm resetting my previous review status to unblock this PR since I'll be largely unavailable over the next couple weeks.

rt4914:
@TheRealJessicaLi Assign me this PR once you have addressed Ben's and Vinita's comments so that I can cross verify them in my full PR pass. Thanks.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: @Vinita @BenHenning My reasoning for dividing both the numerator and denominator by the multiplier was to make sure the final FractionGrade's numerator and denominator are consistent with the values that we would have gotten on Oppia web. Since Chantel said she may want to display the score as a fraction (e.g. "you got x / y points"), I thought it would be better to do that here so the app layer doesn't have to simplify it. 

I made a separate constant for this multiplier because it's possible that the maxScorePerQuestionConstant can be different from the multiplier if the scoring system were to the change in the future. As an example right now, for mastery calculations, maxMasteryGainPerQuestion is 10, but we multiplied all the constants by a factor of 100 to maintain integer values. In the case that they are different (like for masteries), we would need to use both values to calculate the final scores/masteries. I thought keeping this multiplier separate would be good for consistency and also leave room for future changes so the math still checks out. 

Let me know what you think, thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: Fixed!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
TheRealJessicaLi: When I make the field private, Dagger complains and says "error: Dagger does not support injection into private fields", so that's why I made it internal. Do you know any workarounds for this?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentCalculation.kt
TheRealJessicaLi: See comment above!

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
TheRealJessicaLi: Fixed, thanks!

domain/src/main/java/org/oppia/android/domain/question/QuestionConstantsProvider.kt
TheRealJessicaLi: Added, thanks!

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
TheRealJessicaLi: Fixed

model/src/main/proto/question.proto
TheRealJessicaLi: I kept it as a double because I think Chantel may want to display the scores as a fraction in the UI (e.g. "you got score x / y"), so in that case I thought it would be better to calculate the final values to be consistent with the way Oppia web does it  (which may not be integers). To do the calculation, we would also need access to the multiplier constant, which is in the domain module. I thought it made more sense to complete the calculations here before sending off the final score (ready for display) to the app module. Let me know what you think!

domain/src/test/java/org/oppia/android/domain/question/QuestionAssessmentProgressControllerTest.kt
TheRealJessicaLi: Refactored, thanks!

BenHenning:
I won't be able to take a look at this until I return on or after the 24th of May but I'm happy to defer to @vinitamurthi for domain layer design decisions.

vinitamurthi:
I will take a look once again to ensure Ben's comments are addressed, but otherwise the approach LGTM. @rt4914 can you please take a look? We are blocked on your codeowner review

vinitamurthi:
LGTM from my side, @rt4914 PTAL

rt4914 (approved):
LGTM for codeowners. Thanks @TheRealJessicaLi

vinitamurthi (approved):

PR #3043, opened 2021-04-06
Fix #2780: Call checkIfLoadingIsPossible() after inject() in LessonThumbnailImageView

### LINKED ISSUE #2780: Thumbnail not visible
**Describe the bug**
Ratios and Proportional Reasoning Thumbnail is not visible on landscape

**To Reproduce**
1. Go to Home Screen
2. Landscape Orientation
3. Scroll to the bottom
4. last card, the `Ratios and Proportional Reasoning` image is not visible

**Expected behavior**
The image should be visible no matter which orientation

**Screenshots**
https://photos.app.goo.gl/6ZYnTvtBNv2UPAb79

ArpitShukIa:
## Explanation
Fixes #2780: The problem was that few dependencies weren't available at the time of calling `checkIfLoadingIsPossible()` so `loadLessonThumbnail()` was not getting called for that particular image. I simply added a call to `checkIfLoadingIsPossible()` after Dagger inject() call and that fixed the issue.

#### Robolectric:
image

#### Expresso:
image

#### Robolectric:
image

#### Expresso:
image

### Before:
https://user-images.githubusercontent.com/51705072/113724562-199e2e80-9710-11eb-9a84-410350363a7f.mp4

### After:
https://user-images.githubusercontent.com/51705072/113724749-494d3680-9710-11eb-960a-58b53e323ac4.mp4

## Checklist


ArpitShukIa:
@anandwana001 PTAL.

ArpitShukIa:
The lint test has failed because of the semi-colon that I used in the empty while loop. That was necessary to mark that loop as empty.
Should I use { } instead?

anandwana001:
I guess `{};` will prevent the failure. Also, please complete all the installation process.

ArpitShukIa:
@anandwana001 Sorry, I don't understand what you mean by installation process.

FareesHussain:
> @anandwana001 Sorry, I don't understand what you mean by installation process.

To setup pre-push hooks to prevent the ktlint check errors

Ref: 4th point in  https://github.com/oppia/oppia-android/wiki#installation

anandwana001:
Deferring to Rajat for inital pass, as he is the codeowner.

rt4914 (changes requested):
@ArpitShukIa Adding an infinite loop doesn't seem like a correct approach to me at the same time I understand this issue is not that easy too.

@BenHenning Do you have any comments on this approach?

app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt
```diff
@@ -69,17 +73,26 @@ class LessonThumbnailImageView @JvmOverloads constructor(
...
       ::imageLoader.isInitialized &&
       ::logger.isInitialized
+
+  private fun checkIfLoadingIsPossible() {
+    if (::entityId.isInitialized &&
+      ::entityType.isInitialized &&
+      ::lessonThumbnail.isInitialized
     ) {
-      loadLessonThumbnail()
+      GlobalScope.launch(Dispatchers.Default) {
```
BenHenning: We should never use dispatchers other than the ones we centrally manage, otherwise it's easy to introduce performance regressions, and it will be much harder to reliably test changes since unsynchronized dispatchers can introduce flakes.

Introducing an infinite loop is also potentially verify problematic, as is the post back to the main thread. The former could cause performance issues on the user's device if it never resolves since it'll be burning 100% CPU usage for that thread, and the post back to the main thread could introduce crashes if the view is being torn down right when the push occurs (which is why we only use DataProviders + LiveData for these situations).

Per the issue described [here](https://github.com/oppia/oppia-android/issues/2780#issuecomment-796791569), could we add a call to ``checkIfLoadingIsPossible`` after inject() is called? Unless I'm missing something, that should resolve the race when databinding calls everything before injection. Or is there something else that makes this problematic?

BenHenning (changes requested):
Hi @ArpitShukIa. Left one comment--PTAL when you can.

app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt
ArpitShukIa: > Per the issue described here, could we add a call to checkIfLoadingIsPossible after inject() is called?

@BenHenning This worked. Thanks.

Should I close this draft PR and create a new one OR revert commits in this and mark it Ready for review?

BenHenning: I think recycling this PR is fine/preferred. No need to revert commits--just undo the changes & add the new ones that work + tests to verify that it works (the latter is actually really important--you should always make sure to have tests that fail without your fix and pass with it when introducing a bug fix).

It's also better to keep the same PR since that retains history. We expect code to evolve during a code review--no problem with that. :)

ArpitShukIa: Okay. Will do that.
Also, can you plz tell what should be the test? 
Do I need to iterate over all items in the recycler view and check if the thumbnail has been loaded in the view or not?
Or is there some better way to test this fix?

BenHenning: You'll probably want to force the situation in order to verify that this is correct. That would be introducing a LessonThumbnailImageViewTest that creates a new LessonThumbnailImageView, then calls onAttachedToWindow/the binding methods in different orders for different test cases & then verify that the thumbnail was correctly initialized in the image view. The Dagger injection setup might be a bit tricky--let me know if you run into any issues there.

If you're unsure which test cases to add, I suggest just starting with one, get that passing, and then ask for additional feedback.

ArpitShukIa: I am having some trouble figuring this out. The `onAttachedToWindow()` function finds the fragment corresponding to current view for dagger injection. Will I have to launch the activity first? And if I launch the activity, how to replace the original view with this test view?

BenHenning:
For later: please add a TODO in the new LessonThumbnailImageViewTest for #3098 & repurpose that issue to be "add missing test cases".

app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt
BenHenning: @ArpitShukIa just to clarify, is your question in the context of the new ``LessonThumbnailImageViewTest``? If so, you'll need both a test activity & test fragment where the test fragment can point to a view that contains a ``LessonThumbnailImageView``. This isn't much different how we set up real activities/fragments elsewhere in the codebase, or how we use ``LessonThumbnailImageView`` in real layouts (e.g. ``topic_revision_summary_view.xml``).

ArpitShukIa:
@BenHenning I have added the tests and updated the description as well to include the test results screenshots.
Also, I don't understand why the checks failed, the tests are passing in my local machine.
PTAL

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
```diff
@@ -0,0 +1,174 @@
...
+  @get:Rule
+  val activityTestRule = ActivityTestRule(
+    LessonThumbnailImageViewTestActivity::class.java,
+    /* initialTouchMode= */ true,
+    /* launchActivity= */ false
+  )
+
+  @Test
+  fun callDataBindingFunctions_thenCallInject_thumbnailIsLoadedCorrectly() {
+    LessonThumbnailImageView.injectAtStart = false
```
BenHenning: We shouldn't punch holes into the production class for testing purposes. Could we, instead, add the view at runtime (not through layouts), and call the bindable adapter functions before & after adding the view for each case? I think adding the view will automatically trigger injection due to the view lifecycle, and manually calling the databinding methods helps us avoid the race condition we see in production cases. Would that work?

BenHenning (commented):
Thanks @ArpitShukIa. I think this is definitely on the right track. Left one main comment for the test.

Regarding the failure, LessonThumbnailImageViewTest seems to be using wrong imports (such as for TestDispatcherModule and RobolectricModule). These were moved in a recent PR--I'm surprised that it's not failing for you locally. Did you sync your branch with the latest develop & check the test again? I would expect them to fail.

Please follow up once the new comment is addressed & the test is passing, or if you run into any issues/need clarifications.

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
ArpitShukIa: Done

ArpitShukIa:
@BenHenning Looks like I made a mistake. I merged develop into my branch, but now the diff is showing changes for all recent commits in develop branch.
Is there a way to fix this, or should I create a fresh PR?

BenHenning:
@ArpitShukIa can you try following the instructions listed here: https://github.com/oppia/oppia-android/wiki/Fork-and-Clone-Oppia-Android#updating-your-local-repository. I think you'll need to:
1. Merge in the latest develop from oppia/oppia-android into your fork's version of develop
2. Pull your fork's version of develop to your local copy
3. Merge your local copy of develop into your  branch
4. Push your branch

This should properly fix the base of your branch.

ArpitShukIa:
@BenHenning PTAL

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestActivity.kt
```diff
@@ -0,0 +1,18 @@
+package org.oppia.android.app.testing
+
+import android.os.Bundle
+import org.oppia.android.R
+import org.oppia.android.app.activity.InjectableAppCompatActivity
+
+class LessonThumbnailImageViewTestActivity : InjectableAppCompatActivity() {
```
BenHenning, line 9: Please add a brief KDoc here explaining what this is (even though it's sort of self-explanatory), e.g.:

```
/** Test-only activity for verifying behaviors of LessonThumbnailImageView. */
```

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestFragment.kt
```diff
@@ -0,0 +1,21 @@
+package org.oppia.android.app.testing
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import org.oppia.android.app.fragment.InjectableFragment
+import org.oppia.android.databinding.LessonThumbnailImageViewTestFragmentBinding
+
+class LessonThumbnailImageViewTestFragment : InjectableFragment() {
```
BenHenning, line 12: Ditto for this fragment (should have a KDoc).

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestFragment.kt
```diff
@@ -0,0 +1,21 @@
...
+
+  override fun onCreateView(
+    inflater: LayoutInflater,
+    container: ViewGroup?,
+    savedInstanceState: Bundle?
+  ): View {
+    val binding = LessonThumbnailImageViewTestFragmentBinding
+      .inflate(inflater, container, false)
```
BenHenning: Nit: please include an argument comment for the ``false`` value, e.g. ``/* attachToRoot= */ false`` to provide context to readers.

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools">
+
+  <LinearLayout
+    android:id="@+id/lesson_thumbnail_image_view_holder"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="horizontal"
+    tools:ignore="UselessLeaf" />
```
BenHenning: Is this needed?

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools">
+
+  <LinearLayout
```
BenHenning: Is a linear layout actually needed or could we just use a FrameLayout here?

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    override fun getApplicationInjector(): ApplicationInjector = component
+  }
+
+  /** Provides test dependencies (including a mock for [ImageLoader] to capture its operations). */
+  @Module
+  class TestModule {
+    @Provides
+    @Singleton
+    fun provideMockImageLoader() = Mockito.mock(ImageLoader::class.java)
```
BenHenning: I actually suggest using the changes from #2981: https://github.com/oppia/oppia-android/pull/2981/files#diff-90aff075f37a890b5a50c2e1bd541f4f357fd29185d65c2495a4ff22648f7a1bR68. I'll be merging that later tonight my time so it should be available on develop pretty soon. You can avoid mocking ImageLoader & instead use ``TestImageLoaderModule``. It makes things more consistent across the codebase.

BenHenning (changes requested):
Thanks @ArpitShukIa! I had a few comments, but I think the overall testing approach looks really good.

BenHenning:
@rt4914 & @anandwana001 you could probably take a look now, too. The solution seems final.

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestActivity.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestFragment.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestFragment.kt
ArpitShukIa: Done

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
ArpitShukIa: Removed it.

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
ArpitShukIa: Done. Used FrameLayout.

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestActivity.kt
```diff
@@ -0,0 +1,19 @@
+package org.oppia.android.app.testing
+
+import android.os.Bundle
+import org.oppia.android.R
+import org.oppia.android.app.activity.InjectableAppCompatActivity
+
+/** Test-only activity for verifying behaviors of LessonThumbnailImageView. */
```
anandwana001:
```kt
import org.oppia.android.app.customview.LessonThumbnailImageView

/** Test-only activity for verifying behaviors of [LessonThumbnailImageView]. */
```

This way we can directly jump to LessonThumbnailImageView.

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
+
+  <FrameLayout
+    android:id="@+id/lesson_thumbnail_image_view_holder"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"/>
```
anandwana001:
```suggestion
    android:layout_height="match_parent" />
```

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
```
anandwana001: we can skip using layout, if no data binding work required

anandwana001 (changes requested):
LGTM, nit changes suggested

app/src/main/java/org/oppia/android/app/testing/LessonThumbnailImageViewTestActivity.kt
ArpitShukIa: Done

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
ArpitShukIa: Done

app/src/main/res/layout/lesson_thumbnail_image_view_test_fragment.xml
ArpitShukIa: Done

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
ArpitShukIa: I used the TestGlideImageLoader and the withDrawable() Matcher. 
Robolectric tests are working fine (i.e. failing before fix and passing after fix) but the Expresso tests keep on running indefinitely. 
I have pushed the updated test file. PTAL

app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt
BenHenning: The latest solution looks really solid!

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
BenHenning: Ah, looks like you didn't even need #2981.

I suspect you'll need to add a block like this one to get Glide set up correctly on Espresso:

https://github.com/oppia/oppia-android/blob/9782e2a526a55763d09786f821e739fac171e77a/app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt#L179-L190

BenHenning (commented):
Thanks @ArpitShukIa! My only comment is with regards to the Espresso test hanging. PTAL.

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
ArpitShukIa: Just a small doubt. Are we using Glide here? I have used the TestImageLoaderModule for the test instead of GlideImageLoaderModule. And the loadDrawable() function of TestGlideImageLoader doesn't seem to be using Glide.

https://github.com/oppia/oppia-android/blob/9782e2a526a55763d09786f821e739fac171e77a/utility/src/main/java/org/oppia/android/util/parser/TestGlideImageLoader.kt#L48-L62

app/src/main/AndroidManifest.xml
```diff
@@ -188,6 +188,7 @@
     <activity
       android:name=".app.walkthrough.WalkthroughActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
+    <activity android:name=".app.testing.LessonThumbnailImageViewTestActivity"/>
```
rt4914: Add space before `/>`

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import org.oppia.android.util.gcsresource.GcsResourceModule
+import org.oppia.android.util.logging.LoggerModule
+import org.oppia.android.util.logging.firebase.FirebaseLogUploaderModule
+import org.oppia.android.util.parser.HtmlParserEntityTypeModule
+import org.oppia.android.util.parser.ImageParsingModule
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+import javax.inject.Singleton
+
+@RunWith(AndroidJUnit4::class)
```
rt4914, line 64: Add a Kdoc comment something like
`/** Tests for [LessonThumbnailImageViewTestAcivitiy]. */`

rt4914 (changes requested):
I don't know the reason but tests are not running for me. Can anyone else confirm if tests work or not?
Screenshot 2021-03-17 at 5 22 51 PM

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
ArpitShukIa: Done

app/src/main/AndroidManifest.xml
ArpitShukIa: Done

ArpitShukIa:
> I don't know the reason but tests are not running for me. Can anyone else confirm if tests work or not?

Not sure why that's happening. Tests are passing on my device.
Did you run tests on my branch? Are they passing for the develop branch code? 
Are all of the tests failing or some specific ones?

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
BenHenning: Yes, the test image loader still proxies back to Glide itself for loading (we intentionally don't mock Glide out if we can help it since it's otherwise a bit difficult to make sure our integration with Glide is working correctly).

BenHenning:
@ArpitShukIa I believe you have a few errors to correct in CI:

```
/home/runner/work/oppia-android/oppia-android/app/build/tmp/kapt3/stubs/debugUnitTest/org/oppia/android/app/customview/LessonThumbnailImageViewTest.java:73: error: [Dagger/MissingBinding] @org.oppia.android.app.topic.EnablePracticeTab java.lang.Boolean cannot be provided without an @Provides-annotated method.
```

I suggest making sure the new test builds & passes locally before sending this back to reviewers. This is probably just an issue from syncing to the latest develop changes (I suggest looking at other tests to see how they fill in these bindings).

Also, can you confirm whether the new test is passing on Espresso? That + the green CI are my main blockers before approving the PR since it otherwise LGTM.

anandwana001 (approved):
LGTM, as per my earlier comment.

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
ArpitShukIa: I added the code block you linked. The problem still persists. Expresso tests keep on running indefinitely.
Should I push the changes?

ArpitShukIa:
> Also, can you confirm whether the new test is passing on Espresso? That + the green CI are my main blockers before approving the PR since it otherwise LGTM.

All the checks have passed but the Expresso problem is still there. The tests keep on running indefinitely.

https://user-images.githubusercontent.com/51705072/116355653-c6ebfa00-a817-11eb-954c-d0bab5239525.mov

BenHenning:
@ArpitShukIa we might need to make this a Robolectric-only test. I'm not entirely sure why it doesn't work on Espresso, but given that the original regression triggers a test failure in Robolectric I think it being Robolectric-only is fine. Can you move it to be in the same directory structure except ``test/...`` instead of ``sharedTest/...``?

BenHenning:
Though thinking on this more, it either indicates a real issue that Robolectric isn't covering, or it's not liking how we're creating the view.

@ArpitShukIa you should also step-debug the Espresso test if you can to try and narrow down where the test is starting to hang. I think those findings should be identified before we move the test (though that's a valid option if we can't solve the Espresso issue).

ArpitShukIa:
@BenHenning The expresso tests are passing now :)
 I just had to make the assertions outside the activityTestRule.runOnUiThread {...} block.

BenHenning:
Ah, that's really subtle @ArpitShukIa. Nice find. Could you actually file an issue to the thread of: we should introduce a static analysis check or somehow trigger test failures in these situations to avoid the test hanging? It might require us to wrap Espresso's API, but I think that'd be worth it to avoid people running into this in the future. This seems tricky to debug/track down.

BenHenning (approved):
Given that the tests are now passing for both Robolectric & Espresso, this LGTM. Thanks @ArpitShukIa!

BenHenning:
@rt4914 sending this over to you since I think only your review is needed now.

ArpitShukIa:
> Could you actually file an issue to the thread of: we should introduce a static analysis check or somehow trigger test failures in these situations to avoid the test hanging?

@BenHenning I am a little confused about the description of the issue. Could you instead file the issue plz?

The problem was that `activityTestRule.runOnUiThread {...}` block runs on UIThread and `onView` waits for the UIThread to idle. Calling `onView()` inside that block results in a kind of deadlock and the test execution gets stuck.

rt4914 (approved):
LGTM, thanks.

BenHenning:
@ArpitShukIa we should have a test or check to ensure that such cases fail immediately with a good error message for developers to fix the situation rather than running into the deadlock (the deadlock is exactly what we want to avoid since it's difficult to understand what's going wrong). Static analysis or specific failure tests are a good way to prevent these sorts of situations from happening.

Edit: I went ahead and filed #3150 to make sure this context is captured.

PR #3064, opened 2021-04-09
[RunAllTests] Attempt to fix GitHub actions cache flake

BenHenning:
## Explanation
Flakes like [this](https://github.com/oppia/oppia-android/runs/2300787189?check_suite_focus=true) are now being observed after caching was introduced in #3035:

```
Error: Unable to process file command 'env' successfully.
Error: Invalid environment variable format 'util'
```

The variable format change probably won't make a difference, it's just a "correctness" thing per what I saw others doing. I'm hoping that using a bash shell will help, but I'm not confident. It's weird because it seems like echo is outputting the wrong data.

This fix is just experimental--I haven't actually found anyone hitting this problem yet.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -86,7 +86,8 @@ jobs:
           echo "Test target: $TEST_TARGET"
           TEST_CATEGORY=$(echo "$TEST_TARGET" | grep -oP 'org/oppia/android/(.+?)/' | cut -f 4 -d "/")
           echo "Test category: $TEST_CATEGORY"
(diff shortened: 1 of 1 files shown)


BenHenning:
Reopening to kick-off full Bazel test run to see if the issue can repro (chances are we'll need to wait for a few runs after merging this PR to be sure it actually fixed the problem).

BenHenning:
Aha! This has nothing to do with GitHub. It's the same issue as https://github.com/oppia/oppia-android/issues/2691#issuecomment-816322355 just manifesting itself differently. Closing this since it's not actually relevant.

PR #3065, opened 2021-04-09
Fixes #2824: [A11y] Add label for RecentlyPlayedActivity

### LINKED ISSUE #2824: [A11y] Add label for RecentlyPlayedActivity [Blocked on #3095]
## Task
Introduce label for `RecentlyPlayedActivity` as either `Stories for you` or `Recently played stories`. 

## General Explanation

### Example

rt4914:
## Explanation

Fixes #2824

For `RecentlyPlayedActivity` the label can be either `Stories for you` or `Recently Played Stories`.
This activity can be visited via two screens: (a.) Click on `View All` in `HomeActivity` and (b.) Click on `View All` in `ProfileProgressActivity`.

The main problem is that we need to send the string in advance via intent as we need it for screen readers. To do this I have created `RecentlyPlayedTitleEnum ` enum which will send the correct title. But this was not working and even I found one bug with Toolbar too in current develop. The toolbar title fluctuates. Check this below gif and focus on toolbar.

bug1

Earlier we were using toolbar in fragment and we were fetching the content from domain layer and based on the content received we were setting the toolbar. Now because in the new implementation we already have the title we can shift this toolbar to activity so that the flakiness is removed and the screen readers can work properly.

## Screenreader output on this branch

https://user-images.githubusercontent.com/9396084/114113154-a22cf280-98fb-11eb-82ea-6cb5dac1eff8.mp4

## Note to Reviewer
In this draft PR I am just confirming the approach and therefore I done following things which will be fixed in final PR:
1. Removed all other layout files related to `recently_played` so that I don't have to worry about them.
2. While calling the intent for `RecentlyPlayed` I am passing the hard-coded/fixed enum value if the approach is finalised this will be sent correctly.
3. I am not updated the test cases which might be failing.

## Checklist

- [ ] The PR does not contain any unnecessary auto-generated code from Android Studio.


rt4914:
@BenHenning PTAL at approach.

BenHenning:

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedActivity.kt
```diff
@@ -21,18 +21,29 @@ class RecentlyPlayedActivity : InjectableAppCompatActivity(), RouteToExploration
...
     /** Returns a new [Intent] to route to [RecentlyPlayedActivity]. */
-    fun createRecentlyPlayedActivityIntent(context: Context, internalProfileId: Int): Intent {
+    fun createRecentlyPlayedActivityIntent(
+      context: Context,
+      internalProfileId: Int,
+      recentlyPlayedTitleEnum: RecentlyPlayedTitleEnum
+    ): Intent {
       val intent = Intent(context, RecentlyPlayedActivity::class.java)
       intent.putExtra(RECENTLY_PLAYED_ACTIVITY_INTERNAL_PROFILE_ID_KEY, internalProfileId)
+      intent.putExtra(RECENTLY_PLAYED_ACTIVITY_TITLE_ENUM_KEY, recentlyPlayedTitleEnum)
```
BenHenning: This seems reasonable to me, but I think we want to do a push toward using protos more aggressively for sending arguments & intent extras. Maybe just go ahead and define an intent extra proto for this activity & use that, instead? It means a less-clean enum but you could still create an extension function for ``getTitleFromStringRes`` to keep dependent code clean.

app/src/main/res/layout-land/recently_played_fragment.xml
```diff
@@ -1,58 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<layout xmlns:android="http://schemas.android.com/apk/res/android"
-  xmlns:app="http://schemas.android.com/apk/res-auto">
-
-  <androidx.constraintlayout.widget.ConstraintLayout
```
BenHenning, line 5: Did we ever need landscape-specific specialization for this layout or did you just noticed it wasn't needed?

app/src/main/res/layout/recently_played_activity.xml
```diff
@@ -1,7 +1,55 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+
+      <View
+        android:layout_width="match_parent"
+        android:layout_height="6dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/recently_played_app_bar_layout" />
```
BenHenning: Is it value to have constraints when within a layout being constrained? Do we eve need these?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedTitleEnum.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/** Represents different titles for RecentlyPlayedActivity used by screen readers. */
+enum class RecentlyPlayedTitleEnum(@StringRes private var title: Int) {
+  RECENTLY_PLAYED_STORIES(R.string.recently_played_stories),
+  STORIES_FOR_YOU(R.string.stories_for_you);
+
+  /** Returns the string corresponding to this error's string resources, or null if there is none. */
+  fun getTitleFromStringRes(context: Context): String? {
+    return title.let(context::getString)
+  }
```
BenHenning:
```suggestion
  fun getTitleFromStringRes(context: Context): String? = context.getString(title)
```

The scope function way seems less readable to me, but YMMV. I also changed this to an optional single-assignment syntax.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedTitleEnum.kt
```diff
@@ -0,0 +1,16 @@
+package org.oppia.android.app.home.recentlyplayed
+
+import android.content.Context
+import androidx.annotation.StringRes
+import org.oppia.android.R
+
+/** Represents different titles for RecentlyPlayedActivity used by screen readers. */
+enum class RecentlyPlayedTitleEnum(@StringRes private var title: Int) {
+  RECENTLY_PLAYED_STORIES(R.string.recently_played_stories),
```
BenHenning: Why isn't this using recently_played_activity_title?

BenHenning (commented):
Thanks @rt4914. Left some thoughts, but the overall approach seems good.

BenHenning:
Also just to check: is the bug in the first video with or without the intent approach? I wouldn't expect the flicker to happen with your current implementation--is that the case?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedTitleEnum.kt
rt4914: Correct. Used `recently_played_activity_title `

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedTitleEnum.kt
rt4914: Updated as per suggestion.

app/src/main/res/layout/recently_played_activity.xml
rt4914: These constraints are incorrect. Removed now.

app/src/main/res/layout-land/recently_played_fragment.xml
rt4914: Actually we do need other xml files too but if this approach gets approved the xml files will change a lot and therefore I thought to remove them directly for now, finish the implementation, then again add those files back with all the changes. That way i won't have to make changes to various versions of xml files again and again.

rt4914:
> Also just to check: is the bug in the first video with or without the intent approach? I wouldn't expect the flicker to happen with your current implementation--is that the case?

The bug is without this approach, in current develop. With current implementation is it fixed.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedActivity.kt
rt4914: I have tried to use protos. I am not sure if its correct so do let me know WDYT.

1. `DestinationScreen` will helps to identify which activity to open and also with what intent extra values.
2. For example for this implementation, `RecentlyPlayedActivityIntentExtras` is a value inside `DestinationScreen` which contains information about the destination activity as well as its intent extras.
3. In `activity` package I have created `ActivityRouter` which will be injected to all activities / fragments from where we want to start the next activity. In this we will call `routeToScreen` with correct `DestinationScreen` and then this will decide which activity should be started.
4. One added code clarity point: With this approach we can change `RecentlyPlayedActivityIntentExtras` proto anytime it will be very easy to make changes to code. For example in current develop as I had to send an enum I had to introduced a new key and change the `createRecentlyPlayedActivityIntent` function by introducing a new argument which in turn resulted in changes in all the instances where that function was getting called. But with this new approach these changes will be comparatively less.

BenHenning: Ah, I was just thinking that you could pass a RecentlyPlayedActivityIntentExtras proto here & proceed similarly. The router approach seems much nicer. Do you want to do that change as part of this PR or separately?

BenHenning:
@rt4914 I think the approach seems fine, though I do suggest going with protos instead (one way or another with regards to introducing a router or not--I'll defer to you on how/when you want to introduce that). I think the overall solution is actually quite nice, and am happy to see the bug get fixed.

app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.oppia.android.app.model.DestinationScreen
+import org.oppia.android.app.model.RecentlyPlayedActivityIntentExtras
+import javax.inject.Inject
+
+/** Central utility to manage routing to different activities. */
+class ActivityRouter @Inject constructor(private val activity: AppCompatActivity) {
+
+  /** Checks the value of [DestinationScreen] and routes to different activities accordingly. */
+  fun routeToScreen(destinationScreen: DestinationScreen) {
+    if (destinationScreen.destinationScreenCase ==
```
BenHenning, line 14: Should this be a ``when`` statement,  instead? Maybe also use 'return' to force it to be exhaustive.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedActivity.kt
```diff
@@ -17,22 +18,27 @@ class RecentlyPlayedActivity : InjectableAppCompatActivity(), RouteToExploration
...
 
     /** Returns a new [Intent] to route to [RecentlyPlayedActivity]. */
-    fun createRecentlyPlayedActivityIntent(context: Context, internalProfileId: Int): Intent {
+    fun createRecentlyPlayedActivityIntent(
+      context: Context,
+      recentlyPlayedActivityIntentExtras: RecentlyPlayedActivityIntentExtras
+    ): Intent {
       val intent = Intent(context, RecentlyPlayedActivity::class.java)
-      intent.putExtra(RECENTLY_PLAYED_ACTIVITY_INTERNAL_PROFILE_ID_KEY, internalProfileId)
+      intent.putExtra(
```
BenHenning, line 38: Prefer using this utility instead: https://github.com/oppia/oppia-android/blob/develop/utility/src/main/java/org/oppia/android/util/extensions/BundleExtensions.kt#L13.

model/src/main/proto/route.proto
```diff
@@ -0,0 +1,36 @@
...
+
+// Corresponds to intent extras of destination activity.
+message DestinationScreen {
+  oneof destination_screen {
+    RecentlyPlayedActivityIntentExtras recently_played_activity_intent_extras = 1;
+  }
+}
+
+// Intent extras for RecentlyPlayedActivity.
+message RecentlyPlayedActivityIntentExtras{
```
BenHenning, line 18: Nit: space before ``{``

PR #3063, opened 2021-04-09
Revert "Fix #1824: Lowfi story activity tablet"

rt4914:
Reverts oppia/oppia-android#2993

PR #2993, opened 2021-03-23
Fix #1824: Lowfi story activity tablet

rt4914:
## Explanation

Fixes #1824 : Lowfi UI fixes in Story Activity Tablet

### Mocks
https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/64ef596d-f0ee-48fe-afc2-2ae40367e382/
https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/d70a3ade-826a-4ac6-95e6-907a24d7b896/

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3066, opened 2021-04-09
Fix #1824: Lowfi story activity tablet

### LINKED ISSUE #1824: Lowfi: Tablet: StoryActivity Landscape + Portrait
This issue is tracking the full low-fi work of `StoryActivity` Tablet (Landscape + Portrait)

Mock: https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/64ef596d-f0ee-48fe-afc2-2ae40367e382/
https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/d70a3ade-826a-4ac6-95e6-907a24d7b896/

Major work items:
- [ ] Vertical dash line with progress
- [x] Inactive card - lock/blur effect.
- [ ] RTL Supportive
**Note : Please test UI for RTL support, prior to merging this fix.**

rt4914:
## Explanation

Fixes #1824: Lowfi story activity tablet

This PR was approved and merged in #2993 but one of Bazel checks was failing and therefore reverted back in #3063 

## Checklist


BenHenning:
Ah. I think I know why the Bazel build is failing for ImageViewBindingAdaptersTest @rt4914. We perform a genrule for all utilities that need their R imports to be changed (due to the source files currently depending on Gradle's packaging layout), and this results in a rename of the file. In this case, the build graph is actually using ImageViewBindingAdapters_updates not ImageViewBindingAdapters, and because it's a utility file and not a class the symbol is wrong when referencing it.

Could we maybe use a databinding-enabled layout to reference the adapters by their adapter name (e.g. ``android:src`` for ``setImageDrawable``) for each of the different test cases & test them that way? This is a more correct utilization of the utility, anyway, since we shouldn't be calling it directly. It should hopefully be straightforward to do this in Gradle.

Once that's working, I can take an attempt on trying to make this work for Bazel as well (which will require additional work since at the moment we don't support databinding in tests, but I think this will be possible since we'll need to introduce new test-only layouts to make the above work).

Please let me know if you don't think this could work.

rt4914:
> Ah. I think I know why the Bazel build is failing for ImageViewBindingAdaptersTest @rt4914. We perform a genrule for all utilities that need their R imports to be changed (due to the source files currently depending on Gradle's packaging layout), and this results in a rename of the file. In this case, the build graph is actually using ImageViewBindingAdapters_updates not ImageViewBindingAdapters, and because it's a utility file and not a class the symbol is wrong when referencing it.
> 
> Could we maybe use a databinding-enabled layout to reference the adapters by their adapter name (e.g. `android:src` for `setImageDrawable`) for each of the different test cases & test them that way? This is a more correct utilization of the utility, anyway, since we shouldn't be calling it directly. It should hopefully be straightforward to do this in Gradle.
> 
> Once that's working, I can take an attempt on trying to make this work for Bazel as well (which will require additional work since at the moment we don't support databinding in tests, but I think this will be possible since we'll need to introduce new test-only layouts to make the above work).
> 
> Please let me know if you don't think this could work.

@BenHenning I understand the cause of the error.
To understand the solution correctly: I should create a separate activity which will be used for testing right?

BenHenning:
> > Ah. I think I know why the Bazel build is failing for ImageViewBindingAdaptersTest @rt4914. We perform a genrule for all utilities that need their R imports to be changed (due to the source files currently depending on Gradle's packaging layout), and this results in a rename of the file. In this case, the build graph is actually using ImageViewBindingAdapters_updates not ImageViewBindingAdapters, and because it's a utility file and not a class the symbol is wrong when referencing it.
> > Could we maybe use a databinding-enabled layout to reference the adapters by their adapter name (e.g. `android:src` for `setImageDrawable`) for each of the different test cases & test them that way? This is a more correct utilization of the utility, anyway, since we shouldn't be calling it directly. It should hopefully be straightforward to do this in Gradle.
> > Once that's working, I can take an attempt on trying to make this work for Bazel as well (which will require additional work since at the moment we don't support databinding in tests, but I think this will be possible since we'll need to introduce new test-only layouts to make the above work).
> > Please let me know if you don't think this could work.
> 
> @BenHenning I understand the cause of the error.
> To understand the solution correctly: I should create a separate activity which will be used for testing right?

That might be needed. I think the important thing is setting up the separate layout that references the adapters. If that needs a new test activity, then we should also introduce that.

PR #2929, opened 2021-03-16
Fixes #2628: Reading Text Size[A11y]

Arjupta:
## Explanation
Fixes #2628 based on the mock provided [here](https://github.com/oppia/oppia-android/issues/2628#issuecomment-778538267)

|New Layout|Accessibility Scan|
|:-:|:-:|
|111349733-68275280-86a7-11eb-959f-accf17d8ef4f|Screenshot_1616252207|

Tests passing- 

Untitled

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3114, opened 2021-04-22
Fix #3103: Segregate OptionsFragment tests

### LINKED ISSUE #3103: Test Segregation in OptionsFragmentTest
Currently in `OptionsFragmentTest` tests are working across multiple activities but thats not correct.
Each file should contain tests only related to that file and should not check functionalities of other activities.

Segregate tests of OptionsFragmentTest into 
- OptionsFragmentTest
- AudioLanguageFragmentTest
- AppLanguageFragmentTest

Also we should add more tests to make sure that we are able to hit 100% code coverage for these files which means that if we combine all tests than they should have traversed/used each line written for the functionality.

Sparsh1212:
## Explanation

Fixes #3103: Tests Segregation in OptionsFragmentTest

Espresso Test results snapshots:

image
image
image
image
image
image
image
image

## Checklist


Sparsh1212:
@rt4914 
I have completed the Part-A of the issue which involves:
1. Write a test in OptionsFragmentTest to check that the correct default Language is displayed in the recycler view of the Options file. (Repeat this a total of three times to check for App Language, Audio Language, and ReadingTextSize)
2. Repeat the above three tests for a configuration change.
3. Write a test for clicking an item in the Options Recycler View and check if it is sending the correct intent based on the default Language (Repeat this three times to check for App Language, Audio Language, and ReadingTextSize)
4. Also, repeat the above three tests (Point number three) for a configuration change.

Will proceed to Part-B of the issue after getting an initial pass on Part-A.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -206,118 +207,289 @@ class OptionsFragmentTest {
...
+            "English"
+          ),
+          hasComponent(AudioLanguageActivity::class.java.name)
+        )
+      )
+    }
+  }
+
+  @Test
+  fun openOptionsActivity_configChange_selectAudioLanguage_showAudioLanguageActivitySuccessfully() {
```
rt4914: Rename to `openOptionsActivity_configChange_clickAudioLanguage_opensAudioLanguageActivitySuccessfully`

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -206,118 +207,289 @@ class OptionsFragmentTest {
...
+            "English"
+          ),
+          hasComponent(AppLanguageActivity::class.java.name)
+        )
+      )
+    }
+  }
+
+  @Test
+  fun openOptionsActivity_selectAudioLanguage_showAudioLanguageActivitySuccessfully() {
```
rt4914: Instead of select use `click` when clicking on items in OptionsFragment

rt4914: Also use `open` instead of `show` when new activity is opened.

rt4914 (changes requested):
The tests are failing on Roboelectric. 
Each test should pass on Roboelectric as well as Espresso. Also make sure that CI checks also pass.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

Sparsh1212:
> The tests are failing on Roboelectric.
> Each test should pass on Roboelectric as well as Espresso. Also make sure that CI checks also pass.

Apologies! will remember to check each of them before making a commit.
Update: All the checks and tests are now passing. I have updated the PR with the requested changes. PTAL.

rt4914:
@Sparsh1212 I will review this tomorrow but meanwhile you can move to part B too.

Sparsh1212:
@rt4914 
Here's the update about the Part-B work that I have initiated:
- All the previous tests in `AppLanguageFragmentTest` are removed and these four new tests are added keeping in mind the ideal practice (i.e not to launch the other Activity and only use the Test File Activity): 
  1. `testAppLanguage_selectedLanguageIsEnglish()`
  2. `testAppLanguage_changeConfiguration_selectedLanguageIsEnglish()`
  3. `testAppLanguage_changeLanguageToFrench_selectedLanguageIsFrench()`
  4. `testAppLanguage_changeLanguageToFrench_changeConfiguration_selectedLanguageIsFrench() `
 
- I also tried adding this test:
```kt
   @Test 
   fun testApp_Language_pressBackBtn_SendsBackCorrectInfo(){
      launch<AppLanguageActivity>(createAppLanguageActivityIntent("English")).use {
          pressBack()
          intended(hasExtra(MESSAGE_APP_LANGUAGE_ARGUMENT_KEY, "English"))
    }
   }
```
But it is failing in Roboelectric with the error that 0 intends matched.
It looks like the Back press is getting ignored totally.
I also tried many other things to click the back button or to run the `onBackPress` function but nothing seems to work. Can you please take a look at this and guide me on what I am doing wrong here or if this approach isn't correct then how can we achieve checking that it is sending back the correct info on the back press through another approach.

- I have not made any changes to `ReadingTextSizeFragmentTest` because currently there is a PR (#2929) that will propose some changes in this file, so I think it will be better to wait to rewrite this file until those changes are merged.

- Will make all the similar changes for `AudioLanguageFragmentTest` once I get a pass on the above changes and the getting back correct info on the back press problem is resolved.  

Thanks!

rt4914:
> ```kotlin
> @Test 
>    fun testApp_Language_pressBackBtn_SendsBackCorrectInfo(){
>       launch<AppLanguageActivity>(createAppLanguageActivityIntent("English")).use {
>           testCoroutineDispatchers.runCurrent()
>           pressBack()
>           intended(hasExtra(MESSAGE_APP_LANGUAGE_ARGUMENT_KEY, "English"))
>     }
>    }
> ```

@Sparsh1212 Have you checked ShadowActivity? https://stackoverflow.com/a/8991026

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -239,37 +251,62 @@ class OptionsFragmentTest {
...
+  @Test
+  fun testOptionsFragment_configChange_appLanguageIsDisplayed() {
+    launch<OptionsActivity>(
+      createOptionActivityIntent(
+        internalProfileId = 0,
+        isFromNavigationDrawer = true
+      )
+    ).use {
+      testCoroutineDispatchers.runCurrent()
+      onView(isRoot()).perform(orientationLandscape())
```
anandwana001:
```suggestion
      onView(isRoot()).perform(orientationLandscape())
      testCoroutineDispatchers.runCurrent()
```
For stabilizing  after changing orientation

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -278,72 +315,202 @@ class OptionsFragmentTest {
...
       )
+    }
+  }
 
-      it.onActivity { activity ->
-        activity.startActivityForResult(
-          createAppLanguageActivityIntent(summaryValue = "English"),
-          REQUEST_CODE_APP_LANGUAGE
+  @Test
+  fun openOptionsActivity_clickReadingTextSize_opensReadingTextSizeActivitySuccessfully() {
```
anandwana001:
```suggestion
  fun openOptionsActivity_clickReadingTextSize_opensReadingTextSizeActivity() {
```

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -83,6 +84,13 @@ import org.robolectric.annotation.LooperMode
 import javax.inject.Inject
 import javax.inject.Singleton
 
+private const val KEY_READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE =
+  "READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE"
+private const val APP_LANGUAGE_PREFERENCE_SUMMARY_VALUE_EXTRA_KEY =
+  "AppLanguageActivity.app_language_preference_summary_value"
+private const val KEY_AUDIO_LANGUAGE_PREFERENCE_SUMMARY_VALUE =
```
anandwana001: can we use these original items only?
For example check `AdministratorControlsActivity` at `getIntentKey` which is used in `testNavDrawer_withAdminProfile_adminMenu_opensAdminControlsActivity`

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -278,72 +315,202 @@ class OptionsFragmentTest {
...
+          ),
+          hasComponent(ReadingTextSizeActivity::class.java.name)
+        )
+      )
+    }
+  }
 
-      getInstrumentation().waitForMonitorWithTimeout(activityMonitor, 3)
+  @Test
+  fun openOptionsActivity_configChange_clickTextSize_opensReadingTextSizeActivitySuccessfully() {
```
anandwana001:
```suggestion
  fun openOptionsActivity_configChange_clickTextSize_opensReadingTextSizeActivity() {
```

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
```diff
@@ -97,30 +96,46 @@ class AppLanguageFragmentTest {
     setUpTestApplicationComponent()
     Intents.init()
```
anandwana001:
```suggestion
```

I think we don't need `Intents.init()` and `Intents.release()` as there is no test which testing anything related to intents.

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
```diff
@@ -97,30 +96,46 @@ class AppLanguageFragmentTest {
...
   @Test
-  fun testAppLanguage_changeAppLanguageToFrench_changeConfiguration_selectedLanguageIsFrench() {
+  fun testAppLanguage_selectedLanguageIsEnglish() {
+    launch<AppLanguageActivity>(createAppLanguageActivityIntent("English")).use {
+      checkSelectedLanguage(ENGLISH)
+    }
+  }
+
+  @Test
+  fun testAppLanguage_changeConfiguration_selectedLanguageIsEnglish() {
```
anandwana001:
```suggestion
  fun testAppLanguage_configChange_selectedLanguageIsEnglish() {
```

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
```diff
@@ -97,30 +96,46 @@ class AppLanguageFragmentTest {
...
     launch<AppLanguageActivity>(createAppLanguageActivityIntent("English")).use {
-      selectLanguage(FRENCH)
       rotateToLandscape()
```
anandwana001, line 112: Let's use the similar function in the other file as well `app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt`

anandwana001 (changes requested):
regarding the `testApp_Language_pressBackBtn_SendsBackCorrectInfo `, we cannot use ShadowActivity, this will make this test robolectric only. 

The reason, I am seeing here why the test fails is:

```kt
override fun onBackPressed() {
    val message = appLanguageActivityPresenter.getLanguageSelected()
    val intent = Intent()
    intent.putExtra(MESSAGE_APP_LANGUAGE_ARGUMENT_KEY, message)
    setResult(REQUEST_CODE_APP_LANGUAGE, intent)
    finish()
  }
```

on back press, the activity gets finish, so test is basically trying to test after activity gets finished. 
I don't think this test will pass on espresso also. This test seems like to come under e2e test. 

Also, add all the espresso test result screenshots in PR description.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: There is no `getIntentKey` function present in AppLangaugeActivity and AudioLanguageActivity, therefore I have to again declare the constants here. Shall I add `getIntentKey` function to AppLanguageActivity in the companion object or go by the current approach? Please let me know.

Sparsh1212:
> regarding the `testApp_Language_pressBackBtn_SendsBackCorrectInfo `, we cannot use ShadowActivity, this will make this test robolectric only.
> 
> The reason, I am seeing here why the test fails is:
> 
> ```kotlin
> override fun onBackPressed() {
>     val message = appLanguageActivityPresenter.getLanguageSelected()
>     val intent = Intent()
>     intent.putExtra(MESSAGE_APP_LANGUAGE_ARGUMENT_KEY, message)
>     setResult(REQUEST_CODE_APP_LANGUAGE, intent)
>     finish()
>   }
> ```
> 
> on back press, the activity gets finish, so test is basically trying to test after activity gets finished.
> I don't think this test will pass on espresso also. This test seems like to come under e2e test.
> 
> Also, add all the espresso test result screenshots in PR description.

Thanks, will update and make all the suggested changes.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
anandwana001: introducing `getIntentKey`, will be helpful. @rt4914 could you please confirm?

rt4914: Actually we should shift keys to `companion object` of `OptionsFragment` and from there we can use them wherever we want using `OptionsFragment. KEY_AUDIO_LANGUAGE_PREFERENCE_SUMMARY_VALUE `

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
Sparsh1212: Done.

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
Sparsh1212: Done.

app/src/test/java/org/oppia/android/app/testing/options/AppLanguageFragmentTest.kt
Sparsh1212: Done.

Sparsh1212:
@rt4914 @anandwana001 I have made all the suggested changes. PTAL.

app/src/main/java/org/oppia/android/app/options/OptionsFragment.kt
```diff
@@ -25,6 +25,13 @@ class OptionsFragment : InjectableFragment() {
   lateinit var optionsFragmentPresenter: OptionsFragmentPresenter
 
   companion object {
+    const val KEY_READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE =
+      "READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE"
+    const val APP_LANGUAGE_PREFERENCE_SUMMARY_VALUE_EXTRA_KEY =
+      "AppLanguageActivity.app_language_preference_summary_value"
+    const val KEY_AUDIO_LANGUAGE_PREFERENCE_SUMMARY_VALUE =
```
rt4914: If you are introducing these value then please remove the previous consts mentioned at the top of the kotlin file.
Or alternate can be to not introduce these consts and use the values mentioned at top of kotlin file in your test file.

rt4914 (changes requested):
@Sparsh1212 Suggested change.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -207,70 +208,97 @@ class OptionsFragmentTest {
   }
 
   @Test
-  fun testOptionsFragment_readingTextSize_testOnActivityResult() {
+  fun testOptionsFragment_readingTextSizeIsDisplayed() {
```
anandwana001:
```suggestion
  fun testOptionsFragment_defaultReadingTextSizeIsDisplayed() {
```

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -207,70 +208,97 @@ class OptionsFragmentTest {
...
       )
+    }
+  }
 
-      it.onActivity { activity ->
-        activity.startActivityForResult(
-          createAudioLanguageActivityIntent(summaryValue = "Hindi"),
-          REQUEST_CODE_AUDIO_LANGUAGE
+  @Test
+  fun testOptionsFragment_configChange_appLanguageIsDisplayed() {
```
anandwana001:
```suggestion
  fun testOptionsFragment_configChange_defaultAppLanguageIsDisplayed() {
```

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -279,72 +307,203 @@ class OptionsFragmentTest {
...
       ).check(
-        matches(withText("French"))
+        matches(withText("English"))
       )
     }
   }
 
   @Test
-  fun testOptionsFragment_appLanguage_testOnActivityResult() {
+  fun testOptionsFragment_configChange_audioLanguageIsDisplayed() {
```
anandwana001:
```suggestion
  fun testOptionsFragment_configChange_defaultAudioLanguageIsDisplayed() {
```

anandwana001: ditto for other test names.

anandwana001 (changes requested):
1. Add Espresso Test result screenshot of all the test files that exists in this PR. 
2. In the tablet, we use Master-Detail design, so I think we also need to add a test for tablet-specific. 
```kt
  @Test
  @Config(qualifiers = "sw600dp")
fun testAppLanguage_clickAppLanguage_changeAppLanguage_checkOptionsFragmentIsUpdatedCorrectly()
```
and other edge cases too. 

3. For `app/src/test/java/org/oppia/android/app/testing/options/ReadingTextSizeFragmentTest.kt`, could you mention what all cases does #2929 covering and what cases are need to be covered as a part of this PR. This way we can keep track of all the cases to get covered in testing. 

4. Let's remove the unused variables. 
```kt
private const val NO_AUDIO = 0
private const val HINDI = 3
private const val CHINESE = 4
```

5. Add kdoc for test files 
reference :
```kt
/** Tests for [OptionsFragment]. */
``` 

6. Could you confirm if we need `ReadingTextSizeActivityTest`?

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

Sparsh1212:
@anandwana001 

Regarding:
1. I tried running the Espresso tests by right-clicking the file and running the test for `AppLanguageFragmentTest` and `AudioLanguageFragmentTest` but it is running the Roboelectric tests instead of Espresso. I also tried running them by making the Espresso configuration in the Edit Configuration but the tests are still not running, it displays `No tests found in class`. This problem isn't occurring for `OptionsFragmentTest`. Can you please cross-check if you are facing the same problem?
2. Tests added for tablet configuration.
3. It currently has the following tests:
    1. `testTextSize_checkTextSizeOfAllFourItems_textSizeMatchedCorrectly()` 
    2. `testTextSize_changeTextSizeToLarge_changeConfiguration_checkTextSizeLargeIsSelected()`
    3. `testTextSize_changeTextSizeToMedium_mediumItemIsSelected()`
    The recent changes in this PR has covered all the required tests. So, I think we don't have to make any further changes to it.
4. Removed unused variables.
5. Added kdocs.
6. Yes we need the `ReadingTextSizeActivityTest` to check that the accessibility label is defined properly. There's also a corresponding issue for this: #2817

app/src/main/java/org/oppia/android/app/options/OptionsFragment.kt
Sparsh1212: I have removed the previous consts mentioned at the top of the test file. PTAL.

anandwana001:
Looking into the file path, `AudioLanguageFragmentTest` and `AppLanguageFragmentTest` are always in the Local unit tests folder which is `app/testing/..`. that's why there are not running on espresso. 

If you look here there are 2 files:
1. src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
2. src/test/java/org/oppia/android/app/testing/options/OptionsFragmentTest.kt

`sharedTest` directory is used for instrumentation test and `test` directory is used for unit test, and the test we are working for in this PR is usually instrumentation test. 
You need to shift the required test to `sharedTest` directory.

Sparsh1212:
> Looking into the file path, `AudioLanguageFragmentTest` and `AppLanguageFragmentTest` are always in the Local unit tests folder which is `app/testing/..`. that's why there are not running on espresso.
> 
> If you look here there are 2 files:
> 
> 1. src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
> 2. src/test/java/org/oppia/android/app/testing/options/OptionsFragmentTest.kt
> 
> `sharedTest` directory is used for instrumentation test and `test` directory is used for unit test, and the test we are working for in this PR is usually instrumentation test.
> You need to shift the required test to `sharedTest` directory.

Thanks a lot.
I have shifted them to `sharedTest` directory and all the Espresso tests are passing as well. 
Updated the PR description. PTAL.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -207,31 +208,56 @@ class OptionsFragmentTest {
...
-      }
+      ).check(
+        matches(withText("Medium"))
+      )
+    }
+  }
 
-      getInstrumentation().waitForMonitorWithTimeout(activityMonitor, 3)
+  @Test
+  @Config(qualifiers = "sw600dp")
```
anandwana001, line 253: please add espresso test result for these test too, on tablet.

app/src/sharedTest/java/org/oppia/android/app/options/AppLanguageFragmentTest.kt
```diff
@@ -96,18 +90,52 @@ class AppLanguageFragmentTest {
...
+  @Test
+  fun testAppLanguage_configChange_selectedLanguageIsEnglish() {
+    launch<AppLanguageActivity>(createAppLanguageActivityIntent("English")).use {
+      rotateToLandscape()
+      checkSelectedLanguage(ENGLISH)
+    }
+  }
+
+  @Test
+  @Config(qualifiers = "sw600dp")
```
anandwana001, line 118: ditto

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
```diff
@@ -97,32 +89,60 @@ class AudioLanguageFragmentTest {
...
+      checkSelectedLanguage(ENGLISH)
+      selectLanguage(FRENCH)
       rotateToLandscape()
-      checkSelectedLanguage(HINDI)
+      checkSelectedLanguage(FRENCH)
     }
   }
 
   @Test
   @Config(qualifiers = "sw600dp")
```
anandwana001, line 139: ditto

anandwana001 (changes requested):
Thanks @Sparsh1212 
shifting to sharedTest is good. 

Need tablet test result screenshot. Please run these tests on tablet config and see if they are passing correctly.

rt4914:
Un-assigning until @anandwana001 approves it.

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/AppLanguageFragmentTest.kt
Sparsh1212: Done.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
Sparsh1212: Done.

Sparsh1212:
> Thanks @Sparsh1212
> shifting to sharedTest is good.
> 
> Need tablet test result screenshot. Please run these tests on tablet config and see if they are passing correctly.

Added.

anandwana001 (approved):
LGTM, Thanks @Sparsh1212

app/src/main/java/org/oppia/android/app/options/OptionsFragment.kt
rt4914: Actually you have removed it from test file. But instead what we want it we should either use only 
```
const val MESSAGE_READING_TEXT_SIZE_ARGUMENT_KEY = "OptionsFragment.message_reading_text_size"
const val MESSAGE_APP_LANGUAGE_ARGUMENT_KEY = "OptionsFragment.message_app_language"
const val MESSAGE_AUDIO_LANGUAGE_ARGUMENT_KEY = "OptionsFragment.message_audio_language"
```
in tests and production code

or we should use only
```
const val KEY_READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE =
      "READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE"
    const val APP_LANGUAGE_PREFERENCE_SUMMARY_VALUE_EXTRA_KEY =
      "AppLanguageActivity.app_language_preference_summary_value"
    const val KEY_AUDIO_LANGUAGE_PREFERENCE_SUMMARY_VALUE =
      "AUDIO_LANGUAGE_PREFERENCE_SUMMARY_VALUE"
```


Currently you are using the above for the app layer code and the below for test cases which is incorrect. We should use only one of these sets in both code.

rt4914 (changes requested):
@Sparsh1212 Replied to my one of the open comments.

app/src/main/java/org/oppia/android/app/options/OptionsFragment.kt
Sparsh1212: Actually, we are passing the key `KEY_READING_TEXT_SIZE_PREFERENCE_SUMMARY_VALUE` while creating the `ReadingTextSizeActivity` intent. In the `OptionsFragment` test, we are checking that the correct intent has passed or not when the user clicks on the `RecyclerView` item, so we need the value of this key in the test file to match the intent passed, but the problem with accessing this key in the test file is that it has been declared as an `internal const` in the companion object of `ReadingTextSizeActivity` therefore the test fails in Bazel roboelectric because of the Test file not being able to access those internal consts, therefore I have to again declare them in `companion object` of `ReadingTextSizeFragment`. Please suggested an alternative approach to this if this is incorrect so that we can access the key in the test file.

rt4914: You can use the values of removing the `internal` in `ReadingTextSizeActivity` . Won't that work? @Sparsh1212 

Also the solution where you have suggested introducing them in `ReadingTextSizeFragment ` is incorrect because its not part of intent or activity so the testing would be incorrect.

Sparsh1212: @rt4914  Yes removing `internal` in `ReadingTextSizeActivity` should work. Also, apologies about the alternative approach, I meant `OptionsFragment` not `ReadingTextSizeFragment`. Made a typo there. 
I have updated the changes. PTAL.
Also, the CI checks are failing, I hope this isn't due to this PR.

rt4914 (approved):
LGTM, thanks.
Unfortunately we cannot merge this until #3157 is fixed.

Sparsh1212:
@rt4914 I have taken the pull from the latest develop branch. The checks are also passing. Looks like we can merge this now :)

PR #3116, opened 2021-04-23
Fix #3075 and #2290: Use default keyboards for interactions with current nonstandard keyboards

### LINKED ISSUE #2290: Keyboard not showing all input options that are needed for the answer
**Describe the bug**
'/', ' ' are not being shown as an input option on the keyboard despite the required input being of the type 'p/q' of 'r p/q'

**To Reproduce**
1. Open the App and start the exploration on First Test Topic 
2. Click on 'Prototype Exploration'
3. First question is 'What fraction represents half of something?'
4. Keyboard only shows inputs -> 0123456789

**Expected behaviour**
The keyboard is also supposed to show space (' ') and 'forward slash ('/') as input options.

**Screenshots**
 

**Device**
- Samsung A50
- SDK version 29

**Additional context**
Important point to notice here is that it seems to occur **only** on Samsung keyboard. It worked fine when I download Gboard and tried it on that. But many users who use the default samsung keyboard may get stuck on similar cases.

### LINKED ISSUE #3075: Use default keyboard in alpha
Fraction interaction can't be inputted properly on certain devices due to Android showing the wrong keyboard. We should disable the filtered keyboards for now while we consider next steps.

BenHenning:
## Explanation
Fix #3075
Fix #2290

This fixes issues with two of our interactions: fractions & ratios, both of which use less common keyboards in non-standard ways by limiting the characters the user can type. This seems to outright break the input experience for some users on certain devices (i.e. they can't progress), so we're preferring to move back to the standard full keyboard (or numeric for numeric input) until a custom math keyboard can be introduced (which is currently being researched by our product team).

To keep things simple, I added a few change detector tests to just verify that the input type is being properly set for both fractions & ratios (to standard text).

Tests failing without fixes:
image

Tests passing with fixes:
image

The changes in StateFragmentTest were also verified as passing when run locally with Espresso.


BenHenning:
@rt4914 PTAL when you get a chance.

BenHenning:
Note I will be adding change detector tests for input type for fractions & ratios.

rt4914 (approved):
LGTM.

Couple of things that needs to be done. Add tests for inputType.

I will try to check this manually on Samsung device and follow up.

Thanks.

BenHenning:
Note that the Bazel/Gradle TopicControllerTest failure is due to https://github.com/oppia/oppia-android/pull/3017#issuecomment-826033445. After that PR is merged & this one updated, the failure should go away.

BenHenning:
Similar to https://github.com/oppia/oppia-android/pull/2986#issuecomment-826033128 I am going to verify the Bazel tests locally & rely on previously passing runs + Gradle passing as a proxy so that the PR can go in a bit faster.

BenHenning:
Confirming StateFragmentTest passes locally with Bazel. Confirming also that StateFragmentTest passes with Espresso (or rather the same tests that were failing in #2985 are still failing as expected/there are no new failures).

BenHenning:
Thanks @rt4914. The new tests are added and seem to be working. I'm going to go ahead and submit this since then we can start gathering QA feedback. If you notice any issues when testing on a Samsung device please file a bug & add it to the alpha MR2 milestone.

PR #3119, opened 2021-04-26
Fix #2427: Create LogoutDialogFragment

### LINKED ISSUE #2427: Logout Dialog disappears on orientation change

ArpitShukIa:
## Explanation
Fixes #2427: I created a new DialogFragment replacing the old AlertDialog in AdministratorControlsAccountActionsViewModel and removed the IntentFactoryShim dependency as it is no longer being used. Also added test for the fix.

### Test Results:
Robolectric:
image

Expresso:
image

## Checklist

### DIFF
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
@@ -16,6 +16,7 @@ const val SELECTED_CONTROLS_TITLE_SAVED_KEY =
 const val LAST_LOADED_FRAGMENT_KEY = "LAST_LOADED_FRAGMENT_KEY"
 const val PROFILE_LIST_FRAGMENT = "PROFILE_LIST_FRAGMENT"
 const val APP_VERSION_FRAGMENT = "APP_VERSION_FRAGMENT"
+const val LOGOUT_DIALOG_FRAGMENT = "LOGOUT_DIALOG_FRAGMENT"
 
 /** Activity for Administrator Controls. */
 class AdministratorControlsActivity :
@@ -23,7 +24,8 @@ class AdministratorControlsActivity :
   RouteToProfileListListener,
   RouteToAppVersionListener,
   LoadProfileListListener,
-  LoadAppVersionListener {
+  LoadAppVersionListener,
+  ShowLogoutDialogListener {
   @Inject
   lateinit var administratorControlsActivityPresenter: AdministratorControlsActivityPresenter
   private lateinit var lastLoadedFragment: String
@@ -79,6 +81,10 @@ class AdministratorControlsActivity :
     administratorControlsActivityPresenter
       .setExtraControlsTitle(getString(R.string.administrator_controls_app_version))
     administratorControlsActivityPresenter.loadAppVersion()
+  }
+
+  override fun showLogoutDialog() {
+    LogoutDialogFragment.newInstance().showNow(supportFragmentManager, LOGOUT_DIALOG_FRAGMENT)
   }
 
   override fun onSaveInstanceState(outState: Bundle) {
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
@@ -84,7 +84,7 @@ class AdministratorControlsActivity :
 
   override fun showLogoutDialog() {
     LogoutDialogFragment.newInstance()
-      .showNow(supportFragmentManager, LogoutDialogFragment.LOGOUT_DIALOG_FRAGMENT)
+      .showNow(supportFragmentManager, LogoutDialogFragment.TAG_LOGOUT_DIALOG_FRAGMENT)
   }
 
   override fun onSaveInstanceState(outState: Bundle) {
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
@@ -16,7 +16,6 @@ const val SELECTED_CONTROLS_TITLE_SAVED_KEY =
 const val LAST_LOADED_FRAGMENT_KEY = "LAST_LOADED_FRAGMENT_KEY"
 const val PROFILE_LIST_FRAGMENT = "PROFILE_LIST_FRAGMENT"
 const val APP_VERSION_FRAGMENT = "APP_VERSION_FRAGMENT"
-const val LOGOUT_DIALOG_FRAGMENT = "LOGOUT_DIALOG_FRAGMENT"
 
 /** Activity for Administrator Controls. */
 class AdministratorControlsActivity :
@@ -84,7 +83,8 @@ class AdministratorControlsActivity :
   }
 
   override fun showLogoutDialog() {
-    LogoutDialogFragment.newInstance().showNow(supportFragmentManager, LOGOUT_DIALOG_FRAGMENT)
+    LogoutDialogFragment.newInstance()
+      .showNow(supportFragmentManager, LogoutDialogFragment.LOGOUT_DIALOG_FRAGMENT)
   }
 
   override fun onSaveInstanceState(outState: Bundle) {
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
@@ -1,0 +1,25 @@
+package org.oppia.android.app.administratorcontrols
+
+import android.app.Dialog
+import android.os.Bundle
+import androidx.appcompat.app.AlertDialog
+import androidx.fragment.app.DialogFragment
+import org.oppia.android.R
+import org.oppia.android.app.profile.ProfileChooserActivity
+
+/** [DialogFragment] that gives option to either cancel or logout from current profile. */
+class LogoutDialogFragment : DialogFragment() {
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    return AlertDialog.Builder(requireContext(), R.style.AlertDialogTheme)
+      .setMessage(R.string.log_out_dialog_message)
+      .setNegativeButton(R.string.log_out_dialog_cancel_button) { dialog, _ ->
+        dialog.dismiss()
+      }
+      .setPositiveButton(R.string.log_out_dialog_okay_button) { _, _ ->
+        // TODO(#762): Replace [ProfileChooserActivity] to [LoginActivity] once it is added.
+        val intent = ProfileChooserActivity.createProfileChooserActivity(activity!!)
+        startActivity(intent)
+      }.create()
+  }
+}
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
@@ -10,6 +10,12 @@ import org.oppia.android.app.profile.ProfileChooserActivity
 /** [DialogFragment] that gives option to either cancel or logout from current profile. */
 class LogoutDialogFragment : DialogFragment() {
 
+  companion object {
+    fun newInstance(): LogoutDialogFragment {
+      return LogoutDialogFragment()
+    }
+  }
+
   override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
     return AlertDialog.Builder(requireContext(), R.style.AlertDialogTheme)
       .setMessage(R.string.log_out_dialog_message)
--- a/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
+++ b/app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
@@ -7,7 +7,7 @@ import androidx.fragment.app.DialogFragment
 import org.oppia.android.R
 import org.oppia.android.app.profile.ProfileChooserActivity
 
-/** [DialogFragment] that gives option to either cancel or logout from current profile. */
+/** [DialogFragment] that gives option to either cancel or log out from current profile. */
 class LogoutDialogFragment : DialogFragment() {
 
   companion object {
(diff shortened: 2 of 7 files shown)


ArpitShukIa:
@rt4914 PTAL.

ArpitShukIa:
The bazel checks are failing. Probably because LogoutDialogFragment is not visible to AdministratorControlsAccountActionsViewModel.
How to fix this?

rt4914:
Will review it once it gets review by @prayutsu or @FareesHussain

app/src/main/java/org/oppia/android/app/administratorcontrols/administratorcontrolsitemviewmodel/AdministratorControlsAccountActionsViewModel.kt
```diff
@@ -1,29 +1,21 @@
...
-      }
-      .setPositiveButton(R.string.log_out_dialog_okay_button) { _, _ ->
-        // TODO(#762): Replace [ProfileChooserActivity] to [LoginActivity] once it is added.
-        val intent = intentFactoryShim.createProfileChooserActivityIntent(fragment.activity!!)
-        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
-        fragment.activity!!.startActivity(intent)
-        fragment.activity!!.finish()
-      }.create().show()
+    LogoutDialogFragment().showNow(
+      fragment.requireActivity().supportFragmentManager,
```
prayutsu: In the future, we may probably want to add other logic with LogoutDialogFragment, so suggest making a method to provide the LogOutDialogFragment instance inside a Companion Object. You can see other DialogFragment for reference.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
```diff
@@ -0,0 +1,25 @@
...
+import android.app.Dialog
+import android.os.Bundle
+import androidx.appcompat.app.AlertDialog
+import androidx.fragment.app.DialogFragment
+import org.oppia.android.R
+import org.oppia.android.app.profile.ProfileChooserActivity
+
+/** [DialogFragment] that gives option to either cancel or logout from current profile. */
+class LogoutDialogFragment : DialogFragment() {
+
```
prayutsu, line 12: Not sure, but shouldn't the letter 'o' be uppercase?

@rt4914 @ArpitShukIa WDYT?

prayutsu (changes requested):
Left some comments. PTAL @ArpitShukIa

app/src/main/java/org/oppia/android/app/administratorcontrols/administratorcontrolsitemviewmodel/AdministratorControlsAccountActionsViewModel.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
ArpitShukIa: Which 'o'?
Do you mean "LogOutDialogFragment"?

app/src/main/java/org/oppia/android/app/administratorcontrols/administratorcontrolsitemviewmodel/AdministratorControlsAccountActionsViewModel.kt
```diff
@@ -1,29 +1,21 @@
 package org.oppia.android.app.administratorcontrols.administratorcontrolsitemviewmodel
 
-import android.content.Intent
-import androidx.appcompat.app.AlertDialog
 import androidx.fragment.app.Fragment
-import org.oppia.android.R
-import org.oppia.android.app.shim.IntentFactoryShim
+import androidx.lifecycle.ViewModel
+import org.oppia.android.app.administratorcontrols.AdministratorControlsFragment
```
FareesHussain: It looks like a unused import (ktlint might have ignored it as it is used in the KDoc)

FareesHussain (changes requested):
@ArpitShukIa Implementation LGTM,

PTAL there is a bazel build error.

As `AdministratorControlsAccountActionsViewModel` no longer imports the resources move it from [VIEW_MODELS_WITH_RESOURCE_IMPORTS](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L128) to [VIEW_MODELS](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L157) source set to optimize the build but this doesn't solve the bazel error, as we don't have Activities or Fragments in the viewmodels so far there is no `dep` added to the [view_models](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L554) target. 

To fix this we need to add the `:app` dep to the `:view_models` but this is definitely not a  good practice.

I suggest creating a new Listner and use the LogoutDialogFragment within the AdministratorControlsActivity or AdministratorControlsFragment

@BenHenning WDYT?

app/src/main/java/org/oppia/android/app/administratorcontrols/administratorcontrolsitemviewmodel/AdministratorControlsAccountActionsViewModel.kt
ArpitShukIa: What's to be done here?

ArpitShukIa: Removed them

FareesHussain: Thanks, you can resolve this now

FareesHussain (approved):
LGTM, Thanks

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
prayutsu: Yes, I meant LogOutDialogFragment but I am not really sure if we want to change it or not.

prayutsu (approved):
LGTM, Thanks @ArpitShukIa

app/src/main/java/org/oppia/android/app/administratorcontrols/administratorcontrolsitemviewmodel/AdministratorControlsAccountActionsViewModel.kt
prayutsu: This can be resolved now :)

BenHenning:
Sorry, I think I'll need to take a pass on this tomorrow, but I can follow up to the comment above now.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
BenHenning: The noun form is correct here (i.e. this is the dialog where logouts occur). The documentation should say 'log out' since the verb form is the action the user is taking. Another way to check: methods almost certainly should be logOut since they are verb phrases, and classes will usually be logout since they are noun phrases.

BenHenning:
> @ArpitShukIa Implementation LGTM,
> 
> PTAL there is a bazel build error.
> 
> As `AdministratorControlsAccountActionsViewModel` no longer imports the resources move it from [VIEW_MODELS_WITH_RESOURCE_IMPORTS](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L128) to [VIEW_MODELS](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L157) source set to optimize the build but this doesn't solve the bazel error, as we don't have Activities or Fragments in the viewmodels so far there is no `dep` added to the [view_models](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L554) target.
> 
> To fix this we need to add the `:app` dep to the `:view_models` but this is definitely not a good practice.
> 
> I suggest creating a new Listner and use the LogoutDialogFragment within the AdministratorControlsActivity or AdministratorControlsFragment
> 
> @BenHenning WDYT?

I agree with moving this out of the resources view model list if R is no longer needed. Regarding the dependency, I'm fairly certain you can't add :app to :view_models since that will likely result in a circular dependency.

I'm not sure I'm following what the need for the listener is. What specific problem do you see here @FareesHussain?

FareesHussain:
> I'm not sure I'm following what the need for the listener is. What specific problem do you see here @FareesHussain?

https://github.com/oppia/oppia-android/runs/2445936672

Before using the listener. the `AdministratorControlsAccountActionsViewModel` imports the LogoutDialogFragment. I had a look into the error and found that the LogoutDialogFragment comes under the `:app` target, I noticed none of the other viewmodels have imported from the `:app` target so to fix the Bazel error, I suggested adding a listener inorder to call the `showLogoutDialog` function. I'm not sure if there is any other way to fix it.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
prayutsu: Sounds good, thanks for the clarification @BenHenning.
This can be resolved now.

ArpitShukIa: > The documentation should say 'log out' since the verb form is the action the user is taking

Fixed it.

BenHenning:
Ah I see, thanks for clarifying @FareesHussain. Yes, that seems like the correct way to solve this since the interface provides indirection (plus it's a bit cleaner not to directly interact with the activity/fragment lifecycles except at in that activity or fragment implementation class).

BenHenning:
I'm actually going to take a pass on just my codeowners here (presumably Bazel) after @rt4914 has a chance to review it in case there are any big changes.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
```diff
@@ -16,14 +16,16 @@ const val SELECTED_CONTROLS_TITLE_SAVED_KEY =
 const val LAST_LOADED_FRAGMENT_KEY = "LAST_LOADED_FRAGMENT_KEY"
 const val PROFILE_LIST_FRAGMENT = "PROFILE_LIST_FRAGMENT"
 const val APP_VERSION_FRAGMENT = "APP_VERSION_FRAGMENT"
+const val LOGOUT_DIALOG_FRAGMENT = "LOGOUT_DIALOG_FRAGMENT"
```
rt4914: This tag should be shifted to `LogoutDialogFragment` companion object and then it should be used wherever requried.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    return AlertDialog.Builder(requireContext(), R.style.AlertDialogTheme)
+      .setMessage(R.string.log_out_dialog_message)
+      .setNegativeButton(R.string.log_out_dialog_cancel_button) { dialog, _ ->
+        dialog.dismiss()
+      }
+      .setPositiveButton(R.string.log_out_dialog_okay_button) { _, _ ->
+        // TODO(#762): Replace [ProfileChooserActivity] to [LoginActivity] once it is added.
```
rt4914: This TODO is not needed here. You can remove this.

rt4914 (approved):
LGTM, thanks.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/administratorcontrols/ShowLogoutDialogListener.kt
```diff
@@ -0,0 +1,6 @@
+package org.oppia.android.app.administratorcontrols
+
+/** Listener for when an activity should show [LogoutDialogFragment]. */
```
BenHenning: Suggest: ``Listener for when the logout dialog should be shown.``

app/src/main/java/org/oppia/android/app/administratorcontrols/ShowLogoutDialogListener.kt
```diff
@@ -0,0 +1,6 @@
...
+interface ShowLogoutDialogListener {
+  fun showLogoutDialog()
```
BenHenning, line 7: This should have a KDoc, too, such as: ``Shows the logout dialog. This cannot be called at times when there's an ongoing fragment transaction.`` (the latter providing some context on when it's valid to call this).

BenHenning (changes requested):
Thanks @ArpitShukIa! Generally LGTM for Bazel changes, but had 2 nit comments to address before approving.

app/src/main/java/org/oppia/android/app/administratorcontrols/ShowLogoutDialogListener.kt
ArpitShukIa: Done

app/src/main/java/org/oppia/android/app/administratorcontrols/ShowLogoutDialogListener.kt
ArpitShukIa: Done

ArpitShukIa:
@BenHenning I have committed the requested changes. PTAL.

BenHenning (approved):
Thanks @ArpitShukIa. LGTM for codeowners.

BenHenning:
The Bazel issues are existing problems and are unrelated to this PR, so merging this as-is.

PR #3147, opened 2021-04-30
Fixes #3145: ExplorationActivityTest failing tests fixed

### LINKED ISSUE #3145: Audio related tests failing on espresso
`testAudioWithNoVoiceover_openPrototypeExploration_checkAudioButtonIsHidden`
`testAudioWithNoVoiceover_openPrototypeExploration_configurationChange_checkAudioButtonIsHidden`

Above mentioned tests are failing on Espresso. The implementation should be updated such that all these tests pass.

The tests are failing mostly because of commit: https://github.com/oppia/oppia-android/commit/cb1ddc81006bc599a8f527b1f67e6b1522fef362

where we. have changed the proto/json files.

rt4914:
## Explanation

Fixes #3145 

The two tests were failing because the `createVoiceoverMappingsFromJson` was returning map for empty `content-ids` too.

This PR updates the implementation to check if `voiceover/contentid` json is empty then do not add that as `Voiceover` to the map.

## Screenshot
In current develop the `audio` icon is available in toolbar in all explorations. The below screenshot shows that it is no more visible for for states which do not contain audio.
  

Screenshot 2021-04-30 at 11 50 37 AM

Screenshot 2021-05-04 at 7 15 08 PM

## Checklist


domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -191,17 +191,21 @@ class StateRetriever @Inject constructor() {
   private fun createVoiceoverMappingsFromJson(
     recordedVoiceovers: JSONObject
   ): Map<String, VoiceoverMapping> {
+    val voiceoverMappingMap = HashMap<String, VoiceoverMapping>()
     val voiceoverMappingJson = recordedVoiceovers.getJSONObject("voiceovers_mapping")
-    return voiceoverMappingJson.keys().asSequence().associateWith { contentId ->
+    voiceoverMappingJson.keys().asSequence().associateWith { contentId ->
```
BenHenning: Could we maybe just add a filter on the map to filter out content IDs that map to empty voiceover mappings? Prefer not using a mutable map with a non-mutable map construction; this is a bit hard to follow and I suspect may be confusing in the future.

BenHenning (changes requested):
Thanks @rt4914 & nice catch. I would ask for a test to be added, but given this is temporary code I think time is better spent not bothering if ExplorationActivityTest confirms the fix.

Just have 1 comment to address--PTAL.

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
rt4914: PTAL at updated logic.

domain/src/main/java/org/oppia/android/domain/util/StateRetriever.kt
```diff
@@ -192,7 +192,9 @@ class StateRetriever @Inject constructor() {
     recordedVoiceovers: JSONObject
   ): Map<String, VoiceoverMapping> {
     val voiceoverMappingJson = recordedVoiceovers.getJSONObject("voiceovers_mapping")
-    return voiceoverMappingJson.keys().asSequence().associateWith { contentId ->
+    return voiceoverMappingJson.keys().asSequence().filter { contentId ->
+      voiceoverMappingJson.getJSONObject(contentId).length() != 0
```
BenHenning, line 196: Probably not the most efficient, but it seems like a reasonable stop-gap.

BenHenning (approved):
Thanks @rt4914. LGTM!

PR #3166, opened 2021-05-06
Introducing Oppiabot configure file

anandwana001:
## Checklist


anandwana001:
@vojtechjelinek @gp201 PTAL

.github/workflows/oppiabot.yml
```diff
@@ -0,0 +1,21 @@
+# Contains jobs corresponding to Oppiabot.
+
+name: Oppiabot
+
+on:
+  pull_request_target:
+    branches:
+      - develop
```
gp201, line 8: On what branches should the CLA check be enabled?

gp201 (approved):
left a comment. LGTM

.github/workflows/oppiabot.yml
anandwana001: As we need CLA (Oppiabot) for all the Pull request to take action, how should we mention the branch?
Also, will it work for fork branches as well?

gp201: When you mention "develop" here: https://github.com/oppia/oppia-android/blob/f8a5a936c8d9adf8f1ada6227362d223b3cb5604/.github/workflows/oppiabot.yml#L5-L8

It means that the CLA Check will run on any PR that has the develop branch as its base.
Do we need the check to run on all branches?
And could you rephrase "Also, will it work for fork branches as well?"?? I'm not able to understand?

anandwana001: This yml file is using secrets, which PRs from fork might not have access, so will the CLA works as we required?

gp201: Yes, the PR's from a fork will be able to access the secrets since it runs on the event. `pull_request_target`. ([Source](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target)).

If you have doubts regarding the security please read the doc I had shared.

gp201 (changes requested):
Adding myself as blocking so @vojtechjelinek can work with Sean to set the secret up once this is reviewed.

.github/workflows/oppiabot.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  oppiabot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: CLA Check
+        uses: oppia/oppiabot@1.2.0
+        env:
+          SHEETS_TOKEN: ${{ secrets.SHEETS_TOKEN }}
```
BenHenning, line 18: I don't think this is going to work. We can't use secrets like this for forks since they will be exposable to users.

gp201:
@BenHenning PTAL

.github/workflows/oppiabot.yml
```diff
@@ -0,0 +1,21 @@
...
+          SHEETS_CRED: ${{ secrets.SHEETS_CRED }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SPREADSHEET_ID: ${{ secrets.SPREADSHEET_ID }}
```
gp201:
```suggestion
        uses: oppia/oppiabot@1.3.0
        with:
          repo-token: ${{secrets.GITHUB_TOKEN}}
        env:
          SHEETS_TOKEN: ${{ secrets.SHEETS_TOKEN }}
          SHEETS_CRED: ${{ secrets.SHEETS_CRED }}
          SPREADSHEET_ID: ${{ secrets.SPREADSHEET_ID }}
```

Oppiabot made a new release with some changes to CLA Check Github Action

gp201:
@anandwana001 left a change

.github/workflows/oppiabot.yml
BenHenning, line 20: Please follow up with me regarding which secrets need to be added & what their values are, otherwise this will fail when checked in.

BenHenning (approved):
Thanks @anandwana001. This LGTM given it's using pull_request_target. Please follow up with me regarding the secrets before checking this in.

Also, please follow up with me with examples for both fork & non-fork PRs to validate that the check is working (we should have an example for both contributor & non-contributor, fork & non-fork).

.github/workflows/oppiabot.yml
anandwana001: Done

.github/workflows/oppiabot.yml
anandwana001: @gp201 Could you follow up here?
> which secrets need to be added & what their values are, otherwise this will fail when checked in.

gp201: Secrets can be added at the organization level and the repository level. I'm not sure what approach is being taken currently. We have to consult @vojtechjelinek for the secrets.

vojtechjelinek: Since we share the CLA sheet between web and Android I think we can add this on org. level. So @seanlip should be able to add those.

BenHenning: We actually have a lot of repos under the org now. It would probably better to keep these repo level to restrict access. 

@seanlip do you have any thoughts on this?

gp201: @BenHenning Just an FYI, We have to give permission for the GitHub action to run on a repo in [this file](https://github.com/oppia/oppiabot/blob/65f1717d62f4f3358db40909e3c4228398ee14c1/constants.js#L61) in oppiabot.

BenHenning: Also, shouldn’t we be passing these values into the action rather than setting them as environment variables? While the env context is limited to this script, it seems a bit more secure to pass the secrets directly to the action like we are for the github token.

seanlip: > We actually have a lot of repos under the org now. It would probably better to keep these repo level to restrict access.
> @seanlip do you have any thoughts on this?

Would it be possible to explain more why having an org-level secret results in additional access issues -- who can access it and how would they do so? Do forks have access to org-based actions but not repo-based actions?

vinitamurthi (approved):

.github/workflows/oppiabot.yml
seanlip: Hi -- I had a quick chat with @BenHenning on this and I think the core concern is about secrets for one repo in the org being leaked by another repo accidentally through careless handling.

I think this makes sense, we don't know what future repos we may create and what our policies on those would be. So let's go with repo-level for now.

gp201: > Also, shouldn’t we be passing these values into the action rather than setting them as environment variables? While the env context is limited to this script, it seems a bit more secure to pass the secrets directly to the action like we are for the github token.

@BenHenning 
I could add this as an issue on Github and make the change in a later PR so that it will change in the next release. I don't think it would make much of a difference, but it's always better to have it.

seanlip:
Just a heads-up -- secrets have been added. How do we test that things are working correctly?

.github/workflows/oppiabot.yml
BenHenning: @gp201 I'm fine with that. Can you please link to that issue here?

BenHenning (commented):
Generally LGTM, but waiting for the open thread to get wrapped up before approving.

BenHenning (changes requested):
Hmm, flipping back to 'request changes' per previous comment. I'm also interested in seeing this tested before merging it into develop.

gp201:
Steps to test the GitHub action in the Oppia-Android repository.

1. Create a new branch in the Oppia-Andoird repo for testing the action
2. Change the branch `develop` to the new branch created. https://github.com/oppia/oppia-android/blob/c497bb3429c45705f016f6191d746dedae674d6b/.github/workflows/oppiabot.yml#L7-L8
3. Merge the PR to the **NEW** branch.
4. Make a PR to the **NEW** branch
5. The GitHub action should run.

This has already been performed in the Oppia repository.
PR that was used for testing: https://github.com/oppia/oppia/pull/13136
To test this please close and re-open the PR.

.github/workflows/oppiabot.yml
gp201: @BenHenning The issue: https://github.com/oppia/oppiabot/issues/275

gp201:
@BenHenning PTAL

BenHenning:
Thanks for the instructions @gp201 they were really helpful! Per #3438 & #3439 it appears that Oppiabot is working correctly. Going ahead and approving this.

BenHenning (approved):
Thanks @anandwana001. LGTM.

PR #3249, opened 2021-05-25
[RunAllTests] Fix #3217, part of #2923: migrate flexbox dependency to pull from Google's Maven repository instead of JCenter/bintray

BenHenning:
Fix #3217.
Fix part of #2923.

## Explanation
This updates the Bazel & Gradle builds to use the new version & group ID for flexbox-layout so that the build pulls the dependency from Google Maven instead of JCenter/Bintray (per https://github.com/google/flexbox-layout/issues/566#issuecomment-844630491). While Bintray is promising long-term hosting for dependencies, it's immutable which means new versions can't be received. Beyond that, we've had reliability issues with our JCenter/Bintray dependencies. This removes one JCenter/Bintray dependency.

This should have no side effects with the current project since 3.0.0 is meant to be identical with 2.0.1. I'll be using CI & a manual verification of the APK (to ensure the walkthrough experience, our only screen that uses flexbox layout, works correctly) for validity checking.

PR #3259, opened 2021-05-27
[RunAllTests] Fix #2923: Remove direct bintray/jcenter dependence

### LINKED ISSUE #2923: Move project off of jcenter
See https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/. JCenter is turning down over the next 12 months. Most, if not all, of our packages should be available on other Maven repositories. This issue is tracking updating both Gradle & Bazel to:
1. Remove our bintray/jcenter repository references
2. Investigating all missing dependencies to find alternate repositories
3. Updating Gradle/Bazel to use the new repositories in place of JCenter
4. Ensure that the app builds with both Gradle & Bazel

BenHenning:
Fix #2923 

## Explanation
This PR removes dependence on JCenter/Bintray given the service is now frozen & can no longer receive updated dependencies (and will likely be turned down at some point). This required effectively no work on the Bazel side since it seems Bazel builds already didn't depend on JCenter as of #3249.

Our CircularImageview dependency was the only one exclusively available on JCenter, but we fortunately already had an Oppia-specific fork of it that, with a few updates, became jitpack compatible. Interestingly, we also had a bunch of other dependencies coming from JCenter that could easily be replaced with Maven Central. This will likely improve build reliability since JCenter has had some service issues over the past few months.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -116,7 +116,7 @@ google_services_workspace_dependencies()
 
 git_repository(
     name = "circularimageview",
-    commit = "8a65ba42b3fee21b5e19ca5c8690185f7c60f65d",
+    commit = "35d08ba88a4a22e6e9ac96bdc5a68be27b55d09f",
     remote = "https://github.com/oppia/CircularImageview",
 )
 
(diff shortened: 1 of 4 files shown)


BenHenning:
Needed to retrigger the build since it was hanging on identifying Bazel targets again.

The re-run requires a second re-run of Gradle tests since it hit a known Gradle test flake.

rt4914 (approved):
LGTM, thanks.

fsharpasharp (approved):
LGTM!

PR #3269, opened 2021-06-07
Fixes #3289: New Proto classes for Platform Parameters

### LINKED ISSUE #3289: New Protos for the Platform Parameter Project
Add new protos for the platform parameter project. Namely 
- PlatformParameter
- PlatformParameterWithName
- PlatformParameterDatabase
- PlatformParameterMap

Target PR date: 7 June 2021
Target completion date: 11 June 2021

Arjupta:
## Explanation

Fixes #3289 
This PR introduces new Proto Classes which will be used to implement Platform Parameter Project in Oppia. This PR is reffered as PR-1.1 in the Milestones. ( /cc @Sarthak2601 )

## Checklist


model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+
+// It will correspond to a single Platform Parameter Value.
+message Parameter {
```
jcqli: Parameter is a very generic term -- I'd suggest using something more specific to this feature, like `PlatformParameter` for parity with the backend as well as explicitly identifying what type of parameter it is. I'm not sure `PlatformParameter` is specific enough either, though, @vinitamurthi WDYT?
(Note that if this changes we would also want to apply the new name to the below protos as well)

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
...
+
+    // Value of a parameter can be of any one of the following type.
+    oneof value{
+
+        // Boolean type parameter value.
+        bool boolean = 3;
+
+        // Integer type parameter value.
+        int32 integer = 2;
+
```
jcqli: nit: remove the empty lines between the fields and above `value`

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
...
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+
+// It will correspond to a single Platform Parameter Value.
+message Parameter {
+
+    // Value of a parameter can be of any one of the following type.
+    oneof value{
```
jcqli: nit: `oneof value {`

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+
```
jcqli: nit: can remove one of the empty lines

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
...
+// Storage format of platform parameters while the app is running.
+message ParameterMap{
+    map<string,Parameter> parameter = 1;
+}
+
+// Pair of Platform Parameter Value and Name.
+message ParameterWithName{
+    string name = 1;
+    Parameter value = 2;
```
jcqli: Could this be:
```
oneof value {
  boolean
  int
  string
}
```
instead?

This eliminates a layer of nesting and could provide more readability: you would access the value as `ParameterWithName.value` instead of `ParameterWithName.value.value`. Let me know if there is a reason we want a distinct`Parameter` type, I might have missed it in the proposal.

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Final storage format of parameters in the disk. This is made in the form of list so as to closely
+// resemble the JSON response in cache.
+message ParameterDatabase {
+
```
jcqli: nit: can remove this empty line

jcqli (changes requested):
Reviewing this now while in Draft since we discussed this in a meeting earlier. Mostly nits and changes for clarity.

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: This sounds better I guess. @vinitamurthi WDYT?

Arjupta: Also @jcqli the usage for current ParameterWithName will be like  `ParameterWithName.value.string` OR `ParameterWithName.value.boolean` OR `ParameterWithName.value.integer`

model/src/main/proto/platform_parameter.proto
vinitamurthi: +1 on this name being generic. I think PlatformParameter should be ok though because that is what we intend to call it in the rest of the feature

model/src/main/proto/platform_parameter.proto
vinitamurthi: I agree with Jacque, even if it is `ParameterWithName.value.string` and not `Parameter.value.value`, @jcqli 's suggestion reduces that nesting, and moreover hides the implementation detail well

vinitamurthi (changes requested):

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
```diff
@@ -6,32 +6,39 @@ option java_package = "org.oppia.android.app.model";
...
 // Storage format of platform parameters while the app is running.
-message ParameterMap{
-    map<string,Parameter> parameter = 1;
+message PlatformParameterMap{
+    map<string,PlatformParameter> parameter = 1;
 }
 
 // Pair of Platform Parameter Value and Name.
-message ParameterWithName{
+message PlatformParameterWithName{
```
jcqli: Ah I wonder if naming this `PlatformParameter` and naming the above `PlatformParameter` => `PlatformParamaterValue` or `PlatformParameterValueWrapper` might better represent these data objects. Since this object itself is identifying the parameter flag (both its name and its value) and the above proto defines the value type and abstracts it out for the map. This would look something like:
```
message PlatformParameterValueWrapper { oneof value {...} }

message PlatformParamterMap { map<string, PlatformParameterValueWrapper> }

message PlatformParamter { string name, oneof value {...} }
```

model/src/main/proto/platform_parameter.proto
```diff
@@ -6,32 +6,39 @@ option java_package = "org.oppia.android.app.model";
...
         // String type parameter value.
         string string = 4;
     }
 }
 
 // Storage format of platform parameters while the app is running.
-message ParameterMap{
-    map<string,Parameter> parameter = 1;
+message PlatformParameterMap{
+    map<string,PlatformParameter> parameter = 1;
```
jcqli: nit: space before PlatformParameter

jcqli (changes requested):
I now understand why we need the additional proto to wrap the parameter values -- so that the map can store all the platform parameters in the single database.

I see why the `PlatformParameterWithName` proto also had a `PlatformParameter` instead of `oneof {...}` -- for compatibility with the map and converting the value types.

I think there's a tradeoff here between data type compatibility and readability / efficiency. Both seem like reasonable choices, but I wanted to bring this up in case we wanted to discuss. LMK if you have questions about this.

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: @jcqli although I am not sure about the names but we will have to revert back to the original structures. Because while retrieving the values from the ParameterDatabase (in a domain level controller) we will get individual PlatfromParameterWithName type objects (which you are calling in the comment above as PlatformParameter). These objects will be needed to be stored in a PlatformParameterMap before using them. But PlatformParameterMap is map of String and PlatformParameter (which you are calling in the comment above as PlatformParameterValueWrapper). Hence we will be needed to first take out the value from PlatformParameterWithName and then make a PlatformParameter object out of it before getting mapped. This could have been avoided if we went back to previous structure of storing an Object of PlatformParameter in the PlatformParameterWithName Object. @vinitamurthi WDYT?

jcqli: Yup, this was the tradeoff I was thinking about with data type conversions. I think you're right that using the PlatformParameter proto would be better and streamline the conversion, though I think that you'll have to convert to / from these types later on anyways when receiving the JSON and when using the values.

Arjupta: Yeah agreed.

Arjupta:
@BenHenning PTAL on the current proto structure which is finalized after a discussion ( [comment1](https://github.com/oppia/oppia-android/pull/3269#discussion_r646769582) , [comment2](https://github.com/oppia/oppia-android/pull/3269#pullrequestreview-679872385))

jcqli (approved):
Looks good -- I think it's worth thinking about refactoring the name `PlatformParameterWithName` => `PlatformParameter` and updating the associated protos but the current names are clear enough where this isn't a blocking issue for me.

vinitamurthi (approved):

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// It will correspond to a single Platform Parameter Value.
+message PlatformParameter {
+    // Value of a parameter can be of any one of the following type.
```
BenHenning: Formatting seems off--we use 2 spaces for indentation in proto files (see others for reference).

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// It will correspond to a single Platform Parameter Value.
```
BenHenning: This should describe what the message represents in a fragment statement, and provide context around what the thing is:

``Corresponds to platform parameters that can be used to configure application behavior (including gating features).``

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+message PlatformParameter {
+    // Value of a parameter can be of any one of the following type.
+    oneof value {
```
BenHenning: Should be ``value_type`` per conventional use of oneofs elsewhere in the codebase.

model/src/main/proto/platform_parameter.proto
BenHenning: Just: ``Value of the parameter.`` is probably sufficient. The fact that it's a oneof implies that it can be one of the other types (or unspecified).

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+        // Boolean type parameter value.
```
BenHenning: This can probably be a bit more specific: ``Indicates a boolean-typed platform parameter.``

Ditto for the other 2.

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+        bool boolean = 2;
+        // Integer type parameter value.
+        int32 integer = 3;
+        // String type parameter value.
+        string string = 4;
+    }
+}
+
+// Storage format of platform parameters while the app is running.
+message PlatformParameterMap{
```
BenHenning: Nit here & below: space before ``{``.

I'm curious this & the indentation passed buf. @anandwana001 do we need to enable more strict checks for proto files?

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+    map<string, PlatformParameter> parameter = 1;
```
BenHenning: I suggest 'parameters' since it's a map. This is counter to protobuf best practices, but is inline with ours (happy to discuss why I prefer this approach if anyone disagrees (: ).

BenHenning: All proto fields should have documentation. In this case, we should explain what the map is mapping from & to.

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+        // Boolean type parameter value.
+        bool boolean = 2;
+        // Integer type parameter value.
+        int32 integer = 3;
+        // String type parameter value.
+        string string = 4;
+    }
+}
+
+// Storage format of platform parameters while the app is running.
```
BenHenning: Do we actually need this class if it's only for when the app is running? Why not use a Kotlin map in that case?

model/src/main/proto/platform_parameter.proto
BenHenning: Possibly a clearer name: ``PlatformParameters`` though I'm not sure since I'm a bit unclear how this type will be used. Can you provide a bit more context or a direct link to the part in your proposal that explains this?

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+message PlatformParameterMap{
+    map<string, PlatformParameter> parameter = 1;
+}
+
+// Pair of Platform Parameter Value and Name.
```
BenHenning: Why not always pair a name with platform parameters? They are expected to have one, and for compile-time only parameters we can have a special case for when the name is undefined (empty), e.g.:

```proto
// The name of the platform parameter, or empty if the parameter is only defined locally.
string name = 1;
...
```

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+message PlatformParameterWithName{
+    string name = 1;
+    PlatformParameter value = 2;
+}
+
+// Final storage format of parameters in the disk. This is made in the form of list so as to closely
+// resemble the JSON response in cache.
```
BenHenning: Nit: ``Format of platform parameters stored on disk.`` Is probably sufficient. For the second part, do we specifically benefit from matching the JSON response? It seems unimportant here (though using a 'repeated' still seems fine regardless, but I don't think we care about retaining any particular order--do we?).

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,37 @@
...
+message PlatformParameterDatabase {
+    // List of Parameters values with their names.
```
BenHenning: Or maybe: ``Platform parameters cached from the backend.`` might be clearer on the context.

Given that, it might also be worthwhile to name the overall structure ``RemotePlatformParameterDatabase`` since we actually have 2 types of parameters (local & remote), and it seems this is only storing the latter--is that correct?

BenHenning (changes requested):
Thanks @ARJUPTA. Left some thoughts.

Beyond the proto file, do we also need to introduce a new BUILD file target for this proto library?

BenHenning:
Also, per the email I sent out a few minutes ago, could you please add the missing task item number from the PR title once it's filed?

model/src/main/proto/platform_parameter.proto
Arjupta: Yeah a normal Map aould work too. /cc @vinitamurthi

model/src/main/proto/platform_parameter.proto
Arjupta: We didn't included the name of the Parameter in this proto class becuase when it was being used in the ParameterMap so we had a repetition of names, once as a key and second inside the object being stored as value. /cc @vinitamurthi

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done

model/src/main/proto/platform_parameter.proto
Arjupta: Done.
No we don't need that specifically I just mentioned it because I think by following this structure we are keeping ourselves open to consume Parameters in any Data structure we want (currently a Map). Also Yes, this format will retain the order of parameters from the JSON.

model/src/main/proto/platform_parameter.proto
Arjupta: That's True that it is stoting the Remote Parameters. Done

model/src/main/proto/platform_parameter.proto
Arjupta: Not an issue now as We will replace it with a normal Kotlin Map

model/src/main/proto/platform_parameter.proto
Arjupta: Not an issue now as We will replace it with a normal Kotlin Map

model/src/main/proto/platform_parameter.proto
Arjupta: Not an issue now as We will replace it with a normal Kotlin Map

Arjupta: Not an issue now as We will replace it with a normal Kotlin Map

Arjupta:
@vinitamurthi PTAL at the changes.

vinitamurthi:
Hi @ARJUPTA I think you need Ben to take a look, so once you make the changes, please assign the PR to Ben

model/src/main/proto/platform_parameter.proto
BenHenning: Ah I see. How important is the map? An alternate approach might be just to store a repeated list of the parameters and build the map at runtime.

model/src/main/proto/platform_parameter.proto
BenHenning: Got it. I think this also helps address the way to store it, too

Arjupta: Yeah

model/src/main/proto/platform_parameter.proto
Arjupta: Actually the map is formed only at the runtime. It has a keyset that corresponds to all the platform names. The value set of this map contains PlatformParameter objects which currently wrap only the value of a particular parameter. But if we would have also stored the names inside PlatformParameter objects then there would have been unnecessary repititon of the names in the key-value pairs of map. We avoided this problem by using a complete new proto (PlatformParameterWithName) class which contains both the name and value of Platform Parameters.

fsharpasharp (approved):
BUILD.bazel change looks good to me.

model/BUILD.bazel
```diff
@@ -223,6 +223,18 @@ java_lite_proto_library(
     deps = [":exploration_proto"],
 )
 
+format_import_proto_library(
+    name = "platform_parameter",
+    src = "src/main/proto/platform_parameter.proto",
+    visibility = ["//:oppia_api_visibility"],
```
fsharpasharp: You can omit this visibility since it will be used by java_lite_proto_library in this same package.

Arjupta: Done

model/src/main/proto/platform_parameter.proto
```diff
@@ -0,0 +1,31 @@
...
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Corresponds to platform parameters that can be used to configure application behavior (including gating features).
+message PlatformParameter {
+  // Value of the parameter.
+  oneof value_type {
+      // Indicates a boolean-typed platform parameter.
```
BenHenning: Proto files are indented with 2 spaces, not 4--please fix.

model/src/main/proto/platform_parameter.proto
BenHenning: Hmm, I think I might be missing something. Is this the compile-time name of the parameter, or the server name? If the former, it seems wrong not to include it in the PlatformParameter since that is, conceptually, a parameter. I see the point that it would be duplicated, but I don't see that it's an actual issue here. Having access to the parameter name in PlatformParameter seems like a good design, and to some extent expected. PlatformParameter providing a value without the context of the name is a bit confusing.

Or am I missing some key design aspect of this?

@vinitamurthi in case you have thoughts.

BenHenning (commented):
Thanks @ARJUPTA. Apologies for the delayed review. Following up with a few more comments.

model/src/main/proto/platform_parameter.proto
Arjupta: Done. Its just because of default formatting that is done by andorid studio

model/src/main/proto/platform_parameter.proto
Arjupta: These are simply Parameter names as we are not using different Proto structure based on whether it is a compile time or runtime parameter.

vinitamurthi: I think its okay if we add the name in the internal message. We would be duplicating the name but I think that's not a big issue

BenHenning (approved):
Thanks @ARJUPTA! Apologies for the delayed reviews--this looks really good.

BenHenning:
The Robolectric failure seems due to #2032. Restarting those. The build test failure seems like a flake in one of the dependent actions--restarting that as well. Both of these failures seem independent of the changes done here.

Issue #2391, opened 2021-01-05
FAQs crashes in RTL(psuedolocale) 


FareesHussain:
**Describe the bug**
FAQ crashes in [Psuedolocale](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines)  

**To Reproduce**
1. Set language to [Psuedolocale](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines)  
2. Open 'Navigation Drawer'
3. Click on 'Help'
4. Open FAQs
5. Click on any of the following questions
    - How can I create a new profile?
    - How can I delete a profile?
    - Why is the Exploration player not loading?
    - Why is my audio not playing?
6. See the crash

**Expected behavior**
Get FAQs with RTL supported

**Additional context**
Error message
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.oppia.android, PID: 14209
    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.oppia.android/org.oppia.android.app.help.faq.faqsingle.FAQSingleActivity}: java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary (405 follows ‮)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3270)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3409)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2016)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary (405 follows ‮)
        at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:693)
[...]

PR #3320, opened 2021-06-12
Fix #2193, #2391: [RTL] High-fi Faq Bullet points

### LINKED ISSUE #2193: [RTL] High-fi Bullet Points
Custom Bullet Points needs RTL support as shown in below screen.
FAQ Single RTL

Make sure that this works correctly in mobile (landscape + portrait) and tablet (landscape + portrait).

## How to Test
1. Go to `Help` in `Navigation Drawer`
2. Click on `FAQs`
3. Click on `1st` FAQ
4. See the bullets.

## File Reference:

https://github.com/oppia/oppia-android/blob/9221c80f33d94c38a69bf055dbde5d3138ec5ada/utility/src/main/java/org/oppia/android/util/parser/CustomBulletSpan.kt#L20

https://github.com/oppia/oppia-android/blob/9221c80f33d94c38a69bf055dbde5d3138ec5ada/utility/src/main/java/org/oppia/android/util/parser/HtmlParser.kt#L92

veena14cs:
## Explanation
This PR fixes #2193 #2391 . This PR fixes RTL support for Bullet points in FAQ screen. It also fixes the crash on opening of Faq Answer screen.

** Following  Issues are resolved **

- Gap after new line in RTL is fixed by replacing `<p>` with `<span>`
- To fix the crash which I suspected was probably because of custom oppia tag, where we are replacing `<li>` tag with `<oppia-li>` tag. The actual error was due to missing of line break `<br>` we need to have after every bullet closing tag.
- As per @anandwana001 [comment](https://github.com/oppia/oppia-android/issues/2193#issuecomment-742226970) I tried modifying `strings.xml` for rtl with `<li dir=rtl>` tag to support rtl.

## Screenshot LTR RTL
Mobile:
screenshot  ......Screenshot_1627979944

Screenshot_1627981323.......Screenshot_1627979953

Tablet

Screenshot_1627980215......Screenshot_1627980273

Screenshot_1627980221......Screenshot_1627980279

## Screenshots of Test with RTL and without RTL code

Screenshot (3)
Screenshot (4)

## Checklist


rt4914 (changes requested):
@veena14cs We need custom oppia li tag because the UI looks more correct after that. If we use only `li` tag it looks bad. Therefore we will need to use our custom implementation in LTR and RTL both.

anandwana001:
Agree with @rt4914 comment.
Assig me back once done

veena14cs:
@rt4914 and @anandwana001 PTAL. There is one issue I am facing here. Faq crash for RTL is fixed and also it is showing bullets. But its showing double spacing between the every new line when device is in RTL. I am not able to figure out this. Could please check what could be the problem.

anandwana001:
@rt4914 @FareesHussain I think you might have faced a similar issue earlier, could you help here?

FareesHussain:
> @rt4914 and @anandwana001 PTAL. There is one issue I am facing here. Faq crash for RTL is fixed and also it is showing bullets. But its showing double spacing between the every new line when device is in RTL. I am not able to figure out this. Could please check what could be the problem.

@veena14cs Can you add the screenshot here

veena14cs:
> > @rt4914 and @anandwana001 PTAL. There is one issue I am facing here. Faq crash for RTL is fixed and also it is showing bullets. But its showing double spacing between the every new line when device is in RTL. I am not able to figure out this. Could please check what could be the problem.
> 
> @veena14cs Can you add the screenshot here

@FareesHussain Here is the screenhot for RTL

Screenshot_1627302732

FareesHussain:
My bad I didn't notice the screenshots in the description earlier.
I did face this extra spacing when I filed this issue https://github.com/oppia/oppia-android/issues/2391

I'll check if I can find a solution for this

veena14cs:
> My bad I didn't notice the screenshots in the description earlier.
> I did face this extra spacing when I filed this issue #2391
> 
> I'll check if I can find a solution for this

@FareesHussain any update on this?

FareesHussain:
> > My bad I didn't notice the screenshots in the description earlier.
> > I did face this extra spacing when I filed this issue #2391
> > I'll check if I can find a solution for this
> 
> @FareesHussain any update on this?

I tried looking into this but I was unable to figure it out.

app/src/main/res/values/faqs.xml
```diff
@@ -9,13 +9,13 @@
...
+  <string name="faq_answer_2"><![CDATA[<p>Once a profile is deleted:</p> <p><br></p> <p><li dir="rtl> The profile cannot be recovered. </li> </p> <p><li dir="rtl> Profile information such as name, photos, and progress will be permanently deleted. </li> </p> <p><br></p> <p>To delete a profile (excluding the <u>Administrator\'s</u>):</p> <p>1. From the Administrator\'s Home Page, tap on the menu button on the top left.</p> <p>2. Tap on <strong>Administrator Controls</strong>.</p> <p>3. Tap on <strong>Edit Profiles</strong>.</p> <p>4. Tap on the Profile you would like to delete.</p> <p>5. At the bottom of the screen, tap <strong>Profile Deletion</strong>.</p> <p>6. Tap <strong>Delete</strong> to confirm deletion.</p><p><br></p><p>Note: Only the <u>Administrator</u> is able to manage profiles.</p>]]></string>
```
rt4914: @BenHenning Can you confirm if this approach is correct or not?

veena14cs:
> > > My bad I didn't notice the screenshots in the description earlier.
> > > I did face this extra spacing when I filed this issue #2391
> > > I'll check if I can find a solution for this
> > 
> > 
> > @FareesHussain any update on this?
> 
> I tried looking into this but I was unable to figure it out.

Thanks @FareesHussain. Will look into it.

veena14cs:
> > I tried looking into this but I was unable to figure it out.


@FareesHussain I figured out that `<p>` is giving extra spacing due to which there was double spacing after and before every new paragraph tag. I replaced this tag with `<span>` and used line breaks `<br/>` to give new line as `<span>` tag was not adding any spacing.

app/src/main/res/values/faqs.xml
veena14cs: @BenHenning can you please confirm on this as it will unblock me.

app/src/main/res/values/faqs.xml
```diff
@@ -9,13 +9,13 @@
...
+  <string name="faq_answer_2"><![CDATA[ <span> Once a profile is deleted: </span><ul><li dir="rtl"> The profile cannot be recovered. </li><li dir="rtl"> Profile information such as name, photos, and progress will be permanently deleted. </li> </ul> <span>To delete a profile (excluding the <u>Administrator\'s</u>): </span> <br/><br/> <span>1. From the Administrator\'s Home Page, tap on the menu button on the top left. </span> <br/><br/> <span>2. Tap on <strong>Administrator Controls</strong>. </span> <br/><br/> <span>3. Tap on <strong>Edit Profiles</strong>. </span> <br/><br/> <span>4. Tap on the Profile you would like to delete. </span> <br/><br/> <span>5. At the bottom of the screen, tap <strong>Profile Deletion</strong>. </span> <br/><br/> <span>6. Tap <strong>Delete</strong> to confirm deletion. </span><br/> <span>Note: Only the <u>Administrator</u> is able to manage profiles. </span>]]></string>
```
BenHenning: Might it be possible for us to automatically add the rtl directionality via HtmlParser? One concern I have with this is it may be lost across translations once these strings go through Translatewiki.

utility/src/main/java/org/oppia/android/util/parser/html/BulletTagHandler.kt
```diff
@@ -23,10 +23,17 @@ class BulletTagHandler : CustomHtmlContentHandler.CustomTagHandler {
   override fun handleClosingTag(output: Editable) {
     output.append("\n")
     output.getSpans(0, output.length, Bullet::class.java).lastOrNull()?.let {
+      val spanEnd = output.getSpanEnd(it)
       val start = output.getSpanStart(it)
-      output.removeSpan(it)
+      // no need for newline if empty span. This fix 'PARAGRAPH span must start at paragraph boundary'.
+      if (start == spanEnd) {
+        output.removeSpan(it)
+      }
```
BenHenning: When do we encounter an empty span?

utility/src/main/java/org/oppia/android/util/parser/html/BulletTagHandler.kt
```diff
@@ -23,10 +23,17 @@ class BulletTagHandler : CustomHtmlContentHandler.CustomTagHandler {
   override fun handleClosingTag(output: Editable) {
     output.append("\n")
     output.getSpans(0, output.length, Bullet::class.java).lastOrNull()?.let {
+      val spanEnd = output.getSpanEnd(it)
       val start = output.getSpanStart(it)
-      output.removeSpan(it)
+      // no need for newline if empty span. This fix 'PARAGRAPH span must start at paragraph boundary'.
```
BenHenning: I'm not sure I fully understand this comment--could you please rephrase it?

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
```diff
@@ -19,69 +15,7 @@ import org.oppia.android.util.R
...
-      }
-
-      paint.style = style
-    }
-  }
-
+class CustomBulletSpan {
   /**
    * Update and return the [SpannableStringBuilder] by replacing all the [BulletSpan]s in the
    * [SpannableStringBuilder] with [CustomBulletSpan].
```
BenHenning: Needs an update--there's no longer a CustomBulletSpan anymore. Ditto elsewhere.

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
```diff
@@ -91,6 +25,12 @@ class CustomBulletSpan(context: Context) : LeadingMarginSpan {
       spannableStringBuilder: SpannableStringBuilder,
       context: Context
     ): SpannableStringBuilder {
+      val resources = context.resources
+      val bulletRadius = resources.getDimensionPixelSize(R.dimen.bullet_radius)
+      val gapWidth = resources.getDimensionPixelSize(R.dimen.bullet_gap_width)
+      /** The space between the start of the line and the bullet. */
```
BenHenning:
```suggestion
      // The space between the start of the line and the bullet.
```

This shouldn't be a KDoc.

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
```diff
@@ -100,12 +40,17 @@ class CustomBulletSpan(context: Context) : LeadingMarginSpan {
...
-          start,
-          end,
-          Spanned.SPAN_INCLUSIVE_EXCLUSIVE
+          LeadingMarginSpan.Standard(spacingBeforeBullet), start, end,
+          Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
+        )
+        spannableStringBuilder.setSpan(
+          BulletSpan(
+            gapWidth, Color.BLACK, bulletRadius
+          ),
```
BenHenning: Can this fit on one line?

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
```diff
@@ -19,69 +15,7 @@ import org.oppia.android.util.R
...
-  private val bulletRadius = resources.getDimensionPixelSize(R.dimen.bullet_radius)
-  private val gapWidth = resources.getDimensionPixelSize(R.dimen.bullet_gap_width)
-  private val yOffset = resources.getDimensionPixelSize(R.dimen.bullet_y_offset)
-  /** The space between the start of the line and the bullet. */
-  private val spacingBeforeBullet = resources.getDimensionPixelSize(R.dimen.spacing_before_bullet)
-  /** The space between the bullet and the text. */
-  private val spacingBeforeText = resources.getDimensionPixelSize(R.dimen.spacing_before_text)
-  /** The total spacing between the start of the line and the text. */
-  private val totalSpacingToTextStart = spacingBeforeBullet + spacingBeforeText + bulletRadius * 2
-  private var bulletPath: Path? = null
```
BenHenning, line 33: With these deletions, are there any resource dimensions that we can now remove?

app/src/main/res/values/faqs.xml
```diff
@@ -9,13 +9,13 @@
...
+  <string name="faq_answer_8"><![CDATA[ <span> To download an Exploration: </span> <br/><br/> <span>1. From the Home Page, tap on a Topic or Exploration. </span> <br/><br/> <span>2. From that Topic Page, tap on the <strong>Info</strong> tab. </span> <br/><br/> <span>3. Tap on <strong>Download Topic</strong>.  </span> <br/><br/> <span>4. Depending on your app settings, you may need Administrator approval or stable Wifi connection to complete your download. If needed, once these requirements are satisfied, the Topic has been downloaded onto the device and can be used offline by all profiles.  </span>]]></string>
```
BenHenning: I'm not sure that I understand these changes--why are we switching from paragraphs to spans & breaks?

BenHenning (changes requested):
Thanks @veena14cs. This is a really nice potential simplification. I have some questions around the faq changes to better understand them. Also, could we add tests to verify these fixes?

Finally, please add a before screenshot for RTL/non-RTL spacing so that we can compare (I'm not sure what to look for with your known issues list without an image to compare to).

app/src/main/res/values/faqs.xml
BenHenning: Followed up in a few comments below.

utility/src/main/java/org/oppia/android/util/parser/html/BulletTagHandler.kt
veena14cs: If there were empty tags or if there is no end tag. It was crashing with error `PARAGRAPH span must start at paragraph boundary`. But, now as I have formatted the faqs.xml file there are no any empty tags and paragraph tags which would fail here. This is no longer needed so removing it.

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
veena14cs: I did not get this. What needs to be updated.?

utility/src/main/java/org/oppia/android/util/parser/html/BulletTagHandler.kt
veena14cs: Followed up in comment below.

app/src/main/res/values/faqs.xml
veena14cs: `<p>` has extra spacing due to which there was double spacing after and before every new paragraph tag. I replaced this tag with `<span>` and used line breaks `<br/>` to give new line, as `<span>` tag was not adding any spacing.

You can check the screenshot below that used paragraphs tag.

Screenshot_1627302732

Screenshot with spans and break tags.
Screenshot_1627979944

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
veena14cs: Removed unused dimensions.

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
veena14cs: Done.

veena14cs:
> Finally, please add a before screenshot for RTL/non-RTL spacing so that we can compare (I'm not sure what to look for with your known issues list without an image to compare to).

The issue is been resolved I have updated in PR description. I have followed up with below comments and screenshot for the same.

app/src/main/res/values/faqs.xml
veena14cs: Do you mean like as below
```
htmlContent = htmlContent.replace("<li>", "<$CUSTOM_BULLET_LIST_TAG dir=\"rtl\">")
          .replace("</li>", "</$CUSTOM_BULLET_LIST_TAG>")
```
This crashes with runtime exception.

veena14cs:
> Thanks @veena14cs. This is a really nice potential simplification. I have some questions around the faq changes to better understand them. Also, could we add tests to verify these fixes?

What tests I have to add? As its not straight forward in terms what has to be tested.

app/src/main/res/values/faqs.xml
BenHenning: What's the exception? Without that context it's hard to know whether the solution can be made to work.

app/src/main/res/values/faqs.xml
BenHenning: Do we have an understanding of why ``<p>`` introduces extra spacing in RTL mode? I think we need to understand that to see if there's a way to make the current formatting work before changing it. ``span`` doesn't actually introduce any visual changes to the document (or at least, it's not meant to unless there is styling to define it) unlike a paragraph or div. I want to better understand how paragraph is meant to behave and why it changes in RTL vs. LTR since that will provide us guidance with how to properly format this.

Do you have a specific page or reference that led you to choosing span here?

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
BenHenning: There's a reference here to a class that no longer exists: ``CustomBulletSpan``

BenHenning (commented):
Thanks @veena14cs. Following up with some questions.

Regarding the tests, we should make sure that the bullet list items have the RTL property set since we're primarily relying on the system to do the right thing. We can't verify visual correctness without screenshot testing, so this is probably a reasonable midway solution.

app/src/main/res/values/faqs.xml
veena14cs: Yes here is the link https://stackoverflow.com/questions/6874402/remove-space-above-and-below-p-tag-html and 
 https://www.w3schools.com/tags/tag_p.asp

app/src/main/res/values/faqs.xml
veena14cs: `java.lang.RuntimeException: PARAGRAPH span must end at paragraph boundary`

veena14cs:
@BenHenning I have addressed some of your comments PTAL.

app/src/main/res/values/faqs.xml
BenHenning: @veena14cs when providing an exception, you should always provide the full stack trace & error since the error alone doesn't provide the full context.

Also, you should provide context on why this exception can't be worked around. The replace solution is still ideal, and we should try to make it work unless it's clear that & why it won't work (simply running into an exception doesn't tell us this--that could indicate something else is being set up incorrectly).

app/src/main/res/values/faqs.xml
BenHenning: That doesn't necessarily explain why this happens in Android, though. Do we know why paragraphs have padding/margins on Android?

BenHenning (commented):
Hi @veena14cs. Followed up.

veena14cs:
> Hi @veena14cs. Followed up.

I used trim method and other solutions to remove extra spacing [reference](https://github.com/oppia/oppia-android/pull/3320#discussion_r685536329) but it didn't work out.

utility/src/main/java/org/oppia/android/util/parser/html/CustomBulletSpan.kt
veena14cs: Removed.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -87,13 +89,67 @@ class HtmlParser private constructor(
...
+      )
+    }
+    return spannableBuilder
+  }
+
+  private fun setTextDirection(htmlContentTextView: TextView, htmlContent: String): String {
+    when (getLayoutDirection(htmlContentTextView)) {
+      ViewCompat.LAYOUT_DIRECTION_RTL -> {
+        htmlContentTextView.textDirection = View.TEXT_DIRECTION_ANY_RTL
+        if ("<li>" in htmlContent) {
```
BenHenning, line 137: Here & below: is the ``if`` check needed, or can we just perform the replacement?

utility/src/test/java/org/oppia/android/util/parser/html/CustomHtmlContentHandlerTest.kt
```diff
@@ -148,9 +149,26 @@ class CustomHtmlContentHandlerTest {
     assertThat(parsedHtml.toString()).isEqualTo("some other content")
   }
 
-  @Test
+  @Test @Config(qualifiers = "ldltr")
```
BenHenning: Newline before the ``@Config``.

Ditto below.

app/src/sharedTest/java/org/oppia/android/app/parser/BulletSpanTest.kt
```diff
@@ -65,13 +67,21 @@ import javax.inject.Singleton
 
 @RunWith(AndroidJUnit4::class)
 @LooperMode(LooperMode.Mode.PAUSED)
-@Config(application = CustomBulletSpanTest.TestApplication::class, qualifiers = "port-xxhdpi")
-class CustomBulletSpanTest {
+@Config(application = BulletSpanTest.TestApplication::class, qualifiers = "port-xxhdpi")
+class BulletSpanTest {
```
BenHenning: We shouldn't actually be testing Android's bullet span--we can assume that's correct.

Instead, we should be testing that the bullet span code (now in HtmlParser) sets up bullets correctly. Could you remove this suite & move relevant tests over to HtmlParserTest such that those properly cover the changed behaviors in HtmlParser?

app/src/sharedTest/java/org/oppia/android/app/parser/HtmlParserTest.kt
```diff
@@ -219,31 +230,93 @@ class HtmlParserTest {
...
+    assertThat(bulletSpan0Margin.getLeadingMargin(true)).isEqualTo(72)
 
-    val bulletSpan1 = bulletSpans[1] as CustomBulletSpan
+    val bulletSpan1 = bulletSpans[1]
     assertThat(bulletSpan1).isNotNull()
+    assertThat(htmlResult.toString())
+      .isEqualTo(
+        "You should know the following before going on:\n\n" + "The counting numbers" +
+          " (1, 2, 3, 4, 5 ….)\n\n" + "How to tell whether one counting number is bigger or " +
+          "smaller than another"
```
BenHenning: We should only be doing string concatenation when needing to line break. The between-string concatenation isn't needed--please remove them and just have the strings fully defined per-line/not broken up except when they need to be wrapped.

BenHenning (changes requested):
Thanks @veena14cs. Can you please add screenshots to the opening comment also showing:
- The new RTL-specific tests failing without the fixes in place
- The before screenshots in addition to the current (after) screenshots to show how this fixes the issue

> > Hi @veena14cs. Followed up.
> 
> I used trim method and other solutions to remove extra spacing [reference](https://github.com/oppia/oppia-android/pull/3320#discussion_r685536329) but it didn't work out.

I'm actually looking for an explanation of why this solution is the correct one beyond "we found it works." We should always be making changes intentionally by understanding why another solution is insufficient, and then be able to justify why the fix is the best approach given the information we've found. I'm not convinced yet that the solution proposed in this PR is the correct one because I haven't seen a clear explanation for why we have to use spans vs. paragraphs.

I also want to clarify my thinking slightly: paragraphs seem like they're the correct semantic representation to use here per https://stackoverflow.com/a/1908251, so I think we need a strong justification for moving toward spans with an idea on how they affect these other considerations (semantic meaning, screenreader support, etc.).

utility/src/test/java/org/oppia/android/util/parser/html/CustomHtmlContentHandlerTest.kt
veena14cs: Done.

app/src/sharedTest/java/org/oppia/android/app/parser/HtmlParserTest.kt
veena14cs: Done.

app/src/sharedTest/java/org/oppia/android/app/parser/BulletSpanTest.kt
veena14cs: Done.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
veena14cs: We need it as we need to replace for `<li>` tags only.

veena14cs:
> > I used trim method and other solutions to remove extra spacing [reference](https://github.com/oppia/oppia-android/pull/3320#discussion_r685536329) but it didn't work out.
> 
> I'm actually looking for an explanation of why this solution is the correct one beyond "we found it works." We should always be making changes intentionally by understanding why another solution is insufficient, and then be able to justify why the fix is the best approach given the information we've found. I'm not convinced yet that the solution proposed in this PR is the correct one because I haven't seen a clear explanation for why we have to use spans vs. paragraphs.
> 
> I also want to clarify my thinking slightly: paragraphs seem like they're the correct semantic representation to use here per https://stackoverflow.com/a/1908251, so I think we need a strong justification for moving toward spans with an idea on how they affect these other considerations (semantic meaning, screenreader support, etc.).

I found out that the flag which we are using[ `FROM_HTML_MODE_LEGACY`  ](https://developer.android.com/reference/android/text/Html.html#FROM_HTML_MODE_LEGACY) it adds 2 new lines. By this I think we either need to change this to `FROM_HTML_MODE_COMPACT`  or `FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH` here you can see the screenshots for each of the flags https://medium.com/@yair.kukielka/html-tags-in-android-nougat-textview-9f3554d7b705.

Please let me know your suggestions.

Edit: This is only when device is changed to RTL with` <p>`.  I am not getting the exact bug. When I use` <span>` it is same for both LTR and RTL there no spacing issue.

 **Screenshot with `<p> `and Html flag set to FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH**

`   <p> If it is your first time creating a profile and not have a PIN:  </p>  <p> 1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>.  </p>  <p> 2. Create a PIN and <strong>Save</strong>.  </p>  <p> 3. Fill in all fields for the profile.  </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul>  <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! </p> <p> If you have created a profile before and have a PIN:  </p>  <p> 1. From the Profile Chooser, tap on <strong>Add Profile</strong>.  </p>  <p> 2. Enter your PIN and tap <strong>Submit</strong>.  </p>  <p> 3. Fill in all fields for the profile.  </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul>  <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser!   Note: Only the <u>Administrator</u> is able to manage profiles. </p>      `


Screenshot_1628864248...Screenshot_1628864287

BenHenning:
Will need to look at this in a couple of days--apologies.

veena14cs:
> Will need to look at this in a couple of days--apologies.

@BenHenning can you please have a look at this PR today?

BenHenning:
> > > I used trim method and other solutions to remove extra spacing [reference](https://github.com/oppia/oppia-android/pull/3320#discussion_r685536329) but it didn't work out.
> > 
> > 
> > I'm actually looking for an explanation of why this solution is the correct one beyond "we found it works." We should always be making changes intentionally by understanding why another solution is insufficient, and then be able to justify why the fix is the best approach given the information we've found. I'm not convinced yet that the solution proposed in this PR is the correct one because I haven't seen a clear explanation for why we have to use spans vs. paragraphs.
> > I also want to clarify my thinking slightly: paragraphs seem like they're the correct semantic representation to use here per https://stackoverflow.com/a/1908251, so I think we need a strong justification for moving toward spans with an idea on how they affect these other considerations (semantic meaning, screenreader support, etc.).
> 
> I found out that the flag which we are using[ `FROM_HTML_MODE_LEGACY` ](https://developer.android.com/reference/android/text/Html.html#FROM_HTML_MODE_LEGACY) it adds 2 new lines. By this I think we either need to change this to `FROM_HTML_MODE_COMPACT` or `FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH` here you can see the screenshots for each of the flags https://medium.com/@yair.kukielka/html-tags-in-android-nougat-textview-9f3554d7b705.
> 
> Please let me know your suggestions.
> 
> Edit: This is only when device is changed to RTL with` <p>`. I am not getting the exact bug. When I use` <span>` it is same for both LTR and RTL there no spacing issue.
> 
> **Screenshot with `<p> `and Html flag set to FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH**
> 
> ` <p> If it is your first time creating a profile and not have a PIN: </p> <p> 1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>. </p> <p> 2. Create a PIN and <strong>Save</strong>. </p> <p> 3. Fill in all fields for the profile. </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul> <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! </p> <p> If you have created a profile before and have a PIN: </p> <p> 1. From the Profile Chooser, tap on <strong>Add Profile</strong>. </p> <p> 2. Enter your PIN and tap <strong>Submit</strong>. </p> <p> 3. Fill in all fields for the profile. </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul> <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! Note: Only the <u>Administrator</u> is able to manage profiles. </p> `
> 
> Screenshot_1628864248...Screenshot_1628864287

Interesting. I suspect that there's some weird attribute somewhere in our styles (likely inherited) that's causing this spacing inconsistency across LTR/RTL, but it's definitely a problem & will be hard to track down where it's coming from.

What happens when we use FROM_HTML_MODE_COMPACT instead of LEGACY? Is it then consistent between RTL & LTR?

Also, I'm not sure I understand why we're even using paragraphs for the list. Could you try removing the ``<p>`` elements entirely for the list items and instead just use them for top-level sections of text in each FAQ?

veena14cs:
> Interesting. I suspect that there's some weird attribute somewhere in our styles (likely inherited) that's causing this spacing inconsistency across LTR/RTL, but it's definitely a problem & will be hard to track down where it's coming from.
> 
> What happens when we use FROM_HTML_MODE_COMPACT instead of LEGACY? Is it then consistent between RTL & LTR?
> 
> Also, I'm not sure I understand why we're even using paragraphs for the list. Could you try removing the `<p>` elements entirely for the list items and instead just use them for top-level sections of text in each FAQ?


`<p>` tag is been used only at the top level and the numbered list which is been numbered in strings itself and not using any tags for numbered list. 

I have added screenshots for each of the flags. As you can see every RTL screenshot has spacing. I have checked for the attributes in textviews but it doesn't have any specific thing that is causing the spacing. 
This is something weird with `<p>` which I am not understanding. If I use `<span`> with line breaks in the string, the behavior is same for both LTR and RTL.

Screenshot **FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH**

Screenshot_1629284940......Screenshot_1629285066

Screenshot **FROM_HTML_MODE_COMPACT**
Screenshot_1629285279.....Screenshot_1629285243

Screenshot **FROM_HTML_MODE_LEGACY**
Screenshot_1629285380......Screenshot_1629285418

BenHenning:
> FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH

Thanks for the screenshots @veena14cs --they're really helpfu.

One thing I'm confused about: your screenshot in https://github.com/oppia/oppia-android/pull/3320#issuecomment-898474890 doesn't match https://github.com/oppia/oppia-android/pull/3320#issuecomment-901035109 despite both using FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and ``<p>`` tags. Do you know why this inconsistency exists?

I'm asking because the first screenshot actually looks workable if we moved the numbered items to a proper numbered list (or just kept them within a single paragraph rather than having per-item paragraph tags). The first screenshot seems to only have extra spacing for paragraphs and not bullets which is probably fine then.

I'm not keen on the inconsistency and we should definitely fix it, but I think we can live with a small difference. The spacing between the numbers isn't good, though.

veena14cs:
> > FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH
> 
> Thanks for the screenshots @veena14cs --they're really helpfu.
> 
> One thing I'm confused about: your screenshot in [#3320 (comment)](https://github.com/oppia/oppia-android/pull/3320#issuecomment-898474890) doesn't match [#3320 (comment)](https://github.com/oppia/oppia-android/pull/3320#issuecomment-901035109) despite both using FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and `<p>` tags. Do you know why this inconsistency exists?
> 
> I'm asking because the first screenshot actually looks workable if we moved the numbered items to a proper numbered list (or just kept them within a single paragraph rather than having per-item paragraph tags). The first screenshot seems to only have extra spacing for paragraphs and not bullets which is probably fine then.
> 
> I'm not keen on the inconsistency and we should definitely fix it, but I think we can live with a small difference. The spacing between the numbers isn't good, though.

Screenshots with single paragraph tag between numbered list with the flag FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH
Screenshot_1629405631......Screenshot_1629405568


Faq text

`<p> If it is your first time creating a profile and not have a PIN:  </p>  <p> 1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>.   <br/>2. Create a PIN and <strong>Save</strong>.  <br/>3. Fill in all fields for the profile.  </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul>  <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! <br/> If you have created a profile before and have a PIN:  </p> <br/><p> 1. From the Profile Chooser, tap on <strong>Add Profile</strong>.  <br/> 2. Enter your PIN and tap <strong>Submit</strong>.   <br/> 3. Fill in all fields for the profile.  </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul>  <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser!   Note: Only the <u>Administrator</u> is able to manage profiles. </p>`

BenHenning:
> > > FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH
> > 
> > 
> > Thanks for the screenshots @veena14cs --they're really helpfu.
> > One thing I'm confused about: your screenshot in [#3320 (comment)](https://github.com/oppia/oppia-android/pull/3320#issuecomment-898474890) doesn't match [#3320 (comment)](https://github.com/oppia/oppia-android/pull/3320#issuecomment-901035109) despite both using FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and `<p>` tags. Do you know why this inconsistency exists?
> > I'm asking because the first screenshot actually looks workable if we moved the numbered items to a proper numbered list (or just kept them within a single paragraph rather than having per-item paragraph tags). The first screenshot seems to only have extra spacing for paragraphs and not bullets which is probably fine then.
> > I'm not keen on the inconsistency and we should definitely fix it, but I think we can live with a small difference. The spacing between the numbers isn't good, though.
> 
> Screenshots with single paragraph tag between numbered list with the flag FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH
> Screenshot_1629405631......Screenshot_1629405568
> 
> Faq text
> 
> `<p> If it is your first time creating a profile and not have a PIN: </p> <p> 1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>. <br/>2. Create a PIN and <strong>Save</strong>. <br/>3. Fill in all fields for the profile. </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul> <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! <br/> If you have created a profile before and have a PIN: </p> <br/><p> 1. From the Profile Chooser, tap on <strong>Add Profile</strong>. <br/> 2. Enter your PIN and tap <strong>Submit</strong>. <br/> 3. Fill in all fields for the profile. </p> <ul><li> (Optional) Upload a photo. <br></li><li> Enter a name. <br></li><li> (Optional) Assign a 3-digit PIN. <br></li> </ul> <p> 4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! Note: Only the <u>Administrator</u> is able to manage profiles. </p>`

I actually think this looks pretty good minus the spacing between the numbered list & bullets. Contextually, the bullets should actually be part of the numbered list. Coould we maybe reformat this to:

```
<p>If it is your first time creating a profile and not have a PIN:
  <br />1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>.
  <br />2. Create a PIN and <strong>Save</strong>.
  <br />3. Fill in all fields for the profile. </p>
  <ul>
    <li>(Optional) Upload a photo.</li>
    <li>Enter a name.</li>
    <li>(Optional) Assign a 3-digit PIN.</li>
  </ul>
  <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser!
</p>
<br/>
<p>If you have created a profile before and have a PIN:
  <br />1. From the Profile Chooser, tap on <strong>Add Profile</strong>.
  <br />2. Enter your PIN and tap <strong>Submit</strong>.
  <br />3. Fill in all fields for the profile.
  <ul>
    <li>(Optional) Upload a photo.</li>
    <li>Enter a name.</li>
    <li>(Optional) Assign a 3-digit PIN.</li>
  </ul>
  <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! Note: Only the <u>Administrator</u> is able to manage profiles.
</p>
```

(Also note that I'm using multiple lines to make it more readable; I don't think this will have an impact on the final formatting).

Interestingly, I noticed extra breaks in the bullet list that probably shouldn't have been there.

Can you try out this version on RTL & LTR with FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and add screenshots for it?

veena14cs:
> ```
> <p>If it is your first time creating a profile and not have a PIN:
>   <br />1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>.
>   <br />2. Create a PIN and <strong>Save</strong>.
>   <br />3. Fill in all fields for the profile. </p>
>   <ul>
>     <li>(Optional) Upload a photo.</li>
>     <li>Enter a name.</li>
>     <li>(Optional) Assign a 3-digit PIN.</li>
>   </ul>
>   <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser!
> </p>
> <br/>
> <p>If you have created a profile before and have a PIN:
>   <br />1. From the Profile Chooser, tap on <strong>Add Profile</strong>.
>   <br />2. Enter your PIN and tap <strong>Submit</strong>.
>   <br />3. Fill in all fields for the profile.
>   <ul>
>     <li>(Optional) Upload a photo.</li>
>     <li>Enter a name.</li>
>     <li>(Optional) Assign a 3-digit PIN.</li>
>   </ul>
>   <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! Note: Only the <u>Administrator</u> is able to manage profiles.
> </p>
> ```
> 
> (Also note that I'm using multiple lines to make it more readable; I don't think this will have an impact on the final formatting).
> 
> Interestingly, I noticed extra breaks in the bullet list that probably shouldn't have been there.
> 
> Can you try out this version on RTL & LTR with FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and add screenshots for it?


I tried out above version and this is how it formats

Edit:

Screenshot_1629407955....Screenshot_1629407905

veena14cs:
> > ```
> > <p>If it is your first time creating a profile and not have a PIN:
> >   <br />1. From the Profile Chooser, tap on <strong>Set up Multiple Profiles</strong>.
> >   <br />2. Create a PIN and <strong>Save</strong>.
> >   <br />3. Fill in all fields for the profile. </p>
> >   <ul>
> >     <li>(Optional) Upload a photo.</li>
> >     <li>Enter a name.</li>
> >     <li>(Optional) Assign a 3-digit PIN.</li>
> >   </ul>
> >   <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser!
> > </p>
> > <br/>
> > <p>If you have created a profile before and have a PIN:
> >   <br />1. From the Profile Chooser, tap on <strong>Add Profile</strong>.
> >   <br />2. Enter your PIN and tap <strong>Submit</strong>.
> >   <br />3. Fill in all fields for the profile.
> >   <ul>
> >     <li>(Optional) Upload a photo.</li>
> >     <li>Enter a name.</li>
> >     <li>(Optional) Assign a 3-digit PIN.</li>
> >   </ul>
> >   <br />4. Tap <strong>Create</strong>. This profile is added to your Profile Chooser! Note: Only the <u>Administrator</u> is able to manage profiles.
> > </p>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Also note that I'm using multiple lines to make it more readable; I don't think this will have an impact on the final formatting).
> > Interestingly, I noticed extra breaks in the bullet list that probably shouldn't have been there.
> > Can you try out this version on RTL & LTR with FROM_HTML_SEPARATOR_LINE_BREAK_PARAGRAPH and add screenshots for it?
> 
> I tried out above version and this is how it formats
> 
> Edit:
> 
> Screenshot_1629407955....Screenshot_1629407905

There is lot of difference if we have lists inside `<p>`. 

If I have separate versions of LTR and RTL  there is no difference in the output. You can check the updated code for the Faq text.

BenHenning:
@veena14cs I might actually need to take this over to try and dig more deeply into what's going on with the spacing.

veena14cs:
> @veena14cs I might actually need to take this over to try and dig more deeply into what's going on with the spacing.

May be yes you are right. I couldn't find the solution for this.
In that case do I need to close this PR?

PR #3323, opened 2021-06-13
Fix #3322: Platform parameter singleton and Generic Interface

### LINKED ISSUE #3322: Introduce a Singleton Class and a Generic Interface for Platform Parameters
Add new singleton class and an interface for the platform parameter project. Namely

- PlatformParameterSingleton
- PlatformParameter

Target PR date: 14 June 2021
Target completion date: 11 June 2021

Arjupta:
## Explanation

Fixes #3322 
This PR introduces a new domain layer Singleton class which will help with storing and providing platform parameter values. This PR also includes a Generic Interface whose implementations will be provided containing Platform Parameters. (This PR is stacked over #3269). /cc @Sarthak2601 

## Checklist


domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import javax.inject.Inject
+import javax.inject.Singleton
+import org.oppia.android.app.model.PlatformParameter as ParameterValue
+import org.oppia.android.util.platformparameter.PlatformParameter
+
+/** Singleton class which helps in storing and providing Platform Parameters at runtime. */
+@Singleton
+class PlatformParameterSingleton @Inject constructor() {
+  var platformParameterMap: Map<String, ParameterValue> = mapOf()
```
jcqli: Can this be made private? Since we'll be accessing individual values we probably want to hide this.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  /** Initialize platformParameterMap if it is not yet initialised. */
```
jcqli: Be sure to add any [KDoc blocks](https://developer.android.com/kotlin/style-guide#documentation) for public functions (ditto to below function). Main ones to document are parameters, return values, and exceptions, there are plenty of examples in the codebase for this :)

It might also be better to name this `setPlatformParameterMap` since `init` suggests that the initialization of the class which isn't quite accurate (and also update the test names).

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,31 @@
...
+  fun initPlatformParameterMap(map: Map<String, ParameterValue>) {
+    if (platformParameterMap.isEmpty()) platformParameterMap = map
```
jcqli: I'm wondering if we might want to introduce some error-handling with this -- particularly if this gets called twice, we might want to handle that since that indicates something wrong with setting platform parameters. @vinitamurthi  WDYT?

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,142 @@
...
+  lateinit var platformParameterSingleton: PlatformParameterSingleton
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveStringParameter_verifyItsValue() {
+    assertThat(platformParameterSingleton.platformParameterMap.isEmpty()).isTrue()
```
jcqli: This could probably be it's own test `testSingleton_initialPlatformParameterMap_isEmpty` and then you won't need to check it for all the tests.

Also, if we rename the `initPlatformParameterMap` function above this test could then better represent the "initial" state case.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    platformParameterSingleton.initPlatformParameterMap(makePlatformParameterMap())
+    assertThat(platformParameterSingleton.platformParameterMap.isEmpty()).isFalse()
```
jcqli: This can also be it's own test `testSingleton_init/setPlatformParameterMap_isNotEmpty()` and then testing for each value type can just check the value.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.oppia.android.app.model.PlatformParameter as ParameterValue
+import org.oppia.android.testing.robolectric.RobolectricModule
+import org.oppia.android.testing.threading.TestDispatcherModule
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+
+const val stringPlatformParameterName = "string_platform_parameter_name"
```
jcqli: I recommend putting these in the class since they're specific to this class being tested.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveIncorrectParameter_verifyItsValue() {
+    assertThat(platformParameterSingleton.platformParameterMap.isEmpty()).isTrue()
+
+    platformParameterSingleton.initPlatformParameterMap(makePlatformParameterMap())
+    assertThat(platformParameterSingleton.platformParameterMap.isEmpty()).isFalse()
+
+    val incorrectPlatformParameter = platformParameterSingleton.getPlatformParameter<String>(
+      "incorrectPlatformParameterName"
```
jcqli: This can be a class constant too for consistency.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    )
+    assertThat(incorrectPlatformParameter).isEqualTo(null)
```
jcqli: I think there's an `isNull()` check you can use here.

model/BUILD.bazel
```diff
@@ -223,6 +223,17 @@ java_lite_proto_library(
...
 )
 
+format_import_proto_library(
+    name = "platform_parameter",
+    src = "src/main/proto/platform_parameter.proto",
+)
+
+java_lite_proto_library(
+    name = "platform_parameter_java_proto_lite",
+    visibility = ["//visibility:public"],
```
jcqli: We probably want this to be `oppia_api_visibility`

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
```diff
@@ -0,0 +1,6 @@
+package org.oppia.android.util.platformparameter
+
+/** Generic interface whose implementations will be provided as Platform Parameter Values. */
```
jcqli: This might need some more detail to explain how they correspond to the defined protos and why we want this interface.

jcqli (changes requested):
Nice job with the tests! A couple things
1. Try to further split up test cases so that they're only checking on thing at a time. I like to think of it as if we want to minimize the number of assertions in a test so each test can be as specific as possible.
2. I see that protos are included in this PR -- be sure to list only the newly-changed in the PR description in case reviewers are different.
3. Be sure to build & test with both Gradle & Bazel to help speedup the review process. I also forget this sometimes :)

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: It cannot be private, as the inline function (getPlatformParameter) can't acces it. But there can be a workaround like this code snippet - 

      /** Retrieve individual PlatformParameter, if it exists in the platformParameterMap. */
      fun <T : Any> getPlatformParameter(name: String): PlatformParameter<T>? {
        if (platformParameterMap.isEmpty()) return null
        val parameter = platformParameterMap[name] ?: return null
        return object : PlatformParameter<T> {
           override val value: T
            get() = when (getClass<KClass<T>>()) {
              0 -> parameter.string as T
              1 -> parameter.integer as T
              2 -> parameter.boolean as T
              else -> null as T
            }
        }
      }
      
      private inline fun <reified T> getClass(): Int {
         return when (T::class) {
          String::class -> 0
          Int::class -> 1
          Boolean::class -> 2
          else -> 3
       }
      }

Now we can keep the map to be a private var, does this Looks good to you? @jcqli

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: After changing the name of the function to setPlatformParameterMap a clash occured between the setter for platformParameterMap and the function.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Okay got it

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Sure will do that

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: You mean to say I should put these into PlatformParameterSingleton? If yes then isn't it wrong to put the test specific constants in a Production class. @jcqli

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Sure will do it

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Sure will do it

model/BUILD.bazel
Arjupta: Oh! will update this in the base PR ( as this change came after rebasing this PR over the branch that it is stacked upon)

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: Okay

Arjupta:
@fsharpasharp can you take a look at the bazel changes. I am not able to resolve the issues regarding it. Also just for your information, this PR is stacked over #3269. So I will actually commit the changes for Bazel file in the base PR, and then rebase this PR over it.

fsharpasharp:
> @fsharpasharp can you take a look at the bazel changes. I am not able to resolve the issues regarding it. Also just for your information, this PR is stacked over #3269. So I will actually commit the changes for Bazel file in the base PR, and then rebase this PR over it.

You will have to add the newly created proto as a dependency to the domain module's BUILD.bazel file and the corresponding library. Let me know if this is enough to get you started or if I need to provide more details.

fsharpasharp (approved):
Approving for bazel changes. Please take a look at the lint error.

Arjupta:
@jcqli PTAL

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
jcqli: Oh I meant below in this test class! Around line 38.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
jcqli: Nice! I think this can be written as `assertThat(platformParameterSingleton.platformParameterMap).isEmpty()`

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: Hm interesting. I've never used inline functions but I would say the original implementation is probably better since the private function isn't adding any code clarity. @vinitamurthi WDYT?

Also, what's the reason for using an inline function?

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,140 @@
...
+      DaggerPlatformParameterSingletonTest_TestApplicationComponent.builder()
+        .setApplication(this)
+        .build()
+    }
+
+    fun inject(platformParameterSingletonTest: PlatformParameterSingletonTest) {
+      component.inject(platformParameterSingletonTest)
+    }
+  }
+}
```
jcqli: nit: add newline at the end of the file

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,140 @@
...
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveIncorrectParameter_verifyItsValue() {
+    platformParameterSingleton.initPlatformParameterMap(makePlatformParameterMap())
+    val incorrectPlatformParameter = platformParameterSingleton.getPlatformParameter<String>(
+      incorrectPlatformParameterName
+    )
+    assertThat(incorrectPlatformParameter).isNull()
+  }
+
+  private fun makePlatformParameterMap(): Map<String, ParameterValue> {
```
jcqli: This might be better as a class constant since it's being used in nearly every test. That way we don't have to reconstruct it every time.

jcqli (changes requested):
Left a few more comments. Also a few open threads from my last pass as well, wasn't sure how to include them in the same review 🙃

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: Ah got it, is there a reason we are writing our own setter then?  If this is meant to initialize the map only once then I recommend adding error handling if the map has already been initialized. Otherwise why not use the existing setter?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: The reason for inline function is to use \<reified T\>, This is because we want to know which class does T belongs to, which was not possible without using a  \<reified T\>

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Yeah we can use the default setter also. I will find about how to implement the same functionality with it

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Yeah, Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Yeah that can be done its just that in previous tests that I worked, they had constant like this outside of the class ([ test file ](https://github.com/ARJUPTA/oppia-android/blob/9a691d448ed7f7eaeeb7471938669c78e8a875ff/app/src/sharedTest/java/org/oppia/android/app/options/ReadingTextSizeFragmentTest.kt#L79)from my earlier PR)

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Yeah, Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
jcqli: That's fair -- I don't think we have any specific conventions on how to do this so I'll leave it up to you. If you do choose to leave as is, they probably should be `private` and should be named `STRING_PLATFORM_PARAMETER_NAME` etc.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: I did some additional research and I believe I understand now why we have to use the inline function. I'll leave this thread open in case Vinita has additional thoughts.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
jcqli: Hm maybe `Generic interface that is used to provide Platform Parameter Values corresponding to the [PlatformParameter] proto values. Objects that implement this interface will override the [value] property to store the actual platform parameter value.`

As another thought, could this be a generic class rather than an interface? That way we won't have different classes for each value type.

jcqli (commented):
Left a few follow-ups, thanks

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: @vinitamurthi PTAL

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Okay noted

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: This was point was throughly discussed during proposal reviews and also for the submission of additional info. This approach has been suggested by @BenHenning and @vinitamurthi collectively and it matches the way in which we implement a testable feature. (Help us to provide mocks easily). So @vinitamurthi is the best person to take a look at your suggestion.

jcqli: Sounds good, I must've missed that part  -- thanks for the clarification! Could you point me to where that's explained in the proposal?

Arjupta: We will discuss on this in the team meet as there was a email thread regarding the discussion

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Sarthak2601: Instead of using an `inline function` and `<reified T>` to get the platform value type, can't we make use of the proto objects here? We already have defined value types through which we can decide which kind of value we'll be returning. This way the variable can also be made private.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Sarthak2601: I think the `if` clause works fine here -- might be worth adding tests for this to check for regression.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: We can make use of the Proto classes but doing so will include a bug. For example when we use getPlatformParameter\<String\>("param1") , we expect a parameter with name -> "param1" and a value of String type. But in case we depend on Proto class to set the type of value that is going inside PlatformParameter implementation then this can create a RunTime exception when the Proto has a different data type (than String) stored in the same name of "param1"

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: I guess the current tests have this use case. (indirectly tested)

anandwana001:
I will check my code owners file tomorrow.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: Which test is this in? I'm not sure this edge case is tested.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,64 @@
...
+  fun getBooleanPlatformParameter(name: String): PlatformParameter<Boolean>? {
+    if (platformParameterMap.isEmpty()) return null
+    val parameter = platformParameterMap[name]?.boolean ?: return null
+    return object : PlatformParameter<Boolean> {
+      override val value: Boolean
+        get() = parameter
+    }
+  }
+
+//  /** Retrieve individual [ParameterValue], if it exists in the [platformParameterMap]. */
```
jcqli: Be sure to remove any commented-out code here.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: I see this has been renamed, were there collisions with the setter?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: It looks like we are using the proto objects for type-checking now, is that right?

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
jcqli: It looks like we're sticking with the interface implementation -- could you highlight the main reasons for this decision? I know we discussed it but could be useful as a record for others viewing this PR.

jcqli (commented):
Took another pass, I think it's getting close :)
Also be sure to fix the linter errors

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
vinitamurthi: I think we should add a test specific to this use case. The existing tests dont seem to add that

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
vinitamurthi: +1

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,140 @@
...
+  fun testSingleton_initialPlatformParameterMap_isEmpty() {
+    assertThat(platformParameterSingleton.getPlatformParameterMap()).isEmpty()
+  }
+
+  @Test
+  fun testSingleton_initPlatformParameterMap_isNotEmpty() {
+    assertThat(platformParameterSingleton.getPlatformParameterMap().isEmpty()).isTrue()
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    assertThat(platformParameterSingleton.getPlatformParameterMap().isEmpty()).isFalse()
+  }
```
vinitamurthi: Add a test where we call init twice

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,140 @@
...
+
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveIncorrectParameter_verifyItsValue() {
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    val incorrectPlatformParameter = platformParameterSingleton.getStringPlatformParameter(
+      INCORRECT_PLATFORM_PARAMETER_NAME
+    )
+    assertThat(incorrectPlatformParameter).isNull()
+  }
```
vinitamurthi: What happens when we do a getStringPlatformParameter but pass BOOLEAN_PLATFORM_PARAMETER_NAME to it? Can we add a test like this?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: I guess No, we have now made three different functions one each for different data type (String/Int/Boolean). This way we are sure about the type of PlatformParameter to be expected and also in case of any mismatch of the required value type and the cached value type we provide defalt parameter

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: As the property is now private we need a setter to access it

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Okay I will add a test for this too

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: We are sticking with the Interface impl of PlatformParamter so as to be consistent in the way we test features in our codebase. After discussion there was specific point on why we cannot use a Generic Class instead of Generic Interface.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: I have kept the code here in order to compare the previous approach of using a single generic function instead of three different functions. But as everyone seems okay with this new approach so I can remove it now

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: it should return null. Added the test for it

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Added

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+  /** Initialize [platformParameterMap] if it is not yet initialised. */
+  fun setPlatformParameterMap(map: Map<String, ParameterValue>) {
+    if (platformParameterMap.isEmpty()) platformParameterMap = map
+  }
+
+  /** Retrieve a String type [ParameterValue], if it exists in the [platformParameterMap]. */
+  fun getStringPlatformParameter(name: String): PlatformParameter<String>? {
+    if (platformParameterMap.isEmpty()) return null
+    val parameter = platformParameterMap[name] ?: return null
+    if(!parameter.valueTypeCase.equals(ParameterValue.ValueTypeCase.STRING)) return null
```
jcqli: nit: add space after `if`

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
jcqli: What was the specific point? I remember there was a chain of errors occurring when I played around with it but it might be useful to document those here as well if you remember!

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: Ah the calls to `parameter.valueTypeCase.equals(ParameterValue.ValueTypeCase.STRING)` are what I mean by using the protos for type-checking, as opposed to using the function generics to get the types. I like this more than the earlier iteration since I thin kit's more aligned with what we want this getter to be, nice job!

Arjupta: Thanks.

anandwana001 (changes requested):
@ARJUPTA 
Please don't bypass the pre-push check locally. 

Assign me back once all the static analysis checks pass for full PR review.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

Arjupta:
> @ARJUPTA
> Please don't bypass the pre-push check locally.
> 
> Assign me back once all the static analysis checks pass for full PR review.

@anandwana001 I am not able to resolve the error for the lint check about rearranging the imports in a lexicographical order. I have tried rearranging it manually and also the way Android Studio auto formatter tries to. Both not worked and this didn't allowed to me push my code. Can you help me over that?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
+package org.oppia.android.domain.platformparameter
+
+import javax.inject.Inject
```
jcqli: Re: linter errors, I believe the `javax` imports should go at the end.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,160 @@
...
+  fun testSingleton_initPlatformParameterMap_isNotEmpty() {
+    assertThat(platformParameterSingleton.getPlatformParameterMap().isEmpty()).isTrue()
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    assertThat(platformParameterSingleton.getPlatformParameterMap().isEmpty()).isFalse()
+  }
+
+  @Test
+  fun testSingleton_initPlatformParameterMapTwice_checkIsNotUpdatedTwice() {
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    assertThat(platformParameterSingleton.getPlatformParameterMap().isEmpty()).isFalse()
```
jcqli: I believe you can do `isNotEmpty()` (ditto elsewhere this is used).

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    val emptyPlatformParameterMap = mapOf<String, ParameterValue>()
+    platformParameterSingleton.setPlatformParameterMap(emptyPlatformParameterMap)
+    assertThat(
+      platformParameterSingleton.getPlatformParameterMap()
+    ).isEqualTo(mockPlatformParameterMap)
```
jcqli, line 90: I'd also add a check that the map != empty map to guarantee nothing weird is happening with second map.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
jcqli: I see `testSingleton_initPlatformParameterMapTwice_checkIsNotUpdatedTwice`, left a quick follow-up.

jcqli (commented):
Left a few more comments, nothing major.

anandwana001:
> > @ARJUPTA
> > Please don't bypass the pre-push check locally.
> > Assign me back once all the static analysis checks pass for full PR review.
> 
> @anandwana001 I am not able to resolve the error for the lint check about rearranging the imports in a lexicographical order. I have tried rearranging it manually and also the way Android Studio auto formatter tries to. Both not worked and this didn't allowed to me push my code. Can you help me over that?

Suggesting to follow this here, setting this macro will clean everything - https://github.com/oppia/oppia-android/wiki/Ktlint-Guide#macros

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Although if we check that the value of map is equal to mockMap that means it is not equal to emptyMap, but adding that check just to be sure

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
```diff
@@ -0,0 +1,10 @@
+package org.oppia.android.util.platformparameter
+
+/**
+ * Generic interface that is used to provide Platform Parameter Values corresponding to the
+ * [PlatformParameter] proto values. Objects that implement this interface will override the [value]
+ * property to store the actual platform parameter value.
+ * */
+interface PlatformParameter<T> {
```
jcqli: Could we rename this as PlatformParameterValue? Since it's just holding the value, and that lets us avoid naming collisions with the proto type

jcqli:
Left one more comment but otherwise LGTM -- there's one more [comment thread](https://github.com/oppia/oppia-android/pull/3323#discussion_r658066864) still open, though, PTAL.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,163 @@
...
+
+private const val BOOLEAN_PLATFORM_PARAMETER_NAME = "boolean_platform_parameter_name"
+private const val BOOLEAN_PLATFORM_PARAMETER_VALUE = true
+
+private const val INCORRECT_PLATFORM_PARAMETER_NAME = "incorrect_platform_parameter_name"
+
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = PlatformParameterSingletonTest.TestApplication::class)
+class PlatformParameterSingletonTest {
```
anandwana001, line 40: add kdoc

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,163 @@
...
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveIncorrectTypeParameter_verifyIsNull() {
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    val incorrectPlatformParameter = platformParameterSingleton.getStringPlatformParameter(
+      BOOLEAN_PLATFORM_PARAMETER_NAME
+    )
+    assertThat(incorrectPlatformParameter).isNull()
+  }
+
+  fun setUpTestApplicationComponent() {
```
anandwana001: private function

anandwana001 (changes requested):
Test file LGTM,
nit updates

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+import javax.inject.Singleton
+import org.oppia.android.app.model.PlatformParameter as ParameterValue
+
+/** Singleton class which helps in storing and providing Platform Parameters at runtime. */
+@Singleton
+class PlatformParameterSingleton @Inject constructor() {
+  private var platformParameterMap: Map<String, ParameterValue> = mapOf()
+
+  /** Get the current [platformParameterMap]. */
```
vinitamurthi: Get --> Gets. Also add a `@return` annotation (See https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax for an example)

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+  fun getPlatformParameterMap() = platformParameterMap
+
+  /** Initialize [platformParameterMap] if it is not yet initialised. */
```
vinitamurthi: Change comment to -> `Initializes a [platformParameterMap] in case it is empty.`
Technically you can initialize it as an empty map, and if you do that it will keep initializing it if this method is called

vinitamurthi: Also add the @param in KDoc (See https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax for an example of `@param` and `@return` annotations)

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+  fun setPlatformParameterMap(map: Map<String, ParameterValue>) {
+    if (platformParameterMap.isEmpty()) platformParameterMap = map
+  }
+
+  /** Retrieve a String type [ParameterValue], if it exists in the [platformParameterMap]. */
```
vinitamurthi: Retrieve --> Retrieves. Add `@param` and `@return` annotations

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+    if (platformParameterMap.isEmpty()) return null
+    val parameter = platformParameterMap[name] ?: return null
+    if (!parameter.valueTypeCase.equals(ParameterValue.ValueTypeCase.STRING)) return null
+    return object : PlatformParameter<String> {
+      override val value: String
+        get() = parameter.string
+    }
+  }
+
+  /** Retrieve an Integer type [ParameterValue], if it exists in the [platformParameterMap]. */
```
vinitamurthi: Same comment as above

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,53 @@
...
+    if (platformParameterMap.isEmpty()) return null
+    val parameter = platformParameterMap[name] ?: return null
+    if (!parameter.valueTypeCase.equals(ParameterValue.ValueTypeCase.INTEGER)) return null
+    return object : PlatformParameter<Int> {
+      override val value: Int
+        get() = parameter.integer
+    }
+  }
+
+  /** Retrieve a Boolean type [ParameterValue], if it exists in the [platformParameterMap]. */
```
vinitamurthi: Same comment as above

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,163 @@
...
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = PlatformParameterSingletonTest.TestApplication::class)
+class PlatformParameterSingletonTest {
+
+  @Inject
+  lateinit var platformParameterSingleton: PlatformParameterSingleton
+
+  private val mockPlatformParameterMap by lazy {
+    val stringPlatformParameter = ParameterValue.getDefaultInstance().toBuilder()
+      .setString(STRING_PLATFORM_PARAMETER_VALUE).build()
```
vinitamurthi, line 47: Can you not do `ParameterValue.newBuilder().setString(...).build()`? That should create the default instance anyway (same comment below)

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    assertThat(platformParameterSingleton.getPlatformParameterMap()).isNotEmpty()
+  }
+
+  @Test
+  fun testSingleton_initPlatformParameterMapTwice_checkIsNotUpdatedTwice() {
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    assertThat(platformParameterSingleton.getPlatformParameterMap()).isNotEmpty()
+    val emptyPlatformParameterMap = mapOf<String, ParameterValue>()
+    platformParameterSingleton.setPlatformParameterMap(emptyPlatformParameterMap)
+    assertThat(
```
vinitamurthi, line 85: Add a new line before the assert. This is just to separate the lines of code that is setting up the test, and the lines of code that is checking if the result is as expected

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
```diff
@@ -0,0 +1,10 @@
+package org.oppia.android.util.platformparameter
+
+/**
+ * Generic interface that is used to provide Platform Parameter Values corresponding to the
```
vinitamurthi: Platform Parameter Values --> platform parameter values

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
```diff
@@ -0,0 +1,10 @@
...
+ * [PlatformParameter] proto values. Objects that implement this interface will override the [value]
+ * property to store the actual platform parameter value.
+ * */
```
vinitamurthi:
```suggestion
 */
```

vinitamurthi (changes requested):

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: Yeah that will be better. Done
/cc @vinitamurthi

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: Sorry for the mistake. I actually meant that there was "no specific point on why we cannot use a Generic Class"

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Yeah this is better. Done

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: Done.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: Done

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: This is neccessary for KDoc. So cannot remove this asterisk

anandwana001: The format suggested by Vinita here is a standard format of the comment block. 
Check here for reference - `promotedlist/ComingSoonTopicsListView.kt`

Official reference - https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax

Don't know why ktlint doesn't catch this, creating an issue for tracking this.

anandwana001: Issue created https://github.com/oppia/oppia-android/issues/3394

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+  private val mockPlatformParameterMap by lazy {
+    val stringPlatformParameter = PlatformParameter.newBuilder()
+      .setString(STRING_PLATFORM_PARAMETER_VALUE).build()
+    val integerPlatformParameter = PlatformParameter.newBuilder()
+      .setInteger(INTEGER_PLATFORM_PARAMETER_VALUE).build()
+    val booleanPlatformParameter = PlatformParameter.newBuilder()
+      .setBoolean(BOOLEAN_PLATFORM_PARAMETER_VALUE).build()
+
+    mapOf(
```
anandwana001: mention the type to prevent `Type inference` warnings. 

Official reference - https://kotlinlang.org/spec/type-inference.html

anandwana001 (changes requested):
nit suggestion

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: I have made the change, but just confirming that Android Studio was not giving any warnings on this point. So was this a manual check thing or something which can be detected automatically in some way

anandwana001: oh, It was giving a warning for me, `red line` under the `mapOf()`

anandwana001 (changes requested):
LGTM, keeping the approval blocked on 

https://github.com/oppia/oppia-android/pull/3323#discussion_r661613460

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameter.kt
Arjupta: Done

Arjupta:
> LGTM, keeping the approval blocked on
> 
> [#3323 (comment)](https://github.com/oppia/oppia-android/pull/3323#discussion_r661613460)

Done the required changes

vinitamurthi (approved):
LGTM!

jcqli (approved):
LGTM!

BenHenning:
I plan to look at this tomorrow.

rt4914 (approved):
I haven't checked this PR. 

Approving as oppia-bot is assigning it back to me considering me as `code-owner` but there are no files related to me.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
```diff
@@ -0,0 +1,164 @@
...
+    ).isEqualTo(mockPlatformParameterMap)
+  }
+
+  @Test
+  fun testSingleton_initPlatformParameterMap_retrieveStringParameter_verifyItsValue() {
+    platformParameterSingleton.setPlatformParameterMap(mockPlatformParameterMap)
+    val stringPlatformParameter = platformParameterSingleton.getStringPlatformParameter(
+      STRING_PLATFORM_PARAMETER_NAME
+    )
+    assertThat(stringPlatformParameter?.value).isEqualTo(STRING_PLATFORM_PARAMETER_VALUE)
```
anandwana001, line 99: ah, I missed this earlier, is passing a nullable in `assertThat` is a good idea? 

What happens if we pass a null value to `assertThat`? 
Also, does the testing goes correctly if passing a null value?
Or can we modify the test something like where we confirm that we will get the desired value and null never shows up?

anandwana001 (changes requested):

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonTest.kt
Arjupta: @anandwana001 This nullable result is part of our logic.
We pass a name to the gettter method (inside singleton) then this getter method looks for the parameter corresponding to this name in the map. If it find this map then it returns that PlatformParameter instance but if in case there is no such parameter in the map then in that case this getter method returns null. Now it is the responsibility of the dagger module to provide a default PlatformParameter whenever singleton returns null. 

So for our test -> We are checking that after passing in the correct name (means name of a parameter which exists in map) do we have the required value. This required value is not going to be null and should be equal to true value (which we set at the start of the test). 

Hence this single assertThat covers a test for both cases ie it checks that the returned value should not be null and it should be equal to the true value.

vinitamurthi (approved):

vinitamurthi:
Note that tests are failing

Arjupta:
> Note that tests are failing

@vinitamurthi that test is failing due to some env reading error not because of the code changes

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterSingleton.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Gets the current platformParameterMap.
+   * @return [Map<String, PlatformParameter>]
+   */
+  fun getPlatformParameterMap(): Map<String, PlatformParameter>
+
+  /**
+   * Initializes a platformParameterMap in case it is empty.
+   * @param platformParameterMap [Map<String, PlatformParameter>]
```
jcqli, line 18: nit: add newline above KDoc `@` fields (but each `@` line should be next to each other).

jcqli (approved):
LGTM, just be sure to update the KDocs!

anandwana001 (approved):
LGTM

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterSingleton.kt
Arjupta: Done

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonImpl.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.oppia.android.app.model.PlatformParameter
+import org.oppia.android.util.platformparameter.PlatformParameterSingleton
+import org.oppia.android.util.platformparameter.PlatformParameterValue
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Singleton which helps in storing and providing Platform Parameters at runtime. */
+@Singleton
+class PlatformParameterSingletonImpl @Inject constructor() : PlatformParameterSingleton {
+  private var platformParameterMap: Map<String, PlatformParameter> = mapOf()
```
BenHenning: This could be a val with a mutable list, instead, where you clear & perform a ``putAll`` to reinitialize it versus changing the parameter itself.

BenHenning (approved):
Approved for codeowners (BUILD file changes). Left one unrelated comment for one thing I noticed.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonImpl.kt
Arjupta: Done

PR #3338, opened 2021-06-17
[DO NOT MERGE] Fix #3285: Create script to compile list of maven dependencies [WIP]

### LINKED ISSUE #3285: Create a script to generate the list of Maven dependencies, versions, and license links
Create a Kotlin script that will contain a complete list of Maven dependencies (direct + transitive) for which the licenses will be extracted.

**Target PR date**: 2021-06-19
**Target completion date**: 2021-06-24

prayutsu:
## Explanation

Fixes #3285:
Created a kotlin script that compiles a list of Maven dependencies for which the licenses need to be fetched.

## CAUTION

**This PR has only temporary changes.**

## Notes for Reviewers:
The script first creates a list **L1** of dependencies by parsing the maven_install.json file and another list **L2** of dependencies by running the bazel query command. The list **L2** contains only the names of artifacts and that too separated by underscores and with an unwanted prefix e.g -
`@maven//:com_squareup_moshi_moshi`.

Now we iterate over **L1**, and we parse the name of each dependency so that we can search for it in the list **L2**. If the parsed name is present in **L2**, then we add this dependency of **L1** to a new list **L3**.

_Now we try to make a list **L4** and each object of **L4** will contain the dependency's name, version, list of license names, and list of license links._

We have a final list **L3** of dependencies and each dependency also has a URL associated with it that generally corresponds to the dependencies' jar file. We change the extension from the .jar to .pom to read the POM file of the dependency, which can potentially give us the license links for that dependency. If we find the license links in the POM file of the dependency, we simply add this dependency from to **L4** (we extract the version name from the name of the artifact and we provide the corresponding license names and links), else we try to find the links from our backup mechanism. 

The backup mechanism works as follows - 
1. We have a backup_license_links.json file that contains an array of dependencies for which the licenses have to be provided manually.
2. The script parses this JSON file and stores all the dependencies listed in this file in a list **L5**.
3. When we don't find any licenses tag for any dependency, we then try to find the dependency in **L5** and the possibilities are - 

```kotlin
if(dependency is present in L5) {
  // check if the corresponding list of licenses and their links is non-empty
  if(licenseList is non-empty()) {
    // copy these license links list and add the dependency to L5
  } else {
    // print message - "Please provide the license names and links for the artifact - $artifactName"
    // script fails.
  }
} else {
  // add this dependency to L5.
  // script fails.
}
```

So, script will fail if the backup.json is empty or it does not contain all the dependencies for which our strategy failed. Also, with failure, it also lists down all the dependencies in the backup.json for which we need to provide the license links manually. If we provide the links for all these dependencies, the script will not fail and the list of dependencies will be complete.
     

## Checklist


prayutsu:
@BenHenning @rt4914 I have explained the current working of the first script in detail in the PR explanation section. 
I am assigning you the PR for initial feedback.

prayutsu:
@BenHenning @rt4914 Till now, we have build up a mechanism that will provide us the license links for all the Maven dependencies. Now the challenge is - how do we detect that the license link contains plain text or not (because if it doesn't then we can't scrape the text from the URL directly).
I made a set of license links to see how many different licenses we will attribute and the set involves these 11 links - 

1. http://www.apache.org/licenses/LICENSE-2.0.txt
2. http://try.crashlytics.com/terms/terms-of-service.pdf
3. http://www.opensource.org/licenses/bsd-license
4. https://www.apache.org/licenses/LICENSE-2.0.txt
5. https://developer.android.com/studio/terms.html
6. https://fabric.io/terms
7. https://github.com/javaee/javax.annotation/blob/master/LICENSE
8. https://github.com/DanielMartinus/Konfetti/blob/main/LICENSE
9. http://www.gnu.org/software/classpath/license.html
10. http://opensource.org/licenses/MIT
11. https://asm.ow2.io/license.html

1 and 4 are the same links, the only difference is -  1 contains http and 4 contains https.
So, out of these 10 links only one is directly scrapable - (link corresponding to index 1).
And 2. and 6. are not even valid license links.

**I also have a doubt here -- the 3. link has some placeholders in the actual license text, do we also need to fill these placeholders?**

As per the proposal, we'll have to make a separate repository to handle such cases and replace those deps which do not have any licenses.

To handle this, I have the following strategy in mind - 
We create a JSON file that will contain all the license links as keys and their value will be another link that will contain the plain text of that license. (We'll have to manually update this JSON)

So, this way we can ensure that the license link is always easily scrapable.

WDYT?

rt4914:
@BenHenning This strategy has been discussed by me and Abhay but we need you thoughts on it.

prayutsu:
@BenHenning @rt4914 
I also tried this [Gradle Plugin](https://developers.google.com/android/guides/opensource) for attributing Copyright licenses of open-source libraries.
This plug-in shows plain license texts for some dependencies like this - 


And for some deps, it only shows the license link (even if the license link is the same as in the above case) and upon clicking on the link, we are directed to the web page displaying the license of the dependency.



If we maintain a repository of licenses, and if the content of the license changes in the future, we may not be able to detect it. 
That's why the best option according to me here is to only display the licenses' links.

scripts/maven/Dependencies.kt
```diff
@@ -0,0 +1,10 @@
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/** Data class that stores the list of dependencies present in maven_install.json. */
+@JsonClass(generateAdapter = true)
+data class Dependencies(
+
+  @Json(name = "dependencies") val dependencyList: MutableList<Dependency>
```
BenHenning: For this & other files: why MutableList? Why not just List?

scripts/maven/Dependency.kt
```diff
@@ -0,0 +1,19 @@
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Data class that contains all the details relevant to a dependency that is present
+ * in maven_intall.json.
+ */
+@JsonClass(generateAdapter = true)
+data class Dependency(
```
BenHenning: I suggest differentiating between maven_install & dependency_list Moshi classes by prefixing them as so:

MavenListDependency
LicenseDependency
...

WDYT?

scripts/maven/Dependency.kt
```diff
@@ -0,0 +1,19 @@
+import com.squareup.moshi.Json
```
BenHenning: See my doc comment, but these classes should be organized in a package structure similar to #3340.

scripts/maven/DependencyTree.kt
```diff
@@ -0,0 +1,8 @@
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/** Helper Data class to parse the list of dependencies from maven_intall.json. */
```
BenHenning: This is actually a structure definition class, not a helper class (e.g. it has no logic to 'help' with anything, but it can be used to parse the list of dependencies from maven_install).

scripts/maven/maven_dependencies_list.kts
```diff
@@ -0,0 +1,315 @@
+#!/usr/bin/env kscript
```
BenHenning: See my doc comment: I think it's preferable to change this to be an actual Kotlin binary file (i.e. with a main() function & a Bazel definition) so that it can be run through Bazel.

However, that may pose a slight problem since this script is also executing Bazel commands. Can you maybe see if running this script with bazel run & it, in turn, using bazel query works, or if it hangs? If it hangs, you may need to set the ``output_base`` of the bazel query command to be something like ``/tmp`` / different from the script so that they run separately.

BenHenning (changes requested):
Left some comments on the code.

Regarding the issue above, we should prefer showing the licenses when we can. It seems we have the following cases:
1. License is scrapable & plain text
2. License is plain text, but not easy to scrape
3. License is not plain text, but available
4. License cannot be found (either because there is no license linked, or because the link goes to a missing page)

For case (1), we should scrape it. For case (2), I'm assuming we will need to copy the license manually & indicate to the script that those licenses should pull from our local copy (using an association like you indicated seems like a good idea). For (3) & (4), we should probably should provide the link since it's the best that we can do in these cases.

This being said, I think we also need a way to manually override the licenses for individual dependencies. If, for example, a dependency has a filled out license in their repository, we should be able to scrape that instead of the one defined in Maven. If we manually override a dependency's license, it shouldn't then be removed by the script.

Would this work?

scripts/maven/Dependencies.kt
prayutsu: Because once they are generated, we also need to sort them. And I think we can't sort a simple List.

scripts/maven/maven_dependencies_list.kts
prayutsu: @rt4914 @fsharpasharp 
I tried setting up the output_base to `/tmp` and `/home/prayutsu/c1` (random) but it still isn't getting executed properly.
The full command is - 

`bazel --output_base=/tmp query 'deps(deps(//:oppia) intersect //third_party/...) intersect @maven//...'`

When run on terminal directly, both got executed successfully.
It is not hanging, the process is returning a non-zero status code and thus the program is getting terminated.

scripts/maven/Dependency.kt
prayutsu: Sounds good to me, but the refactoring here works only at some places and it requires a little manual work, so can I do it once the script runs correctly?

app/src/main/java/org/oppia/android/app/maven/GenerateMavenDependenciesList.kt
```diff
@@ -0,0 +1,395 @@
...
+        }
+        bazelQueryDepsNames.sort()
+
+        val exitValue = process.waitFor()
+        if (exitValue != 0) {
+          System.err.println(
+            "There was some unexpected error while " +
+              "running the bazel Query command."
+          )
+          throw Exception("Unexpected error.")
```
prayutsu: The command is returning non-zero exitValue even after changing the output_base.

scripts/maven/Dependency.kt
prayutsu: Refactoring done in the fake package.

scripts/maven/Dependency.kt
prayutsu: Will use the same structure when script is configured properly.

scripts/maven/DependencyTree.kt
prayutsu: Fixed in the fake package's class.

prayutsu:
@BenHenning I am trying to change the script to a .kt program, but Android Studio is not providing autocomplete and error highlighting in the scripts directory, so I have temporarily created a maven package in the app module but and implemented the suggested changes there, once the script gets configured properly, I'll shift this to scripts package and make the changes as per the requirements.

Command used to run this temporary script- 
` bazel run //app/src/main/java/org/oppia/android/app/maven:generate_maven_dependencies_list`

prayutsu:
> Left some comments on the code.
> 
> Regarding the issue above, we should prefer showing the licenses when we can. It seems we have the following cases:
> 
> 1. License is scrapable & plain text
> 2. License is plain text, but not easy to scrape
> 3. License is not plain text, but available
> 4. License cannot be found (either because there is no license linked, or because the link goes to a missing page)
> 
> For case (1), we should scrape it. For case (2), I'm assuming we will need to copy the license manually & indicate to the script that those licenses should pull from our local copy (using an association like you indicated seems like a good idea). For (3) & (4), we should probably should provide the link since it's the best that we can do in these cases.
> 
> This being said, I think we also need a way to manually override the licenses for individual dependencies. If, for example, a dependency has a filled out license in their repository, we should be able to scrape that instead of the one defined in Maven. If we manually override a dependency's license, it shouldn't then be removed by the script.
> 
> Would this work?

@BenHenning By plain text, you mean a web page that contains raw text only, right?

For case (2), does local copy mean a separate Github repository or we are going to add these license text files in our codebase?
If you meant a separate repository then I think I'll need some more rights to create one. I also have another concern with this -- we would be displaying the old version of the license (the copied one) if the content of the license changes, is it okay?

For case (4), if the license could not be found, then what link we should provide?

prayutsu:
The final list obtained has a total of 75 dependencies which seems a little less to me. Also, ideally, to obtain the parsed name of an artifact, we should replace only ':' and '.' with '\_' and omit the version from it. But when I did this, the final list contained only 31 dependencies, and then I tried replacing '-' with '_' and this gives a list of 75 dependencies.

Also, I notice here https://github.com/oppia/oppia-android/blob/847d52cfa6d65eb43d851c99ca781a218930bed9/data/BUILD.bazel#L29 that to use artifact `    "com.squareup.moshi:moshi-kotlin": "1.11.0"` we referred it as `//third_party:moshi` but it does not match the example given here https://github.com/oppia/oppia-android/blob/847d52cfa6d65eb43d851c99ca781a218930bed9/third_party/versions.bzl#L10

@fsharpasharp Do you have some idea how this thing works? Why we are not referring to the Moshi dependency as `//third_party:com_squareup_moshi_moshi-kotlin` and how is it even working?

fsharpasharp:
> The final list obtained has a total of 75 dependencies which seems a little less to me. Also, ideally, to obtain the parsed name of an artifact, we should replace only ':' and '.' with '_' and omit the version from it. But when I did this, the final list contained only 31 dependencies, and then I tried replacing '-' with '_' and this gives a list of 75 dependencies.
> 
> Also, I notice here
> 
> https://github.com/oppia/oppia-android/blob/847d52cfa6d65eb43d851c99ca781a218930bed9/data/BUILD.bazel#L29
> 
> that to use artifact ` "com.squareup.moshi:moshi-kotlin": "1.11.0"` we referred it as `//third_party:moshi` but it does not match the example given here
> https://github.com/oppia/oppia-android/blob/847d52cfa6d65eb43d851c99ca781a218930bed9/third_party/versions.bzl#L10
> 
> @fsharpasharp Do you have some idea how this thing works? Why we are not referring to the Moshi dependency as `//third_party:com_squareup_moshi_moshi-kotlin` and how is it even working?

There's a library called moshi under third_party:

https://github.com/oppia/oppia-android/blob/847d52cfa6d65eb43d851c99ca781a218930bed9/third_party/BUILD.bazel#L67

rt4914:
@prayutsu Un-assigned myself until the PR is ready for review

prayutsu:
Got this, thanks @fsharpasharp

app/src/main/java/org/oppia/android/app/maven/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_binary", "kt_jvm_library", "kt_jvm_test")
```
BenHenning, line 1: This & related files should be part of the script package, e.g. ``script/src/java/org/oppia/android/script/maven``. See #3374 for an idea on how this should be set up.

app/src/main/java/org/oppia/android/app/maven/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+    srcs = [
+        "GenerateMavenDependenciesList.kt",
+    ],
+    deps = [
+        ":backup_dependency",
+        ":backup_license",
+        ":license_dependency",
+        ":maven_list_dependencies",
+        ":maven_list_dependency",
+        ":maven_list_dependency_tree",
```
BenHenning: Why split all of these up?

Also, I suggest having a moshi subpackage for all of the Moshi model definitions. Then, the controller can live as part of the main script package.

app/src/main/java/org/oppia/android/app/maven/GenerateMavenDependenciesList.kt
BenHenning: Try copying in BazelClient from #3374 and create any new query methods you need using the same approach as the other ones & see if that works.

scripts/maven/Dependencies.kt
BenHenning: In that case we should be copying the data when we need to change it. Models generally should be immutable--that offers a bunch of benefits in data management & reduction in complexity.

scripts/maven/maven_dependencies_list.kts
BenHenning: See my other comment--try using the same BazelClient approach as #3374. I found that the output_base wasn't actually needed.

BenHenning (commented):
Thanks @prayutsu. Left some thoughts on getting the script working & organized. Once the script is working and the PR is considered finalized, I'll perform a full review.

Please also be sure to add tests for the script. I'll be adding tests to #3374 momentarily to provide some examples for how to do this.

prayutsu:
@BenHenning @fsharpasharp @rt4914 
When I print a proto class object it also includes values corresponding to the enums contained in it. e.g - 

```proto
message License {
  string license_name = 1;
  PrimaryLinkType primary_link_type = 4;
}

enum PrimaryLinkType {
  UNSPECIFIED = 0;
  SCRAPE_DIRECTLY = 1;
  SCRAPE_FROM_LOCAL_COPY = 2;
  SHOW_LINK_ONLY = 3;
  NEEDS_INTERVENTION = 4;
}
```

Now when I do this in my Kotlin program - 

```kotlin
val license = License
            .newBuilder()
            .setLicenseName("Some Licenses")
            .setPrimaryLinkType(PrimaryLinkType.SCRAPE_DIRECTLY)
            .build()
println(license)
```

Output - 
```
license_name: "Some License"
primary_link_type: SCRAPE_DIRECTLY
primary_link_value: 1
```
The output also has the primary_link_value attribute which throws an error when the script tried to read it from the textproto file.
I don't want primary_link_value in the output, is there any way to configure it to not include value in the output?

BenHenning:
@prayutsu are you outputting that textproto using TextFormat (e.g.: https://www.javadoc.io/doc/com.google.protobuf/protobuf-java/3.5.1/com/google/protobuf/TextFormat.html)? Please follow up with me to get access to an example for outputting textproto.

prayutsu:
> @prayutsu are you outputting that textproto using TextFormat (e.g.: https://www.javadoc.io/doc/com.google.protobuf/protobuf-java/3.5.1/com/google/protobuf/TextFormat.html)? Please follow up with me to get access to an example for outputting textproto.

@BenHenning Actually, I tried to bring TextFormat in my script but it seems like I can not access this class in the Oppia-android repository, is there any other dependency that I need to add in order to use this class?

Also, I took a look at its methods but I couldn't find anything that can make the output not include values for enums.
It would be great if you can give an example.

BenHenning:
@prayutsu just sent you a message on chat & shared the code reference for you. Please follow up on chat if you have questions or get stuck.

prayutsu:
Created a new PR #3434 as the structure has changed significantly.

PR #3340, opened 2021-06-18
Fix #3290: Add support for generic regex pattern matching

Sparsh1212:
## Explanation

Fixes #3290: Add support for generic regex pattern matching

The script traverses different layers of the codebase to check for any prohibited file content and file naming patterns.
To run the script, use:

`bazel run //scripts:regex_pattern_validation_check -- $(pwd)`

For testing the script, automated tests have been added.
To execute the tests, use:

`bazel test //scripts/src/javatests/org/oppia/android/scripts/regex:RegexPatternValidationCheckTest`

**Note:** We are generating the test assets dynamically at the time of executing them. The test assets are automatically deleted, when the test finishes.

Screenshot of all new tests added passing locally:

image

image

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3341, opened 2021-06-18
Fix #3291: Add check for XML syntax validation

### LINKED ISSUE #3291: Create a Kotlin script to validate the Xml files for syntax correctness
We want to make sure that, all the Xml files are syntactically correct as per the Xml syntax. Introduce a script that checks all the Xml files for syntax correctness.

**Target PR date:** 2021-06-23
**Target completion date:** 2021-06-27

Sparsh1212:
## Explanation

Fixes #3291: Add check for XML syntac validation

The script traverses different layers of the codebase to check for the syntax correctness of XML files
To run the script, use:

`bazel run //scripts:xml_syntax_check -- $(pwd)`

For testing the script, automated tests have been added.
To execute the tests, use:

`bazel test //scripts/src/javatests/org/oppia/android/scripts/xml:XmlSyntaxCheckTest`

**Note:** We are generating the test assets dynamically at the time of executing them. The test assets are automatically deleted, when the test finishes.

**Blocked on:** #3340 

Screenshot of new tests passing locally:

image

## Checklist


.github/workflows/static_checks.yml
```diff
@@ -107,3 +107,7 @@ jobs:
       - name: Regex Patterns Validation Check
         run: |
           bazel run //scripts:pattern_validation_check -- $(pwd) app data utility domain
+
+      - name: XML Syntax Validation Check
+        run: |
+          bazel run //scripts:xml_syntax_check -- $(pwd) app data utility domain
```
anandwana001: data domain doesn't have any res file, utility does file dimen and string XML file if we are checking that else no needed, we can keep the check only for app module.

anandwana001 (changes requested):
Thanks @Sparsh1212 

Same comment from the previous PR review, for script `kt` file and test file.

Sparsh1212:
> Thanks @Sparsh1212
> 
> Same comment from the previous PR review, for script `kt` file and test file.

Thanks, @anandwana001 for the suggestions. 
The PR is now updated. Please take a look.

.github/workflows/static_checks.yml
Sparsh1212: Yes! they don't have any res files, but my only concern was that these layers do have the `AndroidManifest.xml` file present, so I thought of including them too in the XML syntax check. 
WDYT?

fsharpasharp (approved):
Approving for BUILD.bazel change

BenHenning:
Please address my comments from #3340 & apply the similar feedback to this & your next PRs, then assign this back to me. Please make sure to ensure all documentation & code is properly formatted, clear, and that you add all test cases that you can consider.

anandwana001:
@Sparsh1212 Follow guide - https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section

Sparsh1212:
@anandwana001 @BenHenning The PR is now updated taking into account all the suggestions received on #3340. Please, take a look.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,146 @@
+package org.oppia.android.scripts
+
+import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.parsers.DocumentBuilder
+import java.io.File
+import org.xml.sax.SAXParseException
+import org.xml.sax.ErrorHandler
+import org.oppia.android.scripts.ScriptResultConstants
+
+class XMLSyntaxCheck {
```
anandwana001: add kdocs

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,146 @@
...
+  companion object {
+    @JvmStatic
+    fun main(vararg args: String) {
+      /** path of the repo to be analyzed. */
```
anandwana001: single line comment `//`

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,146 @@
...
+      /** builder factory which provides the builder to parse the XMl. */
+      val builderFactory = DocumentBuilderFactory.newInstance()
+
+      /** document builder which parses the XMl. */
+      val docBuilder = builderFactory.newDocumentBuilder()
+
+      docBuilder.setErrorHandler(SyntaxErrorHandler(repoPath = repoPath))
+
+      /** check if the repo has any syntactically incorrect XML. */
+      val hasXmlSyntaxFailure = searchFiles.fold(initial = false) { isFailing, file->
```
anandwana001:
```suggestion
      val hasXmlSyntaxFailure = searchFiles.fold(initial = false) { isFailing, file ->
```

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,146 @@
...
+       * Collects the paths of all the files which are needed to be checked.
+       *
+       * @param repoPath the path of the repo
+       * @param allowedDirectories a list of all the directories which needs to be checked
+       * @return all files which needs to be checked
+       */
+      fun collectSearchFiles(
+        repoPath: String,
+        allowedDirectories: List<String>)
+        : List<File> {
```
anandwana001:
```suggestion
      fun collectSearchFiles(
        repoPath: String,
        allowedDirectories: List<String>
      )
```

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,146 @@
...
+        return File(repoPath).walk().filter { it ->
+          checkIfAllowedDirectory(
+            retrieveFilePath(it, repoPath),
+            allowedDirectories)
```
anandwana001:
```suggestion
            allowedDirectories
          )
```

anandwana001 (changes requested):
Suggesting to follow my suggestion from the https://github.com/oppia/oppia-android/pull/3340 latest PR review

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: done

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: done

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: done

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: done

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: done

Sparsh1212:
> Suggesting to follow my suggestion from the #3340 latest PR review

Thanks, @anandwana001. 
The PR is updated on the basis of comments received on #3340 and new suggestions too. PTAL.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      if (hasXmlSyntaxFailure) {
+        throw Exception(ScriptResultConstants.XML_SYNTAX_CHECK_FAILED)
+      } else {
+        println(ScriptResultConstants.XML_SYNTAX_CHECK_PASSED)
+      }
+    }
+
+    /** Check if a file has a broken XML syntax.
```
anandwana001: correct format - https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,105 @@
...
+      println(failureMessage)
+      println()
+    }
+  }
+
+  /**
+   * Class for custom error handling of the parse exception thrown
+   * by the parser, to log the failure message in a cleaner way.
+   */
+  private class SyntaxErrorHandler(val repoPath: String) : ErrorHandler {
```
anandwana001: can we separate out this class?

anandwana001 (changes requested):

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Separated. PTAL.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Apologies! Corrected. PTAL.

anandwana001 (changes requested):
Please check on failing checks.

Sparsh1212:
> Please check on failing checks.

@anandwana001 
The checks are failing to due to a bug in `compute_affected_tests.sh` in the CI workflow. There is a PR
 https://github.com/oppia/oppia-android/pull/3374 
open currently which will fix this issue. It is passing successfully on local. This applies to all the PRs: #3340 #3341 #3343 #3352

vinitamurthi (approved):
LGTM for codeowner files

.github/workflows/static_checks.yml
anandwana001: Looking for @BenHenning thoughts on this

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Logs the failures for XML syntax validation.
+   *
+   * @param e the parsing exception thrown by the parser
+   * @param repoPath the path of the repo to be analyzed
+   * @return log the failure
+   */
+  private fun logCheckFailure(e: SAXParseException, repoPath: String) {
```
anandwana001: arguments name should be readable

anandwana001 (changes requested):

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: Done. Made them more descriptive.

anandwana001 (approved):
LGTM

.github/workflows/static_checks.yml
```diff
@@ -107,3 +107,8 @@ jobs:
       - name: Regex Patterns Validation Check
         run: |
```
BenHenning, line 109: Should this also be ``if: always()`` in #3340?

.github/workflows/static_checks.yml
BenHenning: Hmm. I just left a comment in #3340 about passing in the root directory and just recursing among that. If we validate all XML files then this slight oddity is less noticeable (since we do ultimately want all XML files to be valid). Would that mitigate the concern here?

scripts/BUILD.bazel
```diff
@@ -46,3 +46,34 @@ kt_jvm_test(
...
+    testonly = True,
+    srcs = ["src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt"],
+    test_class = "org.oppia.android.scripts.XMLSyntaxCheckTest",
+    deps = [
+        ":xml_syntax_check_lib",
+        "//testing:assertion_helpers",
+        "//third_party:com_google_truth_truth",
+        "//third_party:org_jetbrains_kotlin_kotlin-test-junit",
+    ],
+)
```
BenHenning: Ditto here--these should be at the file level.

scripts/src/java/org/oppia/android/scripts/ScriptResultConstants.kt
```diff
@@ -5,4 +5,8 @@ object ScriptResultConstants {
   /** RegexPatternValidationCheck output message constants. **/
   val REGEX_CHECKS_PASSED: String = "REGEX PATTERN CHECKS PASSED"
   val REGEX_CHECKS_FAILED: String = "REGEX PATTERN CHECKS FAILED"
+
+  /** XMLSyntaxCheck output message constants **/
+  val XML_SYNTAX_CHECK_PASSED: String = "XML SYNTAX CHECK PASSED"
+  val XML_SYNTAX_CHECK_FAILED: String = "XML SYNTAX CHECK FAILED"
```
BenHenning: Both of these need KDocs.

scripts/src/java/org/oppia/android/scripts/ScriptResultConstants.kt
```diff
@@ -5,4 +5,8 @@ object ScriptResultConstants {
   /** RegexPatternValidationCheck output message constants. **/
   val REGEX_CHECKS_PASSED: String = "REGEX PATTERN CHECKS PASSED"
   val REGEX_CHECKS_FAILED: String = "REGEX PATTERN CHECKS FAILED"
+
+  /** XMLSyntaxCheck output message constants **/
+  val XML_SYNTAX_CHECK_PASSED: String = "XML SYNTAX CHECK PASSED"
```
BenHenning: Maybe ``XML_SYNTAX_CHECK_PASSED_OUTPUT_INDICATOR`` might be a bit clearer on how this is being used. WDYT?

Ditto for the others.

scripts/BUILD.bazel
```diff
@@ -46,3 +46,34 @@ kt_jvm_test(
...
+    srcs = [
+        "src/java/org/oppia/android/scripts/RepoFile.kt",
+        "src/java/org/oppia/android/scripts/ScriptResultConstants.kt",
+        "src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt",
+        "src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt",
+    ],
+)
+
+kt_jvm_test(
+    name = "test_xml_syntax_check",
```
BenHenning: For this & the regex pattern: see other test targets in the codebase to ensure we naming these consistently.

This should be XmlSyntaxCheckTest (to match the name of the test).

/cc @fsharpasharp as well since tests are an intentional deviation in naming strategy so that it's easier to create grouped test commands.

scripts/BUILD.bazel
```diff
@@ -46,3 +46,34 @@ kt_jvm_test(
...
+    testonly = True,
+    srcs = ["src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt"],
+    test_class = "org.oppia.android.scripts.XMLSyntaxCheckTest",
```
BenHenning: Hmm. Could you try something: if you add an empty BUILD.bazel file in org/, org/oppia, and org/oppia/android, can you omit this test_class and still run the test?

scripts/BUILD.bazel
```diff
@@ -46,3 +46,34 @@ kt_jvm_test(
...
+    testonly = True,
+    main_class = "org.oppia.android.scripts.XMLSyntaxCheck$Companion",
+    runtime_deps = [":xml_syntax_check_lib"],
+)
+
+kt_jvm_library(
+    name = "xml_syntax_check_lib",
+    testonly = True,
+    srcs = [
+        "src/java/org/oppia/android/scripts/RepoFile.kt",
```
BenHenning: This should be in its own library & added as a dep so it can be shared between different scripts that need it. Ditto for any of the other generic classes (prefer using libraries over duplicating sources--this will make the build perform more efficiently).

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
+package org.oppia.android.scripts
+
+import org.xml.sax.ErrorHandler
+import org.xml.sax.SAXParseException
+
+/**
+ * Class for custom error handling of the parse exception thrown
+ * by the parser, to log the failure message in a cleaner way.
+ */
```
BenHenning: Here & everywhere else: KDocs should fit the full 100 character limit before being line-wrapped just like code (prefer not prematurely line-wrapping).

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+class SyntaxErrorHandler(val repoPath: String) : ErrorHandler {
```
BenHenning: Does ``repoPath`` need to be public? In general, prefer to keep everything with as minimal visibility as necessary to simplify your public contract (put another way: everything that's public needs to be tested thoroughly since other components can use them).

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+  override fun warning(e: SAXParseException) {
```
BenHenning: It seems that this class is specific to XML. Consider naming it: XmlSyntaxErrorHandler so that it's not confused among other error handling happening for other scripts.

This might also be a good indication that we need to do further subpackaging. What do you think about:

org.oppia.android.apps.scripts.regex
org.oppia.android.apps.scripts.xml
org.oppia.android.apps.scripts.common <-- common utilities shared between the scripts

For organizing the scripts?

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+  override fun fatalError(e: SAXParseException) {
+    logCheckFailure(e, repoPath)
+  }
+
+  /**
+   * Logs the failures for XML syntax validation.
+   *
+   * @param e the parsing exception thrown by the parser
+   * @param repoPath the path of the repo to be analyzed
+   * @return log the failure
```
BenHenning: Is this correct? Nothing seems to be returned. (If the return value is Unit or Nothing then ``@return`` shouldn't be in the KDoc).

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+   */
+  private fun logCheckFailure(parseException: SAXParseException, repoPath: String) {
+    val failureMessage =
```
BenHenning: I suggest formatting this message similarly to the version suggested in #3340.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
BenHenning: I apparently need to reopen this since GitHub now longer allows multiple conversations for a single line of code (see: https://github.community/t/please-re-add-multiple-comments-per-line/187397).

Nit on naming: ``XmlSyntaxCheck`` per camel-casing rules: https://developer.android.com/kotlin/style-guide#camel_case

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Script for ensuring that all the XML files
+ * in the repo are syntactically correct.
+ */
+class XMLSyntaxCheck {
+  companion object {
+    @JvmStatic
+    fun main(vararg args: String) {
+      // path of the repo to be analyzed.
```
BenHenning: Please fix single line comment casing & line wrapping per earlier comments here & in #3340.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Check if a file has a broken XML syntax.
+   *
+   * @param docBuilder the builder which will parse the XML file
+   * @param file the file to be checked for
+   * @return whether a file is syntactically correct or not
+   */
+  private fun checkIfFileHasBrokenXml(docBuilder: DocumentBuilder, file: File): Boolean {
+    try {
+      docBuilder.parse(file)
```
BenHenning: Should we maybe create a new builder for each file instead of sharing them? I'm not sure whether the builders are meant to be reused like this (or are they?).

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,71 @@
...
+    } catch (e: SAXParseException) {
```
BenHenning: I wonder if maybe a cleaner way of handling this is to collect all the errors from the error handler, and if any are present then we determine it's a failure & log them. This keeps the logging & the logic in the same file since the error handler just becomes a cataloging utility, and it avoids using exceptions for control flow (which is considered a bad practice in Effective Java).

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
BenHenning: Also, please add tests for this file (but I suggest looking at my comment in the main script class first since this class's contract could be simplified which would result in simpler tests).

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
```diff
@@ -0,0 +1,71 @@
...
+import javax.xml.parsers.DocumentBuilderFactory
+
+/**
+ * Script for ensuring that all the XML files
+ * in the repo are syntactically correct.
+ */
+class XMLSyntaxCheck {
+  companion object {
+    @JvmStatic
+    fun main(vararg args: String) {
```
BenHenning: For this & the regex check script: the main function should be documented as well. See my PR for an idea on how to do this--it's nice to document the actual usage of the script in the KDoc of main() since that's essentially how to "call" it.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    java.lang.System.setOut(PrintStream(outContent))
+  }
+
+  @After
+  fun restoreStreams() {
+    java.lang.System.setOut(originalOut)
+  }
+
+  @Test
+  fun testXmlSyntax_validXML_xmlSyntaxIsCorrect() {
```
BenHenning: ``validXml``

Ditto elsewhere (make sure 'xml' is following expected camel case naming rules for all identifiers where it's used, including variables).

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    val validXML =
+      "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n" +
+        "<shape xmlns:android=\"http://schemas.android.com/apk/res/android\"\n" +
+        "  android:shape=\"rectangle\">\n" +
+        "  <solid android:color=\"#3333334D\" />\n" +
+        "  <size android:height=\"1dp\" />\n" +
+        "</shape>"
```
BenHenning: Maybe use ``"""`` syntax, instead? Ditto for the other tests.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    val tempFile = tempFolder.newFile("testfiles/TestFile.xml")
+    tempFile.writeText(invalidXML)
+
+    val exception = assertThrows(Exception::class) {
+      runScript()
+    }
+
+    assertThat(exception).hasMessageThat().contains(
+      ScriptResultConstants.XML_SYNTAX_CHECK_FAILED
+    )
```
BenHenning: This can fit on one line--suggest doing that here & elsewhere (including #3340).

Also, maybe directly import the constant to avoid needing to qualify it (though that may be eliminated, anyway, if we stop using the constants object).

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    assertThat(outContent.toString().trim()).isEqualTo(
+      "XML syntax error: The content of elements must consist of well-formed" +
+        " character data or markup.\n" +
+        "lineNumber: 4\n" +
+        "columnNumber: 4\n" +
+        "File: [ROOT]/testfiles/TestFile.xml"
```
BenHenning: I'm starting to change my thinking on this. Instead of ``//`` maybe we should actually just fully qualify it (e.g. absolute path).

For the tests, you can just use multiple ``contains`` checks instead of validating the exact output, and the contains can just verify ``testfiles/TestFile.xml`` is present. Ditto for other tests & #3340.

WDYT?

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    )
+  }
+
+  @Test
+  fun testXmlSyntax_wrongClosingTag_xmlSyntaxIsInCorrect() {
```
BenHenning: ``xmlSyntaxIsIncorrect``

Ditto elsewhere.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    val exception = assertThrows(Exception::class) {
+      runScript()
+    }
+
+    assertThat(exception).hasMessageThat().contains(
+      ScriptResultConstants.XML_SYNTAX_CHECK_FAILED
+    )
+    assertThat(outContent.toString().trim()).isEqualTo(
+      "XML syntax error: The end-tag for element type \"shape\" must end" +
+        " with a '>' delimiter.\n" +
```
BenHenning: This error actually seems wrong--shouldn't it be mentioning that the closing tag is missing or wrong rather than missing a ``>``?

BenHenning (changes requested):
Thanks @Sparsh1212! Left a bunch of thoughts. Note that a lot of these comments also apply to #3340.

scripts/BUILD.bazel
BenHenning: Following up: you can simplify this a bit (and the extra BUILD files aren't needed since I think this directory structure is better fitting what Bazel expects). See:

```bazel
kt_jvm_test(
    name = "ComputeAffectedTestsTest",
    srcs = ["ComputeAffectedTestsTest.kt"],
    deps = [
        ":compute_affected_tests_lib",
        "//third_party:com_google_truth_truth",
        "//third_party:org_jetbrains_kotlin_kotlin-test-junit",
    ],
)
```

That works for me when running locally:

```shell
bazel test //scripts/src/java/org/oppia/android/scripts/ci:ComputeAffectedTestsTest
```

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: Yes, this might seem a bit incorrect, but the same error message can be reproduced on several of the online XML validators too. When we do the same syntax error, all of them report the same error message: 
"The end-tag for element type "shape" must end with a '>' delimiter."
https://www.xmlvalidation.com/index.php?id=1&L=0
https://www.liquid-technologies.com/online-xml-validator
https://www.truugo.com/xml_validator/
https://onlinexmltools.com/validate-xml

Will it be fine, to go with this? as it seems pretty difficult to dig on how we can change this, as this is reported by the parser.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: This sounds neat to me. Here is what I am thinking, we will maintain a mutable list of errors in the `XmlSyntaxErrorHandler`, and whenever any exception is raised by the parser, we will simply add it to the list. After the parsing is done, we will just collect the list of all the errors received by the error handler and then log them and fail the script if it's the case. How does this sound?

Also, one thing that we have to take care of is that, for every `fatalError` in the Xml syntax, if the override `fatalError` method of our error handler isn't throwing any exception,  then the default error handler throws a `SaxParseException`. So, in order to make our script continue to work without being terminated in between, we have to wrap the parsing line of code in a try-catch block, just like this:
```kotlin
try {
 builder.parse('path_to_xml')
} catch (e: SaxParseException){}
```
Please let me know your thoughts on this overall approach.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: > Yes, this might seem a bit incorrect, but the same error message can be reproduced on several of the online XML validators too. When we do the same syntax error, all of them report the same error message:
> "The end-tag for element type "shape" must end with a '>' delimiter."
> https://www.xmlvalidation.com/index.php?id=1&L=0
> https://www.liquid-technologies.com/online-xml-validator
> https://www.truugo.com/xml_validator/
> https://onlinexmltools.com/validate-xml
> 
> I think it might be fine to go with this.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
BenHenning: Both sound good to me, and I agree with the overall approach being proposed here.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
BenHenning: Yeah, I don't think we have any choice here. I just wanted to make sure that this is working as intended, even if it isn't ideal. Thanks for clarifying.

.github/workflows/static_checks.yml
Sparsh1212: Done.

scripts/BUILD.bazel
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/ScriptResultConstants.kt
Sparsh1212: Added.

scripts/src/java/org/oppia/android/scripts/ScriptResultConstants.kt
Sparsh1212: SGTM. DONE.

scripts/BUILD.bazel
Sparsh1212: Done.

scripts/BUILD.bazel
Sparsh1212: Updated.

scripts/BUILD.bazel
Sparsh1212: Outdated.

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: Done. noted.

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: Outdated. Tests added too.

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: SGTM. DONE.

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: Removed. noted.

scripts/src/java/org/oppia/android/scripts/SyntaxErrorHandler.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: I think it's fine to use a single builder, as we are doing nothing more than checking an Xml file can be parsed successfully or not.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Implemented.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Done.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: Done.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: Done.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: Done.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: SGTM. DONE.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: Done.

scripts/src/test/org/oppia/android/scripts/XMLSyntaxCheckTest.kt
Sparsh1212: I think we can resolve this now.

Sparsh1212:
> Thanks @Sparsh1212! Left a bunch of thoughts. Note that a lot of these comments also apply to #3340.

Thanks, @BenHenning.
PR is updated after all the suggestions.
Please check.

scripts/src/java/org/oppia/android/scripts/common/ScriptResultConstants.kt
```diff
@@ -4,3 +4,7 @@ package org.oppia.android.scripts.common
 const val REGEX_CHECK_PASSED_OUTPUT_INDICATOR: String = "REGEX PATTERN CHECKS PASSED"
 /** RegexPatternValidationCheck failed indicator. */
 const val REGEX_CHECK_FAILED_OUTPUT_INDICATOR: String = "REGEX PATTERN CHECKS FAILED"
+/** XmlSyntaxCheck successfuly passed indicator. */
```
BenHenning: Nit: please add spaces between the constants such that it looks like this:

```kotlin
/** KDoc...*/
constant

/** KDoc... */
constant

...
```

scripts/src/java/org/oppia/android/scripts/common/ScriptResultConstants.kt
```diff
@@ -4,3 +4,7 @@ package org.oppia.android.scripts.common
...
+const val XML_SYNTAX_CHECK_PASSED_OUTPUT_INDICATOR: String = "XML SYNTAX CHECK PASSED"
+/** XmlSyntaxCheck failed indicator. */
+const val XML_SYNTAX_CHECK_FAILED_OUTPUT_INDICATOR: String = "XML SYNTAX CHECK FAILED"
```
BenHenning: EOF newline here also.

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
+"""
+Package for Xml syntax based checks.
```
BenHenning:
```suggestion
Package for XML syntax based checks.
```

(XML is technically capitalized when documenting it, but it should be 'Xml' when in an identifier per identifier camel-casing guidelines).

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_library", "kt_jvm_test")
+
+kt_jvm_library(
+    name = "xml_syntax_error_handler_lib",
+    testonly = True,
+    srcs = ["XmlSyntaxErrorHandler.kt"],
```
BenHenning, line 11: Why not just include this in xml_syntax_check_lib? Will it ever be used anywhere else?

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.oppia.android.scripts.common.RepositoryFile
+import org.oppia.android.scripts.common.XML_SYNTAX_CHECK_FAILED_OUTPUT_INDICATOR
+import org.oppia.android.scripts.common.XML_SYNTAX_CHECK_PASSED_OUTPUT_INDICATOR
+import org.xml.sax.SAXParseException
+import java.io.File
+import javax.xml.parsers.DocumentBuilder
+import javax.xml.parsers.DocumentBuilderFactory
+
+/**
+ * Script for ensuring that all Xml files in the repo are syntactically correct.
```
BenHenning: Ditto here & elsewhere for XML.

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Logs the failures for Xml syntax validation.
+ *
+ * @param errorList a list of all the errors collected by the error handler
+ */
+private fun logXmlSyntaxFailures(errorList: List<SAXParseException>) {
+  if (errorList.isNotEmpty()) {
+    errorList.forEach { error ->
+      val failureMessage =
+        "${error.getSystemId().removePrefix("file:")}:${error.getLineNumber()}:${error.getColumnNumber()}: ${error.message}"
```
BenHenning: Hmm parsing the system ID like this seems a bit hacky. I wonder if we could instead create a separate error handler for each file (by using a separate document builder) so that we know exactly which file corresponds to each error? While a bit more cumbersome, that relies less on an identifier happening to include the filename (which sounds like something that could potentially change in the future).

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandler.kt
```diff
@@ -0,0 +1,33 @@
+package org.oppia.android.scripts.xml
+
+import org.xml.sax.ErrorHandler
+import org.xml.sax.SAXParseException
+
+/**
+ * Class for custom error handling of the parse exceptions thrown by the parser. It collects all
+ * the syntax errors thrown by the parser, which can then later be retrieved.
```
BenHenning: Could be made a bit more concise I think:

```suggestion
 * Custom XML parser error handler which collects syntax errors as they occur for later processing.
```

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  override fun fatalError(e: SAXParseException) {
+    syntaxErrorList.add(e)
+  }
+
+  /**
+   * Retrieves all the errors collected by the handler.
+   *
+   * @return List<SAXParseException> a list of all the errors collected by the error handler
```
BenHenning:
```suggestion
   * @return a list of all the errors collected by the error handler
```

We don't specify the type in `@return` clauses.

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    val tempFile = tempFolder.newFile("testfiles/TestFile.xml")
+    tempFile.writeText(validXml)
+    val docBuilder = builderFactory.newDocumentBuilder()
+    val xmlSyntaxErrorHandler = XmlSyntaxErrorHandler()
+    docBuilder.setErrorHandler(xmlSyntaxErrorHandler)
+
+    parseXml(docBuilder = docBuilder, file = tempFile)
+
+    val errorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    assertThat(errorList.isEmpty()).isEqualTo(true)
```
BenHenning: Prefer:

```kotlin
assertThat(errorList).isEmpty()
```

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    tempFile.writeText(invalidXml)
+    val docBuilder = builderFactory.newDocumentBuilder()
+    val xmlSyntaxErrorHandler = XmlSyntaxErrorHandler()
+    docBuilder.setErrorHandler(xmlSyntaxErrorHandler)
+
+    parseXml(docBuilder = docBuilder, file = tempFile)
+
+    val errorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    assertThat(errorList.size).isEqualTo(1)
+    assertThat(errorList.elementAt(0).message).isEqualTo(
```
BenHenning: Bit more Kotlin-idiomatic to do ``errorList.first()`` or ``errorList.single()`` in these cases.

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    )
+    assertThat(errorList.elementAt(0).getLineNumber()).isEqualTo(6)
+    assertThat(errorList.elementAt(0).getColumnNumber()).isEqualTo(9)
+    assertThat(errorList.elementAt(0).getSystemId()).isEqualTo(
+      "file:${retrieveTestFilesDirectoryPath()}/TestFile.xml"
+    )
+  }
+
+  @Test
+  fun testXmlErrorHandler_multipleinvalidXmlFiles_errorHandlerShouldCollectAllErrors() {
```
BenHenning:
```suggestion
  fun testXmlErrorHandler_multipleInvalidXmlFiles_errorHandlerShouldCollectAllErrors() {
```

Typo in test name.

BenHenning (commented):
Thanks @Sparsh1212. Took another pass on the PR.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: Update: We are now using a new builder for every file.

scripts/src/java/org/oppia/android/scripts/common/ScriptResultConstants.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
Sparsh1212: Ah! this lib is also used in `XmlSyntaxErrorHandlerTest`, so that's why I decided to separate it out.

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandler.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandler.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Done

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: This is outdated, as we are using a separate error handler for every file, so we are not using the same handler for multiple files.

Sparsh1212:
Thanks, @BenHenning.
Changes done.

scripts/src/java/org/oppia/android/scripts/common/ScriptResultConstants.kt
Sparsh1212: Done

BenHenning:
@Sparsh1212 please assign back to me once this PR's chain (& subsequently this PR) has been updated to be based off of #3374.

Sparsh1212:
@BenHenning This PR's directory structure is now updated based on #3374.
PTAL.

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
BenHenning: Got it. In that case, let's strike "_lib" from the library name since it's not needed here. We only use that for the root script libraries since they have binary targets with the same name which could introduce confusion in the build graph. By default, we don't include "_lib" since it's redundant.

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+are syntactically correct.
+"""
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_library")
+
+kt_jvm_library(
+    name = "xml_syntax_error_handler_lib",
+    testonly = True,
+    srcs = ["XmlSyntaxErrorHandler.kt"],
+    visibility = ["//scripts:oppia_script_binary_visibility"],
```
BenHenning: Could this be ``oppia_script_test_visibility`` instead?

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
```diff
@@ -0,0 +1,92 @@
...
+
+  searchFiles.forEach { file ->
+    val docBuilder = builderFactory.newDocumentBuilder()
+    val xmlSyntaxErrorHandler = XmlSyntaxErrorHandler()
+    docBuilder.setErrorHandler(xmlSyntaxErrorHandler)
+    parseXml(docBuilder, file)
+    val fileErrorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    if (fileErrorList.isNotEmpty()) {
+      fileErrorList.forEach { error ->
+        allErrorsList.add(Pair(error, file))
```
BenHenning: It'd be more Kotlin-idiomatic to compute the errors list as a result of this overall expression (similar to how we use fold/reduce/map/flatMap elsewhere). I suggest trying to rearrange this logic to compute the resulting list rather than mutating it within lambdas.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxCheckTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    tempFile.writeText(invalidXml)
+
+    val exception = assertThrows(Exception::class) {
+      runScript()
+    }
+
+    assertThat(exception).hasMessageThat().contains(XML_SYNTAX_CHECK_FAILED_OUTPUT_INDICATOR)
+    assertThat(outContent.toString().trim()).isEqualTo(
+      """
+      ${retrieveTestFilesDirectoryPath()}/TestFile.xml:6:8: The end-tag for element type "shape" must end with a '>' delimiter.
```
BenHenning: Here & elsewhere: we should try to avoid exceeding the column limit, and in this case I think it's avoidable. Consider changing this to:

```kotlin
assertThat(outContent.toString()).contains("${retrieveTestFilesDirectoryPath()}/TestFile1.xml:6:8: The end-tag")
assertThat(outContent.toString()).contains("The end-tag for element type "shape" must end with a '>' delimiter.")
```

In this case I don't think it's important to necessarily verify the exact error since that's dependent on the XML parser, but it is important to verify the formatting & presence of the error since that's influenced by code we maintain.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    val tempFile = tempFolder.newFile("testfiles/TestFile.xml")
+    tempFile.writeText(invalidXml)
+    val docBuilder = builderFactory.newDocumentBuilder()
+    val xmlSyntaxErrorHandler = XmlSyntaxErrorHandler()
+    docBuilder.setErrorHandler(xmlSyntaxErrorHandler)
+
+    parseXml(docBuilder = docBuilder, file = tempFile)
+
+    val errorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    assertThat(errorList.size).isEqualTo(1)
```
BenHenning:
```suggestion
    assertThat(errorList.size).hasSize(1)
```

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    docBuilder.setErrorHandler(xmlSyntaxErrorHandler)
+
+    parseXml(docBuilder = docBuilder, file = tempFile)
+
+    val errorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    assertThat(errorList).isEmpty()
+  }
+
+  @Test
+  fun testXmlErrorHandler_invalidXml_errorListShouldBeEmpty() {
```
BenHenning: Suggest adding tests for multiple errors, and for warning/error/fatal error cases (consider that if any lines of code in XmlSyntaxErrorHandler are commented out then at least 1 test should fail).

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Parses a given XML file.
+   *
+   * @param docBuilder the builder which will parse the XML file
+   * @param file the file to be checked for
+   */
+  private fun parseXml(docBuilder: DocumentBuilder, file: File) {
+    try {
+      docBuilder.parse(file)
+    } catch (e: SAXParseException) { }
```
BenHenning: I suggest adding a comment both here within the catch block & in the production version to explain why it's okay & correct to ignore the caught exception.

BenHenning (changes requested):
Thanks @Sparsh1212! PR generally LGTM--I think the latest iteration of the checker is really clear & easy to follow. My main concern is that we need to verify the new tests are passing in CI which is blocked on the issues #3340 is running into.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Ah! I just dug a bit deep into this and realized that, for our use case, only `fatalError` is essential to us. Because, whenever there is any syntactically incorrect XML present, then always a `fatalError` is thrown, and for this case as we are only concerned of the syntax error hence, all our errors in testing will be of type `fatalError` only. 
Also, there won't be the case of multiple failures in a single XML file because, as soon as the parser encounters a `fatalError` it stops the further processing of XML.
Hence, it seems like we won't need tests for multiple failures as we will never get more than a single error and also the use cases of `warning` and `error` are irrelevant for our case. 
I was probably thinking of removing them from the syntax error handler but it is compulsory to override them.

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/xml/BUILD.bazel
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/xml/XmlSyntaxCheck.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxCheckTest.kt
Sparsh1212: Revamped the tests to check output format, pass status and failures.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Added.

Sparsh1212:
> Thanks @Sparsh1212! PR generally LGTM--I think the latest iteration of the checker is really clear & easy to follow. My main concern is that we need to verify the new tests are passing in CI which is blocked on the issues #3340 is running into.

Addressed all the nits. PTAL.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Added tests for all cases. PTAL.

.github/workflows/static_checks.yml
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/XMLSyntaxCheck.kt
Sparsh1212: This is corrected now.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    xmlSyntaxErrorHandler.warning(SAXParseException("test_error_message", "", "", 1, 1))
+    val errorList = xmlSyntaxErrorHandler.retrieveErrorList()
+    assertThat(errorList).hasSize(1)
+    assertThat(errorList.first().message).isEqualTo("test_error_message")
+    assertThat(errorList.first().getLineNumber()).isEqualTo(1)
+    assertThat(errorList.first().getColumnNumber()).isEqualTo(1)
+  }
+
+  @Test
+  fun testXmlErrorHandler_invokeErrorCase_errorShouldBeCollected() {
```
BenHenning, line 89: It feels a bit like fatal error is missing here. Maybe add a separate test for it even though it's technically covered in parsing test above (consider that implementation changes may result in the above error no longer being a fatal error which technically would reduce behavioral coverage in this test suite).

BenHenning (commented):
Thanks @Sparsh1212! Same as #3340. This is effectively 'LGTM' with one more comment to resolve. Please resolve that & assign back once all CI tests are passing in this PR.

scripts/src/javatests/org/oppia/android/scripts/xml/XmlSyntaxErrorHandlerTest.kt
Sparsh1212: Added.

BenHenning (commented):
Thanks @Sparsh1212! LGTM, but waiting to approve until #3340 is merged. Also, per our discussion I'm cancelling the workflows for this & the next PR since that will relieve pressure on the CI system and once you merge #3340 into develop and update these PRs they should have much fewer tests to run..

BenHenning (approved):
Thanks @Sparsh1212! CI is green so going ahead and merging this.

PR #3374, opened 2021-06-26
Fix part of #2691: Migrate compute affected tests script to Kotlin

BenHenning:
Fix part of #2691

## Explanation
This PR introduces a new version of the compute_affected_tests script that exists in Kotlin instead of Bash.

The main reason for doing this is the new script includes code to fix situations where Bazel outputs malformed query results (e.g. when two targets are on the same line). It's not yet clear whether these issues are caused by Bazel or Bash, but they only happen in CI (i.e. can't be reproduced locally) on occasional PRs (such as #3340). This version of the script builds in resilience to such failures.

There are a number of other benefits that we see moving to Kotlin, as well:
- It's easier to test the script
- The script is normalizing on an existing language well understood in the codebase (Kotlin) vs. using an alternative language (such as Python)
- The script is easier to understand since advanced Bash syntax is hard to read and understand vs. Kotlin which is a much more self-documenting language
- The script establishes the foundation for introducing more Kotlin-based scripts in the future

Beyond that, three other things to note:
- I noticed that the per-file query loop in the original Bash script can be simplified to a single command which seems to be more performant. Beyond that, the script overall seems much faster through Bazel vs. Bash (though CI will be the true test, especially since the binary also needs to be build via Bazel itself).
- The linters have been updated to include the scripts directory as actual code
- I adjusted the output of the ktlint command to include a copy-and-pastable line for fixing errors based on the version of ktlint used to report the errors
- The command runs in a different directory than it did with Bash hence the updated usage arguments. This is actually safer from a sandboxing perspective as it reduces the chance of the script having accidental side effects in the repository.
- The original command may have had a bug in it when an empty set is passed for BUILD files (such as if the script is run with no files changed). This has been addressed in the new script.
- The original script would sometimes hang on CI for unknown reasons. While this could still happen, additional debug lines have been added to the script to help with this & other issues as they crop up. Further, the new script has a timeout if subcommands take longer than 60 seconds, so CI should never indefinitely hang due to the script now.
- The new script is actually testable, and the new tests are creating a real Bazel & Git repository locally in order to simulate different situations.
- The new script is slower since it's no longer in Bash, but it seems fast enough for our purpose (it actually seems to be comparable, but this PR is simpler from a dependency graph perspective than some PRs we expect this to run in).

Long-term, we'll likely want to move all of our scripts over to Kotlin to reduce the total number of languages needed for the repository by 1.

## Checklist


BenHenning:
@Sparsh1212 are you still hitting the problem we discussed in #3340? It might be nice to do a trial run on your PR, but that might be tricky since the code needs to be copied into the branch and uploaded to GitHub.

Keeping this as a draft until I write tests, though keep in mind that tests are particularly hard for this script since it's highly dependent on a real Bazel environment.

Sparsh1212:
@BenHenning I copied the code from `regex-checks` branch to this branch and initiated a draft PR to see the CI checks are working properly or not.
PR: #3375

Sparsh1212:
The parsing of the added `kt_jvm_test`: `test_pattern_validation_check` is now correctly taking place. Also, the tests are passing as well.

BenHenning:
Ah excellent--thanks @Sparsh1212. In that case, this just needs tests & can then go into review.

Sparsh1212:
> Ah excellent--thanks @Sparsh1212. In that case, this just needs tests & can then go into review.

Thanks! for the fix Ben!

BenHenning:
Added tests for the core script. Still need to refactor, added more documentation, and add tests for the test & script utilities.

BenHenning:
Also interesting: I changed the merge-base command to use ``develop`` instead of ``origin/develop`` since the test repository doesn't have an origin, but that doesn't work in the CI environment. The CI script probably needs to check out the develop branch so that it exists locally.

BenHenning:
Unfortunately the extra checkout action didn't work. We may need to pass the reference object for merge-base in as a parameter unfortunately.

BenHenning:
Okay, I think the script is fully ready for review now since it seems to be operating as expected, and it + its constituent components are now fully documented & tested.

BenHenning:
PTAL @vinitamurthi @anandwana001 & @Sparsh1212 (FYI that this blocks Sparsh's milestone 1 work since one of his PRs isn't running on CI correctly due to issues with the existing Bash version of the script, and this arranges some of the script directory components which will help normalize patterns in downstream script PRs for his & other GSoC projects).

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
```diff
@@ -0,0 +1,231 @@
...
+    check(libTargetName !in libraryFileMap) { "Library '$dependencyName' already exists" }
+    val prereqFiles = ensureWorkspaceIsConfiguredForKotlin()
+    prepareBuildFileForLibraries(rootBuildFile)
+    return rootBuildFile.appendingPrintWriter().use { writer ->
+      val depFile = temporaryRootFolder.newFile("$dependencyName.kt")
+      libraryFileMap[libTargetName] = depFile
+      writer.println("kt_jvm_library(")
+      writer.println("    name = \"$libTargetName\",")
+      writer.println("    srcs = [\"${depFile.name}\"],")
+      writer.println(")")
```
Sparsh1212: or maybe, can we do this?
```kotlin
val ktJvmLibrary = 
"""
kt_jvm_library(
    name = "$libTargetName",
    srcs = ["${depFile.name}"],
)
"""
writer.println(ktJvmLibrary)
```
This might look a bit neat. Ditto elsewhere.

Sparsh1212:
@BenHenning 
The CI checks are failing for some of the `javatests`. PTAL.

anandwana001 (changes requested):
Everything seems reasonable, just one of the test cases is failing in `BazelClientTest `.

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
...
+kt_jvm_library(
+    name = "compute_affected_tests_lib",
+    testonly = True,
+    srcs = [
+        "ComputeAffectedTests.kt",
+    ],
+    visibility = ["//scripts:oppia_script_binary_visibility"],
+    deps = [
+      "//scripts/src/java/org/oppia/android/scripts/common:bazel_client",
+      "//scripts/src/java/org/oppia/android/scripts/common:git_client",
```
anandwana001:
```suggestion
        "//scripts/src/java/org/oppia/android/scripts/common:bazel_client",
        "//scripts/src/java/org/oppia/android/scripts/common:git_client",
```

This will fix the Bazel lint error

anandwana001 (changes requested):

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
BenHenning: Nice idea--this also let me get rid of the ``appendingPrintWriter`` helper function.

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
BenHenning: Done.

BenHenning:
Interesting. The failing test actually may be the same timeout issue we're seeing for certain BUILD file-based PRs for the script in CI. We may have found a deterministic means of reproducing the test in CI (though probably not yet locally unfortunately). I'm going to upload the latest changes to try and see if the test failure repros.

BenHenning:
Looks like this might be consistent: https://github.com/oppia/oppia-android/pull/3374/checks?check_run_id=2976768230.

BenHenning:
I was able to isolate the problem. I've added context to https://github.com/bazelbuild/bazel/issues/906. I might need to file a new issue since that one is old, but I'm really interested in seeing whether the isolated repository will make it easier to investigate the issue.

In the meantime, I'll disable the test with a TODO to unblock this & downstream work.

BenHenning:
Now that the test is ignored, I expect all remaining tests to pass. Will reassign once I confirm the failing suite in question is now passing on CI.

BenHenning:
BazelClientTest is now passing: https://github.com/oppia/oppia-android/pull/3374/checks?check_run_id=2976899202. Assigning this back over since the PR should have a fully green CI now.

BenHenning:
@Sparsh1212 / @anandwana001 PTAL at my follow-up comments & mark as resolved if you're happy with the changes made.

vinitamurthi (approved):
LGTM!

Sparsh1212:
LGTM!

Sparsh1212 (approved):

anandwana001 (approved):
LGTM

BenHenning:
Thanks all! I might follow up with a workaround to the issue with the query hanging (I have an idea that might mitigate it).

PR #3408, opened 2021-07-05
Fix #3325: Domain layer mechanism for saving checkpoints.

### LINKED ISSUE #3325: Implement domain layer mechanism to save checkpoints
Modify the class of the domain layer to implement checkpointing. 

Target | Date
--|--
**PR creation:** | 2021-06-24
**PR completion:** | 2021-06-29

MaskedCarrot:
## Explanation

Fixes #3325 

## Checklist

## Tests

test class | Screenshot
--- | ---
ExplorationActivityTest (Tests ignored here are due to an existing issue.) | Screenshot from 2021-07-07 22-36-43


MaskedCarrot:
@BenHenning please take a look

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -343,6 +426,11 @@ class ExplorationProgressController @Inject constructor(
           "Cannot navigate to a next state if an answer submission is pending."
         }
         explorationProgress.stateDeck.navigateToNextState()
+
+        // only mark checkpoint if current state is pending state. This ensures that checkpoint will
+        // not be marked on any of the completed states.
+        if (explorationProgress.stateDeck.isCurrentStateTopOfDeck())
+          saveExplorationCheckpoint()
```
BenHenning, line 389: Separate point, but please make sure to add braces for single-line cases like this. We only omit braces if the entire if + statement fits on the same line. See: https://developer.android.com/kotlin/style-guide#braces.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +440,112 @@ class ExplorationProgressController @Inject constructor(
...
+        explorationProgress.currentExploration.title,
+        oppiaClock.getCurrentTimeMs()
+      )
+
+    CoroutineScope(backgroundDispatcher).launch {
+      val result = explorationCheckpointController.recordExplorationCheckpoint(
+        profileId,
+        explorationId,
+        checkpoint
+      ).retrieveData()
```
BenHenning: This seems hacky. We shouldn't call retrieveData() unless we actually want to observe the result (& we probably should here--consider that any failures in this block of code will likely silently fail and not be surfaced to the application or user which will result in broken behavior that we can't investigate). In the very least, we should always log the result of a background operation.

I don't have anything against this idea in principle (that we want to entirely save the checkpoint in the background), but if it's important to observe the result then we should probably combine it with one of the existing DataProviders already being observed by StateFragment. Further, the operation to call recordExplorationCheckpoint itself isn't particularly expensive--it's already deferring execution to a background thread. To simplify, consider:
- Calling recordExplorationCheckpoint directly rather than deferring to the background dispatcher
- Update recordExplorationCheckpoint to return a Deferred (we should also make sure that this method cannot be called by the UI layer which might require some refactoring of your controller, maybe into multiple classes)
- Observe the deferred here; when it finishes you can pull the save state & notify that the ephemeral state DataProvider has an update
- Update EphemeralState to include details on the current checkpoint save status
- In the async method for the ephemeral state DataProvider, have it compute the EphemeralState such that it includes the current save state
- Use the new value in StateFragment to make decisions based on when the user wants to exit the exploration

I think the results in a net simplification:
- No need for the extra threading here
- We can avoid a call to retrieveData() by just using Deferred directly since all of this is happening in the domain layer
- We can avoid the extra DataProvider to fetch the checkpoint save status
- We can avoid blocking the UI on the above DataProvider to determine what to show the user when they try to exit

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -249,6 +330,8 @@ class ExplorationProgressController @Inject constructor(
           solution = explorationProgress.stateGraph.computeSolutionForResult(state)
           explorationProgress.stateDeck.pushStateForSolution(state)
         } finally {
+          // mark a checkpoint in the exploration if the solution is revealed.
```
BenHenning: Note comment capitalization for consistency.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -154,6 +228,11 @@ class ExplorationProgressController @Inject constructor(
             )
           }
         } finally {
+          // If the answer was submitted on behalf of the Continue interaction, don't save
+          // checkpoint because it will be saved when the learner moves to the next state.
+          if (answerOutcome.state.interaction.id != "Continue")
```
BenHenning: This seems hacky. There are probably more elegant alternatives, but before I get into that: do we actually need to special case this? Why not just treat it like any other interaction?

BenHenning (commented):
@MaskedCarrot leaving some thoughts. PTAL.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: @BenHenning even if we return a deferred from `recordExplorationCheckpoint` function we still have to use a suspend function to wait for it to complete. So in turn we will have to use a background dispatcher.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: saving works as expected using `invokeOnCompletion` on the deferred.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: If we do this like any other interaction, then checkpoints will be saved twice, once when the answer is submitted on behalf of the continue interaction, and then the second time when the learner moves to the next state.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -166,7 +189,13 @@ class ExplorationProgressControllerTest {
   @Test
   fun testPlayExploration_invalid_returnsSuccess() {
     val resultLiveData =
-      explorationDataController.startPlayingExploration("invalid_exp_id")
+      explorationDataController.startPlayingExploration(
+        profileId.internalId,
+        "invalid_topic_id",
```
anandwana001: can we have some constants for these values
invalid_topic_id,invalid_story_id, invalid_exp_id

domain/src/test/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointControllerTest.kt
```diff
@@ -100,6 +102,9 @@ class ExplorationCheckpointControllerTest {
   @Captor
   lateinit var checkpointDetailsCaptor: ArgumentCaptor<AsyncResult<ExplorationCheckpointDetails>>
 
+  @Inject
+  lateinit var dataProviders: DataProviders
```
anandwana001: are we using this?

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -1408,13 +1779,400 @@ class ExplorationProgressControllerTest {
...
+    )
+    testCoroutineDispatchers.runCurrent()
+
+    val currentStateLiveData =
+      explorationProgressController.getCurrentState().toLiveData()
+    currentStateLiveData.observeForever(mockCurrentStateLiveDataObserver2)
+    testCoroutineDispatchers.runCurrent()
+
+    // The new observer should observe the completed second state since it's the current pending
+    // state.
```
anandwana001:
```suggestion
    // The new observer should observe the completed second state
    // since it's the current pending state.
```

anandwana001: ditto for others

anandwana001 (changes requested):
Thanks @MaskedCarrot 

Left nit suggestions
Also,
1. Add screenshot showing tests are passing on espresso for sharedTest which are UI test.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -792,7 +936,13 @@ class ExplorationActivityTest {
         FRACTIONS_EXPLORATION_ID_0
       )
     )
-    explorationDataController.startPlayingExploration(FRACTIONS_EXPLORATION_ID_0)
+    explorationDataController.startPlayingExploration(
+      internalProfileId,
+      FRACTIONS_TOPIC_ID,
+      FRACTIONS_STORY_ID_0,
+      FRACTIONS_EXPLORATION_ID_0,
+      isCheckpointingEnabled = false
```
aggarwalpulkit596: Can't we just keep the default value instead of setting it for every function call

domain/build.gradle
```diff
@@ -26,6 +26,9 @@ android {
 
   kotlinOptions {
     jvmTarget = JavaVersion.VERSION_1_8
+    freeCompilerArgs += [
+        "-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi"
```
aggarwalpulkit596, line 30: Why are we using experimental API ? @BenHenning @rt4914 can we use it ?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -14,8 +16,15 @@ private const val TERMINAL_INTERACTION_ID = "EndExploration"
  * instances, but calling code is responsible for ensuring it is properly reset.
  */
 internal class ExplorationProgress {
+  internal lateinit var currentProfileId: ProfileId
+  internal lateinit var currentTopicId: String
+  internal lateinit var currentStoryId: String
```
aggarwalpulkit596, line 21: Also, do we really need to add 3 different variables? Seems like a lot of added dependency.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -154,6 +181,12 @@ class ExplorationProgressController @Inject constructor(
             )
           }
         } finally {
+          // If the answer was submitted on behalf of the Continue interaction, don't save
+          // checkpoint because it will be saved when the learner moves to the next state.
+          if (answerOutcome.state.interaction.id != "Continue") {
```
aggarwalpulkit596: Why don't we need to save it for continue state? I'm a bit confused here.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
aggarwalpulkit596: indentation seems incorrect

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +396,110 @@ class ExplorationProgressController @Inject constructor(
...
+  private fun processSaveCheckpointResult(
+    profileId: ProfileId,
+    topicId: String,
+    storyId: String,
+    explorationId: String,
+    lastPlayedTimestamp: Long,
+    newCheckpointState: CheckpointState
+  ) {
+    explorationProgressLock.withLock {
+      if (explorationProgress.checkpointState != newCheckpointState) {
```
aggarwalpulkit596, line 476: Could you add a comment here for the underlying logic involved?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: explorationId was already present for loading the exploration, profiled, topicId and storyId are added to mark the chapter as IN_PROGRESS_SAVED/NOT_SAVED. ProfileId is also used to save checkpoints.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
MaskedCarrot: this value has to be changed to true if we are testing anything related to checkpointing. The next PR contains the changes and tests that use this.

domain/src/test/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointControllerTest.kt
MaskedCarrot: removed this.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: We are not saving checkpoint after the answer is submitted on behalf of continue interaction because once the function submitAnswer completes and the exploration goes into viewing state, the stateFragment moves the exploration to the next state without any user interaction. So saving here as we do in any other interaction will cause us to save one unnecessary checkpoint.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: indentation is auto-formated.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: Isn't that also true for other interactions? I'm not sure that I see an important distinction here for why 'Continue' should be handled differently.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: For posterity: we discussed this over chat & suggested that we should use invokeOnCompletion with getCompleted rather than a DataProvider or suspend function.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: for any other interaction, after submitting the correct answer the user has  to press the continue button to go to the next state, so the user interacts with the UI twice before reaching the next state but with states with continue interaction, the user only interacts once with the UI when they press the continue button.

so since their are no user interactions between the two checkpoints of continue interaction, we can skip saving one checkpoint

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -172,6 +186,37 @@ internal class StateDeck internal constructor(
       .build()
   }
 
+  /**
+   * @return an [ExplorationCheckpoint] which contains all the latest values of variables of
```
BenHenning: KDocs should always have a base summary fragment. Suggest just saying something like:

```kotlin
/**
 * Returns an [ExplorationCheckpoint] ...
 */
```

Also, note that spacing continuation for an at-clause is +4 spaces.

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -280,4 +264,22 @@ class ExplorationCheckpointController @Inject constructor(
 
   @VisibleForTesting(otherwise = VisibleForTesting.NONE)
   fun getExplorationCheckpointDatabaseSizeLimit(): Int = explorationCheckpointDatabaseSizeLimit
+
+  @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
BenHenning: We ought to never use VisibleForTesting. Can we test this through the methods that call it, instead? Ie, that verifies the expected production behavior: under some conditions, those public API functions should result in a checkpoint being saved.

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -69,65 +68,55 @@ class ExplorationCheckpointController @Inject constructor(
...
-   * @return a [DataProvider] that indicates the success/failure of the save operation.
-   *
-   * Success result is returned if the checkpoint is saved successfully. The success result
-   * returned has the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_EXCEEDED]
-   * if the database has exceeded the size limit of [explorationCheckpointDatabaseSizeLimit],
-   * otherwise the success result is returned with the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_NOT_EXCEEDED].
+   * @return a [Deferred] that upon completion indicates the current state of the checkpoint
+   *         database. If the size of the checkpoint database is less than the allocated limit of
```
BenHenning: Continuation is wrong here for at-clause: should be +4 from the '@'.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -47,23 +59,38 @@ class ExplorationProgressController @Inject constructor(
   //  to avoid cases in tests where the exploration load operation needs to be fully finished before
   //  performing a post-load operation. The current state of the controller is leaking this
   //  implementation detail to tests.
+  // TODO(): Update the saving mechanism for checkpoints once internal locking of this controller is
```
BenHenning: Please add an issue for this TODO or resolve it.

BenHenning (commented):
Thanks @MaskedCarrot. Took another quick pass and left a few thoughts. I didn't review the domain changes in detail, but the updated Deferred approach seems good.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: @BenHenning In the earlier approach, we had to acquire `exploration progress lock` twice, first to create a checkpoint and then again to update the `checkpointState` once the save operation has completed. This created a race condition that if the user exits the exploration after the save operation has started but has not finished, we might end up with the wrong `checkpointState`.

But with the current approach, we kick off the saving process, wait for it to complete and then update the checkpoint state without releasing the lock, so I think this eliminates the race condition. WDYT?

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: done.
I will keep this in mind in future.

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: For this PR, this function is being used to test if checkpoints are being saved, since the function to save checkpoints in the production code returns a deferred so for testing, this function creates a dataProvider from that deferred.

For the next PR this function will be used to save checkpoints so that we can have multiple checkpoints saved in the database to properly test those features of checkpointing that are not directly related with saving checkpoints.

If we remove this function, we can create `dataProviders` from the deferred in the classes `ExplorationCheckpointControllerTest` and `ExplorationCheckpointTestHelper` (added in the next PR).

@BenHenning WDYT?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +394,113 @@ class ExplorationProgressController @Inject constructor(
...
+      profileId,
+      explorationId,
+      checkpoint
+    )
+
+    deferred.invokeOnCompletion {
+      checkpointState = if (it == null) {
+        deferred.getCompleted()
+      } else {
+        oppiaLogger.e("CHECKPOINTING", "Failed to save checkpoint in exploration", it)
```
vinitamurthi: Can we make the tag a bit clearer. CHECKPOINTING to me seems vague. Also avoid keeping it in all caps

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -69,65 +68,55 @@ class ExplorationCheckpointController @Inject constructor(
...
    *
-   * @return a [DataProvider] that indicates the success/failure of the save operation.
-   *
-   * Success result is returned if the checkpoint is saved successfully. The success result
-   * returned has the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_EXCEEDED]
-   * if the database has exceeded the size limit of [explorationCheckpointDatabaseSizeLimit],
-   * otherwise the success result is returned with the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_NOT_EXCEEDED].
+   * @return a [Deferred] that upon completion indicates the current state of the checkpoint
```
vinitamurthi: Shouldn't it be a `Deferred<CheckpointState>` here?

domain/build.gradle
rt4914: @aggarwalpulkit596 Not sure about this. Will need to defer to @BenHenning on this.

rt4914 (approved):
Approving for code-owner files.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: updated CHECKPOINTING to Lightweight checkpointing

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: updated the docs

MaskedCarrot:
I have updated this branch as per the changes requested. @vinitamurthi @BenHenning @anandwana001 PTAL

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
BenHenning: Either this function is part of the class's API (which means it should be public/internal & have a KDoc explaining out to use it) in which case it'll have dedicated tests, or it's a helper for other functions in this controller & should be tested indirectly through those. In neither case should VIsibleForTesting be needed.

It seems like the latter since you have the async version of this function above. Could we use that to test this, instead?

domain/build.gradle
BenHenning: I think it's necessary here since otherwise it's quite difficult to use Deferred. We also technically use the experimental API throughout the codebase (mostly for tests), and it seems pretty stable since it's been around for a few years now. This seems fine to me.

However, @prayutsu we should also add this argument to the Bazel builds for tests & binaries, too.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: I don't think it does--we're still waiting for the save checkpoint Deferred to finish and we don't block the DataProvider on that. Things could change between submit answer and the saving completed which could introduce more data races. Ideally, we should ensure eventual consistency (i.e. regardless of the order of operations, the final outcome is still the correct one).

BenHenning (commented):
Left more thoughts @MaskedCarrot. PTAL.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: @BenHenning  I have one question out of curiosity.  We will have to acquire lock to change anything in an exploration and since we are only releasing lock once the saving process is complete, shouldn't this ensure that the there is no race condition,

MaskedCarrot:
@BenHenning PTAL

aggarwalpulkit596 (approved):

anandwana001 (approved):
Thanks @MaskedCarrot 

LGTM

Please follow the PR description and correct the PR title.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: added a TODO and filled a new issue for this #3467

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -40,25 +41,14 @@ class ExplorationCheckpointController @Inject constructor(
...
-   * such that if the deferred result contains
-   *
-   * CHECKPOINT_SAVED_DATABASE_SIZE_LIMIT_EXCEEDED, the
-   * [ExplorationCheckpointState.CHECKPOINT_SAVED_DATABASE_EXCEEDED_LIMIT] will be
-   * passed to a successful AsyncResult.
-   *
-   * CHECKPOINT_SAVED_DATABASE_SIZE_LIMIT_NOT_EXCEEDED,
-   * [ExplorationCheckpointState.CHECKPOINT_SAVED_DATABASE_NOT_EXCEEDED_LIMIT] will be
-   * passed to a successful AsyncResult.
+   * These Statuses correspond to the exception above such that if the deferred result contains
```
vinitamurthi: This statement to me is a bit unclear. What `exception above` is this referring to? Maybe instead of mentioning it that way, provide the actual exception name

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -69,69 +59,79 @@ class ExplorationCheckpointController @Inject constructor(
...
-   * if the database has exceeded the size limit of [explorationCheckpointDatabaseSizeLimit],
-   * otherwise the success result is returned with the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_NOT_EXCEEDED].
+   * @return a [Deferred] that upon completion indicates the current [CheckpointState]. If the size
+   *     of the checkpoint database is less than the allocated limit of
+   *     [ExplorationStorageDatabaseSize] then the deferred upon completion gives the result
+   *     [CheckpointState.CHECKPOINT_SAVED_DATABASE_NOT_EXCEEDED_LIMIT]. If the size of the
+   *     checkpoint database exceeded [ExplorationStorageDatabaseSize] then
+   *     [CheckpointState.CHECKPOINT_SAVED_DATABASE_EXCEEDED_LIMIT] is returned upon successful
+   *     completion of deferred
```
vinitamurthi: Indentation here seems off. Maybe follow the same pattern as the lines above (i.e. put `If the size...` onwards as a new paragraph

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -26,6 +28,8 @@ internal class StateDeck internal constructor(
   private val hintList: MutableList<Hint> = ArrayList()
   private lateinit var solution: Solution
   private var stateIndex: Int = 0
+  private var revealedHintIndex: Int = -1
```
vinitamurthi, line 32: Maybe add the same comment for why the hint index is -1 over here

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -108,6 +114,8 @@ internal class StateDeck internal constructor(
     currentDialogInteractions.clear()
     hintList.clear()
     pendingTopState = state
+    revealedHintIndex = -1
```
vinitamurthi, line 122: Add a comment as to why the hint index is -1 here

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -1408,11 +1781,304 @@ class ExplorationProgressControllerTest {
...
+
+    val result = explorationProgressController.submitSolutionIsRevealed(currentState.state)
+    result.observeForever(mockAsyncSolutionObserver)
+    testCoroutineDispatchers.runCurrent()
+
+    verifyCheckpointHasCorrectValueOfIsSolutionRevealed(
+      profileId,
+      TEST_EXPLORATION_ID_2,
+      isSolutionRevealed = true
+    )
```
vinitamurthi, line 2081: Do we need a test for the case where the player reaches 'Continue' ? I think you have some special casing for that scenario right?

model/src/main/proto/exploration_checkpoint.proto
```diff
@@ -73,3 +73,21 @@ message ExplorationCheckpointDetails{
...
+  // Progress made in the exploration is saved and the size of the checkpoint database has
+  // not exceeded the allocated limit.
+  CHECKPOINT_SAVED_DATABASE_NOT_EXCEEDED_LIMIT = 1;
+
+  // Progress made in the exploration is saved and the size of the checkpoint database has
+  // exceeded the allocated limit.
+  CHECKPOINT_SAVED_DATABASE_EXCEEDED_LIMIT = 2;
+
+  // Progress made in the exploration is not saved.
+  UNSAVED = 3;
```
vinitamurthi: Maybe CHECKPOINT_UNSAVED?

MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: changed the kdoc explanation.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: I don't think we need any additional test case for that because in the case of `Continue` we have just optimized a bit by saving one checkpoint instead of two.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
```diff
@@ -44,9 +44,21 @@ class ExplorationDataController @Inject constructor(
    * @return a one-time [LiveData] to observe whether initiating the play request succeeded. The exploration may still
    *     fail to load, but this provides early-failure detection.
    */
-  fun startPlayingExploration(explorationId: String): LiveData<AsyncResult<Any?>> {
+  fun startPlayingExploration(
+    internalProfileId: Int,
+    topicId: String,
+    storyId: String,
+    explorationId: String,
+    isCheckpointingEnabled: Boolean
```
BenHenning: Please add these as ``@param``s to the KDoc above.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,114 @@ class ExplorationProgressController @Inject constructor(
...
+   *  checkpoint with the latest progress and saves it using [ExplorationCheckpointController].
+   *
+   *  This function also waits for the save operation to complete, upon completion this function
+   *  uses the function [processSaveCheckpointResult] to mark the exploration as
+   *  IN_PROGRESS_SAVED or IN_PROGRESS_NOT_SAVED depending upon the result.
+   */
+  private fun saveExplorationCheckpoint() {
+    // Do not save checkpoints if checkpointing is not enabled. This is expected to happen when
+    // the current exploration has been already completed previously.
+    if (!explorationProgress.isLightweightCheckpointingEnabled) return
```
BenHenning: Maybe it might be clearer to instead pass in a value that indicates whether exploration progress should be saved, such as ``shouldSavePartialProgress``. The current name reads more as "is this feature enabled" rather than providing a condition under which the feature behaves differently. Ditto everywhere (including the parameter name for startPlayingExploration).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,114 @@ class ExplorationProgressController @Inject constructor(
...
+        storyId,
+        explorationId,
+        oppiaClock.getCurrentTimeMs(),
+        checkpointState
+      )
+    }
+  }
+
+  /**
+   * This function process the result obtained upon complete execution of the function
```
```suggestion
   * Processes the result obtained upon complete execution of the function
```

Note the verb tense and the fact that the initial sentence of a method's KDoc summary is a sentence fragment. Ditto elsewhere.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,114 @@ class ExplorationProgressController @Inject constructor(
...
+   * [saveExplorationCheckpoint].
+   *
+   * Marks the exploration as in_progress_saved or in_progress_not_saved if it is not already marked
+   * correctly. This function also updates the checkpoint state of the exploration to the
+   * specified new checkpoint state.
+   *
+   * @param profileId is the profile id currently playing the exploration.
```
BenHenning: Nit: at-clause summaries don't end in punctuation unless they have multiple snetences.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,114 @@ class ExplorationProgressController @Inject constructor(
...
+      )
+
+    val deferred = explorationCheckpointController.recordExplorationCheckpointAsync(
+      profileId,
+      explorationId,
+      checkpoint
+    )
+
+    deferred.invokeOnCompletion {
+      checkpointState = if (it == null) {
```
BenHenning: Suggest defining ``val`` here rather than using a ``lateinit var`` (the latter doesn't seem necessary in this context).

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -69,69 +59,79 @@ class ExplorationCheckpointController @Inject constructor(
...
-   * if the database has exceeded the size limit of [explorationCheckpointDatabaseSizeLimit],
-   * otherwise the success result is returned with the value
-   * [ExplorationCheckpointDatabaseState.CHECKPOINT_DATABASE_SIZE_LIMIT_NOT_EXCEEDED].
+   * @return a [Deferred] that upon completion indicates the current [CheckpointState].
+   *     If the size of the checkpoint database is less than the allocated limit of
+   *     [ExplorationStorageDatabaseSize] then the deferred upon completion gives the result
+   *     [CheckpointState.CHECKPOINT_SAVED_DATABASE_NOT_EXCEEDED_LIMIT]. If the size of the
+   *     checkpoint database exceeded [ExplorationStorageDatabaseSize] then
+   *     [CheckpointState.CHECKPOINT_SAVED_DATABASE_EXCEEDED_LIMIT] is returned upon successful
+   *     completion of deferred
```
BenHenning: Nit: end with punctuation.

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -172,6 +187,37 @@ internal class StateDeck internal constructor(
...
+    return explorationCheckpointBuilder
+      .setPendingStateName(pendingTopState.name)
+      .setHintIndex(revealedHintIndex)
+      .setSolutionIsRevealed(solutionIsRevealed)
+      .addAllPendingUserAnswers(currentDialogInteractions)
+      .setStateIndex(stateIndex)
+      .setExplorationVersion(explorationVersion)
+      .setExplorationTitle(explorationTitle)
+      .setTimestampOfFirstCheckpoint(timestamp)
+      .build()
```
BenHenning:
```suggestion
    return ExplorationCheckpoint.newBuilder().apply {
      addAllCompletedStatesInCheckpoint(previousStates.map { state ->
        CompletedStateInCheckpoint.newBuilder().apply {
          completedState = state.completedState
          stateName = state.state.name
        }.build()
      })
      pendingStateName = pendingTopState.name
      hintIndex = revealedHintIndex
      this.solutionIsRevealed = solutionIsRevealed
      addAllPendingUserAnswers(currentDialogInteractions)
      this.stateIndex = stateIndex
      this.explorationVersion = explorationVersion
      this.explorationTitle = explorationTitle
      timestampOfFirstCheckpoint = timestamp
    }.build()
```

More Kotlin-idiomatic (per the ``map``), and conveniently in one expression to avoid saving the builder to a val (per the ``apply``).

domain/src/test/java/org/oppia/android/domain/topic/StoryProgressControllerTest.kt
```diff
@@ -115,6 +115,34 @@ class StoryProgressControllerTest {
...
+      profileId,
+      FRACTIONS_TOPIC_ID,
+      FRACTIONS_STORY_ID_0,
+      FRACTIONS_EXPLORATION_ID_0,
+      fakeOppiaClock.getCurrentTimeMs()
+    ).toLiveData().observeForever(mockRecordProgressObserver)
+    testCoroutineDispatchers.runCurrent()
+
+    verifyRecordProgressSucceeded()
+  }
```
BenHenning: There are a lot of branches in recordChapterAsInProgressNotSaved. Please make sure to add more tests to cover all of the different scenarios in which that method runs to ensure it's well covered.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: @MaskedCarrot that would be the case if not for the gap. The current flow is essentially:

- Main thread (from UI class): perform some state change (such as submitting an answer)
- Main thread (from progress controller): initiate save; acquire lock
- Main thread (from checkpoint controller): call storeDataWithCustomChannelAsync
- Main thread (from progress controller): call invokeOnCompletion
- Main thread (from progress controller): return back to UI class with DataProvider corresponding to operation; release lock (not exactly; the critical section is smaller than indicated here but this better illustrates the point)
- Background thread A (from checkpoint controller): initiate save operations, then finish
- Background thread A (from progress controller via invokeOnCompletion callback): acquire lock, retrieve results & notify UI layer that the state data provider has changed, release lock

Notice that two different threads are acquiring the lock at different times. The time it takes from when the UI thread releases the lock to invokeOnCompletion being called is potential for a race condition to occur.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -434,5 +585,40 @@ class ExplorationProgressController @Inject constructor(
...
+  }
+
+  private fun markExplorationAsInProgressNotSaved(
+    profileId: ProfileId,
+    topicId: String,
+    storyId: String,
+    explorationId: String,
+    lastPlayedTimestamp: Long
+  ) {
+    storyProgressController.recordChapterAsInProgressNotSaved(
```
BenHenning, line 636: This seems like a dependency inversion, but I'm not sure what the alternative is at the moment.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,114 @@ class ExplorationProgressController @Inject constructor(
...
+   *
+   * @param profileId is the profile id currently playing the exploration.
+   * @param topicId is the id of the topic which contains the story with the current exploration.
+   * @param storyId is the id of the story which contains the current exploration.
+   * @param lastPlayedTimestamp timestamp of the time when the checkpoints state for the exploration
+   *     was last updated.
+   * @param newCheckpointState the latest state obtained after saving checkpoint successfully or
+   *     unsuccessfully.
+   */
+  private fun processSaveCheckpointResult(
```
BenHenning, line 466: Shouldn't this method also notify that the state DataProvider has changed since there's now new data to report to the UI? That part seems important to ensure the UI gets the most up-to-date snapshot of the current state of the controller has possible (to ensure the 'eventual consistency' principle I mentioned in my other comment).

BenHenning (changes requested):
Thanks @MaskedCarrot. Left more thoughts. Please assign back once you have a chance to reply to all of my open comments that aren't already communicated as completed.

domain/src/test/java/org/oppia/android/domain/topic/StoryProgressControllerTest.kt
MaskedCarrot: added more tests

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: The expression `this.stateIndex = stateIndex`  is being interpreted as `ExplorationCheckpointBuilder.stateIndex =  ExplorationCheckpointBuilder.stateIndex`

So reverting back to original.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: As per your earlier comment, we should modify the ephemeral state to store the save state and then use the currentStateDataProvider to pull the save state in the state fragment when the user exits the exploration. But the exploration exits from the ExplorationActivtity and the currentStateDataProvider is in StateFragmentPresenter. So in order to get the save state in ExplorationActivity, we will have to get the currentStateDataProvider in ExplorationActivity, observe it and then pull the save state from there, now since we are reacquiring lock to compute ephemeral state I think we should just check the save state directly without modifying the ephemeral state.

MaskedCarrot:
@BenHenning I have made the requested changes PTAL.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: Understood. Thanks for the explanation.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
vinitamurthi: I think we should then verify that the optimization works as intended right?

MaskedCarrot: added a test that verifies that the optimization works as intended.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -218,7 +218,11 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
 
   private fun playExploration(topicId: String, storyId: String, explorationId: String) {
     explorationDataController.startPlayingExploration(
-      explorationId
+      internalProfileId,
+      topicId,
+      storyId,
+      explorationId,
+      shouldSavePartialProgress = false
```
BenHenning, line 227: Here & elsewhere: should this have a TODO for conditionally setting it true/false based on the completion status?

domain/build.gradle
BenHenning: Reminder to address this comment fully @prayutsu.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
```diff
@@ -41,12 +41,31 @@ class ExplorationDataController @Inject constructor(
...
-   *     fail to load, but this provides early-failure detection.
+   * @param internalProfileId the ID corresponding to the profile for which exploration has to be
+   *     played
+   * @param topicId the ID corresponding to the topic for which exploration has to be played
+   * @param storyId the ID corresponding to the story for which exploration has to be played
+   * @param explorationId the ID of the exploration which has to be played
+   * @param shouldSavePartialProgress the boolean that indicates if partial progress has to be saved
+   *     for the current exploration
+   * @return a one-time [LiveData] to observe whether initiating the play request succeeded.
+   *     The exploration may still ail to load, but this provides early-failure detection.
```
BenHenning: Typo in word 'fail'

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -70,6 +79,17 @@ internal class ExplorationProgress {
     }
   }
 
+  /**
+   * Updates the checkpointState to a new state depending upon the result of save operation for
+   * checkpoints.
+   *
+   * @param newCheckpointState is the latest checkpoint state that is returned upon
+   *        completion of the save operation for checkpoints either successfully or unsuccessfully.
```
BenHenning:
```suggestion
   * @param newCheckpointState is the latest checkpoint state that is returned upon
   *      completion of the save operation for checkpoints either successfully or unsuccessfully.
```

Continuation is +4 from the normal KDoc spacing.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -47,23 +59,37 @@ class ExplorationProgressController @Inject constructor(
   //  to avoid cases in tests where the exploration load operation needs to be fully finished before
   //  performing a post-load operation. The current state of the controller is leaking this
   //  implementation detail to tests.
+  // TODO(#3467): Update the mechanism to save checkpoints to eliminate the race condition.
```
BenHenning: Please provide context on what race condition is being referred to here.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -343,6 +380,12 @@ class ExplorationProgressController @Inject constructor(
           "Cannot navigate to a next state if an answer submission is pending."
         }
         explorationProgress.stateDeck.navigateToNextState()
+
+        // Only mark checkpoint if current state is pending state. This ensures that checkpoint
```
```suggestion
        // Only mark checkpoint if current state is pending state. This ensures that checkpoints
```

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -352,6 +395,113 @@ class ExplorationProgressController @Inject constructor(
     }
   }
 
+  /**
+   *  Checks if checkpointing is enabled, if checkpointing is enabled this function creates a
```
```suggestion
   * Checks if checkpointing is enabled, if checkpointing is enabled this function creates a
```

Spacing is wrong for the KDoc (should just be 1 space from the '*'). Please fix here & anywhere else in the PR.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: I'm confused--wasn't the entire point of moving the saving to this controller to avoid the complexity of observing the save state in the UI layer? Couldn't we just save a value in StateFragment and call down to it from ExplorationActivity to see if it's set? That seems a lot simpler to me unless I'm missing something.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: My concern is that this special case is a bit too specific. I think what might be helpful is if you introduce a helper method to check whether the interaction auto-continues and, if it doesn't, then you save the exploration checkpoint. That helper could be as simple as:

```kotlin
private fun doesInteractionAutoContinue(interactionId: String): Boolean =
  interactionId == "Continue"
```

This provides context as to why we're skipping it, and isolates the non-generic bit to a separate helper that could, in the future, be refactored if we add more auto-continuation interactions.

BenHenning (dismissed):
Thanks @MaskedCarrot. I haven't been focusing as much on the specifics of the controllers or even code quality (though I've commented on things when I notice them), and instead have been focused on the actual data flow mechanisms. My remaining comments cover all of my remaining concerns; please address & respond to all of them (some older ones haven't yet been responded to by you--please make sure to respond to these also). Once all are addressed, I'm happy to approve this.

Also, please follow up in chat regarding the bit about not needing a DataProvider in ExplorationActivity if we should discuss that further.

vinitamurthi (approved):
LGTM thanks!

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
MaskedCarrot: Do you mean these words: `early-failure` ?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
MaskedCarrot: Do you mean these `early-failure` words?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
MaskedCarrot: Do you mean these `early-failure` words?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: changed the approach to the dataProvider approach.

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
MaskedCarrot: removed `visible for testing`. Now, this function is part of the class's API and is being used in testing.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: added the helper function.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -589,6 +603,10 @@ class ExplorationProgressController @Inject constructor(
     saveExplorationCheckpoint()
   }
 
+  /** Checks if the answer was submitted to the state on behalf of continue interaction. */
```
```suggestion
  /** Returns whether the specified interaction automatically continues the user to the next state upon completion. */
```

Needs line wrapping.

model/src/main/proto/exploration.proto
```diff
@@ -331,3 +334,21 @@ message HelpIndex {
     bool everything_revealed = 3;
   }
 }
+
+// Different states in which exploration checkpoint can exist.
+enum CheckpointState {
+
```
BenHenning: Nit: remove extra newline at the beginning of the message for consistency with other enums/messages.

model/src/main/proto/exploration.proto
```diff
@@ -233,6 +233,9 @@ message EphemeralState {
     // message structure if additional data needs to be passed along to the terminal state card.
     bool terminal_state = 6;
   }
+
+  // The current state of checkpoint of an exploration.
```
```suggestion
  // The current status of checkpointing the exploration.
```

Might be a bit clearer.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
BenHenning: ``may still fail to load``

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
BenHenning: Latest changes now include this per discussion in the team meeting.

domain/build.gradle
BenHenning: @prayutsu reminder.

BenHenning (commented):
Thanks @MaskedCarrot. I have 4 nits and 1 reminder comment--please make sure to address all of these. The PR otherwise LGTM.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

model/src/main/proto/exploration.proto
MaskedCarrot: done

model/src/main/proto/exploration.proto
MaskedCarrot: done

MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt
MaskedCarrot: done

domain/build.gradle
aggarwalpulkit596: @MaskedCarrot Kindly follow up with @prayutsu and make sure it get resolved.

model/src/main/proto/exploration.proto
aggarwalpulkit596: Marking as resolved as already done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
aggarwalpulkit596: Marking as resolved as already done

domain/build.gradle
BenHenning: Sorry, I'm not sure why I kept at-mentioning Abhay. @MaskedCarrot I meant to at-mention you. This needs to be resolved.

Issue #2961, opened 2021-03-21
Set up alpha-specific configurations


BenHenning:
There are various alpha-specific configurations that need to be checked directly into the release branch:
1. Enabling local asset laoding
2. Enabling loading from protos
3. Disabling the questions tab
4. Setting up the alpha lesson-specific recommendation graph
6. Setting up app auto expiration

BenHenning:
Forgot to close this, but it was addressed.

PR #3416, opened 2021-07-06
Fix part of #2691: Mitigate affected tests script CI hang problem

BenHenning:
Fix part of #2691.

## Explanation
Introduces a mitigation for the continued issue of certain PRs causing the ComputeAffectedTests target to hang. There's a bunch of context of different situations where this was observed in #2961 and connected PRs, but principally the issue seems to be that sibling queries that return large numbers of results can cause Bazel to hang in CI environments. My theory is that this has something to do with both fewer environment resources & a difference in the filesystem used in CI.

The mitigation tries to reduce the chance the error happens in two ways:
1. By splitting up the query such that we first compute the list of rbuildfile deps, then for *each* dependency we calculate siblings before combining the list together for the final rdeps & test computation (moving from 1 query to 2+N for N related build files).
2. By introducing filtering on the siblings query to only apply to tests and Android libraries.

Note that (1) probably means the query is overall slower in non-CI environments, but it lets us do (2) which is probably the actual fix. For some files, the siblings call alone would expand to >1400 results. With filtering, this goes down to <100 for the same target. There's one important caveat to note here: in special cases where a non-test, non-library target is affected we will not find corresponding tests. This seems fairly unlikely given that this entire mechanism is only the secondary means of finding affected tests, anyway.

Regarding testing, one new test was introduced to verify a scenario where bzl file changes affect tests. This test passed without the fix in CI, but it's still a nice test to have. Further, the WORKSPACE test that previously consistently hung is now passing in CI with this mitigation which provides a fairly strong indication that this fix should work for us at least in the medium term. Finally, #3417 demonstrated the fix works in a real situation where an originating PR's (#3340) tests would not be computed properly in CI.

## Checklist

### DIFF
--- a/scripts/src/java/org/oppia/android/scripts/ci/ComputeAffectedTests.kt
+++ b/scripts/src/java/org/oppia/android/scripts/ci/ComputeAffectedTests.kt
@@ -21,7 +21,7 @@ import kotlin.system.exitProcess
  *     Generally, this is 'origin/develop'.
  *
  * Example:
- *   bazel run //scripts:compute_affected_tests -- $(pwd) /tmp/affected_tests.log
+ *   bazel run //scripts:compute_affected_tests -- $(pwd) /tmp/affected_tests.log origin/develop
  */
 fun main(args: Array<String>) {
   if (args.size < 3) {
--- a/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
+++ b/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
@@ -55,6 +55,38 @@ class BazelClient(
     val buildFileList = buildFiles.joinToString(",")
     // Note that this check is needed since rbuildfiles() doesn't like taking an empty list.
     return if (buildFileList.isNotEmpty()) {
+      val referenceFiles =
+        executeBazelCommand(
+          "query",
+          "--noshow_progress",
+          "--universe_scope=//...",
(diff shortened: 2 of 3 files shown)


BenHenning:
NB: I'm keeping this in draft state to investigate the underlying problem. We discovered that while the new Kotlin script introduced in #3374 mitigated the formatting issue that prevented CI from running in #3340, it didn't stop a new issue of the script hanging. The current working theory is that the rbuildfiles routine has issues in CI when trying to find the dependencies for the WORKSPACE file, however #3340 doesn't change WORKSPACE. I'm now thinking that any changes to bzl files might have the same effect, so I've introduced a new test where we change a bzl file to determine if it also causes CI to hang.

BenHenning:
Hmm. Unfortunately, the tests passed. Still thinking about this.

BenHenning:
The WORKSPACE test is now passing with the mitigation. :) That introduces a lot more confidence that this fix could work longer term.

BenHenning:
@anandwana001 & @Sparsh1212 PTAL.

Akshay: note that this is blocking Sparsh's and Abhay's downstream PRs since their PR's tests can't be verified in CI due to the compute affected tests CI issue.

Sparsh1212:
LGTM!

Sparsh1212 (approved):

anandwana001 (approved):
As all ci test as passing, LGTM

PR #3438, opened 2021-07-08
Temp PR to test new Oppiabot CI workflow

BenHenning:
Temp PR to test Oppiabot CI workflow from #3166.

PR #3439, opened 2021-07-08
Create tempfile

BenHenningGithubTest:
See #3438 for context. Testing the same scenario except without having signed the CLA to verify that it fails correctly.


BenHenning:
Aha. Excellent: the failure seems to be working as expected per the comment above & https://github.com/oppia/oppia-android/pull/3439/checks?check_run_id=3014680921.

PR #3434, opened 2021-07-07
Fix #3285: Create Kotlin script to compile list of maven dependencies and license links

prayutsu:
## Explanation

Fixes #3285:
Created a Kotlin script that compiles a list of Maven dependencies (direct + transitive) along with their licenses links.

To run the script, run the following command - 
```
bazel run //scripts:generate_maven_dependencies_list -- $(pwd) third_party/maven_install.json 
```

Expected Output - 
As maven_dependencies.textproto has been updated manually by me, the script should complete its execution and print - 
```
Script executed succesfully: maven_dependencies.textproto updated successfully."
```

Screenhots of passing `GenerateMavenDependenciesListTest` - 
Screenshot from 2021-07-16 06-39-38

Screenhots of passing `LicenseFetcherImplTest` - 
Screenshot from 2021-07-16 06-41-40

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3450, opened 2021-07-09
Fix #3287: Introduce Kotlin script to extract license texts

### LINKED ISSUE #3287: Create a Kotlin script to extract the licenses for the Maven dependencies
Create a Kotlin script that will extract the license texts for all the Maven dependencies and also generate resource XML files that will be used to display the actual license texts.

**Target PR date**: 2021-06-30
**Target completion date**: 2021-07-06

prayutsu:
## Explanation

Fixes #3287:
Created a Kotlin script that extracts the license texts for the third-party maven dependencies.

To run the script, run the following command - 
```
bazel run //scripts:generate_license_texts -- $(pwd)/app/src/main/res/values scripts/assets/maven_dependencies.pb
```

As an output of the script, 5 files would get generated (resource XML files) that will be used to finally show these licenses in the UI. Note that these files should never be directly checked into the codebase.
They will only be generated by the maintainers at the time of preparing for a release.

Screenshot of passing test cases- 
Screenshot from 2021-07-21 16-12-36

Files generated by the script can be seen here - 
https://gist.github.com/prayutsu/7693dabae98370ca000d959efaff0058

## Checklist


prayutsu:
PTAL @BenHenning @rt4914 @vinitamurthi @anandwana001

prayutsu:
Please note that this and #3434 require some parts to be shifted into new utility files for testing. Since both the scripts are fully functional now, I am working on shifting some parts to separate utilities so that the script can be tested.
The tests to both the PRs will be added simultaneously very soon.
Till then please add an initial review.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * Script to extract the licenses for the third-party Maven dependencies (direct and indirect both)
+ * on which Oppia Android depends.
+ *
+ * Usage:
+ *   bazel run //scripts:generate_license_texts -- <path_to_directory_values>
+ *   <name_of_value_resource_file_1> <name_of_value_resource_file_2>
+ *   <name_of_value_resource_file_2> <name_of_value_resource_file_4>
+ *   <name_of_value_resource_file_2> <path_to_pb_file>
```
rt4914: I think all of these can be directly defined in the script.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+          .setLicenseText(licenseText)
+          .setLicenseLink(licenseLink)
+          .build()
+      )
+    }
+  }
+  return copyrightLicensesSet
+}
+
+fun scrapeLicenseText(url: String): String {
```
rt4914: Make this private.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  val text = URL(url).openStream().bufferedReader().readText()
+  return addEscapeCharactersToLicenseText(text)
+}
+
+fun addEscapeCharactersToLicenseText(licenseText: String): String {
```
rt4914: Ditto

rt4914 (changes requested):
Took first pass. Will need a detailed review once #3434 is updated as per testing strategy. Mostly this looks good.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  val licenseTextBuilder = StringBuilder()
+  for (c in licenseText) {
+    when (c) {
+      '<', '>', '\'', '\"', '&' -> licenseTextBuilder.append(escapeCharactersMap[c])
```
BenHenning: Could we use Guava for this, instead? https://stackoverflow.com/a/26572556

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+}
+
+private fun retrieveArtifactsVersionsList(
+  mavenDependencyList: List<MavenDependency>
+): List<String> {
+  val artifactVersionsList = mutableListOf<String>()
+  mavenDependencyList.forEach { dependency ->
+    artifactVersionsList.add(dependency.artifactVersion)
+  }
+  return artifactVersionsList.toList()
```
BenHenning: More Kotlin idiomatic:

```suggestion
  return mavenDependencyList.map { it.artifactVersion }
```

Ditto elsewhere in this file.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  proto: MavenDependencyList
+): MavenDependencyList {
+  val protoBinaryFile = File(pathToPbFile)
+  val builder = proto.newBuilderForType()
+  return FileInputStream(protoBinaryFile).use {
+    builder.mergeFrom(it)
+  }.build() as MavenDependencyList
+}
+
+fun writeDependenciesNamesXml(
```
BenHenning: This should be private (only public API methods should be public which, for this script, is just ``main``).

Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  val mavenDependencyList = retrieveMavenDependencyList(pathToPbFile)
+  if (mavenDependencyList.isEmpty()) {
+    throw Exception(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE)
+  }
+  val copyrightLicenseSet = retrieveAllLicensesSet(mavenDependencyList)
+
+  writeDependenciesNamesXml(pathToNamesXml, retrieveArtifactsNamesList(mavenDependencyList))
+
+  writeDependenciesVersionsXml(
+    pathToVersionsXml,
```
BenHenning: Why have different XML files for the string resources vs. bundling them in one? The latter seems easier to manage.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  licenseLinkToIndexNameMap: HashMap<String, String>,
+  mavenDependenciesList: List<MavenDependency>
+) {
+  val file = File(pathToResourceXml)
+  val docBuilder: DocumentBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder()
+  val doc: Document = docBuilder.newDocument()
+  val rootResourcesElement: Element = doc.createElement("resources")
+
+  // Add all string-arrays of license texts for each dependency.
+  for (index in mavenDependenciesList.indices) {
```
BenHenning: Here & elsewhere: it may be simpler to use forEachIndexed:

```kotlin
mavenDependenciesList.forEachIndexed { index, dependency ->
  ...
}
```

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+    val stringArrayElement = doc.createElement("string-array")
+    stringArrayElement.setAttribute("name", "third_party_dependency_license_texts_$index")
+    dependency.licenseList.forEach { license ->
+      val licenseLink = if (license.primaryLinkType == PrimaryLinkType.SCRAPE_DIRECTLY ||
+        license.primaryLinkType == PrimaryLinkType.SHOW_LINK_ONLY
+      ) {
+        license.primaryLink
+      } else {
+        license.alternativeLink
+      }
```
BenHenning: This still isn't making sense to me--why would we keep a link that can't be shown? What's the case where we want to show the primary but not the secondary, or vice versa?

If the purpose is to retain the original link and allow for a local copy, then maybe using a ``oneof`` would be a better approach:

```proto
oneof license_link_type {
  ScrapableLink scrapable_link = 1;
  LocalCopy local_copy = 2;
  LinkOnly link_only = 3;
}

message ScrapableLink {
  string url = 1;
}

message LocalCopy {
  string filename = 1;
}

message DirectLink {
  string url = 1;
}
```

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+  rootResourcesElement.appendChild(arrayElement)
+  doc.appendChild(rootResourcesElement)
+  doc.xmlStandalone = true
+  getTransformer().transform(DOMSource(doc), StreamResult(file))
+}
+
+fun retrieveAllLicensesSet(
+  mavenDependencyList: List<MavenDependency>
+): Set<CopyrightLicense> {
+  val copyrightLicensesSet = mutableSetOf<CopyrightLicense>()
```
BenHenning: This is another case we can avoid needing to use a mutable set.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,423 @@
...
+        }
+        else -> throw Exception(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE)
+      }
+      copyrightLicensesSet.add(
+        CopyrightLicense
+          .newBuilder()
+          .setLicenseName(license.licenseName)
+          .setLicenseText(licenseText)
+          .setLicenseLink(licenseLink)
+          .build()
```
BenHenning: Optional, but a slightly nicer way to construct a proto is:

```suggestion
        CopyrightLicense.newBuilder().apply {
          licenseName = license.licenseName
          this.licenseText = licenseText
          this.licenseLink = licenseLink
        }.build()
```

scripts/src/java/org/oppia/android/scripts/proto/copyright_license.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package proto;
+
+option java_package = "org.oppia.android.scripts.proto";
+option java_multiple_files = true;
+
+// Stores the details of a copyright license.
+message CopyrightLicense {
```
BenHenning: This doesn't ever seem to be stored on disk. Why not use a data class instead of a proto?

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+import org.oppia.android.scripts.proto.MavenDependencyList
+import org.oppia.android.scripts.proto.PrimaryLinkType
+import org.oppia.android.testing.assertThrows
+import java.io.ByteArrayOutputStream
+import java.io.File
+import java.io.PrintStream
+
+/** Tests for [GenerateLicenseTexts]. */
+class GenerateLicenseTextsTest {
+  private val TOO_LESS_ARGS_FAILURE = "Too less arguments passed"
```
BenHenning: Nit: For the actual error & related variable news, it should be "too few" not "too less".

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+
+    val pbFile = tempFolder.newFile("assets/test_maven_dependencies.pb")
+    mavenDependencyList.writeTo(pbFile.outputStream())
+
+    val exception = assertThrows(Exception::class) { runScript() }
+    assertThat(exception).hasMessageThat().contains(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE_FAILURE)
+  }
+
+  @Test
+  fun testScript_licenseListEmpty_failsWithNotUpToDateException() {
```
BenHenning: Is this a sufficient check for the list not being up-to-date? It could have entries but not have been recently updated, and it appears the check in the script would pass.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+        primaryLink = VALID_LINK
+      )
+    )
+    val dependencyList = listOf<MavenDependency>(
+      getMavenDependency(
+        artifactName = "artifact:name",
+        licenseList = licenseList
+      )
+    )
+    val mavenDependencyList = getMavenDependencyList(dependencyList)
```
BenHenning: Here & elsewhere: please only use ``main()`` when testing the script. We should never punch holes in the private implementation of a component when testing it, and instead should only test it through its public API.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+    runScript()
+
+    assertXmlGenerated("test_third_party_dependency_names.xml")
+    assertXmlGenerated("test_third_party_dependenct_versions.xml")
+    assertXmlGenerated("test_third_party_dependency_license_texts.xml")
+    assertXmlGenerated("test_third_party_dependency_license_texts_array.xml")
+    assertXmlGenerated("test_third_party_dependency_license_names_array.xml")
+  }
+
+  private fun assertXmlGenerated(filename: String) {
```
BenHenning: This doesn't seem sufficient. We should probably parse the XML to verify it's valid, and perform basic checks to make sure the correct attribute names & contents are being added. At the moment, the tests don't seem to be verifying the correctness of the generated XML.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+import java.io.PrintStream
+
+/** Tests for [GenerateLicenseTexts]. */
+class GenerateLicenseTextsTest {
+  private val TOO_LESS_ARGS_FAILURE = "Too less arguments passed"
+  private val MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE_FAILURE =
+    "maven_dependencies.textproto is not up-to-date"
+  private val SCRIPT_PASSED_INDICATOR = "Script execution completed."
+  private val VALID_LINK = "https://www.apache.org/licenses/LICENSE-2.0.txt"
+  private val INVALID_LINK = "https://fabric.io/terms"
```
BenHenning: Please move the HTTP download portion of the script to a utility that's abstracted by an interface (maybe LicenseFetcher?) so that we can mock it out in the test. We should never be relying on network connectivity during a test since that both introduces flakiness and can prevent people from developing the codebase while offline.

BenHenning (commented):
Thanks @prayutsu. Took a first pass on the PR.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+  private fun getMavenDependencyList(
+    dependenciesList: List<MavenDependency> = listOf<MavenDependency>()
+  ): MavenDependencyList {
+    return MavenDependencyList
+      .newBuilder()
+      .addAllMavenDependency(dependenciesList)
+      .build()
+  }
+
+  private fun getLicense(
```
anandwana001: add kdoc

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+  ): MavenDependency {
+    return MavenDependency
+      .newBuilder()
+      .setArtifactName(artifactName)
+      .setArtifactVersion("1.0")
+      .addAllLicense(licenseList)
+      .build()
+  }
+
+  private fun getMavenDependencyList(
```
anandwana001: add kdoc

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+    val xmlHeader =
+      """
+      <?xml version="1.0" encoding="UTF-8"?>
+      """.trimIndent()
+
+    assertThat(text).contains(xmlHeader)
+  }
+
+  private fun getMavenDependency(
```
anandwana001: add kdoc

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,259 @@
...
+  private fun assertXmlGenerated(filename: String) {
+    val text = File("${tempFolder.root}/values/$filename").bufferedReader().use {
+      it.readText()
+    }
+    val xmlHeader =
+      """
+      <?xml version="1.0" encoding="UTF-8"?>
+      """.trimIndent()
+
+    assertThat(text).contains(xmlHeader)
```
anandwana001: is this a helper function or a test?

If this a helper function, try to keep the test at a separate function. 

Also, add kdoc

anandwana001 (changes requested):

scripts/src/java/org/oppia/android/scripts/proto/copyright_license.proto
vinitamurthi: +1 dont store it as a proto if it isn't necessary

vinitamurthi (approved):
LGTM as codeowner!

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Changed script to write only a single resource XML file.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Sorry, I got a bit confused here, can you give an example here?

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: No more needed as we are not using this proto class, but made changes for other proto classes.

scripts/src/java/org/oppia/android/scripts/proto/copyright_license.proto
prayutsu: Done, removed.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Actually, this script can not verify whether maven_dependencies.pb is up-to-date or not, we rely on the first script and the CI check for this.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

prayutsu:
@BenHenning @rt4914 PTAL.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: the function is not needed now.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -558,6 +558,8 @@ exempted_file_path: "domain/src/main/java/org/oppia/android/domain/util/JsonAsse
 exempted_file_path: "domain/src/test/java/org/oppia/android/domain/classify/InteractionObjectTestBuilder.kt"
 exempted_file_path: "scripts/src/java/org/oppia/android/scripts/common/CommandExecutor.kt"
 exempted_file_path: "scripts/src/java/org/oppia/android/scripts/common/CommandResult.kt"
+exempted_file_path: "scripts/src/java/org/oppia/android/scripts/maven/data/CopyrightLicense.kt"
+exempted_file_path: "scripts/src/java/org/oppia/android/scripts/maven/data/Dependency.kt"
 exempted_file_path: "scripts/src/java/org/oppia/android/scripts/maven/data/MavenListDependency.kt"
```
prayutsu: These are data classes and hence are being exempted.

BenHenning:
Apologies @prayutsu I'll probably need to take a look at this tomorrow instead.

If possible, it might be best if @rt4914 could take a first pass, too, before I take a look at this in case there are any larger changes that need to be addressed.

Beyond that, would it be useful for @vinitamurthi to take another pass? I'm not sure how much this PR has changed given the base PR changed a lot before it was merged late last week.

prayutsu:
> Apologies @prayutsu I'll probably need to take a look at this tomorrow instead.
> 
> If possible, it might be best if @rt4914 could take a first pass, too, before I take a look at this in case there are any larger changes that need to be addressed.

Ack, no problem.
> 
> Beyond that, would it be useful for @vinitamurthi to take another pass? I'm not sure how much this PR has changed given the base PR changed a lot before it was merged late last week.

I'll re-request the review. As the proto has been removed, she doesn't appear to be a codeowner for any file now. Should I instead assign her the PR?

prayutsu:
@vinitamurthi Re-requesting review as the PR has changed significantly.

BenHenning:
Ah, @vinitamurthi's review probably isn't needed then (but still welcome if you have time Vinita!).

Also, please assign back once @rt4914 is happy with this. I have a bunch of PRs coming in at the moment & need to load-balance.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,547 @@
...
+  }
+
+  private fun verifyList(itemList: List<String>, expectedSize: Int, expectedList: List<String>) {
+    assertThat(itemList.size).isEqualTo(expectedSize)
+    itemList.forEachIndexed { index, item ->
+      assertThat(item).isEqualTo(expectedList[index])
+    }
+  }
+
+  /** Returns a [MavenDependency]. */
```
anandwana001: MavenDependency is no argument so we can remove the brackets. 
As scripts is not a module so putting the brackets will not help us to directly jump over the class mentioned in the brackets. IDE doesn't help in such cases. 

What best I see here is `Returns an instance of MavenDependency proto message`

Applies to other kdoc as well.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,547 @@
...
+    licenseList: List<License> = listOf<License>()
+  ): MavenDependency {
+    return MavenDependency.newBuilder().apply {
+      this.artifactName = artifactName
+      this.artifactVersion = artifactVersion
+      this.addAllLicense(licenseList)
+    }.build()
+  }
+
+  /** Returns a [MavenDependencyList]. */
```
anandwana001: MavenDependencyList is no argument so we can remove the brackets.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,547 @@
...
+    directLinkOnlyUrl: String
+  ): License {
+    return License.newBuilder().apply {
+      this.licenseName = licenseName
+      this.originalLink = originalLink
+      this.directLinkOnly = DirectLinkOnly.newBuilder().setUrl(directLinkOnlyUrl).build()
+    }.build()
+  }
+
+  private fun retreiveListOfStrings(
```
anandwana001: add a complete kdoc here with an explanation on what these arguments are and when to pass what.

anandwana001 (changes requested):
Thanks @prayutsu 

Left few nit thoughts on kdocs part. Else all LGTM as all the test cases are passing.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,401 @@
...
+      dependencyVersionsList,
+      copyrightLicenseSet,
+      dependencyList
+    )
+
+    println("\nScript execution completed successfully.")
+  }
+
+  /** Retrieves the list of [MavenDependency] from maven_dependencies.textproto. */
+  private fun retrieveMavenDependencyList(pathToPbFile: String): List<MavenDependency> {
```
rt4914: Please mentioned `@param` and `@return` similar to below function

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
rt4914: Typo in function name `retrieveListOfStrings`

rt4914 (changes requested):
Minor changes. Also please update the PR description with correct command.

Also we discussed about the update the format of xml to better work with UI code, is that not necessary now? (We discussed about using `index` based integer-array for licences)

prayutsu:
@rt4914 @BenHenning We shall block this PR on #3501, so that we can ensure that the generated XML file is correct.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

rt4914:
> @rt4914 @BenHenning We shall block this PR on #3501, so that we can ensure that the generated XML file is correct.

@prayutsu Sounds good to me.

prayutsu:
One test case is failing in CI, but all are passing locally for me (even after I ran 5-6 times) - 
Screenshot from 2021-07-29 03-01-33

prayutsu:
@anandwana001 @BenHenning @rt4914 PTAL, this is no more blocked and is ready for a full review.

BenHenning:
Thanks @prayutsu. Will take a pass on this tomorrow my time.

Could you also update to the latest develop? There seem to be pending conflicts in scripts/BUILD.bazel.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,576 @@
...
+      expectedList = listOf(
+        "Apache License",
+        "Android Terms of Service",
+        "BSD License"
+      )
+    )
+
+    val dependencyNamesArray = retrieveArray(
+      xmlContent,
+      "third_party_dependency_names_array",
```
anandwana001: add named argument for this and other similar cases

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
```diff
@@ -0,0 +1,576 @@
...
+    )
+    verifyArray(
+      itemList = arrayOfLicenseTextsArrays,
+      indicesList = List(dependencyList.size) { it },
+      expectedSize = dependencyList.size,
+      prefix = "@array/third_party_dependency_license_texts_"
+    )
+  }
+
+  private fun verifyArray(
```
anandwana001: add kdoc explaining what each argument is and what it will return. Follow the format of using annotation `@param ...  @returns ..`

anandwana001 (approved):
Thanks @prayutsu Test file LGTM,
Left two nit comments which you can resolve on my behalf once done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/GenerateLicenseTextsTest.kt
prayutsu: Done.

rt4914 (approved):
LGTM, thanks.

BenHenning:
Hi. I'll need to review this a bit later--sorry about that.

scripts/BUILD.bazel
```diff
@@ -118,6 +118,14 @@ kt_jvm_binary(
     runtime_deps = ["//scripts/src/java/org/oppia/android/scripts/maven:generate_maven_dependencies_list_lib"],
 )
 
+kt_jvm_binary(
+    name = "generate_license_texts",
```
BenHenning: Maybe instead: ``retrieve_license_texts`` or similar? I think my earlier comment about 'generation' applies here, too. Ditto for all the library/class names elsewhere.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+    val pathToThirdPartyDependenciesXml = "$pathToValuesDirectory/third_party_dependencies.xml"
+    val pathToMavenDependenciesPb = args[1]
+
+    val mavenDependencyList = retrieveMavenDependencyList(pathToMavenDependenciesPb)
+    if (mavenDependencyList.isEmpty()) {
+      throw Exception(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE)
+    }
+    val copyrightLicenseSet = retrieveAllLicensesSet(mavenDependencyList)
+    val dependencyList =
+      retrieveDependencyList(mavenDependencyList)
```
BenHenning: Can this fit on one line?

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+
+    println("\nScript execution completed successfully.")
+  }
+
+  /**
+   * Retrieve the list of maven dependencies from maven_dependencies.textproto.
+   *
+   * @param pathToPbFile path to the pb file to be parsed
+   *
+   * @return list of [MavenDependency]
```
BenHenning:
```suggestion
   * @return list of [MavenDependency]s
```

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+    )
+
+    println("\nScript execution completed successfully.")
+  }
+
+  /**
+   * Retrieve the list of maven dependencies from maven_dependencies.textproto.
+   *
+   * @param pathToPbFile path to the pb file to be parsed
+   *
```
BenHenning: Nit: remove extra newline between at-clauses. Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+   */
+  private fun retrieveMavenDependencyList(pathToPbFile: String): List<MavenDependency> {
+    return parseTextProto(
+      pathToPbFile,
+      MavenDependencyList.getDefaultInstance()
+    ).mavenDependencyList
+  }
+
+  /**
+   * Helper function to parse the textproto file to a proto class.
```
BenHenning: This & the helper name don't match the implementation--it's parsing a binary proto, not a textproto.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+            licenseLink,
+            licenseText,
+          )
+        )
+      }
+    }
+    return copyrightLicensesSet
+  }
+
+  private fun retrieveDependencyList(
```
BenHenning: This method seems really duplicative to the one above--can we can consolidate some of this common logic into helpers or somehow combine the methods?

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+      println(
+        """
+        Usage: bazel run //scripts:generate_license_texts -- <path_to_directory_values>
+        <path_to_pb_file>  
+        """.trimIndent()
+      )
+      throw Exception("Too few arguments passed.")
+    }
+    val pathToValuesDirectory = args[0]
+    val pathToThirdPartyDependenciesXml = "$pathToValuesDirectory/third_party_dependencies.xml"
```
BenHenning: Suggest using File to compute a more robust path (to avoid depending on the argument to be passed a certain way:

```suggestion
    val valuesDirectory = File(pathToValuesDirectory) // Maybe verify this exists & is a directory with a check?
    val thirdPartyDependenciesXml = File(valuesDirectory, "third_party_dependencies.xml")
```

Ditto for other files (if applicable) to simplify & improve robustness.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
```diff
@@ -0,0 +1,435 @@
...
+    )
+    rootResourcesElement.appendChild(stringArrayElement)
+  }
+
+  private fun writeDependenciesVersionsXml(
+    doc: Document,
+    dependencyVersionsList: List<String>,
+    rootResourcesElement: Element
+  ) {
+
```
BenHenning: Nit: remove extra newline (for consistency). Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/maven/data/BUILD.bazel
```diff
@@ -1,5 +1,6 @@
 """
-Libraries corresponding to parse maven_install.json file.
+Libraries corresponding to parse maven_install.json file and storing license details to be used
+within GenerateLicenseTexts.kt.
```
BenHenning:
```suggestion
Libraries corresponding to parse maven_install.json file and representeing license details.
```

Prefer not to be too specific since implementations change over time (especially class names).

scripts/src/java/org/oppia/android/scripts/maven/data/BUILD.bazel
```diff
@@ -15,3 +16,13 @@ kt_jvm_library(
     visibility = ["//scripts:oppia_script_library_visibility"],
     deps = ["//third_party:moshi"],
 )
+
+kt_jvm_library(
+    name = "copyright_license_lib",
```
BenHenning: Might this be a potentially better fit under a license/model package, instead?

Similarly, this data package might actually be better represented if it were called "model".

BenHenning (changes requested):
Thanks @prayutsu. Took a first pass on everything minus the tests. No major concerns with the approach, I mostly had nits so far. PTAL.

scripts/BUILD.bazel
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Fixed.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Made a new helper method for the common part of both methods.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/data/BUILD.bazel
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/data/BUILD.bazel
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/license/model/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
+"""
+Libraries corresponding to storing license details of the third-party dependencies.
+"""
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_library")
+
+kt_jvm_library(
+    name = "data",
```
BenHenning: Should be "model"

scripts/src/java/org/oppia/android/scripts/maven/RetrieveLicenseTexts.kt
```diff
@@ -0,0 +1,413 @@
...
+      if (licenseList.isEmpty()) {
+        throw Exception(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE)
+      }
+      licenseList.forEach { license ->
+        copyrightLicensesSet.add(
+          retrieveCopyrightLicense(license)
+        )
+      }
+    }
+    return copyrightLicensesSet
```
BenHenning:
```suggestion
    return mavenDependencyList.flatMap { dependency ->
      val licenseList = dependency.licenseList
      check(licenseList.isNotEmpty()) { MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE }
      return@flatMap licenseList.map { license ->
        retrieveCopyrightLicense(license)
      }
    }
```

This should be (effectively) equivalent, and avoids the mutability. Ditto elsewhere in this file that uses mutable lists (prefer to avoid them everywhere if we can).

scripts/src/java/org/oppia/android/scripts/maven/RetrieveLicenseTexts.kt
```diff
@@ -203,6 +158,31 @@ class GenerateLicenseTexts(
...
+        licenseText = license.directLinkOnly.url
+        licenseLink = license.directLinkOnly.url
+      }
+      else -> throw Exception(MAVEN_DEPENDENCY_LIST_NOT_UP_TO_DATE)
+    }
+    return CopyrightLicense(
+      license.licenseName,
+      licenseLink,
+      licenseText
+    )
```
BenHenning:
```suggestion
    return CopyrightLicense(license.licenseName, licenseLink, licenseText)
```

Prefer fitting things on one line if they don't exceed the column limit.

scripts/src/java/org/oppia/android/scripts/maven/model/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+"""
+Libraries corresponding to parse maven_install.json file.
+"""
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_library")
+
+kt_jvm_library(
+    name = "data",
```
BenHenning: This should also be "model".

scripts/src/java/org/oppia/android/scripts/maven/model/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+"""
+Libraries corresponding to parse maven_install.json file.
```
BenHenning:
```suggestion
Data structures corresponding to parsing maven_install.json.
```

Slight clarification & grammar fixes.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+    val exception = assertThrows(Exception::class) {
+      RetrieveLicenseTexts(mockLicenseFetcher).main(arrayOf())
+    }
+
+    assertThat(exception).hasMessageThat().contains(TOO_FEW_ARGS_FAILURE)
+    assertThat(outContent.toString()).contains("Usage:")
+  }
+
+  @Test
+  fun testScript_oneArguments_printsUsageStringAndThrowsException() {
```
BenHenning, line 71:
```suggestion
  fun testScript_oneArgument_printsUsageStringAndThrowsException() {
```

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+    assertThat(exception).hasMessageThat().contains(TOO_FEW_ARGS_FAILURE)
+    assertThat(outContent.toString()).contains("Usage:")
+  }
+
+  @Test
+  fun testScript_dependencyListEmpty_failsWithNotUpToDateException() {
+    val mavenDependencyList = getMavenDependencyList()
+
+    val pbFile = tempFolder.newFile("scripts/assets/maven_dependencies.pb")
+    mavenDependencyList.writeTo(pbFile.outputStream())
```
BenHenning: Does writeTo() close the stream? If not, this should probably be:

```suggestion
    pbFile.outputStream().use { mavenDependencyList.writeTo(it) }
```

To ensure that resources are properly closed. Ditto for other tests.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+        "${tempFolder.root}/scripts/assets/maven_dependencies.pb"
+      )
+    )
+    assertThat(outContent.toString()).contains(SCRIPT_PASSED_INDICATOR)
+    val xmlContent = xmlFile.inputStream().bufferedReader().use { it.readText() }
+    val dependencyNamesList = retrieveListOfStrings(xmlContent, "third_party_dependency_name_")
+    val dependencyVersionsList =
+      retrieveListOfStrings(xmlContent, "third_party_dependency_version_")
+    val licenseTextsList = retrieveListOfStrings(xmlContent, "license_text_")
+    val licenseNamesList = retrieveListOfStrings(xmlContent, "license_name_")
```
BenHenning, line 245: Might it be possible to move some of these bits to helpers to just retrieve the elements that the test needs to verify? If it helps, we can leverage data classes to represent some of the substructures that need to be verified, too.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+
+    verifyList(
+      itemList = dependencyNamesList,
+      expectedSize = 3,
+      expectedList = listOf(
+        "artifact.name",
+        "artifact.name",
+        "artifact.name"
+      )
+    )
```
BenHenning: Suggest instead using Truth:

```suggestion
    assertThat(dependencyNamesList).containsExactly("artifact.name", "artifact.name", "artifact.name")
```

Ditto everywhere else--prefer not using custom verifier methods unless needed.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+      retrieveListOfStrings(xmlContent, "third_party_dependency_version_")
+    val licenseTextsList = retrieveListOfStrings(xmlContent, "license_text_")
+    val licenseNamesList = retrieveListOfStrings(xmlContent, "license_name_")
+
+    verifyList(
+      itemList = dependencyNamesList,
+      expectedSize = 3,
+      expectedList = listOf(
+        "artifact.name",
+        "artifact.name",
```
BenHenning: Might it be possible to set this up to have different artifact names for each? It looks a bit like a bug here (and might hide real issues if we fail to pass data through properly).

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+      expectedList = listOf(
+        "2.0.1",
+        "4.1.1",
+        "3.1.1"
+      )
+    )
+    verifyList(
+      itemList = licenseTextsList,
+      expectedSize = 3,
+      expectedList = listOf(
```
BenHenning: Is this still the correct expected content given that the license XML was replaced with test data?

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+
+  private fun verifyList(itemList: List<String>, expectedSize: Int, expectedList: List<String>) {
+    assertThat(itemList.size).isEqualTo(expectedSize)
+    itemList.forEachIndexed { index, item ->
+      assertThat(item).isEqualTo(expectedList[index])
+    }
+  }
+
+  /** Returns an instance of MavenDependency proto message. */
+  private fun getMavenDependency(
```
BenHenning: For this & other methods, this is actually a create*() method. Prefer that over 'get' since 'get' implies that something is being retrieved, not created.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
```diff
@@ -0,0 +1,585 @@
...
+    }.build()
+  }
+
+  /** Parses the XML to return a list of values enclosed between string tags.
+   *
+   * @param xmlContent XML content to be parsed
+   * @param attributePrefix the prefix of the attribute of <string> tag that needs to be parsed
+   *
+   * @return list of parsed values of <string> nodes with the given attributrPrefix
+   */
```
BenHenning:
```suggestion
  /**
   * Parses the XML to return a list of values enclosed between string tags.
   *
   * @param xmlContent XML content to be parsed
   * @param attributePrefix the prefix of the attribute of <string> tag that needs to be parsed
   * @return list of parsed values of <string> nodes with the given attributrPrefix
   */
```

BenHenning (changes requested):
Thanks @prayutsu. Checked the changes & fully reviewed the tests. PTAL.

scripts/src/java/org/oppia/android/scripts/license/model/BUILD.bazel
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/RetrieveLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/RetrieveLicenseTexts.kt
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/model/BUILD.bazel
prayutsu: Done.

scripts/src/java/org/oppia/android/scripts/maven/model/BUILD.bazel
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Sorry, I didn't get that completely, aren't we already using helpers to retrieve a list of strings that need to be matched?

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Done.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
prayutsu: Actually, we are not testing the output of third_party_dependencies.xml, we are creating a new values XML file in every test case, and we are mocking the `LicenseFetcher` that fetches the text of the URL for us. I have kept it original license text only because test data is completely in our hands and this way it looks closer to the real scenarios.

scripts/src/java/org/oppia/android/scripts/maven/GenerateLicenseTexts.kt
BenHenning: This was addressed in a later comment (the idea is to use methods like ``map{}`` to avoid needing to use mutable structures).

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
BenHenning: I was thinking more of isolating the XML reading & retrieval of higher-level elements to helper methods so that all the test needs to care about is retrieving exactly the data it needs. That being said, I don't feel strongly about this.

scripts/src/javatests/org/oppia/android/scripts/maven/RetrieveLicenseTextsTest.kt
BenHenning: Got it--thanks for clarifying.

BenHenning (approved):
Thanks @prayutsu. LGTM!

PR #3489, opened 2021-07-17
Fix #3485: Added GaeModels along with PlatformParameterService and its Mock

### LINKED ISSUE #3485: Network layer changes for hooking up the backend for PlatformParameters
PlatformParameter API & Gae Models, also a MockPlatformParameterSerivice for fake response.

PlatformParameterService
GaePlatformParameters
GaePlatformParameter
MockPlatformParameterService

Target PR date: 17 July 2021
Target completion date: 24 July 2021

Arjupta:
## Explanation

Fixes #3485 
PlatformParameter API & Gae Models for hooking up the backend and including Mock Platform Parameter Service for representing fake network response.

## Checklist


data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
```diff
@@ -0,0 +1,12 @@
+package org.oppia.android.data.backends.gae.model
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/** Data class for List of [GaePlatformParameter]. */
+@JsonClass(generateAdapter = true)
+data class GaePlatformParameters(
```
jcqli: Maybe `GaePlatformParameterList` so it's more distinct from `GaePlatformParameter`?

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.oppia.android.data.backends.gae.model.GaePlatformParameters
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/** Service that provides access to platform parameter endpoint. */
+interface PlatformParameterService {
+
+// TODO("Change the url to point to the correct endpoint when the backend is ready")
```
jcqli: Is there no backend for this yet? I had thought the backend was done in a previous project.

jcqli (changes requested):
Looks pretty good, left a few comments. Otherwise, be sure to include those data files in the test exemption file if we won't have tests for them.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -70,6 +71,17 @@ class NetworkModule {
     return retrofit.create(FeedbackReportingService::class.java)
   }
 
+  /**
+   * Provides the [PlatformParameterService] implementation.
+   * @param retrofit the Retrofit object used to instantiate the service
```
BenHenning, line 77: Newline between the summary fragment & at-aprams.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
```diff
@@ -70,6 +71,17 @@ class NetworkModule {
...
+   * @return the [PlatformParameterService] implementation.
```
BenHenning:
```suggestion
   * @return the [PlatformParameterService] implementation
```

Nit: at-params don't end in punctuation unless they're multi-lined.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
+package org.oppia.android.data.backends.gae.api
+
+import org.oppia.android.data.backends.gae.model.GaePlatformParameters
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/** Service that provides access to platform parameter endpoint. */
```
BenHenning:
```suggestion
/** Service that provides access to the Oppia platform parameter endpoint. */
```

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
BenHenning: There is: https://github.com/oppia/oppia/blob/develop/core/controllers/platform_feature.py#L28.

Or, is this blocking on updating that API? (in which case, shouldn't that happen first?)

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
...
+interface PlatformParameterService {
+
+// TODO("Change the url to point to the correct endpoint when the backend is ready")
+  @GET("platform_features_evaluation_handler")
+  fun getPlatformParametersByVersion(
+    @Query("app_version") version: String
```
BenHenning: platform_type also should be populated. It's important for the backend to know the request is originating from Android.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameter.kt
```diff
@@ -0,0 +1,13 @@
+package org.oppia.android.data.backends.gae.model
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/** Data class for a Platform Parameter. */
```
BenHenning: Nit: platform & parameter should be lowercase here.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameter.kt
```diff
@@ -0,0 +1,13 @@
...
+@JsonClass(generateAdapter = true)
+data class GaePlatformParameter(
+
+  @Json(name = "parameter_name") val parameterName: String,
+  @Json(name = "parameter_value") val parameterValue: Any
```
BenHenning: Please add KDocs for these public vals. It may be useful to link to Oppia's backend domain objects to help provide context for these.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
```diff
@@ -0,0 +1,12 @@
...
+
+  @Json(name = "platform_parameters") val platformParameters: List<GaePlatformParameter>?
```
BenHenning: Ditto here--should have KDoc.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
```diff
@@ -0,0 +1,12 @@
+package org.oppia.android.data.backends.gae.model
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/** Data class for List of [GaePlatformParameter]. */
```
BenHenning: Perhaps instead: ``Data class for a list of [GaePlatformParameter]s evaluated by Oppia's backend.``

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    assertThat(responseBody).isNotNull()
+    verifyResponseBody(responseBody!!)
+  }
+
+  // Checks for the individual platform parameters in the response body and compares them with the
+  // expected network response body
+  private fun verifyResponseBody(responseBody: GaePlatformParameters) {
+    val gaePlatformParameterList = responseBody.platformParameters!!
+    assertThat(gaePlatformParameterList.size).isEqualTo(expectedNetworkResponse.size)
+    for (platformParameter in expectedNetworkResponse) {
```
BenHenning: In general, we want to avoid control flow in tests since it makes them harder to verify as correct.

Instead, I suggest having separate tests for each case we want to verify: string, integer, and boolean. Further, you can also have a separate initial test to verify that the response for version 1 is successful to avoid needing to repeat that assertion in each test.

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
```diff
@@ -0,0 +1,34 @@
...
+) : PlatformParameterService {
+  override fun getPlatformParametersByVersion(version: String): Call<GaePlatformParameters> {
+    val parameters = createMockGaePlatformParameters()
+    return delegate.returningResponse(parameters).getPlatformParametersByVersion(version)
+  }
+
+  // Creates a Mock Response containing GaePlatformParameter for testing
+  private fun createMockGaePlatformParameters(): GaePlatformParameters {
+    return GaePlatformParameters(
+      arrayListOf(
```
BenHenning: Just ``listOf`` should be sufficient here.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
```diff
@@ -0,0 +1,25 @@
+package org.oppia.android.testing.platformparameter
+
+import javax.inject.Qualifier
+
+// These constants are only used for testing purpose.
```
BenHenning: I suggest instead splitting these into separate files for each annotation, and pairing the name/default/server value with those files. You can then add KDocs for each one to explain what that test parameter is.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
```diff
@@ -0,0 +1,25 @@
...
+@Qualifier
+annotation class TestStringParam
+
+val TEST_STRING_PARAM_NAME = "test_string_param_name"
+val TEST_STRING_PARAM_DEFAULT_VALUE = "test_string_param_default_value"
+val TEST_STRING_PARAM_VALUE = "test_string_param_value"
```
BenHenning: Perhaps instead: ``TEST_STRING_PARAM_SERVER_VALUE``? Note it would be helpful in the KDoc I suggest adding per the above to explain that this value is only used if the mock parameter service is set up in a test (& how to do that, e.g. by adding its corresponding module).

BenHenning (changes requested):
Thanks @ARJUPTA! PTAL at my comments. Also, you have a failing static check--could you please take a look & address that?

Thanks!

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: @BenHenning and @jcqli currently we have an endpoint in Oppia for platform parameters but making any changes to it was discussed to be put in the Future Works in the Proposal. Thats why I have mentioned this as TODO becuase we have not confirmed that whether we will be sending version number to the backend or the names of paramaters to be returned in the response. //cc: @vinitamurthi

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
Arjupta: Yes, that will be better

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/NetworkModule.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done. 
These are the reference files for the allowed value of platform_type

- [https://github.com/oppia/oppia/blob/b1846a839bfb72a5430031fcd2e27638004ad92b/core/domain/platform_parameter_domain.py#L78](https://github.com/oppia/oppia/blob/b1846a839bfb72a5430031fcd2e27638004ad92b/core/domain/platform_parameter_domain.py#L78)
- [https://github.com/oppia/oppia/blob/b1846a839bfb72a5430031fcd2e27638004ad92b/assets/constants.ts#L5830](https://github.com/oppia/oppia/blob/b1846a839bfb72a5430031fcd2e27638004ad92b/assets/constants.ts#L5830)

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameter.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
Arjupta: Done

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameter.kt
Arjupta: Done. 
As currently this response structure is not followed in the Oppia backend so we can't have any reference to such objects.

data/src/main/java/org/oppia/android/data/backends/gae/model/GaePlatformParameters.kt
Arjupta: Done.
Ditto as above

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: Done. Though I am not sure about the difficulty to test, can you explain a bit more?

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
jcqli: Ack, make an issue for it an include the issue # here then.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
Arjupta: I am confused about this suggested change. Can you explain what is the need to split the annotations into a separate file from the constants. 
Also these are just Test constants so I guess it will be okay to put them in one file and we can follow this suggestion with the Prod Constants.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
Arjupta: I am taking up on the suggestion for renaming the constant but for the latter part (documenting their use) I will include it in the Wiki.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
jcqli: Each annotation usually is its own file (ex: https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/feedbackreporting/ReportSchemaVersion.kt)
That said, it also makes sense to me to keep them altogether in this file as they're only test constants but I'll defer to Ben on his thoughts on this.

jcqli (approved):
Overall LGTM, although there is a build error that I'm not familiar with.. couldn't find anything concrete about it online so I re-ran the checks. Recommend keeping an eye on it.

jcqli:
> Overall LGTM, although there is a build error that I'm not familiar with.. couldn't find anything concrete about it online so I re-ran the checks. Recommend keeping an eye on it.

I saw your message that this might be an issue with Bazel & CI. Noting here it might not be resolvable in this PR.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
BenHenning: This actually seems concerning to me. We shouldn't be setting up the model layer for an endpoint format that doesn't exist. Either these API endpoint representations should match the current backend, or the current backend should be updated to fit these. Otherwise, the system isn't going to work. 'Future work' should be optional, not required, pieces of work beyond the defined requirements for the system to work within spec.

BenHenning:
Per https://github.com/oppia/oppia-android/pull/3489#discussion_r673716626 I'm now concerned about the approach for interacting with the Oppia backend. We should be designing against the correct API, and it seems like we might not be.

@ARJUPTA @vinitamurthi

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
Arjupta: Yeah understood now. I will include this thing in the Wiki (as a step while creating a Platform Parameter). Other than that I also prefer to keep test constants in one file.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,122 @@
...
+  @field:[Inject MockPlatformParameterService]
+  lateinit var mockPlatformParameterService: PlatformParameterService
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testPlatformParameterService_getPlatformParameterUsingMockService_verifyTheResponse() {
```
vinitamurthi: Maybe change to `testPlatformParameterService_getPlatformParameterUsingMockService_verifyNonNullResponse` or `testPlatformParameterService_getPlatformParameterUsingMockService_verifySuccessfulResponse` to make it clearer

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    val responseBody = response.body()!!
+    assertThat(responseBody[TEST_INTEGER_PARAM_NAME]).isEqualTo(TEST_INTEGER_PARAM_SERVER_VALUE)
+  }
+
+  @Test
+  fun testPlatformParameterService_getPlatformParameterUsingMockService_checkForBooleanParam() {
+    val response = mockPlatformParameterService.getPlatformParametersByVersion("1").execute()
+    val responseBody = response.body()!!
+    assertThat(responseBody[TEST_BOOLEAN_PARAM_NAME]).isEqualTo(TEST_BOOLEAN_PARAM_SERVER_VALUE)
+  }
```
vinitamurthi, line 104: Do we need a test for a failure case? For e.g. getting a wrong version number?

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: Okay sure

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: A failed response will be handled by Work Manager and not the service itself. Also while testing we use MockRetrofit instance with 100% success rate. So I guess we don't need any failure handling tests.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,122 @@
...
+  lateinit var mockPlatformParameterService: PlatformParameterService
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testPlatformParameterService_getPlatformParameterUsingMockService_verifySuccessfulResponse() {
+    val response = mockPlatformParameterService.getPlatformParametersByVersion("1").execute()
```
anandwana001: named arguments will look good clean here for function `getPlatformParametersByVersion`

anandwana001 (changes requested):
I don't have much here, just a nit suggestion.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: Done. PTAL @anandwana001

vinitamurthi (approved):

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
+package org.oppia.android.data.backends.gae.api
+
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/** Service that provides access to the Oppia platform parameter endpoint. */
+interface PlatformParameterService {
+
+// TODO(#3506): Change the url to point to the correct endpoint when the backend is ready
```
BenHenning: 2 spaces before this TODO (it should be aligned with other members of the class).

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
...
+  @GET("platform_features_evaluation_handler")
+  fun getPlatformParametersByVersion(
```
BenHenning, line 22: This method ought to have a KDoc. I suggest explaining briefly what this does, but if possible I suggest including a permalink to the Oppia web documentation for this endpoint.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    assertThat(responseBody).isNotNull()
+  }
+
+  @Test
+  fun testPlatformParameterService_getPlatformParameterUsingMockService_checkForStringParam() {
+    val response = mockPlatformParameterService
+      .getPlatformParametersByVersion(version = "1")
+      .execute()
+    val responseBody = response.body()!!
+    assertThat(responseBody[TEST_STRING_PARAM_NAME]).isEqualTo(TEST_STRING_PARAM_SERVER_VALUE)
```
BenHenning: More idiomatic:

```suggestion
    assertThat(responseBody).containsEntry(TEST_STRING_PARAM_NAME, TEST_STRING_PARAM_SERVER_VALUE)
```

(May need line wrapping)

Ditto for the other tests.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
```diff
@@ -0,0 +1,130 @@
...
+import javax.inject.Qualifier
+import javax.inject.Singleton
+
+/**
+ * Tests for [PlatformParameterService] retrofit instance using
+ * [org.oppia.android.testing.network.MockPlatformParameterService].
+ */
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+class PlatformParameterServiceTest {
```
BenHenning, line 41: I suggest also testing:
- Different payloads based on different parameters passed in
- Cases when the query fails and an error is thrown (such as 4xx and 5xx errors)

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.oppia.android.testing.platformparameter.TEST_BOOLEAN_PARAM_SERVER_VALUE
+import org.oppia.android.testing.platformparameter.TEST_INTEGER_PARAM_NAME
+import org.oppia.android.testing.platformparameter.TEST_INTEGER_PARAM_SERVER_VALUE
+import org.oppia.android.testing.platformparameter.TEST_STRING_PARAM_NAME
+import org.oppia.android.testing.platformparameter.TEST_STRING_PARAM_SERVER_VALUE
+import retrofit2.Call
+import retrofit2.mock.BehaviorDelegate
+
+/** Mock [PlatformParameterService] to check that the service is properly requested. */
+class MockPlatformParameterService(
```
BenHenning, line 14: You may need additional methods in this class orchestrate it for specific scenarios (like arranging the parameter map, handling cases with different parameters, introducing errors, etc. that we will want to introduce in some test cases.

Also, the different behaviors introduced into this mock ought to be tested with a dedicated MockPlatformParameterServiceTest suite.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
BenHenning: All public members should have KDocs--that's actually expected everywhere in the codebase. It also solves a different purpose than the wiki: wikis provide high-level context or instructions, not specific details on how to use a particular value or component, or what that thing represents.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
BenHenning: What benefit do we gain from keeping them in one file? I think we can evaluate the merits of both approaches.

I lean toward the convention of keeping files as specific as possible (oftentimes having overly generic packages/classes/utilities can result in them growing unboundedly due to not having a clear idea what should go into that component; maintaining specific files helps keep things focused).

I also find the current name confusing. This file really contains annotations and constants, not just contains. Either we should rename it, or keep the annotations and constants in 2 separate files. If we do the latter, I suspect organizing the annotations as individual files with their constants included will be clearer and easier to find when searching for code vs. the alternative considerations. This is another benefit from the per-file convention: it makes code discovery easier.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
BenHenning: Sure. One benefit you can derive from automated tests is that they verify production code is working correctly. However, what verifies that the test itself is correct? It's not reasonable to write tests for the test, so we need to use humans to verify that the tests are correct (i.e. by code reviews, ensuring they pass/fail when expected, and other processes). However, we have a problem: the fact that humans make a lot of mistakes is the reason we all need tests in the first place. How can we rely on ourselves to ensure tests are also correct?

Generally, it comes down to following different coding practices that make tests as easy to understand as possible so that we can quite literally "run them in our heads" when reading them. A large part of test correctness validation comes from code review (unlike production code which comes from running the code and verifying the tests we verified as correct pass against that production code).

Something like control flow is both easy to mess up when writing it, and easy to miss mistakes for reviewers. For that reason, we try to avoid it to make the tests clearer. You may notice a general principle of preferring code de-duplication for production code, but code duplication for tests since generic coding patterns that reduce code duplication can be complicated and difficult to verify as a human with confidence.

Does this help provide more context into the 'why'?

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
BenHenning: Also, this should probably say something more like "Change the URL to point to the updated endpoint."

The current backend _is_ ready, we just want to change it.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,16 @@
...
+    @Query("app_version") version: String,
+    @Query("platform_type") platformType: String = "Android"
```
BenHenning, line 24: We ought to verify that this is being set correctly via the mock.

BenHenning (commented):
Thanks @ARJUPTA! I think besides the few style nits, most of my feedback is digging into improving the testing for the PR.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: Yeah it was a great explanation. Thank you very much

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
Arjupta: Yeah this explaination helps us to understand the need to have separate files. Including this change

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterConstants.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: Done

anandwana001 (approved):
Tests LGTM

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
Arjupta: A failed response will be handled by Work Manager and not the service itself. Also while testing we use MockRetrofit instance with 100% success rate. So I guess we don't need any failure handling tests.

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
Arjupta: Ditto as above

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * This function returns a Map of Platform Parameters where the keys corresponds to parameter
+   * names and values corresponds to their server value.
+   * [Reference](https://github.com/oppia/oppia/blob/develop/core/controllers/platform_feature.py)
+   *
+   * @param version [String] - App version at the time of Build.
+   * @param platformType [String] - Type of client from which the request is made. It should have
+   * a default value of "Android" for Oppia-Android.
+   *
+   * @return [Map]<[String], [Any]> - Platform parameter values mapped to their names.
```
BenHenning:
```suggestion
   * @param version app version at from build time
   * @param platformType type of client from which the request is made. It should have
   *     a default value of "Android" for Oppia-Android.
   * @return platform parameter values mapped to their names
```

Main changes:
- Type information is not explicitly defined (since Kotlin is a strongly typed language, we just need to look at the function declaration to get the types)
- at-param clauses stay together
- No '-'
- Param fragments don't end in punctuation unless they are multi-sentence
- KDoc continuation for at-params is +4 spaces

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,27 @@
...
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/** Service that provides access to the Oppia platform parameter endpoint. */
+interface PlatformParameterService {
+
+  /**
+   * This function returns a Map of Platform Parameters where the keys corresponds to parameter
+   * names and values corresponds to their server value.
+   * [Reference](https://github.com/oppia/oppia/blob/develop/core/controllers/platform_feature.py)
```
BenHenning: Can you pin this to a specific commit hash rather than referencing develop? If the file moves or is renamed in the future, this doc string will become invalid.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/** Service that provides access to the Oppia platform parameter endpoint. */
+interface PlatformParameterService {
+
+  /**
+   * This function returns a Map of Platform Parameters where the keys corresponds to parameter
```
BenHenning:
```suggestion
   * Retrieves map of platform parameters where the keys corresponds to parameter
```

"This function" is implied for the summary fragment and can be omitted (in the same way that "This class" is omitted from class-level KDoc summary fragments).

testing/src/main/java/org/oppia/android/testing/platformparameter/TestStringPlatformParameter.kt
```diff
@@ -0,0 +1,24 @@
+package org.oppia.android.testing.platformparameter
+
+import javax.inject.Qualifier
+
+/**
+ * Qualifier for Test String Platform Parameter. Only used in tests related to Platform Parameter.
```
BenHenning: s/Platform Parameter/platform parameter/

and

s/Test String/test string/

In general, we should only capitalize things which are names. Ditto everywhere else in this PR, plus similar cases (like Integer, Boolean, etc.).

testing/src/main/java/org/oppia/android/testing/platformparameter/BUILD.bazel
```diff
@@ -0,0 +1,41 @@
+# TODO(#1532): Rename file to 'BUILD' post-Gradle.
+"""
+Package for PlatformParameter-specific test utilities and helpers.
```
BenHenning:
```suggestion
Package for test-specific platform parameter utilities and helpers.
```

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
BenHenning: You're welcome--I'm glad that helped.

data/src/test/java/org/oppia/android/data/backends/gae/api/PlatformParameterServiceTest.kt
BenHenning: I think that's reasonable, though testing failure cases via the mock probably is valuable (see my comment below).

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
BenHenning: Which comment are you referring to? I think the needs of testing this mock are quite different than for PlatformParameterService. In particular, we do need a way to simulate different conditions including failures to ensure that the downstream work manager PR handles them correctly. The mock is used to orchestrate these.

While you may not need to add that functionality to the mock in this PR (it could be added in the later PR when it's needed), we should still have a test suite for verifying that this mock properly sets up the platform parameter service. Testing test utilities is pretty important--test utilities with bugs can result in false passing/failing suites for tests that depend on them.

BenHenning (changes requested):
Thanks @ARJUPTA. Had a few nits, but my main feedback is that I think we still need at least a basic test suite for the new mock especially since it'll probably become more complex in subsequent PRs & in the future.

Also, apologies for not reviewing this earlier. I think it slipped off my radar this week somehow.

testing/BUILD.bazel
```diff
@@ -64,6 +64,9 @@ TEST_DEPS = [
     ":dagger",
     ":testing",
     "//domain",
+    "//testing/src/main/java/org/oppia/android/testing/network:network",
```
BenHenning:
```suggestion
    "//testing/src/main/java/org/oppia/android/testing/network",
```

Can be simplified.

BenHenning (changes requested):
Thanks @ARJUPTA. Were all of my previous comments addressed? Some of the comment threads didn't have any replies, and don't seem to correspond to recent changes.

testing/src/main/java/org/oppia/android/testing/network/MockPlatformParameterService.kt
Arjupta: Understood. Added the Tests for MockPlatformParameterService for now

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Understood. Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done. Pointing to the latest commit for this file

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: Done

testing/src/main/java/org/oppia/android/testing/platformparameter/BUILD.bazel
Arjupta: Done

testing/BUILD.bazel
Arjupta: Done

testing/src/main/java/org/oppia/android/testing/platformparameter/TestStringPlatformParameter.kt
Arjupta: Done

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
```diff
@@ -8,15 +8,14 @@ import retrofit2.http.Query
 interface PlatformParameterService {
 
   /**
-   * This function returns a Map of Platform Parameters where the keys corresponds to parameter
-   * names and values corresponds to their server value.
-   * [Reference](https://github.com/oppia/oppia/blob/develop/core/controllers/platform_feature.py)
+   * Retrieves map of platform parameters where the keys corresponds to parameter names and values
+   * corresponds to their server value.
+   * [Reference](https://github.com/oppia/oppia/blob/50f5cffdfccf19dad4d7144addb44a74cad23d3c/core/controllers/platform_feature_test.py)
```
BenHenning: Did you mean to change this to point to the test file? That's not expected, I think, since tests don't technically define the actual contract of a component's interface.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
BenHenning: Also: keep in mind you can just use the first ~6 characters of the hash to keep it shorter (to check, just navigate to the shorter URL & make sure it works correctly).

BenHenning (commented):
Thanks @ARJUPTA!  Just a couple a super small nits left.

data/src/main/java/org/oppia/android/data/backends/gae/api/PlatformParameterService.kt
Arjupta: No I didn't meant to point to test file. My mistake. Changing it to actual code file. (https://github.com/oppia/oppia/blob/50f5cffdfccf19dad4d7144addb44a74cad23d3c/core/controllers/platform_feature.py)

BenHenning (approved):
Thanks @ARJUPTA. LGTM!

PR #3352, opened 2021-06-21
Fix #3300: Add check for accessibility label for activities

Sparsh1212:
## Explanation

Fixes #3300: Add check for accessibility labels for activities

The script parses the `AndroidManifest.xml` file of the `app` layer to ensure that the `activity` elements are defined with accessibility labels.

**Note:** The Activities listed in the `ACCESSIBILITY_LABEL_CHECK_EXEMPTIONS_LIST` will be skipped for this check, and currently, all the activities which were failing this check, have been added to this list so that the workflow doesn't fail.
We can remove any Activity from the exemption list, where we think that we can't go without a label for that Activity, or if a label is added for that Activity.

To run the script, use:

`bazel run //scripts:accessibility_label_check -- $(pwd) app/src/main/AndroidManifest.xml`

For testing the script, automated tests have been added.
To execute the tests, use:

`bazel test //scripts:AccessibilityLabelCheckTest`

**Note:** We are generating the test assets dynamically at the time of executing them. The test assets are automatically deleted when the test finishes.

SS when the CI check fails:

image

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3499, opened 2021-07-17
Fix #3304: Add check to ensure KDoc presence for non-private members

### LINKED ISSUE #3304: Create a Kotlin script to ensure the presence of KDoc for every non-private declaration.
KDoc must be present for every:

- Classes
- Values/fields
- Explicit constructors
- Companion objects
- Nested classes
- Enums
- Annotations
- Interfaces
- Extension functions

Create a script which ensures the same.

**Target PR date:** 2021-07-30
**Target completion date:** 2021-08-04

Sparsh1212:
## Explanation

Fixes #3304: Add check to ensure KDoc presence for non-private members

Script for ensuring that KDocs are present on all non-private:
- Classes
- Functions
- Values/fields
- Explicit constructors
- Companion objects
- Nested classes
- Enums
- Annotations
- Interfaces

Note: If any of the above member has an override modifier present, then it is automatically
exempted for the KDoc check.

Usage:
`bazel run //scripts:kdoc_check -- <path_to_directory_root>`

Arguments:
- path_to_directory_root: directory path to the root of the Oppia Android repository.

Example:
`bazel run //scripts:kdoc_check -- $(pwd)`

For testing the script, automated tests have been added.
To execute the tests, use:

`bazel test //scripts/src/javatests/org/oppia/android/scripts/kdoc:KDocCheckTest`

**Note:** We are generating the test assets dynamically at the time of executing them. The test assets are automatically deleted when the test finishes.

SS when the CI check fails:

image

## Checklist


Sparsh1212:
@BenHenning PTAL.

BenHenning:
Will need to take a look tomorrow, I think. Apologies for the delay.

BenHenning:
Apologies, I think I might need to delay reviewing this one more day.

.github/workflows/static_checks.yml
```diff
@@ -123,3 +123,8 @@ jobs:
         if: always()
         run: |
           bazel run //scripts:accessibility_label_check -- $(pwd) app/src/main/AndroidManifest.xml
+
+      - name: KDoc Check
+        if: always()
+        run: |
+          bazel run //scripts:kdoc_check -- $(pwd)
```
BenHenning: Maybe instead: 'KDoc validity check' would be clearer? The current name is a bit too non-specific, and we do want to eventually extend this script to include other checks.

Ditto for the step name above & class/library names elsewhere in the PR.

scripts/src/java/org/oppia/android/scripts/kdoc/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+"""
+Libraries corresponding to kdoc presence check, that ensures all public members (except overrides)
```
BenHenning: This & the test counterpart should be more generic: this package contains scripts corresponding to KDocs in the codebase. Presence is one such example, and there may be more in the future.

That being said, I actually suggest making this package even more generic. Could we call it 'docs'? We probably will only ever have one script for verifying KDocs, but I could see the need for validating other forms of documentation & comments in the future.

third_party/versions.bzl
```diff
@@ -90,6 +90,7 @@ MAVEN_TEST_DEPENDENCY_VERSIONS = {
     "com.squareup.okhttp3:mockwebserver": "4.1.0",
     "com.squareup.retrofit2:retrofit-mock": "2.5.0",
     "junit:junit": "4.12",
+    "org.jetbrains.kotlin:kotlin-compiler-embeddable": "1.5.0",
```
BenHenning, line 97: @prayutsu will we need to also update the license textproto file, or will you be doing that in a later PR (given that there's no CI check currently to ensure it stays up-to-date?).

scripts/script_assets.bzl
```diff
@@ -112,3 +112,27 @@ def generate_accessibility_label_assets_list_from_text_protos(
         proto_dep_bazel_target_prefix = "//scripts/src/java/org/oppia/android/scripts/proto",
         proto_package = "proto",
     )
+
+def generate_kdoc_assets_list_from_text_protos(
+        name,
+        kdoc_exemptions_name):
+    """
+    Converts multiple lists of text proto assets to binary.
```
BenHenning: Is this docstring correct?

scripts/script_assets.bzl
```diff
@@ -112,3 +112,27 @@ def generate_accessibility_label_assets_list_from_text_protos(
...
+
+    Args:
+        name: str. The name of this generation instance. This will be a prefix for derived targets.
+        kdoc_exemptions_name: list of str. The list of kdoc exemptions file names.
```
BenHenning: It's just one file, right?

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+  config.put(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, MessageCollector.NONE)
+  KotlinCoreEnvironment.createForProduction(
+    Disposer.newDisposable(),
+    config,
+    EnvironmentConfigFiles.JVM_CONFIG_FILES
+  ).project
+}
+
+/**
+ * Genrates a [KtFile] from a Kotlin file.
```
BenHenning: Typo: Generates

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ */
+private fun File.asString(): String {
+  val fileContents = StringBuilder(this.length().toInt())
+  Scanner(this).use { scanner ->
+    while (scanner.hasNextLine()) {
+      fileContents.append(scanner.nextLine() + System.lineSeparator())
+    }
+    return fileContents.toString()
+  }
+}
```
BenHenning: Could we use [``File.readText()``](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/read-text.html) instead?

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * @param file the file to be checked
+ * @param kDocNotRequiredAnnotationEntryList the list of annotation entries which when present
+ *     on an element, the element does not needs to be checked for a KDoc.
+ * @return whether the file has a KDoc presence failure
+ */
+private fun hasKdocFailure(file: File, kDocNotRequiredAnnotationEntryList: List<String>): Boolean {
+  val ktFile = createKtFile(file.asString(), file.name)
+  return ktFile.children.fold(initial = false) { isFailing, elem ->
+    val childFailingKdoc = if (elem is KtElement) elementIterator(
```
BenHenning: This if-check should have curly braces. We only omit curly braces on statements if they fit entirely on one line.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ * @param kDocNotRequiredAnnotationEntryList the list of annotation entries which when present
+ *     on an element, the element does not needs to be checked for a KDoc.
+ * @return whether the element is missing a KDoc
+ */
+private fun elementIterator(
+  elem: KtElement,
+  file: File,
+  kDocNotRequiredAnnotaionEntryList: List<String>
+): Boolean {
+  if (elem is KtClass) {
```
BenHenning: I suggest converting this if-else-if block into a ``when`` block.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * Recursively iterates over an element and checks for a KDoc presence.
+ *
+ * @param elem the element over which we have to iterate
+ * @param file the file to be checked for the KDoc presence
+ * @param kDocNotRequiredAnnotationEntryList the list of annotation entries which when present
+ *     on an element, the element does not needs to be checked for a KDoc.
+ * @return whether the element is missing a KDoc
+ */
+private fun elementIterator(
```
BenHenning: I'm not sure how to understand what this method is meant to do from the name alone. Could you rename it to be something more specific as to what it's actually doing? Generally, method names are verb phrases since they describe an action that's taking place.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * Checks if an element is missing a KDoc.
+ *
+ * @param elem the element to be checked
+ * @param file the file to be checked for KDoc presence
+ * @param kDocNotRequiredAnnotationEntryList the list of annotation entries which when present
+ *     on an element, the element does not needs to be checked for a KDoc.
+ * @return whether the element is missing a KDoc
+ */
+private fun checkIfKDocMissing(
```
BenHenning:
```suggestion
private fun checkIfKDocIsMissing(
```

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+    return classKdocMissing || memberMissingKdoc
+  } else if (elem is KtObjectDeclaration && elem.isCompanion()) {
+    val memberMissingKdoc = elem.declarations.fold(initial = false) { isMissingKdoc, childElem ->
+      val childMissingKdoc = elementIterator(childElem, file, kDocNotRequiredAnnotaionEntryList)
+      isMissingKdoc || childMissingKdoc
+    }
+    return memberMissingKdoc
+  } else if (elem is KtNamedFunction ||
+    elem is KtVariableDeclaration ||
+    elem is KtSecondaryConstructor
```
BenHenning: Can this fit on one line?

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+  kDocNotRequiredAnnotaionEntryList: List<String>
+): Boolean {
+  if (elem is KtClass) {
+    val classKdocMissing = checkIfKDocMissing(elem, file, kDocNotRequiredAnnotaionEntryList)
+    val memberMissingKdoc = elem.declarations.fold(initial = false) { isMissingKdoc, childElem ->
+      val childMissingKdoc = elementIterator(childElem, file, kDocNotRequiredAnnotaionEntryList)
+      isMissingKdoc || childMissingKdoc
+    }
+    return classKdocMissing || memberMissingKdoc
+  } else if (elem is KtObjectDeclaration && elem.isCompanion()) {
```
BenHenning: Is the companion check needed? I'm guessing regular object declarations should also have KDocs.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+  }
+  if (elem.docComment == null) {
+    println("$file:${retrieveLineNumberForElement(elem, false)}: missing KDoc")
+    return true
+  }
+  return false
+}
+
+/**
+ * Returns whether an element is required to be checked for a KDoc presence
```
BenHenning:
```suggestion
 * Returns whether an element is required to be checked for a KDoc presence.
```

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ * @param statement the [KtElement] for which we want to get the line number
+ * @param markEndOffset Whether to mark the end offset for the given element. For example:
+ *     when set to true, the function will return the line number where the statement ends.
+ *     Similarly, when this is set to false, the function returns the line number where the
+ *     statement starts.
+ * @return the line number
+ */
+private fun retrieveLineNumberForElement(statement: KtElement, markEndOffset: Boolean): Int? {
+  val file = statement.containingFile
+  if (file is KtFile && file.doNotAnalyze != null) {
```
BenHenning: Why is the doNotAnalyze check needed?

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+    return null
+  }
+  val document = file.viewProvider.document
+  return document?.getLineNumber(
+    if (markEndOffset) statement.textRange.endOffset else statement.textOffset
+  )?.plus(1)
+}
+
+/**
+ * Loads the KDoc exemptions list to proto.
```
BenHenning: From a proto file?

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * @param kdocExemptiontextProto the location of the kdoc exemption textproto file
+ * @return proto class from the parsed textproto file
+ */
+private fun loadKdocExemptionsProto(kdocExemptiontextProto: String): KDocExemptions {
+  val protoBinaryFile = File("$kdocExemptiontextProto.pb")
```
BenHenning: Typo: TextProto

Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/proto/kdoc_exemptions.proto
```diff
@@ -0,0 +1,22 @@
...
+option java_package = "org.oppia.android.scripts.proto";
+option java_multiple_files = true;
+
+// Exemptions for the KDoc check. Exemptions indicate files for which we should
+// not check for the presence of a KDoc.
+message KDocExemptions {
+  // List of all the files which should be exempted for the KDoc check.
+  // For exempting any file, provide its relative path to root in
+  // 'script/assets/kdoc_exemptions.textproto'.
+  // Also, note that the exemptions in the textproto file are maintained in lexicographical order.
```
BenHenning:
```suggestion
  // Also, note that the exemptions in the text proto file are maintained in lexicographical order.
```

Ditto below.

scripts/src/javatests/org/oppia/android/scripts/kdoc/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+"""
+Tests corresponding to kdoc presence check, that ensures all public members (except overrides) must
+have a corresponding KDoc.
```
BenHenning: As mentioned earlier, this docstring is a bit too specific to the exact check being implemented. Please generalize it.

scripts/src/javatests/org/oppia/android/scripts/kdoc/KDocCheckTest.kt
```diff
@@ -0,0 +1,487 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import org.oppia.android.testing.assertThrows
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+
+/** Tests for [KDocCheck]. */
+class KDocCheckTest {
```
BenHenning: Other tests that I think are worth adding:
- A block comment instead of a KDoc (to ensure KDocs are specifically checked).
- Inject primary constructor (just to make sure nothing special is triggered in that situation)
- Object vals
- Constants
- Companion objects
- Nested interfaces
- Nested annotations
- All other special case annotations not already tested below (all of them ought to be checked)
- The special case of the exemptions list having a redundant file in it (see my comment above)

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
```diff
@@ -0,0 +1,284 @@
...
+ * - path_to_directory_root: directory path to the root of the Oppia Android repository.
+ *
+ * Example:
+ *   bazel run //scripts:kdoc_check -- $(pwd)
+ */
+fun main(vararg args: String) {
+  // Path of the repo to be analyzed.
+  val repoPath = "${args[0]}/"
+
+  val kdocExemptiontextProto = "scripts/assets/kdoc_exemptions"
```
BenHenning: One scenario I think we missed in past checks that is especially important for this one is outputting an error when the exemption list contains something no longer relevant. I think we actually want to make sure that, in general, exemption lists are always the complete picture on the status of the codebase.

BenHenning (changes requested):
Thanks @Sparsh1212! Took a full pass.

third_party/versions.bzl
prayutsu: As this is a test dependency, we'll not need to update the textproto file.

vinitamurthi (approved):
LGTM for the codeowner files on the assumption that Ben's comments will be addressed

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Actually, if we use `File.readText()`, then it gives this error while parsing one of the files because of the line separator, hence I had to go by this approach.

```
Exception in thread "main" java.lang.AssertionError: Wrong line separators: '....gae.model\r\n\r\nimport...' at offset 49
        at org.jetbrains.kotlin.com.intellij.openapi.util.text.StringUtil.assertValidSeparators(StringUtil.java:2450)
        at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.assertValidSeparators(DocumentImpl.java:690)
        at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.<init>(DocumentImpl.java:119)
        at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.<init>(DocumentImpl.java:114)
        at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.<init>(DocumentImpl.java:105)
        at org.jetbrains.kotlin.com.intellij.core.CoreApplicationEnvironment.lambda$new$1(CoreApplicationEnvironment.java:94)
        at org.jetbrains.kotlin.com.intellij.mock.MockFileDocumentManagerImpl.getDocument(MockFileDocumentManagerImpl.java:42)
        at org.jetbrains.kotlin.com.intellij.psi.AbstractFileViewProvider.getDocument(AbstractFileViewProvider.java:180)
        at org.oppia.android.scripts.docs.KDocValidityCheckKt.retrieveLineNumberForElement(KDocValidityCheck.kt:260)
        at org.oppia.android.scripts.docs.KDocValidityCheckKt.checkIfKDocMissing(KDocValidityCheck.kt:208)
        at org.oppia.android.scripts.docs.KDocValidityCheckKt.elementIterator(KDocValidityCheck.kt:169)
        at org.oppia.android.scripts.docs.KDocValidityCheckKt.hasKdocFailure(KDocValidityCheck.kt:130)
        at org.oppia.android.scripts.docs.KDocValidityCheckKt.main(KDocValidityCheck.kt:86)
```

.github/workflows/static_checks.yml
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/BUILD.bazel
Sparsh1212: Done.

scripts/script_assets.bzl
Sparsh1212: Corrected now.

scripts/script_assets.bzl
Sparsh1212: Corrected now.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: As per the discussion, we had, its now updated.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Apologies! this was redundant. Removed.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/proto/kdoc_exemptions.proto
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/kdoc/BUILD.bazel
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/kdoc/KDocCheckTest.kt
Sparsh1212: Added.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
Sparsh1212: Added.

Sparsh1212:
@BenHenning PTAL.
- All comments addressed
- Added check for redundant exemptions
- Added tests for redundant exemptions
- Added new tests as suggested for KDocs

scripts/src/java/org/oppia/android/scripts/docs/BUILD.bazel
```diff
@@ -1,18 +1,17 @@
 """
-Libraries corresponding to kdoc presence check, that ensures all public members (except overrides)
-must have a corresponding KDoc.
+Libraries corresponding docs validity check, that ensures the documentations are valid.
```
BenHenning:
```suggestion
Libraries corresponding to documentation-related scripts.
```

Can be simpler. Ditto for the tests version.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
```diff
@@ -1,4 +1,4 @@
-package org.oppia.android.scripts.kdoc
+package org.oppia.android.scripts.docs
```
BenHenning: File: ``KdocValidityCheck``

(note camel-casing/capitalization). Ditto for the proto message name.

We consider "kdoc" to be one word which means as a unit it's capitalized/lowercase rather than being partly.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
```diff
@@ -72,24 +74,40 @@ fun main(vararg args: String) {
   )
 
   // A list of all the files to be exempted for this check.
-  val kdocExemptionList = loadKdocExemptionsProto(kdocExemptiontextProto).getExemptedFilePathList()
+  val kdocExemptionList = loadKdocExemptionsProto(pathToProtoBinary)
+    .getExemptedFilePathList()
```
BenHenning:
```suggestion
  val kdocExemptionList =
    loadKdocExemptionsProto(pathToProtoBinary).getExemptedFilePathList()
```

Wrapping should be at the higher syntactical element (in this case the ``=``).

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
```diff
@@ -122,17 +140,23 @@ private fun createKtFile(codeString: String, fileName: String): KtFile {
  * @param file the file to be checked
  * @param kDocNotRequiredAnnotationEntryList the list of annotation entries which when present
  *     on an element, the element does not needs to be checked for a KDoc.
- * @return whether the file has a KDoc presence failure
+ * @return Returns list of elements in the file which are missing KDoc
```
BenHenning:
```suggestion
 * @return a list of elements in the file which are missing KDocs
```

Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
```diff
@@ -158,33 +182,52 @@ private fun File.asString(): String {
...
+    elem is KtClass -> {
+      val classKdocMissing = checkIfKDocIsMissing(elem, file, kDocNotRequiredAnnotationEntryList)
+      val memberMissingKdoc = elem.declarations.flatMap { childElem ->
+        recursiveKdocPresenceChecker(
+          childElem,
+          file,
+          kDocNotRequiredAnnotationEntryList
+        )
+      }
+      return listOf(listOf(classKdocMissing), memberMissingKdoc).flatten()
```
BenHenning: Can be simplified I think to:

```suggestion
      return memberMissingKdoc + classKdocMissing
```

Ditto elsewhere.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
```diff
@@ -122,17 +140,23 @@ private fun createKtFile(codeString: String, fileName: String): KtFile {
...
-    isFailing || childFailingKdoc
+  return ktFile.children.flatMap { elem ->
+    if (elem is KtElement) {
+      recursiveKdocPresenceChecker(
+        elem,
+        file,
+        kDocNotRequiredAnnotationEntryList
+      )
+    } else {
+      listOf(null)
```
BenHenning: Why not return empty list, instead? Ditto below.

Prefer to avoid nullable unless needed.

scripts/src/java/org/oppia/android/scripts/proto/script_exemptions.proto
```diff
@@ -35,3 +35,19 @@ message AccessibilityLabelExemptions {
   // at the correct lexicographical position in the textproto file.
   repeated string exempted_activity = 1;
 }
+
+// Exemptions for the KDoc validity check. Exemptions indicate files for which we should
+// not check for the presence of a KDoc.
```
BenHenning:
```suggestion
// not check for the presence of KDocs.
```

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -118,6 +123,119 @@ class KDocCheckTest {
...
+    val tempFile = tempFolder.newFile("testfiles/TempFile.kt")
+    tempFile.writeText(testContent)
+
+    runScript()
+
+    assertThat(outContent.toString().trim()).isEqualTo(KDOC_CHECK_PASSED_OUTPUT_INDICATOR)
+  }
+
+  @Test
+  fun testKDoc_fieldWithRuleAnnotationWithoutKDoc_checkShouldPass() {
```
BenHenning:
```suggestion
  fun testKdoc_fieldWithRuleAnnotationWithoutKdoc_checkShouldPass() {
```

Ditto for other tests (per camel-casing guidelines).

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -118,6 +123,119 @@ class KDocCheckTest {
...
+
+    assertThat(outContent.toString().trim()).isEqualTo(KDOC_CHECK_PASSED_OUTPUT_INDICATOR)
+  }
+
+  @Test
+  fun testKDoc_functionWithProvidesAnnotationWithoutKDoc_checkShouldPass() {
+    val testContent =
+      """
+      @Provides
+      @ExplorationStorageDatabaseSize
```
BenHenning: What about a test where something just has a qualifier? That seems like a nice failure case to cover.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -154,6 +272,22 @@ class KDocCheckTest {
     assertThat(outContent.toString().trim()).isEqualTo(KDOC_CHECK_PASSED_OUTPUT_INDICATOR)
   }
 
+  @Test
+  fun testKDoc_injectPrimaryConstructor_noSpecialBehaviour_checkShouldPass() {
```
BenHenning: _withKdoc?

Ditto for other passing cases (to contrast with the 'withoutKdoc' tests). Similarly, I suggest separating out "_withoutKdoc" in those test names to be a bit clearer that it's a separate condition being implemented.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -154,6 +272,22 @@ class KDocCheckTest {
...
+    val testContent =
+      """
+      /** Test Class. */
+      class SelectionFragmentModel @Inject constructor(){
```
BenHenning:
```suggestion
      class SelectionFragmentModel @Inject constructor() {
```

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -246,8 +383,84 @@ class KDocCheckTest {
...
+      - ${retrieveTestFilesDirectoryPath()}/TempFile.kt:2
+      """.trimIndent()
+    assertThat(outContent.toString().trim()).isEqualTo(failureMessage)
+  }
+
+  @Test
+  fun testKDoc_blockCommentInsteadOfKDoc_checkShouldFail() {
+    val testContent =
+      """
+      // Test block comment.
```
BenHenning: This is a line comment, not a block comment (suggest updating the test name & adding a test for a block comment).

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -463,25 +696,101 @@ class KDocCheckTest {
...
+    val exemptions = builder.build()
+    exemptions.writeTo(exemptionFile.outputStream())
 
     runScript()
 
     assertThat(outContent.toString().trim()).isEqualTo(KDOC_CHECK_PASSED_OUTPUT_INDICATOR)
   }
 
+  @Test
+  fun testKDoc_redundantExemption_checkShouldFail() {
```
BenHenning: I suggest framing this as "fileWithAllExpectedKdocs_redundantExemption_..." and then add a test for the case where it's exempted and only has some of the kdocs--that shouldn't fail since the file hasn't been fully resolved yet.

BenHenning (changes requested):
Thanks @Sparsh1212. Had a few more comments--PTAL.

scripts/src/java/org/oppia/android/scripts/kdoc/KDocCheck.kt
BenHenning: Ah. You could maybe do something like:

```kotlin
file.readLines().joinToString(separateor = System.lineSeparator())
```

Would that work?

third_party/versions.bzl
BenHenning: Ah good point. Thanks!

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
```diff
@@ -0,0 +1,796 @@
...
+      - ${retrieveTestFilesDirectoryPath()}/TempFile.kt:3
+      - ${retrieveTestFilesDirectoryPath()}/TempFile.kt:5
+      - ${retrieveTestFilesDirectoryPath()}/TempFile.kt:7
+      - ${retrieveTestFilesDirectoryPath()}/TempFile.kt:8
+      """.trimIndent()
+    assertThat(outContent.toString().trim()).isEqualTo(failureMessage)
+  }
+
+  @Test
+  fun testKDoc_companionObjectMissingKDoc_checkShouldFail() {
```
BenHenning: I suggest making this test specific to just the companion object and add another test for a companion object with constituent values & functions to make it more clear the specific expected behavior for companion objects.

Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Added.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/proto/script_exemptions.proto
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/docs/KDocValidityCheck.kt
Sparsh1212: Done.

scripts/src/java/org/oppia/android/scripts/docs/BUILD.bazel
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
BenHenning: Was this addressed? The latest code version doesn't seem to include the space.

BenHenning (changes requested):
Thanks @Sparsh1212! I think this looks really great--the test name changes make them *much* clearer, I think.

I just noticed one comment didn't seem resolved, but I otherwise have nothing new to add.

Could you please fix the deltas on this PR & its title since #3352 was merged?

scripts/src/javatests/org/oppia/android/scripts/docs/KDocValidityCheckTest.kt
Sparsh1212: Apologies! Just missed this. Now updated.

Sparsh1212:
> Thanks @Sparsh1212! I think this looks really great--the test name changes make them _much_ clearer, I think.
> 
> I just noticed one comment didn't seem resolved, but I otherwise have nothing new to add.
> 
> Could you please fix the deltas on this PR & its title since #3352 was merged?

Thanks @BenHenning the PR is now updated and the one left a comment is also addressed. PTAL.
Also assigning to @anandwana001 for a pass.

.github/workflows/static_checks.yml
```diff
@@ -123,3 +123,8 @@ jobs:
         if: always()
         run: |
           bazel run //scripts:accessibility_label_check -- $(pwd) scripts/assets/accessibility_label_exemptions.pb app/src/main/AndroidManifest.xml
+
+      - name: KDoc validity Check
```
anandwana001:
```suggestion
      - name: KDoc Validation Check
```

scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt
```diff
@@ -0,0 +1,347 @@
...
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.KtSecondaryConstructor
+import org.jetbrains.kotlin.psi.KtVariableDeclaration
+import org.oppia.android.scripts.common.RepositoryFile
+import org.oppia.android.scripts.proto.KdocValidityExemptions
+import java.io.File
+import java.io.FileInputStream
+
+/**
+ * Script for ensuring that KDocs validity on all non-private:
```
anandwana001:
```suggestion
 * Script for ensuring the KDocs validity on all non-private:
```

scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt
```diff
@@ -0,0 +1,347 @@
...
+ * Arguments:
+ * - path_to_directory_root: directory path to the root of the Oppia Android repository.
+ * - path_to_proto_binary: relative path to the exemption .pb file.
+ *
+ * Example:
+ *   bazel run //scripts:kdoc_validity_check -- $(pwd) scripts/assets/kdoc_validity_exemptions.pb
+ */
+fun main(vararg args: String) {
+  // Path of the repo to be analyzed.
+  val repoPath = "${args[0]}/"
```
anandwana001, line 54: use the same name given above in comment - `path_to_directory_root` in camel case

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
```diff
@@ -0,0 +1,875 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.File
+import java.io.PrintStream
+
+/** Tests for [KdocValidityCheck]. */
+class KdocValidityCheckTest {
+  private val outContent: ByteArrayOutputStream = ByteArrayOutputStream()
+  private val originalOut: PrintStream = System.out
+  private val KDOC_CHECK_PASSED_OUTPUT_INDICATOR = "KDOC VALIDITY CHECK PASSED"
+  private val KDOC_CHECK_FAILED_OUTPUT_INDICATOR = "KDOC VALIDITY CHECK FAILED"
```
anandwana001: `private const val ` and make them global

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
```diff
@@ -0,0 +1,875 @@
...
+      /** 
+       * Returns the string corresponding to this error's string resources, or null if there 
+       * is none.
+       */
+      fun getErrorMessageFromStringRes(context: Context): String? {
+        return error?.let(context::getString)
+      }
+      """.trimIndent()
+    val tempFile = tempFolder.newFile("testfiles/TempFile.kt")
+    tempFile.writeText(testContent)
```
anandwana001, line 54: you can create a helper function at bottom of this file and use it here for writing text file

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
```diff
@@ -0,0 +1,875 @@
...
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import org.oppia.android.scripts.proto.KdocValidityExemptions
+import org.oppia.android.testing.assertThrows
+import java.io.ByteArrayOutputStream
+import java.io.File
+import java.io.PrintStream
+
+/** Tests for [KdocValidityCheck]. */
+class KdocValidityCheckTest {
```
anandwana001, line 19:
```suggestion
class KDocValidityCheckTest {
```

scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt
anandwana001: Also, make file name `KDocValidityCheck.kt`

anandwana001 (changes requested):
Thanks @Sparsh1212 
Took a quick pass on this. Left few suggestions. Also, please check on failing actions.

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
Sparsh1212: Ah! this has been discussed previously, and it was decided to name it as "Kdoc", as "Kdoc" is a single entity in itself.
Ditto for the script name.

scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt
Sparsh1212: Actually, we are using the same name "repoPath" for all the scripts till now, so will it be better to keep this the same in order to maintain consistency across all the scripts?

scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt
Sparsh1212: Explained below. PTAL.

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
Sparsh1212: Hmm, we just want to use `writeText` here in order to write the test file, so do we actually need to create a separate function for that? Also, this will make this script a bit inconsistent with all the other scripts, as there we are following the current approach.

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
Sparsh1212: Done.

.github/workflows/static_checks.yml
Sparsh1212: Done.

Sparsh1212:
> Thanks @Sparsh1212
> Took a quick pass on this. Left few suggestions. Also, please check on failing actions.

Thanks, @anandwana001 followed up on all the suggestions. PTAL.

.github/workflows/static_checks.yml
Sparsh1212: Done.

scripts/src/javatests/org/oppia/android/scripts/docs/KdocValidityCheckTest.kt
BenHenning: For context, see also: https://developer.android.com/kotlin/style-guide#camel_case for why it should be 'Kdoc' and not 'KDoc' here.

BenHenning (approved):
Thanks @Sparsh1212! LGTM.

anandwana001 (approved):
LGTM

Issue #2476, opened 2021-01-14
Revealing second hint after orientation change before the timer


FareesHussain:
**Describe the bug**
For the questions with more than one hint, the second hint is revealed before 10 seconds after orientation change

**To Reproduce**
1. Go to `What is a Fraction` story inside `Fractions`
2. click on `continue`, `continue`, `Three fifths`, `Four sevenths`, `2nd option (green)`
3. Here select the option `7/2` twice and reveal the hint
4. go back to question
5. change to the landscape
6. see the 2nd revealed hint

**Expected behavior**
Only reveal second hint after 10sec timer

FareesHussain:
cc: @BenHenning

BenHenning:
So I think that this might because of [this line](https://github.com/oppia/oppia-android/blob/74db288/app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt#L218). It seems like in some cases it might force a new hint to be shown when compute() is called (which will happen on configuration change since the adapter & list is recreated). I'm guessing that we need to build in some additional check to ensure we're only ever showing a hint when there's legitimately a new one to show.

BenHenning:
We should also make sure that this specific cases & related cases are thoroughly tested, too.

FareesHussain:
@BenHenning Can we do something like restart the timer after orientation change

imraghav20:
@FareesHussain can I work on this issue?

imraghav20:
I tried to make few changes, although the issue is still not solved.
Should I share the changes in code in a comment or make a PR to review and suggest appropriate changes.

FareesHussain:
I'm sorry I forgot to assign myself actually I'm currently working on it

Issue #2571, opened 2021-01-27
Make app resilient to state loss due to low-memory process deaths


BenHenning:
Android will kill apps when they are backgrounded & the system runs low on memory. This manifests as a process death, and returning to the app (e.g. via backstack navigation or from the launcher) will trigger the app reopening & restoring via a bundle that the OS had captured & persisted on-disk prior to process death.

Note that we can simulate this scenario by (note that this requires a rooted device/emulator):
- Run ``adb root`` (if your ADB isn't already in root mode; this may not be needed for emulators)
- Running ``adb shell ps | grep oppia``
- Pass the PID found via the above to ``adb shell kill -9 <pid>`` to simulate a kill-on-OOM device

Non-rooted way to simulate this (thanks to https://stackoverflow.com/a/17829677):
- Run ``adb shell ps -A | grep oppia``
- Run ``adb shell run-as org.oppia.android kill -9 <pid>``

Bash one-liner:
```
adb shell run-as org.oppia.android kill -9 "$(adb shell ps -A | grep oppia | awk '{print $2}')"
```

(The current flow causes a crash in Oppia when in an exploration).

We can fix this issue by:
- Introducing a common infrastructure in the domain layer to track all ephemeral app state
- Introduce a mechanism that all Oppia activities can default opt into to pull/restore the app state (this will be easier after #2561 is merged in since all of the state can be represented via a single proto message)

This requires some architectural considerations both in the domain & app layer, so considering this to be a 'mini project'.

prayutsu:
@BenHenning I am interested to work on this.

kartikeysaran:
@BenHenning Can I work on this project if no one working on this !

rt4914:
@prayutsu Can you please confirm if you are working on this or not, else it can be assigned to @kartikeysaran

prayutsu:
> @prayutsu Can you please confirm if you are working on this or not, else it can be assigned to @kartikeysaran

I think I would not be able to work on this as I already have some other issues. So, we can assign this to 
@kartikeysaran.

prayutsu:
@rt4914 As I am busy with other issues, I may not be able to prioritize this, so @kartikeysaran can take it. Thanks!

kartikeysaran:
@rt4914 , I am working on this issue

FareesHussain:
@kartikeysaran any updates on this?

kartikeysaran:
Was working on other issues will work on this

BenHenning:
[...]

PR #2557, opened 2021-01-24
Fix #2476: Fixed Revealing Second hint due to orientation change [Blocked on #2572]

FareesHussain:
## Explanation

Fixes #2476 

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3507, opened 2021-07-20
Fix #3488: Domain layer mechanism to resume explorations. [Blocked: #2476]

### LINKED ISSUE #3488: Implement domain layer mechanism to resume explorations with checkpoints.
Create a mechanism to resume explorations with a saved checkpoint.

Target | Date
--|--
**PR creation:** | 2021-07-27
**PR completion:** | 2021-08-01

MaskedCarrot:
## Explanation

Fixes #3488 

## Checklist


MaskedCarrot:
@BenHenning Can you please take a look at these changes. I wanted your review because I am not sure if passing null (done in TopicLesssonsFragmentPresenter) is a good way to do this but right now this seems to be the only option to me.

So what I am doing here is `TopicLessonsFragment` ( or any other app module class that starts the exploration) will either launch `ExplorationActivity` or `ResumeExplorationActivity` (a new activity that will be added in one of the following PRs) depending upon the chapterPlayState ( if chapter play state is `in_progress_saved/not_saved` then `ResumeExplorationActivity` will be launched) 

If ExplorationActivity is started there is no need to fetch the checkpoint so we will start the exploration and pass `null` in `startPlayingExploration` method of `ExplorationDataController`. 

If the ResumeExplorationActivity is launched, ExplorationCheckpoint will be fetched. Then null or the fetched checkpoint will be passed the `startPlayingExploration` method of `ExplorationDataController` depending upon the option that the user chooses (start over or resume ).

Once the exploration is loaded, we will check if the checkpoint is null, then we will reset the StateDeck but if it is not null we will resume the exploration with the saved checkpoint.

BenHenning:
@MaskedCarrot I feel like feeding the checkpoint state back down to the domain layer is a bit of an inversion. Why not have the controller automatically fetch the checkpoint & resume if it's present? That seems like it would be an overall simplification to the system.

If there are other requirements I'm missing, that context might help provide insight on potentially a better communication strategy here vs. sending null.

MaskedCarrot:
@BenHenning Checkpoint state will not be used anywhere. Chapter play state will be used to decide if the exploration can be resumed or not ( exploration can only be resumed if it is IN_PROGRESS_SAVED/NOT_SAVED and the user chooses to resume the exploration). Chapter play state will not be sent to the domain layer. It will be used in the app layer to either launch exploration activity (when exploration does not resume) or ResumeExploration activity ( when the user has to decide if they wish to resume).

If we retrieve checkpoints in the domain layer then we will again face a similar problem of performing an async task in the domain layer as we did while saving checkpoints. To solve this we can change the retrieveCheckpoints function to return a deferred as we did to save checkpoints and then in invoke on completion method of this deferred we can load the state deck. But doing this will introduce some delay (I think it would be negligible to the user) in the exploration being loaded.

We can not load the exploration and retrieve the checkpoint simultaneously, we will have to start the function to fetch the checkpoint once the exploration is loaded because in invoke on completion method of the deferred we will have to load the statedeck and for that, the exploration has to be loaded first.

BenHenning:
I see. In that case, passing through the app layer sounds good. Could we pass an empty checkpoint for the case when no checkpoint is available rather than null? That's probably a bit cleaner since proto defines an automatic "blank" sentinel value in its default instances. Would that work?

MaskedCarrot:
yes that will work

MaskedCarrot:
@BenHenning: An existing bug in the hint handler allows the user to view a new hint immediately on configuration change if the first hint was revealed. This is happening because if the configuration is changed, the variables of the hint handler are reset (because StatePlayerRecyclerViewAssember is recreated), then the app shows a new hint immediately (since the number of wrong answers > 1 so the first hint should have been revealed immediately). Although the variables of the hint handler are reset, the correct hint is shown because the hint index for the next hint is saved in StateFragmentViewModel. But this is a problem in the case of resume exploration with checkpoints because in that case, the ViewModel will not have the index of the next hint.

I made some changes to StatePlayerRecyclerViewAssember and StateFragmentPresenter in this PR, and I tested this locally, and these changes seem to fix this issue and also allows resuming explorations with the correct hints visible. Can you also take a look to verify if this is the correct way of doing this?

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
```diff
@@ -112,7 +112,7 @@ class HintsAndSolutionDialogFragmentPresenter @Inject constructor(
         bindingAdapter = createRecyclerViewAdapter()
         adapter = bindingAdapter
       }
-      if (viewModel.newAvailableHintIndex.get() != -1) {
+      if (viewModel.newAvailableHintIndex.get() != -2) {
```
BenHenning: This seems slightly hacky to me--why is '-2' correct? It seems like a magic number since there isn't any context of where it's coming from.

BenHenning:
@MaskedCarrot I think #2557 from @FareesHussain was meant to fix that bug. Would the fix proposed in that PR address the issue? Rather than restoring the previous index in this way, I'm wondering whether we should store the checkpoint in the saved instance state bundle since then we can keep it in sync with other ephemeral data, and restore hints at the same time as retrieving the most recent known checkpoint state. WDYT?

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
MaskedCarrot: @BenHenning I think this value should be -2 instead of -1 because we are multiplying the newAvailableHintIndex with 2 in handleOnCreate method to account for the divider, so this value will never be -1.
```
    viewModel.newAvailableHintIndex.set(
      newAvailableHintIndex * RECYCLERVIEW_INDEX_CORRECTION_MULTIPLIER
    )
```

MaskedCarrot:
When an exploration is resumed the variable previousHintIndex will always be either HelpIndex.getDefaultInstance (when no hint is revealed and checkpoint has hintIndex as -1) or HelpIndex.setEverytingRevealed(true).build() (when some/all hint or solution are revealed and checkpiont has hintIndex != -1).

This means that if a user exits the exploration when an unrevealed hint was visible, the unravelled hint will not be visible immediately when they resume the exploration. I can think of two reasons why we should not show unravelled hints immediately when the user resumes the exploration.
1. The user will need some time to get the context of the question again.
2. The checkpoint is created in the StateDeck, and the stateDeck does not know when a new unrevealed hint is visible to the user, it only has information about those hints that the user has revealed.

To restore the number of hints revealed, we will have to restore the StateFragmentViewModel.

All of these variables (of hint handler and stateFragmentViewModel) only have to be restored once when the StateFragment is created for the first time.

One way of doing this can be: 
1. we are retrieving the exploration checkpoint in the ResumeLessonActivity (new activity added in the next PR).
2. We move the checkpiont through the following path ResumeLessonActivity --> ExplorationActivity ---> ExplorationFragment --> StateFragment.
3. in the handleOnCreate method of the StateFragment, we check if saved instance state == null; this will be true iff StateFragment is created for the first time. 
4. If the condition is true, we set the values of the variable previousHelpIndex, trackedWrongAnswerCount (both in hint handler) and the variables  StateFragmntViewModel.newHintAvailableIndex and StateFragmentViewModel.allHintsExhausted with the values saved in the checkpoint. 

To fix the bug of the hint being revealed on configuration change, we can use Farees's approach in #2557. But that approach is blocked because we can't write tests for it yet.

@BenHenning WDYT? 

One more thing that I want to ask can there be any case when the saved instance state == null but the fragment is re-created? because if there is the above approach won't work.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
BenHenning: That doesn't seem correct @MaskedCarrot. The index isn't the index in the recycler view, it's the hint index within the index list. We shouldn't be conflating the two since one is a domain concept and the other is a view concept.

BenHenning:
> When an exploration is resumed the variable previousHintIndex will always be either HelpIndex.getDefaultInstance (when no hint is revealed and checkpoint has hintIndex as -1) or HelpIndex.setEverytingRevealed(true).build() (when some/all hint or solution are revealed and checkpiont has hintIndex != -1).
> 
> This means that if a user exits the exploration when an unrevealed hint was visible, the unravelled hint will not be visible immediately when they resume the exploration. I can think of two reasons why we should not show unravelled hints immediately when the user resumes the exploration.
> 
> 1. The user will need some time to get the context of the question again.
> 2. The checkpoint is created in the StateDeck, and the stateDeck does not know when a new unrevealed hint is visible to the user, it only has information about those hints that the user has revealed.
> 
> To restore the number of hints revealed, we will have to restore the StateFragmentViewModel.
> 
> All of these variables (of hint handler and stateFragmentViewModel) only have to be restored once when the StateFragment is created for the first time.
> 
> One way of doing this can be:
> 
> 1. we are retrieving the exploration checkpoint in the ResumeLessonActivity (new activity added in the next PR).
> 2. We move the checkpiont through the following path ResumeLessonActivity --> ExplorationActivity ---> ExplorationFragment --> StateFragment.
> 3. in the handleOnCreate method of the StateFragment, we check if saved instance state == null; this will be true iff StateFragment is created for the first time.
> 4. If the condition is true, we set the values of the variable previousHelpIndex, trackedWrongAnswerCount (both in hint handler) and the variables  StateFragmntViewModel.newHintAvailableIndex and StateFragmentViewModel.allHintsExhausted with the values saved in the checkpoint.
> 
> To fix the bug of the hint being revealed on configuration change, we can use Farees's approach in #2557. But that approach is blocked because we can't write tests for it yet.
> 
> @BenHenning WDYT?

Wouldn't your approach also fix the configuration change option if we saved the checkpoint in the instance state bundle? We probably should, anyway, since it's possible for the instance to be lost in cases other than configuration changes (the most notable example being out of memory cases, but we don't handle this correctly today, anyway -- see #2571 for the tracking issue).

Otherwise, this approach seems reasonable to me.

> One more thing that I want to ask can there be any case when the saved instance state == null but the fragment is re-created? because if there is the above approach won't work.

There are two potential cases that I'm aware of where this can happen/kind of happen:
1. When we handle configuration changes (but technically in this case, the fragment isn't recreated though it could look like it is depending on how we handle them)
2. When fragments are transacted with state loss

We don't do either of these, so you can use the presence of saved instance state as an indication that the fragment has been recreated.

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
```diff
@@ -75,6 +77,9 @@ class StateFragmentPresenter @Inject constructor(
   private val routeToHintsAndSolutionListener = activity as RouteToHintsAndSolutionListener
   private val hasConversationView = true
 
+  private var hasExplorationJustLoaded: Boolean = false
```
MaskedCarrot: This variable is initialized as follows  ```hasExplorationJustLoaded = savedInstanceState == null```

my original intention for having this variable was that it will only be true if the exploration has just loaded and will be false after that. so we can have something like this in processEphemeralState method

```

if (hasExplorationJustLoaded) 
    recyclerViewAssembler.restoreHintHandlerWithCheckpoint(ephemeralState.explorationCheckpoint)
```

This should have ensured that we only restore the hint handler with the checkpoint once when the exploration is loaded for the first time after it is resumed, and then the hint handler will restore again from the checkpoint only if the checkpoint is saved in the saved instance state. 

But later I found out that this will be false until the fragment is recreated because it is only then that we will reinitialize this variable.  So now I am not really sure that even if we save the checkpoint in the saved instance state how will we restore the exploration for the first time without having it restore the hint handler every time a new checkpoint is available.

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -206,15 +266,33 @@ internal class StateDeck internal constructor(
...
+      hintCount = pendingTopState.interaction.hintCount
+      indexOfLastRevealedHint = revealedHintsAndSolutionIndex
+      this.hintIndex =
+        if (isUnrevealedHintVisible) {
+          revealedHintsAndSolutionIndex + 1
+        } else {
+          revealedHintsAndSolutionIndex
+        }
+      this.isUnrevealedSolutionVisible = this@StateDeck.isUnrevealedSolutionVisible
+
```
MaskedCarrot: This method of creating checkpoints is becoming over complicated. I am thinking of having separate variables for hint and solution (2 for each)  
--  revealed hint/solution index: Int
--  unrevealed hint/solution Index: Int

WDYT?

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
```diff
@@ -322,6 +366,12 @@ class StateFragmentPresenter @Inject constructor(
 
     val ephemeralState = result.getOrThrow()
     explorationCheckpointState = ephemeralState.checkpointState
+    explorationCheckpoint = ephemeralState.explorationCheckpoint
+
+    if (hasExplorationJustLoaded) {
+      recyclerViewAssembler.restoreHintHandlerFromCheckpoint(explorationCheckpoint)
+    }
+
```
MaskedCarrot: here in the line ```explorationCheckpoint = ephemeralState.explorationCheckpoint``` I am initializing the variable `explorationCheckpoint` with the latest checkpoint, this variable will be saved in saved instance state in the override method if the configuration is changed so that it can be restored when the fragment is recreated.

MaskedCarrot (commented):
@BenHenning PTAL

MaskedCarrot:
Basic Data flow: 

`hint revelled, unrevealed hint visible, or any other interaction` ----> `new checkpoint is created and stored in the ephemeral state ` ----> `current state dataProvider notified and state is re - queried by the state Fragment` ----> `fragment uses this checkpoint to restore the values of the hint handler.`

MaskedCarrot:
There are a couple of questions that I have regarding this. 
1. If we are retrieving the checkpoint from the ephemeral state. Why do we need to save it in the saved instance bundle as well. We can restore the values of the hint handler the next time the ephemeral state is retrieved. We only need to save the hintSequenceNumber to keep track of the timer for showing when we restore the hint handler in case of configuration changes.

2. How to restore the hint handler for the first time when the exploration is loaded. If we do this in processEphemeralState, then the hint handler will be restored everytime a new ephemeral state is loaded this will be unnecessary because the hint handler already has the correct values.

MaskedCarrot:
@BenHenning can you please take a look ( mainly at State deck, exploration progress controller, state fragment presenter and StatePlayerRecylerViewAssembler)

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
BenHenning: Could we maybe leverage HelpIndex for this purpose? It was originally meant to help simplify this mechanism in the UI layer, but it seems like it could be used here as well.

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
BenHenning: If we are introducing a new ExplorationProgressController method for tracking whether the hint/solution can be shown (rather than just whether it was shown), then I don't think you need to bother with saved instance state at all since we can just rely on the EphemeralState LiveData to restore the entirety of the StateFragment state (the only thing today we don't get from EphemeralState is whether a hint can be shown which is why Farees needed to work on that PR in the first place; the controller & EphemeralState change suggested above completely mitigates the need for that fix).

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
MaskedCarrot: resolving this because changed the approach to handle hints

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
MaskedCarrot: resolving this because changed the approach to handle hints

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
MaskedCarrot: resolving this because changed the approach to handle hints

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
MaskedCarrot: resolving this because changed the approach to handle hints

MaskedCarrot:
I have added two new progress controller methods, one to notify the state deck that an un-revealed solution is visible and the other to notify the state deck that an un-revealed hint is visible. I have also created a new proto, "HintState" which acts as a container to save all the local variables of the `statePlayerRecyclerViewAssembler.hintHandler` every time the state deck is notified (when an answer is submitted or a new hint/solution is visible or reveleaed). HintState is returned back with the ephemeral state to the UI layer and all the variables of `statePlayerRecyclerViewAssembler.hintHandler` are restored. (we have to create `HintState` in the UI layer instead of computing it in the domain layer because of the variables hintSequenceNumber and trackedAnswerCount which are not part of the domain layer and are essential if we want to restore the `HintHandler` back to the previous state).

Since HintHandler is restored every time an ephemeral state is returned, therefore this solves any issue with configuration changes. In future, this approach will also allow us to show hints selectively depending upon the checkpoint by modifying the `HintState` in the domain layer before sending it with the ephemeral state to the UI layer.

MaskedCarrot:
@BenHenning I have changed the approach to the one explained in the above comment. PTAL (mainly at ExplorationProgressController, StateFragmentPresenter and StatePlayerRecyclerViewAssemebler)

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
MaskedCarrot: @BenHenning We are multiplying the viewModel.newHintIndex with 2 in line number 93 and then calling the function ` loadHintsAndSolution(state)`  so I think it should be `-2` PTAL.

BenHenning:
Apologies, will need to look at this a bit later today (my time). That being said, your approach seems sound @MaskedCarrot, I just haven't yet verified the code.

app/src/main/java/org/oppia/android/app/player/state/listener/ShowHintAvailabilityListener.kt
```diff
@@ -4,6 +4,6 @@ import org.oppia.android.app.model.HelpIndex
 
 /** Callback interface for when hints can be made available to the learner. */
 interface ShowHintAvailabilityListener {
-  /** Called when a hint is available to be shown, or null if all hints have been revealed. */
+  /** Called when a hint is available to be shown, or if hints have been revealed. */
```
BenHenning, line 7: Nice fix.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -205,10 +223,65 @@ class ExplorationProgressController @Inject constructor(
     }
   }
 
+  fun submitUnrevealedHintIsVisible(
```
BenHenning, line 226: Honestly, I'm starting to think these changes are more complicated than just moving the hint/solution logic to this controller.

I'm partly wondering if we could just move HintHandler to this class and, when a hint/solution availability changes, notify that the ephemeral state DataProvider has changed (and ensure the necessary hint/solution data is included within EphemeralState). At that point, all we need to track is which hints are revealed & whether the solution is revealed. All the extra piping to keep the UI & domain layers in sync would become much simpler.

Beyond that, we also can probably then simplify what needs to be stored to retain hint/solution state. One HintHandler is needed per session (& we should be resetting it across session states) which means we don't need to keep track of hint sequence number or tracked wrong answer count since these can be reasonably reset upon resuming a checkpoint. Ditto I think for isHintVisibleInLatestState (which would mean only HelpIndex actually needs to be stored within the checkpoint).

BenHenning (commented):
Thanks @MaskedCarrot. Left one high-level comment on specifically the hints & solutions piece, and didn't look closely at the resume bits yet since I think my comment may result in some more major changes to the piping (but hopefully an overall simplification).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: @BenHenning Moving hint handler to the domain layer will simplify the flow but it might create clashes with @yashraj-01's project. Because I think in one of the PRs he has to move the handler out of the StatePlayerRecyclerViewAssemebler as a separate class to show hints in developer options.

@BenHenning @yashraj-01 WDYT?

yashraj-01: @MaskedCarrot I'm not sure if this is a good idea. I also thought of moving the HintHandler to the domain layer but it interacts with the UI in some cases. For example, it calls the listeners while showing a hint. Also, in my PR #3613, we are putting the HintHandler modules in the FragmentComponent so, I'm not sure how it will affect your new implementation or vice-versa.

@BenHenning any thoughts on this?

BenHenning: Good point--this is an unexpected clash between the two projects. Unfortunately, the complexity we're seeing here trying to make this all work might force our hand toward trying to move it to the domain layer.

My thinking here is that we change the flow to the following: StateFragment reactively shows hints/solution entirely based on EphemeralState (mitigating the need for the listener at all--instead, we would notify that the DataProvider has changed in that same location when the class is moved to the domain layer).

The module changes should be workable: we can just use ApplicationComponent, instead, since this becomes an application-level thing instead of a fragment-level thing. I suspect it won't change too much for your project @yashraj-01, but it will definitely change some of it. I also suspect that it will probably be easier to base your work off of @MaskedCarrot's, but we probably should have a separate, preliminary PR that moves the handler to the domain layer so that it's not coupled with the rest of this resume work & so that your work isn't blocked for as long.

Would this be workable?

MaskedCarrot: I think moving the hint handler to the domain layer might get complicated. We have lifecycle aware timers that the hint handler uses to show hints. Since these timers are being observed as live data, they will have to remain in the app layer. Also, the delay for which the timer runs depends upon the current state, i.e. it is 60 seconds for the first hint, 30 for others and 10 seconds if the user submits an answer. Since the delay is variable and not fixed, we will have to compute that in the app layer. 

So I think we will have to distribute the hint handler between the app and domain layers, which might complicate the solution.

@BenHenning  WDYT?

MaskedCarrot: Actually, I was able to move the hint handler to the domain layer, and I think it does simplify the implementation of both the hint handler and the resuming mechanism. 
There are a couple of edge cases left that I still need to take care of. I will probably send a PR in a few hours or by tomorrow.

MaskedCarrot: Created a new PR (#3659) to move the hint handler to the domain module.

MaskedCarrot:
Closing this PR because now that the implementation hint handler has changed quite significantly, so there will be a lot of merge conflicts in this and it will be much easier to do everything in a new PR.

PR #3529, opened 2021-07-26
Fix #3528: Setup Infrastructure to test using local dev server

FareesHussain:
## Explanation

Fixes #3528
This pr introduces a Testing infrastructure that enables instrumentation tests for downloading explorations and loading images from the local dev server.

This introduces a TestNetworkModule, TestImageParsingModule, and a TestGcsResourceModule.
These modules are used by the OppiaTestApplication and added to a manifest used by the oppia_test android_binary.

oppia_test is similar to the oppia target but it uses the above three modules and a different AndroidManifest

Here I've used two different AndroidManifest in the instrumentation module
one under java directory used for the OppiaTestApplication
and under javatest directory used for InstrumentationTests

**Build the test app using:**
```
bazel build //instrumentation:oppia_test
```

### Screenshot
image

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3505, opened 2021-07-20
Fix #3315: Set up UIAutomator with Bazel

FareesHussain:
## Explanation

Fixes #3315

This pr introduces UI Automator with a BaseTest generated using Bazel

### How to run UI Automator tests
**Prerequisites:**
1. To be able to build oppia using bazel 
    ```
    bazel build oppia
    ```
2. Add adb to the environment (platform-tools) i.e, add the following line to the .bashrc or the file path according to the Operating System.
    ```
    export PATH=/home/<username>/Android/Sdk/platform-tools:$PATH
    ``` 
3. java version 8 (Optional, only for uiautomatorviewer)
    ```
    java -version
    ```
    *output:*
    openjdk version "1.8.0_292"
    OpenJDK Runtime Environment (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
    OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)

**Steps to run the tests:**
1. Build the BaseTest android_binary from the instrumentation module
    ```
    bazel build :oppia_test && bazel build //instrumentation:ExplorationPlayerTestBinary
    ```
2. install the oppia_test.apk and the ExplorationPlayerTestBinary.apk
    ```
    adb install -r bazel-bin/oppia_test.apk && adb install -r bazel-bin/instrumentation/ExplorationPlayerTest.apk
    ```
3. Run the instrumentation tests using am instrument command
    ```
    adb shell am instrument -w org.oppia.android.app.instrumentation/androidx.test.runner.AndroidJUnitRunner
    ```

https://user-images.githubusercontent.com/54740946/126880278-f6410f13-f26a-49e7-ab76-a3f036d03583.mp4

 

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
[...]

Issue #1611, opened 2020-08-11
Enable proper image region selection testing


BenHenning:
Image region selection currently can't be tested on Robolectric due to a bug in how Robolectric manages matrix transformations in its RenderNode class. The matrix is always identity, which results in touch events not being properly transformed according to the view's bounds & position. Changes are needed in Robolectric in order to fix this issue.

PR #3608, opened 2021-07-29
Fix #3607: ExplorationPlayer End to End tests

### LINKED ISSUE #3607: Write e2e tests for explorationplayer
Write e2e tests for exploration player using bazel + UiAutomator

FareesHussain:
## Explanation

Fixes #3607

This pr introduces e2e tests in ExplorationPlayerTests using Ui Automator

### How to run UI Automator tests
**Prerequisites:**
1. To be able to build oppia using bazel 
    ```
    bazel build oppia
    ```
2. Add adb to the environment (platform-tools) i.e, add the following line to the .bashrc or the file path according to the Operating System.
    ```
    export PATH=/home/<username>/Android/Sdk/platform-tools:$PATH
    ``` 
3. download and install [test-services-1.1.0.apk](https://mvnrepository.com/artifact/androidx.test.services/test-services/1.1.0) and [orchestrator-1.1.0](https://mvnrepository.com/artifact/androidx.test/orchestrator/1.1.0) in the emulator. (run command in the directory where both apk are downloaded)
    ```
    adb install -r test-services-1.1.0.apk && adb install -r orchestrator-1.1.0.apk
    ```
4. java version 8 (Optional, only for uiautomatorviewer)
    ```
    java -version
    ```
    *output:*
    openjdk version "1.8.0_292"
    OpenJDK Runtime Environment (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
    OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)

**Steps to run the tests:**
1. Build the BaseTest android_binary from the instrumentation module
    ```
    bazel build :oppia_test && bazel build //instrumentation:ExplorationPlayerTest
    ```
2. install the oppia_test.apk and the ExplorationPlayerTest.apk
    ```
    adb install -r bazel-bin/oppia_test.apk && adb install -r bazel-bin/instrumentation/ExplorationPlayerTest.apk
    ```
3. Run the instrumentation tests using am instrument command
```
adb shell 'CLASSPATH=$(pm path androidx.test.services) app_process / \
androidx.test.services.shellexecutor.ShellMain am instrument -w -e clearPackageData true \
-e targetInstrumentation org.oppia.android.app.instrumentation/androidx.test.runner.AndroidJUnitRunner \
androidx.test.orchestrator/.AndroidTestOrchestrator'
```

image

https://user-images.githubusercontent.com/54740946/129954451-a6ae277f-10cd-4d39-949a-a645ea46b9f1.mp4

Ran using single testcase for both with argument and without argument

https://user-images.githubusercontent.com/54740946/130116663-76873c41-248f-4954-a82f-0586c7b7f80b.mp4

Ran using duplicate testcase for both without argument and with argument

https://user-images.githubusercontent.com/54740946/130140822-60f98258-e215-4f71-968d-455322044077.mp4

Note: There are few flacky tests which is obvious for E2E tests but each test passes on individual run. 

## Checklist


anandwana001:
Assigning to @FareesHussain as well, to keep track of adding screenshots.

FareesHussain:
> Assigning to @FareesHussain as well, to keep track of adding screenshots.

Adding screenshots might take a time here as the whole test suite has 188 testcases, it might almost take more than 2 hrs

instrumentation/src/javatest/org/oppia/android/instrumentation/ExplorationPlayerTest.kt
```diff
@@ -0,0 +1,4631 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
+
+/** Tests for Explorations. */
+class ExplorationPlayerTest {
```
BenHenning: Where did these tests come from? Did you reference other test suites?

I actually think there are too many tests here. Per the testing triangle end-to-end tests are more expensive and slower to run, and thus there should be fewer of them. The advantage of these tests is they can very complex, whole-user journeys that unit tests are unable to verify. We should focus the tests we add to this suite on verifying whole user journeys rather than very specific details like whether a button is enabled.

For example, playing through an entire exploration would be a single test since it encapsulates a full user journey.

Note also that the standard 'arrange-act-assert' flow doesn't always fit for these kinds of tests. They often imply correctness simply by passing since that means the full flow worked correctly.

BenHenning (changes requested):
Thanks @FareesHussain. Left one comment, but in general I think that:
1. We need to leverage helpers to simplify the tests (per a comment I added in one of the base PRs)
2. The tests we add for UiAutomator should be higher-level and cover general user flows rather than testing at the same level of specificity of Robolectric. Please follow up with me & Akshay if you're unsure how to proceed, but I did add one example to my comment. PTAL.

instrumentation/src/javatest/org/oppia/android/instrumentation/EndToEndTestHelper.kt
```diff
@@ -0,0 +1,67 @@
...
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.uiautomator.By
+import androidx.test.uiautomator.UiDevice
+import androidx.test.uiautomator.UiScrollable
+import androidx.test.uiautomator.UiSelector
+import androidx.test.uiautomator.Until
+import org.junit.Assert.assertNotNull
+
+/** This object contains common operations used for end-to-end tests. */
+object EndToEndTestHelper {
```
anandwana001: Can we go with the other approach where we can make `EndToEndTestHelper` a class and put the function in companion object making them static to access.

anandwana001 (changes requested):
Could you add a screenshot showing all new tests are passing? 

Other than this, test LGTM.

BenHenning:
Apologies, will need to look at this a bit later today (my time).

FareesHussain:
@BenHenning can you review https://github.com/oppia/oppia-android/pull/3529 before this.
This might need some utilities to be added which would be better after 2.1 gets merged.

BenHenning:
@FareesHussain that PR isn't assigned to me--is it ready for review?

De-assigning myself here given there's a blocker. Please re-assign to me once it's ready for review.

FareesHussain:
> @FareesHussain that PR isn't assigned to me--is it ready for review?
> 
> De-assigning myself here given there's a blocker. Please re-assign to me once it's ready for review.

it is also a branch of 2.1 but not related to the tests so it can be reviewed

FareesHussain:
@BenHenning @anandwana001 
I've updated this pr with all the changes from 2.1 PR

PTAL

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+    // Assert Topic Completed.
+    scrollRecyclerViewTextIntoView("Chapter 1: Prototype Exploration")
+    val chapterCompletedTick = device.findObjectByText(
+      "Chapter 1: Prototype Exploration"
+    )?.parent?.children?.get(2)
+    assertThat(chapterCompletedTick).isNotNull()
+  }
+
+  @Test
+  fun testImageRegionSelectionInteraction_answerAllInteractionsAndCompleteExploration() {
```
FareesHussain: @BenHenning Shall I remove this test case as this requires internet and we will be using localhost for images once 2.3 is merged.

> We should be using the test lesson that you introduced into Oppia backend, but we can't since the download piece isn't finished. Given that, I think it's reasonable to file a TODO issue for enabling the image region selection test, and disabling it for now (we should still implement it, but it can be reenabled and fixed once we can download the correct content to test).

This is what you mentioned in the mail.

FareesHussain: can we also remove the test case used to check the exploration title, as it was only used to verify if the e2e tests are running in the 2.1 pr

anandwana001: Deferring to Ben for adding TODO as images will load from the local host after merging the last PR of this project.

anandwana001 (approved):
LGTM, 

1. Need to have TODO for image region test cases. 
2. PR #3505 will fix the app data clear for every test. 

Carefully merge this PR as this might get merge conflict after #3505  get merged.

FareesHussain:
Files to review

- ExplorationPlayerTest.kt
- EndToEndTestHelper.kt

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: Could we instead ``@Ignore`` this & add a TODO pointing to the issue tracking properly hooking up to the backend once downloads are supported? (If there isn't such an issue, please file one).

That way we don't lose the actual context of the test in the meantime.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+  @Test
+  fun testProtoTypeExploration_answerAllInteractionsAndCompleteExploration() {
+    navigateToPrototypeExploration()
+    completePrototypeExploration()
+
+    // Assert Topic Completed.
+    scrollRecyclerViewTextIntoView("Chapter 1: Prototype Exploration")
+    val chapterCompletedTick = device.findObjectByText(
+      "Chapter 1: Prototype Exploration"
+    )?.parent?.children?.get(2)
```
BenHenning: Why 2? Seems like a magic number.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+  @Test
+  fun testExploration_prototypeExploration_toolbarTitle_isDisplayedSuccessfully() {
+    navigateToPrototypeExploration()
+    assertThat(device.findObjectByRes("exploration_toolbar_title")).isNotNull()
+  }
+
+  @Test
+  fun testProtoTypeExploration_answerAllInteractionsAndCompleteExploration() {
+    navigateToPrototypeExploration()
+    completePrototypeExploration()
```
BenHenning: I think this is hiding a bit too much from the test. Could you instead introduce multiple helper methods (similar to StateFragmentLocalTest) for each part of the flow, and then expand the actual completion bits in the test so that context isn't lost?

BenHenning (changes requested):
Thanks @FareesHussain. Left some follow-up comments. PTAL.

instrumentation/src/javatest/org/oppia/android/instrumentation/ExplorationPlayerTest.kt
FareesHussain: removed

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: image
the `chapter_completed_tick` might be multiple if there are more than one chapters.

so I took the chapter title which is index 0 in the view group, took the parent reference to the the title, and then get its child at index 2 which makes it accurate

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: https://github.com/oppia/oppia-android/issues/3697

BenHenning:
@FareesHussain please update this with the latest develop changes now that #3505 is merged & assign back once done.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -42,7 +49,9 @@ class ExplorationPlayerTest {
     assertThat(chapterCompletedTick).isNotNull()
   }
 
+  // TODO(#3697): Update e2e tests when backend support is introduced.
   @Test
+  @Ignore("Need backend support to test the ImageRegionSelectionInteraction")
```
BenHenning:
```suggestion
  @Ignore("Need backend connection support to test the ImageRegionSelectionInteraction")
```

To be a bit more specific.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: Thanks!

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -78,60 +87,77 @@ class ExplorationPlayerTest {
     device.findObjectByText("Chapter 1: Prototype Exploration")?.click()
   }
 
-  /** Answers all the interactions and ends the exploration. */
-  private fun completePrototypeExploration() {
-    // Exploration description.
+  /** Continue Description of ProtoTypeExploration. */
+  private fun playExplorationDescription() {
```
BenHenning: Suggest: ``playContinueInteraction`` (since this is a valid interaction state).

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -78,60 +87,77 @@ class ExplorationPlayerTest {
     device.findObjectByText("Chapter 1: Prototype Exploration")?.click()
   }
 
-  /** Answers all the interactions and ends the exploration. */
-  private fun completePrototypeExploration() {
-    // Exploration description.
+  /** Continue Description of ProtoTypeExploration. */
```
BenHenning: Here & below: I actually don't think these KDocs are adding anything--I suggest removing them since they aren't required for private members, anyway, and these helpers are reasonably self-documenting in their naming.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -78,60 +87,77 @@ class ExplorationPlayerTest {
...
-    // Fraction Input Interaction.
+  /** Play FractionInputInteraction for PrototypeExporation. */
+  private fun playFractionInputInteraction() {
     device.findObjectByRes("fraction_input_interaction_view")?.text = "1/2"
     device.findObjectByText("SUBMIT")?.click()
     device.findObjectByText("CONTINUE")?.click()
-
-    // Multiple Choice Interaction - 1.
+  }
+  /** Play first MultipleChoiceInteraction for PrototypeExporation. */
```
BenHenning: Here & elsewhere: please ensure there is a single blank newline separating each method of the class.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -78,60 +87,77 @@ class ExplorationPlayerTest {
...
+  private fun playDragDropMergeInteraction() {
     device.findObjectByDesc("Link to item 2")?.click()
     device.findObjectByDesc("Move item down to 3")?.click()
     device.findObjectByText("SUBMIT")?.click()
     device.findObjectByText("CONTINUE")?.click()
+  }
 
-    // End Exploration.
+  /** End exploration for PrototypeExploration. */
+  private fun endExploration() {
```
BenHenning: Suggest: ``playEndExplorationInteraction`` (technically it's another interaction).

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -147,7 +173,20 @@ class ExplorationPlayerTest {
...
+    playFractionInputInteraction()
+    playMultipleChoiceIntearction1()
+    playMultipleChoiceIntearction2()
+    playItemSelectionInteraction()
+    playNumericInputInteraction()
+    playRatioInputInteraction()
+    playTextInputInteraction()
+    playDragAndDropInteraction()
+    playDragDropMergeInteraction()
+    endExploration()
```
BenHenning: In this case I would actually prefer you reintroduce completePrototypeExploration since these details are not important for the purpose of this test, they're just prerequisites to satisfy (unlike the other test where they are the key actions to complete).

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
```diff
@@ -67,10 +67,10 @@ class TestBazelWorkspace(private val temporaryRootFolder: TemporaryFolder) {
       createLibrary("${testName}Dependency")
     } else null to listOf()
     val buildFile = if (subpackage != null) {
-      if (!File(temporaryRootFolder.root, subpackage).exists()) {
-        temporaryRootFolder.newFolder(subpackage)
+      if (!File(temporaryRootFolder.root, subpackage.replace(".", "/")).exists()) {
```
BenHenning: Why are these changes duplicated from your other PR?

FareesHussain:
I didn't see that I pushed the TestBazelWorkspace here
Reverted now

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: Is there any way that we can retrieve the checkbox directly? It ought to at least have a content description for screenreader users if it doesn't already (/cc @rt4914). Could we add a content description & retrieve the object based on that?

I'm not keen on the '2' because it's non-specific (without this screenshot, someone coming across this isn't going to have any idea why or if this is correct), and it's a bit fragile (if we move elements in the recyclerview around, that almost guarantees breaking this test even if the UX is still more or less the same).

BenHenning (changes requested):
Thanks @FareesHussain. Took a full pass now that this is based on develop--had some comments. PTAL.

I'll also follow up over chat regarding clearing app data.

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
BenHenning: Ah, you reverted right before I submitted this comment! Marking as resolved.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -23,9 +25,52 @@ class ExplorationPlayerTest {
   }
 
   @Test
-  fun testExploration_prototypeExploration_toolbarTitle_isDisplayedSuccessfully() {
+  fun testProtoTypeExploration_answerAllInteractionsAndCompleteExploration() {
```
BenHenning: We should probably try to keep the usual test name structure even for these. What about:

testPlayExploration_prototypeExploration_playedFullyThrough_finishesSuccessfully()

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
FareesHussain: My bad I didn't see that, 
Reverted now

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: I agree, actually the content description is a bit off here, it shows %s instead of actual values.
Maybe add a todo to keep track of it of fix it in this pr

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done, I had some reference for this

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: image

app/src/main/res/layout-land/story_chapter_view.xml
```diff
@@ -91,7 +91,7 @@
             android:layout_height="wrap_content"
             android:layout_marginEnd="8dp"
             android:layout_marginBottom="4dp"
-            android:contentDescription="@{@string/chapter_completed}"
+            android:contentDescription="@{String.format(@string/chapter_completed, (viewModel.index + 1), viewModel.name)}"
```
BenHenning: Wow, nice find. I didn't realize that we weren't formatting this correctly.

@rt4914 in case something was missed somewhere or if there are other views we might want to check.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: Ah, thanks for fixing that! Can we retrieve the object now by its content description rather than using the magic number?

Also strangely, the full deltas for this PR are not showing these changes from your latest commit. Any idea why?

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -51,8 +39,8 @@ class ExplorationPlayerTest {
 
   // TODO(#3697): Update e2e tests when backend support is introduced.
   @Test
-  @Ignore("Need backend support to test the ImageRegionSelectionInteraction")
-  fun testImageRegionSelectionInteraction_answerAllInteractionsAndCompleteExploration() {
+  @Ignore("Need backend connection support to test the ImageRegionSelectionInteraction")
+  fun testPlayExploration_imageRegionInteraction_playedFullyThrough_finishesSuccessfully() {
```
BenHenning:
```suggestion
  fun testPlayExploration_imageRegionInteractionExp_playedFullyThrough_finishesSuccessfully() {
```

Just to avoid the potential confusion between testing interactions vs. full explorations.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: To be a bit more clear: this test should go through all the steps because they are key actions for the test itself. The next test can just rely on the helper since it's part of the arrangement of that test rather than the actual action.

BenHenning (commented):
Thanks @FareesHussain. Pretty close to LGTM--just had a few more comments. PTAL.

BenHenning:
Also @FareesHussain I think you have tests that may need to be updated per the app layer changes (see failing CI).

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
```diff
@@ -23,22 +25,161 @@ class ExplorationPlayerTest {
...
+  @Test
+  @Ignore("Need backend connection support to test the ImageRegionSelectionInteraction")
+  fun testPlayExploration_imageRegionInteractionExp_playedFullyThrough_finishesSuccessfully() {
+    navigateToImageRegionSelectionInteraction()
+
+    // Image Region Selection Interaction.
+    val imageSelectionView = device.findObjectByRes("interaction_container_frame_layout")
+    device.waitForRes("image_click_interaction_image_view")
+    imageSelectionView!!.children!!.get(2)!!.click()
+    device.findObjectByText("SUBMIT")!!.click()
```
FareesHussain: Here I'm using `!!.` instead of `?.`
So that in case the view doesn't exist the test fails immediately, else the test runs as normal till it reaches the assertion and then gives a failure which might be confusing

app/src/main/res/layout-land/story_chapter_view.xml
rt4914: Actually I had changed this string earlier in my PR https://github.com/oppia/oppia-android/commit/9ad3f85b4648d2bb1ab82c50ca196d0e67e4cf34

But in another it was changed back: https://github.com/oppia/oppia-android/commit/630a433adddc478b8aff20120b7325e168618b44#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103


The final solution is as follows:
1. Do not change `contentDescription` in `story_chapter_view`, i.e., keep this code as `android:contentDescription="@{@string/chapter_completed}"`
2. In `strings.xml` file, change `chapter_completed ` and `chapter_in_progress` values to `Chapter is completed` and `Chapter is in progress` respectively.

This will result in following outcome:

`Chapter 1: What is a Fraction? This is outline/summary of What is a Fraction?` , `Chapter is Completed`, `2 of 3`




The currently applied change in this PR will result in 
`Chapter 1: What is a Fraction? This is outline/summary of What is a Fraction?` , `Chapter 1 with title What is a Fraction is completed`, `2 of 3`
And we can see that there is too much repetition here which does not make sense.


Does this solve the issue?

FareesHussain: I've reverted all changes related to `chapter_completed_tick` since we can't use distinct content description 
changes requested by @rt4914 would result in having same content description ("Chapter is completed") 

and in this case we can't use the function findObjectByDesc to retrieve the object here.

I've created a separate helper to get the subview from the parent view.
i.e, UiObject2.findObjectByRes()

app/src/main/java/org/oppia/android/app/utility/ClickableAreasImage.kt
```diff
@@ -116,6 +116,7 @@ class ClickableAreasImage(
         newView.isFocusable = true
         newView.isFocusableInTouchMode = true
         newView.tag = clickableArea.label
+        newView.contentDescription = clickableArea.label
```
FareesHussain: @aggarwalpulkit596 PTAL

This enables content description for each image region in ImageRegionSelectionInteraction
image

app/src/sharedTest/java/org/oppia/android/app/testing/ImageRegionSelectionInteractionViewTest.kt
```diff
@@ -135,6 +137,26 @@ class ImageRegionSelectionInteractionViewTest {
     }
   }
 
+  @Test
+  // TODO(#1611): Fix ImageRegionSelectionInteractionViewTest
+  @Ignore
+  fun testImageRegionSelectionInteractionView_clickDescRegion3_region3Clicked() {
```
FareesHussain: Adding screenshot since the test can't run on Robolectric

image

FareesHussain: Wasn't able to exactly test the content description here

aggarwalpulkit596: What was the exact error that you. are getting here?

FareesHussain: Re: robolectric I was referring to the issue #1611

Is there a way to retrieve content description of the image region??

app/src/sharedTest/java/org/oppia/android/app/testing/ImageRegionSelectionInteractionViewTest.kt
```diff
@@ -135,6 +136,19 @@ class ImageRegionSelectionInteractionViewTest {
...
+  @Test
+  fun testImageRegionSelectionInteractionView_region3_region3HasCorrectContentDescription() {
+    launch(ImageRegionSelectionTestActivity::class.java).use {
+      it.onActivity {
+        it.findViewById<ImageRegionSelectionInteractionView>(R.id.clickable_image_view)
+          .setListener(onClickableAreaClickedListener)
+      }
+      onView(withTagValue(`is`("Region 3"))).check(
+        matches(withContentDescription("Region 3"))
+      )
```
FareesHussain: I couldn't think of any other way to test the content description for an image region

Removed the ignore function as this can be tested in robolectric

rt4914 (approved):
Approved for code owner file.

app/src/sharedTest/java/org/oppia/android/app/testing/ImageRegionSelectionInteractionViewTest.kt
aggarwalpulkit596: @FareesHussain Could you explain the issue here ?

FareesHussain: @aggarwalpulkit596 
The main requirement here is to click on an image region in ImageRegionSelectionInteraction
the only way to exactly retrieve the image region is to add a content description to each image region.
Also included a test case to check the content description

Assigned to review the changes.

aggarwalpulkit596 (approved):
LGTM

app/src/main/java/org/oppia/android/app/utility/ClickableAreasImage.kt
BenHenning: This isn't correct. The label will give away the answer, so we introduced another property for content descriptions in image region.

However, looking at https://github.com/oppia/oppia/blob/develop/extensions/interactions/ImageClickInput/ImageClickInput.py it seems we never did add this. Per https://github.com/oppia/oppia/issues/9924 I guess we didn't finish this which introduces a problem. We unfortunately won't be able to use this method.

Can you file an issue to add support for https://github.com/oppia/oppia/issues/9924 once it's ready & add that as a TODO here to fix the reference? We probably aren't going to be able to do better than the direct indexing without content description support.

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
BenHenning: I suggest returning only non-nullable types from your helper class and utilize checkNotNull with clearer error messages if an object isn't found. That'll give you the quick-failure you're looking for with clearer context, and doesn't even give the choice to test authors to potentially avoid the null-check (since we do want things to be really crisp).

app/src/main/res/layout-land/story_chapter_view.xml
BenHenning: Ack thanks @rt4914. Does this mean R/strings/chapter_completed still needs to be updated to no longer use variables?

BenHenning (changes requested):
Thanks @FareesHussain. Couple more comments, but this looks pretty close to LGTM. Once you address these I'll take on more full pass (check open conversation threads; I didn't submit the review correctly this time).

BenHenning:
Also @rt4914 PTAL at the comment I replied to above re: the content description.

app/src/main/java/org/oppia/android/app/utility/ClickableAreasImage.kt
FareesHussain: Done
https://github.com/oppia/oppia-android/issues/3712

instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt
FareesHussain: Done

app/src/main/res/layout-land/story_chapter_view.xml
BenHenning: Resolving for now to unblock the PR.

BenHenning (approved):
Thanks @FareesHussain. This LGTM!

BenHenning:
Since no reviewer threads are open & everyone's approved, merging this.

PR #3637, opened 2021-08-02
Fix #635: Prototype for Handle Multiple clicks throughout the app

### LINKED ISSUE #635: Handle Multiple clicks throughout the app
On devices that are a bit slow, pressing a button multiple times leads to the opening of multiple activities, or performing the same task in a repeated manner.
As soon as a button is clicked once, the button should be disabled from receiving further input until the task is successfully performed.

TheSwarnim:
## Explanation

Fixes part of #635
In this pr, I've added some code that handles the multiple clicks in the help activity FAQ recycler view item. In brief what we were doing when we clicked the 'FAQs' item then it will disable the item from further click so avoiding the cause of multiple clicks. After we resume the activity then it will reenable the recycle view button by overriding the onResume in HelpActivity by the help of HelpFragmentPresenter and the cycle repeats.

## Checklist

### DIFF
--- a/app/src/main/java/org/oppia/android/app/help/HelpFragmentPresenter.kt
+++ b/app/src/main/java/org/oppia/android/app/help/HelpFragmentPresenter.kt
@@ -12,6 +12,8 @@ import org.oppia.android.app.viewmodel.ViewModelProvider
 import org.oppia.android.databinding.HelpFragmentBinding
 import org.oppia.android.databinding.HelpItemBinding
 import javax.inject.Inject
+import kotlinx.android.synthetic.main.help_item.*
+import kotlinx.android.synthetic.main.help_item.view.*
 
 /** The presenter for [HelpFragment]. */
 @FragmentScope
@@ -53,5 +55,11 @@ class HelpFragmentPresenter @Inject constructor(
 
   private fun getHelpListViewModel(): HelpListViewModel {
     return viewModelProvider.getForFragment(fragment, HelpListViewModel::class.java)
+  }
+
+  fun handleOnResume(){
+    if(fragment.help_item_text_view != null){
+      fragment.help_item_text_view.isEnabled = true
+    }
   }
 }
--- a/app/src/main/java/org/oppia/android/app/help/HelpItemViewModel.kt
+++ b/app/src/main/java/org/oppia/android/app/help/HelpItemViewModel.kt
@@ -1,6 +1,11 @@
 package org.oppia.android.app.help
 
+import android.view.View
+import android.widget.AdapterView
 import androidx.appcompat.app.AppCompatActivity
+import kotlinx.android.synthetic.main.help_fragment.*
+import kotlinx.android.synthetic.main.help_item.*
+import kotlinx.android.synthetic.main.help_item.view.*
 import org.oppia.android.R
 import org.oppia.android.app.viewmodel.ObservableViewModel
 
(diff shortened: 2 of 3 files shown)


rt4914:
@BenHenning This PR tries to fix #635 . The approach here is to disable the button as soon as its clicked. Once the approach is finalized the corresponding test cases will be added and same approach will be applied on other clicks too.

Can you pleas confirm if this is a good approach?

app/src/main/java/org/oppia/android/app/help/HelpFragmentPresenter.kt
```diff
@@ -12,6 +12,8 @@ import org.oppia.android.app.viewmodel.ViewModelProvider
 import org.oppia.android.databinding.HelpFragmentBinding
 import org.oppia.android.databinding.HelpItemBinding
 import javax.inject.Inject
+import kotlinx.android.synthetic.main.help_item.*
+import kotlinx.android.synthetic.main.help_item.view.*
```
BenHenning, line 16: We shouldn't use synthetic generated code anymore--it's been deprecated in favor of databinding or Jetpack Compose (we use the former currently in Oppia Android).

app/src/main/java/org/oppia/android/app/help/HelpItemViewModel.kt
```diff
@@ -13,6 +18,7 @@ class HelpItemViewModel(
     if (title == activity.getString(R.string.frequently_asked_questions_FAQ)) {
       val routeToFAQListener = activity as RouteToFAQListListener
       routeToFAQListener.onRouteToFAQList()
+      activity.help_item_text_view.isEnabled = false
```
BenHenning, line 21: /cc @rt4914 

I have 3 main concerns with this approach:
1. The mechanism for reenabling the button is in an entire different file, and assumes a particular lifecycle order that may not be guaranteed. One example where this may break (in theory): if the button hangs due to a long operation. It will stay disabled and won't be reenabled since onResume won't be called (unless the user navigates away, backgrounds & foregrounds the app, or triggers a configuration change e.g. via a rotation). By being in a different file, it's likely people will forget to add that part which may result in broken & potentially difficult-to-debug scenarios arising.
2. We aren't tying the "enable/re-enable" state to the desired outcome. We have to assume that operations can fail if we assume they might take too long. For this reason, we may actually need to consider introducing loading interstitials that are tied directly to DataProviders/LiveDatas to help indicate when it's time to complete the transition. That approach may help us generalize the pattern, and make it more robust by introducing timeouts to ensure the button does eventually get re-enabled. This also might help guard against a potential broken scenario where the user clicks a button and, while rotating, tries clicking it again (since the current code allows this to happen).
2. As written, it's not generalizable. Ideally, we'd come up with a plug-and-play solution that can be easily adapted to multiple button scenarios with minimal effort or as the default behavior to try and cover as many scenarios where this can happen as possible.

BenHenning (changes requested):
Thanks @TheSwarnim for this initial prototype!

I left some high-level feedback. I actually suspect that #635 will require input from our product team on how the app should specifically behave in this case, and more investigation work to understand what exactly is the expected user experience on these slower devices. It may be possible that we don't actually need to do any work here if the experience isn't too bad.

In the event it is bad & needs to be addressed, this is probably closer to a 'mini-project' since there's a clear design component in figuring out a generic & robust mechanism that we can plug-and-play in all cases where this could potentially happen. Those requirements will need to be determined before we can dig into this further.

@rt4914 do you have any thoughts/things to add?

rt4914:
@BenHenning Sounds good. And I agree with possible issues in this approach. Lets close this for now and see if this issues comes from user side.

PR #3643, opened 2021-08-04
Fix #3642: UI-Driven flow for Topic Download

### LINKED ISSUE #3642: UI-Driven flow for Topic Download
1. Select a topic
2. See the topic preview page
3. Click download
4. Wait a few seconds
5. Transition to the full topic experience

anandwana001:
## Explanation

Fixes #3642 

## Checklist


app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+      .setBoolean(MY_DOWNLOADS_IS_DISABLE)
+      .build()
+    listOf<PlatformParameter>(
+      disableMyDownloadsFeature
+    )
+  }
+
+  /** Starts TopicActivity with downloaded topic */
+  fun viewDownloadedTopic(@Suppress("UNUSED_PARAMETER") v: View) {
+    platformParameterController.updatePlatformParameterDatabase(mockPlatformParameterList)
```
anandwana001: @ARJUPTA Could you help you out here, seems like we cannot update value, any quick fix here?

anandwana001: @vinitamurthi tagging for any quick suggestion

Arjupta: @anandwana001 this updatePlatformParameterDatabase will update the cache store. I think you intended to update the singleton.

anandwana001: Thanks @ARJUPTA it works

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -78,15 +82,24 @@ class TopicFragmentPresenter @Inject constructor(
...
+      )
     viewPager2.adapter = adapter
-    TabLayoutMediator(tabLayout, viewPager2) { tab, position ->
-      val topicTab = TopicTab.getTabForPosition(position, enablePracticeTab)
-      tab.text = fragment.getString(topicTab.tabLabelResId)
-      tab.icon = ContextCompat.getDrawable(activity, topicTab.tabIconResId)
-    }.attach()
-    if (!isConfigChanged && topicId.isNotEmpty()) {
-      setCurrentTab(if (storyId.isNotEmpty()) TopicTab.LESSONS else TopicTab.INFO)
+    if (!myDownloadsFeatureFlag.get().value) {
```
BenHenning: Will this break the post-download flow? The tabs should still work in that case.

app/src/main/java/org/oppia/android/app/topic/TopicViewModel.kt
```diff
@@ -20,6 +20,7 @@ class TopicViewModel @Inject constructor(
 ) : ObservableViewModel() {
   private var internalProfileId: Int = -1
   private lateinit var topicId: String
+  var enableMyDownloads: Boolean = false
```
BenHenning: Why not inject the flag, instead? Ditto for other injectable view models. For non-injectable ones, prefer passing in the flag as a val to the constructor.

This value cannot change at runtime, so it ought to be a 'val' everywhere.

app/src/main/java/org/oppia/android/app/topic/ViewPagerAdapter.kt
```diff
@@ -13,24 +13,41 @@ class ViewPagerAdapter(
...
-      }
-      TopicTab.LESSONS -> {
-        TopicLessonsFragment.newInstance(internalProfileId, topicId, storyId)
-      }
-      TopicTab.PRACTICE -> {
-        TopicPracticeFragment.newInstance(internalProfileId, topicId)
-      }
-      TopicTab.REVISION -> {
-        TopicRevisionFragment.newInstance(internalProfileId, topicId)
+    if (enableMyDownloads) {
```
BenHenning: I'm a bit confused why these tab changes are needed. Shouldn't we be navigating to the preview fragment on the outset rather than first navigating to TopicActivity? This might be a bit of a UI architecture question (/cc @rt4914 in case you have thoughts), but I'd expect the routing location (i.e. HomeActivity) to decide whether to route to the preview or full experience based on whether the topic is downloaded (to avoid complex fragment management, or transitioning subsequent activities).

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+  /** Starts TopicActivity with downloaded topic */
+  fun viewDownloadedTopic(@Suppress("UNUSED_PARAMETER") v: View) {
+    val disableMyDownloadsFeature = PlatformParameter.newBuilder()
+      .setName(MY_DOWNLOADS_FLAG)
+      .setBoolean(MY_DOWNLOADS_IS_DISABLE)
+      .build()
+
+    val platformParameterMap = mutableMapOf<String, PlatformParameter>()
+    platformParameterMap.put(disableMyDownloadsFeature.name, disableMyDownloadsFeature)
+    platformParameterSingleton.setPlatformParameterMap(platformParameterMap)
```
BenHenning: This we **definitely** do not want to ever do. Changing state of platform parameters mid-app could cause really bad behaviors since the system is meant to guarantee immutability. We should be using a different code design to manage the state instead of changing the gate flag like this.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
```diff
@@ -16,3 +16,10 @@ annotation class SplashScreenWelcomeMsg
 val SPLASH_SCREEN_WELCOME_MSG = "splash_screen_welcome_msg"
 val SPLASH_SCREEN_WELCOME_MSG_DEFAULT_VALUE = false
 val SPLASH_SCREEN_WELCOME_MSG_VALUE = true
+
+@Qualifier
+annotation class MyDownloads
+
+val MY_DOWNLOADS_FLAG = "my_downloads_flag"
+val MY_DOWNLOADS_IS_ENABLED_DEFAULT = true
+val MY_DOWNLOADS_IS_DISABLE = false
```
BenHenning: This shouldn't be needed, I think. The default is sufficient since it defines the compile-time on/off status for the flag.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
```diff
@@ -16,3 +16,10 @@ annotation class SplashScreenWelcomeMsg
 val SPLASH_SCREEN_WELCOME_MSG = "splash_screen_welcome_msg"
 val SPLASH_SCREEN_WELCOME_MSG_DEFAULT_VALUE = false
 val SPLASH_SCREEN_WELCOME_MSG_VALUE = true
+
+@Qualifier
+annotation class MyDownloads
+
+val MY_DOWNLOADS_FLAG = "my_downloads_flag"
+val MY_DOWNLOADS_IS_ENABLED_DEFAULT = true
```
BenHenning, line 24: This should be default false--we don't want to change the developer experience since this workaround is really impactful (& will probably break tests).

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -25,4 +28,13 @@ class PlatformParameterModule {
     return platformParameterSingleton.getBooleanPlatformParameter(SPLASH_SCREEN_WELCOME_MSG)
       ?: PlatformParameterValue.createDefaultParameter(SPLASH_SCREEN_WELCOME_MSG_DEFAULT_VALUE)
   }
+
+  @Provides
+  @MyDownloads
+  fun provideMYDownloadsParam(
```
BenHenning:
```suggestion
  fun provideMuDownloadsParam(
```

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -25,4 +28,13 @@ class PlatformParameterModule {
...
+    platformParameterSingleton: PlatformParameterSingleton
+  ): PlatformParameterValue<Boolean> {
+    return platformParameterSingleton.getBooleanPlatformParameter(MY_DOWNLOADS_FLAG)
+      ?: PlatformParameterValue.createDefaultParameter(MY_DOWNLOADS_IS_ENABLED_DEFAULT)
```
BenHenning, line 38: @ARJUPTA should this just be creating the default parameter since we have no intention to control this flag (yet) with Oppia's backend?

app/src/sharedTest/java/org/oppia/android/app/topic/info/TopicInfoFragmentTest.kt
```diff
@@ -354,6 +470,82 @@ class TopicInfoFragmentTest {
...
+    launchTopicActivityIntent(internalProfileId, RATIOS_TOPIC_ID).use {
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.download_story_count_text_view))
+        .check(matches(withText(R.string.topic_info_topic_downloaded)))
+    }
+  }
+
+  @Test
+  fun testTopicInfoFragment_enableMyDownloads_skillsIsDisplayed() {
+    updateMyDownloadsFeatureFlag(MY_DOWNLOADS_IS_ENABLED_DEFAULT)
```
BenHenning: I suggest defining test-specific enabled/disabled constants for this purpose.

app/src/main/res/values/strings.xml
```diff
@@ -110,6 +113,21 @@
   <string name="status_completed">Completed</string>
   <string name="show_chapter_list">Show chapter list</string>
   <string name="hide_chapter_list">Hide chapter list</string>
+  <string name="topic_info_skills_heading">Skills</string>
+  <string name="topic_info_download_to_play">DOWNLOAD TO PLAY</string>
+  <string name="topic_info_topic_downloaded">Downlaoded</string>
+  <string name="topic_info_download_topic">Downlaod Topic</string>
```
BenHenning: Typos in strings.

app/src/main/res/values/strings.xml
```diff
@@ -110,6 +113,21 @@
...
+  <string name="topic_info_downloaded_message">Your download for %s is complete!</string>
+  <string name="show_hide_chapter_list">Show/Hide Chapter list of %s</string>
+  <plurals name="story_play_heading">
```
BenHenning: This should be contextualized to the view it's displayed on (otherwise the name is a bit generic).

BenHenning (changes requested):
Thanks @anandwana001. Left some high-level feedback.

I think my main take away is that this seems far more invasive than I expected. Why not utilize a separate activity/fragment for the preview flow rather than adding all this complex fragment management bits? I recall discussing the high-level UI architecture back when you were working on the design document, but I thought that we were fine with starting with an activity even though the transition isn't as nice since it's far simpler to build & maintain. Please let me know your thoughts on this.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
anandwana001: fixed

app/src/main/java/org/oppia/android/app/topic/TopicViewModel.kt
anandwana001: fixed

app/src/main/java/org/oppia/android/app/topic/ViewPagerAdapter.kt
anandwana001: fixed

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedViewModel.kt
anandwana001: fixed

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
anandwana001: fixed

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
anandwana001: fixed

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
anandwana001: fixed

app/src/main/res/values/strings.xml
anandwana001: fixed

app/src/main/res/values/strings.xml
anandwana001: fixed

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
Arjupta: Yes. Any flag which is not defined in the backend will be using default value always. But keeping the provider method in this format gives us the flexibility to introduce that flag in the backend at any point of time

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import org.oppia.android.R
+import org.oppia.android.app.activity.ActivityScope
+import javax.inject.Inject
+
+/** Profile id argument key helps in passing to TopicPreviewFragment. */
+const val PROFILE_ID_ARGUMENT_KEY = "profile_id"
+/** Topic id argument key help in passing to TopicPreviewFragment. */
```
rt4914, line 11: nit: `helps`

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
```diff
@@ -0,0 +1,34 @@
...
+/** The presenter for [TopicPreviewActivity]. */
+@ActivityScope
+class TopicPreviewActivityPresenter @Inject constructor(
+  private val activity: AppCompatActivity
+) {
+
+  /** Shows the TopicPreviewFragment */
+  fun handleOnCreate(internalProfileId: Int, topicId: String) {
+    activity.setContentView(R.layout.topic_preview_activity)
+    // Bundle object to pass to TopicPreviewFragment.
```
rt4914, line 23: I thinks this can be removed as its self-explanatory.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
```diff
@@ -0,0 +1,34 @@
...
+    val args = Bundle()
+    args.putInt(PROFILE_ID_ARGUMENT_KEY, internalProfileId)
+    args.putString(TOPIC_ID_ARGUMENT_KEY, topicId)
+    val topicPreviewFragment = TopicPreviewFragment()
```
rt4914, line 27: Also, suggest not creating fragment like this. Instead create a companion functions in `TopicPreviewFragment` named `newInstance` and call that function directly. This follows the same concept as that of `createXYZActivityIntent`.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewFragmentPresenter.kt
```diff
@@ -0,0 +1,214 @@
...
+  private val oppiaLogger: OppiaLogger
+) {
+
+  private var internalProfileId: Int = -1
+  private lateinit var topicId: String
+
+  private lateinit var binding: TopicPreviewFragmentBinding
+  private val topicPreviewViewModel = getTopicPreviewViewModel()
+
+  /** Bind TopicDownloadedFragmentBinding with the TopicDownloadedFragment */
```
rt4914, line 46: `.` at the end

rt4914 (changes requested):
From overview it looks good. Some major changes are:
1. Need screenshots to confirm if its RTL compatible.
2. Need video for A11y output.
3. I think the `preview` and `topicdownloaded` package should be inside `topic`. WDYT? Also it is suppose remain outside then `topicpreview` would be much more appropriate.

anandwana001:
@rt4914 Keeping `preview` and `topicdownload` outside because 
1. We can access `preview` from other activities like `MyDownloads` once the `updates` get in.
2. We can access `topicdownload` from screens like `MyDownloads` as well once it get in.
These are the reason, I am keeping both of these packages independent of Topic. 

Regarding the RTL and a11y, I am keeping these off from this PR, as this is a Quick PR just to bring the UI-driven flow of topic download. Currently, no bandwidth for RTL and a11y work.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
anandwana001: Can't remove, as every non-private must have kdoc.
Though, I need to fix this using `/** */`.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
anandwana001: ah, are we following this starter pattern for Fragment also? 
Till now what I saw is that we follow this pattern only for Activity.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewFragmentPresenter.kt
anandwana001: nice catch. Thanks.

anandwana001:
@BenHenning Requesting for review. 

Regarding the Bazel failure, `TopicPreviewViewModel` has `import org.oppia.android.R` and I added this file in `VIEW_MODELS_WITH_RESOURCE_IMPORTS` in `app/BUILD` but the build is failing. Any specific reason or something I am missing here?

BenHenning:
Hi. Will need to look at this tomorrow.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -144,7 +150,10 @@ class HomeFragmentPresenter @Inject constructor(
   }
 
   fun onTopicSummaryClicked(topicSummary: TopicSummary) {
-    routeToTopicListener.routeToTopic(internalProfileId, topicSummary.topicId)
+    if (myDownloadsFeatureFlag.get().value)
+      routeToTopicPreviewListener.routeToTopicPreview(internalProfileId, topicSummary.topicId)
+    else
+      routeToTopicListener.routeToTopic(internalProfileId, topicSummary.topicId)
```
BenHenning, line 156: Missing curly braces per https://developer.android.com/kotlin/style-guide#braces.

app/src/main/java/org/oppia/android/app/home/RouteToTopicPreviewListener.kt
```diff
@@ -0,0 +1,7 @@
+package org.oppia.android.app.home
+
+/** Listener for when an activity should route to a topic preview. */
+interface RouteToTopicPreviewListener {
+  /** Helps to start TopicPreviewActivity. */
```
BenHenning, line 5:
```suggestion
  /** Called when a user wants to open a topic preview. */
```

These KDocs should specifically explain things fromt he user (calling) perspective, not the implementation perspecitve (since technically there could be multiple implementations of this interface).

app/src/main/java/org/oppia/android/app/home/promotedlist/PromotedStoryViewModel.kt
```diff
@@ -37,11 +40,15 @@ class PromotedStoryViewModel(
...
 
   fun clickOnStoryTile() {
-    routeToTopicPlayStoryListener.routeToTopicPlayStory(
-      internalProfileId,
-      promotedStory.topicId,
-      promotedStory.storyId
-    )
+    if (myDownloadsFeatureFlagValue)
+      routeToTopicPreviewListener.routeToTopicPreview(internalProfileId, promotedStory.topicId)
+    else {
```
BenHenning, line 45: Ditto here (for missing braces).

app/src/main/java/org/oppia/android/app/completedstorylist/CompletedStoryListViewModel.kt
```diff
@@ -22,7 +25,8 @@ class CompletedStoryListViewModel @Inject constructor(
   private val intentFactoryShim: IntentFactoryShim,
   private val topicController: TopicController,
   private val oppiaLogger: OppiaLogger,
-  @StoryHtmlParserEntityType private val entityType: String
+  @StoryHtmlParserEntityType private val entityType: String,
+  @MyDownloads private val myDownloadsFeatureFlag: Provider<PlatformParameterValue<Boolean>>
```
BenHenning, line 29: Is this being used or is it needed in this view model?

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -144,7 +150,10 @@ class HomeFragmentPresenter @Inject constructor(
   }
 
   fun onTopicSummaryClicked(topicSummary: TopicSummary) {
-    routeToTopicListener.routeToTopic(internalProfileId, topicSummary.topicId)
+    if (myDownloadsFeatureFlag.get().value)
+      routeToTopicPreviewListener.routeToTopicPreview(internalProfileId, topicSummary.topicId)
+    else
```
BenHenning, line 155: Could you please add a test to verify that this flow properly opens the preview activity when the activity is enabled? We should also make sure there's already or is now a test for the opposite case (i.e. that the preview activity does not open).

app/src/main/java/org/oppia/android/app/home/promotedlist/PromotedStoryViewModel.kt
```diff
@@ -37,11 +40,15 @@ class PromotedStoryViewModel(
   }
 
   fun clickOnStoryTile() {
-    routeToTopicPlayStoryListener.routeToTopicPlayStory(
-      internalProfileId,
-      promotedStory.topicId,
-      promotedStory.storyId
-    )
+    if (myDownloadsFeatureFlagValue)
+      routeToTopicPreviewListener.routeToTopicPreview(internalProfileId, promotedStory.topicId)
```
BenHenning, line 44: Ditto here (please make sure there are tests covering this & its opposite cases).

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.os.Bundle
+import org.oppia.android.app.activity.InjectableAppCompatActivity
+import javax.inject.Inject
+
+/** TopicPreviewFragment TAG. */
+const val TOPIC_PREVIEW_FRAGMENT_TAG = "TopicPreviewFragment"
+private const val TOPIC_PREVIEW_ACTIVITY_TOPIC_ID_ARGUMENT_KEY = "TopicPreviewActivity.topic_id"
+private const val TOPIC_PREVIEW_ACTIVITY_PROFILE_ID_ARGUMENT_KEY = "TopicPreviewActivity.profile_id"
+
+/** Activity display preview of topic when topic is not downloaded. */
```
BenHenning, line 14:
```suggestion
/** Provides a preview of a topic for when it's not yet downloaded. */
```

Suggestion to make this a bit clearer.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+class TopicPreviewActivity : InjectableAppCompatActivity() {
+
+  companion object {
+    /**
+     * This function creates the intent of [TopicPreviewActivity].
+     *
+     * @param [internalProfileId]: Id of the Profile.
+     * @param [topicId]: Id of the Topic.
+     *
```
BenHenning, line 23: Nit: no blankline between at-param clauses in KDocs.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+private const val TOPIC_PREVIEW_ACTIVITY_PROFILE_ID_ARGUMENT_KEY = "TopicPreviewActivity.profile_id"
+
+/** Activity display preview of topic when topic is not downloaded. */
+class TopicPreviewActivity : InjectableAppCompatActivity() {
+
+  companion object {
+    /**
+     * This function creates the intent of [TopicPreviewActivity].
+     *
+     * @param [internalProfileId]: Id of the Profile.
```
BenHenning, line 21:
```suggestion
     * @param internalProfileId Id of the profile
```

Note:
- The brackets aren't needed here since it's already expected that the name will match a parameter in the parameter list
- The colon should not be here
- Most words should not be capitalized unless they are proper nouns ('profile' is not a proper noun in this context)
- At-clauses don't end in punctuation unless they are multiple sentences

Ditto elsewhere in this PR & future PRs.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+     * @param [topicId]: Id of the Topic.
+     *
+     * @return [Intent]: TopicPreviewActivity Intent object.
```
BenHenning, line 24:
```suggestion
     * @return an [Intent] to open the [TopicPreviewActivity]
```

We do not explicitly specify the type in the return at-clause (it's not needed since KDoc generation will pull the type from the method declaration itself).

Ditto elsewhere.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewActivityPresenter.kt
BenHenning: Not sure I understand--this isn't a KDoc?

app/src/main/java/org/oppia/android/app/preview/TopicPreviewChapterItemViewModel.kt
```diff
@@ -0,0 +1,9 @@
+package org.oppia.android.app.preview
+
+import org.oppia.android.app.viewmodel.ObservableViewModel
+
+/** [ObservableViewModel] for the chapter recycler view of Topic Preview. */
```
BenHenning, line 5: Here here also on capitalization--please fix (& in other KDocs in the PR).

app/src/main/res/layout/topic_preview_chapter_list_item.xml
```diff
@@ -0,0 +1,28 @@
...
+  xmlns:tools="http://schemas.android.com/tools">
+
+  <data>
+
+    <variable
+      name="viewModel"
+      type="org.oppia.android.app.preview.TopicPreviewChapterItemViewModel" />
+  </data>
+
+  <LinearLayout
```
BenHenning, line 12: There's only one view in this layout--why is a linear layout needed?

app/src/main/res/layout/topic_preview_chapter_list_item.xml
```diff
@@ -0,0 +1,28 @@
...
+    android:layout_height="wrap_content"
+    android:layout_marginStart="16dp"
+    android:layout_marginBottom="16dp">
+
+    <TextView
+      android:id="@+id/chapter_name"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:fontFamily="sans-serif"
+      android:text="@{String.format(@string/topic_chapter_index, (viewModel.index + 1), viewModel.chapterName)}"
```
BenHenning, line 23:
```suggestion
      android:text="@{@string/topic_chapter_index((viewModel.index + 1), viewModel.chapterName)}"
```

Data-binding can call format() directly.

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+
+    <com.google.android.material.appbar.AppBarLayout
+      android:id="@+id/topic_preview_app_bar_layout"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar"
+      app:layout_constraintBottom_toTopOf="@+id/topic_tabs_viewpager_container"
+      app:layout_constraintEnd_toEndOf="parent"
+      app:layout_constraintStart_toStartOf="parent"
+      app:layout_constraintTop_toTopOf="parent">
```
BenHenning, line 28: The outer layout isn't a constraint layout--remove.

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+          android:fadingEdge="horizontal"
+          android:fadingEdgeLength="20dp"
+          android:focusable="false"
+          android:focusableInTouchMode="true"
+          android:fontFamily="sans-serif"
+          android:marqueeRepeatLimit="1"
+          android:requiresFadingEdge="horizontal"
+          android:scrollHorizontally="true"
+          android:singleLine="true"
+          android:text="@{String.format(@string/topic_preview_toolbar_title, viewModel.topic.name)}"
```
BenHenning, line 55: Ditto here for format (see my other comment).

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+      </androidx.appcompat.widget.Toolbar>
+    </com.google.android.material.appbar.AppBarLayout>
+
+    <FrameLayout
+      android:layout_width="match_parent"
+      android:layout_height="match_parent">
+
+      <ScrollView
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
```
BenHenning, line 67: Is it correct to use match_parent here when the parent is also match_parent & the previous element isn't 0? I think that means we're trying to make the outer linear layout larger than it can be. Or, does linear layout automatically fix this layout height for us?

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+          android:paddingBottom="24dp">
+
+          <org.oppia.android.app.customview.LessonThumbnailImageView
+            android:id="@+id/topic_preview_thumbnail_image_view"
+            android:layout_width="match_parent"
+            android:layout_height="0dp"
+            android:adjustViewBounds="true"
+            app:entityId="@{viewModel.topic.topicId}"
+            app:entityType="@{viewModel.entityType}"
+            app:layout_constraintDimensionRatio="16:9"
```
BenHenning, line 86: Is this always true? Is this what we use for other thumbnails elsewhere in the app?

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@id/topic_preview_download_image_view"
+            app:layout_constraintTop_toTopOf="parent"
+            app:lessonThumbnail="@{viewModel.topic.topicThumbnail}"
+            tools:src="@tools:sample/avatars" />
+
+          <FrameLayout
+            android:id="@+id/thumbnail_gradient_container"
```
BenHenning, line 95: ID should be namespaced based on being part of the topic preview, too, like others here since IDs are global.

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+            android:gravity="start"
+            android:maxLines="@{viewModel.isDescriptionExpanded() ? @integer/topic_description_expanded : @integer/topic_description_collapsed}"
+            android:text="@{viewModel.topicDescription}"
+            android:textColor="@color/oppiaPrimaryText"
+            android:textDirection="locale"
+            android:textSize="16sp"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/topic_preview_story_count_text_view"
+            tools:text="@android:string/httpErrorBadUrl" />
```
BenHenning, line 156: Suggest removing this since it seems unrelated to the content.

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+            android:paddingBottom="16dp"
+            android:scrollbars="none"
+            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="@+id/topic_preview_download_image_view"
+            app:layout_constraintTop_toBottomOf="@+id/stories_heading"
+            app:list="@{viewModel.storyItemList}" />
+
+          <TextView
+            android:id="@+id/skills_heading"
```
BenHenning, line 251: Aren't these actually subtopics, not skills? I think we just refer to them as 'skills' in the UI to simplify things for users, but technically they're subtopics.

app/src/main/res/layout-sw600dp-port/topic_preview_fragment.xml
```diff
@@ -0,0 +1,277 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BenHenning, line 1: Is it possible to consolidate this & the other topic preview fragments into a single layout file? We have an active effort to do that with many of our other layouts, so it seems like duplicating this new one is the wrong direction since it just pushes back that work.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+  fragment: Fragment,
+  @TopicHtmlParserEntityType val entityType: String
+) : ObservableViewModel() {
+
+  private val topicPreviewListener = fragment as TopicPreviewListener
+
+  /** Topic observable instance. */
+  val topic = ObservableField<Topic>(Topic.getDefaultInstance())
+  /** Size of the topic. */
+  val topicSize = ObservableField<String>("")
```
BenHenning, line 29: Needs a unit--should it be bytes?

Also, the size is presumably an Int, not a String.

utility/src/main/java/org/oppia/android/util/formattor/FileSizeConversionUtil.kt
```diff
@@ -0,0 +1,35 @@
+package org.oppia.android.util.formattor
+
+import android.content.Context
+import org.oppia.android.util.R
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Utility to calculate file size. */
+@Singleton
+class FileSizeConversionUtil @Inject constructor(
```
BenHenning, line 10: I don't think this will work correctly. Using R requires special regeneration in Bazel at the moment for proper interoperability with Gradle. I think this conversion needs to stay in the app layer.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
BenHenning: Still seems true. Reminder to make this false. (I'll probably keep this comment open until the end to make sure it stays false up to the final check-in of the code).

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
```diff
@@ -16,3 +16,9 @@ annotation class SplashScreenWelcomeMsg
 val SPLASH_SCREEN_WELCOME_MSG = "splash_screen_welcome_msg"
 val SPLASH_SCREEN_WELCOME_MSG_DEFAULT_VALUE = false
 val SPLASH_SCREEN_WELCOME_MSG_VALUE = true
+
+@Qualifier
+annotation class MyDownloads
```
BenHenning, line 21: Please add a KDoc to explain what feature this corresponds to.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
BenHenning: Thanks for clarifying.

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+    fragmentComponent.inject(this)
+  }
+
+  override fun onCreateView(
+    inflater: LayoutInflater,
+    container: ViewGroup?,
+    savedInstanceState: Bundle?
+  ): View? {
+    val internalProfileId = arguments?.getInt(PROFILE_ID_ARGUMENT_KEY, -1)!!
+    val topicId = arguments?.getString(TOPIC_ID_ARGUMENT_KEY)!!
```
BenHenning, line 28: Prefer adding checkNotNull(){} calls here with more contextual errors in case a null is actually encountered (see other checkNotNull cases in onCreateView() calls for reference).

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  private val routeToTopicListener = activity as RouteToTopicListener
+
+  /** Name of the downloaded topic */
+  val topicName = ObservableField<String>("")
+
+  /** Id of the profile which downloaded a topic */
+  var internalProfileId: Int = -1
+
+  /** Id of the downloaded topic */
```
BenHenning, line 26:
```suggestion
  /** ID of the downloaded topic. */
```

Summary fragments do end in punctuation. Ditto elsewhere in this file & PR.

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedFragmentPresenter.kt
```diff
@@ -0,0 +1,91 @@
...
+    topicLiveData.observe(
+      fragment,
+      Observer { result ->
+        topicDownloadedViewModel.topicName.set(result.name)
+      }
+    )
+  }
+
+  private fun getTopicDownloadedViewModel(): TopicDownloadedViewModel {
+    return viewModelProvider.getForFragment(fragment, TopicDownloadedViewModel::class.java)
```
BenHenning, line 89: We should not be using view model provider. Instead, mark the view model as FragmentScoped and inject it. ViewModelProvider will leak view models that are also bound to Dagger.

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedFragmentPresenter.kt
```diff
@@ -0,0 +1,91 @@
...
+        "TopicDownloadedFragment",
+        "Failed to retrieve topic",
+        topic.getErrorOrNull()!!
+      )
+    }
+    return topic.getOrDefault(Topic.getDefaultInstance())
+  }
+
+  private val topicResultLiveData: LiveData<AsyncResult<Topic>> by lazy {
+    topicController.getTopic(profileId, topicId = topicId).toLiveData()
```
BenHenning, line 76:
```suggestion
    topicController.getTopic(profileId, topicId).toLiveData()
```

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedFragmentPresenter.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+    this.topicId = topicId
+    profileId = ProfileId.newBuilder().setInternalId(internalProfileId).build()
+    subscribeToTopicLiveData()
+    topicDownloadedViewModel.internalProfileId = internalProfileId
+    topicDownloadedViewModel.topicId = topicId
+    return binding.root
+  }
+
+  private val topicLiveData: LiveData<Topic> by lazy { getTopic() }
```
BenHenning, line 58: Nit: move all vals above the methods & keep them together for consistency.

Ditto elsewhere in the PR.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewListener.kt
```diff
@@ -0,0 +1,7 @@
+package org.oppia.android.app.preview
+
+/** Listener interface for topic preview screen. */
+interface TopicPreviewListener {
+  /** Start topic download process */
```
BenHenning, line 5: Ditto here (please update similarly to my earlier comment).

app/src/main/java/org/oppia/android/app/preview/TopicPreviewListener.kt
```diff
@@ -0,0 +1,7 @@
+package org.oppia.android.app.preview
+
+/** Listener interface for topic preview screen. */
+interface TopicPreviewListener {
```
BenHenning, line 4: This is a bit too vague. What exactly is being listened for, opening the topic preview page? Please be more specific in class name & documentation.

app/src/main/java/org/oppia/android/app/preview/TopicPreviewFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+    fragmentComponent.inject(this)
+  }
+
+  override fun onCreateView(
+    inflater: LayoutInflater,
+    container: ViewGroup?,
+    savedInstanceState: Bundle?
+  ): View? {
+    val internalProfileId = arguments?.getInt(PROFILE_ID_ARGUMENT_KEY)!!
+    val topicId = arguments?.getString(TOPIC_ID_ARGUMENT_KEY)!!
```
BenHenning, line 28: Ditto here & elsewhere (please add checkNotNull checks).

app/src/main/java/org/oppia/android/app/topicdownloaded/TopicDownloadedActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.oppia.android.app.home.RouteToTopicListener
+import org.oppia.android.app.topic.TopicActivity
+import javax.inject.Inject
+
+private const val TOPIC_DOWNLOADED_ACTIVITY_TOPIC_ID_ARGUMENT_KEY =
+  "TopicDownloadedActivity.topic_id"
+private const val TOPIC_DOWNLOADED_ACTIVITY_INTERNAL_PROFILE_ID_ARGUMENT_KEY =
+  "TopicDownloadedActivity.internal_profile_id"
+
+/** The activity for displaying [TopicDownloadedFragment]. */
```
BenHenning, line 16: The fragment is an implementation detail. Similar to my earlier activity KDoc comment, please rephrase this to explain what it's meant to show.

app/src/main/res/values/strings.xml
```diff
@@ -89,6 +90,9 @@
   <string name="number_with_units_input_hint_text">Write numbers with units here.</string>
   <string name="enable_audio_playback_content_description">Enable audio voiceover for this lesson.</string>
   <string name="recently_played_stories">Recently-Played Stories</string>
+  <string name="topic_download_page">Topic Download page</string>
+  <string name="topic_preview_page">Topic Preview page</string>
```
BenHenning, line 94: Do the mocks include the word 'page'?

app/src/main/res/values/strings.xml
```diff
@@ -110,6 +114,19 @@
...
+  <string name="show_hide_chapter_list">Show/Hide Chapter list of %s</string>
+  <plurals name="topic_preview_story_heading">
+    <item quantity="one">
+      Story You Can Play
+    </item>
+    <item quantity="other">
+      Stories You Can Play
+    </item>
+  </plurals>
+  <string name="topic_download_view">View</string>
```
BenHenning, line 129: If this is temporary, should it be marked as not translatable?

app/src/main/res/layout/topic_preview_fragment.xml
```diff
@@ -0,0 +1,288 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="?attr/actionBarSize"
+        android:background="?attr/colorPrimary"
+        android:fontFamily="sans-serif"
+        android:minHeight="?attr/actionBarSize"
+        app:navigationContentDescription="@string/navigate_up"
+        app:navigationIcon="?attr/homeAsUpIndicator">
+
+        <TextView
+          android:id="@+id/topic_preview_toolbar_title"
```
BenHenning, line 41: This seems to be duplicated with the string name--can we use a unique name for the view to differentiate?

app/src/main/AndroidManifest.xml
```diff
@@ -69,6 +69,13 @@
       android:name=".app.ongoingtopiclist.OngoingTopicListActivity"
       android:label="@string/ongoing_topic_list_activity_title"
       android:theme="@style/OppiaThemeWithoutActionBar" />
+    <activity
+      android:name=".app.topicdownloaded.TopicDownloadedActivity"
+      android:label="@string/topic_download_page"
```
BenHenning, line 74: Should this be topic_downloaded_page?

BenHenning (changes requested):
Thanks @anandwana001. Took a partial pass on the PR. However, there are some things that I need to be addressed before I can finish taking a full pass. In particular:
- Please fix the CI checks (some of the failures are related to scripts which I think may indicate missing test and/or KDoc coverage)
- Please make sure all new classes have appropriate tests (even placeholders)
- Please address the pending conflicts reported below
- Please add screenshots for all the new screens being introduced or changed here, along with corresponding configurations (portrait/landscape, handheld/tablet) to make it easier to contextualize the UI changes & verify mock correctness
- Please add a video demonstrating the new implementation to make the cross-activity flows clearer

> @BenHenning Requesting for review.
> 
> Regarding the Bazel failure, `TopicPreviewViewModel` has `import org.oppia.android.R` and I added this file in `VIEW_MODELS_WITH_RESOURCE_IMPORTS` in `app/BUILD` but the build is failing. Any specific reason or something I am missing here?

One of your listeners may be missing from the listener list in Bazel per the error. Further, there is probably some wiring difficulty in getting the util component working correctly so I suggest instead not going that route (see my comment on the utility for an alternate suggestion).

anandwana001:
Keeping this in low priority right now compared to the proto size calculation project and updating the email thread for the milestone change for the End-to-End testing project.

anandwana001:
Will take a look at this next week.

anandwana001:
closing this PR for now, as I am not working on this for the next couple of weeks. I Will check again later this month.

Issue #1273, opened 2020-06-08
Introduce UI tests for hints & solutions


BenHenning:
Hints & solutions logic was updated in #500, but an unanticipated regression delayed its merge by a week.
This regression could have been caught earlier if robust UI tests were in place. Currently, there is no comprehensive test coverage for hint & solution flows in the app, which risks future regressions going undetected.

## Solution Approach

Introduce **dedicated UI tests** that:

Introduce **dedicated UI tests** that:

1. **Verify hint bulb visibility depending on user actions**
   * `testStateFragment_forHintsAndSolution_incorrectInput_hintBulbContainerIsNotVisible`
   * `testStateFragment_forHintsAndSolution_incorrectInputTwice_hintBulbContainerIsVisible`
   * `testStateFragment_forHintsAndSolution_incorrectInput_thenCorrectInput_doesNotShowHint`
   * `testStateFragment_withNoHints_doesNotShowHintBulb`

2. **Ensure correct accessibility labels for hints & solutions**
   * `testStateFragment_showHintsAndSolutionBulb_dotHasCorrectContentDescription`
   * `testStateFragment_showHintsAndSolutionBulb_bulbHasCorrectContentDescription`

3. **Validate the lifecycle of hints and solutions (timing, showing, dismissal)**
   * `testStateFragment_hintShown_afterDelayAndSubsequentHintChanges_doNotRedrawActiveAnswer`
   * `testStateFragment_revealSolution_afterHints_showsSolutionAndDisablesFurtherHints`
   * `testStateFragment_repeatedHints_doNotStackOrDuplicate`
   * `testStateFragment_hint_withNullText_showsFallbackOrNoCrash`

4. **Test navigation scenarios (moving between questions/explorations)**
   * `testQuestionPlayer_hintFlow_acrossQuestionNavigation_restoresCorrectState`

5. **Cover edge cases like no hints, null hints, or duplicate updates**
   * (Already covered by: `testStateFragment_withNoHints_doesNotShowHintBulb`,  
     `testStateFragment_hint_withNullText_showsFallbackOrNoCrash`,  
     `testStateFragment_repeatedHints_doNotStackOrDuplicate`)

6. **Add sanity Espresso flows to ensure correctness on device-like conditions**
   * `testHintsSolution_basicFlow_espresso_endToEnd`

[...]

PR #3659, opened 2021-08-08
Fix #2476, #3312 and #3703: Move hint handler to domain layer

### LINKED ISSUE #2476: Revealing second hint after orientation change before the timer

### LINKED ISSUE #3312: Take out HintHandler from StatePlayerRecyclerViewAssembler
Take out HintHandler from StatePlayerRecyclerViewAssembler and use the interface approach to use it as a prod implementation of Hint Handler.

**Target PR date:** 2021-07-28
**Target completion date:** 2021-08-04

### LINKED ISSUE #3703: Non-modularized testing module tests not being built in Bazel today
Found this out in passing.

``bazel query 'kind(test,//testing/...)'`` only returns the tests that have been migrated rather than all. From some digging this seems to be because having a BUILD.bazel file (even a blank one) under src/test changes the directory graph such that the glob can no longer find files under that directory. This is actually really unexpected, and may pose an issue as we continue to modularize tests.

MaskedCarrot:
## Explanation

Fixes #2476
Fixes #3312
Fixes #3703

Note that this PR was made in collaboration between @MaskedCarrot, @yashraj-01, and @BenHenning (see commits for specifics). Further, much of this was built as part of #3696 as a separate rewrite to address specific technical needs for both downstream dependent projects (lightweight checkpointing & developer options menu).

At a high-level, this PR moves & rebuilds HintHandler from the UI layer to being a separate, independent logic unit within the domain layer. Key changes that are meant to simplify hint handling:
1. HintHandler was moved to be an interface so that multiple implementations can be provided
2. HelpIndex is now the singular source of truth for all hint/solution state everywhere (domain, checkpoint, and UI); the previous revealed states in Hint/Solution have been removed (which now means those structures are properly immutable for the lifetime of a play session)
3. HintHandler is responsible for everything hints & solution (when there are new ones available, and managing when they've been shown) by being the sole provider of HelpIndex. This includes both computing the available hints/solution as well as managing when hints & the solution have been revealed. This has led to lots of simplifications in the UI.
4. HintHandler's API has been reinvented to be a bit more operation-specific and less generic
5. A monitor was introduced for HintHandler so that the calling controllers know when to re-query for the latest state. There's definitely a data race here, but the data model is eventually consistent so it should be fine.
6. HintHandler's tests make use of lots of new highly specific explorations. These explorations are not accessible from the topic list; they are only included for testing purposes (and are meant to exemplars for future tests).

This PR introduces a few other examples for future changes:
- Module-specific test to verify bindings
- Extensions class tests

While neither of those are unique, the relevant part is that these were the default approaches taken to address the test file presence requirement.

Other details:
- This also fixes #3703 since it was found in passing that //testing module tests were not being built or run in Bazel (except those that are already modularized), leading to some build errors that weren't caught by the CI environment as they should have been.
- QuestionAssessmentProgressControllerTest has been updated to include new scores since it was previously testing impossible scenarios (i.e. assuming hints could be shown/revealed when they wouldn't actually show up)

## Checklist

#### Without this fix, tests (added in this PR because they were not present) fail on the latest develop.

Screenshot from 2021-08-12 03-36-02


MaskedCarrot:
@BenHenning PTAL

MaskedCarrot:
The one failing CI is because there are no tests for the file hintHandler (there are that ensure the proper functioning of the hint handler in state fragment local tests)

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -224,9 +205,6 @@ class StatePlayerRecyclerViewAssembler private constructor(
         /* isCorrectAnswer= */ false,
         gcsEntityId
       )
-      if (playerFeatureSet.hintsAndSolutionsSupport) {
```
BenHenning, line 227: I feel like we still want to have a feature for this, but maybe to gate whether hints & solutions show up for the user instead of activating the handler.

model/src/main/proto/exploration.proto
```diff
@@ -233,9 +233,10 @@ message EphemeralState {
     // message structure if additional data needs to be passed along to the terminal state card.
     bool terminal_state = 6;
   }
-
   // The current status of checkpointing the exploration.
   CheckpointState checkpoint_state = 7;
+  // The state of hints for the current exploration.
```
BenHenning: Nit: newline before this for spacing.

Ditto below for new protos.

model/src/main/proto/exploration.proto
```diff
@@ -335,6 +336,27 @@ message HelpIndex {
   }
 }
 
+// A structure that contains details about the hint that is visible to the learner.
+message HintIndex {
```
BenHenning: Can we consolidate this with HelpIndex to avoid having two different structures?

Note that I specifically called 'HelpIndex' that name so that it could represent both hints & solutions.

model/src/main/proto/exploration.proto
```diff
@@ -335,6 +336,27 @@ message HelpIndex {
...
+// training session.
+message HintState {
+  // HelpIndex for the current state.
+  HelpIndex help_index = 1;
+  // Sequence number of the most recently scheduled task to show help to the learner.
+  int32 hint_sequence_number = 2;
+  // Count of wrong answers that the hint handler has been notified about.
+  int32 tracked_answer_count = 3;
+  // Delay for scheduling a new task to show more help to the learner.
+  int64 delay_to_show_next_hint_and_solution = 4;
```
BenHenning: Why do these need to be passed to the UI layer? They all seem like domain-only implementation details (e.g. for the last one, if we don't want a particular hint to show up until after some time then we shouldn't indicate in the ephemeral state that that hint is ready to be viewed by the learner). This may require some minor adjustments to the way we present hint/solution visibility readiness to the UI layer.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -204,11 +222,7 @@ class ExplorationProgressController @Inject constructor(
     }
   }
 
-  fun submitHintIsRevealed(
-    state: State,
-    hintIsRevealed: Boolean,
-    hintIndex: Int
-  ): LiveData<AsyncResult<Hint>> {
+  fun submitHintIsRevealed(hintIsRevealed: Boolean, hintIndex: Int): LiveData<AsyncResult<Hint>> {
```
BenHenning: While not entirely related to your change, can you please add a KDoc to this method explaining what it does, what parameters are expected, and what it can be expected to return?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -230,15 +244,28 @@ class ExplorationProgressController @Inject constructor(
...
-            state,
+            ephemeralState.state,
             hintIsRevealed,
             hintIndex
           )
-          explorationProgress.stateDeck.pushStateForHint(state, hintIndex)
+          explorationProgress.stateDeck.pushStateForHint(ephemeralState.state, hintIndex)
         } finally {
+          hintHandler.notifyHintIsRevealed(hintIndex)
+          // Schedule a new hints or solution or show a new hint or solution immediately based on
```
```suggestion
          // Schedule, or show immediately, a new hint or solution based on
```

(May require some re-wrapping with the next line).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -255,9 +282,7 @@ class ExplorationProgressController @Inject constructor(
     }
   }
 
-  fun submitSolutionIsRevealed(
-    state: State
-  ): LiveData<AsyncResult<Solution>> {
+  fun submitSolutionIsRevealed(): LiveData<AsyncResult<Solution>> {
```
BenHenning: Ditto for this method (please add a KDoc).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -397,110 +437,31 @@ class ExplorationProgressController @Inject constructor(
...
    */
-  private fun processSaveCheckpointResult(
-    profileId: ProfileId,
-    topicId: String,
-    storyId: String,
-    explorationId: String,
-    lastPlayedTimestamp: Long,
-    newCheckpointState: CheckpointState
-  ) {
+  fun hintAndSolutionTimerCompleted(trackedSequenceNumber: Int, state: State) {
```
BenHenning: I'm wondering if it might be cleaner & simpler to instead send signals to this controller when the timer should be paused/resumed (i.e. during the activity pausing/resuming), but to otherwise perform the timer tracking in the domain layer. My concern with the current implementation is a lot still depends on the UI to do the right things in order for this mechanism to work correctly.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -762,6 +762,45 @@ class StateFragmentLocalTest {
     }
   }
 
+  @Test
+  fun testStateFragment_nextState_viewHint_configChange_noNewHintAvailable() {
```
BenHenning: It's great to see these tests added. Can you briefly verify that these fail without the other fixes in place & add screenshots to your main PR comment showing these tests passing/failing based on the fix? It's useful to do that for bug fixes to be really clear that the tests will prevent the behavior from regressing.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -762,6 +762,45 @@ class StateFragmentLocalTest {
...
+    launchForExploration(FRACTIONS_EXPLORATION_ID_1).use {
+      startPlayingExploration()
+      playThroughState1()
+      produceAndViewFirstHint()
+
+      submitWrongAnswerToState2()
+      onView(isRoot()).perform(orientationLandscape())
```
BenHenning, line 775: I'm actually wondering whether this will correctly repro the issue. I believe @FareesHussain had another way to solve the problem but couldn't actually write tests for it due to Robolectric limitations. Curious to see per the above if these tests actually pass/fail as expected.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -762,6 +762,45 @@ class StateFragmentLocalTest {
...
+      testCoroutineDispatchers.advanceTimeBy(TimeUnit.SECONDS.toMillis(60))
+      onView(withId(R.id.dot_hint)).check(matches(isDisplayed()))
+      onView(isRoot()).perform(orientationLandscape())
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.dot_hint)).check(matches(isDisplayed()))
+    }
+  }
+
+  @Test
+  fun testStateFragment_nextState_viewHint_prevState_wait30seconds_newHintIsAvailable() {
```
BenHenning: Should this read as not available?

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -762,6 +762,45 @@ class StateFragmentLocalTest {
...
+  fun testStateFragment_nextState_viewHint_configChange_noNewHintAvailable() {
+    launchForExploration(FRACTIONS_EXPLORATION_ID_1).use {
+      startPlayingExploration()
+      playThroughState1()
+      produceAndViewFirstHint()
+
+      submitWrongAnswerToState2()
+      onView(isRoot()).perform(orientationLandscape())
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.dot_hint)).check(matches(not(isDisplayed())))
```
BenHenning, line 777: Should the test name indicate that a hint is actually not available given this?

BenHenning (changes requested):
Thanks @MaskedCarrot. I think this is a really awesome improvement. Left an initial round of comments--will take a full pass after these are addressed. I looked at mostly everything, but I didn't look at the new test suite in detail (yet).

Note that after these initial comments are resolved, please make sure that all public API method changes are covered in tests as well (as I'll likely request them to be added in my next review, anyway).

model/src/main/proto/exploration.proto
MaskedCarrot: The UI layer doesn't need all of these, Right now, it only needs help index, hint_sequence_number and delay_to_show_next_hint_and_solution. These are needed because we have the timer to show help in the UI layer, so to schedule a new task to show help, we need to know the delay and the sequence number (if we move the timer to the domain layer, it will then just need help index)

model/src/main/proto/exploration.proto
MaskedCarrot: removed the message `hintIndex` and updated `helpIndex` as follows
```
message HelpIndex {
  oneof index_type {
    int32 hint_index = 1;
    bool show_solution = 2;
    bool everything_revealed = 3;
}
  bool is_all_visible_help_revealed = 4;
}

```

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
MaskedCarrot: updated as follows 
```
    if (recyclerViewAssembler.playerHasSupportForHintsAndSolution()) {
      scheduleShowHintAndSolution(ephemeralQuestion.ephemeralState.hintState)
      showHintsAndSolutions(ephemeralQuestion.ephemeralState.hintState)
    }
```

model/src/main/proto/exploration.proto
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
MaskedCarrot: thanks, I must have missed that

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
MaskedCarrot: No, this test indicates that if the user views a hint and then changes the device configuration, the next hint should not be available. Updated the function name to clearly indicate this.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: I am not sure how we will move the timer to the domain layer, specifically how we will stop and resume it when the activity resumes or pauses. We can use the override methods onPause and onResume, but how will we notify the timer to pause.

I looked into the default timer that android has,
```
 object : CountDownTimer(dealyToShowHints, 1000) {

     override fun onTick(millisUntilFinished: Long) {}

     override fun onFinish() {
         // reveal hint or solution 
     }
 }.start()
 
```



 Alternatively, we can also use coroutine,
```
    backgroundCoroutineScope.launch {
      delay(delayToShowHints)
      // reveal hint/solution
    }
```

But neither of them have pause and resume methods.

One possible way to handle pause and resume cases is to save the delay left and cancel the timer on pause and rerun a new timer with the remaining delay when the activity is resumed.

```
 object : CountDownTimer(dealyToShowHints, 1000) {

     override fun onTick(millisUntilFinished: Long) {
          if (hintSequenceNumber == mostRecentSequenceNumber) {
               explorationProgress.timeElapsed += 1
         }
     }

     override fun onFinish() {
         // reveal hint or solution 
     }
 }.start()
```
But I think we can do it like this only with the default timers and not with coroutines.

Can you please suggest if the above-mentioned way is the correct way to handle this or give a few points on how I should approach this problem?

MaskedCarrot: Update: 

I think keeping track of the time elapsed (shown in the above block of code in the line `explorationProgress.timeElapsed += 1` ) is not a good way of solving this because then every second lock will have to be acquired to update the elapsed time, this will delay the other functions that are waiting to acquire lock (for eg submit answer)

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
MaskedCarrot: This approach solves the two existing issues that were present in the hint hander
1. Hints that were scheduled to be visible when the timer runs out, are visible even on the completed state.
2. new hints were revealed (after the first hint was viewed) upon configuration change.

I have verified that without this fix the tests that check the above two cases fail (added screenshot in pr description).

All tests pass for the current approach.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
```diff
@@ -323,8 +323,11 @@ class StateFragmentPresenter @Inject constructor(
       ephemeralState.stateTypeCase == EphemeralState.StateTypeCase.PENDING_STATE
 
     showOrHideAudioByState(ephemeralState.state)
-    scheduleShowHintAndSolution(ephemeralState.hintState)
-    showHintsAndSolutions(ephemeralState.hintState)
+
+    if (recyclerViewAssembler.playerHasSupportForHintsAndSolution()) {
+      scheduleShowHintAndSolution(ephemeralState.hintState)
+      showHintsAndSolutions(ephemeralState.hintState)
```
BenHenning: Why not call these as part of the assembler's compute method? It's a bit odd/unexpected to be implementing assembler feature behavior checking outside of the assembler, and it seems possible to fold this directly into the assembler itself (& also gain the benefit of sharing code between the exploration & question player as is the purpose of the assembler).

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -273,6 +273,8 @@ class StatePlayerRecyclerViewAssembler private constructor(
     return Pair(conversationPendingItemList, extraInteractionPendingItemList)
   }
 
+  fun playerHasSupportForHintsAndSolution(): Boolean = playerFeatureSet.hintsAndSolutionsSupport
```
BenHenning: Can be omitted, I think, per my other comment.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
BenHenning: Interesting. Thanks for checking @MaskedCarrot. I'm surprised at the discrepancy in how Robolectric is handling these config changes, but I'm glad to see the tests are correctly covering the failure case.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
BenHenning: Much clearer--thanks.

model/src/main/proto/exploration.proto
```diff
@@ -334,25 +350,23 @@ message HelpIndex {
...
-}
 
-// A structure that contains details about the hint that is visible to the learner.
-message HintIndex {
-  // Index of the hint that is visible to the learner.
-  int32 index = 1;
-  // Boolean to indicate if the learner has revealed the hint or not.
-  bool is_hint_revealed = 2;
+  // Boolean to indicate if the learner has revealed the latest  help or not.
+  bool is_all_visible_help_revealed = 4;
```
BenHenning: Nit: double space between 'latest' and 'help'.

Also, the name & the description seem at odds. The doc comment seems to indicate whether the latest one was revealed, but the name suggests "everything" being revealed. I think you meant the former since we have 'everything_revealed' above for the latter. Can you rename this for clarity?

Also, I'm wondering if we could bake this into the oneof somehow. Maybe:

```proto
message HelpIndex {
  oneof index_type {
    int32 available_next_hint_index = 1;
    int32 latest_revealed_hint_index = 2;
    bool show_solution = 3;
    bool everything_revealed = 4;
  }
}
```

Where the default state (unspecified) is "nothing is ready yet" and each of the states implies all previous hints are available. What do you think about this alternative structure?

scripts/assets/kdoc_validity_exemptions.textproto
```diff
@@ -376,7 +376,6 @@ exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rule
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/InteractionObjectTypeExtractorRepository.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/continueinteraction/ContinueModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt"
-exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt"
```
BenHenning, line 379: Awesome! Glad to see one more file removed from this list. :)

Thanks for filling in the missing KDocs.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -245,6 +255,12 @@ class QuestionAssessmentProgressController @Inject constructor(
     }
   }
 
+  /**
+   *  Notifies the [StateDeck] and the [HintHandler] that the solution has been revealed by the user.
```
BenHenning:
```suggestion
   * Notifies the [StateDeck] and the [HintHandler] that the solution has been revealed by the user.
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -109,14 +108,16 @@ class HintHandler @Inject constructor(
           hintSequenceNumber++
           delayShowInitialHintMs
         } else {
+          // Update the sequence number to cancel any pending tasks ans show new help immediately.
```
BenHenning:
```suggestion
          // Update the sequence number to cancel any pending tasks and show new help immediately.
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -109,14 +108,16 @@ class HintHandler @Inject constructor(
...
+          hintSequenceNumber++
```
BenHenning: This seems...unexpected. We are now unconditionally always incrementing the sequence number for this index type--is that actually correct? I wouldn't expect a behavior change since we're just moving this to the domain layer.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: Hmm. I'm not worried about the locking mechanism (yet) since the functions return relatively quickly.

Regarding how to do this, I think there are a couple ways we can do this that will match the current behavior. However, it occurs to me that the current behavior may be wrong (I hypothesize that if you were to pause the app after triggering the timer & wait for the timer's duration before returning, a hint should show up since the timer continues while the app is backgrounded).

For the sake of time, I suggest we go with a parity approach and file an issue tracking this bug since it'll require more work to solve. Could you please file that issue?

Regarding the solution, I don't think we actually need to do anything extra. The timer will continue while the app is backgrounded, recompute the HelpIndex & EphemeralState, notify subscribers, and post the data to the UI. Fortunately, that last part happens via LiveData in the same way the lifecycle-safe timer does, so the result shouldn't actually be observed until the user returns. I think the current batch of tests should cover these cases well enough that if they pass, the behavior should be working as expected with the timer moved to the domain layer.

BenHenning (commented):
Thanks @MaskedCarrot. Followed up on existing discussions.

model/src/main/proto/exploration.proto
MaskedCarrot: this will work. Thanks, changed the structure.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: moved the timer to the domain layer.

model/src/main/proto/exploration.proto
MaskedCarrot: removed all variables except help_index

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: we don't need this because no new help will be scheduled until a delay other than -1 is available. Removed this.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: removed this comment because removed the line below it.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
MaskedCarrot: removed.

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
MaskedCarrot: having the listener works, updated the PR.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -0,0 +1,210 @@
+package org.oppia.android.domain.hintsandsolution
```
vinitamurthi, line 1: Just want to clarify -- this is an exact copy of the existing hint handler correct?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -180,11 +191,29 @@ class ExplorationProgressController @Inject constructor(
               explorationProgress.stateGraph.getState(answerOutcome.stateName),
               prohibitSameStateName = true
             )
+            // Reset the hintState if pending top state has changed.
+            explorationProgress.hintState = hintHandler.reset()
+          } else {
+            //  Schedule, or show immediately, a new hint or solution based on the current ephemeral
```
vinitamurthi: Dont we need test cases in ExplorationProgressControllerTest for this scenario?

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
```diff
@@ -0,0 +1,468 @@
+package org.oppia.android.domain.hintsandsolution
```
vinitamurthi: Again, to clarify -- this is an exact copy of the app module test?

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
```diff
@@ -157,13 +171,31 @@ class QuestionAssessmentProgressController @Inject constructor(
...
+              progress.hintState = hintHandler.reset()
             } else {
               // Otherwise, push a synthetic state for the end of the session.
               progress.stateDeck.pushState(
                 State.getDefaultInstance(),
                 prohibitSameStateName = false
               )
             }
+          } else {
+            // Schedule a new hints or solution or show a new hint or solution immediately based on
```
vinitamurthi, line 176: Do we need tests for this scenario in QuestionAssessmentProgressControllerTest?

model/src/main/proto/exploration.proto
```diff
@@ -319,22 +336,40 @@ message HelpIndex {
   // This type is uninitialized in cases when no index is currently available (either because a hint
   // or solution hasn't yet been triggered, or because there are none to trigger).
   oneof index_type {
-    // Indicates this help index corresponds to the index of a hint within the hint list of a state.
-    int32 hint_index = 1;
+    // Indicates this help index corresponds to the index of the next available hint within the hint
+    // list of a state.
+    int32 available_next_hint_index = 1;
```
vinitamurthi: In general we shouldn't do modifications to protos like this (see more here https://developers.google.com/protocol-buffers/docs/proto3#updating) . Can we instead deprecate hint_index and add the new fields in the end? i.e.
```
reserved 1;

// ...
bool show_solution = 2;

// ...
bool everything_revealed = 3;

// ...
int32 available_next_hint_index = 4;

// ...
int32 latest_revealed_hint_index = 5;
```

model/src/main/proto/exploration.proto
```diff
@@ -319,22 +336,40 @@ message HelpIndex {
...
+// training session.
+message HintState {
+  // HelpIndex for the current state.
+  HelpIndex help_index = 1;
+
+  // Sequence number of the most recently scheduled task to show help to the learner.
+  int32 hint_sequence_number = 2;
+
+  // Delay for scheduling a new task to show more help to the learner.
+  int64 delay_to_show_next_hint_and_solution = 4;
```
vinitamurthi:
```suggestion
  int64 delay_to_show_next_hint_and_solution = 3;
```

model/src/main/proto/exploration_checkpoint.proto
```diff
@@ -46,6 +46,8 @@ message ExplorationCheckpoint {
 
   // The timestamp in milliseconds of when the checkpoint was saved for the first time.
   int64 timestamp_of_first_checkpoint = 9;
+
+  HelpIndex help_index = 10;
```
vinitamurthi, line 51: Add a comment explaining the field

vinitamurthi (changes requested):

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: No this is not an exact copy of the previous hint handler. 

The previous hint handler was responsible for the following
- compute which hint (if any) should be revealed to the user.
- schedule a new timer, observe it and show new hints when the timer runs out.

The new hint handler computes which hint should be revealed but it is no longer responsible to schedule timer to show help, instead, the hint handler computes the delay after which the hint should be revealed. The timer is scheduled by the exploration progress controller using the delay computed by the hint handler.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
MaskedCarrot: We did not have these tests previously, this is a new file.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
MaskedCarrot: We have another issue tracking this (#1273). The state fragment also uses the same logic for hints and solutions and we do have tests in state fragment local tests that test this behaviour.

model/src/main/proto/exploration.proto
MaskedCarrot: done

model/src/main/proto/exploration.proto
MaskedCarrot: done

model/src/main/proto/exploration_checkpoint.proto
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: added more tests to test this.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
```diff
@@ -51,7 +51,7 @@ class QuestionPlayerFragmentPresenter @Inject constructor(
   private val oppiaLogger: OppiaLogger,
   @QuestionResourceBucketName private val resourceBucketName: String,
   private val assemblerBuilderFactory: StatePlayerRecyclerViewAssembler.Builder.Factory,
-  private val splitScreenManager: SplitScreenManager
+  private val splitScreenManager: SplitScreenManager,
```
rt4914, line 54: Why is there `,` at the end?

rt4914 (approved):
Approving for code owner files. Just left one comment. Will mainly defer to @vinitamurthi and @BenHenning for approval.

model/src/main/proto/exploration.proto
```diff
@@ -332,9 +350,30 @@ message HelpIndex {
     // have been revealed and there's no other help to provide. The boolean value here has no
     // importance and is always 'true'.
     bool everything_revealed = 3;
+
+    // Indicates this help index corresponds to the index of the next available hint within the hint
+    // list of a state.
+    int32 available_next_hint_index = 4;
```
```suggestion
    int32 next_available_hint_index = 4;
```

Ditto for the comment above.

model/src/main/proto/exploration.proto
```diff
@@ -332,9 +350,30 @@ message HelpIndex {
...
+
+    // Indicates this help index corresponds to the index of the last revealed hint within the hint
+    // list of a state.
```
BenHenning: I think we're missing some important context in this & the above doc comment. The case above specifically indicates that a new hint is available to be revealed whereas this case represents that all available hints have been revealed. Without that distinction these cases seem like they could be combined into one.

model/src/main/proto/exploration.proto
```diff
@@ -332,9 +350,30 @@ message HelpIndex {
...
+message HintState {
+  // HelpIndex for the current state.
+  HelpIndex help_index = 1;
+
+  // Sequence number of the most recently scheduled task to show help to the learner.
+  int32 hint_sequence_number = 2;
+
+  // Delay for scheduling a new task to show more help to the learner.
+  int64 delay_to_show_next_hint_and_solution = 3;
+}
```
BenHenning: I thought per our discussion only HelpIndex needs to be exposed--why is this needed?

model/src/main/proto/exploration_checkpoint.proto
```diff
@@ -46,6 +46,9 @@ message ExplorationCheckpoint {
 
   // The timestamp in milliseconds of when the checkpoint was saved for the first time.
   int64 timestamp_of_first_checkpoint = 9;
+
+  // The saved help index for the exploration.
```
BenHenning, line 50:
```suggestion
  // The saved help index for the exploration indicating the revealed & available hints/solution of the latest state.
```

BenHenning: Also, are 'solution_is_revealed' & 'hint_index' still needed now that this is here?

BenHenning (commented):
Sending some initial comments--I might actually follow up with direct code changes here.

model/src/main/proto/exploration.proto
MaskedCarrot: only help index is being exposed to the UI, these values are being stored in exploration Progress. we need to have these in the progress controller to correctly handler the timer that shows new help

vinitamurthi (approved):
LGTM for codeowner files

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
```diff
@@ -0,0 +1,468 @@
...
+        currentEphemeralState.state,
+        currentEphemeralState.pendingState.wrongAnswerCount
+      )
+
+    // Now that the first hint is visible, the HintIndex.IndexTypeCase should be equal to HINT_INDEX.
+    assertThat(currentHintState.helpIndex.indexTypeCase).isEqualTo(
+      HelpIndex.IndexTypeCase.AVAILABLE_NEXT_HINT_INDEX
+    )
+    assertThat(currentHintState.helpIndex.availableNextHintIndex).isEqualTo(0)
+    // The delay should now be equal to -1 because there is unrevealed help available.
```
anandwana001:
```suggestion
    // The delay should now be equal to -1 because there is unrevealed hint available.
```

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
```diff
@@ -0,0 +1,468 @@
...
+    )
+    assertThat(currentHintState.helpIndex.latestRevealedHintIndex).isEqualTo(0)
+    // The delay should now be equal to 30 seconds because first hint was revealed.
+    assertThat(currentHintState.delayToShowNextHintAndSolution)
+      .isEqualTo(delayShowAdditionalHintsMs.get())
+  }
+
+  /**
+   * Creates a blank subscription to the current state to ensure that requests to load the
+   * exploration complete, otherwise post-load operations may fail. An observer is required since
```
anandwana001:
```suggestion
   * exploration is complete, otherwise, post-load operations may fail. An observer is required since
```

anandwana001 (approved):
LGTM for code owner files,

nit changes suggested below. Fix the merge conflicts.

BenHenning:
I've verified that the incoming changes build & pass on Bazel. I verified that a subset of them pass all CI checks & Gradle, though with recent merging & other changes it's possible some of these have regressed.

BenHenning:
@rt4914 @anandwana001 @vinitamurthi unfortunately, I think you will each need to take one more pass on this PR. It's been almost entirely rewritten, and mostly by me which means I shouldn't be the one to approve it for merge. I'll still take a pass to double check my work & ensure the bits of code I didn't write mix correctly with the bits I did.

PTAL with priority since this blocks two GSoC projects. Note that this solution has been heavily vetted and I'm confident we can proceed with it; we just need the review completed at this point.

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkStoriesCompletedActivityTest.kt
```diff
@@ -157,13 +158,13 @@ class MarkStoriesCompletedActivityTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
anandwana001: nit fix 
```suggestion
      NetworkConnectionDebugUtilModule::class
```

anandwana001: All other places as well.

anandwana001 (approved):
Test files LGTM so far, keeping eye on checks for the result of all the running tests.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -24,7 +24,9 @@ internal class QuestionAssessmentProgress {
   internal var questionSessionMetrics: MutableList<QuestionSessionMetrics> = mutableListOf()
 
   /** Initialize the assessment with the specified list of questions. */
-  internal fun initialize(questionsList: List<Question>) {
+  internal fun initialize(
+    questionsList: List<Question>
+  ) {
```
vinitamurthi: Is this wrapping needed?

vinitamurthi (approved):
LGTM. Note that the approval is for domain layer and the proto files (since thats what I am a codeowner for)

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
```diff
@@ -51,7 +49,7 @@ class QuestionPlayerFragmentPresenter @Inject constructor(
...
```
rt4914: Remove `,` at the end.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
```diff
@@ -264,13 +265,13 @@ class AppVersionActivityTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Remove `,` at the end

app/src/sharedTest/java/org/oppia/android/app/completedstorylist/CompletedStoryListActivityTest.kt
```diff
@@ -484,13 +485,13 @@ class CompletedStoryListActivityTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
```diff
@@ -142,13 +143,13 @@ class LessonThumbnailImageViewTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
```diff
@@ -280,13 +281,13 @@ class MarginBindingAdaptersTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerMarginBindingAdaptersTest.kt
```diff
@@ -468,13 +469,13 @@ class StateAssemblerMarginBindingAdaptersTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt
```diff
@@ -466,13 +467,13 @@ class StateAssemblerPaddingBindingAdaptersTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/databinding/ViewBindingAdaptersTest.kt
```diff
@@ -200,13 +201,13 @@ class ViewBindingAdaptersTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsActivityTest.kt
```diff
@@ -268,13 +268,13 @@ class DeveloperOptionsActivityTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
```diff
@@ -518,13 +519,13 @@ class DeveloperOptionsFragmentTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkChaptersCompletedActivityTest.kt
```diff
@@ -157,13 +158,13 @@ class MarkChaptersCompletedActivityTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkChaptersCompletedFragmentTest.kt
```diff
@@ -861,13 +862,13 @@ class MarkChaptersCompletedFragmentTest {
...
-      ViewBindingShimModule::class, RatioInputModule::class,
+      ViewBindingShimModule::class, RatioInputModule::class, WorkManagerConfigurationModule::class,
       ApplicationStartupListenerModule::class, LogUploadWorkerModule::class,
-      WorkManagerConfigurationModule::class, HintsAndSolutionConfigModule::class,
+      HintsAndSolutionConfigModule::class, HintsAndSolutionModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
rt4914: Ditto

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkStoriesCompletedActivityTest.kt
rt4914: +1

rt4914 (approved):
LGTM. There is one consistent change across multiple files: there is one unnecessary `,` at the end. Please remove that from all files.

model/src/main/proto/exploration_checkpoint.proto
BenHenning: Those fields have been removed since only HelpIndex is needed now.

BenHenning:
Note that this will now also fix #3703 which was found in passing while addressing review comments.

app/src/main/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerFragmentPresenter.kt
BenHenning: Done.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
BenHenning: Nope; not sure where this happened but fixed.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
BenHenning: Similarly, HintHandlerImplTest is thoroughly being tested and ensures consistent behavior between questions & explorations. Unfortunately, we have only surface-level checks to ensure that the fully questions stack is properly integrating with these systems. We could definitely improve testing coverage for the latter in the future.

model/src/main/proto/exploration.proto
BenHenning: Can this be resolved now?

model/src/main/proto/exploration.proto
BenHenning: Has been resolved in the latest changes.

yashraj-01:
linking issue #3312 as this PR essentially covers the scope of the issue.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/completedstorylist/CompletedStoryListActivityTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerMarginBindingAdaptersTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/ViewBindingAdaptersTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsActivityTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkChaptersCompletedActivityTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkChaptersCompletedFragmentTest.kt
BenHenning: Done.

app/src/sharedTest/java/org/oppia/android/app/devoptions/MarkStoriesCompletedActivityTest.kt
BenHenning: Done.

model/src/main/proto/exploration.proto
BenHenning: Done (in the latest changes).

app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
BenHenning: Has been resolved.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
BenHenning: Has been resolved.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: Can this now be marked as resolved?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
BenHenning: Has been resolved.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
BenHenning: Has been resolved.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgressController.kt
BenHenning: Has been resolved.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
BenHenning: No longer relevant since the suite has been rewritten. Can this be marked as resolved now?

model/src/main/proto/exploration.proto
BenHenning: Has been resolved.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: Has been resolved.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
BenHenning: Can this be marked as resolved?

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
BenHenning: Ditto--suite has been rewritten. Can this be marked as resolved now?

model/src/main/proto/exploration.proto
BenHenning: FWIW this is an ephemeral message so breaking compatibility is fine, but it's a good suggestion in principle.

Also, can this now be marked as resolved?

model/src/main/proto/exploration.proto
BenHenning: No longer needed (I think); HelpIndex is now sufficient due to a bunch of pipelining changes. This is now resolved.

model/src/main/proto/exploration.proto
BenHenning: Addressed in the last changes.

BenHenning (approved):
Thanks for the fast reviews! @rt4914 @vinitamurthi @anandwana001 can you PTAL at the latest changes & make sure they look good (& that the old conversation threads are resolved if so)?

I verified that all static checks, Gradle tests, and Bazel tests seem to work locally before uploading the latest changes. I also synced the branch with develop. If no further questions arise, this PR can probably be merged without waiting for the Bazel tests (to speed things up for @yashraj-01 & @MaskedCarrot).

Also, marking PR as approved since I reviewed all the changes & verified my old comment threads are now addressed.

BenHenning:
Ack @yashraj-01 that SGTM. Thanks!

BenHenning:
Ah excellent. We can see via CI that #3703 is being resolved (all //testing module tests are now showing up in CI as they should have been originally).

anandwana001 (approved):
LGTM,

as per [this](https://github.com/oppia/oppia-android/pull/3659#pullrequestreview-733406199) comment all static and bazel test checked locally, so the PR is gtg.

rt4914 (approved):
LGTM

model/src/main/proto/exploration.proto
BenHenning: This appears to be resolved (mainly since HintState no longer exists). Closing the thread with permission from @vinitamurthi to verify & close out old conversation threads.

model/src/main/proto/exploration.proto
BenHenning: This appears to be resolved (mainly since the old index has now been reserved). Closing the thread with permission from @vinitamurthi to verify & close out old conversation threads.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerTest.kt
BenHenning: This appears to be resolved (the test suite is entirely new, and has been vetted/written by me). Closing the thread with permission from @vinitamurthi to verify & close out old conversation threads.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: This appears to be resolved (tests in ExplorationProgressControllerTest include cases for wrong answer submission--I verified by removing the new version of this code & observed 9 test failures in ExplorationProgressControllerTest). Closing the thread with permission from @vinitamurthi to verify & close out old conversation threads.

BenHenning:
All conversation threads have been verified & closed, required CI checks are passing, and required approvals are in. Going ahead and merging this since I verified that the new testing tests are correctly in CI now, and I verified locally that all Bazel tests are passing. This is blocking two separate streams of time-sensitive work, so I think we can take the risk of merging this in now.

Note that it's theoretically possible that the testing module tests that were previously missing from CI now fail on develop. If this occurs, I'll fix forward to avoid reverting this large, blocking PR & because those tests are not yet blocking (we're running them in Gradle & they've been passing on CI--that check _is_ required).

Thanks again all for the repeated reviews & collaboration on this tricky unexpected design challenge. I'm glad we found a workable solution quickly.

PR #3667, opened 2021-08-10
Fix #3656: [RTL] High-fi Add support for RTL in Story Activity

### LINKED ISSUE #3656: [RTL] High-fi Add support for RTL in Story Activity
## Explanation
Chapter summary in Story view need to support RTL

image

veena14cs:
## Explanation
This PR fixes #3656 . This PR focuses on textalignment in `storyView`.

## Screenshots LTR and RTL
Mobile 
Screenshot_1629114810....Screenshot_1629114759

Screenshot_1629114799.....Screenshot_1629114768

Tablet
Screenshot_1629115032......Screenshot_1629115311
Screenshot_1629115325.....Screenshot_1629115023

## Checklist


rt4914 (changes requested):
@veena14cs The RTL screenshot does not look correct. The text is left aligned. Any specific reason behind that?

veena14cs:
Rajat that's what I was discussing in the meeting. I am not understanding why it is not aligning to right even if I use textdirection.

seanlip:
@veena14cs, I think you'll need to be the one debugging this. What specific investigations have you made / what things have you tried, in order to get to the root of the issue?

veena14cs:
> @veena14cs, I think you'll need to be the one debugging this. What specific investigations have you made / what things have you tried, in order to get to the root of the issue?

I tried setting textdirection to RTL, but it is not changing the alignment. I tried to set width to wrap_content and used horizontal_bias 0.0 but there is no any changes in the alignment. I am not able to find the issue. Because when I had tested it earlier the textalignment was correct for RTL I have added screenshots also for the same in the [doc](https://docs.google.com/document/d/1Fl1ar5vcdLvay7ZIJLUFQro1wEf1yUEicwF-CKcvwJ0/edit#bookmark=id.xqlfwuybao1l) . I am not finding where is the bug.

seanlip:
If it worked before, but doesn't work now, then the general way to debug such cases is to figure out which specific change you made caused the behaviour to stop working. 

You can do this by undoing your changes and applying them one by one until you find the point where the behaviour differs. That will tell you where you'd need to dig in more detail.

veena14cs:
> If it worked before, but doesn't work now, then the general way to debug such cases is to figure out which specific change you made caused the behaviour to stop working.
> 
> You can do this by undoing your changes and applying them one by one until you find the point where the behaviour differs. That will tell you where you'd need to dig in more detail.

Actually there were no changes from my side. There were some PRs after that which were done. I digged into it and tried to set the previous changes. But failed to fix. I will check again in detail if I can find the bug.

seanlip:
Sounds good, it might be that some other PR broke it. You can use `git bisect` (see https://git-scm.com/docs/git-bisect) to help you figure out which one did.

veena14cs:
> Sounds good, it might be that some other PR broke it. You can use `git bisect` (see https://git-scm.com/docs/git-bisect) to help you figure out which one did.

Thanks @seanlip I will try this.

veena14cs:
@rt4914 PTAL I have fixed this issue with textalignment.

app/src/main/res/layout-land/story_chapter_view.xml
```diff
@@ -68,13 +68,14 @@
...
             android:id="@+id/chapter_summary"
-            android:layout_width="0dp"
+            android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_marginStart="8dp"
             android:layout_marginTop="8dp"
             android:layout_marginEnd="8dp"
             android:layout_marginBottom="4dp"
             android:fontFamily="sans-serif"
+            android:textAlignment="viewStart"
```
rt4914, line 78: @BenHenning Can you confirm if we can use this attribute or not? Because in past we have discussed not to use this attribute to fix RTL

app/src/main/res/layout-sw600dp-port/story_chapter_view.xml
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rt4914: @veena14cs Why is this file getting added? The landscape mocks are quite different and therefore I suggest not to create this file.

rt4914 (changes requested):
@veena14cs PTAL

app/src/main/res/layout-sw600dp-port/story_chapter_view.xml
veena14cs: There was no portrait file so I created one. I will revert this file.

veena14cs: Done,

app/src/main/res/layout-land/story_chapter_view.xml
BenHenning: Seems fine, though I'm unsure when we should be using ``textStart`` vs. ``viewStart``. @veena14cs does ``textStart`` have the same behavior? Per [documentation](https://developer.android.com/reference/android/view/View#attr_android:textAlignment) it may be better to always default to ``textStart`` since it's more likely to look correct (since it provides alignment based on the paragraph rather than the view), while also accounting for RTL scenarios.

BenHenning (changes requested):
Left one follow-up comment.

@veena14cs can you please also add tests for these changes to make sure they're well tested?

app/src/main/res/layout-land/story_chapter_view.xml
veena14cs: Here, what I seem to be reasonable is, we are aligning based on the layout direction and if we want based on textdirection we need to add direction for paragraph. As it does not understand if the paragraph is LTR or RTL.

veena14cs:
> Left one follow-up comment.
> 
> @veena14cs can you please also add tests for these changes to make sure they're well tested?

Followed up with your comment and added tests for the same.

rt4914 (approved):
LGTM, thanks.

app/src/main/res/layout-land/story_chapter_view.xml
rt4914: I think `viewStart` is correct for this. http://androidsbs.blogspot.com/2013/11/androidtextalignment.html#:~:text=textStart%20%2D%20Align%20to%20the%20start%20of%20the%20paragraph%2C%20e.g.%20ALIGN_NORMAL.&text=center%20%2D%20Center%20the%20paragraph%2C%20e.g.,is%20LTR%2C%20and%20ALIGN_RIGHT%20otherwise.

veena14cs: Thanks for this reference @rt4914 .

veena14cs:
> LGTM, thanks.

Thanks @rt4914

app/src/main/res/layout-land/story_chapter_view.xml
BenHenning: Oh, I assumed that the paragraph's start with align with RTL/LTR. If that's not the case, that's a bit weird.

Either way, I think this approach is sensible.

BenHenning (approved):
Thanks @veena14cs. This LGTM.

app/src/main/res/layout-land/story_chapter_view.xml
veena14cs: Thanks Ben!

PR #3668, opened 2021-08-11
Fix #3488: Domain layer mechanism to resume explorations.

### LINKED ISSUE #3488: Implement domain layer mechanism to resume explorations with checkpoints.

MaskedCarrot:
## Explanation

Fix #3488 

## Checklist

#### ExplorationActivityTest
Screenshot from 2021-08-20 03-04-28

#### StateFragmentTest (only the new test added)
Screenshot from 2021-08-20 03-24-41


BenHenning:
Will need to take a look at this tomorrow.

app/src/main/java/org/oppia/android/app/player/state/testing/StateFragmentTestActivityPresenter.kt
```diff
@@ -98,7 +99,8 @@ class StateFragmentTestActivityPresenter @Inject constructor(
       topicId,
       storyId,
       explorationId,
-      shouldSavePartialProgress = false
+      shouldSavePartialProgress = false,
+      explorationCheckpoint = ExplorationCheckpoint.getDefaultInstance()
```
aggarwalpulkit596, line 106: Since we are passing this to every function call and the value is the same too, shouldn't we make it a default argument in the function definition itself? 
@BenHenning @rt4914 Your thoughts?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -720,4 +728,160 @@ class ExplorationProgressController @Inject constructor(
       lastPlayedTimestamp
     )
   }
+
```
aggarwalpulkit596: Kdocs missing

aggarwalpulkit596 (changes requested):

app/src/main/java/org/oppia/android/app/player/state/testing/StateFragmentTestActivityPresenter.kt
MaskedCarrot: This is a temporary measure, it is changed in the following PRs ( when we add update the app layer mechanism to start exploration.)

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

anandwana001:
Assign me back once the UI part gets reviewed, to prevent multiple reviews for test cases. 
Also add screenshot of espresso test result for app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt

app/src/main/java/org/oppia/android/app/player/state/testing/StateFragmentTestActivityPresenter.kt
BenHenning: I think I lean toward this being explicit given that one "happy path" is passing something non-default in.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
BenHenning: FWIW, KDocs aren't required for private members (per https://developer.android.com/kotlin/style-guide#usage), but they're definitely fine to add.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -735,4 +743,164 @@ class ExplorationProgressController @Inject constructor(
...
+      progress.stateDeck.resetDeck(progress.stateGraph.getState(exploration.initStateName))
+    } else {
+      progress.stateDeck.resumeDeck(
+        createPendingTopStateFromCheckpoint(progress),
+        getPreviousStatesFromCheckpoint(progress),
+        progress.currentExplorationCheckpoint.pendingUserAnswersList,
+        progress.currentExplorationCheckpoint.stateIndex
+      )
+    }
+  }
```
BenHenning: This feels a bit like something that should be in the progress class itself since it's all about properly initializing the progress state. I suggest moving it over to keep this controller a bit more focused on its direct public API implementations.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -735,4 +743,164 @@ class ExplorationProgressController @Inject constructor(
...
+      HelpIndex.IndexTypeCase.SHOW_SOLUTION, HelpIndex.IndexTypeCase.EVERYTHING_REVEALED -> {
+        // All the hints are visible and revealed if helpIndex.indexTypeCase is equal to
+        // SHOW_SOLUTION or EVERYTHING_REVEALED.
+        pendingStateHintList.forEach { hint ->
+          updatedHintList.add(hint.toBuilder().setHintIsRevealed(true).build())
+        }
+      }
+      else -> updatedHintList.addAll(pendingStateHintList)
+    }
+    return updatedHintList
```
```suggestion
    return when (helpIndex.indexTypeCase) {
      HelpIndex.IndexTypeCase.HINT_INDEX -> {
        pendingStateHintList.forEachIndexed { index, hint ->
          if (index < helpIndex.hintIndex) {
            // Mark all hints as visible and revealed which have an index less than that stored in
            // the HintState.
            listOf(hint.toBuilder().setHintIsRevealed(true).build())
          } else if (index == helpIndex.hintIndex) {
            // Add the currently shown hint to the updated hint list and set hintIsRevealed
            // depending upon if the current hint was revealed or not.
            listOf(hint.toBuilder().setHintIsRevealed(helpIndex.isAllVisibleHelpRevealed).build())
          } else {
            // Add all the remaining hints that are not yet visible to the user to the updated hint
            // list.
            listOf(hint.toBuilder().setHintIsRevealed(false).build())
          }
        }
      }
      HelpIndex.IndexTypeCase.SHOW_SOLUTION, HelpIndex.IndexTypeCase.EVERYTHING_REVEALED -> {
        // All the hints are visible and revealed if helpIndex.indexTypeCase is equal to
        // SHOW_SOLUTION or EVERYTHING_REVEALED.
        pendingStateHintList.forEach { hint ->
          listOf(hint.toBuilder().setHintIsRevealed(true).build())
        }
      }
      else -> pendingStateHintList
    }
```

Better to avoid mutable state. Ditto elsewhere.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -735,4 +743,164 @@ class ExplorationProgressController @Inject constructor(
...
+          .build()
+      }
+      else -> pendingTopState.interaction.solution
+    }
+  }
+
+  /**
+   * Creates a list of completed states from the saved [ExplorationCheckpoint].
+   *
+   * @return [List] of [EphemeralState] containing all the states that were completed before the
```
```suggestion
   * @return [List] of [EphemeralState]s containing all the states that were completed before the
```

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -735,4 +743,164 @@ class ExplorationProgressController @Inject constructor(
...
+   *     checkpoint was created
+   */
+  private fun getPreviousStatesFromCheckpoint(
+    progress: ExplorationProgress
+  ): List<EphemeralState> {
+    val previousStates: MutableList<EphemeralState> = ArrayList()
+    progress.currentExplorationCheckpoint.completedStatesInCheckpointList
+      .forEachIndexed { index, state ->
```
BenHenning: Per the above suggestion, you could use ``mapIndexed`` here, instead, to avoid the mutable list.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -735,4 +743,164 @@ class ExplorationProgressController @Inject constructor(
...
+  private fun loadHintState(progress: ExplorationProgress) {
+    progress.hintState =
+      if (progress.currentExplorationCheckpoint == ExplorationCheckpoint.getDefaultInstance()) {
+        HintState.getDefaultInstance()
+      } else {
+        HintState.newBuilder().apply {
+          helpIndex = progress.currentExplorationCheckpoint.helpIndex
+          trackedAnswerCount = progress.currentExplorationCheckpoint.pendingUserAnswersCount
+          hintSequenceNumber = 0
+          delayToShowNextHintAndSolution = -1
```
BenHenning: I think this means that the next hint will default to the first hint's initial triggering criteria vs. the "next hint" criteria (per the diagram in HintHandler's documentation). Is that expected from a UX/product perspective?

BenHenning (changes requested):
Thanks @MaskedCarrot. Reviewed everything per my codeowners except ExplorationProgressControllerTest. The mechanism overall seems good, but I suspect it may change a bit per the base PR.

Please also update the title of this PR & downstream ones to include the PR that they're subsequently blocked on & using as a base branch (to make it clear that the PR is blocked & not yet ready for merging).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: changed to mapIndexed

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: changed to mapIndexed

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: moved to exploration progress

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
MaskedCarrot: actually, it does not matter what we set the delay here because the delay is recomputed based on the ephemeral state before anything is visible to the user. I set it to -1 as a default value because elsewhere we are using -1 to indicate that no timer should be scheduled.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -113,4 +119,144 @@ internal class ExplorationProgress {
...
+
+  /**
+   * Initializes the variables of [StateDeck]. If the [ExplorationCheckpoint] is of type default
+   * instance, the values of [StateDeck] are reset. Otherwise, the variables of [StateDeck] are
+   * re-initialized with the values created from the saved [ExplorationCheckpoint].
+   *
+   * This function expects explorationProgress.hintState to be initialized with the correct values,
+   * so it should only be called after the function [loadHintState] has executed.
+   */
+  internal fun loadStateDeck(exploration: Exploration) {
```
vinitamurthi: Add a `@param` annotation here

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -113,4 +119,144 @@ internal class ExplorationProgress {
...
+        .build()
+    return pendingTopState.toBuilder().setInteraction(interactionBuilder).build()
+  }
+
+  /**
+   * Mark all hints as reveled in the pendingState that were revealed for the current state pending
+   * state before the checkpoint was saved.
+   *
+   * @param pendingStateHintList the list of hint for the current pending state
+   * @param helpIndex the state of hints for the exploration which was generated using the saved
```
vinitamurthi: Add a `@return` annotation as well

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -113,4 +119,144 @@ internal class ExplorationProgress {
...
+      else -> pendingStateHintList
+    }
+  }
+
+  /**
+   * Set solution is reveled in the pendingState to true or false depending upon if solution was
+   * revealed for the current state pending state before the checkpoint was saved.
+   *
+   * @param pendingTopState the pending state created from the checkpoint
+   * @param helpIndex the state of solution for the exploration which was generated using the saved
```
vinitamurthi: Add an `@return` annotation too

model/src/main/proto/exploration_checkpoint.proto
```diff
@@ -30,24 +30,17 @@ message ExplorationCheckpoint {
...
-  // of the hint that the learner has revealed.
-  int32 hint_index = 5;
-
-  // Stores true if the solution to the pending top state has been revealed by the learner otherwise
-  // it stores false.
-  bool solution_is_revealed = 6;
-
   // The title of the exploration whose checkpoint is saved.
-  string exploration_title = 7;
+  string exploration_title = 5;
```
vinitamurthi: Avoid modifying existing proto fields. Instead you can deprecate fields 5 and 6 and keep 7 as is.
i.e. 
```
reserved 5,6;

// ....
string exploration_title = 7;

// ...
string exploration_version = 8;

.....
HelpIndex help_index = 10;
```

vinitamurthi (changes requested):

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

model/src/main/proto/exploration_checkpoint.proto
MaskedCarrot: done

aggarwalpulkit596 (approved):

rt4914 (approved):
Approved for code-owner files.

vinitamurthi (approved):
LGTM for codeowner files

BenHenning:
Sorry, I don't have enough focus to review this in detail right now. However, I also suspect that this is going to change a bit after #3659 has my changes incorporated since I removed the non-HelpIndex pieces from ExplorationCheckpoint & changed a few other things. I'm certain that we no longer need to pass along the sequence number or other data, and I'm pretty sure the changes from #3659 will be sufficient for hint/solution restoration to "just work" when loading a checkpoint. @MaskedCarrot please update this PR with the changes from there once they are ready & then assign back.

anandwana001 (approved):
LGTM for code owners file. 

Could you add a espresso test result screenshot for future reference fo this file - app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -48,6 +48,15 @@ import org.oppia.android.app.model.State
  */
 interface HintHandler {
 
+  /**
+   * Restores the local variables of hint handler to a specific point in the exploration.
```
BenHenning:
```suggestion
   * Restores the hint handler to a previously cached state so that the user can continue where they
   * left off.
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -48,6 +48,15 @@ import org.oppia.android.app.model.State
...
+   *
+   * @param trackedWrongAnswerCount the count of wrong answers saved in the checkpoint
+   * @param helpIndex the [HelpIndex] saved in the checkpoint.
```
BenHenning:
```suggestion
   * @param helpIndex the cached state of hints/solution from the checkpoint
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -48,6 +48,15 @@ import org.oppia.android.app.model.State
...
+   * @param hintCount the total number of hint available on the saved pending state
```
BenHenning:
```suggestion
   * @param hintCount the total number of hints available in the restored pending state
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -48,6 +48,15 @@ import org.oppia.android.app.model.State
...
+   * */
+  fun restoreHintHandler(trackedWrongAnswerCount: Int, helpIndex: HelpIndex, hintCount: Int)
```
BenHenning: Please add HintHandler-specific tests for the new behaviors introduced in this method.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
```diff
@@ -33,11 +35,55 @@ class HintHandlerImpl private constructor(
...
+      }
+    }
+  }
+
   override fun startWatchingForHintsInNewState(state: State) {
     handlerLock.withLock {
       pendingState = state
       hintMonitor.onHelpIndexChanged()
-      maybeScheduleShowHint(wrongAnswerCount = 0)
+      maybeScheduleShowHint(trackedWrongAnswerCount)
```
BenHenning: This is semantically a bit strange. We'd expect a new state to have no answers to continue with.

Maybe instead we should have an API for HintHandler that's resuming for a given state, e.g. something like:

``fun resumeHintsForSavedState(trackedWrongAnswerCount: Int, helpIndex: HelpIndex, state: State)``

where we call that instead of this method in the event that we're restoring things. I think that'll simplify stuff a bit since it also avoids needing to restore the HintHandler in ExplorationProgress (which seems a bit odd since that class otherwise doesn't care at all about HintHandler).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
```diff
@@ -110,4 +114,67 @@ internal class ExplorationProgress {
...
     SUBMITTING_ANSWER
   }
+
+  /**
+   * Loads [StateDeck] and [HintHandler] for the current exploration.
+   *
+   * @param exploration the current exploration that is being played
+   * @param hintHandler the [HintHandler] that will be used to show help in the current exploration
+   */
+  internal fun resetOrResumeExploration(exploration: Exploration, hintHandler: HintHandler) {
```
BenHenning: I suggest instead just making this resume & doing the check at the callsite since it avoids needing to recheck the state in multiple places. You can also reduce the hedging in the documentation since things become much more crisp: it's just restoring state, no conditions.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+    navigateToPrototypeFractionInputState()
+    // Submit 2 wrong answers to trigger the hint.
+    submitWrongAnswerForPrototypeState2()
+    submitWrongAnswerForPrototypeState2()
+
+    verify(
+      mockCurrentStateLiveDataObserver,
+      atLeastOnce()
+    ).onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
```
BenHenning: Here & elsewhere in the updated/new tests: what is this verification attempting to ensure? It seems unnecessary.

Edit: I realized this is needed for tests that then use the retrieved state so I revise this comment to only correspond to tests that don't use the current state.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+      explorationCheckpoint = ExplorationCheckpoint.getDefaultInstance()
+    )
+    testCoroutineDispatchers.runCurrent()
+
+    verify(mockCurrentStateLiveDataObserver, atLeastOnce())
+      .onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
+    val currentState = currentStateResultCaptor.value.getOrThrow()
+    assertThat(currentState.checkpointState)
+      .isEqualTo(CheckpointState.CHECKPOINT_UNSAVED)
```
BenHenning: Can this fit on one line?

```suggestion
    assertThat(currentState.checkpointState).isEqualTo(CheckpointState.CHECKPOINT_UNSAVED)
```

Prefer to pack everything onto one line unless it needs to wrap. Ditto elsewhere.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+  }
+
+  @Test
+  fun testCheckpointing_saveCheckpoint_checkCheckpointStateIsCorrect() {
```
BenHenning:
```suggestion
  fun testCheckpointing_saveCheckpoint_checkCheckpointStateIsSuccessful() {
```

We should be a bit more specific as to what 'correct' is when we can be.

Ditto for the other new tests.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+    playThroughPrototypeState2AndMoveToNextState()
+
+    testCoroutineDispatchers.runCurrent()
+
+    verify(mockCurrentStateLiveDataObserver, atLeastOnce())
+      .onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
+    val currentState = currentStateResultCaptor.value.getOrThrow()
+    assertThat(currentState.checkpointState)
+      .isEqualTo(CheckpointState.CHECKPOINT_SAVED_DATABASE_EXCEEDED_LIMIT)
```
BenHenning, line 2530: Why was this hit in this test? I'm not seeing anything in the arrangement that would result in the database being full.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
-    testCoroutineDispatchers.runCurrent()
+    playThroughPrototypeState1AndMoveToNextState()
+    submitWrongAnswerForPrototypeState2()
+    submitWrongAnswerForPrototypeState2()
+    // Reveal the hint, then submit another wrong answer to trigger the solution.
+    verifyOperationSucceeds(explorationProgressController.submitHintIsRevealed(hintIndex = 0))
+
+    testCoroutineDispatchers.advanceTimeBy(TimeUnit.SECONDS.toMillis(30))
+
+    endExploration()
```
BenHenning:
```suggestion
    testCoroutineDispatchers.advanceTimeBy(TimeUnit.SECONDS.toMillis(30))
    endExploration()
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+
+    playExploration(
+      profileId.internalId,
+      TEST_TOPIC_ID_0,
+      TEST_STORY_ID_0,
+      TEST_EXPLORATION_ID_2,
+      shouldSavePartialProgress = true,
+      explorationCheckpoint = retrieveExplorationCheckpoint(profileId, TEST_EXPLORATION_ID_2)
+    )
+    // Verify that the helpIndex.IndexTypeCase is equal EVERYTHING_IS_REVEALED because all available
```
BenHenning, line 3146:
```suggestion

    // Verify that the helpIndex.IndexTypeCase is equal EVERYTHING_IS_REVEALED because all available
```

For this & the above comment, we should make sure the arrange, act, and assert blocks are clearly spearated in the test.

Ditto for the other tests.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2235,154 +2338,797 @@ class ExplorationProgressControllerTest {
...
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
+    val currentState = currentStateResultCaptor.value.getOrThrow()
+    assertThat(currentState.pendingState.helpIndex.indexTypeCase)
+      .isEqualTo(HelpIndex.IndexTypeCase.SHOW_SOLUTION)
+    assertThat(currentState.isHintRevealed(0)).isTrue()
+    assertThat(currentState.isSolutionRevealed()).isFalse()
+  }
 
+  @Test
+  fun testCheckpointing_revealedSolution_resumeExp_revealedSolIsVisibleOnPendingState() {
```
BenHenning, line 3159: Overall, the new tests are really good.

I suggest also adding 1 test that completes a few states, resumes, completes a few more, then resumes again and finishes the exploration to demonstrate multiple start/stops & that a restored exploration can be successfully finished.

I also suggest adding a test that shows finishing an exploration results in no/an empty checkpoint being saved (unless this test already exists?).

BenHenning (changes requested):
Thanks @MaskedCarrot! I really like how much the implementation has overall simplified.

Took a full pass and left comments--PTAL.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: I have changed the size limit allocated to the checkpoint database, for testing this controller the size is set to 150 Bytes. This value is set so that the checkpoint database exceeds the allocated size limit when the second checkpoint is saved for TEST_EXPLORATION_ID_2

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: added more tests.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: updated test names

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -1468,6 +1472,19 @@ class StateFragmentTest {
     }
   }
 
+  @Test
+  @RunOn(TestPlatform.ESPRESSO) // TODO(#1612): Enable for Robolectric.
```
BenHenning, line 1478: Why doesn't this work on Robolectric?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -23,21 +23,23 @@ import org.oppia.android.app.model.State
...
   /**
    * Starts watching for potential hints to be shown (e.g. if a user doesn't submit an answer after
    * a certain amount of time). This is meant to only be called once at the beginning of a state.
    */
   fun startWatchingForHintsInNewState(state: State)
 
+  /**
+   * Starts watching for potential hints to be shown when the exploration is resumed. (e.g. if a
+   * user doesn't submit an answer after a certain amount of time). This is meant to only be called
+   * once when the the exploration is r resumed.
```
BenHenning:
```suggestion
   * once when the the exploration is resumed.
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -23,21 +23,23 @@ import org.oppia.android.app.model.State
...
+   *
+   * @param trackedWrongAnswerCount the count of wrong answers saved in the checkpoint
+   * @param helpIndex the cached state of hints/solution from the checkpoint
+   * @param state the restored pending state.
```
BenHenning:
```suggestion
   * @param state the restored pending state
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
```diff
@@ -23,21 +23,23 @@ import org.oppia.android.app.model.State
...
-  fun restoreHintHandler(trackedWrongAnswerCount: Int, helpIndex: HelpIndex, hintCount: Int)
-
   /**
    * Starts watching for potential hints to be shown (e.g. if a user doesn't submit an answer after
    * a certain amount of time). This is meant to only be called once at the beginning of a state.
    */
   fun startWatchingForHintsInNewState(state: State)
 
+  /**
+   * Starts watching for potential hints to be shown when the exploration is resumed. (e.g. if a
```
BenHenning:
```suggestion
   * Starts watching for potential hints to be shown when the exploration is resumed (e.g. if a
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2982,11 +3017,49 @@ class ExplorationProgressControllerTest {
...
+      .onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
+    val currentState = currentStateResultCaptor.value.getOrThrow()
+    assertThat(currentState.pendingState.helpIndex.indexTypeCase)
+      .isEqualTo(HelpIndex.IndexTypeCase.LATEST_REVEALED_HINT_INDEX)
+    assertThat(currentState.isHintRevealed(0)).isTrue()
+  }
 
+  @Test
+  fun testCheckpointing_revealedHintIsVisible_resumeExp_Wait10Seconds_solutionIsNotVisible() {
```
BenHenning:
```suggestion
  fun testCheckpointing_revealedHintIsVisible_resumeExp_wait10Seconds_solutionIsNotVisible() {
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -3009,6 +3083,45 @@ class ExplorationProgressControllerTest {
     assertThat(currentState.isHintRevealed(0)).isTrue()
   }
 
+  @Test
+  fun testCheckpointing_revealedHintIsVisible_resumeExp_Wait30Seconds_solutionIsNotVisible() {
```
BenHenning: Ditto

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -3009,6 +3083,45 @@ class ExplorationProgressControllerTest {
     assertThat(currentState.isHintRevealed(0)).isTrue()
```
BenHenning, line 3072: Should this instead be verifying that the solution isn't visible, and that the index type is not SHOW_SOLUTION?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -498,22 +500,34 @@ class ExplorationProgressController @Inject constructor(
...
     // Mark a checkpoint in the exploration once the exploration has loaded.
     saveExplorationCheckpoint()
   }
 
   private fun computeCurrentEphemeralState(): EphemeralState =
     explorationProgress.stateDeck.getCurrentEphemeralState(computeCurrentHelpIndex())
 
-  private fun computeCurrentHelpIndex(): HelpIndex = hintHandler.getCurrentHelpIndex()
+  private fun computeCurrentHelpIndex(): HelpIndex =
+    hintHandler.getCurrentHelpIndex()
```
BenHenning: Suggest reverting since this can still fit on one line.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
```diff
@@ -77,6 +79,53 @@ class HintHandlerImpl private constructor(
...
+          solutionIsAvailable = true
+          solutionIsRevealed = false
+        }
+        EVERYTHING_REVEALED -> {
+          // 1 is subtracted from the hint count because hints are indexed from 0.
+          lastRevealedHintIndex = state.interaction.hintCount - 1
+          latestAvailableHintIndex = state.interaction.hintCount - 1
+          solutionIsAvailable = true
+          solutionIsRevealed = true
+        }
```
BenHenning:
```suggestion
        SHOW_SOLUTION, EVERYTHING_REVEALED -> {
          // 1 is subtracted from the hint count because hints are indexed from 0.
          lastRevealedHintIndex = state.interaction.hintCount - 1
          latestAvailableHintIndex = state.interaction.hintCount - 1
          solutionIsAvailable = true
          solutionIsRevealed = helpIndex.indexTypeCase == EVERYTHING_REVEALED
        }
```

Simplification.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
```diff
@@ -77,6 +79,53 @@ class HintHandlerImpl private constructor(
...
+          latestAvailableHintIndex = helpIndex.nextAvailableHintIndex
+          solutionIsAvailable = false
+          solutionIsRevealed = false
+        }
+        LATEST_REVEALED_HINT_INDEX -> {
+          lastRevealedHintIndex = helpIndex.latestRevealedHintIndex
+          latestAvailableHintIndex = helpIndex.latestRevealedHintIndex
+          solutionIsAvailable = false
+          solutionIsRevealed = false
+        }
```
BenHenning, line 100:
```suggestion
        NEXT_AVAILABLE_HINT_INDEX, LATEST_REVEALED_HINT_INDEX -> {
          lastRevealedHintIndex = if (helpIndex.indexTypeCase == NEXT_AVAILABLE_HINT_INDEX) {
            helpIndex.nextAvailableHintIndex - 1
          } else helpIndex.nextAvailableHintIndex
          latestAvailableHintIndex = helpIndex.nextAvailableHintIndex
          solutionIsAvailable = false
          solutionIsRevealed = false
        }
```

Simplification (may require directly importing NEXT_AVAILABLE_HINT_INDEX).

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -1622,7 +1662,7 @@ class ExplorationActivityTest {
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class, PracticeTabModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
       TestExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
-      NetworkConnectionDebugUtilModule::class
+      NetworkConnectionDebugUtilModule::class,
```
BenHenning:
```suggestion
      NetworkConnectionDebugUtilModule::class
```

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
```diff
@@ -199,6 +199,239 @@ class HintHandlerImplTest {
...
+    waitFor60Seconds()
+
+    // Verify that the monitor is called again (since there's a hint now available).
+    verify(mockHintMonitor).onHelpIndexChanged()
+  }
+
+  @Test
+  fun testResumeHints_stateWithHints_hintVisible_wait10Seconds_doesNotCallMonitorAgain() {
+    val state = expWithHintsAndSolution.getInitialState()
+
```
BenHenning, line 350:
```suggestion
```

The ``val state`` bit is part of the arrangement block. Ditto for other similar tests.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
```diff
@@ -199,6 +199,239 @@ class HintHandlerImplTest {
...
+        latestRevealedHintIndex = 0
+      }.build(),
+      state
+    )
+    reset(mockHintMonitor)
+
+    waitFor30Seconds()
+
+    // Verify that the monitor is called again (since there's a solution now available).
+    verify(mockHintMonitor).onHelpIndexChanged()
```
BenHenning, line 480: For each of these new tests where the monitor is called, please also add corresponding tests that verify the current HelpIndex state.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
```diff
@@ -199,6 +199,239 @@ class HintHandlerImplTest {
...
+    )
+    reset(mockHintMonitor)
+
+    waitFor10Seconds()
+
+    verifyNoMoreInteractions(mockHintMonitor)
+  }
+
+  @Test
+  fun testResumeHints_stateWithHints_hintReveled_wait30Seconds_callsMonitorAgain() {
```
BenHenning: These are really nice tests to verify that restoration works as expected. Please also add the two other missing cases & related wait tests) for SHOW_SOLUTION and EVERYTHING_REVEALED.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2379,8 +2541,744 @@ class ExplorationProgressControllerTest {
...
+    playThroughPrototypeState8AndMoveToNextState()
+    playThroughPrototypeState9AndMoveToNextState()
+    playThroughPrototypeState10AndMoveToNextState()
+
+    // Verify that the last state is terminal.
+    verify(mockCurrentStateLiveDataObserver, atLeastOnce())
+      .onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
+    val currentState = currentStateResultCaptor.value.getOrThrow()
+    assertThat(currentState.stateTypeCase).isEqualTo(TERMINAL_STATE)
```
BenHenning, line 3246: Can we also verify that the checkpoint is empty?

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
BenHenning: Ah. Can you add a single line comment explaining that here to help contextualize it?

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -2227,6 +2327,36 @@ class ExplorationProgressControllerTest {
...
+    )
+
+    navigateToPrototypeFractionInputState()
+    // Submit 2 wrong answers to trigger the hint.
+    submitWrongAnswerForPrototypeState2()
+    submitWrongAnswerForPrototypeState2()
+
+    verify(mockCurrentStateLiveDataObserver, atLeastOnce())
+      .onChanged(currentStateResultCaptor.capture())
+    assertThat(currentStateResultCaptor.value.isSuccess()).isTrue()
```
BenHenning: Copying from https://github.com/oppia/oppia-android/pull/3668#discussion_r691835241 since that comment seemed to break on GitHub somehow.

Here & elsewhere in the updated/new tests: what is this verification attempting to ensure? It seems unnecessary.

Edit: I realized this is needed for tests that then use the retrieved state so I revise this comment to only correspond to tests that don't use the current state.

BenHenning (changes requested):
Thanks @MaskedCarrot. Took another pass--I think the PR is looking quite good. Just had a variety of suggestions, nits, and some requests for improved testing. PTAL.

BenHenning:
Also @MaskedCarrot please sync with develop & resolve conflicts.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
MaskedCarrot: This does not work on robolectric because we can't move past the drag and drop state of the exploration on robolectric yet. 

Also the logic for deleting progress is being tested in exploration activity test by the testing the altert dialog boxes.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: There will be no empty checkpoint saved here, the most recent checkpoint that is saved in this case is when the user comes to the terminal state (i.e. the pending state changes). After that when the user clicks on the "Return to Topic" button, the saved progress for this lesson will be deleted.

MaskedCarrot: We are verifying that the progress is deleted successfully on clicking the "Return to topic" button in StateFragmentTest

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: added an assert statement to ensure the solution is not revealed. 
Also, checking that the index type is `last revealed hint index`  instead of verifying that it is not `show solution` also ensures that the index type case is not some other value (for example index type not set)

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandler.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
MaskedCarrot: simplified

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
MaskedCarrot: done while resolving conflicts

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
```diff
@@ -86,31 +86,23 @@ class HintHandlerImpl private constructor(
...
-          lastRevealedHintIndex = helpIndex.nextAvailableHintIndex - 1
-          latestAvailableHintIndex = helpIndex.nextAvailableHintIndex
+        NEXT_AVAILABLE_HINT_INDEX, LATEST_REVEALED_HINT_INDEX -> {
+          if (helpIndex.indexTypeCase == NEXT_AVAILABLE_HINT_INDEX) {
+            lastRevealedHintIndex = helpIndex.nextAvailableHintIndex - 1
+            latestAvailableHintIndex = helpIndex.nextAvailableHintIndex
+          } else {
+            lastRevealedHintIndex = helpIndex.latestRevealedHintIndex
+            latestAvailableHintIndex = helpIndex.latestRevealedHintIndex
+          }
```
BenHenning: Ah, I misread the code before. Honestly, it's probably more readable the other way since we can't simplify it like the solution cases.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
```diff
@@ -414,9 +424,26 @@ class HintHandlerImplTest {
...
+    reset(mockHintMonitor)
+
+    waitFor30Seconds()
 
+    // Verify that the monitor is called again (since there's a solution now available).
+    verify(mockHintMonitor).onHelpIndexChanged()
+  }
+
+  @Test
+  fun testResumeHints_stateWithHints_firstHintReveled_wait30Seconds_helpIndexHasSolution() {
```
BenHenning:
```suggestion
  fun testResumeHints_stateWithOneHint_hintRevealed_wait30Seconds_helpIndexHasSolution() {
```

Otherwise it's slightly confusing that it's jumping straight to the solution.

Also, typo in 'revealed' here & elsewhere. Please fix.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
BenHenning: Thanks for clarifying that! Sometimes it's hard to review merge commits.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
BenHenning: Ah, drag and drop is definitely an issue then. Thanks for clarifying.

And FWIW it should be possible to test AlertDialogs in Robolectric.

BenHenning (commented):
Thanks @MaskedCarrot. LGTM + 2 nits.

Also, PlatformParameterSyncUpWorkerTest seems like a potential unrelated flake (/cc @ARJUPTA). I'm following up with Arjun, but fixing the above should also retrigger the CI workflow.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerImpl.kt
MaskedCarrot: done

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerImplTest.kt
MaskedCarrot: done

BenHenning (approved):
Thanks @MaskedCarrot. LGTM!

PR #3595, opened 2021-07-27
Fix #3490: Add resume lesson activity

MaskedCarrot:
## Explanation

**Please do not review, this PR contains a few unnecessary changes**, but please take a look at this comment https://github.com/oppia/oppia-android/pull/3595#issuecomment-897231167

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3670, opened 2021-08-12
Fix #3490 Add resume lesson activity [Blocked on #3488]

### LINKED ISSUE #3490: Create activity to resume explorations
Create an activity, its presenter and layout file to resume explorations.

Target | Date
--|--
**PR creation:** | 2021-08-02
**PR completion:** | 2021-08-06

MaskedCarrot:
## Explanation

Fixes #3490 

## Checklist

#### ResumeLessonActivityTest
Screenshot from 2021-08-20 14-57-26

#### ResumeLessonFragmentTest
Screenshot from 2021-08-20 14-56-19
Screenshot from 2021-08-20 15-11-46


MaskedCarrot:
Comment in the previous PR: #3595 

> Thanks @MaskedCarrot. Took a full pass on my codeowner files & related. Besides the comments I've left, I have a few other questions:
> 
> 1. Is the UI implementation finalized? The text spacing & background colors in your screenshots seem incomplete & inconsistent with other screens in the app.


I created the mobile portrait UI with the help of the mock, mobile landscape and tablet mocks are not available in the PRD so i created those by looking at the other screens in the app (primarily topic info fragment because that is similar to this fragment as it has lessonsThumbanilImageVeiw. )

Also, the image in the PR description may look a bit stretched because the screenshot was too large, so I changed the height and width of the image in markdown to make the image smaller.

## Mock
Screenshot from 2021-07-28 14-02-14

Note- The overflow menu in the above mock was removed as per the discussions in PRD doc comments.




@BenHenning  PTAL

MaskedCarrot:
The failing tests are

testResumeLessonFragment_lessonThumbnailIsDisplayed()
testResumeLessonFragment_onTablet_lessonThumbnailIsDisplayed()
testResumeLessonFragment_onTablet_configChange_lessonThumbnailIsDisplayed()

I tried this in the following two ways:

changing the XML to show the drawable with src.
changing the file LessonThumbnailImageView to force it to show this drawable.
All tests pass for the first case but all of them fail for the second.

Based on the above observation I think this might be happening because the image does not load in time, to verify this I tried adding multiple testCoroutineDispatchers.runCurrent() but the tests still fail.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -143,6 +144,30 @@ class TopicController @Inject constructor(
     )
   }
 
+  /**
+   * Retrieves a chapter given a topic ID, story ID and exploration ID.
```
BenHenning:
```suggestion
   * Retrieves a chapter given a topic ID, story ID, and exploration ID.
```

Nit on Oxford comma.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -143,6 +144,30 @@ class TopicController @Inject constructor(
...
+   *
+   * @param topicId the ID corresponding to the topic which contains this story
+   * @param storyId the ID corresponding to the story which needs to be returned
+   * @param explorationId the ID corresponding to the exploration which needs to be returned
+   * @return a [DataProvider] for [ChapterSummary]
+   */
+  fun retrieveChapter(
```
BenHenning, line 160: Please add tests for this new method.

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+    private const val RESUME_LESSON_FRAGMENT_STORY_ID_KEY =
+      "ResumeExplorationFragmentPresenter.resume_exploration_fragment_story_id"
+    private const val RESUME_LESSON_FRAGMENT_EXPLORATION_ID_KEY =
+      "ResumeExplorationFragmentPresenter.resume_Lesson_fragment_exploration_id"
+    private const val RESUME_LESSON_FRAGMENT_BACKFLOW_SCREEN_KEY =
+      "ResumeLessonFragmentPresenter.resume_lesson_fragment_backflow_screen"
+
+    /**
+     * Creates new instance of [ResumeLessonFragment].
+     * @param internalProfileId is used by the ResumeLessonFragment to retrieve saved checkpoint
```
BenHenning, line 34: Nit: a blank doc line should exist between the at-clause block and the summary fragment. Ditto elsewhere in the PR.

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import org.oppia.android.app.fragment.InjectableFragment
+import javax.inject.Inject
+
+/** Fragment that allows the user to resume a saved exploration. */
+class ResumeLessonFragment :
+  InjectableFragment() {
```
BenHenning: Should be on one line (in general, prefer to keep all statements on one line unless they need to be wrapped.

Ditto elsewhere in the PR (including for ResumeLessonActivity).

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonActivity.kt
```diff
@@ -0,0 +1,105 @@
...
+      "ResumeLessonActivity.topic_id"
+    private const val RESUME_LESSON_ACTIVITY_STORY_ID_ARGUMENT_KEY =
+      "ResumeLessonActivity.story_id"
+    private const val RESUME_LESSON_ACTIVITY_EXPLORATION_ID_ARGUMENT_KEY =
+      "ResumeLessonActivity.exploration_id"
+    private const val RESUME_LESSON_ACTIVITY_BACKFLOW_SCREEN_KEY =
+      "ResumeLessonActivity.backflow_screen"
+
+    /**
+     *  Returns a new [Intent] to route to [ResumeLessonActivity] for a specified exploration.
```
BenHenning: Nit: extra space before ``Returns``

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.databinding.ObservableField
+import org.oppia.android.app.fragment.FragmentScope
+import org.oppia.android.app.model.ChapterSummary
+import org.oppia.android.app.model.ExplorationCheckpoint
+import org.oppia.android.app.viewmodel.ObservableViewModel
+import org.oppia.android.util.parser.html.ExplorationHtmlParserEntityType
+import javax.inject.Inject
+
+/**
+ * The viewModel that providers [ChapterSummary] and [ExplorationCheckpoint] for the exploration
```
BenHenning:
```suggestion
 * The viewModel that provides a [ChapterSummary] and [ExplorationCheckpoint] for the exploration
```

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+@FragmentScope
+class ResumeLessonViewModel @Inject constructor(
+  @ExplorationHtmlParserEntityType val entityType: String
+) : ObservableViewModel() {
+
+  /** The chapter summary for the exploration that may be resumed. */
+  val chapterSummary = ObservableField(ChapterSummary.getDefaultInstance())
+
+  /** The [ExplorationCheckpoint] that may be used to resume exploration. */
```
BenHenning:
```suggestion
  /** The [ExplorationCheckpoint] that may be used to resume the exploration. */
```

BenHenning (changes requested):
Thanks @MaskedCarrot. Looked at all of my codeowner files & some of the related ones. Just had a few comments, but otherwise the PR is pretty close to LGTM from my perspective.

BenHenning:
> The failing tests are
> 
> testResumeLessonFragment_lessonThumbnailIsDisplayed()
> testResumeLessonFragment_onTablet_lessonThumbnailIsDisplayed()
> testResumeLessonFragment_onTablet_configChange_lessonThumbnailIsDisplayed()
> 
> I tried this in the following two ways:
> 
> changing the XML to show the drawable with src.
> changing the file LessonThumbnailImageView to force it to show this drawable.
> All tests pass for the first case but all of them fail for the second.
> 
> Based on the above observation I think this might be happening because the image does not load in time, to verify this I tried adding multiple testCoroutineDispatchers.runCurrent() but the tests still fail.

@MaskedCarrot

I missed this earlier. As discussed in chat, I think the issue is that LessonThumbnailImageView (used by ResumeLessonFragmentTest) uses ImageLoader internally and, because that uses Glide, the image won't actually load deterministically without setting Glide up a certain way in the test (which is a bit of a pain and not hermetic or fully reliable). I actually suggest switching to TestImageLoaderModule instead of GlideImageLoaderModule and see if that just works (I think it will). If it doesn't, please let me know since that means additional verification needs to be added.

aggarwalpulkit596:
> > The failing tests are
> > testResumeLessonFragment_lessonThumbnailIsDisplayed()
> > testResumeLessonFragment_onTablet_lessonThumbnailIsDisplayed()
> > testResumeLessonFragment_onTablet_configChange_lessonThumbnailIsDisplayed()
> > I tried this in the following two ways:
> > changing the XML to show the drawable with src.
> > changing the file LessonThumbnailImageView to force it to show this drawable.
> > All tests pass for the first case but all of them fail for the second.
> > Based on the above observation I think this might be happening because the image does not load in time, to verify this I tried adding multiple testCoroutineDispatchers.runCurrent() but the tests still fail.
> 
> @MaskedCarrot
> 
> I missed this earlier. As discussed in chat, I think the issue is that LessonThumbnailImageView (used by ResumeLessonFragmentTest) uses ImageLoader internally and, because that uses Glide, the image won't actually load deterministically without setting Glide up a certain way in the test (which is a bit of a pain and not hermetic or fully reliable). I actually suggest switching to TestImageLoaderModule instead of GlideImageLoaderModule and see if that just works (I think it will). If it doesn't, please let me know since that means additional verification needs to be added.

@BenHenning I'm a bit curious here since the image is part of local drawable and even if we loading it with Glide it should be almost immediate and the test shouldn't fail because of that just as it's not failing in the rest of the places where it's being used.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: added tests for this method

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonActivity.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonViewModel.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonViewModel.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragment.kt
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragment.kt
MaskedCarrot: done

app/src/main/res/drawable/ic_start_over_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+  android:width="24dp"
+  android:height="24dp"
+  android:viewportWidth="24"
+  android:viewportHeight="24">
+  <path
+    android:fillColor="#FFFFFF"
+    android:pathData="M6,6h2v12L6,18zM9.5,12l8.5,6L18,6z"/>
+</vector>
```
rt4914: EOF is missing, add one blank line at the end.

app/src/main/res/drawable/start_over_button_background.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+  android:shape="rectangle">
+  <corners android:radius="@dimen/state_previous_next_button_radius" />
+  <stroke
+    android:width="4px"
+    android:color="@color/oppiaPrimaryDark" />
+</shape>
```
rt4914: EOF is missing, add one blank line at the end.

app/src/main/res/drawable/start_over_button_background.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+  android:shape="rectangle">
+  <corners android:radius="@dimen/state_previous_next_button_radius" />
+  <stroke
+    android:width="4px"
```
rt4914: Why is this `px`? Normally we keep all values as `dp`

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonActivity.kt
```diff
@@ -0,0 +1,103 @@
...
+      storyId,
+      explorationId,
+      backflowScreen,
+    )
+  }
+
+  // TODO(#1655): Re-restrict access to fields in tests post-Gradle.
+  companion object {
+    private const val RESUME_LESSON_ACTIVITY_PROFILE_ID_ARGUMENT_KEY =
+      "ResumeLessonActivity.profile_id"
```
rt4914: Instead of `profileId`/ `ProfileId`/`PROFILE_ID`, etc. use `internalProfileId`/`InternalProfileId`/`INTERNAL_PROFILE_ID` everywhere.

Basically when its `Int` we use `internalProfileId` as `ProfileId` is actually a model in our proto.

rt4914 (changes requested):
@MaskedCarrot PLAT at following points.
1. Refer to [mock](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/1c7d4ec6-a5f1-4462-886d-e47c1457b69e/) for current xml values.
2. File two issues (1.) for Highfi-Mobile-Landscape and (2.) Highfi Tablet and mark both of them as blocked and then send an email to Chantel asking for mock. Keep me, Ben, Sean, Pulkit in cc.
3. Please add `RTL` screenshots too for all LTR images so that we can verify that it works correctly on RTL too.

MaskedCarrot:
@rt4914 using the exact values from the mock, the button doesn't show the full text. Can you please take a look
Screenshot from 2021-08-18 00-09-14

app/src/main/res/drawable/ic_start_over_24dp.xml
MaskedCarrot: done

app/src/main/res/drawable/start_over_button_background.xml
MaskedCarrot: done

app/src/main/res/drawable/start_over_button_background.xml
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonActivity.kt
MaskedCarrot: done

rt4914:
> @rt4914 using the exact values from the mock, the button doesn't show the full text. Can you please take a look
> Screenshot from 2021-08-18 00-09-14

@MaskedCarrot Actually you will need to write attributes in a different way.
layout_width = wrap_content
layout_height = wrap_content
paddingStart = 12dp
paddingEnd = 12dp
For above two paddings you might need to add (drawablePadding) too as you have a drawable in this button.
Also make `min_height` = "48dp". This won't match correctly with the mocks but we need this to make sure buttons are correctly accessible.

MaskedCarrot:
@rt4914 PTAL, I have used a fixed layout_height of 144dp instead of wrap content, because since the text lengths in the two buttons are different using wrap content will make them have unequal lengths

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentPresenter.kt
```diff
@@ -133,6 +133,7 @@ class ResumeLessonFragmentPresenter @Inject constructor(
       fragment,
       Observer<ChapterSummary> { chapterSummary ->
         resumeLessonViewModel.chapterSummary.set(chapterSummary)
+        val t = chapterSummary
```
BenHenning: Unused val?

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -156,16 +161,18 @@ class TopicController @Inject constructor(
...
     storyId: String,
     explorationId: String
-  ): DataProvider<ChapterSummary> {
-    return dataProviders.createInMemoryDataProviderAsync(GET_CHAPTER_PROVIDER_ID) {
-      return@createInMemoryDataProviderAsync AsyncResult.success(
-        fetchChapter(
-          topicId,
-          storyId,
-          explorationId
+  ): LiveData<AsyncResult<ChapterSummary>> {
```
BenHenning: Please don't do this--domain controllers should return DataProviders. LiveData are for the UI layer, not the domain layer.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -156,16 +161,18 @@ class TopicController @Inject constructor(
...
+    return MutableLiveData(
+      try {
+        AsyncResult.success(fetchChapter(topicId, storyId, explorationId))
+      } catch (e: Exception) {
+        AsyncResult.failed(
+          ChapterNotFoundException(
+            "Chapter for exploration $explorationId not found in story $storyId and topic $topicId"
+          )
```
BenHenning, line 175: If you need a failure case then I suggest returning AsyncResult.failure() in the original DataProvider.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -95,6 +95,11 @@ class TopicController @Inject constructor(
   @LoadLessonProtosFromAssets private val loadLessonProtosFromAssets: Boolean
 ) {
 
+  /**
+   * Indicates that chapter for the specified exploration Id, story ID, and topicID was not found.
```
BenHenning:
```suggestion
   * Indicates that the chapter for the specified exploration, story, and topic ID was not found.
```

Nits on keeping consistency when referring to structure IDs.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -156,16 +161,18 @@ class TopicController @Inject constructor(
...
-      return@createInMemoryDataProviderAsync AsyncResult.success(
-        fetchChapter(
-          topicId,
-          storyId,
-          explorationId
+  ): LiveData<AsyncResult<ChapterSummary>> {
+    return MutableLiveData(
+      try {
+        AsyncResult.success(fetchChapter(topicId, storyId, explorationId))
+      } catch (e: Exception) {
```
BenHenning: We shouldn't be catching a generic exception like this--how do you know the correct one was caught? Also, it's considered an antipattern to use exceptions for control flow. Instead, return a result or introduce a mechanism to check for existence to make the determination that the chapter doesn't exist.

BenHenning (changes requested):
Thanks @MaskedCarrot. Took another pass. PTAL.

I'll need to review the TopicControllerTest changes more closely after this latest round of comments is addressed.

> > > The failing tests are
> > > testResumeLessonFragment_lessonThumbnailIsDisplayed()
> > > testResumeLessonFragment_onTablet_lessonThumbnailIsDisplayed()
> > > testResumeLessonFragment_onTablet_configChange_lessonThumbnailIsDisplayed()
> > > I tried this in the following two ways:
> > > changing the XML to show the drawable with src.
> > > changing the file LessonThumbnailImageView to force it to show this drawable.
> > > All tests pass for the first case but all of them fail for the second.
> > > Based on the above observation I think this might be happening because the image does not load in time, to verify this I tried adding multiple testCoroutineDispatchers.runCurrent() but the tests still fail.
> > 
> > 
> > @MaskedCarrot
> > I missed this earlier. As discussed in chat, I think the issue is that LessonThumbnailImageView (used by ResumeLessonFragmentTest) uses ImageLoader internally and, because that uses Glide, the image won't actually load deterministically without setting Glide up a certain way in the test (which is a bit of a pain and not hermetic or fully reliable). I actually suggest switching to TestImageLoaderModule instead of GlideImageLoaderModule and see if that just works (I think it will). If it doesn't, please let me know since that means additional verification needs to be added.
> 
> @BenHenning I'm a bit curious here since the image is part of local drawable and even if we loading it with Glide it should be almost immediate and the test shouldn't fail because of that just as it's not failing in the rest of the places where it's being used.

@aggarwalpulkit596 

I think you might have a false assumption here (that loading a local drawable via Glide is immediate--it's not). Glide's pipeline has a lot of black boxing to it including caching & potentially utilizing custom loading pipelines, even for local files (which we see with SVGs, for example). It seems Glide makes heavy use of its executors for all image loading, and getting these to behave deterministically in the test environment is tricky.

app/src/main/res/values/styles.xml
```diff
@@ -327,4 +327,39 @@
...
+    <item name="android:textColor">@color/oppiaPrimaryDark</item>
+    <item name="android:textSize">14sp</item>
+  </style>
+  <!-- Continue Lesson Button -->
+  <style name="ContinueLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
+    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_height">wrap_content</item>
+    <item name="android:drawablePadding">4dp</item>
+    <item name="android:paddingRight">12dp</item>
+    <item name="android:paddingLeft">12dp</item>
```
rt4914: Please use `paddingStart` and `paddingEnd` here and everywhere else in this PR.

Left/Right does not translate well for RTL

app/src/main/res/values/styles.xml
```diff
@@ -327,4 +327,39 @@
     <item name="android:textAllCaps">false</item>
     <item name="android:textColor">@color/oppiaPrimaryText</item>
   </style>
+  <!-- Start Over Lesson Button -->
+  <style name="StartOverLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
+    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_height">wrap_content</item>
+    <item name="android:paddingRight">12dp</item>
+    <item name="android:paddingLeft">12dp</item>
```
rt4914: Do not mention paddingTop and paddingBottom. Instead mention `minHeight = 48dp` as we need this to make this correctly accessible

app/src/main/res/values/styles.xml
```diff
@@ -327,4 +327,39 @@
...
+  </style>
+  <!-- Continue Lesson Button -->
+  <style name="ContinueLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
+    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_height">wrap_content</item>
+    <item name="android:drawablePadding">4dp</item>
+    <item name="android:paddingRight">12dp</item>
+    <item name="android:paddingLeft">12dp</item>
+    <item name="android:paddingTop">10dp</item>
+    <item name="android:paddingBottom">10dp</item>
```
rt4914: Do not mention paddingTop and paddingBottom. Instead mention `minHeight = 48dp` as we need this to make this correctly accessible

app/src/main/res/values/styles.xml
```diff
@@ -327,4 +327,39 @@
     <item name="android:textAllCaps">false</item>
     <item name="android:textColor">@color/oppiaPrimaryText</item>
   </style>
+  <!-- Start Over Lesson Button -->
+  <style name="StartOverLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
+    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_height">wrap_content</item>
+    <item name="android:paddingRight">12dp</item>
```
rt4914: Please use paddingStart and paddingEnd here and everywhere else in this PR.

Left/Right does not translate well for RTL

rt4914 (changes requested):
@MaskedCarrot PTAL

app/src/main/res/values/styles.xml
MaskedCarrot: done

app/src/main/res/values/styles.xml
MaskedCarrot: done

app/src/main/res/values/styles.xml
MaskedCarrot: done

app/src/main/res/values/styles.xml
MaskedCarrot: done

app/src/main/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentPresenter.kt
MaskedCarrot: removed

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: updated retrieveChapter

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: updated retrieveChapter

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: updated retrieveChapter

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: updated retrieveChapter

MaskedCarrot:
@BenHenning Can you please also take a look at the test `fun testRetrieveChapter_validChapter_returnsCorrectChapterSummary()` present in `topicControllerTest`. 

This test is currently failing on Bazel but passing on Gradle.

This test checks the following: 
1. chapter name 
2. chapter summary 
3. chapter thumbnail
out of the 3 the test only fails for the last one (chapter thumbnail) with the error message expected was 4 actual was 0.


[link to the test](https://github.com/oppia/oppia-android/blob/4b68d28882a7c8c5e778f38770487ecb9db943a2/domain/src/test/java/org/oppia/android/domain/topic/TopicControllerTest.kt#L416)

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: done

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -161,18 +161,20 @@ class TopicController @Inject constructor(
...
-    return MutableLiveData(
-      try {
-        AsyncResult.success(fetchChapter(topicId, storyId, explorationId))
-      } catch (e: Exception) {
+  ): DataProvider<ChapterSummary> {
+    return dataProviders.createInMemoryDataProviderAsync(GET_STORY_PROVIDER_ID) {
+      return@createInMemoryDataProviderAsync AsyncResult.success(retrieveStory(topicId, storyId))
+    }.transformAsync(GET_CHAPTER_PROVIDER_ID) { storySummary ->
+      val chapterSummary = fetchChapter(storySummary, explorationId)
+      if (chapterSummary == null)
```
BenHenning: This if/else needs braces per https://developer.android.com/kotlin/style-guide#braces.

BenHenning: A nit: but I also suggest inverting the if/else condition so that the default cause is success (since that's the expected happy path).

domain/src/test/java/org/oppia/android/domain/topic/TopicControllerTest.kt
```diff
@@ -431,19 +430,11 @@ class TopicControllerTest {
...
 
     verifyRetrieveChapterFailed()
-    assertThat(chapterSummaryResultCaptor.value.getErrorOrNull()).isInstanceOf(
-      TopicController.ChapterNotFoundException::class.java
-    )
-    assertThat(chapterSummaryResultCaptor.value.getErrorOrNull()).hasMessageThat()
-      .isEqualTo(
-        "Chapter for exploration $INVALID_EXPLORATION_ID_1 not found in story " +
-          "$INVALID_STORY_ID_1 and topic $INVALID_TOPIC_ID_1"
-      )
```
BenHenning: It's probably still worth verifying that AsyncResult includes the correct failure.

BenHenning (changes requested):
Thanks @MaskedCarrot! Just had a few comments, otherwise the PR LGTM (barring any issues when upstream PRs are merged).

> @BenHenning Can you please also take a look at the test `fun testRetrieveChapter_validChapter_returnsCorrectChapterSummary()` present in `topicControllerTest`.
> 
> This test is currently failing on Bazel but passing on Gradle.
> 
> This test checks the following:
> 
> 1. chapter name
> 2. chapter summary
> 3. chapter thumbnail
>    out of the 3 the test only fails for the last one (chapter thumbnail) with the error message expected was 4 actual was 0.
> 
> [link to the test](https://github.com/oppia/oppia-android/blob/4b68d28882a7c8c5e778f38770487ecb9db943a2/domain/src/test/java/org/oppia/android/domain/topic/TopicControllerTest.kt#L416)

This is probably because TopicControllerTest uses the textproto versions of the test explorations in tests when run on Bazel & the JSON versions when run in Gradle. The Bazel versions intentionally dropped the temporary thumbnails, and in cases when we test with both types we usually don't validate thumbnails for this reason.

Is verifying the name sufficient for your test? I suggest dropping the thumbnail assertion altogether if it the test is still valid without it. If you need to verify the thumbnail, reach out to me on chat and we'll try to figure out a solution.

rt4914 (approved):
LGTM, thanks

domain/src/test/java/org/oppia/android/domain/topic/TopicControllerTest.kt
MaskedCarrot: updated the test to verify the error on failure.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
MaskedCarrot: done

vinitamurthi:
I dont think I am required here because I dont own any of these files, removing myself as a reviewer!

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
```diff
@@ -0,0 +1,233 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(
+  application = ResumeLessonActivityTest.TestApplication::class,
+  qualifiers = "port-xxhdpi"
+)
+class ResumeLessonActivityTest {
```
anandwana001, line 97: add kdoc and espresso test result screenshot

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
```diff
@@ -0,0 +1,233 @@
...
+  @Test
+  fun testResumeLessonActivity_toolbarTitleIsDisplayed() {
+    launch<ResumeLessonActivity>(createResumeLessonActivityIntent()).use {
+      onView(
+        allOf(
+          instanceOf(TextView::class.java),
+          withParent(withId(R.id.resume_lesson_activity_toolbar))
+        )
+      )
+        .check(matches(withText("Resume Lesson")))
```
anandwana001:
```suggestion
      ).check(matches(withText("Resume Lesson")))
```

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
```diff
@@ -0,0 +1,233 @@
...
+    launch<ResumeLessonActivity>(createResumeLessonActivityIntent()).use {
+      testCoroutineDispatchers.runCurrent()
+      onView(isRoot()).perform(orientationLandscape())
+      onView(
+        allOf(
+          instanceOf(TextView::class.java),
+          withParent(withId(R.id.resume_lesson_activity_toolbar))
+        )
+      )
+        .check(matches(withText("Resume Lesson")))
```
anandwana001: ditto

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
```diff
@@ -0,0 +1,260 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(
+  application = ResumeLessonFragmentTest.TestApplication::class,
+  qualifiers = "port-xxhdpi"
+)
+class ResumeLessonFragmentTest {
```
anandwana001, line 91: add kdoc and espresso result screenshot for mobile and tablet

anandwana001 (changes requested):
left few points.

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
MaskedCarrot: done

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
```diff
@@ -119,7 +121,7 @@ class ResumeLessonFragmentTest {
   private fun setUpTestApplicationComponent() {
     ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
   }
-
+  @Ignore
```
BenHenning: Newline before this (newlines generally always separate class members except fields which are optional).

BenHenning (commented):
Thanks @MaskedCarrot. Just 1 nit. PTAL.

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
```diff
@@ -0,0 +1,262 @@
...
+
+  @After
+  fun tearDown() {
+    testCoroutineDispatchers.unregisterIdlingResource()
+  }
+
+  private fun setUpTestApplicationComponent() {
+    ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
+  }
+  @Ignore
```
anandwana001: Why are we ignoring this test case?
Add reason in format `@Ignore TODO ... ` and add TODO of issue which will fix this.

anandwana001 (changes requested):

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
MaskedCarrot: ignore is not required, this test is for tablets and it fails on mobile, so I added this while testing the tests on espresso

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
MaskedCarrot: ignore is not required, this test is for tablets and it fails on mobile, so I added this while testing the tests on espresso

MaskedCarrot: removed this

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonFragmentTest.kt
MaskedCarrot: removed ignore

BenHenning (commented):
Thanks @MaskedCarrot. LGTM, but waiting on approving until it's based on develop with passing CI. Please assign back at that point.

anandwana001 (approved):
Approving my last suggestion got fixed. LGTM

BenHenning (approved):
Thanks @MaskedCarrot. LGTM!

BenHenning:
Merging since everyone has approved, CI is passing, branch is up-to-date & looks good, and no conversation threads are open anymore.

PR #3705, opened 2021-08-19
Fix #3396 and #3397: Implement Show All Hints and Solution Feature

### LINKED ISSUE #3396: Introduce Debug Implementation of HintHandler
Implement Debug Implementation of HintHandler. We will extend the `HintHandler` interface created in #3312.

**Target PR date:** 2021-08-02
**Target completion date:** 2021-08-08

### LINKED ISSUE #3397: Implement logic to show all hints and solution
Implement logic to show all hints and solutions on toggling on of the `Hint switch`

**Target PR date:** 2021-08-05
**Target completion date:** 2021-08-12

yashraj-01:
## Explanation

Fixes #3396.
Fixes #3397. 
Implemented feature to show all hints and solution.

**Screenshots**

- When the feature is disabled:

https://user-images.githubusercontent.com/63469151/129990302-f4e922f2-342c-4289-aa19-53075a5d7f47.mp4

- When the feature is enabled:

https://user-images.githubusercontent.com/63469151/129990420-1e543d21-3f8b-4bd7-8ca8-64a8cf2657f2.mp4

- Espresso tests passing for `DeveloperOptionsFragmentTest`
Screenshot from 2021-08-19 06-26-59

## Checklist


yashraj-01:
Some of the test names in `HintHandlerDebugImplTest` exceed the 100 max char limit. I have temporarily exempted them from ktlint check. I request the reviewers to suggest proper names for the test so that the name fits under 100 char, if possible. Thanks.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,2012 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Tests for [HintHandlerDebugImpl]. */
+@Suppress("FunctionName")
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = HintHandlerDebugImplTest.TestApplication::class)
+class HintHandlerDebugImplTest {
```
BenHenning, line 47: While I like the idea of being especially thorough when testing APIs, given it's a debug API I think we can reduce the complexity of these tests. Rather than duplicating the entirety of the prod implementation, could we maybe instead have a few select scenarios where we specifically want to test debug on/off? All other prod behaviors let's just rely on the prod implementation to be correct (i.e. I think it's okay to assume the debug impl is delegating to the prod impl for the sake of simplifying this testing surface).

BenHenning (changes requested):
Thanks @yashraj-01. Haven't taken a full review yet, but I'm going to pull the HintsAndSolutionProdModule rename into the base PR to simplify this a bit, and I had 1 high-level comment to reduce the testing complexity. PTAL.

Please assign back once both are addressed.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: I think this makes sense. I have now removed prod behaviour tests and kept a few of them which we are testing for on/off cases. PTAL. Thanks.

yashraj-01:
> Thanks @yashraj-01. Haven't taken a full review yet, but I'm going to pull the HintsAndSolutionProdModule rename into the base PR to simplify this a bit, and I had 1 high-level comment to reduce the testing complexity. PTAL.
> 
> Please assign back once both are addressed.

@BenHenning I have updated the PR with the latest develop and resolved all merge conflicts. You can take a pass on the PR now.

app/src/main/java/org/oppia/android/app/application/ApplicationComponent.kt
```diff
@@ -77,7 +77,7 @@ import javax.inject.Singleton
     ExpirationMetaDataRetrieverModule::class, RatioInputModule::class,
     UncaughtExceptionLoggerModule::class, ApplicationStartupListenerModule::class,
     LogUploadWorkerModule::class, WorkManagerConfigurationModule::class,
-    HintsAndSolutionConfigModule::class, HintsAndSolutionProdModule::class,
+    HintsAndSolutionConfigModule::class, HintsAndSolutionDebugModule::class,
```
Sarthak2601, line 85: Update the Kdoc above as this module has to be replaced with its corresponding prod module in production build.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
```diff
@@ -24,7 +24,9 @@ internal class QuestionAssessmentProgress {
   internal var questionSessionMetrics: MutableList<QuestionSessionMetrics> = mutableListOf()
 
   /** Initialize the assessment with the specified list of questions. */
-  internal fun initialize(questionsList: List<Question>) {
+  internal fun initialize(
```
Sarthak2601: I think it was fine earlier.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
```diff
@@ -200,7 +208,8 @@ class HintHandlerImpl private constructor(
     solutionIsRevealed = false
   }
 
-  private fun computeCurrentHelpIndex(): HelpIndex {
+  /** Computes the current [HelpIndex]. */
```
Sarthak2601:
```suggestion
  /** Returns the current [HelpIndex]. */
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
```diff
@@ -139,10 +145,12 @@ class HintHandlerImpl private constructor(
     computeCurrentHelpIndex()
   }
 
-  private fun cancelPendingTasks() {
-    // Cancel any potential pending hints by advancing the sequence number. Note that this isn't
-    // reset to 0 to ensure that all previous hint tasks are cancelled, and new tasks can be
-    // scheduled without overlapping with past sequence numbers.
+  /**
+   * Cancel any potential pending hints by advancing the sequence number. Note that this isn't
```
Sarthak2601: small nit: Cancels instead of Cancel

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+      viewHint(helpIndex.nextAvailableHintIndex)
+    }
+    if (hintHandlerProdImpl.pendingState.hasSolution()) {
+      hintHandlerProdImpl.showHintImmediately(hintHandlerProdImpl.getNextHelpIndexToReveal())
+      viewSolution()
+    }
+  }
+
+  /** Debug implementation of [HintHandler.Factory]. */
+  class FactoryImpl @Inject constructor(
```
Sarthak2601: nit: `DebugFactoryImpl`

Sarthak2601 (changes requested):
Thanks @yashraj-01! Left some thoughts.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: I have renamed it to `FactoryDebugImpl` to be consistent with the naming of prod and debug implementations.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
yashraj-01: Done.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
yashraj-01: Done.

domain/src/main/java/org/oppia/android/domain/question/QuestionAssessmentProgress.kt
yashraj-01: This is weird. I don't remember making this change. Anyways, I have changed it back.

app/src/main/java/org/oppia/android/app/application/ApplicationComponent.kt
yashraj-01: Done.

rt4914 (approved):
Implementation wise looks good. Approved for code-owner files.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,795 @@
...
+
+    hintHandler.navigateToPreviousState()
+
+    // The monitor should not be called since the user navigated away from the pending state and all
+    // helps are already revealed and there is nothing to monitor now.
+    verifyNoMoreInteractions(mockHintMonitor)
+  }
+
+  @Test
+  fun testNavigateToPreviousState_showAllHelpsDisabled_multipleTimes_pendingHint_wait60Sec_monitorNotCalled() { // ktlint-disable max-line-length
```
anandwana001: let's try to abbreviate the parts and no print the ktlint check.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,795 @@
...
+
+    // The monitor should not be called since the pending state isn't visible and all helps are
+    // already revealed and there is nothing to monitor now.
+    verifyNoMoreInteractions(mockHintMonitor)
+  }
+
+  /* Tests for navigateBackToLatestPendingState */
+
+  @Test
+  fun testNavigateBackToLatestPendingState_showAllHelpsEnabled_fromPreviousState_monitorNotCalled() { // ktlint-disable max-line-length
```
anandwana001: let's try to abbreviate the parts and no print the ktlint check.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,795 @@
...
+
+    hintHandler.navigateBackToLatestPendingState()
+
+    // The monitor should not be called after returning to the pending state as all helps are
+    // already revealed and there is nothing to monitor now.
+    verifyNoMoreInteractions(mockHintMonitor)
+  }
+
+  @Test
+  fun testNavigateBackToLatestPendingState_showAllHelpsDisabled_fromPreviousState_pendingHint_monitorNotCalled() { // ktlint-disable max-line-length
```
anandwana001: let's try to abbreviate the parts and no print the ktlint check.

anandwana001: ditto for all

anandwana001 (changes requested):
nit fixes.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: Renamed.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: Renamed.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: Renamed for all other instances as well. PTAL.

Sarthak2601 (approved):
Thanks @yashraj-01, LGTM!

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
```diff
@@ -0,0 +1,19 @@
+package org.oppia.android.domain.devoptions
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Monitor to identify whether all hints and solution should be shown. */
+@Singleton
+class ShowAllHintsAndSolutionHelper @Inject constructor() {
```
BenHenning: This isn't a helper, more like a 'Controller' per our typical structure.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
```diff
@@ -0,0 +1,19 @@
+package org.oppia.android.domain.devoptions
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Monitor to identify whether all hints and solution should be shown. */
```
BenHenning: Please detail in this KDoc that this controller is only expected to be interacted with on the main thread.

FWIW if it isn't, then you update it to be threads safe & mention that in its KDoc.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
```diff
@@ -0,0 +1,19 @@
...
+@Singleton
+class ShowAllHintsAndSolutionHelper @Inject constructor() {
+
+  private var showAllHintsAndSolution = false
+
+  /** Returns [showAllHintsAndSolution] indicating whether showing all hints and solution feature is enabled. */
```
BenHenning: This variable isn't accessible to outside users, so this documentation comment may actually be confusing. Suggest isntead:

```suggestion
  /** Returns whether showing all hints and solutions should be shown by default. */
```

Also, the way this was worded before was slightly confusing: this isn't gating hints & solutions, it's gating whether hints & solutions should be always shown. Please update variable names/method names/documentation in this PR accordingly to be clearer on this distinction.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
```diff
@@ -0,0 +1,19 @@
...
+  fun getShowAllHintsAndSolution(): Boolean = showAllHintsAndSolution
+
+  /** Sets [showAllHintsAndSolution]. */
```
BenHenning: Ditto here (KDoc could be improved).

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
```diff
@@ -0,0 +1,19 @@
...
+  fun setShowAllHintsAndSolution(isEnabled: Boolean) {
+    showAllHintsAndSolution = isEnabled
```
BenHenning: Is it possible to access the dev options menu while within an exploration? If so, this doesn't actually notify the handler impl that the state has changed (meaning there will be a delayed update). Is that okay?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+import java.util.concurrent.locks.ReentrantLock
+import javax.inject.Inject
+import kotlin.concurrent.withLock
+
+/**
+ * Debug implementation of [HintHandler] that overrides the default mechanism of hints & solutions
+ * to show all hints and solutions based on availability, bypassing the wrong answer check and
+ * scheduling of hints, if 'Show all hints and solution' functionality is enabled in the
+ * 'Developer Options Menu'. If this functionality is disabled then it will fall back to
+ * [HintHandlerProdImpl] and provide its functionalities.
```
BenHenning:
```suggestion
 * Debug implementation of [HintHandler] that conditionally always shows hints & solutions if 'Show all hints and solution' functionality is enabled in the developer options menu.
 * If this functionality is disabled then it will fall back to [HintHandlerProdImpl].
```

(Needs re-wrapping)

Some suggested changes to make the documentation possibly a bit clearer.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+class HintHandlerDebugImpl private constructor(
+  private val hintHandlerProdImpl: HintHandlerProdImpl,
+  private val showAllHintsAndSolutionHelper: ShowAllHintsAndSolutionHelper,
+  private val hintMonitor: HintHandler.HintMonitor
+) : HintHandler {
+
+  private val handlerLock = ReentrantLock()
+
+  override fun startWatchingForHintsInNewState(state: State) {
+    if (!showAllHintsAndSolutionHelper.getShowAllHintsAndSolution()) {
```
BenHenning: Should we maybe cache this value upon debug impl creation? Otherwise, we might have some whacky behavior if it's ever possible to change this state mid-exploration.

It also could largely simplify this implementation: consider querying for this state in the factory and providing either the debug or prod implementation (which would let you remove all the prod impl calls from the debug version & just perform the debug-only functionality).

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+  override fun handleWrongAnswerSubmission(wrongAnswerCount: Int) {
+    if (!showAllHintsAndSolutionHelper.getShowAllHintsAndSolution()) {
+      hintHandlerProdImpl.handleWrongAnswerSubmission(wrongAnswerCount)
+    }
+  }
+
+  override fun viewHint(hintIndex: Int) {
+    if (!showAllHintsAndSolutionHelper.getShowAllHintsAndSolution()) {
+      hintHandlerProdImpl.viewHint(hintIndex)
+    } else {
```
BenHenning: Isn't this case invalid? I thought we were by default jumping straight to the EVERYTHING_REVEALED state?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
```diff
@@ -62,9 +62,15 @@ class HintHandlerImpl private constructor(
   private val backgroundCoroutineScope = CoroutineScope(backgroundCoroutineDispatcher)
 
   private var trackedWrongAnswerCount = 0
-  private lateinit var pendingState: State
+
+  /** Refers to the current pending state. */
+  internal lateinit var pendingState: State
```
BenHenning: Please don't expose the internal state of this class. I don't think you'll need to with the factory suggestion I made above, but we should otherwise try to design around punching holes into class implementations.

Ditto for other 'internal' visibility changes below.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
```diff
@@ -301,7 +310,8 @@ class HintHandlerImpl private constructor(
           latestAvailableHintIndex = nextHelpIndexToShow.nextAvailableHintIndex
         }
         SHOW_SOLUTION -> solutionIsAvailable = true
-        else -> {} // Nothing else to do.
+        else -> {
+        } // Nothing else to do.
```
BenHenning: Why this change?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    hintHandlerProdImpl.pendingState.interaction.hintList.forEach { _ ->
+      val helpIndex = hintHandlerProdImpl.getNextHelpIndexToReveal()
+      hintHandlerProdImpl.showHintImmediately(helpIndex)
+      viewHint(helpIndex.nextAvailableHintIndex)
+    }
+    if (hintHandlerProdImpl.pendingState.hasSolution()) {
+      hintHandlerProdImpl.showHintImmediately(hintHandlerProdImpl.getNextHelpIndexToReveal())
+      viewSolution()
+    }
```
BenHenning: Do we actually need all this? I'm wondering if we can just default the implementation to returning an EVERYTHING_REVEALED HelpIndex & calling the monitor callback just once when we initialize for a new state.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModule.kt
```diff
@@ -0,0 +1,13 @@
+package org.oppia.android.domain.hintsandsolution
+
+import dagger.Binds
+import dagger.Module
+
+/** Module for providing hints & solution related dependencies in debug build of the app. */
```
BenHenning:
```suggestion
/** Module for providing debug-only hints & solution related dependencies. */
```

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionProdModule.kt
```diff
@@ -3,9 +3,9 @@ package org.oppia.android.domain.hintsandsolution
 import dagger.Binds
 import dagger.Module
 
-/** Production module for providing hints & solution related dependencies. */
+/** Module for providing hints & solution related dependencies in production build of the app. */
```
BenHenning: Ditto here.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModuleTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.robolectric.annotation.LooperMode
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Tests for [HintsAndSolutionDebugModule]. */
+@Suppress("FunctionName")
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = HintsAndSolutionDebugModuleTest.TestApplication::class)
+class HintsAndSolutionDebugModuleTest {
```
BenHenning, line 31: One way you could significantly simplify your debug impl testing surface (if we go with the factory approach) is to actually do the factory selection in the module and test that here. Then, that becomes as simple as:
- Test debug impl provided if setting is off
- Test prod impl is provided if setting is on

Then the debug testing surface is fairly trivial since it's just making sure each of the methods behave correctly in the debug setting. We can assume that the prod impl is correct when the setting is off since this module guarantees that scenario, so no new testing is needed.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -543,6 +543,7 @@ exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rule
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/numericinput/NumericInputRuleModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/ratioinput/RatioInputModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/textinput/TextInputRuleModule.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt"
```
BenHenning: Worth adding tests for, even if they're trivial. (We may complicate the impl in the future, so having a place to force testing considerations is generally good practice).

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
BenHenning: Thanks. Going to hold off taking a pass on this until the latest changes are addressed since there may be further simplifications possible here.

BenHenning (changes requested):
Thanks @yashraj-01. Reviewed everything except the debug impl test. Overall, I think the implementation is solid but I think there are a few things we can do to further simplify it. PTAL.

Once these are addressed, I'll take a pass on the new tests.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
yashraj-01: Renamed to `ShowAllHintsAndSolutionController`.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
yashraj-01: Updated the Kdoc as this is expected to be called on the main thread only.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
yashraj-01: Updated Kdocs. PTAL. Not sure what variables/methods are you referring to here.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
yashraj-01: Updated KDoc. PTAL. Thanks.

domain/src/main/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionHelper.kt
yashraj-01: No, it is not possible to access the menu while within an exploration.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: Done.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: We can't change the state mid-exploration as the developer options menu is not accessible while in an exploration. However, I think we can go with the factory approach you suggested. It will simplify the implementation even further.

yashraj-01: Done. PTAL. Thanks.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: Following your suggestion to return EVERYTHING_REVEALED state, I have removed this case. Thanks for the suggestion.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
yashraj-01: Replaced all internal with private visibility except for `State.offersHelp` (we are using it in debug handler) which I think is fine to use as it is more general rather than specific to prod handler. WDYT?

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
yashraj-01: Well, this is the second instance in this PR that there is a change that I didn't commit (at least not knowingly). I think this was changed when I formatted the code using Ctrl + Alt + L in Android Studio. Anyways, I have reverted this change.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: Ah! Thanks for this @BenHenning. This clarified a lot of misconceptions I had. I have updated the approach to return directly the EVERYTHING_REVEALED HelpIndex state and default instance if there is no help in the particular state. PTAL. Thanks.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModule.kt
yashraj-01: Done.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionProdModule.kt
yashraj-01: KDoc updated.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModuleTest.kt
yashraj-01: I have gone with the factory approach but I wasn't able to simplify the tests as I had to create the handler in each test rather than creating it once in the `setUp` function. This is because now the handler implementation is selected based on the setting so first, we have to specify whether the setting is enabled or disabled and then create the hint handler. Is this correct or am I missing something here?

scripts/assets/test_file_exemptions.textproto
yashraj-01: Added test for `ShowAllHintsAndSolutionController`. PTAL. Thanks.

yashraj-01:
@BenHenning I have addressed all your comments. PTAL. Thanks,

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
```diff
@@ -2,125 +2,69 @@ package org.oppia.android.domain.hintsandsolution
...
   /** Debug implementation of [HintHandler.Factory]. */
   class FactoryDebugImpl @Inject constructor(
     private val hintHandlerProdImplFactory: HintHandlerProdImpl.FactoryProdImpl,
-    private val showAllHintsAndSolutionHelper: ShowAllHintsAndSolutionHelper
+    private val showAllHintsAndSolutionController: ShowAllHintsAndSolutionController
   ) : HintHandler.Factory {
     override fun create(hintMonitor: HintHandler.HintMonitor): HintHandler {
-      val hintHandlerProdImpl: HintHandlerProdImpl =
+      return if (!showAllHintsAndSolutionController.getShowAllHintsAndSolution()) {
         hintHandlerProdImplFactory.create(hintMonitor) as HintHandlerProdImpl
```
BenHenning: Is this case needed?

domain/src/test/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionControllerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+  @Inject
+  lateinit var showAllHintsAndSolutionController: ShowAllHintsAndSolutionController
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testGetShowAllHintsAndSolution_returnsFalse() {
```
BenHenning:
```suggestion
  fun testGetShowAllHintsAndSolution_initalState_returnsFalse() {
```

The condition under test is missing I think.

domain/src/test/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionControllerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    val showAllHintsAndSolution = showAllHintsAndSolutionController.getShowAllHintsAndSolution()
+    assertThat(showAllHintsAndSolution).isFalse()
+  }
+
+  @Test
+  fun testSetShowAllHintsAndSolution_true_returnsTrue() {
```
BenHenning:
```suggestion
  fun testGetShowAllHintsAndSolution_setToShowAll_returnsTrue() {
```

Might be a bit clearer framing from the perspective of testing the getter.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerProdImpl.kt
BenHenning: I think that's reasonable--thanks.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModuleTest.kt
BenHenning: I think you can simplify it by:
- Updating the debug module to do the if check & return either the debug or prod version
- Updating these tests correspondingly to verify the correct impl is returned based on the controller state
- Updating the hint handler debug impl tests to directly inject the impl factory (similar to what the prod version does) so that it doesn't even go through the module; This lets us focus those tests just on the debug impl behaviors.

Does that help clarify?

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
BenHenning: Going to defer partly until one more pass--see my other comment for a suggestion on how to further simplify. I think you can effectively remove all of the prod impl test cases if we go with the module approach (& effectively replace them with just a single module test to verify that the prod version is provided when we expect it to be).

I took one pass on this test from the perspective of just the debug impl.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,814 @@
...
+
+    val state = expWithNoHintsOrSolution.getInitialState()
+
+    hintHandler.startWatchingForHintsInNewState(state)
+
+    verify(mockHintMonitor).onHelpIndexChanged()
+  }
+
+  @Test
+  fun testStartWatchingForHints_showAllHelpsEnabled_stateWithoutHints_callsMonitor() {
```
BenHenning: Here & elsewhere: tests are generally organize into three distinct sections: arrange, act, assert. It often helps to keep code in each of these sections tightly together (new blank newlines) and each of the sections separate by 1 blank newline. For this reason you'll see in many of our tests a clear three sections in each test.

I suggest following that structure here for clarity & consistency.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
```diff
@@ -0,0 +1,814 @@
...
+
+    assertThat(helpIndex).isEqualTo(
+      HelpIndex.newBuilder().apply {
+        nextAvailableHintIndex = 0
+      }.build()
+    )
+  }
+
+  @Test
+  fun testGetCurrentHelpIndex_showAllHelpsEnabled_initialState_allHelpsAreRevealed() {
```
BenHenning:
```suggestion
  fun testGetCurrentHelpIndex_showAllHelpsEnabled_initialState_everythingIsRevealed() {
```

BenHenning: Suggest also adding a test for the other case: should be empty for an empty state.

BenHenning (changes requested):
Thanks @yashraj-01! Per CI it seems HintsAndSolutionDebugModuleTest is failing. Please be sure to double check all changed tests are passing locally before sending the PR back for review to save time waiting on CI.

Overall, the PR looks pretty close to LGTM but I had a few more suggestions & nits. PTAL.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: Well we are providing prod impl when the debug setting is disabled and debug impl when it is enabled. Isn't this correct?

yashraj-01: Instead of passing prod impl in debug impl, I have checked in the factory itself that which impl we should be using based on this check.

yashraj-01:
> Thanks @yashraj-01! Per CI it seems HintsAndSolutionDebugModuleTest is failing. Please be sure to double check all changed tests are passing locally before sending the PR back for review to save time waiting on CI.
> 
> Overall, the PR looks pretty close to LGTM but I had a few more suggestions & nits. PTAL.

Yes. I just saw that the test was failing. It was due to the change in `FactoryDebugImpl`. I'll correct the tests in a few minutes. Apologies for the delay.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModuleTest.kt
yashraj-01: If we do the if check in module then wouldn't it provide the implementation on compile time which will always be prod impl. I think we need to do the if check in the debug factory to provide correct implementation at runtime? WDYT? Correct me if I'm wrong.

yashraj-01: However, I'll try this module approach and let you know if we get the desired result

yashraj-01: Unfortunately, this didn't work. It is only providing prod impl as I said before. Is this the correct way to do it? I also tried injecting the controller but that was throwing errors.
Screenshot from 2021-08-21 15-15-56
Should we go with the factory approach instead?

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: test renamed and new test added.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
yashraj-01: Done. PTAL. Thanks.

domain/src/test/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionControllerTest.kt
yashraj-01: Done.

domain/src/test/java/org/oppia/android/domain/devoptions/ShowAllHintsAndSolutionControllerTest.kt
yashraj-01: Done.

anandwana001 (approved):
Not codeowner but reviewed my last test function name suggestion.

yashraj-01:
> Not codeowner but reviewed my last test function name suggestion.

Thanks @anandwana001. Unassigning you now.

yashraj-01:
`org.oppia.android.domain.platformparameter.syncup.PlatformParameterSyncUpWorkerTest > testSyncUpWorker_databaseIsNotEmpty_getCorrectPlatformParameters_verifyValuesAreUpdated FAILED
    expected: SUCCEEDED
    but was : RUNNING
        at org.oppia.android.domain.platformparameter.syncup.PlatformParameterSyncUpWorkerTest.testSyncUpWorker_databaseIsNotEmpty_getCorrectPlatformParameters_verifyValuesAreUpdated(PlatformParameterSyncUpWorkerTest.kt:240)`

This test was failing even when I re-ran the tests. I don't think this PR affects this test.

yashraj-01:
There was another failure on Bazel `StateFragmentLocalTest`. This test was passing before and is passing locally. So, I have re-run the Bazel tests to see if it is broken by this PR or not (though I don't think this PR is affecting `StateFragmentLocalTest`).

yashraj-01:
> There was another failure on Bazel `StateFragmentLocalTest`. This test was passing before and is passing locally. So, I have re-run the Bazel tests to see if it is broken by this PR or not (though I don't think this PR is affecting `StateFragmentLocalTest`).

This passed after re-running.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImplTest.kt
BenHenning: Resolving since the latest test suite is fully reviewed.

domain/src/test/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModuleTest.kt
BenHenning: Dagger arranges the graph at compile time but the providers are actually called at runtime. Since the Factory isn't scoped as singleton, each time it's injected Dagger will re-call the provider.

However, you do bring up a point: we inject the factory just once in the progress controllers which means they won't change without recreating the controllers. The factory approach will work better without broader changes, and your latest test changes seem well in-line with my thoughts on reducing PR complexity.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
BenHenning: Ah, I had a typo unfortunately. I meant to ask: is this cast* needed.

BenHenning (changes requested):
Thanks @yashraj-01. Latest PR LGTM, but it seems there are still CI failures (possibly due to #3714). Could you resolve my last comment & see if a new run fixes CI? Happy to approve once CI is green.

domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintHandlerDebugImpl.kt
yashraj-01: Removed the unnecessary cast. Thanks.

yashraj-01:
> Thanks @yashraj-01. Latest PR LGTM, but it seems there are still CI failures (possibly due to #3714). Could you resolve my last comment & see if a new run fixes CI? Happy to approve once CI is green.

The CI checks have passed except the failing PlatformParameter test.

BenHenning (approved):
Thanks @yashraj-01. LGTM. Re-running the Gradle test to see if it'll pass this time.

PR #3433, opened 2021-07-07
Fix #3432: [A11y] Content description for rich text based images

rt4914:
## Explanation

Fix #3432: [A11y] Content description for rich text based images

This PR updates the `ImageTagHandler` file so that we can support content description for images inside rich text using `alt-with-value` attribute.

## Output
In below image notice that in Toast in place of image it reads `A boy with a red shirt and blue trousers.`

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3768, opened 2021-09-06
Fix #3765: Revert "Fix #3432: [A11y] Content description for rich text based images"

### LINKED ISSUE #3432: [A11y] Use alt-with-value from <oppia-noninteractive-image> in RichText
Currently `oppia-noninteractive-image` is used to render images in RichText.
We should use its attribute `alt-with-value` as content description which will be readable by screen readers as an alternative to images.

### LINKED ISSUE #3765: Images are not rendering in multiple choice
To repro: play through 'What is a  Fraction?' until the question starting with '"Very good! Now look at these cakes,"'.

The multiple choice bullet list is blank (the images flicker before disappearing when navigating to the state):

image

Note that this was encountered on #3764. I haven't verified whether it is encountered on develop, but it's quite likely to be. We should find the root cause (https://github.com/oppia/oppia-android/wiki/Revert-&-regression-policy#how-to-find-the-offending-commit) and then determine what the fix should be, and whether a rollback is beneficial.

anandwana001:
Reverts oppia/oppia-android#3433

Fix #3765 

Worked for me
Screenshot_20210906_180414

### DIFF
--- a/utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
+++ b/utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
@@ -113,8 +113,6 @@ class CustomHtmlContentHandler private constructor(
         customTagHandlers.getValue(tag).handleClosingTag(output)
         customTagHandlers.getValue(tag)
           .handleTag(attributes, openTagIndex, output.length, output, imageRetriever)
-        customTagHandlers.getValue(tag)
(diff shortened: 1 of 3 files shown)


rt4914 (approved):
LGTM, thanks.

BenHenning (approved):
Nice find. LGTM. Please make sure #3432 is reopened once this is merged since that'll no longer be fixed.

BenHenning:
Ah looks like it's already been reopened. Thanks @anandwana001!

PR #3764, opened 2021-09-02
[DO NOT MERGE] Temp branch combine prework and translations

BenHenning:
Combines #3687 and #3750 to enable localization work. This PR will never be merged.

PR #3754, opened 2021-09-01
Fix #3753: Fix some translatable/untranslatable strings

BenHenning:
## Explanation

Fixes #3753.

This PR:
- Marks 'Oppia' as non-translatable (since it should never be translated) & moves it to a new 'untranslatable_strings.xml' file
- Replaces all occurrences of 'Oppia' in strings to instead be parameters whereby code injects the 'Oppia' app_name string
- Moves FAQ & greeting strings to strings.xml so that they can be translated (the latter of which required moving DateTimeUtil + test to the app module; Bazel modularization for the prod utility was retained)
- Two strings that were not being used were removed
- Tests were updated accordingly, and strings have been verified
- New regex checks + tests were added to ensure strings are placed in the correct places (i.e. all translatable strings must go in strings.xml, 'Oppia' should never should up in a translatable string, non-translatable strings cannot go in the primary strings.xml file)

In passing, this also updates the regex check script to allow filepath-based matching instead of just filename, and updates the test file output for errors to be more relative.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
No UI functionality was changed. See existing home screen + FAQ screens to demonstrate existing behavior has been retained:

image

image

image

PR #3687, opened 2021-08-16
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates

PR #3787, opened 2021-09-14
Fix #3786: Fix regex checks for translated strings

### LINKED ISSUE #3786: Regex checks fail for translations
The new regex checks introduced in #3754 are incorrectly failing for translatewiki (the pattern is set up incorrectly). See: #3687

BenHenning:
## Explanation

Fixes #3786

#3754 introduced new regex checks to make sure translated strings go into the correct XML files, but the check didn't quite work for translated versions of strings.xml. However, the existing regex check didn't have support for matching against file patterns when exempting files (which is needed in this case since the check should exclude translated values directories).

The changes to the check include:
- Adding support for exempting files by one or more patterns
- Rearranging the logic to cache regex pattern compilation & file reading to avoid repeating these. This should be a nice performance improvement, but one drawback is the current implementation requires reading whole files into memory. Given we don't have any super large files in the codebase, this is probably fine. A future improvement would be to parallelize the file searching.

Note that one behavior change with the changes above is that failures will now be organized by file rather than by check (i.e. all failures for a particular file will be printed in order of that file's lines rather than in order of the checks).

See #3764 for a passing demonstration of the check (since the check is currently passing on develop; this is only an issue when trying to merge #3687--the test branch combines this & that branch). Also, new tests were added (one specifically for validating this case). See this test fail/pass without/with the regex change:

regex_test_failing
_Test failing with the old pattern_

regex_test_passing
_Test passing with the new pattern_

Finally, per the review an additional check was added to prohibited nested resource directories to simplify the set of potential matches for strings.xml (i.e. to ensure strings.xml must live immediately under a 'values-<lang>-r<region>/' directory structure).

## Essential Checklist

## For UI-specific PRs only
N/A


scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -828,6 +828,30 @@ class RegexPatternValidationCheckTest {
...
+
+    runScript()
+
+    assertThat(outContent.toString().trim()).isEqualTo(REGEX_CHECK_PASSED_OUTPUT_INDICATOR)
+  }
+
+  @Test
+  fun testFileContent_translatableString_inTranslatedPrimaryStringsFile_fileContentIsCorrect() {
+    val prohibitedContent = "<string name=\"test\">Translatable</string>"
+    tempFolder.newFolder("testfiles", "app", "src", "main", "res", "values-ar")
```
FareesHussain, line 901: can we have a test case for 

`app/src/main/res/values-ar/<str>/strings.xml` too

FareesHussain (changes requested):
LGTM, 
I guess some test cases are needed
since we don't have a test case for the paths where there is a package between 
`values` and `strings.xml`

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
BenHenning: Is this actually allowed by Android resources? I was under the impression that value files should generally live directly under their respective values folders?

BenHenning:
Hmm https://stackoverflow.com/questions/4930398/can-the-android-layout-folder-contain-subfolders lays out contradictory information. I'm actually thinking that maybe we need a check to prohibit subdirectories instead of supporting subdirectories in these types of patterns.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
BenHenning: Hmm https://stackoverflow.com/questions/4930398/can-the-android-layout-folder-contain-subfolders lays out contradictory information. I'm actually thinking that maybe we need a check to prohibit subdirectories instead of supporting subdirectories in these types of patterns.

FareesHussain: image
This is the new android studio template for new projects 
I guess it is allowed

BenHenning: @FareesHussain is that representing directory structure? That doesn't seem right at face value since themes.xml (night) means values-night/themes.xml. What's the directory structure for that folder ('project view' not 'Android view')?

FareesHussain: Ah my bad, it is values-night/themes.xml

BenHenning: New check added.

BenHenning:
New check has been added. Thanks @FareesHussain! PTAL.

FareesHussain (approved):
LGTM, thanks

BenHenning:
Thanks again @FareesHussain!

PR #3834, opened 2021-09-23

translatewiki:
Translation updates


BenHenning:
The string update here is wrong, so closing this. We're working to fix the Arabic plural translations, so this is going to be overwritten once the correct syntax is used in Translatewiki.

PR #3860, opened 2021-09-30
Fix #3800: Mitigate indefinite crashes due to an initial crash or process death in low memory cases

### LINKED ISSUE #3800: Handle low memory death crashes due to localization
#3795 will introduce a crash that occurs when the app is reconstituted after low-memory death due to the locale not being properly retained. This issue is tracking fixing that crash scenario.

Related: #2571.

BenHenning:
Fix #3800

## Explanation

This PR fixes #3800 by introducing a fallback mechanism in AppLanguageWatcherMixin to initialize the profile in AppLanguageLocaleHandler for cases when the app loses the locale that it makes an effort to keep. The two known cases when this can happen:
- App crash (particularly an issue when the app crashes in the foreground and the user reopens the app from the crash dialog)
- Low-memory process death (see #2571 for context)

The fallback mechanism isn't great since the user experience may lead to extra recreations of the activity (see video below, the fallback locale is English so non-English cases will trigger a recreation). However, this seems like the only option since there's no obvious way to pass the locale's context across thread boundaries, or to retrieve a persisted copy of the locale without blocking the main thread (& potentially triggering an ANR).

Some things to note:
- #2571 proposes to use a bundle-based solution; that doesn't work here since both the saved instance state bundle & intents aren't available at the time the locale is needed (onAttachBaseContext) since it's at the beginning of the activity's initialization
- This change sort of makes InitializeDefaultLocaleRule less critical since tests can now default to the fallback mechanism, but the rule is still more correct & provides the much-needed capability to initialize with a particular locale
- Tests can't be added for the mixin due to limitations in the current test activity setup (see the new TODO added in the mixin's test suite for additional context)

## Essential Checklist

## For UI-specific PRs only

This PR only affects crash scenarios, but the videos are really helpful in demonstrating the fixes.

Simulated in-app foreground crash without the fix:
https://user-images.githubusercontent.com/12983742/135440867-0f25696d-7db3-4ce4-af7b-f2ddee6717e0.mp4

Simulated in-app foreground crash with the fix:

https://user-images.githubusercontent.com/12983742/135440914-1cdb7768-1499-49c6-8b2b-2ed81f9547af.mp4

Low-memory death scenarios. Notice that when the app closes/crashes, it's being triggered by a simulated low memory death. Note also that the app is killed after it's backgrounded to simulate a more realistic scenario where the system kills one app to free up resources for another.

Without fix:

https://user-images.githubusercontent.com/12983742/135441142-9c2a6d36-8386-4c1f-8a96-a5ee2e820a55.mp4

With the fix:

https://user-images.githubusercontent.com/12983742/135441212-e3a7fd4c-5229-4069-86d0-e28a73103ea3.mp4

(note that some of the 'without' fixes might look like they partially work since, I think, the system can choose to reopen the app from launcher rather than at the specific activity in certain circumstances).


BenHenning:
Reopening to force CI to fully restart (to avoid false passing CI check for base branch since it wasn't set when CI was kicked off).

BenHenning:
@rt4914 PTAL for codeowners.

rt4914 (approved):
LGTM, thanks.

BenHenning:
Thanks @rt4914! Merging now.

PR #3891, opened 2021-10-06
Fix #3912: Added local property protobuf condition for M1 Mac

### LINKED ISSUE #3912: aarch_64 release of protoc doesn't exist [M1 mac]
Building the app in M1 mac show the following error
image

Here the gradle tries to download the osx-aarch_64 version of protoc which doesn't exist in [protoc releases](https://repo1.maven.org/maven2/com/google/protobuf/protoc/3.8.0/)

Related issues:
- https://github.com/protocolbuffers/protobuf/pull/8557
- https://github.com/grpc/grpc-java/issues/7690

FareesHussain:
Fixes #3912 

protoc-3.8.0-osx-aarch_64.exe is not available in [3.8.0 release](https://repo1.maven.org/maven2/com/google/protobuf/protoc/3.8.0/) and gives the following error on build
image

Here to fix this we the following property to local.properties (Only for M1 Mac) which uses osx release instead for the given condition when built on gradle. 
```
protobuf_platform=osx-x86_64
```

PS: This issue doesn't appear for bazel build

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -3,8 +3,7 @@ apply plugin: 'com.google.protobuf'
 
 protobuf {
   protoc {
-    // To build protoc in M1 mac. For Context, see:
-    // https://github.com/grpc/grpc-java/issues/7690.
+    // To build protoc in M1 mac. For context, see: #3912.
     if (project.rootProject.hasProperty('protobuf_platform')) {
       artifact = "com.google.protobuf:protoc:3.8.0:${project.rootProject.property("protobuf_platform")}"
     } else {
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -3,7 +3,11 @@ apply plugin: 'com.google.protobuf'
 
 protobuf {
   protoc {
-    artifact = 'com.google.protobuf:protoc:3.8.0'
+    if (project.rootProject.hasProperty('protobuf_platform')) { // Only for M1 Mac.
+      artifact = "com.google.protobuf:protoc:3.8.0:${project.rootProject.property("protobuf_platform")}"
+    } else {
+      artifact = "com.google.protobuf:protoc:3.8.0"
+    }
   }
   generateProtoTasks {
     all().each { task ->
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -3,7 +3,9 @@ apply plugin: 'com.google.protobuf'
 
 protobuf {
   protoc {
-    if (project.rootProject.hasProperty('protobuf_platform')) { // Only for M1 Mac.
+    // To build protoc in M1 mac. For Context, see:
+    // https://github.com/grpc/grpc-java/issues/7690.
+    if (project.rootProject.hasProperty('protobuf_platform')) {
       artifact = "com.google.protobuf:protoc:3.8.0:${project.rootProject.property("protobuf_platform")}"
     } else {
       artifact = "com.google.protobuf:protoc:3.8.0"


BenHenning:
Thanks @FareesHussain. Nice fix. Some questions:
1. Is there a tracking issue on protoc's GitHub repo tracking this issue that we can point to?
2. If (1) doesn't exist, could you file one? It seems like a nice thing to do for a fellow open source project.
2. Could you file an issue on our repository to add references to anywhere that helped solve this, to capture the context above, and to link to the GitHub issue for protoc?
3. Could you update the comment to point to the issue from (2) as a reference?
4. Could you update the PR to mark the new issue as fixed?

I think this'll better help centralize the investigation to an issue (which is the default place where investigation contexts should live for each consulting in the future).

vinitamurthi (approved):

FareesHussain:
> Thanks @FareesHussain. Nice fix. Some questions:
> 
> 1. Is there a tracking issue on protoc's GitHub repo tracking this issue that we can point to?

Yes, there is (https://github.com/grpc/grpc-java/issues/7690). this is where i found the solution. 
> 2. If (1) doesn't exist, could you file one? It seems like a nice thing to do for a fellow open source project.
> 3. Could you file an issue on our repository to add references to anywhere that helped solve this, to capture the context above, and to link to the GitHub issue for protoc?

Done, https://github.com/oppia/oppia-android/issues/3912 
> 4. Could you update the comment to point to the issue from (2) as a reference?

Since the issue is already filed I didn't have to create these issues, https://github.com/protocolbuffers/protobuf/pull/8557 https://github.com/grpc/grpc-java/issues/7690
> 5. Could you update the PR to mark the new issue as fixed?

Done
> I think this'll better help centralize the investigation to an issue (which is the default place where investigation contexts should live for each consulting in the future).

model/build.gradle
```diff
@@ -3,7 +3,11 @@ apply plugin: 'com.google.protobuf'
 
 protobuf {
   protoc {
-    artifact = 'com.google.protobuf:protoc:3.8.0'
+    if (project.rootProject.hasProperty('protobuf_platform')) { // Only for M1 Mac.
```
BenHenning: Per my earlier comment, please add a line comment pointing to the issue for context.

BenHenning (commented):
Thanks @FareesHussain. Just one follow-up.

model/build.gradle
FareesHussain: Done

model/build.gradle
```diff
@@ -3,7 +3,13 @@ apply plugin: 'com.google.protobuf'
 
 protobuf {
   protoc {
-    artifact = 'com.google.protobuf:protoc:3.8.0'
+    // To build protoc in M1 mac. For Context, see:
+    // https://github.com/grpc/grpc-java/issues/7690.
```
BenHenning:
```suggestion
    // To build protoc in M1 mac. For context, see: #3912.
```

Prefer linking to our issue for context.

BenHenning (commented):
Thanks @FareesHussain. I think I wasn't very clear in my earlier comment, but I think the latest is a bit clearer. Please resolve & send back. PR otherwise LGTM.

model/build.gradle
FareesHussain: Done

BenHenning (approved):
Thanks again for the fast fix @FareesHussain! LGTM.

PR #3913, opened 2021-10-07

translatewiki:
Translation updates


BenHenning (approved):
This is a continuation from #3883. Verified that each of the removed strings seems correctly removed.

PR #3920, opened 2021-10-11
Fix #3919: Marquee for toolbar titles make  RTL compatible.

### LINKED ISSUE #3919: The marquee for toolbar titles has to be RTL compatible.
The marquee for toolbar titles goes left-to-right in RTL device. The actual behavior has to be from right-to-left for RTL device.

veena14cs:
## Explanation

This PR fixes #3919. This PR fixes Marquee direction for toolbar titles to from right-to-left in RTL devices.

https://user-images.githubusercontent.com/16301028/137023878-e0b8fda5-ed80-4182-b268-131608d74d68.mp4

https://user-images.githubusercontent.com/16301028/137025589-b0b9fae5-ed6b-4105-ad61-fbcf453176f3.mp4

**Device**
Nexus 5 Kitkat

https://user-images.githubusercontent.com/16301028/137274925-00458346-8e80-4fb7-b719-7d14cd09ddf0.mp4

https://user-images.githubusercontent.com/16301028/137276845-98120610-d07b-476f-82ae-6cf42dddde16.mp4

## Screenshots of Testcases after RTL 

Screenshot (15)

Screenshot (16)

Screenshot (17)

Screenshot (22)

Screenshot (27)

##  Screenshots of Testcases before RTL 

Screenshot (18)

Screenshot (19)

Screenshot (21)

Screenshot (23)

Screenshot (26)

## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)

## For UI-specific PRs only


app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -1894,12 +1939,12 @@ class ExplorationActivityTest {
     testCoroutineDispatchers.runCurrent()
   }
 
-  // TODO(#59): Remove these waits once we can ensure that the production executors are not depended on in tests.
-  //  Sleeping is really bad practice in Espresso tests, and can lead to test flakiness. It shouldn't be necessary if we
-  //  use a test executor service with a counting idle resource, but right now Gradle mixes dependencies such that both
-  //  the test and production blocking executors are being used. The latter cannot be updated to notify Espresso of any
-  //  active coroutines, so the test attempts to assert state before it's ready. This artificial delay in the Espresso
-  //  thread helps to counter that.
```
BenHenning: Mis-reformatting?

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
```diff
@@ -94,6 +96,11 @@ class ExplorationActivityPresenter @Inject constructor(
 
     binding.explorationToolbarTitle.setOnClickListener {
       binding.explorationToolbarTitle.isSelected = true
+      if (isRtlLayout(binding.explorationToolbarTitle)) {
+        binding.explorationToolbarTitle.textDirection = View.TEXT_DIRECTION_RTL
+      } else {
+        binding.explorationToolbarTitle.textDirection = View.TEXT_DIRECTION_LTR
+      }
```
BenHenning: This seems really error-prone (it's super easy to forget this when creating or changing activities). Would it be reasonable to create our own toolbar subclass that sets up its text direction automatically?

I'm also a bit confused. Why doesn't this just automatically work in the system?

BenHenning (changes requested):
Thanks @veena14cs. If this doesn't work in an emulator, how have you verified that it's correct? It was my understanding that the built-in marquee in AndroidX's toolbar doesn't properly support RTL. Is that not the case?

veena14cs:
> Thanks @veena14cs. If this doesn't work in an emulator, how have you verified that it's correct? It was my understanding that the built-in marquee in AndroidX's toolbar doesn't properly support RTL. Is that not the case?

By default it is not supporting. If the textview direction is set to RTL then the marquee direction goes right-to-left.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
veena14cs: Created custom marquee toolbar textview for setting textdirection. It is not setting the marquee direction by default when device is in RTL mode. We need to set textdirection to get correct Marquee effect.

veena14cs:
> Thanks @veena14cs. If this doesn't work in an emulator, how have you verified that it's correct? It was my understanding that the built-in marquee in AndroidX's toolbar doesn't properly support RTL. Is that not the case?

It is working on emulator now I will be adding up the screen recording. Initially I think I added it in xml or I mistook it that I couldn't observe the correct behavior on the emulator when I was toggling between the Languages and playing around the same screen without reloading it.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.core.view.ViewCompat
+
+/** The custom Textview class for toolbar with Marquee effect. */
+class MarqueeToolbarTextView @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyle: Int = android.R.attr.textViewStyle
+) : AppCompatTextView(context, attrs, defStyle), View.OnClickListener {
+
+  private var textView = this
```
BenHenning: Why is this needed?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  init {
+    textView.setOnClickListener(this)
+  }
+
+  override fun onClick(v: View?) {
+    textView.isSelected = true
+    setTextDirection(textView)
+  }
```
BenHenning: Just to clarify, do we not initially play the marquee since this only changes direction upon clicking? Also, does this result in any stuttering when the user taps on the text view?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  private fun setTextDirection(view: TextView) {
```
BenHenning: Why not use 'this' instead of passing in a TextView?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+        ViewCompat.LAYOUT_DIRECTION_RTL -> {
+          textView.textDirection = View.TEXT_DIRECTION_RTL
+        }
+        ViewCompat.LAYOUT_DIRECTION_LTR -> {
+          textView.textDirection = View.TEXT_DIRECTION_LTR
+        }
+      }
+    } else {
+      textView.textDirection = View.TEXT_DIRECTION_LOCALE
+    }
```
BenHenning: Is this something that we can do in onAttachedToWindow instead of onClick?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+      when (getLayoutDirection(textView)) {
+        ViewCompat.LAYOUT_DIRECTION_RTL -> {
+          textView.textDirection = View.TEXT_DIRECTION_RTL
+        }
+        ViewCompat.LAYOUT_DIRECTION_LTR -> {
+          textView.textDirection = View.TEXT_DIRECTION_LTR
+        }
+      }
+    } else {
+      textView.textDirection = View.TEXT_DIRECTION_LOCALE
```
BenHenning: I'm wondering if this is actually correct. I'm assuming we actually want to set the text direction using https://github.com/oppia/oppia-android/blob/f266da8dc41c7313b1e220f5eed38d05461cc078/utility/src/main/java/org/oppia/android/util/locale/OppiaLocale.kt#L241. I think we inject this in other views, so see usages of that method to see how to get access to a DisplayLocale.

Without using DisplayLocale, this might get set based on the system locale rather than the app-specified one.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+    textView.setOnClickListener(this)
+  }
+
+  override fun onClick(v: View?) {
+    textView.isSelected = true
+    setTextDirection(textView)
+  }
+
+  private fun setTextDirection(view: TextView) {
+    if (ViewCompat.isLayoutDirectionResolved(view)) {
```
BenHenning: Why is this needed? Can we just set the direction to the one we know it needs to be per my comment below?

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
```diff
@@ -0,0 +1,225 @@
...
+
+  @Test
+  fun testMarqueeToolbarTitle_ltrIsEnabled_textDirectionForLtrIsCorrect() {
+    val marqueeToolbarTitle = activityRule.scenario.runWithActivity {
+      val textView: TextView = it.findViewById(R.id.marquee_toolbar_title)
+      ViewCompat.setLayoutDirection(textView, ViewCompat.LAYOUT_DIRECTION_LTR)
+      return@runWithActivity textView
+    }
+    onView(ViewMatchers.withId(R.id.marquee_toolbar_title))
+      .perform(ViewActions.click())
```
BenHenning, line 159: Prefer directly importing withId()/click().

app/src/main/res/layout/exploration_activity.xml
```diff
@@ -38,7 +38,7 @@
           android:layout_height="wrap_content"
           android:orientation="horizontal">
 
-          <TextView
+          <org.oppia.android.app.customview.MarqueeToolbarTextView
```
BenHenning, line 41: Should we be removing the marquee-specific properties from these uses & instead specify them directly in the custom view? Otherwise it's sort of not a marquee text view if we need to set up the marquee properties each place that it needs to be used.

app/BUILD.bazel
```diff
@@ -367,6 +367,7 @@ VIEW_MODELS = [
 # keep sorted
 VIEWS_WITH_RESOURCE_IMPORTS = [
     "src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt",
+    "src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt",
```
BenHenning: It doesn't import org.oppia.android.R, so it should go in the other VIEW list.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -290,6 +293,48 @@ class ExplorationActivityTest {
...
+        TEST_EXPLORATION_ID_2,
+        shouldSavePartialProgress = false
+      )
+    )
+    val explorationToolbarTitle: TextView =
+      explorationActivityTestRule.activity.findViewById(R.id.exploration_toolbar_title)
+    ViewCompat.setLayoutDirection(explorationToolbarTitle, ViewCompat.LAYOUT_DIRECTION_LTR)
+
+    onView(withId(R.id.exploration_toolbar_title))
+      .perform(click())
```
BenHenning: Should be on one line. Ditto for all other similar lines elsewhere in the PR.

BenHenning (changes requested):
Thanks @veena14cs. Per the videos, I think this looks quite god. Just had some comments about the latest approach, but this seems like a good approach.

Can you please check this on a KitKat emulator to make sure that it's working correctly for KitKat, too? To test, you should use the Bazel build (which you can download from CI: https://github.com/oppia/oppia-android/pull/3920/checks?check_run_id=3875395434 see artifacts)--I'm not sure whether the Gradle build works on KitKat.

The Bazel failure in CI seems to be an infra flake. I restarted it.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Removed this.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Removed this method from onclick and moved to onAttachWindow().

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
veena14cs: Done.

app/src/main/res/layout/exploration_activity.xml
veena14cs: Done.

app/BUILD.bazel
veena14cs: Done.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Done.

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
veena14cs: Done.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
veena14cs: Removed this function.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -1,46 +1,59 @@
...
+  @Inject
+  lateinit var resourceHandler: AppLanguageResourceHandler
+
+  constructor(context: Context) : super(context)
+  constructor(context: Context, attrs: AttributeSet) : super(context, attrs)
+  constructor(context: Context, attrs: AttributeSet, defStyleAttr: Int) : super(
+    context,
+    attrs,
+    defStyleAttr
+  )
```
BenHenning, line 28: Can we keep the previous constructor declaration (as we do for other custom views)?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -1,46 +1,59 @@
...
-        ViewCompat.LAYOUT_DIRECTION_LTR -> {
-          textView.textDirection = View.TEXT_DIRECTION_LTR
-        }
-      }
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    val viewComponentFactory = FragmentManager.findFragment<Fragment>(this) as ViewComponentFactory
+    val viewComponent = viewComponentFactory.createViewComponent(this) as ViewComponentImpl
+    viewComponent.inject(this)
+    if (isRtl) {
```
BenHenning, line 52: Suggest just inlining this isRtl value to avoid potentially accessing it before it's actually initialized (this is a bit clearer when trying to access injectable fields, but the by lazy is once removed & a bit harder to tell when it can be called).

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -1,46 +1,59 @@
...
+      textDirection = View.TEXT_DIRECTION_RTL
     } else {
-      textView.textDirection = View.TEXT_DIRECTION_LOCALE
+      textDirection = View.TEXT_DIRECTION_LTR
```
BenHenning: Could we maybe just do something like ``textDirection = resourceHandler.getLayoutDirection()`` or do the constants differ?

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -1,46 +1,59 @@
...
+  }
 
   init {
-    textView.setOnClickListener(this)
+    setOnClickListener(this)
+    ellipsize = TextUtils.TruncateAt.MARQUEE
+    marqueeRepeatLimit = 1
+    isHorizontalFadingEdgeEnabled = true
+    isHorizontalScrollBarEnabled = true
+    setFadingEdgeLength(20)
```
BenHenning, line 40: Is the '20' here in dp or pixels? We might need to convert for it to be correct.

app/src/main/res/layout/marquee_toolbar_text_view_test_fragment.xml
```diff
@@ -1,5 +1,41 @@
 <?xml version="1.0" encoding="utf-8"?>
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-  android:id="@+id/marquee_toolbar_text_view_holder"
-  android:layout_width="match_parent"
-  android:layout_height="match_parent" />
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
```
BenHenning, line 2: We aren't using databinding to inflate this view--should this top-level declaration instead be the ConstraintLayout declared below?

app/src/main/res/layout/story_fragment.xml
```diff
@@ -39,18 +39,9 @@
           android:layout_width="wrap_content"
           android:layout_height="wrap_content"
           android:layout_marginEnd="20dp"
-          android:ellipsize="marquee"
-          android:fadingEdge="horizontal"
-          android:fadingEdgeLength="20dp"
-          android:focusable="false"
-          android:focusableInTouchMode="true"
```
BenHenning, line 46: Is this one defined programmatically in the custom view? It seems missing. Is it default true for clickable text views?

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
```diff
@@ -119,14 +128,22 @@ class MarqueeToolbarTextViewTest {
   }
 
   @Test
+  @DefineAppLanguageLocaleContext(
+    oppiaLanguageEnumId = OppiaLanguage.ARABIC_VALUE,
+    appStringIetfTag = "ar",
+    appStringAndroidLanguageId = "ar"
+  )
+  @RunOn(TestPlatform.ROBOLECTRIC) // TODO(#3840): Make this test work on Espresso & Robolectric
```
BenHenning, line 136: Is this correct? It should pass on Robolectric, but might not on Gradle. See that other tests are restricted to running on Bazel--can you see if this passes on Bazel before disabling this for Robolectric?

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
```diff
@@ -119,14 +128,22 @@ class MarqueeToolbarTextViewTest {
...
       val textView: TextView = it.findViewById(R.id.marquee_toolbar_title)
-      ViewCompat.setLayoutDirection(textView, ViewCompat.LAYOUT_DIRECTION_RTL)
       return@runWithActivity textView
     }
+    val displayLocale = appLanguageLocaleHandler.getDisplayLocale()
+    val layoutDirection = displayLocale.getLayoutDirection()
+    assertThat(layoutDirection).isEqualTo(ViewCompat.LAYOUT_DIRECTION_RTL)
     onView(ViewMatchers.withId(R.id.marquee_toolbar_title))
-      .perform(ViewActions.click())
+      .perform(click())
```
BenHenning: Please also import withId directly.

And, this can fit on one line (wrapping isn't needed).

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -155,7 +155,6 @@ class TopicFragmentTest {
     }
   }
 
-  @Test
```
BenHenning, line 154: This shouldn't be removed--it stops the test from running.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -207,36 +206,6 @@ class StoryFragmentTest {
     }
   }
 
-  @Test
-  fun testStoryFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly() {
```
BenHenning: Why remove these tests? Ditto for ExplorationActivityTest and RevisionCardActivityTest.

scripts/assets/kdoc_validity_exemptions.textproto
```diff
@@ -28,6 +28,7 @@ exempted_file_path: "app/src/main/java/org/oppia/android/app/completedstorylist/
 exempted_file_path: "app/src/main/java/org/oppia/android/app/completedstorylist/CompletedStoryListFragmentPresenter.kt"
 exempted_file_path: "app/src/main/java/org/oppia/android/app/completedstorylist/CompletedStoryListViewModel.kt"
 exempted_file_path: "app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt"
+exempted_file_path: "app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt"
```
BenHenning, line 31: Don't exempt this--prefer to add the missing documentation, instead.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.appcompat.widget.AppCompatTextView
+import androidx.core.view.ViewCompat
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import org.oppia.android.app.translation.AppLanguageResourceHandler
+import org.oppia.android.app.view.ViewComponentFactory
+import org.oppia.android.app.view.ViewComponentImpl
+import javax.inject.Inject
+
+/** The custom Textview class for toolbar with Marquee effect. */
```
BenHenning, line 16: ``Custom text view that should be configured as the child of a toolbar to display a toolbar's title with an RTL-capable marquee effect.``

Bit more specific on why & how to use this.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
BenHenning: Thanks, but I don't think that actually answers my questions about juttering/initial behavior.

I think I have the answer to the first one (that tapping is necessary), but can you clarify on the second point: is there any juttering when in RTL mode? It doesn't seem like it from the video, but the video might not show jutters.

app/src/main/java/org/oppia/android/app/customview/MarqueeToolbarTextView.kt
BenHenning: Replaced with newer comment (I don't think this was fully resolved).

app/src/sharedTest/java/org/oppia/android/app/customview/MarqueeToolbarTextViewTest.kt
BenHenning: One was done, but not the other. Replaced with a newer comment.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
BenHenning: Not fully resolved, but replaced with a newer comment.

BenHenning (changes requested):
Thanks @veena14cs. Took another full pass--PTAL.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
veena14cs: I removed it because I need to create Fragment test for these activity test as its not present in current code, so I just removed it from here.

rt4914:
For this particular PR I will have to defer to @BenHenning . Once he approves I will take a look at my code-owner files just to be sure.

veena14cs:
Closing this PR as its getting fixed in #3935

PR #3943, opened 2021-10-16
Updated .gitignore and automated setup procedure

FareesHussain:
Updated scripts for setup.sh to automatically update the local.properties for M1 mac (to use protoc using gradle)
Added misc.xml and runconfigurations.xml to .gitignore


scripts/setup.sh
```diff
@@ -25,3 +25,12 @@ bash ../oppia-android/scripts/checkstyle_download.sh
 
 # Download buf
 bash ../oppia-android/scripts/buf_download.sh
+
+# Add protobuf platform for M1 Mac
+arch_name="$(uname -m)"
+if [ "${arch_name}" = "x86_64" ]; then
+    if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
+        echo '\nprotobuf_platform=osx-x86_64' >> ../oppia-android/local.properties
```
FareesHussain: https://github.com/oppia/oppia-android/pull/3891 for context

.gitignore
```diff
@@ -7,8 +7,10 @@
 /.idea/gradle.xml
 /.idea/jarRepositories.xml
 /.idea/libraries
+/.idea/misc.xml
 /.idea/modules.xml
 /.idea/navEditor.xml
+/.idea/runConfigurations.xml
```
BenHenning: Here & above: can you remove the existing ones if we're now ignoring them?

BenHenning (changes requested):
Thanks @FareesHussain! Just had one comment--PTAL.

.gitignore
FareesHussain: I don't get it which ones do you mean?

BenHenning:
Sorry, will need to review this tomorrow.

vinitamurthi (approved):
LGTM thanks Farees!

.gitignore
BenHenning: Ah sorry. Can you please remove misc.xml and runConfigurations.xml? Now that they're being ignored, they no longer need to be tracked in VCS.

However, we need to make sure that removing these files from GitHub does *not* result in the files being deleted locally (since that might break people). I forgot how I did this last time, but I think https://stackoverflow.com/a/20241145/3689782 is what you want to do. To check: after pushing the removal to GitHub, you should verify that:
- The files are now removed on GitHub in your branch
- Checking out develop then checking out your branch in the same repository does not result in the files being removed locally

Once you confirm this & the PR is updated, I think it will look good to check in.

BenHenning (commented):
Apologies; I probably could have followed up yesterday (I didn't realize you were just looking for clarification). PTAL @FareesHussain.

scripts/setup.sh
```diff
@@ -25,3 +25,12 @@ bash ../oppia-android/scripts/checkstyle_download.sh
...
 bash ../oppia-android/scripts/buf_download.sh
+
+# Add protobuf platform for M1 Mac
+arch_name="$(uname -m)"
+if [ "${arch_name}" = "x86_64" ] && [[ "$OSTYPE" == "darwin"* ]]; then
+    if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
+        echo 'protobuf_platform=osx-x86_64' >> ../oppia-android/local.properties
+        echo 'Added protobuf platform to local properties.'
+    fi
+fi
```
anandwana001: Can we put this in a separate script and call it from here, something like `buf_m1_mac_setup.sh`

anandwana001 (changes requested):

scripts/setup.sh
FareesHussain: Sounds good, done

.gitignore
FareesHussain: Done

scripts/buf_m1_mac_setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Add protobuf platform for M1 Mac
+arch_name="$(uname -m)"
+if [ "${arch_name}" = "x86_64" ] && [[ "$OSTYPE" == "darwin"* ]]; then
+    if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
+        echo 'protobuf_platform=osx-x86_64' >> ../oppia-android/local.properties
+        echo 'Added protobuf platform to local properties.'
+    fi
+fi
```
anandwana001: add EOF

scripts/setup.sh
```diff
@@ -25,3 +25,6 @@ bash ../oppia-android/scripts/checkstyle_download.sh
 
 # Download buf
 bash ../oppia-android/scripts/buf_download.sh
+
+# Add protobuf platform for M1 Mac
+bash ../oppia-android/scripts/buf_m1_mac_setup.sh
```
anandwana001, line 30: isn't the `buf_m1_mac_setup` script is dependent on the `buf_download` script? 
We should add a NOTE in the comment here about this.

anandwana001: or can we do this here as `echo Buf setup completed`
https://github.com/oppia/oppia-android/blob/02013abd737cebd8f4931fa9b174ed35c4314eb1/scripts/buf_download.sh#L9


Need to confirm that if this change doesn't break our GitHub action!!

scripts/buf_m1_mac_setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Add protobuf platform for M1 Mac
+arch_name="$(uname -m)"
+if [ "${arch_name}" = "x86_64" ] && [[ "$OSTYPE" == "darwin"* ]]; then
+    if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
+        echo 'protobuf_platform=osx-x86_64' >> ../oppia-android/local.properties
+        echo 'Added protobuf platform to local properties.'
```
anandwana001, line 8: Do we need to echo this? 
if yes, can we say it `Buf` rather than `protobuf`?

scripts/buf_m1_mac_setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Add protobuf platform for M1 Mac
```
anandwana001:
```suggestion
# Setup protobuf platform for M1 Mac
```

scripts/setup.sh
```diff
@@ -25,3 +25,6 @@ bash ../oppia-android/scripts/checkstyle_download.sh
 
 # Download buf
 bash ../oppia-android/scripts/buf_download.sh
+
+# Add protobuf platform for M1 Mac
```
anandwana001, line 29:
```suggestion
# Add buf platform for M1 Mac
```

anandwana001 (changes requested):

FareesHussain:
@anandwana001 PTAL https://github.com/oppia/oppia-android/pull/3891 and https://github.com/oppia/oppia-android/issues/3912

scripts/buf_m1_mac_setup.sh
FareesHussain: Done

scripts/setup.sh
FareesHussain: for this and others
PTAL  https://github.com/oppia/oppia-android/pull/3891 and https://github.com/oppia/oppia-android/issues/3912
here the protobuf only for the protobuf dependency in gradle

anandwana001:
Got it, PTAL at my one last comment for comment update, else all LGTM

FareesHussain:
> Got it, PTAL at my one last comment for comment update, else all LGTM

@anandwana001 can you resolve the other comments I'm unable to understand which one (all of them are related to comment)

scripts/buf_m1_mac_setup.sh
FareesHussain: DONE

anandwana001 (approved):
LGTM

.gitignore
BenHenning: I see the files removed. Can you also confirm that checking out this branch locally does _not_ delete the files when checked out?

.gitignore
```diff
@@ -4,11 +4,14 @@
 /.idea/assetWizardSettings.xml
 /.idea/caches
 /.idea/compiler.xml
+/.idea/deploymentTargetDropDown.xml
```
BenHenning: At this point, should we just allowlist the files that *should* be tracked under .idea/? It may be simpler.

BenHenning (commented):
Thanks @FareesHussain. Just had two more comments, PTAL.

.gitignore
FareesHussain: No, it doesn't get deleted 

I checked out to develop then this branch 
These files weren't deleted

.gitignore
FareesHussain: can we do that?

I found this [ans](https://stackoverflow.com/a/6011532) as a opposite of `.gitignore`
In this case maybe we need a new .gitignore file under the .idea directory

FareesHussain: https://gist.github.com/FareesHussain/b2596f21294570ae129cf0705e225ad7
Can you confirm if this is what we need

.gitignore
BenHenning: Thanks for checking!

.gitignore
BenHenning: That seems correct, except why can't we keep it in the top-level .gitignore? E.g.:

```gitignore
.idea/*
!.idea/encodings.xml
!.idea/codeStyles/Project.xml
!.idea/codeStyleConfig.xml
```

Will that not work?

scripts/setup.sh
BenHenning: Suggest adding this context as a comment if it's important to make the association (it's more important to have that as a permanent record in code vs. a PR comment that might be hard to find in the future).

.gitignore
FareesHussain: It works and updated too

sry for the huge delay

scripts/setup.sh
BenHenning: It seems my comment is no longer relevant--the latest version of this is now using a shell script.

BenHenning (approved):
Thanks @FareesHussain. I think this looks great! Approved for codeowners.

BenHenning:
Merging since there don't seem to be any outstanding comments to address, CI checks are passing, and this should be a big help with contributor PRs (since we keep seeing people changing unrelated .idea files).

PR #3951, opened 2021-10-20
Fix #3798: Introduced files in values-night [Dark Mode Implementation]

### LINKED ISSUE #3798: Introduce dark mode theme files [Dark Mode Implementation].
Create a dark mode theme files under values-night folder.

ayush0402:
## Explanation
Fixes #3798 

Approach :

- Rename all colors with correct naming convention
- Rename `colors.xml` to `colors_defs.xml` and now name all colors in a meaningful way. [IN THIS DO NOT USE `_color` word anywhere in the name.]
- Create color_palette.xml and component_colors.xml file.
- Now choose 3-4 activities and create colors for them correctly and create a PR.
- …. Repeat above such that each PR has 15-20 files of changes.

## Essential Checklist

## For UI-specific PRs only


ayush0402:
@rt4914 PTAL. I think we can also just have night file or colors.xml instead of having it for both themes.xml and colors.xml because themes are referring to colors.xml only.

rt4914:
Assigning to @BenHenning for approach confirmation.

app/src/main/res/values-night/themes.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+  <style name="OppiaTheme" parent="Theme.MaterialComponents.DayNight.DarkActionBar.Bridge">
```
BenHenning: If we're using the DayNight switchable theme, why do we need to redefine all of these styles?

app/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,178 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BenHenning: @rt4914 this seems like the correct approach, but it's a *lot* of colors to maintain. My thoughts at this point are:
1. Would it be possible to reduce these colors to the base ones that we need? For example, I'd more expect to see something like: oppiaProgressChapterFinished as a color that uses something like "oppiaProgressAccentColor" which is then defined with night/day variations. We ought to steer away from defining custom colors like white_80 or, if we do, they should be in a common colors.xml used by both modes. From there, we should have contextually-specific colors to the individual screens which then use commonly defined colors (which also have semantic meaning based on our color palette). These last colors are the ones that generally need to be changed out based on day/night mode since they control the color palette for the entire app. We could conceivably even have separate files for these: color_defs.xml (for literal colors), component_colors.xml (which are UI-specific colors that use colors from color_palette.xml), color_palette.xml (for which we have day/night versions since these define the actual colors that are used, e.g. by pointing to the shared color_defs.xml).
2. Would it be unreasonable to have a basic script CI check that verifies all colors defined in the specific file that needs to have day/night variants are exactly in both files & that neither file has extras/any missing? For example, if we use the approach proposed above then both the day & night modes of color_palette.xml would need to be validated to make sure they both contain exactly the same color definitions. I think other aspects of colors could probably be validated, too (see summary below).

To summaries:
- I think it would be clearer if all actual color definitions (e.g. ``#ff1234``) were defined in a shared color_defs.xml file with corresponding names like 'oppia_red_20'. Note that in general, we should use the colors from material rather than defining our own custom colors (so this is only for cases when material  doesn't provide us a built-in color)
- The colors above would be referenced by an explicit color_palette.xml file that has day/night variants and generic contextual color names, like "oppia_main_heading_color"
- Further, all colors used by layouts should be specific to their context (e.g. "home_screen_background") which, in turn, reference colors from the color palette to ensure that they receive the correct color based on day/night mode
- We would have a CI check to validate various color properties including:
  - Ensuring that all colors in both versions of color_palette.xml are matching in name so that none are missing/extra
  - Ensuring that all colors referenced in layout files come from component_colors.xml
  - Ensuring no colors in component_colors.xml or color_palette.xml directly define colors
  - Ensuring all colors in component_colors.xml only reference colors from color_palette.xml
  - Ensuring that all colors from color_palette.xml only reference colors from color_defs.xml or material library
  - Ensuring that all colors in color_defs.xml are raw colors (e.g. with a hex code)
  - Ensuring that colors are defined nowhere else other than these predetermiend files
- We probably also need to add regex patterns to prohibit directly assigning or computing colors in Kotlin & hide that functionality behind an allowlisted API (similar to what we do for localization) since assigning any color may result in a dark mode issue. As part of this exercise, a lot of existing situations will probably break & we'll need to evaluate for each case how to handle dark mode variations, if necessary. 

Apologies if this comment is a bit long & rambly. I think this PR is an excellent start, and it's led me to think a lot more about how to set this up for the long-term. I'm *not* suggesting that this PR tackle all of the above, Instead, I'd like yours and @ayush0402's thoughts on the above, how we should structure the work, and then what role this PR plays in all the other pieces needed to add dark mode to the app.

BenHenning (commented):
Thanks @ayush0402 & @rt4914! Left two comments, but one main one with a bunch of suggestions for how I think we should proceed from here. Please let me know how this relates to existing plans, and what aspects you think should be considered.

app/src/main/res/values-night/colors.xml
rt4914: The overall approach that you mentioned sounds really great to me. I am unable to understand the difference between `color_palette.xml` and `component_colors .xml`

@ayush0402  Once I get the clarification I will mention the smaller PRs in which this can be done meanwhile you can also leave your comments/doubts here. If you have understood this then I suggest first create a PR which removes all those colors which are not getting used currently.

BenHenning: My thinking is that component_colors.xml contain contextual colors used in components, such as "home_screen_background_color" (which would correspondingly be used by the homescreen). This follows the trend of naming resources (like strings & dimensions) based on their intended usage rather than their value since that provides better inline context.

color_palette.xml, on the other hand, contains colors which correspond to the color palette that Chantel laid out in document form some months ago (though I suspect it may not be complete, so we would need to sync with the UX team on what the final color palette should be).

The idea here is basically:
- Most colors are tied to components, so most of the color declarations will be in component_colors.xml
- Component colors shouldn't actually need to change for day/night mode since they correspond to high-level color palette colors
- We can simplify color_palette.xml (which does need to change based on day/night mode) by having it points to actual color definitions (color_defs.xml)
- It doesn't make sense to change colors in color_defs.xml since something like "oppia_red20" will always be "oppia_red20"
- Having a single small set of colors in color_palette.xml that aligns with a UX-managed document and controls day/night colors for the whole app is a good simplification since it provides a clearer "design language" that we can use when talking to the UX team

Does that help clarify?

rt4914: Okay can you confirm if my interpretation is correct:

**color_defs.xml**
```
  <color name="black_12">#1F000000</color>
  <color name="white_80">#CCFFFFFF</color>
  <color name="oppiaLightGreen">#F0FFFF</color>
  <color name="oppiaDarkBlue">#2D4A9D</color>
  <color name="lightGreenColor">#2D4A9D</color>
  <color name="darkGreenColor">#2D4A9D</color>
```
- All colors in this file should be pure hex values.

**color_palette.xml (day)**
```
   <color name="oppiaToolbarColor">@colors/lightGreenColor</color>
```
- All colors for this file should come from `color_defs.xml` or material colors.

**color_palette.xml (night)**
```
   <color name="oppiaToolbarColor">@colors/darkGreenColor</color>
```
- All colors for this file should come from `color_defs.xml`
- All colors in this file should have a corresponding color in `color_palette.xml (day)` too.

**component_colors.xml**
```
  <color name="somUIColorWhichDoesNotNeedVariationsForDayNight">@colors/white_80</color>
```
- All colors for this file should come from `color_defs.xml`
- All colors should have conceptual name.
- All colors in this file should be **independent** of day/night mode


@BenHenning Is this correct?

BenHenning: All of that looks correct @rt4914 except one thing: component_colors.xml should always reference colors from color_palette.xml (that way components are properly recolored in day/night changes), and layouts/Kotlin code should generally only reference these colors

I was also thinking that "oppiaToolbarColor" might be too "component-like" for the palette, but it actually fits. I could see us doing something like:

**component_colors.xml**
```
<color name="home_screen_toolbar_color">@color/oppia_toolbar_color</color>
```

**color_palette.xml** (day mode)
```
<color name="oppia_toolbar_color">@color/oppia_primary_color</color>
<color name="oppia_primary_color">@color/light_green</color>
```

Or, we could define the toolbar in component_colors.xml if it always depends on the primary color. I think there will be some cases where just have to decide which file these types of colors should go into (& fortunately it's easy to change over time if need to). This also seems to show that it should be valid for colors within a file to reference other colors within that same file; that seems fine.

Separately, two other things come to mind:
- Since themes can also reference colors and they're inherently generic, I think it's fine if they reference colors in either component_colors.xml or color_palette.xml, but I defer to your judgement here (we can always add *more* colors if we want to organize things a particular way)
- We should be consistent about the snake_case vs. lowerCamelCase when naming colors since we've been inconsistent (I prefer snake_case since it will also then match layout IDs & strings)--any thoughts here?

rt4914: Everything above sounds good. The example makes lot of sense. Thanks.
Replying to last two questions:
1. themes should reference only `component_colors.xml` just to keep it consistent with `layout/*.xml` files as they will also be referencing only `component_colors.xml` file.
2. Yes, I agree there is inconsistency and started because of two things: (a.) by default strings creating by android like `app_name` used snake_case and (b.) the colors.xml file had three colors by default at start which are `colorPrimary`, `colorPrimaryDark`, `colorAccent` which followed lowerCamelCase format and because of this the confusion as well as inconsistency started. From now onwards lets make this final: **All files in values folder should follow snake_case naming convention**

rt4914:
@ayush0402 Have a look at the comments and now I think its final and you can start working. I suggest creating a short document or update the issue description summarizing what exactly needs to be done with examples and then start working accordingly. Thanks :)

ayush0402:
@rt4914 Understood. I created a PR removing unused colors from the colors.xml file at #3979.

app/src/main/res/values-night/colors.xml
BenHenning: SGTM @rt4914. If you'd like, I think we can even make a regex check for this by prohibiting capitalization in certain contexts. I'm not sure if this'll have any false positives, but we can try it and see what existing cases are triggered:

```textproto
file_content_checks {
  file_path_regex: ".+?\\.xml"
  prohibited_content_regex: "name=\".+?[A-Z].+?\"
  failure_message: "All resource names must be written in snake_case."
}
file_content_checks {
  file_path_regex: ".+?\\.xml"
  prohibited_content_regex: "android:id=\".+?[A-Z].+?\"
  failure_message: "All layout IDs must be written in snake_case."
}
```

WDYT?

ayush0402:
@rt4914 Right now colors.xml has naming such as ```colorPrimary``` etc. Should these be changed to their actual color names instead (like oppia_green_1) and then called ```colorPrimary``` in color_palette.xml instead?
If yes, then first we should be renaming all the colors in colors.xml (renamed to color_def.xml) to their actual names and at the same time creating their instances in color_palette.xml (day/night)?
Which in turn will be referenced from ```color_palette.xml``` to ```themes.xml```.
Also for the colors used in layouts for different screens we should be referencing colors from ```component_colors.xml```, where ```component_colors.xml```  will be referencing from ```color_palette.xml```.
Is this correct?
Also, what about the colors having same color in both day and night variants? Should they still be created twice in both ```color_palette.xml``` or just directly be declared in ```component_color.xml```?

rt4914:
> @rt4914 Right now colors.xml has naming such as `colorPrimary` etc. Should these be changed to their actual color names instead (like oppia_green_1) and then called `colorPrimary` in color_palette.xml instead? If yes, then first we should be renaming all the colors in colors.xml (renamed to color_def.xml) to their actual names and at the same time creating their instances in color_palette.xml (day/night)? Which in turn will be referenced from `color_palette.xml` to `themes.xml`. Also for the colors used in layouts for different screens we should be referencing colors from `component_colors.xml`, where `component_colors.xml` will be referencing from `color_palette.xml`. Is this correct? Also, what about the colors having same color in both day and night variants? Should they still be created twice in both `color_palette.xml` or just directly be declared in `component_color.xml`?

I suggest following things:
1. Do not change anything in `colors.xml` yet as it will create lot of conflicts. Instead create `colors_defs.xml` first. This will contain all color definitions as mentioned here: https://github.com/oppia/oppia-android/pull/3951#discussion_r736907251
2. Don't change to `oppia_green_1` as `colorPrimary` makes more sense (This will be done in `colors_defs.xml` ) file.
3. Note as per above [this](https://github.com/oppia/oppia-android/pull/3951#discussion_r739768103) all layout/themes xml files will using `component_colors.xml` file only.
4. Colors which have day/night mode both will be defined in `color_palette.xml` file and used in `component_colors.xml` if file. Colors which do not have day/night variations will not be defined in `color_palette.xml` and will only be defined in `component_colors.xml` file.

@BenHenning Is this all correct?

BenHenning:
> > @rt4914 Right now colors.xml has naming such as `colorPrimary` etc. Should these be changed to their actual color names instead (like oppia_green_1) and then called `colorPrimary` in color_palette.xml instead? If yes, then first we should be renaming all the colors in colors.xml (renamed to color_def.xml) to their actual names and at the same time creating their instances in color_palette.xml (day/night)? Which in turn will be referenced from `color_palette.xml` to `themes.xml`. Also for the colors used in layouts for different screens we should be referencing colors from `component_colors.xml`, where `component_colors.xml` will be referencing from `color_palette.xml`. Is this correct? Also, what about the colors having same color in both day and night variants? Should they still be created twice in both `color_palette.xml` or just directly be declared in `component_color.xml`?
> 
> I suggest following things:
> 
> 1. Do not change anything in `colors.xml` yet as it will create lot of conflicts. Instead create `colors_defs.xml` first. This will contain all color definitions as mentioned here: [Fix #3798: Introduced files in values-night [Dark Mode Implementation] #3951 (comment)](https://github.com/oppia/oppia-android/pull/3951#discussion_r736907251)
> 2. Don't change to `oppia_green_1` as `colorPrimary` makes more sense (This will be done in `colors_defs.xml` ) file.
> 3. Note as per above [this](https://github.com/oppia/oppia-android/pull/3951#discussion_r739768103) all layout/themes xml files will using `component_colors.xml` file only.
> 4. Colors which have day/night mode both will be defined in `color_palette.xml` file and used in `component_colors.xml` if file. Colors which do not have day/night variations will not be defined in `color_palette.xml` and will only be defined in `component_colors.xml` file.
> 
> @BenHenning Is this all correct?

Apologies for the delayed follow-up. I agree with all of these points @rt4914.

rt4914:
@ayush0402 Your questions have been answered. Please let us know if there are any other doubts.

ayush0402:
@rt4914 PTAL I tried implementing dark mode for onboarding fragments/slides for now to make sure how things are working.
| Light | Dark |
| ------------- | ------------- |
|Screenshot_1637488244|Screenshot_1637488182|

I tried adding some colors from the mock link you provided but I am not sure If I understand all colors there correctly, right now I am assuming colorPrimary to be Primary color in the mock , colorPrimaryDark to be Secondary color and colorAccent to be Backdrop color as named in the mocks given.

Also we seem to be using same shade of black for headings and subtexts in Light scheme of colors but mock has different shades of whites for Heading and subtext in dark mode so I added them accordingly to the styles.xml file.

Also I am kind of confused about naming of the colors so please suggest what convention do I follow to name them.

Where should this PR be headed next? Am I supposed to do what I did for onboarding fragments for every single layout in this PR or I was wondering it would be good if we could create separate issue for various layouts adding upto this component_colors.xml and color_pallete.xml accordingly? Because without looking at the layout it is confusing as to which colors should go where as there are many colors with variations to handle. Suggest what should be the goal of this PR as of now.

app/src/main/res/values-night/themes.xml
rt4914: +1

app/src/main/res/values-night/color_palette.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <color name="oppia_primary_color">@color/colorPrimary_DarkMode</color>
```
rt4914: The color code `colorPrimary_DarkMode` is not properly named. All color codes should be `snake_case` format.

app/src/main/res/values/color_defs.xml
```diff
@@ -0,0 +1,158 @@
...
+  <color name="onboarding4StatusBar">#3B5893</color>
+  <!-- ICON COLORS -->
+  <color name="mergeIconEnabled">#FF000000</color>
+  <color name="mergeIconDisabled">#999999</color>
+  <!-- HIGHLIGHTED COLORS -->
+  <color name="adminHighlighted">#A5D3EC</color>
+  <color name="comingSoonEndColor">#674172</color>
+  <color name="comingSoonStartColor">#7659B6</color>
+  <!-- DARK MODE COLORS -->
+  <color name="colorPrimary_DarkMode">#3bd1c4</color>
```
rt4914: In `color_defs` do not mentioned `dark` or `light` on the basis of mode in any color name.
Instead it should be like `white`, `black`, `gray_1`, `red`, `oppia_primary`, `oppia_light_green`  or `oppia_dark_green` etc.

app/src/main/res/values/component_colors.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <color name="primary_color">@color/oppia_primary_color</color>
```
rt4914, line 3: This is incorrect as this color is defined in `color_palette.xml` file.
Have a look at this: https://github.com/oppia/oppia-android/pull/3951#discussion_r736907251

`component_colors` can only have colors from `color_defs.xml` file.

rt4914 (changes requested):
@ayush0402 PTAL. Thanks.

app/src/main/res/values/component_colors.xml
ayush0402: @rt4914 I think Ben wanted ```component_colors.xml``` to only reference ```color_palette.xml``` according to this [comment](https://github.com/oppia/oppia-android/pull/3951#discussion_r737740334).

app/src/main/res/values-night/color_palette.xml
ayush0402: Done.

app/src/main/res/values/color_defs.xml
ayush0402: PTAL now.

app/src/main/res/values-night/themes.xml
ayush0402: Deleted this file.

app/src/main/res/values/component_colors.xml
rt4914: Ack.

rt4914:
@ayush0402 This has been discussed so unassigning for now. (Summary for others: Ayush will create new PRs on top of one another in which he will be first refactoring the current code and afterwards implement the new changes for dark mode.)

ayush0402:
Creating some smaller PRs in parallel which will ease the implementation for the discussed approach. 
Starting with renaming all existing colors in colors.xml with proper ```snake_case``` conventions. #4012

ayush0402:
Started working on implementing dark mode to layouts. #4032

BenHenning:

Issue #1050, opened 2020-05-10
Solutions are not displayed correctly except for text and fraction input types


BenHenning:
Per https://github.com/oppia/oppia-android/blob/65b1dc8/app/src/main/java/org/oppia/app/player/state/hintsandsolution/HintsAndSolutionAdapter.kt#L160 it seems that solutions are currently only supported for fractions whereas they should be supported for all interaction answer types.

aggarwalpulkit596:
Hint will always be of type text right ? 🤔

BenHenning:
I don't think #1059 full solves this. The text being sent to the correct answer is still assuming that a solution has a numerator and denominator (that's only true for fractions--what about solutions for numeric input or multiple choice?).

As far as I'm aware, hints can be any answer, which means those answers can correspond to any interaction type.

BenHenning:
Also, we should always use strings for formatting text like "<numerator>/<denominator>" (Android Studio gives a warning for these cases). We need to update hints to generically handle any interaction answer type, including non-text ones, and we should make sure the answer is formatted in a way that's understandable by the user.

veena14cs:
> I don't think #1059 full solves this. The text being sent to the correct answer is still assuming that a solution has a numerator and denominator (that's only true for fractions--what about solutions for numeric input or multiple choice?).
> 
> As far as I'm aware, hints can be any answer, which means those answers can correspond to any interaction type.

@BenHenning There are 2 types of correctAnswer one is object and one is string. Object corresponds to fractions based on json files in assets and rest all interaction types as well as other then fractions have correctAnswer type as string.

seanlip:
Something seems a bit confused here. The bug title says "solutions" but the description talks about "hints". These are two different things.

Hints are always in rich-text format.

Solutions include an "answer" and an "explanation" component. The explanation is always in rich-text format. The type of the "answer" can vary, and it depends on the specific interaction in the corresponding question. The type of the answer would be the same as the type of the answer to the corresponding question (which is string in some cases, and other types of objects in other cases). In multiple-choice I think its an integer. You'll need to check the python definition files in the web codebase to figure out what the type for a particular interaction is.

BenHenning:
Thanks for pointing out the inconsistency in the description @seanlip -- I fixed that. This is only corresponding to solutions.

[...]

PR #3955, opened 2021-10-20
Fix #3946: Fix solution automatically revealing after first hint reveal

### LINKED ISSUE #3946: Solutions seem to immediately appear without prompting
**Describe the bug**
When playing through a lesson in the alpha MR3 build, it was discovered that solutions seem to appear after the first hint is revealed.

**To Reproduce**
1. Open a lesson (e.g. Place Values 1)
2. Trigger a hint to show (e.g. submit 2 wrong answers)
3. Open the hint dialog & reveal the hint
4. Close the hint dialog & reopen it
5. Observe that the solution has been exposed

**Expected behavior**
Solutions shouldn't show up until triggered, and should require explicit revealing.

**Demonstration**
-

**Environment**
- Device/emulator being used: Pixel 5
- Android or SDK version (e.g. Android 5 or SDK 21): SDK 30
- App version (you can get this through system app settings or via the admin controls menu in-app): 0.6-alpha-de345f92ee

**Additional context**
This doesn't seem to repro locally in test lessons, so it seems alpha-specific. More investigation is needed.

BenHenning:
## Explanation

Fix #3946

Per #3946 there's a bug in some contexts where, after revealing a hint, the solution may automatically be revealed as well when the user reopens the solution dialog.

The root cause of this is that the hint handler decides whether to show solutions based on the presence of a correct answer, not based on the presence of a solution itself. This presents a problem specifically for proto-based lessons because in some cases correct_answer will be absent (if the converter can't properly convert the answer into an app-acceptable format, e.g. text or fraction). This is technically a valid case to introduce a failure, but due to #1050 we can't actually assume that correct_answer is properly defined. Thus, the solution is to change the check to instead check for whether solution is present. This *does* allow for an empty solution object to be added to the proto, but this shouldn't realistically happen in practice so it should be fine.

To repro (on this branch):
- Enable proto loading in CachingModule
- Build & install a dev flavor of the app
- Start a new Prototype Exploration play session
- Play to the numeric expression state
- Enter 2 wrong answers
- Reveal the hint
- Close & reopen the dialog
- (If you want to repro the issue, you'll need to replace the implementation of HintHandlerProdImpl with the current develop copy)
- Without the fix, the solution will automatically be revealed. With the fix, the solution won't yet be available to reveal.

Note that this doesn't affect json since we've taken some measure to ensure that correct_answer is filled with *something* even if that value isn't quite correct.

To ensure this regression is properly caught, tests were added both in StateFragmentLocalTest and HintHandlerProdImplTest. Note that the latter required introducing a new test exploration that specifically tests a missing correct_answer, and the former required changes to the prototype exploration to introduce a hint & solution in the numeric expression state to more easily repro this issue. Test results:
- Failing without the fix (Bazel only since it requires textproto loading): https://app.buildbuddy.io/invocation/da48df48-444c-48d5-870f-dec2c75b68b1
- Passing with the fixes: https://app.buildbuddy.io/invocation/92f3739b-0477-43c1-82df-8170b8fd197a

Other miscellaneous changes:
- StateFragmentLocalTest had some cleaning up since its helpers were a bit confusing referencing states for different explorations (now that the test suite uses at least 2 explorations for testing)
- StateFragmentLocalTest was updated to depend on textproto assets in Bazel for better coverage, and so that the failure case could be triggered per the new regression tests
- StateRetriever was updated to allow a missing correct_answer in solutions (for parity with the textproto variant, but the implementation is a bit more permissive and actually results in passing tests regardless of the fix)
- StateRetriever was also changed so that solutions are not set if they aren't loaded (to avoid an empty solution being included which shouldn't actually ever happen for loaded binary protos). This ensures that the updated check (hasCorrectAnswer -> hasSolution) results in consistent & correct behavior for both JSON & textproto runs of the hint tests/app builds.
- HintHandlerProdImpl was updated to use an AtomicInteger for hint sequencing. This isn't related to the original issue, but there's a likely race condition in the current implementation that I'm fixing at the same time. No regression test is reasonably possible to actually test this, so it needs to be a verify-on-inspection type change.

## Essential Checklist

## For UI-specific PRs only
N/A -- this is fixing a UI-affecting issue, but doesn't actually change the UI.

### DIFF
--- a/app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
@@ -2603,6 +2603,7 @@ class StateFragmentTest {
     targetTextViewId: Int
   ) {
     scrollToViewType(SELECTION_INTERACTION)
+    // First, check that the option matches what's expected by the test.
     onView(
       atPositionOnView(
         recyclerViewId = R.id.selection_interaction_recyclerview,
--- a/app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
+++ b/app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
@@ -1146,6 +1146,9 @@ class StateFragmentLocalTest {
     }
   }
 
+  // TODO(#1050): Add a test for verifying that the solution is correct for non-text & non-fraction
+  //  interactions.
+
   @Test
   fun testStateFragment_stateWithNumericSolution_revealHint_reopenDialog_onlyOneHintShown() {
     launchForExploration(TEST_EXPLORATION_ID_2).use {
--- a/app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
+++ b/app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
@@ -72,6 +72,7 @@ import org.oppia.android.app.player.state.itemviewmodel.StateItemViewModel.ViewT
 import org.oppia.android.app.player.state.itemviewmodel.StateItemViewModel.ViewType.PREVIOUS_RESPONSES_HEADER
 import org.oppia.android.app.player.state.itemviewmodel.StateItemViewModel.ViewType.SELECTION_INTERACTION
 import org.oppia.android.app.player.state.itemviewmodel.StateItemViewModel.ViewType.SUBMIT_ANSWER_BUTTON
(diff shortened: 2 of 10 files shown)


BenHenning:
@rt4914 PTAL (for codeowners + other changes if you can since the PR isn't affecting too many files).

rt4914 (approved):
LGTM, thanks.

BenHenning:
Thanks @rt4914!

BenHenning:
SIGSEGV in StateFragmentTest for the Gradle run. Re-running since that test should pass in this PR.

BenHenning:
Looks like everything is passing, so merging now.

BenHenning:
This is confirmed as fixed in the release branch.

Issue #3789, opened 2021-09-14
Sharded Bazel tests are build-flaky in CI


BenHenning:
The issue described in #3759 is not properly mitigated in CI and is causing issues (full runs seem to always have at least 1 failure due to this build error). From looking at the code, I think the retry mechanism isn't being set up correctly.

BenHenning:
Note also that this is new with sharding since before we didn't combine multiple tests in the same build, and this only happens when building several app module tests at the same time.

BenHenning:
Copying from #3759: Aha! I think I know why the retry mechanism isn't working (been puzzling over this for a while). Per a recent failure log (https://pipelines.actions.githubusercontent.com/v2jo6u0i4lUfyxD2aH2pOnpX9TRM6k4RpTudJnUSvFarVABEQV/_apis/pipelines/1/runs/28918/signedlogcontent/16?urlExpires=2021-10-06T20%3A10%3A31.6787136Z&urlSigningMethod=HMACV1&urlSignature=nJkBVgj0uRSbxKSY8WkMsxQEzQFvEOMYn8XJ4kuxe6Y%3D) it's clear that only 1 build run was attempted despite the configuration being set up to attempt 5 (https://github.com/oppia/oppia-android/blob/develop/.github/workflows/unit_tests.yml#L238). It seems that this is because GitHub Actions initializes bash using 'bash -e' which, per https://stackoverflow.com/a/19622569/3689782, means the actions script immediately fails if any of the Bazel commands fail.

I think we'll need to wrap the Bazel command so that the failure doesn't bubble up, but more investigation into possible solutions is needed.

BenHenning:
Reopening per https://github.com/oppia/oppia-android/pull/3908#issuecomment-937132130.

BenHenning:
The last ~10 develop commits seem to not run into this issue (they seem to mostly be hitting this SEGSEGV described in #2844 for StateFragmentTest/StateFragmentLocalTest. PRs I've been working on have also not hit this (in a way that results in a CI failure, as intended by this PR). Closing since the desired outcome seems to have been achieved.

PR #3969, opened 2021-10-25
[RunAllTests] Try to fix/workaround rest of #2844: Add retry mechanism when running tests

BenHenning:
## Explanation
Try to workaround the last issue in #2844: a SIGSEGV thrown sometimes when running StateFragmentTest/StateFragmentLocalTest.

While it would be preferable to fix this issue, it will likely be extremely time consuming since it'll require digging into a JVM bug. After some effort to try and find a root cause, I haven't been able to come up with a single dependable way to work around the issue. It mostly went away with previous mitigations, but after adding test sharding it seems to come up a lot.

This solution introduces the same while-loop retry mechanism as we do for builds (which was effective in addressing #3789). Bazel makes this work well since passing tests won't be re-run (their results are cached--see the CI results for this PR). Given that the issue seems to happen less often when StateFragmentTest/StateFragmentLocalTest runs by itself, this is a reasonable outcome (since it'll generally result in running just those tests for runs 2-5).

That being said, I'm not a complete fan of this solution since it will:
- Result in true flakes being discovered less often, though fortunately the same retry mechanism also applies to the develop branch
- False flakes could be missed which could result in them being checked in (but similarly, this also should result in a more stable situation since develop won't fail from them, but it does mean legitimate issues could be missed)
- Result in failures taking longer to be reported since they'll be retried up to 5 times

While the second outcome is definitely worse, it also seems to happen much less often so it seems like a worthwhile trade-off for the stability benefits that we get from this fix.

#3970 was filed to track the long-term fix of the underlying SIGSEGV so that this mechanism isn't needed for CI runs to reliably pass.

## Essential Checklist

## For UI-specific PRs only
N/A -- infrastructure change

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -277,7 +277,7 @@ jobs:
           BAZEL_TEST_TARGETS: ${{ env.BAZEL_TEST_TARGETS }}
         run: |
           # Attempt to build 5 times in case there are flaky builds.
-          # TODO: Remove this once there are no longer app test build failures.
+          # TODO(#3970): Remove this once there are no longer app test build failures.
           i=0
           # Disable exit-on-first-failure.
           set +e
@@ -298,7 +298,7 @@ jobs:
           BAZEL_TEST_TARGETS: ${{ env.BAZEL_TEST_TARGETS }}
         run: |
           # Attempt to build 5 times in case there are flaky builds.
-          # TODO: Remove this once there are no longer app test build failures.
+          # TODO(#3970): Remove this once there are no longer app test build failures.
           i=0
           # Disable exit-on-first-failure.
           set +e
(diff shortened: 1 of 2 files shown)


BenHenning:
@vinitamurthi could you PTAL as a reviewer since this is tied pretty strongly into developer workflow/happiness?

BenHenning:
Note that I just added a commit to trigger a real failure in StateFragmentLocalTest to make sure the failure is properly represented in CI results.

BenHenning:
Looks like the failure worked as expected. Reverting now so that the PR is passing CI.

vinitamurthi:
Overall this LGTM ..Once question though -- does it make sense to retry only failures? From my understanding , this is retrying all targets right?

vinitamurthi (approved):

BenHenning:
> Overall this LGTM ..Once question though -- does it make sense to retry only failures? From my understanding , this is retrying all targets right?

@vinitamurthi you're correct that it will retry all targets, but Bazel caches passing results so it won't re-run those (unless you pass a specific flag to force it to re-run, or a flag that implies that like "runs_per_test."

BenHenning:
Thanks also for the review @vinitamurthi! Going ahead and merging this since nothing seems outstanding, and it'd be really good to get this merged for improved CI stability.

Issue #3893, opened 2021-10-06
[A11Y] Need DiffUtils with RecyclerView for better experience for a11y users [Blocked on #171]


rt4914:
**Issue**
In exploration player when we use screen-reader and mark correct answer, because the state changes, the screen-reader starts from the top which provides really bad experience.

https://user-images.githubusercontent.com/9396084/136243639-0a762493-de80-4703-95e5-a23c0292909f.mp4

With the use of diffutils after any change the user can remain on the same item of recyclerview and thus providing consistent experience.

**Places where this needs to be fixed**

- Onboarding Screen
- Admin Controls
- Home Screen
- Revision Card
- Exploration Player
- Question Player

BenHenning:
FYI that this is blocked on #171.

vrajdesai78:
Need to improve the experience for A11Y users by handling state in recycler view for the following screens:

1. Onboarding Screen
2. Admin Controls
3. Home Screen
5. Revision Card
6. Exploration Player
7. Question Player

vrajdesai78:
#4650 is also a part of this issue

PR #3353, opened 2021-06-21
Fix #2324: Changes in Submitted Answer List Items [A11y]

rt4914:
## Explanation

Fix #2324

## Problem
If learner submits multiple same incorrect answers or atleast 1 incorrect answer in case of DragAndDrop then the A11y Scanner fails because there are multiple same texts.

## Solution
For this problem user experience is more important than A11y Scanner failures and therefore for all correct answers we append `Correct submitted answer` and for all incorrect answers we append `Incorrect submitted answer`. In this case even if learner enters same incorrect answer multiple types (for which Scanner fails) we are prioritising user experience here.

Have a look at following comments: https://github.com/oppia/oppia-android/issues/2324#issuecomment-780799146 and https://github.com/oppia/oppia-android/issues/2324#issuecomment-780858109

## Output

https://user-images.githubusercontent.com/9396084/122798150-e23a2980-d2dd-11eb-94f6-120b6ac0f83b.mp4

https://user-images.githubusercontent.com/9396084/122798180-ec5c2800-d2dd-11eb-8ada-4006e10039d7.mp4

https://user-images.githubusercontent.com/9396084/122798198-f0884580-d2dd-11eb-9331-6ca43c1b14b5.mp4

Screenshot 2021-06-21 at 10 42 41 PM

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #3980, opened 2021-11-02
Fix #3907: [A11Y] Output Congratulations for screenreader

### LINKED ISSUE #3907: [A11Y] Output `Congratulations` for screenreader
Output `Congratulations` for screenreader.

Whenever a user submits a correct answer we show a `Congratulations` toast. This is currently not readable by screen-reader. 

It should be readable by screen-reader as soon as it appears.

viktoriias:
## Explanation

Fixes #3907: [A11Y] Output Congratulations for screenreader. It's not "Congratulations", but "Correct".

To be able to test new behavior, announceForAccessibility is wrapped in its own container (AccessibilityService), and swapped with a test-only fake to determine if it's been called.

## Essential Checklist

## For UI-specific PRs only


viktoriias:
Screen recording:

https://user-images.githubusercontent.com/10843277/140162067-5bc45b3e-5555-412e-a869-d44e2ab21e69.mp4

rt4914 (commented):
@BenHenning Can you have a look at the output for this PR. I have couple of concerns about this:

1. Should we actually read out `Congratulations`/ `Correct` because most of the time once correct answer is submitted we do have some feedback text which says that the answer was correct.
2. If we do need to announce it, can you suggest some strategy based on which we can test this on Espresso or Robolectic?


Thanks

BenHenning:
> @BenHenning Can you have a look at the output for this PR. I have couple of concerns about this:
> 
> 1. Should we actually read out `Congratulations`/ `Correct` because most of the time once correct answer is submitted we do have some feedback text which says that the answer was correct.
> 2. If we do need to announce it, can you suggest some strategy based on which we can test this on Espresso or Robolectic?
> 
> Thanks

@rt4914 replies below:

1. This is a really good point. I actually need to ask @seanlip about this. Sean: pedagogically, how important is it to communicate the notion of the "correct" banner to the user? I.e. is it actually optional such that it's nice-to-have for sighted users but unimportant for a11y users (since we have redundancy in the feedback as @rt4914 mentioned)? The video @viktoriias linked above demonstrates the behavior with the extra "Correct" announcement (sound needs to be on). FWIW my thinking is that we should not announce it now that we can see what the behavior is here.
2. Regarding testing, once we actually start using announceForAccessibility we should wrap it in our own container, limit usage via regex exemptions, and swap it out with a test-only fake that we can use to determine if it's been called. I think that's the best that we can do here.

seanlip:
For (1), I think it is fine not to announce it. We typically have feedback provided for the correct answer, which should suffice.

If you want to be more pedantic, I would say, announce it if and only if the "correct answer feedback" is empty. But this should only happen in a minority of cases.

BenHenning:
> For (1), I think it is fine not to announce it. We typically have feedback provided for the correct answer, which should suffice.
> 
> If you want to be more pedantic, I would say, announce it if and only if the "correct answer feedback" is empty. But this should only happen in a minority of cases.

Thanks @seanlip. This case is interesting. The current exploration structure does allow feedback to be empty, and so far as I'm aware we don't actually require it in any cases (though maybe it should be required if an answer is labeled as correct?).

Either way, I like the framing of relying on feedback to provide context and otherwise to supplement the experience when we know for sure there's no correct context given. @rt4914 & @viktoriias do you have any concerns with this adjustment?

viktoriias:
> Either way, I like the framing of relying on feedback to provide context and otherwise to supplement the experience when we know for sure there's no correct context given. @rt4914 & @viktoriias do you have any concerns with this adjustment?

I know very little about accessibility. On one hand, I would assume that talkback experience should be the same as visual. On the other hand, it may be overwhelming to hear too many sounds or repeating sounds, while seeing "Correct" confirmation twice is fine.

To make sure I understand suggested adjustment:
 - We only want to pronounce "Correct" when there is no correct answer feedback.
 - "Correct" toast and confetti will remain unchanged and they will appear regardless of feedback.

BenHenning:
Your understanding is correct & matches the proposal.

viktoriias:
I just realized that the feedback is not in the talkback, so we only hear "Correct" once. @rt4914 is not reading the answer feedback intended or will be changed?

https://user-images.githubusercontent.com/10843277/140618985-bc214ea4-711d-480e-a923-1862305c294b.mp4

rt4914:
> > For (1), I think it is fine not to announce it. We typically have feedback provided for the correct answer, which should suffice.
> > If you want to be more pedantic, I would say, announce it if and only if the "correct answer feedback" is empty. But this should only happen in a minority of cases.
> 
> Thanks @seanlip. This case is interesting. The current exploration structure does allow feedback to be empty, and so far as I'm aware we don't actually require it in any cases (though maybe it should be required if an answer is labeled as correct?).
> 
> Either way, I like the framing of relying on feedback to provide context and otherwise to supplement the experience when we know for sure there's no correct context given. @rt4914 & @viktoriias do you have any concerns with this adjustment?

@BenHenning Sounds good. We should announce only when Feedback is not available.

rt4914:
> I just realized that the feedback is not in the talkback, so we only hear "Correct" once. @rt4914 is not reading the answer feedback intended or will be changed?
> Screen.recording.mp4

@Viktoriia-Lozova This is happening for cases where we have ItemSelectionInteraction and I suspect thats because of nested recyclerview and there are chances it will happen for DragAndDrop too.
For other cases like `FractionInput`, it works correctly (though we have to start from top because of DiffUtils issue) 

Both of these should work perfectly once #3893 issue is fixed.

viktoriias:
"Correct" is not announced when feedback is present:

https://user-images.githubusercontent.com/10843277/141225171-ff1a9d6b-cb5b-4542-aa75-0acffd12a0f5.mp4

viktoriias:
"Correct" is announced when there is no feedback

https://user-images.githubusercontent.com/10843277/141225411-9bc563cf-121c-41f4-82b2-cbb30d3130b3.mp4

:

viktoriias:
Please take a look at the new behavior before I add tests (which sounds like a lot of work):

> Regarding testing, once we actually start using announceForAccessibility we should wrap it in our own container, limit usage via regex exemptions, and swap it out with a test-only fake that we can use to determine if it's been called. I think that's the best that we can do here.

rt4914:
> Please take a look at the new behavior before I add tests (which sounds like a lot of work):
> 
> > Regarding testing, once we actually start using announceForAccessibility we should wrap it in our own container, limit usage via regex exemptions, and swap it out with a test-only fake that we can use to determine if it's been called. I think that's the best that we can do here.

@Viktoriia-Lozova That looks good. Once you have submitted an answer and then you traverse back from top to bottom in recyclerview you will notice that the last answer that they submitted outputs "Correct submitted answer: <answer>" which again makes things clear for the user. (We won't need to traverse back from top once #3893  is fixed.)

Have a look at videos in this PR. https://github.com/oppia/oppia-android/pull/3353

WDYT?

viktoriias:
> > Please take a look at the new behavior before I add tests (which sounds like a lot of work):
> > > Regarding testing, once we actually start using announceForAccessibility we should wrap it in our own container, limit usage via regex exemptions, and swap it out with a test-only fake that we can use to determine if it's been called. I think that's the best that we can do here.
> 
> @Viktoriia-Lozova That looks good. Once you have submitted an answer and then you traverse back from top to bottom in recyclerview you will notice that the last answer that they submitted outputs "Correct submitted answer: " which again makes things clear for the user. (We won't need to traverse back from top once #3893 is fixed.)
> 
> Have a look at videos in this PR. #3353
> 
> WDYT?

@BenHenning Do we have similar existing constructs for testing/spying?

BenHenning:
> > > Please take a look at the new behavior before I add tests (which sounds like a lot of work):
> > > > Regarding testing, once we actually start using announceForAccessibility we should wrap it in our own container, limit usage via regex exemptions, and swap it out with a test-only fake that we can use to determine if it's been called. I think that's the best that we can do here.
> > 
> > 
> > @Viktoriia-Lozova That looks good. Once you have submitted an answer and then you traverse back from top to bottom in recyclerview you will notice that the last answer that they submitted outputs "Correct submitted answer: " which again makes things clear for the user. (We won't need to traverse back from top once #3893 is fixed.)
> > Have a look at videos in this PR. #3353
> > WDYT?
> 
> @BenHenning Do we have similar existing constructs for testing/spying?

I agree with @rt4914 that the videos look really good.

@viktoriias we do: FakeAccessibilityChecker & ShadowBidiFormatter are good references. I think this needs to be a fake like the accessibility checker, but the shadow does tracking in a way that's more relevant to what I think you need here (e.g. tracking which strings were announced & in what order).

The other aspect of this to "do it correctly" would be to add new patterns to https://github.com/oppia/oppia-android/blob/develop/scripts/assets/file_content_validation_checks.textproto such that:
1. announceForAccessibility() is prohibited for all uses except the new implementation that you'll be adding
2. All calls to the new utility's announce function should also be prohibited in general so that each use is whitelist approved (to avoid abusing calls to announceForAccessibility; this is something that we want to do, anyway, and given the refactor it seems like a nice time to add that)

One caveat is that we will need to use a different method name than "announceForAccessibility" otherwise the regex will get a bit tricky.

Does this help provide needed context?

BenHenning:
FYI I'm listed as a codeowner for this PR & I'll be unavailable to perform code reviews over the next 2 weeks--thanks for your patience.

viktoriias:
@BenHenning I have a question. If I follow `AccessibilityChecker` example, where would the field of type `AccessibilityChecker` be injected? If I inject it in `StatePlayerRecyclerViewAssembler`, I get `lateinit property has not been initialized` error. Can fields only be injected in objects without constructors, those initialized by the system (Activities, Views)?

BenHenning:
> @BenHenning I have a question. If I follow `AccessibilityChecker` example, where would the field of type `AccessibilityChecker` be injected? If I inject it in `StatePlayerRecyclerViewAssembler`, I get `lateinit property has not been initialized` error. Can fields only be injected in objects without constructors, those initialized by the system (Activities, Views)?

@viktoriias For constructor-injected classes, we always need to add injected fields to the constructor. We only use lateinit var for non-constructor injected classes, and they require setting up an injector which isn't done by Dagger by default, so lateinit var fields will never be initialized in constructor-injected classes (this is more or less what you said, but I added a bit of extra context).

viktoriias:
@BenHenning Please take a look if this is correct setup. If so, we were talking about renaming `AccessibilityChecker` to `AccessibilityService`.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -469,9 +472,15 @@ class StatePlayerRecyclerViewAssembler private constructor(
     val confettiConfig = checkNotNull(congratulationsTextConfettiConfig) {
       "Expected non-null reference to confetti animation configuration"
     }
-
     createBannerConfetti(confettiView, confettiConfig)
     animateCongratulationsTextView(textView)
+
+    if (feedback.html.isBlank()) {
```
BenHenning, line 483: Can you please also update the KDoc to explain what feedback is & how it's used (for context to the readers on how to specify its value)?

domain/src/main/assets/MjZzEVOG47_1.json
```diff
@@ -2694,7 +2694,7 @@
               "dest": "Matthew gets conned",
               "feedback": {
                 "content_id": "feedback_1",
-                "html": "<p>Yes, that's right -- congratulations, you figured out Crumb's trick!</p><p>Let's see what happens to Matthew, though...</p>"
+                "html": " "
```
BenHenning: Nit: prefer empty string for consistency with the textproto version.

That being said, I think it's preferable to set the feedback to null here & absent for textproto (since that's what the backend would generate for such a case). You'll also need to remove any translations or voiceovers corresponding to feedback_1 for this state.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -1233,10 +1243,12 @@ class StatePlayerRecyclerViewAssembler private constructor(
      * answer.
      */
     fun addCelebrationForCorrectAnswers(
+      accessibilityChecker: AccessibilityChecker,
```
BenHenning: I suggest injecting this in the Builder's Factory and passing it along. That way it doesn't need to be null, and it simplifies using this assembler.

Although it means the checker will still be enabled for cases when the celebration isn't enabled, this should be fine since the runtime cost is very low & we'll still not actually use it.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
```diff
@@ -1,7 +1,11 @@
 package org.oppia.android.util.accessibility
 
+import android.view.View
+
 /** Utility for determining various properties of the system's accessibility manager(s). */
 interface AccessibilityChecker {
   /** Returns whether a screen reader (such as TalkBack) is currently enabled. */
   fun isScreenReaderEnabled(): Boolean
+  /** Calls view.announceForAccessibility(text). */
+  fun announceForAccessibilityForView(view: View, text: CharSequence?)
```
BenHenning: I don't think this method actually fits the contract of a 'checker' class. My original intention was for you to either rename this class, or introduce a new one that works similarly to this one.

I think renaming would be a fine option, though it might affect a few other files in the codebase (particularly tests). Perhaps something like 'AccessibilityManager' would be clearer? It sort of accidentally replaces any references to Android's AccessibilityManager since we wouldn't want people to reference that directly, anyway.

Edit: I just saw your earlier comment. I think the approach looks good, and renaming to 'AccessibilityService' also sounds good.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
```diff
@@ -1,7 +1,11 @@
 package org.oppia.android.util.accessibility
 
+import android.view.View
+
 /** Utility for determining various properties of the system's accessibility manager(s). */
 interface AccessibilityChecker {
   /** Returns whether a screen reader (such as TalkBack) is currently enabled. */
   fun isScreenReaderEnabled(): Boolean
+  /** Calls view.announceForAccessibility(text). */
```
BenHenning: We should describe here what the method does from a domain perspective, Perhaps instead:

'Suggests to the screen reader (if any is installed/enabled) to announce the specified text for the given view. This does not guarantee the text will actually be read, and it may interrupt existing text being spoken.'

Also nit: please add newline above this to separate the methods in the interface.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
BenHenning: Also, does CharSequence need to be null? It doesn't seem like we'd ever actually want to pass null in.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityCheckerImpl.kt
```diff
@@ -15,4 +16,8 @@ class AccessibilityCheckerImpl @Inject constructor(
   override fun isScreenReaderEnabled(): Boolean {
     return accessibilityManager.isEnabled
   }
+
+  override fun announceForAccessibilityForView(view: View, text: CharSequence?) {
+    view.announceForAccessibility(text)
```
BenHenning, line 21: Please add a check [here](https://github.com/oppia/oppia-android/blob/develop/scripts/assets/file_content_validation_checks.textproto) to prohibit this from being used anywhere else, and a test [here](https://github.com/oppia/oppia-android/blob/develop/scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt) to verify that check.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
```diff
@@ -1,20 +1,29 @@
...
+
   /**
    * Sets whether a screen reader should be considered currently enabled. This will change the
    * return value of [isScreenReaderEnabled].
    */
   fun setScreenReaderEnabled(isEnabled: Boolean) {
     isScreenReaderEnabled = isEnabled
   }
+
+  override fun announceForAccessibilityForView(view: View, text: CharSequence?) {
```
BenHenning: Given that we're passing the view in, might it be prudent to map the text to the given view?

Further, please also add tests for this new functionality (just to make sure it does what you intend it to).

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
```diff
@@ -1,20 +1,29 @@
...
 /** Fake implementation of [AccessibilityChecker] which should be used in tests. */
 @Singleton
 class FakeAccessibilityChecker @Inject constructor() : AccessibilityChecker {
   private var isScreenReaderEnabled = false
+  private var announcement: CharSequence? = null
 
   override fun isScreenReaderEnabled(): Boolean = isScreenReaderEnabled
 
+  /** Returns announcement. */
+  fun getAnnouncement(): CharSequence? = announcement
```
BenHenning: Perhaps instead: getLatestAnnouncement since the underlying value is overwritten.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
```diff
@@ -1,20 +1,29 @@
...
 
 /** Fake implementation of [AccessibilityChecker] which should be used in tests. */
 @Singleton
 class FakeAccessibilityChecker @Inject constructor() : AccessibilityChecker {
   private var isScreenReaderEnabled = false
+  private var announcement: CharSequence? = null
 
   override fun isScreenReaderEnabled(): Boolean = isScreenReaderEnabled
 
+  /** Returns announcement. */
```
BenHenning: There's more detail that could be added here around how the announcement is defined, and when (i.e. that it's overwritten), plus the fact that it's tied to a given view (depending on whether you implement my suggestion below).

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -352,6 +355,32 @@ class StateFragmentLocalTest {
     }
   }
 
+  @Test
+  @Config(qualifiers = "+port")
+  fun testStateFragment_portrait_submitCorrectAnswer_correctIsAnnounced() {
```
BenHenning, line 388: Can you also add tests for the question player in QuestionActivityTest? You updated a test question but then didn't add a test for it.

BenHenning (changes requested):
Thanks @viktoriias! I think this is definitely on the right track. Most of my feedback is generic around specific parts, but the overall approach looks good. PTAL.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
viktoriias: Done.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
viktoriias: Done.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityCheckerImpl.kt
viktoriias: Done.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
viktoriias: When we discussed this offline, you mentioned possible complications if we do something like that. I don't recall why exactly.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
viktoriias: Added more details.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
viktoriias: Done, moved it.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
viktoriias: CharSequence doesn't need to be null, changed that.
If you don't mind, I will make the renaming change last after everything else is resolved (since it will touch a lot of files).

domain/src/main/assets/MjZzEVOG47_1.json
viktoriias: If I set `"feedback": null`, all tests break. 
I don't think it's necessary to remove all the voiceover and translation, other test depends on those. I think it's fine to have a nonsensical proto for testing purposes, since all we care about is html content?

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
viktoriias: Done.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
viktoriias: Done.

viktoriias:
@BenHenning I addressed your comments, and I will do renaming `AccessibilityChecker > AccessibilityService`  last since it will touch a lot of files.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -1367,12 +1368,14 @@ class StatePlayerRecyclerViewAssembler private constructor(
       private val resourceHandler: AppLanguageResourceHandler,
       private val translationController: TranslationController
     ) {
+      @Inject lateinit var accessibilityChecker: AccessibilityChecker
```
BenHenning: This should be a parameter in the constructor, instead.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -458,6 +458,9 @@ class StatePlayerRecyclerViewAssembler private constructor(
   /**
    * Shows a celebratory animation with a congratulations message and confetti when the learner submits
    * a correct answer.
+   *
+   * @param feedback Oppia's feedback to the learner's most recent answer. If the feedback is empty,
+   * talkback confirms correct answer by announcing "Correct" banner.
```
BenHenning:
```suggestion
   *     talkback confirms correct answers by a separate announcement.
```

Updated to not be overly specific to the implementation, and to indent it per https://developer.android.com/kotlin/style-guide#block_tags.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -26,6 +26,9 @@ class RegexPatternValidationCheckTest {
     "AndroidX should be used instead of the support library"
   private val coroutineWorkerUsageErrorMessage =
     "For stable tests, prefer using ListenableWorker with an Oppia-managed dispatcher."
+  private val announceForAccessibilityUsageErrorMessage =
```
BenHenning, line 29: A test needs to be added below in order to verify this error message (see how the other error messages are used below for a reference).

scripts/assets/file_content_validation_checks.textproto
```diff
@@ -3,6 +3,16 @@ file_content_checks {
   prohibited_content_regex: "^import .+?support.+?$"
   failure_message: "AndroidX should be used instead of the support library"
 }
+file_content_checks {
```
BenHenning, line 6: I was actually thinking we should have two: one that prohibits "announceForAccessibility\(" and suggests using AccessibilityChecker, and a second that checks for "announceForAccessibilityForView\(" with the message you have below. We generally always want people to use the utility, but it still requires approval to avoid over-use.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
```diff
@@ -8,12 +8,15 @@ import javax.inject.Singleton
...
   private var isScreenReaderEnabled = false
-  private var announcement: CharSequence? = null
+  private var announcement: CharSequence = ""
 
   override fun isScreenReaderEnabled(): Boolean = isScreenReaderEnabled
 
-  /** Returns announcement. */
-  fun getAnnouncement(): CharSequence? = announcement
+  /**
+   * Returns latest announcement. Announcement gets overwritten each time
```
BenHenning:
```suggestion
   * Returns latest announcement. Note that the announcement gets overwritten each time
```

(might need line-wrapping adjustment)

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
```diff
@@ -8,12 +8,15 @@ import javax.inject.Singleton
...
+   * announceForAccessibilityForView is called.
+   */
+  fun getLatestAnnouncement(): CharSequence = announcement
```
BenHenning: This might actually be useful to keep as null since then it demonstrates an initial state.

domain/src/main/assets/MjZzEVOG47_1.json
BenHenning: Since you're keeping the feedback, we should also keep the translation and voiceover. It's odd that null breaks it; I suppose that just means that the JSON reader for states isn't robust against this case. Either way, I think leaving this as-is is fine. Thanks for checking!

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityChecker.kt
BenHenning: Yep that sounds good (re: the rename plan)--thanks!

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
BenHenning: Hmm, I don't recall. One potential issue is that tracking views might cause some memory issues, but given this is a test environment that's probably fine. If we're worried about it, though, you could map to something like the view's hashcode. It might have duplicates occasionally, but that should be rare.

Either way, I think it's also fine to keep this initial approach & we can build on it later if needed. Tests would be good to add as well since we'll want them to require updating if we do change this later on.

app/src/test/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityLocalTest.kt
```diff
@@ -166,6 +171,32 @@ class QuestionPlayerActivityLocalTest {
     }
   }
 
+  @Test
+  @Config(qualifiers = "port")
+  fun testQuestionPlayer_portrait_submitCorrectAnswer_correctIsAnnounced() {
```
BenHenning, line 208: For both this & the exploration variant it might be a good idea to add a new test each for verifying that the announcement does *not* happen if there is feedback since that's an important requirement being implemented.

BenHenning (commented):
Thanks @viktoriias! Just a few more comments--PTAL.

BenHenning:

scripts/assets/file_content_validation_checks.textproto
viktoriias: Done, added second check and a test.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
viktoriias: Done.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
viktoriias: Done, I didn't know it could be injected like that.

app/src/test/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityLocalTest.kt
viktoriias: Done, added more tests.

utility/src/main/java/org/oppia/android/util/accessibility/FakeAccessibilityChecker.kt
viktoriias: Ok, back to null

anandwana001:
I will review this today.

anandwana001 (approved):
My code owner file LGTM.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityProdModule.kt
```diff
@@ -7,5 +7,5 @@ import dagger.Module
 @Module
 interface AccessibilityProdModule {
   @Binds
-  fun provideProductionAccessibilityChecker(impl: AccessibilityCheckerImpl): AccessibilityChecker
+  fun provideProductionAccessibilityChecker(impl: AccessibilityServiceImpl): AccessibilityService
```
BenHenning: Nit: please update the provider name.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityTestModule.kt
```diff
@@ -7,5 +7,5 @@ import dagger.Module
 @Module
 interface AccessibilityTestModule {
   @Binds
-  fun provideFakeAccessibilityChecker(impl: FakeAccessibilityChecker): AccessibilityChecker
+  fun provideFakeAccessibilityChecker(impl: FakeAccessibilityService): AccessibilityService
```
BenHenning: Ditto.

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import android.view.View
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.mockito.Mockito.mock
+import org.oppia.android.util.accessibility.FakeAccessibilityService
+
+/** Tests for [FakeAccessibilityService]. */
+class FakeAccessibilityServiceTest {
+  private var accessibilityManager = FakeAccessibilityService()
```
BenHenning: This should be injected instead of constructed (it's more future-proof since we may add injected fields to the constructor in the future).

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+  @Test
+  fun testInitialState() {
```
BenHenning: Prefer splitting this up such that we're verifying the initial state of both properties.

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    assertThat(accessibilityManager.isScreenReaderEnabled()).isFalse()
+    assertThat(accessibilityManager.getLatestAnnouncement()).isNull()
+  }
+
+  @Test
+  fun testSetScreenReaderEnabledTrue() {
```
BenHenning: For all tests in this file: please follow the format ``testAction_withCondition_andCondition_hasOutcome`` & ensure the action, conditions, and outcome are clearly specified for each.

BenHenning (commented):
Hi @viktoriias. Sorry for the delayed review. The PR generally LGTM, just had a few nit comments.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityProdModule.kt
viktoriias: Done.

utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityTestModule.kt
viktoriias: Done.

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
viktoriias: Done.

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
viktoriias: Done

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
viktoriias: Done

viktoriias:
@BenHenning PTAL, failing checks seem unrelated? `Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_cf495b11-c32b-430f-b75e-0be2b55ef582/db5100df-2ebe-4899-b80c-65f91bc396c2.tar.gz. return code: 2.`

BenHenning:
@viktoriias it looks like actions had some issues about an hour ago (https://twitter.com/githubstatus?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Eauthor). I'm restarting the workflows.

BenHenning (approved):
This looks really good, thanks @viktoriias! I think the PR is just blocking on all CI checks passing at this point.

BenHenning:
@rt4914 I believe you still need to approve this--PTAL when convenient.

rt4914 (approved):
Nice clean implementation. Thanks.

viktoriias:
@BenHenning Do you know why I'm getting `WorkManager is already initialized.  Did you try to initialize it manually without disabling WorkManagerInitializer?` error in `FakeAccessibilityServiceTest`?

BenHenning:
> @BenHenning Do you know why I'm getting `WorkManager is already initialized. Did you try to initialize it manually without disabling WorkManagerInitializer?` error in `FakeAccessibilityServiceTest`?

@viktoriias it might be because you're not using a test application (which means Robolectric will use OppiaApplication, and that initializes WorkManager). I'm not certain, but it might be the case that even though Robolectric recreates the test suite object before each test, it still runs in the same process (so static state like WorkManager can be shared across test boundaries). This will result in WorkManager attempting to be initialized multiple times which isn't valid. @Sarthak2601 could probably confirm.

If this is the cause, introducing a test application in your suite & configuring it using ``@Config`` should fix it.

Sarthak2601:
> > @BenHenning Do you know why I'm getting `WorkManager is already initialized. Did you try to initialize it manually without disabling WorkManagerInitializer?` error in `FakeAccessibilityServiceTest`?
> 
> @viktoriias it might be because you're not using a test application (which means Robolectric will use OppiaApplication, and that initializes WorkManager). I'm not certain, but it might be the case that even though Robolectric recreates the test suite object before each test, it still runs in the same process (so static state like WorkManager can be shared across test boundaries). This will result in WorkManager attempting to be initialized multiple times which isn't valid. @Sarthak2601 could probably confirm.
> 
> If this is the cause, introducing a test application in your suite & configuring it using `@Config` should fix it.

Confirming that this is correct and introducing a test application should fix it.

viktoriias:
> > > @BenHenning Do you know why I'm getting `WorkManager is already initialized. Did you try to initialize it manually without disabling WorkManagerInitializer?` error in `FakeAccessibilityServiceTest`?
> > 
> > 
> > @viktoriias it might be because you're not using a test application (which means Robolectric will use OppiaApplication, and that initializes WorkManager). I'm not certain, but it might be the case that even though Robolectric recreates the test suite object before each test, it still runs in the same process (so static state like WorkManager can be shared across test boundaries). This will result in WorkManager attempting to be initialized multiple times which isn't valid. @Sarthak2601 could probably confirm.
> > If this is the cause, introducing a test application in your suite & configuring it using `@Config` should fix it.
> 
> Confirming that this is correct and introducing a test application should fix it.

Thank you, it's passing now!

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
```diff
@@ -64,9 +66,7 @@ class FakeAccessibilityServiceTest {
   }
 
   private fun setUpTestApplicationComponent() {
-    DaggerFakeAccessibilityServiceTest_TestApplicationComponent.builder()
-      .setApplication(ApplicationProvider.getApplicationContext())
-      .build()
+    ApplicationProvider.getApplicationContext<TestApplication>()
       .inject(this)
```
BenHenning: Nit: this can probably fit on one line.

BenHenning (approved):
Thanks @viktoriias! Just one nit. Otherwise, the PR LGTM! Glad to hear that the fix worked.

app/src/test/java/org/oppia/android/app/accessibility/FakeAccessibilityServiceTest.kt
viktoriias: Done

BenHenning (approved):
Thanks @viktoriias. LGTM!

PR #3992, opened 2021-11-10
Fix #3899: Onboarding Flow improvement

### LINKED ISSUE #3899: [A11Y] Onboarding flow improvement
Onboarding flow experience can be improved. Current experience is shown below.

https://user-images.githubusercontent.com/9396084/136245319-7bf6397d-697c-4ed7-a11f-8ddf3a15f6be.mp4

bkaur-bkj:
## Explanation

Fix #3899 
To change accessibility focus to main screen instead of bottom navigation next button in the onboarding flow
## Essential Checklist

### DIFF
--- a/app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
+++ b/app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
@@ -90,8 +90,21 @@ class OnboardingFragmentPresenter @Inject constructor(
           }
           selectDot(position)
           onboardingStatusBarColorUpdate(position)
+          setAccessibilityFocusToViewPager()
         }
       })
+  }
+
+  private fun setAccessibilityFocusToViewPager() {
+//    binding.root.announceForAccessibility(null)
+//
+//    binding.onboardingSlideViewPager.announceForAccessibility(null)
+//
+//    binding.onboardingSlideViewPager.requestFocus()
+//
+//    binding.onboardingSlideViewPager.sendAccessibilityEvent(TYPE_VIEW_CLICKED)
+//
+//    binding.onboardingFragmentNextImageView.sendAccessibilityEvent(TYPE_VIEW_ACCESSIBILITY_FOCUS_CLEARED)
   }
 
   private fun createViewPagerAdapter(): BindableAdapter<OnboardingViewPagerViewModel> {


bkaur-bkj:
@rt4914 sir I tried making a function to change focus of screen reader to main screen but the present options I tried does not work or show any change in screen focus, let me know am I trying in a right way or what else should I try for this change

rt4914:
@bkaur-bkj I will try different solutions mentioned by you tomorrow and get back to you.

bkaur-bkj:
> @bkaur-bkj I will try different solutions mentioned by you tomorrow and get back to you.

Ok sir thanks

rt4914 (changes requested):
@bkaur-bkj 
I have tried different solutions and the final conclusion is as follows:
1. `binding.skipTextView.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)` something like this works perfectly as we are targeting a specifc view but doing same for `ViewPager` which is a `ViewGroup` does not work. It also does not work for any other view group.
2. One possible solution is to get rid of the `BindableAdapter`, create our own custom adapter, and then set the focus on the specifc view (like `slide_title_text_view`) manually which is not a neat solution.

As a result I think its fine to keep the current behavior as final. Even with current behavior we can traverse all views correctly.

In my opinion the cost of changing the code (in a bit untidy way) is not worth it for a feature is a one-time-thing and is normally skipable.

What are your thoughts?

BenHenning:
What precisely is the issue with the existing onboarding experience--ordering? I might be able to weigh in with advice on how to approach the general solution (if one exists), but I'm not sure I fully understand the problem yet @rt4914 (I also left a comment in the issue asking for clarification along the same point).

bkaur-bkj:
> What precisely is the issue with the existing onboarding experience--ordering? I might be able to weigh in with advice on how to approach the general solution (if one exists), but I'm not sure I fully understand the problem yet @rt4914 (I also left a comment in the issue asking for clarification along the same point).

@BenHenning sir Upto what I understood and discussed with @rt4914 sir, the issue is that when we are on screen 1 of onboarding and move to screen 2 through next button the focus  still lies on next button itself for the second page of onboarding and it does not read any content shown on the screen(fragment), thus it makes difficult for a user to know the screen content as the screen just reads "next button" when moved to next screen. to read whats on screen then user needs to make swipes to take focus to fragment from the bottom navigation bar

bkaur-bkj:
> @bkaur-bkj I have tried different solutions and the final conclusion is as follows:
> 
> 1. `binding.skipTextView.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)` something like this works perfectly as we are targeting a specifc view but doing same for `ViewPager` which is a `ViewGroup` does not work. It also does not work for any other view group.
> 2. One possible solution is to get rid of the `BindableAdapter`, create our own custom adapter, and then set the focus on the specifc view (like `slide_title_text_view`) manually which is not a neat solution.
> 
> As a result I think its fine to keep the current behavior as final. Even with current behavior we can traverse all views correctly.
> 
> In my opinion the cost of changing the code (in a bit untidy way) is not worth it for a feature is a one-time-thing and is normally skipable.
> 
> What are your thoughts?

@rt4914 Yes sir exactly as you said in point one it was not working for viewPager.
Also as you said that its just a one time thing and user need not to address this issue frequently so its not much needed to change this at the cost of a bit untidy code. So if we make it final same as it is then we can close the issue then.

BenHenning:
Thanks for clarifying @bkaur-bkj. I'm actually not sure whether we should be proceeding with a change here. The general expectation is to rely on Talkback to do the right thing, and if based on changes in the recycler view doesn't think focus needs to automatically change, then that's probably fine. I know this doesn't match activity navigation, but the user can technically still navigation to the changed elements.

How do other apps handle this case?

bkaur-bkj:
> Thanks for clarifying @bkaur-bkj. I'm actually not sure whether we should be proceeding with a change here. The general expectation is to rely on Talkback to do the right thing, and if based on changes in the recycler view doesn't think focus needs to automatically change, then that's probably fine. I know this doesn't match activity navigation, but the user can technically still navigation to the changed elements.
> 
> How do other apps handle this case?

@BenHenning sir what  I tried out on LinkedIn (as it was listed on google as an app with great onboarding experience). I learnt that there is no such navigation bar there to move on next screen instead we have 3 fragments and it automatically navigates to other fragment  on swipes and reads them all, after that it moves it focus to the signIn by Google button which was constantly present on the bottom of screen below the sliders.

rt4914:
@BenHenning and @bkaur-bkj I think this does not need to be solved as it will make things more complicated and at the same time even the current behavior is good enough and does provide full meaningful access to all views.
Also on other apps is possible because of different fragments for each screen which is not possible for our UI currently.

BenHenning:
Hmm. I think it's fine to close it for now @rt4914 and @bkaur-bkj but I think we should clearly capture the decision in #3899 (specifically, that we're closing it & why we think that's reasonable given current information). We may find as we bring the app to more users that this is actually a bad experience that needs to be fixed in which case having this existing context captured will be helpful. Thanks for digging into this!

rt4914:
> Hmm. I think it's fine to close it for now @rt4914 and @bkaur-bkj but I think we should clearly capture the decision in #3899 (specifically, that we're closing it & why we think that's reasonable given current information). We may find as we bring the app to more users that this is actually a bad experience that needs to be fixed in which case having this existing context captured will be helpful. Thanks for digging into this!

Done.

PR #4038, opened 2021-12-13
Remove self code owner until 27th Dec 2021 

rt4914:
## Explanation
I will have limited availability over the next two weeks, so removing/replacing code owners to ensure we have coverage & availability.

**Code reviewers**: Please confirm your adjusted code owners via approval. If you're ensure of anything, please reach out.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -65,7 +65,7 @@ gradlew.bat @BenHenning
 *.pro @BenHenning
 
 # Lesson assets.
-/domain/src/main/assets/ @BenHenning @BenHenning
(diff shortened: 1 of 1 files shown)


rt4914:
@seanlip @BenHenning PTAL

BenHenning:
Thinking about this more @rt4914 since I'm going to be out after Thursday, I think it might make more sense just to keep things as-is. I'll keep an eye on all PRs independently of this change, and will leave messages to the effect of when folks can expect each of our reviews next (i.e. yours after 27th Dec and mine after 4th Jan for reviews I don't get to this week).

Issue #4044, opened 2021-12-14
Add support for numeric & algebraic expressions/equations interactions support


BenHenning:
Project tracking issue for implementing math-based interactions (https://github.com/oppia/oppia/tree/develop/extensions/interactions -- AlgebraicExpressionInput, MathEquationInput, NumericExpressionInput) per the PRD https://docs.google.com/document/d/1x2vcSjocJUXkwwlce5Gjjq_Z83ykVIn2Fp0BcnrOrTg/edit# and the accessibility version of this feature in Android.

PR #2173, opened 2020-11-26
Fix #4044: Introduce UI support for math expressions & new interactions

BenHenning:
## Explanation
Fixes #4044

This PR introduces the UI interaction views for numeric expressions, algebraic expressions, and algebraic equations, finishing the initial implementation of the broader math expressions project. This work unblocks adding support for 4 additional topics in the app: Multiplication (support was lost after Alpha MR2 due to a change that added numeric expression questions), Addition & Subtraction, Division, and Expressions & Equations.

This PR finishes a long chain of PRs that were needed to provide the domain functionality to support these new interactions, as a significant amount of mathematics functionality needed to be added including:
- Support for representing generic math expressions/equations, polynomials, and reals
- Support for parsing math expressions
- Support for converting between math expressions and:
  - LaTeX (for rendering)
  - Human-readable sentences (for accessibility)
  - Polynomials and real-value evaluation (for equivalence checking)
  - Comparable structures to verify that two expressions/equations match irrespective of associativity or commutativity order
- Robust error detection to support nearly 2 dozen special cased errors with potential for countless more if needed to ensure users receive excellent feedback when inputting expressions

All of the above also required thorough testing to ensure correctness. See the previous PRs corresponding to #4044 for the full context, and thorough PR descriptions explaining past changes.

From a UI perspective, this PR is introducing:
- The new interaction views and relevant view models, including the logic for constructing and displaying errors for invalid answers
- Support for playing a custom content description for math expression answers so that screenreader users can better understand the expression they've entered
- Support for rendering the LaTeX representation of entered expressions
- A new developer options menu to provide support to input arbitrary expressions and equations and convert them to different outputs to test the underlying math engine that's been built (though it can't yet test the classifiers, but it provides enough information to verify the more complex bits behind the classifiers)

From a functional perspective, this PR is introducing:
- Support for loading JSON explorations that include math expressions
- A new test exploration that's been added to test topic 0 to demonstrate all three interactions and each of those interaction's classifiers (9 total states have been added)

The above, and more, are explained in more detail below.

### Background on UI implementation for new interactions

The UI implementation leverages a single new interaction view & model for all three new interaction types since they are very similar to one another, including handling accessibility and LaTeX rendering cases. The customization arguments, however, don't exactly match so there are some inconsistencies there.

### Background on parsing approach
[...]

PR #4047, opened 2021-12-15
Fix part of #4044: Add protos & testing libraries for math expressions/equations

BenHenning:
## Explanation
Fix part of #4044

This PR introduces several new math protos:
- Real
- MathExpression (& its constituent pieces: MathBinaryOperation, MathUnaryOperation, and MathFunctionCall)
- MathEquation

The expression & equation structures are necessary for downstream numeric expression, algebraic expression, and algebraic equation parsing, as well as for implementing the MatchesExactlyWith classifiers for each interaction. For specific details on this classifier, see [the PRD](https://docs.google.com/document/d/1x2vcSjocJUXkwwlce5Gjjq_Z83ykVIn2Fp0BcnrOrTg/edit#heading=h.9mdskev1k1ax). None of that code is introduced in this PR, but will be in future PRs.

The ``Real`` structure being introduced is to help with full and partial expression evaluation that will be introduced in a future PR to ensure infinite precision can be kept for as long as possible (i.e. by keeping values in the form of integers or fractions).

Some caveats regarding the expression structure:
- Parsing information is included for token & subexpression extraction which is needed for certain user errors that will be implemented as part of the interaction UI implementations in a future PR
- A single generic structure is used to represent both numeric and algebraic expressions to simplify a lot of code (this will be more evident in future PRs)

Beyond the protos, new Truth test subject classes are being introduced to make verifying these values easier. Part of this includes a custom Kotlin DSL for verifying expressions & equations in a clean way, e.g.:

```kotlin
    val equation = parseAlgebraicEquationSuccessfully("y = (x+1)(x-1)")
    assertThat(equation).hasLeftHandSideThat().hasStructureThatMatches {
      variable {
        withNameThat().isEqualTo("y")
      }
    }
    assertThat(equation).hasRightHandSideThat().hasStructureThatMatches {
      multiplication {
        leftOperand {
          group {
            addition {
              leftOperand {
                variable {
                  withNameThat().isEqualTo("x")
                }
              }
              rightOperand {
                constant {
                  withValueThat().isIntegerThat().isEqualTo(1)
                }
              }
            }
          }
        }
        rightOperand {
          group {
            subtraction {
              leftOperand {
                variable {
                  withNameThat().isEqualTo("x")
                }
              }
              rightOperand {
                constant {
                  withValueThat().isIntegerThat().isEqualTo(1)
                }
              }
            }
          }
        }
      }
    }
```

The main benefit to the DSL is to avoid cascading checks using subsequent assertion lines that can become quite difficult to read, e.g.:

```kotlin
    val rootExpression = result.getExpectedSuccessfulExpression()
    val mulOp = rootExpression.getExpectedBinaryOperationWithOperator(MULTIPLY)
    val leftConstant = mulOp.leftOperand.getExpectedRationalConstant()
    val rightAddOp = mulOp.rightOperand.getExpectedBinaryOperationWithOperator(ADD)
    val rightAddLeftVar = rightAddOp.leftOperand.getExpectedVariable()
    val rightAddRightConstant = rightAddOp.rightOperand.getExpectedRationalConstant()
    assertThat(leftConstant).isEqualTo(createWholeNumberFraction(2))
    assertThat(rightAddLeftVar).isEqualTo("x")
    assertThat(rightAddRightConstant).isEqualTo(createWholeNumberFraction(1))
```

The above is for the expression "2(x+1)" and was a previous attempt. It leverages helpers and to me seems much harder to read as compared with the DSL. The DSL does result in more verbosity, but I think that actually helps readability in this case.

### Test exemptions
Note that I did not add tests for these. Given that they're test utilities & the importance of getting the downstream work completed, I decided to file #4097 to track adding tests for each. I've already manually verified the subjects work in actual tests downstream, and the code is simple enough that I'm confident in its correctness by inspection. Nevertheless, we should eventually add automated tests for this. Reviewers: please let me know if you have any concerns with this approach.

## Essential Checklist

## For UI-specific PRs only


BenHenning:

BenHenning:
PTAL @vinitamurthi for codeowners, and @anandwana001 for both codeowners and for general math expressions reviewing.

model/src/main/proto/math.proto
```diff
@@ -5,17 +5,186 @@ package model;
...
+// Represents a mathematical expression such as 1+2*7. Expressions are inherently recursive, so the
+// overall expressiveness of this structure (& math expressions as a whole) is defined based on its
+// constituent substructures.
+//
+// This structure is designed to represent both numeric and algebraic expressions.
+//
+// Values of this proto can be analyzed using MathExpressionSubject.
+message MathExpression {
+  // The index within the input text stream at which point the expression starts (it's an inclusive
+  // index). If both this and the end index are zero then no parsing information is included for
```
vinitamurthi:
```suggestion
  // index). If both this and the end index are the same then no parsing information is included for
```

It can be a value other than zero as well right?

model/src/main/proto/math.proto
```diff
@@ -5,17 +5,186 @@ package model;
...
+  // this specific expression.
+  uint32 parse_start_index = 1;
+
+  // The index within the input text stream at which point the expression ends, exclusively. If both
+  // this and the start index are zero then no parsing information is included for this specific
```
vinitamurthi: Same comment here

model/src/main/proto/math.proto
```diff
@@ -5,17 +5,186 @@ package model;
...
+  // Types of supported unary operations.
+  enum Operator {
+    // Represents an unknown operator (which is never supported).
+    OPERATOR_UNSPECIFIED = 0;
+
+    // Represents negating a value, e.g.: -y.
+    NEGATE = 1;
+
+    // Represents indicating a value as positive, e.g.: +y.
+    POSITIVE = 2;
```
vinitamurthi, line 149: Do we actually need this? Can we not say that anything thats not negative will be a positive?

model/src/main/proto/math.proto
```diff
@@ -5,17 +5,186 @@ package model;
...
+
+  // Defines the denominator of the fraction. This should never be zero.
+  uint32 denominator = 4;
 }
 
-// Structure containing a ratio object for eg - [1,2,3] for 1:2:3.
+// Represents a structured real value.
+//
+// Values of this proto can be analyzed using RealSubject.
+message Real {
```
anandwana001, line 28: can we update these message with `DTO` suffix?

anandwana001 (approved):
LGTM

model/src/main/proto/math.proto
BenHenning: These aren't data transfer objects, they're the actual domain representations for these values.

model/src/main/proto/math.proto
BenHenning: Good point; updated to your suggestion.

model/src/main/proto/math.proto
BenHenning: Done.

model/src/main/proto/math.proto
BenHenning: The grammar defines positives as valid expressions for three reasons:
1. This is conventional with other calculators (such as the Ubuntu calculator, Google's, and Wolfram Alpha)
2. It provides parity with unary plus (consider the parity between addition & subtraction)
3. It makes it more straightforward to cleanly detect this as an error case (since we don't actually want learners to do this since it's likely an incomplete addition operation). Rather than special casing a specific invalid token case we can do a post-parsing analysis for this type.

I suggest looking at the tests in #4052 if you want to see examples where unary plus is used.

BenHenning (commented):
Thanks @vinitamurthi & @anandwana001. I addressed all open comments--PTAL.

Also, I synced with develop. I intentionally committed updates before the develop merge so you can view that directly if the merge messes up your diff view.

vinitamurthi (approved):

BenHenning:
This is still active.

PR #4049, opened 2021-12-15
Fix part of #4044: Add protos & testing library for commutative expressions/operations

BenHenning:
## Explanation
Fix part of #4044

This PR introduces the proto & Truth subject for representing math expressions in a way that makes them trivial to compare without commutative or associativity changing the meaning of the operation. See #4047 for details about the test exemptions being made here, and the custom Kotlin DSL. Note that #4098 is tracking adding tests for the new test subject.

The way the structure introduced here works is that it provides a structure for math expressions that collects "accumulations" (i.e. products or summations). Since both multiplication and addition are commutative operators, elements of these accumulation lists can be rearranged (and thus sorted for deterministic sorting that doesn't fail upon commutative reordering). Associativity complicates this since we actually want expressions like 3-4+7 to be equivalent to 7+3-4. In order for that to be the case, the structure treats subtraction as negative addition and division as inverted multiplication so that division and subtraction terms can also be rearranged. This leaves only square roots & exponentiations as non-commutative operations (which are more or less retained in the same hierarchical structure as math expressions).

This structure will be utilized in a later PR when a routine for converting ``MathExpression``s to ``ComparableOperation``s, and then later in MatchesUpToTrivialManipulations classifier implementations for each numeric expression, algebraic expression, and math equation interaction. For specific details on this classifier, see [the PRD](https://docs.google.com/document/d/1x2vcSjocJUXkwwlce5Gjjq_Z83ykVIn2Fp0BcnrOrTg/edit#heading=h.g0bd549g079d).

## Essential Checklist

## For UI-specific PRs only


BenHenning:

BenHenning:
@anandwana001 PTAL for review as part of the math expressions/equations project.

model/src/main/proto/math.proto
```diff
@@ -188,3 +188,105 @@ message MathEquation {
...
+      // operations of the same type that are commutative, e.g. addition or multiplication).
+      CommutativeAccumulation commutative_accumulation = 3;
+
+      // Indicates that this operation is a non-commutative operation and thus cannot be
+      // accumulated (e.g. exponentiation).
+      NonCommutativeOperation non_commutative_operation = 4;
+
+      // Indicates that this operation represents a constant value.
+
+      Real constant_term = 5;
```
anandwana001:
```suggestion
      // Indicates that this operation represents a constant value.
      Real constant_term = 5;
```

model/src/main/proto/math.proto
```diff
@@ -188,3 +188,105 @@ message MathEquation {
...
+
+    // The supported comparison types.
+    oneof comparison_type {
+      // Indicates that this operation is a commutative accumulation (that is, a list of subsequent
+      // operations of the same type that are commutative, e.g. addition or multiplication).
+      CommutativeAccumulation commutative_accumulation = 3;
+
+      // Indicates that this operation is a non-commutative operation and thus cannot be
+      // accumulated (e.g. exponentiation).
+      NonCommutativeOperation non_commutative_operation = 4;
```
anandwana001: isn't there assigned numbers that should be 1 and 2 rather than 3 and 4 and below fields too, as these are parts of `oneOf`?

model/src/main/proto/math.proto
```diff
@@ -188,3 +188,105 @@ message MathEquation {
...
+//
+// 'Comparable' here means that this structure provides a trivial way to compare commutative and
+// associative operations (i.e. by extracting terms from multiple subsequent commutative &
+// associative operations into lists that can be deterministically sorted). This structure is meant
+// to provide a means to compare two expressions without considering associativity or commutativity
+// (though the latter requires the operation lists stored within this structure to be sorted before
+// using standard proto equals checking). Also note that care must be taken when performing equality
+// checking since this structure can contain floating point values that require an epsilon check to
+// approximate equality.
+message ComparableOperationList {
```
anandwana001: As this message name says it is a list so either this message will be repeated somewhere else or inside this message we need a `repeated` value to make it a list!!

What I am understanding here is, this message is a list and list contains multiple items, so inside this message there should be a `repeated` field. 
```proto
repeated ComparableOperation root_operation = 1;
```

anandwana001 (approved):
Overall LGTM,

left some nit thoughts below.

model/src/main/proto/math.proto
BenHenning: I guess I think of this conceptually as a list even though it's a mix between a tree and a list. Do you have a suggestion for a clearer name?

@vinitamurthi as well in case you have any thoughts.

model/src/main/proto/math.proto
BenHenning: The oneof shares indexes with the broader message since it actually is assigning a message-level field. Since ``is_inverted`` has index 2, the fields defined in the oneof need to start at index 3.

model/src/main/proto/math.proto
BenHenning: Ah I noticed this in passing as well--fixed.

BenHenning (commented):
Thanks @anandwana001. I followed up with replies--PTAL.

model/src/main/proto/math.proto
BenHenning: When working on a later PR (that actually introduces the computations for this structure), I thought about this again & decided to remove the outer list wrapper proto (especially before I went too deep into the testing to save time later on). PTAL and let me know your thoughts, and whether the latest changes address your comment.

anandwana001: looks good to me. 
Keeping this conversation open for @vinitamurthi

BenHenning:
This is still active.

BenHenning:
Force-merging this and tracking @vinitamurthi's missing approval in #4265 (which was agreed upon ahead of time).

PR #4050, opened 2021-12-15
Fix part of #4044: Add protos & testing library for polynomials

BenHenning:
## Explanation
Fix part of #4044

This PR introduces the protos and test subject to represent polynomials.

For the purposes of upcoming classifier work, a polynomial is defined as a sum of terms where each term has a coefficient and zero or more variables with positive integer powers. This representation provides support for all real polynomials, and keeps a nicely structured representation for coefficients that actually allows for retaining integers and rational values (this will become more evident when math expression -> polynomial conversion is added).

Furthermore, various extensions files are added to help support some of the fundamental operations (mainly needed by test subjects). These operations are being thoroughly tested, and will be augmented with a lot more functionality in upcoming PRs.

The new polynomial test subject doesn't have new tests to keep this PR focused on production code, and since it's relatively easy to verify as correct via review. #4100 is tracking adding tests.

This structure will be utilized in a later PR in IsEquivalentTo classifier implementations for each numeric expression, algebraic expression, and math equation interaction. For specific details on this classifier, see [the PRD](https://docs.google.com/document/d/1x2vcSjocJUXkwwlce5Gjjq_Z83ykVIn2Fp0BcnrOrTg/edit#heading=h.1q3av9yssyi5). Polynomials are the ideal structure for verifying equivalence since they fully collapse expressions regardless of associativity, commutativity, and distributivity (to some extent--caveats will be noted in the future PR that converts expressions to this new polynomial structure).

Slightly separate from polynomials, ``FloatExtensions`` was updated to include better epsilon values for comparing both floats and doubles (and a separate one is used for doubles). These are loosely based on the computed machine epsilon value listed here: https://en.wikipedia.org/wiki/Machine_epsilon, but it uses a higher order of magnitude and rounding for a bit more "wiggle room" for equality (so that it's not essentially replicating '==' for values close to 1).

## Essential Checklist

## For UI-specific PRs only


BenHenning:

BenHenning:
@vinitamurthi PTAL for codeowners
@anandwana001 PTAL for codeowners & as part of the broader math expression reviews (so please make sure to review the whole PR)

anandwana001:
Reviewing it now.

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
```diff
@@ -0,0 +1,156 @@
...
+    assertWithMessage(
+      "Expected polynomial to be undefined, but was: ${actual?.toPlainText()}"
+    ).that(actual).isNull()
+  }
+
+  /**
+   * Verifies that the represented [Polynomial] is a constant (i.e. [Polynomial.isConstant] and
+   * returns a [RealSubject] to verify the value of the constant polynomial.
+   */
+  fun isConstantThat(): RealSubject {
```
anandwana001, line 49: The prefix we are using `*That` sounds a bit different to me. I don't have any specific suggestions for this.

utility/src/main/java/org/oppia/android/util/math/PolynomialExtensions.kt
```diff
@@ -0,0 +1,57 @@
...
+ * The returned string is guaranteed to be a syntactically correct algebraic expression representing
+ * the polynomial, e.g. "1+x-7x^2").
+ */
+fun Polynomial.toPlainText(): String {
+  return termList.map { it.toPlainText() }.reduce { acc, termAnswerStr ->
+    if (termAnswerStr.startsWith("-")) {
+      "$acc - ${termAnswerStr.drop(1)}"
+    } else "$acc + $termAnswerStr"
+  }
+}
```
anandwana001:
```suggestion
fun Polynomial.toPlainText(): String {
  return termList.map { term ->
    term.toPlainText()
  }.reduce { acc, termAnswerStr ->
    if (termAnswerStr.startsWith("-")) {
      "$acc - ${termAnswerStr.drop(1)}"
    } else "$acc + $termAnswerStr"
  }
}
```

anandwana001: Also, if we can give better name to `acc`.

utility/src/main/java/org/oppia/android/util/math/PolynomialExtensions.kt
```diff
@@ -0,0 +1,57 @@
...
+  val productValues = mutableListOf<String>()
+
+  // Include the coefficient if there is one (coefficients of 1 are ignored only if there are
+  // variables present).
+  productValues += when {
+    variableList.isEmpty() || !abs(coefficient).isApproximatelyEqualTo(1.0) -> when {
+      coefficient.isRational() && variableList.isNotEmpty() -> "(${coefficient.toPlainText()})"
+      else -> coefficient.toPlainText()
+    }
+    coefficient.isNegative() -> "-"
```
anandwana001, line 46: How about keeping this as compile-time constant, and the one below?
```kt
const val NEGATIVE_COEFFICIENT = "-"
```

utility/src/main/java/org/oppia/android/util/math/RatioExtensions.kt
```diff
@@ -21,3 +21,8 @@ fun RatioExpression.toSimplestForm(): List<Int> {
 fun RatioExpression.toAnswerString(): String {
   return ratioComponentList.joinToString(separator = ":")
 }
+
+/** Returns the greatest common divisor between two integers. */
+private fun gcd(x: Int, y: Int): Int {
```
anandwana001, line 26: As we are using GCD operation in RatioExtension and FractionExtension, this might scale in the future, so can we keep this in some file called `CommonExtensions` under math package only?

utility/src/main/java/org/oppia/android/util/math/RealExtensions.kt
```diff
@@ -0,0 +1,84 @@
...
+ * like '1 1/2' can't be written as a numeric expression without converting them to an improper
+ * form: '3/2').
+ */
+fun Real.toPlainText(): String = when (realTypeCase) {
+  // Note that the rational part is first converted to an improper fraction since mixed fractions
+  // can't be expressed as a single coefficient in typical polynomial syntax).
+  RATIONAL -> rational.toImproperForm().toAnswerString()
+  IRRATIONAL -> irrational.toPlainString()
+  INTEGER -> integer.toString()
+  REALTYPE_NOT_SET, null -> ""
```
anandwana001, line 55: Can we use some const here, making it more readable and understandable that why we are passing an empty string here?

utility/src/test/java/org/oppia/android/util/math/FloatExtensionsTest.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    val result = leftFloat.approximatelyEquals(rightFloat)
+
+    assertThat(result).isFalse()
+  }
+
+  @Test
+  fun testFloat_approximatelyEquals_nonZeroValues_veryDifferent_returnsFalse() {
+    val leftFloat = 1.2f
+    val rightFloat = 7.3f
```
anandwana001, line 71: Do we also need a test case where:
```kt
val leftFloat = 0f
val rightFloat = 7.3f
```
returns false?

utility/src/test/java/org/oppia/android/util/math/RealExtensionsTest.kt
```diff
@@ -0,0 +1,414 @@
...
+  integer = value
+}.build()
+
+private fun createRationalReal(value: Fraction) = Real.newBuilder().apply {
+  rational = value
+}.build()
+
+private fun createIrrationalReal(value: Double) = Real.newBuilder().apply {
+  irrational = value
+}.build()
```
anandwana001, line 414: What if we keep these functions static only (inside companion object)?

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
```diff
@@ -0,0 +1,156 @@
...
+   */
+  fun term(index: Int): PolynomialTermSubject = assertThat(nonNullActual.termList[index])
+
+  /**
+   * Returns a [StringSubject] to test the plain-text representation of the [Polynomial] (i.e. via
+   * [Polynomial.toPlainText]).
+   */
+  fun evaluatesToPlainTextThat(): StringSubject = assertThat(nonNullActual.toPlainText())
+
+  companion object {
```
anandwana001, line 78: Can we shift the companion object to the top?

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Verifies that the represented [Polynomial] is a constant (i.e. [Polynomial.isConstant] and
+   * returns a [RealSubject] to verify the value of the constant polynomial.
+   */
+  fun isConstantThat(): RealSubject {
+    assertWithMessage("Expected polynomial to be constant, but was: ${nonNullActual.toPlainText()}")
+      .that(nonNullActual.isConstant())
+      .isTrue()
+    return assertThat(nonNullActual.getConstant())
```
anandwana001, line 53: How about we remove the import and just use the `assertThat` in the below-mentioned fashion? In this way, this is more readable and will not confuse future contributors which `assertThat` is of which subject!! 

```suggestion
    return RealSubject.assertThat(nonNullActual.getConstant())
```

anandwana001 (approved):
Left some nit thoughts on PR. Most of them are just suggestions to improve the readability and understandability.
Overall LGTM.

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
BenHenning: It's conventional for Truth methods that are chaining so that the final assertion reads like a sentence. See examples of this being used, as well as how other Truth subjects are named. Does it seem reasonable to you?

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
BenHenning: It's generally conventional to directly import assertThat(), but as it's within a subject I'm not opposed to this. That being said, the return type is RealSubject which means it has to be RealSubject's assertThat. Do you feel strongly about this being changed?

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
BenHenning: For what reason? The KDoc at the top has a clickable reference to the assertThat, plus it's fairly conventional to expect it to be within the subject class.

utility/src/main/java/org/oppia/android/util/math/PolynomialExtensions.kt
BenHenning: Done (both).

utility/src/main/java/org/oppia/android/util/math/PolynomialExtensions.kt
BenHenning: Not keen on this since we actually have multiple symbols to represent negatives and minus.

utility/src/main/java/org/oppia/android/util/math/RatioExtensions.kt
BenHenning: I thought about this, but my main hesitation is that it's a small method to copy & I couldn't come up with a good place to keep it (it's not really an extension, and I'd rather avoid an overly generic utility file to place it in).

utility/src/main/java/org/oppia/android/util/math/RealExtensions.kt
BenHenning: I added a KDoc and inline comment to contextualize this--is it clearer?

utility/src/test/java/org/oppia/android/util/math/FloatExtensionsTest.kt
BenHenning: Sure that seems sensible--done.

utility/src/test/java/org/oppia/android/util/math/RealExtensionsTest.kt
BenHenning: That would be my preference, too, but classes can only have one companion object and I put it at the top for constants. I didn't want the helpers up there since they're more appendices. This seemed like the best alternative. Any suggestions?

BenHenning (commented):
Thanks for the review @anandwana001. I followed up to your comments, PTAL and resolve any that look good or let me know your follow-up thoughts.

Also, I merged in changes & develop from upstream, but I addressed your comments before that in case you want to view that commit in isolation (just in case GitHub's diff view is weird).

BenHenning:
Note that I updated the PR to include better epsilon checking--see PR description for details.

BenHenning:
Note that I removed a test from NumericInputEqualsRuleClassifierProviderTest since it was testing an edge case that's not actually useful to verify (border values) since the new epsilons allow for very small margins when comparing values (so computing a border value actually runs into a high risk of precision loss).

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
anandwana001: sounds good.

utility/src/test/java/org/oppia/android/util/math/RealExtensionsTest.kt
anandwana001: I guess, in that case let's keep it here only.

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
anandwana001: Got it

testing/src/main/java/org/oppia/android/testing/math/PolynomialSubject.kt
anandwana001: I think we can skip this, as it doesn't bring much improvisation.

vinitamurthi (approved):
LGTM for codeowners

BenHenning:
This is still active.

PR #4051, opened 2021-12-15
Fix part of #4044: Add math tokenizer & parameterized test support

BenHenning:
## Explanation
Fix part of #4044

This PR principally introduces the lexical tokenizer (lexer) for math expressions, as defined by the [formal grammar](https://docs.google.com/document/d/1JMpbjqRqdEpye67HvDoqBo_rtScY9oEaB7SwKBBspss/edit). The tokenizer converts a string into a sequence of ``Token``s which provide context to the characters available in the string. While the tokenizer is not yet hooked up, it will be leveraged in an upcoming PR to add support for parsing math expressions and equations. Note that special care was taken to ensure the parser is an LL(1) parser to simplify implementation, ease future grammar additions, and to allow for a table-based parsing method in the future if increased performance is necessary.

Some caveats of the tokenizer:
- It's implemented on top of a sequence with lazy iteration to ensure that no processing time is spent parsing more than necessary (i.e. if the parser short-circuits due to an error then no additional tokenization cost is paid)
- Due to a limitation in the grammar being LL(1), function name parsing is a bit more nuanced and limiting. The lexer has to assume that the start of a function name is actually a name and not a series of variables. To account for this, a special token exists for invalid function names. See the tokenizer's test suite for more details on valid and invalid character combinations.
- Care is taken to ensure that overflowed integers & doubles result in failing tokens
- The lexer takes on a bit of the actual parsing (for integers & reals) to simplify the parser, but the majority of parsing is not done in the lexer

This PR introduces a new Truth subject for the tokenizer's ``Token`` class, though similar to subjects introduced in previous PRs this one does not include tests. #4121 is tracking adding them in the future.

To enforce the LL(1) grammar, the tokenizer makes use of a PeekableIterator that doesn't allow for lookahead beyond one character. This iterator will also be used by the upcoming parser to keep the LL(1) property.

To enable better testing for the tokenizer, I decided to add parameterized tests. Unfortunately, there wasn't an obvious way on how to exactly do this. While parameterized JUnit tests work, there are some limitations:
- They won't work with instrumentation or Robolectric
- They don't support combining parameterized & non-parameterized tests

Instead, I decided to implement a custom parameterized test runner that addresses both of the points above. I used ``AndroidJUnit4`` and ``ParameterizedRobolectricTestRunner`` as key references into understanding how to actually build this. While the implementation we're now using is quite different than both, they serve as parts of the basis. ``AndroidJUnit4`` is final, so we needed to implement a similar switching mechanism to select Robolectric or Espresso test helpers based on the platform (this is controlled via a test class-level annotation). Further, Robolectric's runner was important to understand how to even set up a parameterized runner, and to better understand how to make Robolectric work in this environment.

The result seems to be a very clean test runner, but I'm keen to hear reviewer's thoughts on this. Note a couple things:
- While there's an Espresso-supported runner, I didn't actually verify that it works since we haven't yet gotten instrumentation tests working natively with Bazel (though I do expect that it'll work)
- I didn't test this on Gradle, but per the CI run it appears to work just fine
- There are no tests for any of the new runners or their utilities. The utilities are fairly trivial, and the runners are difficult to test. I decided not to test them or file a tracking issue as they'll likely be the last components we add tests for when we aim to raise the codebase's code coverage (I didn't feel the difficulty was worth overcoming here over vs. manual verification).
- The runner has very robust error detection to try and reduce potential errors, including enforcing parameter order consistency
- I ran into a little snag with ktlint--see #4122
- The previous iteration of this led to platform selection between Espresso & Robolectric based on Bazel dependencies, but I decided to change this to be an explicit annotation. This has a number of considerations:
  - It leads to automatic build fixes (i.e. you can't compile a reference to a class without the Bazel dependency being complete)
  - It allowed me to introduce a JUnit-specific test suite (after changing MathTokenizerTest to this, I saw an almost 40x decrease in runtime; I plan to use this for all future parameterized math tests since Robolectric isn't actually needed for these and is **much** slower)
  - It doesn't facilitate shared tests like AndroidJUnit4, but we could conceivably create a shared runner in the future that uses either the instrumentation or Robolectric runner based on the current platform (similar to how AndroidJUnit4 works--see the commit history for the old code that did this)
- Finally, see the KDoc for ``OppiaParameterizedTestRunner`` for much more details on the runner, including both a code example and suggestions on when to actually use the runner. Further, ``MathTokenizerTest`` demonstrates real uses of the runner.

The way that the runner works is it generates a test suite that combines one runner for all non-parameterized tests with a runner for each parameterized iteration (so the total number of tests in a suite is ``number_of_non_parameterized_tests + sum(parameterized_iterations_in_all_parameterized_tests)``. The iteration names are appended to parameterized test's names to provide both Bazel filtering support and direct error reporting when tests fail (so that the exact iteration is known & can be run/debug in isolation). Android Studio will run all iterations when filtering on a method (since that should match against all), but this might be environment-specific (I verified this with Bazel in Android Studio, but not with Gradle).

### Script asset changes
Test file exemptions were added for the new JUnit rules & the TokenSubject class--see above for the rationale.

Regex content exemptions were added for ``ParameterizedMethod`` since it uses ``Locale`` and ``capitalize``. It doesn't have access to ``MachineLocale`` (at least, currently), and is only running in tests (so localization correctness isn't as important). Finally, the chosen locale to use for capitalization is ``US`` which should more or less match field names in tests for the Oppia codebase.

Further, a new regex check was added to require all uses of the new test runner to require approval (i.e. by adding an exemption). This will help ensure it doesn't get abused/used too broadly since parameterization should be an exceptional case rather than the norm.

## Essential Checklist

## For UI-specific PRs only
N/A -- This PR introduces a non-user facing utility, and the utility isn't yet hooked up (it will be in a subsequent PR).


BenHenning:

BenHenning:
@anandwana001 PTAL at the whole PR as part of the broader math expressions work, and for your codeowners.

@vinitamurthi optional, but could you take a look at the parameterized test runner & pieces? I'd like your thoughts on the approach.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * The test testParams_multipleVals_isConsistent will be run three times, and before each time the
+ * specified parameter value corresponding to each iteration will be injected into the parameter
+ * field for use by the test.
+ *
+ * Note that with Bazel specific iterations can be run by utilizing the test and iteration name,
+ * e.g.:
+ *
+ * ```bash
+ * bazel run //...:ExampleParameterizedTest --test_filter=testParams_multipleVals_isConsistent-first
```
anandwana001: Should we add a line explaining `//...:` how does the path will come here?
Currently, it is not mentioned in the [wiki](https://github.com/oppia/oppia-android/wiki/Oppia-Bazel-Setup-Instructions#building-the-app) how to run a single test.

utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt
```diff
@@ -0,0 +1,441 @@
...
+          '^' -> tokenizeSymbol(chars) { startIndex, endIndex ->
+            Token.ExponentiationSymbol(startIndex, endIndex)
+          }
+          '=' -> tokenizeSymbol(chars) { startIndex, endIndex ->
+            Token.EqualsSymbol(startIndex, endIndex)
+          }
+          '(' -> tokenizeSymbol(chars) { startIndex, endIndex ->
+            Token.LeftParenthesisSymbol(startIndex, endIndex)
+          }
+          ')' -> tokenizeSymbol(chars) { startIndex, endIndex ->
```
anandwana001: Should we create some static fields for these variables? '=', ')', etc....

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
```diff
@@ -0,0 +1,303 @@
...
+   *
+   * @property name the name of the iteration (this should be short, but meaningful since it's
+   *     appended to the test name)
+   * @property keyValuePairs an array of strings of the form "key=value" where 'key' is the name of
+   *     a [Parameter]-annotated field and 'value' is a stringified conforming value based on the
+   *     type of that field (incompatible values will result in test failures)
+   */
+  @Repeatable(RunParameterized::class)
+  @Target(AnnotationTarget.FUNCTION)
+  annotation class Iteration(val name: String, vararg val keyValuePairs: String)
```
anandwana001, line 265: How are we handling the use case where if any single test gets failed in the iteration or is it handled by `AndroidJUnit4ClassRunner` only?

anandwana001 (changes requested):
I will be looking into this PR in 2 parts (a lot of engineering in one go), for the first, I had covered things around the param iteration way of testing. Loved it very much :heart: 
For the second part, I will be understanding the `TokenSubject` and its relatable files. 

Putting request change, as the second part of the review is left to be done. Left some nit thoughts.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
BenHenning: This feels like something we should fix in the wiki, not in the documentation. The reason I didn't add the fully qualified path is because it could change.

@FareesHussain is this something that we could add to the documentation? Maybe a reference of different useful Bazel command how-tos?

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
BenHenning: If one of the iterations fail, it'll show up as the test's name following by a dash and the name of the iteration. You can actually directly run that iteration with Bazel  (and presumably Gradle, but I didn't try).

I'm not sure that I understand the second part of your question--what do you mean by a test being handled by ``AndroidJUnit4ClassRunner`` only?

utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt
BenHenning: I personally find it more readable to see the actual symbols being parsed, but it might be worth cross referencing for someone else's thoughts as far as readability is concerned if we differ on this point.

Beyond that, do you see any specific advantage to using constants?

BenHenning (commented):
Thanks for the first pass @anandwana001. FYI I also updated the PR to include a regex check to by-default prohibit using parameterized tests (see the PR description & changes for specifics).

BenHenning:
Note: I decided to change the delimiter to '_' instead of '-' so that it doesn't interfere as much with the regex matcher that Bazel uses for test_filter, and for better parity with our test name style.

BenHenning:
Note that I updated the PR to:
- Have the runner require explicit selection of platform
- Add support for a JUnit-only runner

See PR description for details.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
anandwana001: Keeping this conversation open for @FareesHussain

utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt
anandwana001: Not exactly, this readability point sounds good. Let's keep it unchanged.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
anandwana001: I was guessing `AndroidJUnit4ClassRunner` is somehow helping us here in the iteration fail part.

anandwana001 (approved):
LGTM

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
```diff
@@ -0,0 +1,303 @@
...
+ *
+ *   @Test
+ *   @RunParameterized(
+ *     Iteration("first", "parameter=first value"),
+ *     Iteration("second", "parameter=second value"),
+ *     Iteration("third", "parameter=third value")
+ *   )
+ *   fun testParams_multipleVals_isConsistent() {
+ *     val result = performOperation(parameter)
+ *     assertThat(result).isEqualTo(consistentExpectedValue)
```
vinitamurthi: It may be worth explaining how to pass parameters that are strings as well . ANd to document that only primitives are allowed as parameters. But other than that I think this is great

vinitamurthi (approved):

BenHenning:
This is still active.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
BenHenning: Thanks for the feedback! I've updated the documentation to mention that only primitives are allowed, but I'm not sure what you mean by the first bit. Aren't these parameters already strings? In case I misinterpreted, I also updated this to include a non-string in tandem to show multiple parameters, as well.

Issue #1497, opened 2020-07-22
Set up code coverage for Robolectric tests


vinitamurthi:
We would like to get code coverage reports for our robolectric tests. The code coverage tool needs to be compatible with Kotlin and java, as well as work on Gradle and Bazel.
Jacoco seems to be a good pathway, some useful links:

- Jacoco for Kotlin: https://kevcodez.de/posts/2018-08-19-test-coverage-in-kotlin-with-jacoco/
- Jacoco in Bazel: https://github.com/bazelbuild/bazel/tree/master/third_party/java/jacoco

vinitamurthi:
CC @BenHenning @anandwana001 , let me know your thoughts on using Jacoco! Here's the link: https://www.jacoco.org/jacoco/trunk/index.html

BenHenning:
FYI I found out that Bazel's android local tests do not support code coverage at all, and this is something we'll need to contribute directly.

MaskedCarrot:
May I work on this?

peculiaruc:
Thanks for showing interest to work on this task but it already assigned.

BenHenning:
This is resolved with #5343.

PR #4052, opened 2021-12-15
Fix part of #4044: Add math expression/equation parser (with error detection)

BenHenning:
## Explanation
Fix part of #4044

This PR introduces an LL(1) recursive descent parser for numeric expressions, algebraic expressions, and algebraic equations (per [this grammar specification](https://docs.google.com/document/d/1JMpbjqRqdEpye67HvDoqBo_rtScY9oEaB7SwKBBspss/edit#bookmark=id.wtmim9gp20a6)).

Some details about the implementation:
- Error handling is managed through a MathParsingResult sealed class with corresponding transformations and errors specifically tied to MathParsingError. I actually quite like this pattern since it avoids relying on exceptions yet still provides as much context as the frontend needs (this will be more apparent in a future PR that leverages these errors for user-visible errors). I might try to adopt this as the more broad error handling pattern in the future.
- Error detection is split into two levels of strictness: required-only (i.e. irrecoverable) and all errors. 'Optional' errors are ones that can still be represented by the ``MathExpression`` proto, but are likely learner errors. While classifiers will never disable these, tests do in order to test certain scenarios that are possible per the grammar.
- The parser implementation is intentionally generic so that both numeric & algebraic expressions can share the same parser (for simplicity)
- A specific effort is made to force LL(1) parsing for simpler grammar & implementation and long-term maintainability (since it's likely this grammar will need to be changed in the future). This is done by leveraging an LL(1) tokenizer and wrapping the ``Sequence`` provided by the tokenizer in a ``PeekableIterator`` (which doesn't allow a lookahead farther than 1). As a result, the parser will short-circuit and fail at the first encounter of an irrecoverable error (optional errors are processed after parsing is complete).
- The original implementation leveraged the [Shunting-yard algorithm](https://en.wikipedia.org/wiki/Shunting-yard_algorithm) (see implementation: https://github.com/oppia/oppia-android/blob/33d62aaa415bcab92dbecd9d64eda199f2676532/utility/src/main/java/org/oppia/android/util/math/MathExpressionParser.kt#L59). This implementation was far smaller, but it was much harder to make specific complex cases work correctly (such as implicit multiplication, and specialized error detection). A recursive descent parser was picked, instead, for better long-term maintainability and flexibility.
- An attempt was made to leverage a Kotlin DSL to define a generic grammar language to try and more declaratively define the implementation (i.e. by sort of just writing out the grammar and inferring a parser from it), but it ended up being more complex than the direct implementation and the syntax wasn't much better. See: 0ddcd2eda73c15993b9c55c3c1cd4ae7136c6f26.
- Specific effort was made toward maintainability in the implementation by utilizing exhaustive when statements so that new token types trigger many compiler errors (to ensure most cases that handle tokens can correctly manage the new token). This is intentional.

Some details about testing (including testing exemptions):
- Due to the size and complexity of the parser, there are actually four test suites for it: MathExpressionParserTest (verifies that parsing works at a high-level, and thoroughly tests error cases), NumericExpressionParserTest (verifies thoroughly that numeric expressions are parsed correctly), AlgebraicExpressionParserTest (thoroughly verifies variable-based parsing), and AlgebraicEquationTest (verifies equation parsing). This approach helps reduce the size of the overall suites and makes them easier to work with (Android Studio seems to have issues with Kotlin files much larger than 1k LOC).
- Note that the tests are intentionally non-redundant; they leak the implementation detail that a common implementation is used for all three types of math constructs and rely on each other for verification (i.e. AlgebraicExpressionTest assumes that most of the tests in NumericExpressionTest also apply to it, just with a variable added).
- Note that error testing does not include ordering. While it does matter to an extent, it didn't seem important enough to add targeted tests for it right now.
- I used code coverage checking locally (as part of verifying the possibility of #1497) to add a few more tests for the parser, but all the remaining uncovered lines are either impossible to cover (due to cascading when statements where earlier calls already cover such cases; this is a limitation in Kotlin) or are incorrectly instrumented by JaCoCo. The current code coverage for the parser is around 93.5% (but realistically is closer to 100%; though there are many logical cases not covered in current tests).
- MathParsingError was exempted from tests since it's a simple data structure definition.
- A custom Truth subject was added for MathParsingError.
- MathParsingErrorSubject was exempted to keep this PR smaller, and because it's trivially verifiable. #4132 was filed to track adding tests in the future.

A regex exemption check was added to enable parameterized testing in MathExpressionParserTest. These are tests which are verifying the same failure behaviors under many different conditions, so it seems like a good utilization of parameterization.

## Essential Checklist

## For UI-specific PRs only
The parser introduced in this PR is not yet available via any UI features, so it doesn't directly affect UIs yet.


BenHenning:

BenHenning:
@anandwana001 PTAL for codeowners & as part of the broader math expressions project.

BenHenning:
FYI that I pulled in upstream changes including a new sync with develop. I've also added a new test. Hopefully this doesn't affect the review too much.

BenHenning:
Note that the recent changes:
1. Add support for the en-dash for subtraction (& related tests)
2. Updates MathExpressionParserTest to use the faster JUnit-only test runner
3. Fixes unary precedence so that it's higher than multiplication and division (per the changes described [here](https://docs.google.com/document/d/1JMpbjqRqdEpye67HvDoqBo_rtScY9oEaB7SwKBBspss/edit#bookmark=id.iynn3amkkdxo))

anandwana001 (approved):
Overall LGTM, 

Found `MathExpressionSubject.kt` and `MathParsingErrorSubject.kt` very interesting.

BenHenning:
This is still active.

PR #4054, opened 2021-12-15
Fix part of #4044: Add support for expression evaluation & conversion to LaTeX

BenHenning:
## Explanation
Fix part of #4044

This PR introduces support for:
- Converting math expressions & equations to renderable LaTeX
- Evaluating numeric expressions (i.e. turning them into a ``Real`` value)

The LaTeX support is fairly barebones, but it covers all cases that can be represented by the math expression and equation protos. Note that it has a special mode for treating divisions as fractions as this is a requirement for the upcoming math interactions that are being added (as a creator-specified configuration argument).

Some specifics regarding evaluation:
- The expression evaluation is limited to numeric expressions since variables can't be evaluated without values being provided for those variables. This is something we could trivially add in the future, but we don't yet have any applications for such functionality.
- Evaluation is meant to be used mainly in tests for verifying structural correctness of parsed expressions (see the changes in NumericExpressionParserTest). All of the complexity added for computing parts of expressions won't just be limited to evaluation, though; it's also needed for polynomial conversion (which is being added in an upcoming PR) and for downstream classifiers work (one of the classifier types compares the results of evaluation for numeric expressions).
- The constituent computation pieces of evaluation are designed to maximize precision by trying keep integers for as long as possible. For example, adding two integers yields an integer, but dividing two might not (e.g. if the integers correctly divide to another integer, then an integer will be returned; otherwise, a fraction will be returned). When precision cannot accurately be kept, then a double is computed. This approach has nice benefits for polynomial computations since it ensures perfect operations when possible (there are less concerns around carrying rounding errors).
- 0^0 is special cased to be 1 for consistency with the system pow() method and other languages (see https://stackoverflow.com/a/19955996).
- Failures are a bit inconsistent (they can sometimes throw and sometimes return NaN/infinity). We can revise this in the future if needed, but I wanted to err on the side of "failures probably won't happen" since we have parser-based guards in place to prevent them.
- The pow function supports negative powers, non-integer powers, and root powers (e.g. powers whose absolute value is less than 1) where the root computation is the real value root rather than the principal root (i.e. it will take roots of negative values when it can). This actually is inconsistent with the system power function, but matches *some* calculator's behaviors. We can make this configurable or remove the negative support in the future, if needed.

Some notes on testing:
- ``MathEquationSubject`` and ``MathExpressionSubject`` were updated to include convenience functions for evaluation & LaTeX. These aren't directly tested since they're expected to be tested as part of the future test suites for each of these subjects.
- ``RealSubject.assertThat`` was updated to accept a null value (since this can come up with evaluation test cases).
- The fractions & real values tests are extremely detailed to cover a large variety of cases for each operation. This is exacerbated by the fact that each ``Real`` operation has a matrix of different combinations with different outcomes. This test coverage is important since these are the fundamental operational building blocks of polynomial calculations and evaluation.
- RealExtensionsTest is being exempted for parameterized testing. This is a rare case where these tests aren't actually verifying single behaviors, but it's necessary since the test matrix is too large to reasonably fit within a single test suite without maintenance concerns. Fortunately, it seems like the parameterized representation for these tests are mostly readable, and quite concise.

## Essential Checklist

## For UI-specific PRs only
N/A -- This change is introducing test-only functionality (numeric expression evaluation) and support for something that will later be shown on the UI (so it doesn't yet affect any UI screens or flows).


BenHenning:

BenHenning:
@anandwana001 PTAL at the whole PR as part of the math expressions project.

BenHenning:
Recent changes:
- The LaTeX converter test was updated to test the utility directly (vs. via MathExpressionExtensions)
- RealExtensionsTest was updated to use the faster JUnit-only test runner
- Evaluation checks were added for the new NumericExpressionParserTest tests

anandwana001 (approved):
Overall LGTM

BenHenning:
This is still active.

PR #4060, opened 2021-12-16
Fix #3846: Test for TextViewBindingAdapters

### LINKED ISSUE #3846: Add tests for TextViewBindingAdapters
TextViewBindingAdapters today doesn't have any tests for it, and ought to since it may not be obvious to track down future regressions that may occur in this file. These tests also ought to specifically test the binding adapters in their normal context (i.e. used through databinding with a test layout).

This also tracks adding new testing coverage after #3795 is merged since https://github.com/oppia/oppia-android/blob/baf7fd23affbf7dc27192a24fbb881e10519d2e7/utility/src/test/java/org/oppia/android/util/system/OppiaDateTimeFormatterTest.kt is being removed & its behaviors are now part of TextVIewBindingAdapters directly (so the same behaviors from that test suite should be added to the test suite for TextViewBindingAdapters).

rishidyno:
Fix #3846
## Essential Checklist


BenHenning:

rishidyno:
.

rishidyno:
@BenHenning @rt4914 All tests have passed PTAL. 

https://user-images.githubusercontent.com/78549195/149935135-6bb790a8-cbd8-49fd-8904-7b36f1d89708.mp4

app/src/main/AndroidManifest.xml
```diff
@@ -200,7 +200,10 @@
       android:name=".app.testing.TopicTestActivityForStory"
       android:theme="@style/OppiaThemeWithoutActionBar" />
     <activity android:name=".app.testing.TopicRevisionTestActivity" />
-    <activity android:name=".app.testing.ViewBindingAdaptersTestActivity"/>
+    <activity
+      android:name=".app.testing.TextViewBindingAdaptersTestActivity"
```
BenHenning: Is this in lexicographical order by android:name?

app/src/main/res/drawable/test_text_view_drawable_binding_adapter.xml
```diff
@@ -0,0 +1,15 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
BenHenning: Could you maybe use one of the drawables already included in the app rather than adding a new one?

app/src/main/res/layout/test_text_view_bindable_adapter_activity.xml
```diff
@@ -0,0 +1,13 @@
...
+
+    <TextView
+      android:id="@+id/test_text_view"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:fontFamily="sans-serif-medium"
+      android:padding="10dp"
+      android:text="@string/app_name"
+      android:textSize="18sp"
+      />
```
BenHenning: Nit: should be on the line above.

app/src/main/res/values/strings.xml
```diff
@@ -452,6 +452,8 @@
   <string name="ratio_default_hint_text" description="The default hint text for the ratio interaction if the placeholder text is not available.">
     Enter a ratio in the form x:y.
   </string>
+  <!--  TextViewBindingAdaptersTestActivity-->
```
BenHenning: Nit: please make sure there's exactly one space on either side of the activity name.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -0,0 +1,220 @@
...
+    @Component.Builder
+    interface Builder : ApplicationComponent.Builder
+
+    fun inject(textViewBindingAdaptersTest: TextViewBindingAdaptersTest)
+  }
+
+  /**
+   * Class to override a dependency throughout the test application, instead of overriding the
+   * dependencies in every test class, we can just do it once by extending the Application class.
+   */
```
BenHenning: Please also remove the unnecessary KDocs in this test suite.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -0,0 +1,220 @@
...
+import javax.inject.Singleton
+
+/** Tests for [TextViewBindingAdapters]. */
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(
+  application = TextViewBindingAdaptersTest.TestApplication::class,
+  qualifiers = "port-xxhdpi"
+)
+
```
BenHenning: Nit: remove blank newline for consistency.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -0,0 +1,220 @@
...
+import org.oppia.android.util.networking.NetworkConnectionDebugUtilModule
+import org.oppia.android.util.networking.NetworkConnectionUtilDebugModule
+import org.oppia.android.util.parser.html.HtmlParserEntityTypeModule
+import org.oppia.android.util.parser.image.ImageParsingModule
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Tests for [TextViewBindingAdapters]. */
```
BenHenning, line 83: Similar to a comment I left on an earlier PR, this doesn't actually point to a class or object (so it shouldn't use square brackets).

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  @Inject
+  lateinit var fakeOppiaClock: FakeOppiaClock
+
+  // Time: Wed Apr 24 2019 08:22:00
+  private val MORNING_TIMESTAMP = 1556094120000
+
+  @Test
+  fun testTextViewBindingAdapters_ltrIsEnabled_port_profileDataTextIsCorrect() {
```
BenHenning: These tests cover a couple of the happy path cases, but ought to cover all other cases that the production code includes.

One way to do this: you can analyze the production code to determine different cases of input that would hit each branch of the implementation. Then, consider what behaviors would result in those inputs, and then add tests for each behavior you come up with. In this case, you've only add one behavior for setProfileDataText (setting it to the same timestamp as the current time).

BenHenning (changes requested):
Thanks @rishidyno! Left some comments--PTAL.

app/src/main/AndroidManifest.xml
rishidyno: Yup!

app/src/main/res/drawable/test_text_view_drawable_binding_adapter.xml
rishidyno: Ok!

app/src/main/res/layout/test_text_view_bindable_adapter_activity.xml
rishidyno: Done

app/src/main/res/values/strings.xml
rishidyno: Done

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Well, for this It's done this way in other binding adapters test files. Otherwise, makes it in consistence.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Done

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: So I have left it as is. Or will have to change others?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Regarding this, I quite not understand where to get the production code. Earlier when we were in the meet I used resource handler and got the current time in ms with Oppia clock and just used that function which converts it to that format and compared. Will it work if I do that.

rishidyno:
@BenHenning PTAL.

BenHenning:
@rishidyno is this ready for review? I don't see any of my comments responded to. Please make sure to respond to all comments to confirm they've been addressed, or why they haven't in cases when they can't or you don't think they should.

rishidyno:
Reason for me not responding [here](https://github.com/oppia/oppia-android/pull/4077#issuecomment-1018264579) as mention in #4077 Pr.

rishidyno:
> @rishidyno is this ready for review? I don't see any of my comments responded to. Please make sure to respond to all comments to confirm they've been addressed, or why they haven't in cases when they can't or you don't think they should.

Reason for me not responding [here](https://github.com/oppia/oppia-android/pull/4077#issuecomment-1018264579) as mention in #4077 Pr.

rishidyno:
image
image
image

rishidyno (commented):
@BenHenning PTAl

rishidyno:
@BenHenning I think I did not submit it to get a review, that's why they were not showing. Thanks to @yash10019coder he helped me figure it out. PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: We should fix it since it's incorrect. The other test suites ought to be updated, too, but that's outside the scope of this PR.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: The production code for setProfileDataText is https://github.com/oppia/oppia-android/blob/b9d4b13af8c880f5fed8099194f26f9b3f52f0c3/app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapters.java#L26. The way to think about this is any if/when blocks and loops indicate multiple conditions that should have corresponding tests. Sometimes methods without control structures can also vary based on input (such as selecting a different string). We should test all the cases we can think of for these methods.

app/src/main/AndroidManifest.xml
BenHenning: Shouldn't it go before TopicTestActivity ('e' comes before 'o')?

BenHenning (commented):
Thanks @rishidyno! Just a few comments left, including the one requesting to add more tests. PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rt4914: Basically `setProfileDataText ` is only responsible for showing "Created on ....." text in textview, which means we are just sending the timestamp as long and it is converting it as string. Now whatever timestamp we send it will convert it to string, that timestamp value can be `0` or `1556094120000` and it will just covert it to either 

`Created on 1 Jan 1970`, or `Created on April 24, 2019`. So I am not sure what other test we can add. 

Are you thinking of other locales like arabic?

rt4914:
Will need to rely on @BenHenning for review on this one.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: I think testing setProfileDataText for two different inputs is probably sufficient (to demonstrate that the variable part of the string is working as expected). We don't need to test multiple locales here since there's nothing locale-specific.

setProfileLastVisitedText has a lot more behaviors to test, though, since the label that's computed has a bunch of different potential cases. I see that method as being the one that will have the most tests corresponding to it.

app/src/main/AndroidManifest.xml
rishidyno: Ya! My bad.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Added another case of `TIMESTAMP =0`.

rishidyno:
@BenHenning PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: So I think checking `setProfileLastVisitedText` for `just now`, `x minutes ago`, `x hours ago` and `x days ago` will be sufficient?

rishidyno: @BenHenning @rt4914 I have added other test cases for possible outputs. Do I need to structure them?

BenHenning: Thanks @rishidyno! I think this is on the right track.

In terms of testing coverage, I think you did a really good job covering all of the "this should work as expected" cases. The three that I think are also worth covering are when the provided timestamp is in the future (i.e. bigger than the current time), when it's zero, and when it's negative. These are all cases handled by the implementation, so we should verify that they work as expected (as per the adapter's contract).

Beyond that, you should structure each behavior being checked in its own test rather than combining them all into one.

BenHenning (commented):
Thanks @rishidyno! Apologies for the delayed review. I think the PR is quite close, but I had some suggestions to improve upon the tests you recently added--PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: @BenHenning I have separated all test cases and added few more. PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -140,37 +140,91 @@ class TextViewBindingAdaptersTest {
...
       setProfileLastVisitedText(
         textView,
         /* setText= */ getCurrentTimeMs - 86400000
       )
       assertThat(textView.text.toString()).isEqualTo("Last used yesterday")
+    }
+  }
 
+  @Test
+  fun testTextViewBindingAdapters_ltrIsEnabled_port_profileLastVisitedTextIsCorrectCase7() {
```
BenHenning: Rather than case1, case2, etc. you should name this based on the conditions and outcome, per https://github.com/oppia/oppia-android/wiki/Oppia-Android-Testing#naming-convention.

For this specific method, it could be called:
``testSetProfileLastVisitedText_forTimeTwoDaysAgo_setsLastUsedTwoDaysAgoText``

Ditto elsewhere.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -179,6 +233,51 @@ class TextViewBindingAdaptersTest {
...
+    fakeOppiaClock.setFakeTimeMode(FakeOppiaClock.FakeTimeMode.MODE_FIXED_FAKE_TIME)
+    fakeOppiaClock.setCurrentTimeMs(TIMESTAMP)
+    val getCurrentTimeMs = fakeOppiaClock.getCurrentTimeMs()
+    activityRule.scenario.onActivity {
+      val textView: TextView = it.findViewById(R.id.test_text_view)
+      setProfileLastVisitedText(
+        textView,
+        /* setText= */ getCurrentTimeMs + 172800000
+      )
+      assertThat(textView.text.toString()).isEqualTo("Last used ")
```
BenHenning: Perhaps we could add a new string that just says "Last used recently" that we update the adapter to use in these cases, instead. It's nice that we have tests since they demonstrate that the resulting behavior isn't ideal. I know this is a bit outside the scope of the PR, but it seems like a nice thing to fix while someone's looking at this code, anyway. WDYT?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: Left some follow-up comments, so resolving this now.

BenHenning (commented):
Thanks @rishidyno! Just had a few follow-up comments--PTAL.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: I think doing this would be easy, as Instead of returning an empty string in these cases, we can just return `recently`.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -0,0 +1,350 @@
...
+      setProfileLastVisitedText(
+        textView,
+        /* setText= */ getCurrentTimeMs + 172800000
+      )
+      assertThat(textView.text.toString()).isEqualTo("Last used recently")
+    }
+  }
+
+  @Test
+  fun testSetProfileLastVisitedText_forTimeZero_setsLastUsedRecently() {
```
rishidyno, line 252: @BenHenning Is this fun name suitable?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Done.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: I have already made few changes PTAL.

rishidyno: I used google translator for `'recently'` for other languages.

BenHenning:

app/src/main/res/values-ar/strings.xml
```diff
@@ -365,6 +365,7 @@
   <string name="just_now">الآن</string>
   <string name="time_ago">منذ %s</string>
   <string name="yesterday">أمس</string>
+  <string name="recently">مؤخرا</string>
```
BenHenning: Prefer not to add this or the Brazilian Portuguese variant--we'll have translators translate these after the English base string is checked in.

app/src/main/res/values/strings.xml
```diff
@@ -407,6 +407,7 @@
   <string name="reveal">Reveal</string>
   <!-- TextViewBindingAdapters -->
   <string name="just_now">just now</string>
+  <string name="recently">recently</string>
```
BenHenning: The name ought to be clearer (the other names aren't very clear). Perhaps: ``last_logged_in_recently`` would be clearer?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
```diff
@@ -189,7 +189,7 @@ class TextViewBindingAdaptersTest {
   }
 
   @Test
-  fun testTextViewBindingAdapters_ltrIsEnabled_port_profileLastVisitedTextIsCorrectCase5() {
+  fun testSetProfileLastVisitedText_forTimeTwoHourAgo_setsLastUsedTwoHoursAgoText() {
```
BenHenning: Nit: ``TwoHoursAgo``. Ditto elsewhere (I think there are some pluralization issues in other tests).

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: Yup, this is clear--thanks for checking!

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
BenHenning: Followed up with additional comments--thanks!

BenHenning (commented):
Thanks @rishidyno! I think the latest changes look really good! Just had a few nits--PTAL.

app/src/main/res/values-ar/strings.xml
rishidyno: Removed.

app/src/main/res/values/strings.xml
rishidyno: Changed.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextViewBindingAdaptersTest.kt
rishidyno: Done.

rishidyno:
@BenHenning did few nit-fix PTAL.

BenHenning (approved):
Thanks @rishidyno. This LGTM!

BenHenning:
@anandwana001 and @rt4914 I think you need to take a pass for codeowners.

anandwana001 (approved):
My code owner file LGTM

rt4914 (approved):
Thanks a lot @rishidyno 
LGTM

PR #3194, opened 2021-05-17
Fix #3201: Setup Kotlitex library to render raw latex

anandwana001:
## Explanation

Fix #3201 

Need data with the math tag to review this PR. 

[Documentation](https://docs.google.com/document/d/1xEgWGYIUgqozTv8NdFwpmTJW0u5JHo1ezRgsoqvX0Q4/edit?usp=sharing)

Sample Result
```
"html": "<p>This represents <oppia-noninteractive-math math_content-with-value=\"{&amp;quot;raw_latex&amp;quot;: &amp;quot;\\\\frac{2}{5}&amp;quot;, &amp;quot;svg_filename&amp;quot;: &amp;quot;mathImg_20200901_220408_fllx4jv5pn_height_3d329_width_1d666_vertical_1d072.svg&amp;quot;}\"></oppia-noninteractive-math>.</p><p>Then, see if you can figure out what to do next, so that you get a fraction that looks like <oppia-noninteractive-math math_content-with-value=\"{&amp;quot;raw_latex&amp;quot;: &amp;quot;\\\\frac{?}{10}&amp;quot;, &amp;quot;svg_filename&amp;quot;: &amp;quot;mathImg_20200901_220408_wpjlg4ybsl_height_3d329_width_2d495_vertical_1d072.svg&amp;quot;}\"></oppia-noninteractive-math>.</p><p>Test exploration with interactions. <oppia-noninteractive-link text-with-value=\"&quot;Oppia&quot;\" url-with-value=\"&quot;https://oppia.github.io&quot;\"></oppia-noninteractive-link></p>"
```

Alpha Content

### TODO/Q&A
1. Should we remove SVG rendering? - keep both, use raw latex when no svg available
2. Update `MathTagHandlerTest`. 

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
[...]

Issue #3715, opened 2021-08-22
Introduce managed executors to back dispatchers instead of using managed dispatchers


BenHenning:
A prerequisite for #2951 is building a set of executors which have corresponding managed testing utilities rather than only relying on CoroutineDispatchers. This will largely simplify things since it allows us to properly synchronize both Glide & Retrofit, and also removes the need for the complex CoroutineExecutorService. Further, it provides a fairly easy interop story between coroutines & Guava's ListenableFutures (e.g. with SettableFuture).

PR #4068, opened 2021-12-17
Fix part of #4044: Add KotliTeX integration (direct LaTeX rendering)

BenHenning:
## Explanation
Fix part of #4044

This PR introduces support for rendering raw LaTeX in the app using the KotliTeX (https://github.com/karino2/kotlitex) library which is necessary since users may input arbitrary math expressions that need to be pretty-rendered (using the math expression -> LaTeX generation support added in #4054).

This implementation is heavily based on #3194 and leverages a new fork (https://github.com/oppia/kotlitex) of KotliTeX based on @anandwana001's fork.

In order to keep the implementation simpler, this hangs the functionality onto the existing custom Oppia noninteractive math tag except it only renders LaTeX if there's no SVG filename to load (which will be the case for dynamically generated LaTeX). This approach has the advantage of the app always being able to fall back to LaTeX rendering in the unlikely scenario where content has a math tag without an SVG URL being present (or where the SVG fails to load which could allow potential fault tolerance when downloads support is fully implemented). These changes required several changes to KotliTeX itself which are outlined in more detail below.

Furthermore, it was observed that the rendering was noticeably slow on a Nexus 5X for repeated renders (think navigation back and forth between views with rendered LaTeX, or expanding/collapsing multiple answers). This didn't seem acceptable, so this PR also introduces a platform parameter for utilizing Glide to pre-render the LaTeX into a PNG bitmap so that it can cache it. This is not only more performant for repeated exposure and faster to render, but it also offloads the expensive rendering step to a background thread rather than blocking the main thread (see the videos in the UI section below for a before-and-after comparison).

### Details on changes to KotliTeX
A custom fork of KotliTeX (https://github.com/oppia/kotlitex) was needed to include the following changes:
- Support to be built on Bazel
- Lower the min sdk required by the library to match the app's
- A slight hack to ensure KotliTeX doesn't auto-fail for larger square roots (the library actually doesn't support them, so this workaround just stretches the smaller square root rather than leading to error text)
- A hack to expose the outer bounds of drawing requires by KotliTeX for its span (see caching details below for why)

Note that KotliTeX only supports a subset of LaTeX, but it seems to support everything the app needs for now (and it should be extensible by porting other portions of KaTeX as needed).

### Details on the rendering caching & positioning challenges
At a high-level, the approach to caching is to render the KotliTeX drawable to a bitmap that can be saved on-disk in a format that Glide can easily load on its own (PNG). While it's not ideal to perform the extra PNG conversion step, it's much more disk-friendly (and in the future, we could look into offloading just this part to another thread if needed). Each individual load is keyed on the exact raw LaTeX, the line height (up to 2 digits), and whether it's inline or block rendered (see later in this section for the distinction). If all three of these properties match, there should be a Glide cache hit and no need to re-render the LaTeX.

A couple of issues arose when trying to implement this:
- Despite us being able to expose the internal KotliTeX drawable (which wasn't by default), I was having a lot of difficulty getting the text alignment correct. Instead, it was much easier to rely on Android's internal text rendering by just rendering the span in a similar way to what TextView does.
- While the above worked, it led to another problem: figuring out how large the bitmap should be. The actual drawable bounds that KotliTeX computes is an underestimate (its ascent and descent go outside of these bounds), so an additional estimation is needed.
- Because we're now drawing the LaTeX as a bitmap in an ImageSpan, the text aligns the image either with its bottom line or its baseline. This introduces a vertical center alignment issue as shown below (the image with the red text shows one attempt at fixing this by eliminating unnecessary upper space by using a combination of vertical translation and negative bounds). Due to this quirk, we can't actually have whitespace below the text without modifying the text's font metrics (specifically the bottom/top/ascent/descent properties). #4170 is tracking this work. Note also that this is really only an issue for inline rendering which we're unlikely to use in the medium-term since all user-submitted answers will be rendered in block style which seems to work well.

Caching is demonstrably much more performant for large amounts of repetition, and it doesn't result in UI lag like direct rendering does since all rendering is offloaded to a background thread (see the videos below for the comparison).

### Details on testing
The tests are mainly focused around changes to the tag handler since the Glide portions can't be tested, and neither can the rendering routines (hence the exemptions). The fake image loader was updated to include support for the math-based rendering load requests.

An aside change was made to make the Robolectric library test-only (it's something that was noticed in passing during development, and isn't directly needed by this PR; it's just a nice-to-have).

The new interfaces are exempted from tests since they have no logic to verify (their implementations are generated by Dagger and verified at compile time). Furthermore, I didn't add any tests for the changes to the compute affected tests script since it's a bit difficult to test, and already quite an edge case. Please let me know if you have any concerns with this.

### Details on the race condition fix
KotliTeX seems to initialized shared state when its parsing the LaTeX, and this with multiple Glide threads can cause a race condition where the LaTeX fails to render. This has been fixed by leveraging coroutines (where we still try to maximize efficiency and parallelization), but it required exposing new injector pathways for the coroutine dispatchers and ``ConsoleLogger``.

### Details on the ComputeAffectedTests fix
This PR exposes an issue with the ComputeAffectedTests script wherein it can sometimes exceed the system's maximum argument length.  This is being worked around by partitioning values for longer Bazel queries into multiple calls and concatenating the results. I'm fairly certain this is functionally equivalent to the current implementation, and for the vast majority of future PRs only one partition will ever be used. There doesn't seem to be an alternative possible since this is a platform-specific Kernel restriction.

## Essential Checklist

## For UI-specific PRs only
For the most part, this PR doesn't actually affect users yet since this functionality isn't exposed (all LaTeX currently rendered in the app will utilize the existing SVG pipeline). The user-facing UI changes will be thoroughly documented in the PR that's planned for after this one since it'll be integrating all prior PRs along with the changes introduced here.

That being said, the "What is a Fraction?" test revision card has been updated to include rendered LaTeX text in order to demonstrate the functionality more closely. The following is how rendering looks (with bitmap caching):

cached_latex_rendering

Note that the above, when compared with non-cached rendering, demonstrates where some of the spacing issues mentioned above come in:

direct_render_latex

Here's the image showing the bounds when trying to force the drawable downward:

This video shows how long a bunch of LaTeX expressions takes to render with direct rendering:

https://user-images.githubusercontent.com/12983742/153363983-4ff99db5-52da-4360-9630-cc98397dc07b.mp4

Versus the performance of caching for the same experience (note this occurred with a completely clean local app cache):

https://user-images.githubusercontent.com/12983742/153364178-3066d6ce-2a85-45c8-83fe-c0378a786023.mp4

### Notes on rendering approach

Note that rendering during cache time involves a three step process:
1. Approximate the outer bounding volume by tracking all draw calls needed for a given LaTeX expression (these bounds are used to create the bitmap that will store the cacheable render)
2. Render the LaTeX to a canvas with dimensions larger than the approximated bounding (since it's possible for the rendering to exceed these estimates--see notes below)
3. Crop the bitmap to the smallest bounds needed to encapsulate all of the pixels

(1) is needed because KotliTeX actually incorrect computes bounds (it sometimes over and underestimates). This isn't as apparent when directly rendering because Android will handle text expanding past the space KotliTeX expects since it's rendered in-line. I spent some time trying to debug this within KotliTeX but I couldn't make much headway. It may be worth looking into this in the future to see if we can reliably calculate the bounds within KotliTeX (since it will likely be much more performant than the solution used here). Seeing KotliTeX's debug lines makes this clearer (where it thinks the bounds are relative to glyphs; note that this screenshot is using rendering without caching):

kotlitex_wrong_bounds

(2) is needed because the methodology used for (1) is inexact. It's actually really difficult to estimate exactly where the final glyph pixels will be rendered for a given glyph since KotliTeX is directly handling positioning (rather than allowing Android to compute x/y coordinates based on nearby characters to account for tracking, kerning, etc.). The dimension approximation for (1) is much closer to correct than KotliTeX's internal bounds calculations, however it tends to slightly underapproximate the bounds. To counteract this, the app uses a bitmap that's 2x what's actually estimated to be needed and offsets rendering into that bitmap such that 50% of the estimated size is a margin around the drawing area. This provides a large buffer to account for incorrect bounds calculations. The app then crops this to the edges of filled pixels so that the smallest bitmap is used to represent the final rendered equation. While this uses a lot more memory than should actually be required for rendering, it guarantees perfect results. See:

Before (notice the over-sized LaTeX, and that some expressions are partially cut-off from the right and above, or have extra space below):

| latex_rendering_cutoff_problem1 | latex_rendering_cutoff_problem2 |
|------|------|

After (with size estimation & croppingl this also includes a fix in UrlImageParser to ensure that LaTeX images are never auto-resized when rendering in block mode since it distorts them, and that they are properly centered):

| latex_rendering_cutoff_fix1 | latex_rendering_cutoff_fix2 |
|------|------|


BenHenning:
Things left to do on this PR:
- [x] Fix Gradle builds
- [x] Possibly fix the compute affected tests script since it seems to be breaking
- [x] ~Fix the licenses (need to add some new ones)~ (no longer necessary since KotliTeX needed patching--it was inadvertently pulling in unnecessary test deps)
- [x] Fix the race condition in KotliTeX caused by Glide

BenHenning:
Note that there's still an image cutoff for single-line LaTeX, so a bit more iteration is needed here around positioning. That being said, if the latest changes pass CI I think this can go into review.

BenHenning:
@rt4914 @vinitamurthi PTAL for codeowners.

@anandwana001 PTAL at the whole PR as part of the broader math expression work.

rt4914 (approved):
LGTM for code-owners. Also the performance has changed drastically. Great work.

anandwana001 (approved):
LGTM
The custom ModalLoader and caching were amazing.

**Feel free to reply later on for the below query as we are swamped with multiple PR's currently.**
I have a quick query about class `LatexModelDataFetcher`. 
```kt
CoroutineScope(backgroundDispatcher).launch {
        withContext(CoroutineScope(blockingDispatcher).coroutineContext) {}
}
```

Here we launch the coroutine in `backgroundDispatcher` and inside it, we are launching `withContext` on `blockingDispatcher`. 

As per my understanding, `backgroundDispatcher` is going to provide us a thread pool of 4 threads to complete our work and `blockingDispatcher` provides a single thread. 

The question here is:
1. Why do we need `withContext` as we are already passing `backgroundDispatcher` to the coroutine scope which has 4 threads?
2. Why `blockingDispatcher` is passed to `withContext`? Is this somehow blocks our execution on Main thread?

BenHenning:
> LGTM The custom ModalLoader and caching were amazing.
> 
> **Feel free to reply later on for the below query as we are swamped with multiple PR's currently.** I have a quick query about class `LatexModelDataFetcher`.
> 
> ```kotlin
> CoroutineScope(backgroundDispatcher).launch {
>         withContext(CoroutineScope(blockingDispatcher).coroutineContext) {}
> }
> ```
> 
> Here we launch the coroutine in `backgroundDispatcher` and inside it, we are launching `withContext` on `blockingDispatcher`.
> 
> As per my understanding, `backgroundDispatcher` is going to provide us a thread pool of 4 threads to complete our work and `blockingDispatcher` provides a single thread.
> 
> The question here is:
> 
> 1. Why do we need `withContext` as we are already passing `backgroundDispatcher` to the coroutine scope which has 4 threads?
> 2. Why `blockingDispatcher` is passed to `withContext`? Is this somehow blocks our execution on Main thread?

``withContext`` starts a new coroutine asynchronously for the provided scope (just creating a scope isn't sufficient to do anything with a given coroutine context or dispatcher; it needs to be used to execute something).

The blocking dispatcher is a single background thread that ensures all operations executed on it happen on the same thread. This is useful to avoid race conditions without needing to rely on the main thread. Since a specific part of interacting with KotliTeX has a race condition, we use the blocking dispatcher to complete just this part then defer the rest of the operation (which is rather expensive) to the normal background dispatcher to ensure it can be parallelized.

Note that the main thread isn't being used at all here (the calling thread from Glide is actually also a background thread, just one managed by Glide; eventually we'll be providing these threads so that we can properly synchronize both Glide & app threads in tests--see #3715 if you're curious about this).

BenHenning:
Note that the latest changes are fairly extensive in how LaTeX rendering is happening for caching. There were a bunch of problems before that needed to get fix--see the updated PR description for details.

@anandwana001 could you PTAL at these changes and approve if they look good to you, or, leave comments if not? Also, I replied to your questions above.

anandwana001 (approved):
LGTM

vinitamurthi (approved):
LGTM as codeowner

PR #4080, opened 2022-01-05
Fix #4006: Hide edit accounts option from administrator controls

### LINKED ISSUE #4006: Hide Edit Accounts option from AdministratorControls
Hide `Edit Accounts` option from `AdministratorControls` as currently it is of no use. 

Note: Do not remove it, just hide it. Add appropriate test case too.

JishnuGoyal:
## Explanation

Removes general section in `Administrator Controls` by removing AdministratorControlsGeneralViewModel from the itemViewModelList. 

Instead of just commenting out `AdministratorControlsGeneralViewModel`, a  new platform parameter, `ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE` was introduced and used to gate it. (which is the preferred way as suggested by @BenHenning) . 
Now, the Edit Accounts options can be enabled simply by changing the `ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE` to `true`.

Why this method has been used can be found in detail under the comments here: https://github.com/oppia/oppia-android/issues/4006#issuecomment-981965721

The screen looks like this now:
image

Device used: Emulator Pixel 4 API 28

## Essential Checklist

## For UI-specific PRs only


JishnuGoyal:
PTAL @rt4914 @BenHenning 
Thanks!

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -20,7 +21,8 @@ import org.oppia.android.domain.oppialogger.OppiaLogger
 import org.oppia.android.domain.profile.ProfileManagementController
 import org.oppia.android.util.data.AsyncResult
 import org.oppia.android.util.data.DataProviders.Companion.toLiveData
-import javax.inject.Inject
+import org.oppia.android.util.platformparameter.EnableEditAccountsOptionsUi
+import org.oppia.android.util.platformparameter.PlatformParameterValue
```
rt4914, line 23: You will need to reorganise these to pass the lint checks.
`import javax.inject.Inject` will come at the bottom.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -64,15 +67,20 @@ class AdministratorControlsViewModel @Inject constructor(
...
-    itemViewModelList.add(
       AdministratorControlsProfileViewModel(
         routeToProfileListListener,
         loadProfileListListener
       )
     )
+
+    if (enableEditAccountsOptionsUi.value) {
+      itemViewModelList.add(
+        AdministratorControlsGeneralViewModel()
```
rt4914, line 74: If the `enableEditAccountsOptionsUi` value is true won't the `Edit Accounts` option be visible below `Profiles` because of this current implementation.

Can you please turn the value `true` and share screenshot of how it looks.

rt4914 (changes requested):
@JishnuGoyal PTAL at the comments.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
JishnuGoyal: I used a temporary fix (which is now resolved) - For some reason Android Studio cannot import List properly(I think that is an unresolved problem from Android Studio's IDE side). Because of which I could not use ` val itemViewModelList: MutableList<AdministratorControlsItemViewModel> = emptyList` in order to initialise this variable.

> Can you please turn the value true and share screenshot of how it looks.

Now, after turning the value true, it looks like this: 
image

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
JishnuGoyal: PTAL @rt4914 Thanks!

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -64,15 +67,26 @@ class AdministratorControlsViewModel @Inject constructor(
...
+    if (enableEditAccountsOptionsUi.value) {
+      itemViewModelList = mutableListOf(
+        AdministratorControlsGeneralViewModel(),
+        AdministratorControlsProfileViewModel(
+          routeToProfileListListener,
+          loadProfileListListener
+        )
       )
-    )
+    }else{
```
BenHenning: There should be spaces around the 'else'. See https://developer.android.com/kotlin/style-guide#whitespace.

Further, it seems that your lint tests are failing. I suggest setting these up as a pre-push hook to ensure that lint issues are addressed before code review (this helps make things run more smoothly). See: https://github.com/oppia/oppia-android/wiki/Contributing-to-Oppia-android#install-oppia-android (you need to run the setup script in your local repository).

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -64,15 +67,26 @@ class AdministratorControlsViewModel @Inject constructor(
...
+    val itemViewModelList: MutableList<AdministratorControlsItemViewModel>
+
+    if (enableEditAccountsOptionsUi.value) {
+      itemViewModelList = mutableListOf(
+        AdministratorControlsGeneralViewModel(),
+        AdministratorControlsProfileViewModel(
+          routeToProfileListListener,
+          loadProfileListListener
+        )
       )
```
BenHenning:
```suggestion
    val itemViewModelList = mutableListOf<AdministratorControlsItemViewModel>()

    if (enableEditAccountsOptionsUi.value) {
      itemViewModelList += AdministratorControlsGeneralViewModel()
```

More Kotlin-idiomatic.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -64,15 +67,26 @@ class AdministratorControlsViewModel @Inject constructor(
...
-    )
+    }else{
+      itemViewModelList = mutableListOf(
+        AdministratorControlsProfileViewModel(
+          routeToProfileListListener,
+          loadProfileListListener
+        )
```
BenHenning: This is redundant with the above--can you refactor this to avoid the duplication?

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -175,7 +175,7 @@ class AdministratorControlsActivityTest {
   }
 
   @Test
-  fun testAdministratorControlsFragment_generalAndProfileManagementIsDisplayed() {
+  fun testAdministratorControlsFragment_profileManagementIsDisplayed() {
```
BenHenning: You should add another version of this test to verify that the UI shows when the feature is enabled. This verifies that the platform parameter is working as expected.

BenHenning (changes requested):
Thanks @JishnuGoyal. Sorry for the delayed review. The PR looks to be on the right track, but I think there are some changes that can be made both to simplify it a bit & ensure that it works as expected.

Further, please also make sure that tests & lint checks are passing locally before sending this back to reviewers.

rt4914:
@JishnuGoyal Please assign it once its ready for review.

JishnuGoyal:
Hello, 
In order to write test as mentioned here https://github.com/oppia/oppia-android/pull/4080#discussion_r780641386 ,
I created a TestModule in `AdministratorControlsActivityTest.kt` in order to get a copy of `EnableEditAccountsOptionsUI`. 

But I am facing an error when i try to run tests on `AdministratorActivityTest.kt`: 

> ^D:\college\opensource\oppia-android\app\build\tmp\kapt3\stubs\debugAndroidTest\org\oppia\android\app\options\OptionsFragmentTest.java:286: error: [Dagger/MissingBinding] @org.oppia.android.util.platformparameter.EnableEditAccountsOptionsUi org.oppia.android.util.platformparameter.PlatformParameterValue<java.lang.Boolean> cannot be provided without an @Provides-annotated method.

> Execution failed for task ':app:kaptDebugAndroidTestKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution
   > java.lang.reflect.InvocationTargetException (no error message)

I have never made any changes to OptionsFragmentTest.java file (as mentioned in the error above). 
The same copy of the file OptionsFragmentTest.java works fine on the develop branch (but when i try to run it on my local branch, it fails too.) 


What I have already tried:
1. Previously, I was trying to `@Inject @EnableEditAccountOptionsUi` without TestModule (as done in AdministratorControlsViewModel) inside `AdministratorControlsActivityTest`
2. Including `TestModule` in the `modules = [ ...]` list does not solve the problem

I feel that some files that are internaly generated by dagger got changed and are somehow not tracked by git, due to which I'm not able to track where the error is stemming from.

I have been trying to resolve this for a few days... Please give me a hint so I could proceed in the right direction @BenHenning  @rt4914  @FareesHussain .

BenHenning:
> Hello, In order to write test as mentioned here [#4080 (comment)](https://github.com/oppia/oppia-android/pull/4080#discussion_r780641386) , I created a TestModule in `AdministratorControlsActivityTest.kt` in order to get a copy of `EnableEditAccountsOptionsUI`.
> 
> But I am facing an error when i try to run tests on `AdministratorActivityTest.kt`:
> 
> > ^D:\college\opensource\oppia-android\app\build\tmp\kapt3\stubs\debugAndroidTest\org\oppia\android\app\options\OptionsFragmentTest.java:286: error: [Dagger/MissingBinding] @org.oppia.android.util.platformparameter.EnableEditAccountsOptionsUi org.oppia.android.util.platformparameter.PlatformParameterValue<java.lang.Boolean> cannot be provided without an @Provides-annotated method.
> 
> > Execution failed for task ':app:kaptDebugAndroidTestKotlin'.
> > A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution
> > java.lang.reflect.InvocationTargetException (no error message)
> 
> I have never made any changes to OptionsFragmentTest.java file (as mentioned in the error above). The same copy of the file OptionsFragmentTest.java works fine on the develop branch (but when i try to run it on my local branch, it fails too.)
> 
> What I have already tried:
> 
> 1. Previously, I was trying to `@Inject @EnableEditAccountOptionsUi` without TestModule (as done in AdministratorControlsViewModel) inside `AdministratorControlsActivityTest`
> 2. Including `TestModule` in the `modules = [ ...]` list does not solve the problem
> 
> I feel that some files that are internaly generated by dagger got changed and are somehow not tracked by git, due to which I'm not able to track where the error is stemming from.
> 
> I have been trying to resolve this for a few days... Please give me a hint so I could proceed in the right direction @BenHenning @rt4914 @FareesHussain .

You'll need a test module to provide the actual value, such as:

https://github.com/oppia/oppia-android/blob/0caaa7fc9d8dec302b28b94674d8881c233234a5/app/src/test/java/org/oppia/android/app/testing/options/OptionsFragmentTest.kt#L248

Some other fragments & tests may indirectly reference the fragment that you changed. This will, in turn, require the feature to provided in order to complete the dependency graph required by Dagger. For fragments that indirectly need this property but won't change their behavior due to it, you can just add a provider method in a test module that returns a default false to disable the feature.

JishnuGoyal:
1) I made the changes exactly as suggested by referring to `OptionsFragmentTest` to create a `TestModule` inside `AdministratorControlsActivityTest`.

2) Creating another `TestModule` (to provide `enableEditAccountsOptionsUiValue`) leads to an error saying that dagger has multiple bindings to PlatformParamaterModule. 

I inspected upon it and found that it is happening due to the TestApplicationComponent, ie, this block of code:

>   interface TestApplicationComponent : ApplicationComponent {
    @Component.Builder
    interface Builder : ApplicationComponent.Builder
    fun inject(administratorControlsActivityTest: AdministratorControlsActivityTest)
  }

3) The `ApplicationComponent` already includes `PlatformParameterModule` (check its definition).

4) I found this solution online: https://stackoverflow.com/questions/56761359/error-dagger-duplicatebindings-com-example-startrouter-is-bound-multiple-time 
According to this, removing PlatformParameterModule from the components list of `ApplicationComponent` will solve the error.

5) I tried this, but it lead to another missing dagger binding issue with the SyncUpWorkerPeriodHours value

May i suggest, shifting the Writing test part of this PR to a new separate issue, as otherwise this _will_  get a bit messy.
Since the main concern of this PR is solved (the main hide functionality and parameter gate is working fine), how about I push the code **before** I made changes to the _tests_ file and get those merged first? 

I'll write the tests on a fresh new issue. I feel this will make things simpler. 
Please tell what you think.

PTAL @BenHenning @rt4914  , Thanks

BenHenning:
> 1. I made the changes exactly as suggested by referring to `OptionsFragmentTest` to create a `TestModule` inside `AdministratorControlsActivityTest`.
> 2. Creating another `TestModule` (to provide `enableEditAccountsOptionsUiValue`) leads to an error saying that dagger has multiple bindings to PlatformParamaterModule.
> 
> I inspected upon it and found that it is happening due to the TestApplicationComponent, ie, this block of code:
> 
> > interface TestApplicationComponent : ApplicationComponent {
> > @Component.Builder
> > interface Builder : ApplicationComponent.Builder
> > fun inject(administratorControlsActivityTest: AdministratorControlsActivityTest)
> > }
> 
> 3. The `ApplicationComponent` already includes `PlatformParameterModule` (check its definition).
> 4. I found this solution online: https://stackoverflow.com/questions/56761359/error-dagger-duplicatebindings-com-example-startrouter-is-bound-multiple-time
>    According to this, removing PlatformParameterModule from the components list of `ApplicationComponent` will solve the error.
> 5. I tried this, but it lead to another missing dagger binding issue with the SyncUpWorkerPeriodHours value
> 
> May i suggest, shifting the Writing test part of this PR to a new separate issue, as otherwise this _will_ get a bit messy. Since the main concern of this PR is solved (the main hide functionality and parameter gate is working fine), how about I push the code **before** I made changes to the _tests_ file and get those merged first?
> 
> I'll write the tests on a fresh new issue. I feel this will make things simpler. Please tell what you think.
> 
> PTAL @BenHenning @rt4914 , Thanks

Your test isn't using ApplicationComponent, it's using TestApplicationComponent. Changing ApplicationComponent to fix a build error in a test isn't correct.

I think your duplicate binding issue is coming from the fact that your TestApplicationComponent is including PlatformParameterModule--you'll need to remove this in order to define custom overrides for features being tested in this test suite.

Regarding the tests, they need to be fixed before this PR is merged otherwise it will break the test builds for everyone else. Every submitted PR needs to keep the develop branch in a working state otherwise it will block other contributors.

JishnuGoyal:
Thanks for the tips and guidance, @BenHenning . I was able to compile all the tasks as suggested. I particularly learnt a lot from this PR. Thanks.

JishnuGoyal:
PTAL @BenHenning @rt4914 Thanks!

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -68,25 +68,22 @@ class AdministratorControlsViewModel @Inject constructor(
     deviceSettings: DeviceSettings
   ): List<AdministratorControlsItemViewModel> {
 
-    val itemViewModelList: MutableList<AdministratorControlsItemViewModel>
+    val itemViewModelList: MutableList<AdministratorControlsItemViewModel> =
+      emptyList<AdministratorControlsItemViewModel>().toMutableList()
```
BenHenning: You can instead do: ``mutableListOf<AdministratorControlsItemViewModel>()``

You also don't need to specify the val's type in this context since it can be inferred.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -175,7 +185,9 @@ class AdministratorControlsActivityTest {
   }
 
   @Test
-  fun testAdministratorControlsFragment_profileManagementIsDisplayed() {
+  fun testAdministratorControlsFragment_generalOptionsIsDisplayed() {
```
BenHenning: The test name ought to indicate the condition that leads to the expected outcome, e.g.:

``testAdministratorControlsFragment_editAccountFeatureEnabled_generalOptionsIsDisplayed``

Ditto for the next test.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -729,6 +799,42 @@ class AdministratorControlsActivityTest {
     fun inject(administratorControlsActivityTest: AdministratorControlsActivityTest)
   }
 
+  @Module
+  class TestModule {
+    companion object {
+      var forceEnableLanguageSelectionUi: Boolean = true
```
BenHenning: You can just return false in the provider below by default since there's no plans to change this.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -729,6 +799,42 @@ class AdministratorControlsActivityTest {
...
+      return PlatformParameterValue.createDefaultParameter(SPLASH_SCREEN_WELCOME_MSG_DEFAULT_VALUE)
+    }
+
+    @Provides
+    @SyncUpWorkerTimePeriodHours
+    fun provideSyncUpWorkerTimePeriod(): PlatformParameterValue<Int> {
+      return PlatformParameterValue.createDefaultParameter(
+        SYNC_UP_WORKER_TIME_PERIOD_IN_HOURS_DEFAULT_VALUE
+      )
+    }
```
BenHenning: Does the test not actually build without these?

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -729,6 +799,42 @@ class AdministratorControlsActivityTest {
     fun inject(administratorControlsActivityTest: AdministratorControlsActivityTest)
   }
 
+  @Module
+  class TestModule {
+    companion object {
+      var forceEnableLanguageSelectionUi: Boolean = true
+      var forceEnableEditAccountsOptionsUi: Boolean = true
```
BenHenning: Probably should be false (disabled) by default since that's the default production state.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -651,6 +653,14 @@ class OptionsFragmentTest {
     fun provideEnableLanguageSelectionUi(): PlatformParameterValue<Boolean> {
       return PlatformParameterValue.createDefaultParameter(forceEnableLanguageSelectionUi)
     }
+
+    @Provides
+    @EnableEditAccountsOptionsUi
+    fun provideEnableEditAccountsOptionsUi(): PlatformParameterValue<Boolean> {
+      return PlatformParameterValue.createDefaultParameter(
+        forceEnableEditAccountsOptionsUi
```
BenHenning: Ditto here--just default it.

BenHenning (changes requested):
Thanks @JishnuGoyal! Just had mostly nits as far as comments go, but I also noticed CI has a few failing checks. PTAL at the remaining comments & ensure any CI issues are fixed before sending this back.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
JishnuGoyal: SplashScreenWelcomeMsgParam is necessary for the test to compile, I removed syncUpWorkerTimePeriod :)

rt4914:
@JishnuGoyal Please do not mark any conversation which has been started by others as `Resolved`. Instead reply to those conversations so that the original person who started that conversation thread can check your reply and based on that `Resolve` the conversation. For now I have marked all of them as `Unresolved` and you can reply to them now.

JishnuGoyal:
> @JishnuGoyal Please do not mark any conversation which has been started by others as `Resolved`. Instead reply to those conversations so that the original person who started that conversation thread can check your reply and based on that `Resolve` the conversation. For now I have marked all of them as `Unresolved` and you can reply to them now.

I wasn't aware. Thanks for the tip, Noted! @rt4914

JishnuGoyal:
> ensure any CI issues are fixed before sending this back

I am not sure why `Unit Tests (Robolectric -- Gradle) / App Module Robolectric Tests (ubuntu-18.04)` is failing. 
Could you please suggest what changes am I supposed to make here?

PTAL @BenHenning

rt4914:
Will review it tomorrow. Thanks.

BenHenning:
> > ensure any CI issues are fixed before sending this back
> 
> I am not sure why `Unit Tests (Robolectric -- Gradle) / App Module Robolectric Tests (ubuntu-18.04)` is failing. Could you please suggest what changes am I supposed to make here?
> 
> PTAL @BenHenning

The Bazel & Gradle failures seem to be the same--you need to make sure all the broken tests are updated to include a test binding for the new platform parameter (since it's going to be required for a few tests).

JishnuGoyal:
Hello, @BenHenning 
regarding:
> The Bazel & Gradle failures seem to be the same--you need to make sure all the broken tests are updated to include a test binding for the new platform parameter (since it's going to be required for a few tests).

I read the error report from the CI and have added TestModules to 25 files and pushed a commit.

I have next 25 files committed and will be pushing them soon as well.

I just want to check if I'm on the right track because I'm manually trying to add the modules and there are several files in multiples of the numbers that I've already edited.

PTAL , @BenHenning meanwhile  ill continue editing more files

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
```diff
@@ -138,12 +146,43 @@ class LessonThumbnailImageViewTest {
...
+    }
+  }
+
   // TODO(#59): Figure out a way to reuse modules instead of needing to re-declare them.
   @Singleton
   @Component(
     modules = [
       RobolectricModule::class,
-      PlatformParameterModule::class, PlatformParameterSingletonModule::class,
+      TestModule::class, PlatformParameterSingletonModule::class,
```
BenHenning: Why isn't PlatformParameterModule sufficient (for here & other tests)? If it includes the new parameter, I wouldn't expect these tests to actually need to redefine the values unless they specifically need to change them (i.e. to enable the feature).

BenHenning (commented):
Thanks @JishnuGoyal. I'm glad you followed up to confirm, first. While your approach works, I think there may be a simpler and more correct way to approach it. PTAL at my newest comment.

JishnuGoyal:
> If it includes the new parameter, I wouldn't expect these tests to actually need to redefine the values unless they specifically need to change them (i.e. to enable the feature).

_Exactly!_
All of these test files that I edited, already had the `PlatformParameterModule` included - _**Yet**_ these tests were failing, complaining that they couldn't access the `enableEditAccountsOptionsUiValue` (and some other) parameter(s).
Ideally, the CI tests should never have failed since all these files already have `PlatformParameterModule` included.

It was then after which I removed and replaced `PlatformParameterModule` with the `TestModule` -- which is working fine for now.

PTAL @BenHenning and suggest if I should continue replacing them the same way
Thanks!

BenHenning:
> > If it includes the new parameter, I wouldn't expect these tests to actually need to redefine the values unless they specifically need to change them (i.e. to enable the feature).
> 
> _Exactly!_ All of these test files that I edited, already had the `PlatformParameterModule` included - _**Yet**_ these tests were failing, complaining that they couldn't access the `enableEditAccountsOptionsUiValue` (and some other) parameter(s). Ideally, the CI tests should never have failed since all these files already have `PlatformParameterModule` included.
> 
> It was then after which I removed and replaced `PlatformParameterModule` with the `TestModule` -- which is working fine for now.
> 
> PTAL @BenHenning and suggest if I should continue replacing them the same way Thanks!

@JishnuGoyal I think you need to add the new parameter to PlatformParameterModule in order to fix the error, not add new TestModules (except when the value needs to be overwritten). As far as I can tell, PlatformParameterModule isn't being updated here (unless I'm missing something?)

JishnuGoyal:
Hello @BenHenning , 

I was migrating from Windows environment to Linux Ubuntu, and had been facing some conflicts with the IDE and the new OS. Apologies for the conflicting commits, I'll revert these asap..
Please only consider the last 3 commit force pushed commits..

- I noticed that all the tests that are failing are not able to access Platform parameters from the included `PlatformParameterModule`. 
- Error shown (HomeActivityLocalTest.kt): [when `PlatformParameterModule` was included, as suggested in a file which doesn't require overwriting of the parameter values]
image

- It says it cannot access the parameter values unless they are provided with an `@provides` method. That is, it can only read parameter values when they are locally declared [ Hence in a local TestModule].
- I verified this: I created a duplicate TestModule (not local) called ParamModule, just to test if the test files could read parameter values from Modules that weren't locally defined. It failed.
- (I'll delete this ParamModule in the next commit)

JishnuGoyal:
The solution seems to add a local definition of `@provides` method, i.e. a TestModule to each of these files. 
How would you suggest I should proceed?

PTAL @BenHenning

app/src/sharedTest/java/org/oppia/android/app/customview/LessonThumbnailImageViewTest.kt
JishnuGoyal: Yes, I too expected the same behaviour. But for some reason these files aren't able to consume the Platform Parameter Values from `PlatformParameterModule`....But only when I define the functions generating parameter values locally inside the Test, i.e, in a  `TestModule`, the tests finally succeed.

JishnuGoyal:
Hello, @BenHenning 
I took the long route and wrote test modules for all the tests.

> Why isn't PlatformParameterModule sufficient (for here & other tests)? If it includes the new parameter, I wouldn't expect these tests to actually need to redefine the values unless they specifically need to change them (i.e. to enable the feature).

According to the Wiki <A href = "https://github.com/oppia/oppia-android/wiki/Platform-Parameters-&-Feature-Flags#how-to-write-tests-related-platform-parameter">here</A>, a testmodule is being added regardless of using or not a different value for that Platform Parameter (in case different value is used, some more steps are required).

PTAL @BenHenning and please give a hint why the remaining checks are failing...

JishnuGoyal:
PTAL @BenHenning @rt4914 Thanks!

BenHenning:

BenHenning:
@JishnuGoyal I don't think we want the extra test modules; it actually significantly complicates managing new platform parameters since we'd need to add them everywhere each time a new one is added. I think the correct path forward is figuring out why using PlatformParameterModule didn't work as that's the preferred approach.

JishnuGoyal:
Hello, @BenHenning 
I finally found out a way to use the PlatformParameterModule across all the tests.
FYI, removing platform parameter module from _any test_ breaks the other tests. 
I have also removed all the Test Modules.

PTAL @BenHenning  and let me know your thoughts

JishnuGoyal:
I believe some CI checks are breaking after merging with develop? Not sure but the local test on my machine suggests so, I think..

BenHenning:
Unfortunately I'll need to close this @JishnuGoyal. We don't allow force pushing since ti breaks review history, and we close PRs when it occurs. Please reopen the PR and copy over the existing comment threads so that we can track them over there.

I suggest also taking a bit to re-review the guidelines mentioned in https://github.com/oppia/oppia-android/wiki/making-a-code-change (it includes a note on not allowing force pushing, so I wanted to make sure you're familiar with all the other parts, too).

PR #4081, opened 2022-01-06
Fix #4002: Profiles cant have internationalised names

### LINKED ISSUE #4002: Profiles can't have internationalized names
**Describe the bug**
Profiles can't have internationalized names.

**To Reproduce**
1. Open the create profile flow
2. In the name box for the profile, type non-English characters
3. Attempt to create the profile
4. Observe that an error appears indicating that the profile can't be created

**Expected behavior**
Now that Oppia supports non-English languages, profile names should be able to also be internationalized.

**Demonstration**
image

**Environment**
- Android or SDK version (e.g. Android 5 or SDK 21): Android 8.1.0
- App version (you can get this through system app settings or via the admin controls menu in-app): Develop build on 04b8e0381e3d49a4a4a0f17e111386385bf4ffbb.

**Additional context**
None

JishnuGoyal:
## Explanation

Fixes #4002

This PR introduces a new injectable class into the codebase called `ProfileNameValidator`.

This class uses Kotlin functions to verify input, instead of Regex patterns which were used earlier. 

In addition to what functionalities the regex patterns supported, this solution will now accept all International characters(English + non-English letters), and these characters: `.` , `'`, `-`

It will also reject the following:

Numbers (0, 1, 2, ... 9)
Symbols (!@#$%^&*... etc.)
Parentheses and Brackets ( (, [, { )

These built-in Kotlin methods function do not interfere with the RTL markers.

#### Hindi and Arabic profiles successfully created:
image

## Essential Checklist


JishnuGoyal:
Note that this regex pattern _will_ allow mixing of English and non English characters as some languages use of combination of those, for ex, Spanish, German, etc.

Kindly let me know if any further changes are required in the logic part @srushtirk

JishnuGoyal:
PTAL @BenHenning @bkaur-bkj Thanks!

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -722,7 +722,7 @@ class ProfileManagementController @Inject constructor(
   }
 
   private fun onlyLetters(name: String): Boolean {
```
BenHenning: This name no longer matches--please update.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -722,7 +722,7 @@ class ProfileManagementController @Inject constructor(
...
-    return name.matches(Regex("^[ A-Za-z]+\$"))
+    return name.matches(Regex("^(?!.*^[-.'])^(?!.*[.'-]\$).*^(?!.*[0-9!@#\$%^&*()_+={}\\[\\]|\\\\;\",?/<>`~]).*[\\w\\u00BF-\\u1FFF\\u2C00-\\uD7FF.'-]"))
```
BenHenning: This pattern has grown to an unmanageable size, I think. :) Can you instead split this into multiple, more specific patterns and check them one-at-a-time? I suggest moving them to fields to provide semantic meaning (such as having one for verifying that disallowed symbols are absent)l.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -722,7 +722,7 @@ class ProfileManagementController @Inject constructor(
...
   }
```
BenHenning: Please add tests for various names that should be prohibited/allowed to ensure the pattern is working as expected.

BenHenning (changes requested):
Thanks @JishnuGoyal! Left some initial comments--PTAL.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
JishnuGoyal: Hello, @srushtirk , could you give some test cases/ examples of what names are acceptable and which ones are not acceptable?
What I want to make sure is that:
Although you have allowed `.`, `-` and `'` in the name, can the name start/end with one of these symbols?(For ex, is `-jishnu..` allowed?)

This is one of the constraints I found, but I want to know if there are any more of such constraints that are needed to be taken care of. 
PTAL @srushtirk Thanks!

JishnuGoyal:
Hello, @srushtirk  , could you give some test cases/ examples of what names are acceptable and which ones are not acceptable?
What I want to make sure is that:
Although you have allowed `.`, `-` and `'` in the name, can the name start/end with one of these symbols?
(For ex, is `-jishnu..` allowed?)

This is one of the constraints I found, but I want to know if there are any more of such constraints that are needed to be taken care of.
PTAL @srushtirk Thanks!

srushtirk:
Hi @JishnuGoyal let's allow any order for the symbols . , ', -, English and non English characters. As it is difficult to predict where exactly in the user's name would these symbols be. This causes some edge cases where users can create profiles with random names starting or ending with these symbols, which is fine to trade off with. 

We can put the following restrictions though:
1) No two symbols next to each other. Eg: --, '-, .', ..
2) Avoid spaces in profile names

JishnuGoyal:
PTAL @BenHenning , Thanks!

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -721,8 +721,29 @@ class ProfileManagementController @Inject constructor(
     return imageFile.absolutePath
   }
 
-  private fun onlyLetters(name: String): Boolean {
-    return name.matches(Regex("^[ A-Za-z]+\$"))
+  private fun nameAllowed(name: String): Boolean {
```
BenHenning: Given that we're considering also adding a bad word filter in the future, it might be prudent to create a new NameValidator class in the utility module and have name logic added there (along with dedicated tests so that you can more easily cover edge cases vs. having to test those through the profile management controller).

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -721,8 +721,29 @@ class ProfileManagementController @Inject constructor(
...
+        noRepeatedUseOfAllowedSymbols(name)
+      )
+  }
+
+  private fun notEmptyNoSpacesAndContainsLetters(name: String): Boolean {
+    return name.matches(Regex("^.[\\w\\u00BF-\\u1FFF\\u2C00-\\uD7FF.'\\-]+\$"))
+  }
+
+  private fun noNumbers(name: String): Boolean {
+    return name.matches(Regex("^[^0-9]*\$"))
```
BenHenning: You could also do: ``name.none { it.isDigit() }`` which might actually be more performant than the regex check.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -721,8 +721,29 @@ class ProfileManagementController @Inject constructor(
...
+  }
+
+  private fun noSymbols(name: String): Boolean {
+    return name.matches(Regex("^[^#*!@\$%^&()_+=\\\\|\\]\\[\":;?/><,`~{}]*\$"))
+  }
+
+  private fun noRepeatedUseOfAllowedSymbols(name: String): Boolean {
+    return name.matches(Regex("^(?!.*([.'-]{2})).*"))
```
BenHenning: Could you use a containsMatch here and just check again ``[\.'-]{2}``?

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -721,8 +721,29 @@ class ProfileManagementController @Inject constructor(
...
+    return (
+      notEmptyNoSpacesAndContainsLetters(name) &&
+        noNumbers(name) &&
+        noSymbols(name) &&
+        noRepeatedUseOfAllowedSymbols(name)
+      )
+  }
+
+  private fun notEmptyNoSpacesAndContainsLetters(name: String): Boolean {
+    return name.matches(Regex("^.[\\w\\u00BF-\\u1FFF\\u2C00-\\uD7FF.'\\-]+\$"))
```
BenHenning: Similarly to below, you could have a separate check that uses ``isWhitespace`` to simplify this just to checking for letters. That being said, would ``isLetter`` also work instead of using the pattern?

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -721,8 +721,29 @@ class ProfileManagementController @Inject constructor(
...
+  }
+
+  private fun noNumbers(name: String): Boolean {
+    return name.matches(Regex("^[^0-9]*\$"))
+  }
+
+  private fun noSymbols(name: String): Boolean {
+    return name.matches(Regex("^[^#*!@\$%^&()_+=\\\\|\\]\\[\":;?/><,`~{}]*\$"))
```
BenHenning: Nit: you ought to define regex patterns as vals so that they don't need to be recompiled each time.

Also, why doesn't the letter check above cover the symbols case?

BenHenning (changes requested):
Thanks @JishnuGoyal! Added some more comments for ways I think this could be simplified, and made more future-proof.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
JishnuGoyal: This method ensures that the name isn't empty (doesn't allow blank space, next line), but at the same time it matches letters of "any language" -> so isLetters() can't be used. Please Note that this regex pattern doesn't reject any symbols, it merely returns `true` when it there is a `letter` (of any language) is present in the name.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
JishnuGoyal: > Also, why doesn't the letter check above cover the symbols case?

Since the regex pattern mentioned here https://github.com/oppia/oppia-android/pull/4081#discussion_r787941521 doesn't reject symbols, this function was added.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
JishnuGoyal: > along with dedicated tests so that you can more easily cover edge cases vs. having to test those through the profile management controller

Interesting! Should I also move the tests related to allowed and unallowed names from profile management controller's test to name validator class's test or keep duplicate tests as well?

JishnuGoyal:
This was my first time creating a new class in this project. PTAL @BenHenning

BenHenning:
Thanks @JishnuGoyal. Apologies, I'll need to take another pass on this tomorrow.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
+package org.oppia.android.util.namevalidator
```
BenHenning: I suggest putting this under the profile subpackage since it's still relevant to profiles. You'll also need to update the BUILD.bazel file in that package to include a new android_library for this validator.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class NameValidator {
```
BenHenning: I suggest making this an ``@Inject``-able class (see https://github.com/oppia/oppia-android/blob/develop/utility/src/main/java/org/oppia/android/util/profile/DirectoryManagementUtil.kt for a reference, and its usages) since that'll allow for better test & dependency management.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+  companion object {
+    private const val MATCH_INTERNATIONAL_LETTERS_REJECT_SPACES =
+      "^.[\\w\\u00BF-\\u1FFF\\u2C00-\\uD7FF.'\\-]+\$"
```
BenHenning: How did you determine this range? Is there any documentation that you can point to?

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    private const val REJECT_SYMBOLS = "^[^#*!@\$%^&()_+=\\\\|\\]\\[\":;?/><,`~{}]*\$"
```
BenHenning: Ditto here--why reject these specific symbols?

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    private const val REJECT_REPEATED_USE_OF_ALLOWED_SYMBOLS = "[\\.'-]{2}"
+
+    fun validate(name: String): Boolean {
```
BenHenning: Suggest instead: isNameValid.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private fun notEmptyNoSpacesAndContainsLetters(name: String): Boolean {
+      return name.matches(Regex(MATCH_INTERNATIONAL_LETTERS_REJECT_SPACES))
+    }
+
+    private fun noNumbers(name: String): Boolean {
+      return name.none {
+        it.isDigit()
+      }
```
BenHenning: This can fit on one line--please update.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun validate(name: String): Boolean {
+      return (
+        notEmptyNoSpacesAndContainsLetters(name) &&
+          noNumbers(name) &&
+          noSymbols(name) &&
+          noRepeatedUseOfAllowedSymbols(name)
+        )
+    }
+
+    private fun notEmptyNoSpacesAndContainsLetters(name: String): Boolean {
```
BenHenning: Method names should be verb phrases, e.g.: ``containsLettersAndNoSpaces``.

Ditto for the others.

You could also consider making these string extension function, e.g.:

```kotlin
private fun String.containsLettersAndNoSpaces(): Boolean =
  matches(Regex(MATCH_INTERNATIONAL_LETTERS_REJECT_SPACES))
```

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+      return name.matches(Regex(MATCH_INTERNATIONAL_LETTERS_REJECT_SPACES))
```
BenHenning: Suggest saving these ``Regex`` objects in lazy-initialized ``val``s at the companion object level.

utility/src/test/java/org/oppia/android/util/namevalidator/NameValidatorTest.kt
```diff
@@ -0,0 +1,27 @@
+package org.oppia.android.util.namevalidator
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
```
BenHenning: Please use Truth assertions, instead (see uses of ``assertThat`` elsewhere in the codebase).

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
BenHenning: Why wouldn't ``isLetter`` work? It's my understanding that it should be checking for letters in roughly the same range that you're using here.

Also, ``isWhitespace`` should be the same as ``\\w`` (and you could specifically verify this by adding tests to cover various whitespace).

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
BenHenning: I think you should focus on adding new test cases that are separate from the ProfileManagementController's tests (you also get redundancy this way).

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
BenHenning: I'm actually wondering if we could just check that it only contains letters and no repeated allowed symbols. Wouldn't the "else" case there be all other symbols (including whitespace, digits, and these symbols)?

BenHenning (changes requested):
Thanks @JishnuGoyal! Took a first pass. I think this is definitely on the right track, but I have a bunch of suggestions. I'll take a more thorough pass on tests in my next review pass, but in the meantime please also:
- Add KDocs for all non-private classes and methods
- Add tests to more thoroughly cover the various cases that the regex patterns can handle
- Fix the broken tests & CI checks

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
JishnuGoyal: Please have a look at the new android_library and suggest if deps [ ] field is to be added

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
JishnuGoyal: I found a new in-built way of determining if the input is a letter or not. This implementation is essentially a contraction of that range. It is much more clear now :)

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
JishnuGoyal: i'm not sure about this one... Is there a reference in the codebase I could use?

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
JishnuGoyal: Awesome! But what should I `inject` via the constructor as a parameter?

JishnuGoyal:
PTAL @BenHenning 
I will be adding more-tests to check edge cases after the following pass :)
Thanks!

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -167,23 +171,43 @@ class ProfileManagementControllerTest {
   }
 
   @Test
-  fun testAddProfile_addProfileWithNumberInName_checkResultIsFailure() {
+  fun testAddProfile_addProfilesWithUnAllowedNames_checkResultIsFailure() {
```
```suggestion
  fun testAddProfile_addProfilesWithDisallowedNames_checkResultIsFailure() {
```

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -111,6 +111,10 @@ class ProfileManagementControllerTest {
     Profile.newBuilder().setName("Veena").setPin("567").setAllowDownloadAccess(true).build()
   )
 
+  private val allowedNamesList = listOf<String>("जिष्णु", "Ben-Henning", "Rajat.T", "جيشنو")
+
+  private val unAllowedNamesList = listOf<String>("जिष्णु7", "Ben_Henning", "Rajat..T", "جيشنو^&&")
```
BenHenning: Nit: suggest disallowedNames

(strike 'List' from both, and switch 'unAllowed' with 'disallowed').

utility/src/main/java/org/oppia/android/util/profile/NameValidator.kt
```diff
@@ -0,0 +1,33 @@
...
+      return (onlyLettersAndAllowedSymbols(name) && noRepeatedUseOfAllowedSymbols(name))
+    }
+
+    private fun onlyLettersAndAllowedSymbols(name: String): Boolean {
+      return name.matches(Regex(MATCH_LETTERS_AND_ALLOWED_SYMBOLS))
+    }
+
+    private fun noRepeatedUseOfAllowedSymbols(name: String): Boolean {
+      return !name.contains(Regex(REJECT_REPEATED_USE_OF_ALLOWED_SYMBOLS))
+    }
```
BenHenning: These should be member functions, instead, and the validator should be injected.

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
```diff
@@ -0,0 +1,26 @@
...
+  private val allowedNamesList = listOf<String>("जिष्णु", "Ben-Henning", "Rajat.T", "جيشنو")
+
+  private val unAllowedNamesList =
+    listOf<String>("जिष्णु7", "Ben_Henning", "Rajat..T", "جيشنو^&&", " ", ".", "Ben Henning")
+
+  @Test
+  fun testNameValidator_addUnAllowedName_returnFalse() {
+    unAllowedNamesList.forEach {
+      Truth.assertThat(!NameValidator.isNameValid(it))
+    }
```
BenHenning: Each of the names should be a distinct test for each case being tested (such as English characters, Hindi characters, Portuguese characters, etc., allowed symbols, and failing cases such as disallowed symbols or characters).

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  @Test
+  fun testNameValidator_addAllowedName_returnTrue() {
+    allowedNamesList.forEach {
+      Truth.assertThat(NameValidator.isNameValid(it))
```
BenHenning: You're not asserting anything. This should be: assertThat(value).isEqualTo(expectedValue), though because you're testing a boolean you can just use 'isTrue/isFalse'. See  other assertThat() calls in the codebase for reference.

You should also directly import assertThat (per convention & best practices).

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
BenHenning: It would look something like:

```kotlin
private val letterRegex by lazy { Regex("...") }
```

Then you would use ``letterRegex`` when needed.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
BenHenning: You don't need to inject anything into the validator class, but you do need to inject the validator.

utility/src/main/java/org/oppia/android/util/namevalidator/NameValidator.kt
BenHenning: I suggest setting up Bazel locally per the wiki and trying to build the app locally (since there are other BUILD files that will need to be updated here).

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
BenHenning: Reminder to respond to this--isLetter would be a large simplification.

BenHenning (commented):
Thanks @JishnuGoyal. Took another pass.

JishnuGoyal:
I'm migrating from Windows environment to Linux Ubuntu, and have been facing some conflicts with the IDE and the new OS. Apologies for the conflicting commits, I'll revert these asap..

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
JishnuGoyal: > English characters, Hindi characters, Portuguese characters

Could you please clarify which all language characters should I add tests for?

JishnuGoyal:
>  I suggest setting up Bazel locally per the wiki and trying to build the app locally (since there are other BUILD files that will need to be updated here).


It took me a while to set up Bazel and linux environment properly. I think the Bazel build on my current branch is failing because I either haven't properly set up Build.Bazel rules for the new NameValidator class correctly, or injected it incorrectly inside `ProfileManagementController`...
 A little push into the right direction would be very helpful!
image
###### it complains that name validator is unresolved at `ProfileManagmentController`

PTAL @BenHenning

JishnuGoyal:
> Reminder to respond to this--isLetter would be a large simplification.

isLetter() recognizes only english letters --> Every other symbol is "not a letter" to it, including letters from other languages...
Hence it will return false in case anything is input other than the English Alphabet.

PTAL @BenHenning

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -167,23 +171,43 @@ class ProfileManagementControllerTest {
...
+        avatarImagePath = null,
+        allowDownloadAccess = false,
+        colorRgb = -10710042,
+        isAdmin = true
+      ).toLiveData().observeForever(mockUpdateResultObserver)
+      testCoroutineDispatchers.runCurrent()
+
+      verifyUpdateFailed()
+      assertThat(updateResultCaptor.value.getErrorOrNull()).hasMessageThat()
+        .contains("$it does not contain only letters")
```
BenHenning: This error seems no longer correct--we should update it.

@srushtirk do you have suggestions on what the new error message should be?

utility/src/main/java/org/oppia/android/util/profile/BUILD.bazel
```diff
@@ -14,3 +14,11 @@ kt_android_library(
         "//third_party:javax_inject_javax_inject",
     ],
 )
+
+kt_android_library(
+    name = "name_validator",
+    srcs = [
+        "NameValidator.kt",
+    ],
+    visibility = ["//domain:__subpackages__","//utility:__subpackages__"],
```
BenHenning: I suggest just making this API visibility (i.e. ``//:oppia_api_visibility``) since we don't specify more granular visibility unless it's a relative sub-package, test-only, or a module.

utility/src/main/java/org/oppia/android/util/profile/NameValidatorUtil.kt
```diff
@@ -0,0 +1,20 @@
+package org.oppia.android.util.profile
+
+/** Utility class for [NameValidator]. */
+object NameValidatorUtil {
```
BenHenning: I'm not sure I see the point of this class--why not include these properties within NameValidator itself?

utility/src/main/java/org/oppia/android/util/profile/NameValidator.kt
```diff
@@ -0,0 +1,32 @@
+package org.oppia.android.util.profile
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Utility to validate profile names */
+@Singleton
+class NameValidator @Inject constructor() {
```
BenHenning: Nit: ProfileNameValidator would probably be clearer since 'name' is a bit generic in itself.

utility/src/main/java/org/oppia/android/util/profile/NameValidator.kt
```diff
@@ -0,0 +1,32 @@
+package org.oppia.android.util.profile
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Utility to validate profile names */
```
BenHenning:
```suggestion
/** Utility to validate that profile names are correctly formatted. */
```

utility/src/main/java/org/oppia/android/util/profile/NameValidator.kt
```diff
@@ -0,0 +1,32 @@
...
+@Singleton
+class NameValidator @Inject constructor() {
+  private val letterAndSymbolsRegex by lazy {
+    Regex("^.[${NameValidatorUtil.lettersAndSymbolsRegexString}]+\$")
+  }
+
+  private val noRepeatedAllowedSymbolsRegex by lazy { Regex("[.'-]{2}") }
+
+  /**
+   * Validates names for a profile
```
BenHenning:
```suggestion
   * Validates a profile name to ensure that it isn't confusingly formatted or contains invalid characters.
```

(May require line wrapping).

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  @Test
+  fun testNameValidator_addDisallowedName_returnFalse() {
+    disallowedNames.forEach {
+      assertThat(nameValidator.isNameValid(it)).isEqualTo(false)
+    }
+  }
+
+  @Test
+  fun testNameValidator_addAllowedName_returnTrue() {
```
BenHenning: Rather than organizing these tests as pass/fail, you should structure them based on the specific behavior that's being tested, the condition that triggers that behavior, and subsequent outcome. Each behavior should be in its own test. For example:

``testIsNameValid_nameWithSpaces_returnsFalse``

or

```testNameIsValid_nameHasOneDash_returnsTrue``

etc.

File: utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt [hunk 30]
```diff
@@ -0,0 +1,36 @@
...
+    allowedNames.forEach {
+      assertThat(nameValidator.isNameValid(it)).isEqualTo(true)
```
BenHenning: The more Truthy way to do this is:

```suggestion
      assertThat(nameValidator.isNameValid(it)).isTrue()
```

BenHenning (changes requested):
> > Reminder to respond to this--isLetter would be a large simplification.
> 
> isLetter() recognizes only english letters --> Every other symbol is "not a letter" to it, including letters from other languages... Hence it will return false in case anything is input other than the English Alphabet.
> 
> PTAL @BenHenning

Out of curiosity, where did you obverse this behavior @JishnuGoyal? Per https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-letter.html and the corresponding JVM implementation (https://github.com/JetBrains/kotlin/blob/6a670dc5f38fc73eb01d754d8f7c158ae0176ceb/libraries/stdlib/jvm/src/kotlin/text/CharCategoryJVM.kt#L22) isLetter should be conforming to the category "Ll" of the Unicode specification which means every character listed in https://www.compart.com/en/unicode/category/Ll should return true. https://www.compart.com/en/unicode/category/Lu and https://www.compart.com/en/unicode/category/Lo are also matched. Per [this Kotlin](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjEwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiLyoqXG4gKiBZb3UgY2FuIGVkaXQsIHJ1biwgYW5kIHNoYXJlIHRoaXMgY29kZS4gXG4gKiBwbGF5LmtvdGxpbmxhbmcub3JnIFxuICovXG5cbmZ1biBtYWluKCkge1xuICAgIHByaW50bG4oJ2EnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignw78nLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignxJEnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignxaknLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignxbMnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignxr4nLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignyKEnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbign0bEnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbignzrYnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbign0asnLmlzTGV0dGVyKCkpXG4gICAgcHJpbnRsbign4YOTJy5pc0xldHRlcigpKVxuICAgIHByaW50bG4oJ+CkhScuaXNMZXR0ZXIoKSlcbn0ifQ==) code, all of the tested characters across multiple scripts (including on Devanagari at the end) all return true for 'isLetter'. Am I maybe missing something?

Beyond the above, I took another pass on the PR--PTAL.

JishnuGoyal:
> Out of curiosity, where did you obverse this behavior @JishnuGoyal? 

@BenHenning, have a look at <a href = "https://pl.kotl.in/KHcwbnFzy">this code sample</a>. This is a manual test I did using isLetter(). Please feel free to press the `run` option and check the implementation.

> all of the tested characters across multiple scripts (including on Devanagari at the end) all return true for 'isLetter'. Am I maybe missing something?

The first test case, `जिष्णु` [  `Jishnu` in Hindi ] in the above code sample fails only due to using isLetter() , as it cannot recognize complex usage of [HINDI, devnagri] literals, such as in names. 
This is because the formation of `letters` in hindi (and probably some other languages too) changes according to the `word`. 

Although isLetter() works fine for _individual letters_, in case of some words, it returns false when it actually should return true due to complex usage of the letters.
#### Another way of saying is that we are using letters which are beyond the set of letters conforming to the category "Ll" of the Unicode specification.

_Another problem is that isLetter() will also `reject` allowed symbols._ [This is why all the other test cases fail in the code sample]

<a href = "https://pl.kotl.in/3ZFRtuPWQ">Here is a code sample</a> using regex which passes the same tests.

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
JishnuGoyal: Added. PTAL and let me know if more tests are needed

JishnuGoyal:
PTAL @BenHenning

BenHenning:

utility/src/main/java/org/oppia/android/util/profile/BUILD.bazel
```diff
@@ -20,5 +20,5 @@ kt_android_library(
     srcs = [
         "NameValidator.kt",
```
BenHenning: Looks like this also needs to be renamed.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
```diff
@@ -0,0 +1,71 @@
...
+  @Test
+  fun testIsNameValid_nameWithAllowedSymbols_returnsTrue() {
+    val namesWithAllowedSymbol = listOf<String>("Ben-Henning", "Rajat.T", "G'Jishnu")
+    namesWithAllowedSymbol.forEach {
+      assertThat(profileNameValidator.isNameValid(it)).isTrue()
+    }
+  }
+
+  @Test
+  fun testIsNameValid_nameWithRepeatedAllowedSymbols_returnsFalse() {
```
BenHenning, line 79: I think this will be a good candidate for parameterized tests, though that'll need to wait for my chain of math expressions PRs to be merged (which might take another week or so). Are you okay with waiting for that?

BenHenning (commented):
> > Out of curiosity, where did you obverse this behavior @JishnuGoyal?
> 
> @BenHenning, have a look at [this code sample](https://pl.kotl.in/KHcwbnFzy). This is a manual test I did using isLetter(). Please feel free to press the `run` option and check the implementation.
> 
> > all of the tested characters across multiple scripts (including on Devanagari at the end) all return true for 'isLetter'. Am I maybe missing something?
> 
> The first test case, `जिष्णु` [ `Jishnu` in Hindi ] in the above code sample fails only due to using isLetter() , as it cannot recognize complex usage of [HINDI, devnagri] literals, such as in names. This is because the formation of `letters` in hindi (and probably some other languages too) changes according to the `word`.
> 
> Although isLetter() works fine for _individual letters_, in case of some words, it returns false when it actually should return true due to complex usage of the letters.
> 
> #### Another way of saying is that we are using letters which are beyond the set of letters conforming to the category "Ll" of the Unicode specification.
> _Another problem is that isLetter() will also `reject` allowed symbols._ [This is why all the other test cases fail in the code sample]
> 
> [Here is a code sample](https://pl.kotl.in/3ZFRtuPWQ) using regex which passes the same tests.

@JishnuGoyal thanks for the example. Might https://developer.android.com/reference/kotlin/java/lang/Character#isAlphabetic(kotlin.Int) work instead? I think the issue is making sure that modifier and control characters are also considered (since we need to make sure RTL markers also aren't ignored since they can appear).

Besides the above, I took another pass on the PR but didn't focus as much on the util impl and tests since a lot is in-flux at the moment. Please also make sure that the CI checks are passing before sending this back.

utility/src/main/java/org/oppia/android/util/profile/BUILD.bazel
JishnuGoyal: done

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
JishnuGoyal: Sure @BenHenning . Is it ready yet?

JishnuGoyal:
> @JishnuGoyal thanks for the example. Might https://developer.android.com/reference/kotlin/java/lang/Character#isAlphabetic(kotlin.Int) work instead? I think the issue is making sure that modifier and control characters are also considered (since we need to make sure RTL markers also aren't ignored since they can appear).

Thank you so much @BenHenning for this suggestion. 
I must bring into notice that although isAlphabet() handles all the cases well, some hindi characters as mentioned before will not be recognised (I ran tests and verified). [the current impl handles it correctly though]

Also, isAlphabet is only available in Java, so an additional Java class will be introduced to use it. If we are ready to continue with this, may i push the required code?

JishnuGoyal:
PTAL @BenHenning , looking to confirm if im on the right track, i believe making a Java file wouldnt be such a good idea..

utility/src/main/java/org/oppia/android/util/profile/BUILD.bazel
```diff
@@ -16,9 +17,34 @@ kt_android_library(
...
+        "ProfileNameValidatorModule.kt",
+    ],
+    visibility = ["//:oppia_prod_module_visibility"],
+    deps = [
+        ":dagger",
+        ":profile_name_validator_impl",
+    ],
+)
+
+dagger_rules()
```
BenHenning: Please add missing EOF newline.

BenHenning:
Apologies @JishnuGoyal. I'll need to delay my review on this due to being a bit slammed this week due to a release we're preparing for later in the week.

JishnuGoyal:
@seanlip Could you please reopen this as well? 
Thanks!

JishnuGoyal:
PTAL @BenHenning

BenHenning:

I'll take a pass on Monday.In the meantime, you should update to the latest develop branch @JishnuGoyal and address any conflicts.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
BenHenning: Yep! You can look at the following for an example: https://github.com/oppia/oppia-android/blob/develop/domain/src/test/java/org/oppia/android/domain/classify/rules/numericexpressioninput/NumericExpressionInputIsEquivalentToRuleClassifierProviderTest.kt.

BenHenning:
Hi @JishnuGoyal. Sorry for the long review. I spent a while looking into different options, and I think the best option is just to implement our own character-based alphabetic check. This ensures proper Unicode compatibility (which is difficult with Regex), but accounts for the mark characters used in Hindi. See my implementation for reference: https://pl.kotl.in/V7hAJ9lsh. This seems to work correctly for the characters in your name, and it avoids needing to depend on the Android Character.isAlphabetic function so it can be done completely in Kotlin without any warnings. Let me know what you think.

Once we finalize the checking approach, I'll take a full pass on the PR.

JishnuGoyal:
> Hi @JishnuGoyal. Sorry for the long review. I spent a while looking into different options, and I think the best option is just to implement our own character-based alphabetic check. This ensures proper Unicode compatibility (which is difficult with Regex), but accounts for the mark characters used in Hindi. See my implementation for reference: https://pl.kotl.in/V7hAJ9lsh. This seems to work correctly for the characters in your name, and it avoids needing to depend on the Android Character.isAlphabetic function so it can be done completely in Kotlin without any warnings. Let me know what you think.

Thank you so much @BenHenning for this awesome suggestion. CharCategory class actually solves the problem in Kotlin itself. I have added the required necessary changes and pushed them.

JishnuGoyal:
Talking about the CI checks, I do need a little push in the right direction here --
I have gone through the logs but I don't understand why the ProfileNameValidator class is not being recognised properly --
I have injected this class just the way other classes are being injected -- 
For reference, I saw how OppiaClock has been injected and did the same with ProfileNameValidator.

JishnuGoyal:
I have also set up the bazel configuration for `ProfileNameValidator.`
Why is it failing?

Also while we are at it, would it be okay to not _Inject_ `ProfileNameValidator` and simply use an object of it in `ProfileManagementController` ?

PTAL @BenHenning 
I feel its been a while this PR has been open and I hope this could be resolved soon, Thanks!

BenHenning:
@JishnuGoyal I think you don't need a module here (specifically ``ProfileNameValidatorModule``). Modules are only needed when we want to provide different implementations. We can forego the interface and just provide an injectable implementation class.

The error is coming from the fact that the module needs to be included in every suite that depends on ``ProfileManagementControllerTest`` (which is, to say, almost every test in the codebase). Not having a new module avoids all of this and should simplify things quite a bit.

JishnuGoyal:
> @JishnuGoyal I think you don't need a module here (specifically ProfileNameValidatorModule). Modules are only needed when we want to provide different implementations. We can forego the interface and just provide an injectable implementation class.
The error is coming from the fact that the module needs to be included in every suite that depends on ProfileManagementControllerTest (which is, to say, almost every test in the codebase). Not having a new module avoids all of this and should simplify things quite a bit.

Thank you so much for this tip! I have added these changes :)

Other than the lint test, all the tests are failing due to an unrelated class, ProfileEditActivityTest --
Could you please tell how should i proceed with this? 

PTAL @BenHenning

app/src/main/java/org/oppia/android/app/application/ApplicationComponent.kt
```diff
@@ -99,6 +99,7 @@ import javax.inject.Singleton
     LocaleProdModule::class, ActivityRecreatorProdModule::class,
     NumericExpressionInputModule::class, AlgebraicExpressionInputModule::class,
     MathEquationInputModule::class, SplitScreenInteractionModule::class,
+    LocaleProdModule::class, ActivityRecreatorProdModule::class,
```
BenHenning: This appears duplicated with a few lines above. Please remove.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
+package org.oppia.android.util.profile
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Utility to validate that profile names are correctly formatted. */
+@Singleton
```
BenHenning: This can be omitted--scopes are only needed if there's state (otherwise it's fine for the Dagger provider to recreate the class).

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+class ProfileNameValidator @Inject constructor() {
+  private val noRepeatedAllowedSymbolsRegex by lazy { Regex("[.'-]{2}") }
+
+  /**
+   * Validates a profile name to ensure that it isn't confusingly formatted
+   * or contains invalid characters.
```
BenHenning: Two nits:
1. Please make sure documentation lines take up as much horizontal space as possible before line-wrapping (i.e. up to 100 characters which is our line limit).
2. Perhaps add some clarifying details about what 'invalid' means (e.g. non-letter characters, per the Unicode standard for letter categories, plus apostrophes and dashes).

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @param name name of the profile
+   * @return if the profile name is allowed or not
+   */
+  fun isNameValid(name: String): Boolean {
+    return (onlyLettersAndAllowedSymbols(name) && noRepeatedUseOfAllowedSymbols(name))
```
BenHenning:
```suggestion
    return onlyLettersAndAllowedSymbols(name) && noRepeatedUseOfAllowedSymbols(name)
```

Small nit, but we don't typically add redundant parentheses.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /** Validates if the character in the name is an alphabet or an allowed symbol or not. */
+  private fun onlyLettersAndAllowedSymbols(name: String): Boolean {
```
BenHenning: We generally keep method names as verbs or verb phrases. Suggest renaming to something like: ``containsOnlyLettersAndAllowedSymbols``.

Ditto for other methods in this class.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+@Singleton
+class ProfileNameValidator @Inject constructor() {
+  private val noRepeatedAllowedSymbolsRegex by lazy { Regex("[.'-]{2}") }
+
+  /**
+   * Validates a profile name to ensure that it isn't confusingly formatted
+   * or contains invalid characters.
+   *
+   * @param name name of the profile
+   * @return if the profile name is allowed or not
```
BenHenning: Perhaps instead: whether it's a valid, acceptable name?

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /** Validates if the character in the name is an alphabet or an allowed symbol or not. */
+  private fun onlyLettersAndAllowedSymbols(name: String): Boolean {
+    name.forEach {
+      if (!(it.isAlphabetic() || isSymbolAllowed(it))) {
+        return false
+      }
+    }
+    return true
```
BenHenning: Can be simplified and made more Kotlin-idiomatic:

```kotlin
return name.all { it.isAlphabetic() || isSymbolAllowed(it) }
```

Also given the above, perhaps rename the latter helper to ``isAllowedSymbol`` since I think it'll be less confusion ('symbol' could be construed to mean any written symbol which then puts it a bit at odds with allowed alphabetic characters).

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
+package org.oppia.android.util.profile
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Utility to validate that profile names are correctly formatted. */
+@Singleton
+class ProfileNameValidator @Inject constructor() {
+  private val noRepeatedAllowedSymbolsRegex by lazy { Regex("[.'-]{2}") }
```
BenHenning: Does the '.' need to be escaped? I'm assuming that names like "Jerry" would fail right now (would be nice to add a test case for that scenario to verify a fix).

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+  private fun noRepeatedUseOfAllowedSymbols(name: String): Boolean {
+    return !name.contains(noRepeatedAllowedSymbolsRegex)
+  }
+
+  private fun isSymbolAllowed(symbol: Char): Boolean {
+    return symbol == '.' || symbol == '-' || symbol == '\''
+  }
+
+  private fun Char.isAlphabetic(): Boolean {
+    /**
```
BenHenning: Should be a block comment, or multiple line comments. KDocs should only be reserved for top-level members like classes, objects, functions, and values.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
BenHenning: Optional: you could instead make the name a receiver, e.g. ``String.containsOnlyLettersAndAllowedSymbols``. Ditto for the other check method, below.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test
+import javax.inject.Inject
+
+class ProfileNameValidatorTest {
+  @Inject
+  lateinit var profileNameValidator: ProfileNameValidator
+
+  @Before
+  fun setup() {
+    profileNameValidator = ProfileNameValidator()
```
BenHenning: This should be injected as I mentioned earlier. Are you running into any issues with setting this up as a Robolectric test? I suggest referencing other utility and domain tests for a reference (you probably won't need any modules nor a custom component, so it should be a relatively simple setup).

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Test
+  fun testIsNameValid_nameWithNumber_returnsFalse() {
+    val nameWithNumber = "Jishnu7"
+    assertThat(profileNameValidator.isNameValid(nameWithNumber)).isFalse()
+  }
+
+  @Test
+  fun testIsNameValid_nameWithDisallowedSymbol_returnsFalse() {
```
BenHenning, line 59: Since we were waiting for parameterized tests, I suggest using that here, instead, to avoid the for-loop/control flow below. See https://github.com/oppia/oppia-android/blob/develop/utility/src/test/java/org/oppia/android/util/math/MathExpressionParserTest.kt for an example test that uses parameterization.

Ditto for other tests that are verifying multiple things.

BenHenning (commented):
Thanks @JishnuGoyal! Nice tests & overall implementation. I find this much cleaner and easier to understand!

I had a bunch of comments, but most of them are just nits. PTAL and let me know if you have any follow-up questions. I think after the latest batch of comments this PR will probably be more or less LGTM.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/application/ApplicationComponent.kt
JishnuGoyal: done

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: Noted. Done.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: done

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: done

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: done!

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: Yep, done :)

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: noted, thanks. Done.

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: I feel the implementation would be cleaner this way because  otherwise, inside the method, "it" keyword will have to be used twice which will have two different meanings each time (String and char).

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
JishnuGoyal: No, the '.' doesn't need to be escaped -- Kotlin actually complains an redundant escape when I tried doing it. Also, "Jerry" passes the test, and I'm adding a test case just in case.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
JishnuGoyal: done :)

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
JishnuGoyal: done :)

utility/src/main/java/org/oppia/android/util/profile/BUILD.bazel
JishnuGoyal: done

JishnuGoyal:
PTAL @BenHenning is it correct way to write tests? 
Also, are any additional changes required to make parameterised tests work with Build.Bazel ?

utility/src/main/java/org/oppia/android/util/profile/ProfileNameValidator.kt
BenHenning: Ah, did you escape it with ``\\.``? Though, if it is actually being treated as '.' correctly, then this is probably a moot point.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: Is this still unresolved @JishnuGoyal?

Also, I think we should make sure that there are at least a few tests for name validation in ProfileManagementControllerTest, too, such as:
- One valid name with Latin characters
- One valid name with Devanagari characters (which would fail without the new validator)
- One invalid name (with the correct presented)

What do you think?

utility/src/test/java/org/oppia/android/util/profile/NameValidatorTest.kt
BenHenning: I think the latest tests are sufficient.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    val nameWithNumber = "Jishnu7"
+    assertThat(profileNameValidator.isNameValid(nameWithNumber)).isFalse()
+  }
+
+  @Test
+  @RunParameterized(
+    Iteration("Ben#Henning", "name=Ben#Henning"),
+    Iteration("Rajay@T", "name=Rajay@T"),
+    Iteration("جيشنو^&&", "name=جيشنو^&&"),
+    Iteration("_Jishnu", "name=_Jishnu"),
```
BenHenning: Nit: Please remove the trailing comma for here & other parameterized tests below.

BenHenning (commented):
Thanks @JishnuGoyal! Just had a few remaining comments, otherwise this PR looks really close to being done.

> PTAL @BenHenning is it correct way to write tests? Also, are any additional changes required to make parameterised tests work with Build.Bazel ?

They look correct--did they pass for you when run?

For Bazel, you may need to update the test's dependencies here: https://github.com/oppia/oppia-android/blob/develop/utility/BUILD.bazel#L90. For the needed dependencies, see the referenced parameterized tests for which libraries need to be included. You should be able to verify locally if it works by running the following in a terminal (from the repository root):

```sh
bazel test //utility:src/main/java/org/oppia/android/util/profile/ProfileNameValidatorTest
```

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
JishnuGoyal: done

JishnuGoyal:
I have added Build.bazel file for ProfileNameValidatorTest but I'm facing some problems -  This mentioned test is failing even after adding the bazel files.
>bazel test //utility:src/main/java/org/oppia/android/util/profile/ProfileNameValidatorTest

PTAL @BenHenning . Which all places do the bazel files need to be added?

JishnuGoyal:
> Is this still unresolved @JishnuGoyal?

Yes, haven't heard back from @srushtirk yet.

>Also, I think we should make sure that there are at least a few tests for name validation in ProfileManagementControllerTest, too, such as:

    One valid name with Latin characters
    One valid name with Devanagari characters (which would fail without the new validator)
    One invalid name (with the correct presented)

What do you think?

Sure.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
seanlip: I will take over answering this one, since @srushtirk is away.

Let's go with this error message: "Please choose a profile name that doesn't include numbers or symbols."

(Also, @JishnuGoyal, in the future, if you don't hear back from someone within 2 days and it's blocking you, please escalate. Thanks.)

seanlip:
Just a note -- for the first point, I have responded to @JishnuGoyal's comment.

JishnuGoyal:
> Just a note -- for the first point, I have responded to @JishnuGoyal's comment.

@seanlip , I'm not sure what this message is in reference to, could you please provide some more detail?

seanlip:
It refers to: "Is this still unresolved @JishnuGoyal? / Yes, haven't heard back from @srushtirk yet."

BenHenning:
Reopened since this was closed due to my late review.

utility/src/test/java/org/oppia/android/util/profile/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+        "//third_party:com_google_truth_truth",
+        "//third_party:junit_junit",
+        "//third_party:org_robolectric_robolectric",
+        "//third_party:robolectric_android-all",
+        "//utility/src/main/java/org/oppia/android/util/locale:prod_module",
+        "//utility/src/main/java/org/oppia/android/util/logging:prod_module",
+    ],
+)
+
+dagger_rules()
```
BenHenning: Please add a missing EOF line to this file.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
```diff
@@ -0,0 +1,119 @@
...
+@Config(manifest = Config.NONE)
+class ProfileNameValidatorTest {
+  @Inject
+  lateinit var profileNameValidator: ProfileNameValidator
+
+  @Parameter
+  lateinit var name: String
+
+  @Before
+  fun setup() {
```
BenHenning, line 36: Nit: ``setUp`` for consistency with other test suites.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: @JishnuGoyal reminder to please address this comment.

BenHenning (changes requested):
Apologies for the really late review here.

Thanks @JishnuGoyal. I just had a few nits, plus a comment on an existing thread that hasn't yet been resolved. Other than those comments, I think the PR is very close to 'LGTM'. PTAL.

BenHenning:
@JishnuGoyal this would be value to get in, especially since it's so close to being finished up. Do you have a plan for completing this?

BenHenning:
This is actually on my backlog to investigate, but I haven't been able to make the time for it quite yet.

BenHenning:
I'm now taking a look at this (since it's functionality that we'd like to include in the upcoming beta release).

BenHenning:
Created https://github.com/JishnuGoyal/oppia-android/pull/1 to address the conflicts & broken tests. PTAL @JishnuGoyal.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: add_profile_error_name_only_letters still needs to be updated to have a clearer error.

utility/src/test/java/org/oppia/android/util/profile/BUILD.bazel
BenHenning: This has been resolved.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
BenHenning: Still unresolved.

BenHenning (approved):
Latest changes look good and CI is now passing. There are two comments that still need resolution before this can be fixed, though I think I'm going to go ahead and merge this and then address them in a separate PR (to avoid needing another day of iteration).

BenHenning:
Since this PR is part of broader urgent work for the Beta MR1 release, I'm force-merging this without @rt4914's review. I've self-reviewed it and found no major issues. Furthermore, https://github.com/oppia/oppia-android/issues/4567 is tracking ensuring that this does get reviewed by someone else later after it's been merged.

utility/src/test/java/org/oppia/android/util/profile/ProfileNameValidatorTest.kt
BenHenning: Resolved in #4565.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: Resolved in #4565.

Issue #4064, opened 2021-12-17
Add basic learner analytics support


BenHenning:
PRD: https://docs.google.com/document/d/13H_r47Ge6dj_VAszEHQttoyOoCKXeEgodSTv16-7fnY/edit.

Between #4247 and #4248, remaining work items I've identified:
- [x] Disable the feature by default
- [x] Verify that event uploading is actually happening, and correctly, to Firebase
- [x] Shift the implementation such that the domain changes update modules, but don't actually change the application classes until the integration PR (revised to 4 PRs: domain components, integration, events, UI)
- [x] Generating device ID using a GUID that's persisted on disk (this is a **requirement change** since we've discovered that we can't rely on hardware identifiers)
- [x] Update event logging flow to be asynchronous since the device ID has to be persisted on disk and thus can't be retrieved synchronously (see previous item) -- Update: only the IDs part needs to be asynchronous (which conveniently builds nicely into existing controllers now that using ``StateFlow`` is on the table since I branched the PRs off of math expressions).
- [ ] Finish implementing the remaining events, and make sure they're thoroughly tested
- [x] Session ID needs to be piped using data providers (which might introduce a situation where we actually block on #4239 which would not be ideal)
- [ ] All tests need to be checked to pass on Gradle & Bazel
- [ ] Affected UI tests need to be tested with Espresso via Gradle
- [x] We may potentially need the Bazel test fixes introduced in #4068 since these PRs are affecting so many tests it's causing the compute affected tests workflow to fail
- [x] Any other comments I've listed in #4247 and #4248
- [x] Figure out the chaining & merge strategy for the project since changes are sort of messily split between the two PRs (maybe partly because of earlier merging challenges). It'd be nice to keep PR reviews as simple as possible, but the priority needs to be execution speed.
- [x] Since new branches are going to be created per the merging strategy, ensure all the changes from the early project are fully covered within those branches

Statuses on all of the events that need to be added:
- [x] Card Start: not started as of the opening of this issue
- [x] Card End: not started as of the opening of this issue
- [x] Hint Offered: not started as of the opening of this issue
- [x] Hint Accessed: not started as of the opening of this issue
- [x] Solution accessed: not started as of the opening of this issue
- [x] Solution offered: not started as of the opening of this issue
- [x] Submit answer: not started as of the opening of this issue
- [x] Play voice over: not started as of the opening of this issue
- [x] Background app inactivity: not started as of the opening of this issue
[...]

PR #4116, opened 2022-01-20
Fix part of #4064: Learner Analytics proto structures

Sarthak2601:
## Explanation

Fixes part of #4064 
This PR modifies the existing proto structure in `oppia_logger` to accommodate new learner study analytics events. There are 15 new analytics events that are being added as of now. The data related to these events is divided into three parts: Generic Data (common across all events), Exploration Data (common across all exploration-related events) and unique parameters. This PR adds proto definitions for event-specific contexts and actions along with definitions for Generic and Exploration Data. 

For reference: 
<google-sheets-html-origin style="font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0);">

Event Name | Unique Dimensions for each event | Event Category(decides coverage)
-- | -- | --
START_CARD | Skill ID (card_skill_id) | Exploration
END_CARD | Skill ID (card_skill_id) | Exploration
HINT_OFFERED | Hint index (hint_index) | Exploration
ACCESS_HINT | Hint index (hint_index) | Exploration
ACCESS_SOLUTION |   | Exploration
SOLUTION_OFFERED |   | Exploration
SUBMIT_ANSWER | User submitted answer (user_answer)Answer Label (answer_label) | Exploration
PLAY_VOICE_OVER | Content ID (content_id) | Exploration
APP_IN_BACKGROUND |   | Generic
APP_IN_FOREGROUND |   | Generic
EXIT_EXPLORATION |   | Exploration
FINISH_EXPLORATION |   | Exploration
RESUME_EXPLORATION | *Click continue should trigger a new session (new session ID) | Generic
START_OVER_EXPLORATION | *Click start over should trigger a new session (new session ID) | Generic
DELETE_PROFILE |   | Generic

</google-sheets-html-origin>

## Essential Checklist

## For UI-specific PRs only


BenHenning:
Thanks @Sarthak2601. I'll need to follow up with a review on this tomorrow.

model/src/main/proto/oppia_logger.proto
```diff
@@ -31,6 +31,36 @@ message EventLog {
...
+      AppInBackgroundContext app_in_background_context = 15;
+      // The event being logged is related to foregrounding of the application.
+      AppInForegroundContext app_in_foreground_context = 16;
+      // The event being logged is related to exiting an exploration.
+      ExitExplorationContext exit_exploration_context = 17;
+      // The event being logged is related to finishing an exploration.
+      FinishExplorationContext finish_exploration_context = 18;
+      // The event being logged is related to resuming an exploration.
+      ResumeExplorationContext resume_exploration_context = 19;
+      // The event being logged is related to starting over an exploration.
```
BenHenning: Perhaps instead: "to restarting an exploration"? The current wording seems a bit off.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
     int32 sub_topic_id = 2;
   }
 
+  // Structure of a start card context.
+  message StartCardContext {
+    GenericData generic_data = 1;
+    ExplorationData exploration_data= 2;
```
BenHenning: Nit: please make sure there's a space before the ``=``. Ditto elsewhere.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
     int32 sub_topic_id = 2;
   }
 
+  // Structure of a start card context.
```
BenHenning: Perhaps something like: ``Represents the event context for opening a card within an exploration.`` would be clearer context. Ditto for the other new context messages.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+  message StartCardContext {
+    GenericData generic_data = 1;
```
BenHenning: This could benefit from a line comment, even if it's just: ``Defined attributes that are common among other event log contexts.``

Ditto elsewhere (please make sure all fields have comments).

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+    ExplorationData exploration_data= 2;
+    // The linked skill id of the exploration.
+    string skill_id = 3;
+  }
+
+  // Structure of a hint offered context.
+  message HintOfferedContext {
+    GenericData generic_data = 1;
+    ExplorationData exploration_data= 2;
+    // index of the last revealed hint within the hint list of a state (meaning the index
```
BenHenning: Couple of nits here (& ditto for all other relevant places):
- Please make sure that these comments start with a capital letter and end with punctuation
- Please make sure comments fully reach the 100 character line before being wrapped
- Please add blank newlines between fields to help make the documentation comments a bit easier to read.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
     int32 sub_topic_id = 2;
   }
 
+  // Structure of a start card context.
+  message StartCardContext {
+    GenericData generic_data = 1;
+    ExplorationData exploration_data= 2;
+    // The linked skill id of the exploration.
```
BenHenning: More context could probably be added here. What does a linked skill ID mean in the context of a card?

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+    string skill_id = 3;
+  }
+
+  // Structure of a hint offered context.
+  message HintOfferedContext {
+    GenericData generic_data = 1;
+    ExplorationData exploration_data= 2;
+    // index of the last revealed hint within the hint list of a state (meaning the index
+    // corresponding to this value is the latest hint that's been revealed by the user)
+    string hint_index = 3;
```
BenHenning: Suggest also noting in the documentation what the value of this field is when it's a solution offered. Ditto below for accessed.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+  }
+
+  // Structure of an access hint context.
+  message AccessHintContext {
```
BenHenning: Out of curiosity, could we maybe reuse common context structures? E.g. have just a 'HintContext' and differentiate offered vs. accessed based on the oneof above? That might help reduce the number of new contexts actually needed here since it seems really repetitive as-is.

If we collapse these, please make sure that the oneof lines above have clearer documentation per my comment above regarding the top-level message doc comments.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+    ExplorationData exploration_data= 2;
+    // Whether the answer that the learner submitted is the correct answer.
+    bool is_answer_correct = 3;
+  }
+
+  // Structure of a play voiceover context.
+  message PlayVoiceOverContext {
+    GenericData generic_data = 1;
+    ExplorationData exploration_data= 2;
+    // The content id of the voiceover recording.
```
BenHenning:
```suggestion
    // The content id of the voiceover recording being played.
```

This comes across as an incomplete thought.

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,6 +110,105 @@ message EventLog {
...
+    GenericData generic_data = 1;
+  }
+
+  // Structure of a delete profile context.
+  message DeleteProfileContext {
+    GenericData generic_data = 1;
+  }
+
   // Supported event actions for event logging.
   enum EventAction {
```
BenHenning: I'm not necessarily suggesting we change a lot in the analytics controller, but why can't we just use the generated oneof index for these instead of having a separate enum if the two lists are being kept in parity?

model/src/main/proto/oppia_logger.proto
```diff
@@ -118,6 +277,26 @@ message EventLog {
     // The priority of events whose logs can be removed from the storage if the size exceeds a certain limit
     OPTIONAL = 2;
   }
+
+  // Structure of the generic data.
+  message GenericData {
```
BenHenning: This name is a bit too generic. :) Perhaps instead: LearnerDetails might be clearer?

model/src/main/proto/oppia_logger.proto
```diff
@@ -118,6 +277,26 @@ message EventLog {
...
+    // The learner id of the current user. This id is profile-specific.
+    string learner_id = 1;
+    // The device id of the current device. This id is common across all profiles.
+    string device_id = 2;
+  }
+
+  // Structure of the exploration data.
+  message ExplorationData {
```
BenHenning: Is suffixing these with 'Data' conventional? It might be clearer just to call this 'ExplorationDetails' or 'ExplorationContext' (though that might conflate with the event contexts above).

I am rethinking some of these names a bit since earlier naming choice deficiencies are becoming more noticeable now that a lot of new events are being added.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
```diff
@@ -36,6 +36,8 @@ class EventBundleCreator {
         CONCEPT_CARD_CONTEXT -> createConceptCardContextBundle(eventLog)
         REVISION_CARD_CONTEXT -> createRevisionCardContextBundle(eventLog)
         ACTIVITYCONTEXT_NOT_SET -> createNoContextBundle(eventLog)
+        else -> createNoContextBundle(eventLog)
+        // TODO(#4064): create bundle creator functions for new events and replace this with them.
```
BenHenning:
```suggestion
        // TODO(#4064): Create bundle creator functions for new events and replace this with them.
```

Maybe move this above the 'else' since it's not 100% clear what "this" is referring to (generally comments describe the line they're on or the following line, not the preceding line).

BenHenning (changes requested):
Thanks @Sarthak2601! Took a first pass. I had a bunch of thoughts around the documentation context--PTAL.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Collapsed this context to support both card start and end related events. Modified documentation accordingly, PTAL.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Nice catch! I've collapsed all repetitive contexts and used the oneof to differentiate. Regarding documentation, I found it to be okay -- in terms of explaining what the field depicts. However, if there is some specific type of info that you want me to put, please let me know.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Agreed, done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: The idea behind keeping both of them was to indicate what we intend to do (action) and what we intend to collect (context data) in a mutually exclusive manner. We might come across a situation where we collect similar data across different event actions. For example: event actions like `OPEN_INFO_TAB`, `OPEN_LESSONS_TAB`, `OPEN_PRACTICE_TAB` and `OPEN_REVISION_TAB`, correspond to a single context - `TopicContext`. Apart from this, the values of this enum (EventAction) are also the event names that we send to Firebase. Does that help?

model/src/main/proto/oppia_logger.proto
Sarthak2601: Changed this to `LearnerDetailsContext`.

model/src/main/proto/oppia_logger.proto
Sarthak2601: No, it's not conventional. I used because it was mentioned in the TDD. Changed to `ExplorationDetailsContext`.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
```diff
@@ -110,159 +130,156 @@ message EventLog {
...
+    // Defined attributes that are common among other exploration related event log contexts.
+    ExplorationDetailsContext exploration_details = 1;
+
+    // Corresponds to the index of a hint or solution in a state. Help Index is used to determine which
+    // hints can be shown, have been seen, whether the solution can be shown and whether the solution has
+    // been seen. For events tracking:
+    // - offering a hint, this index should be of type = next_available_hint_index.
+    // - accessing a hint, this index should be of type = latest_revealed_hint_index.
+    // - offering a solution, this index should be of type = show_solution.
+    // - accessing a solution, this index should be of type = everything_revealed.
```
BenHenning: Ah, why not just use HelpIndex then if these are the values?

model/src/main/proto/oppia_logger.proto
```diff
@@ -110,159 +130,156 @@ message EventLog {
...
-  message StartOverExplorationContext {
-    GenericData generic_data = 1;
-  }
+    // The name of the current state of the exploration.
+    string state_name = 4;
 
-  // Structure of a delete profile context.
-  message DeleteProfileContext {
-    GenericData generic_data = 1;
+    // Defined attributes that are common among all event log contexts.
```
BenHenning: Not all--please be more specific here (since we already log events that don't include these details).

model/src/main/proto/oppia_logger.proto
```diff
@@ -21,46 +21,66 @@ message EventLog {
...
-      HintOfferedContext hint_offered_context = 9;
+      HintContext hint_offered_context = 9;
+
       // The event being logged is related to accessing a hint.
-      AccessHintContext access_hint_context = 10;
+      HintContext access_hint_context = 10;
+
       // The event being logged is related to offering a solution when it becomes available.
-      SolutionOfferedContext solution_offered_context = 11;
+      ExplorationDetailsContext solution_offered_context = 11;
```
BenHenning: Hmm this is really close in name to ExplorationContext. Is there any way to consolidate the two?

model/src/main/proto/oppia_logger.proto
BenHenning: Much clearer--thanks!

model/src/main/proto/oppia_logger.proto
```diff
@@ -80,32 +130,158 @@ message EventLog {
     int32 sub_topic_id = 2;
   }
 
+  // Represents the event context for interacting with a card within an exploration.
+  message CardContext {
+    // Defined attributes that are common among other exploration related event log contexts.
+    ExplorationDetailsContext exploration_details = 1;
+
+    // The ID of skill to which this exploration is associated.
```
BenHenning: Do you mean 'exploration card'?

model/src/main/proto/oppia_logger.proto
BenHenning: I find the new code/docs much clearer--thanks!

model/src/main/proto/oppia_logger.proto
BenHenning: I don't think that quite addresses why a oneof wouldn't work. In your example, each of those actions would presumably be different oneof cases with the same type (i.e. distinct events with the same logged details), e.g.:

```kotlin
TopicContext open_info_tab = 1;
TopicContext open_lessons_tab = 2;
TopicContext open_practice_tab = 3;
TopicContext open_revision_tab = 4;
...
```

Or, am I missing something?

BenHenning (commented):
Thanks @Sarthak2601. This is looking really good--took a follow-up pass. PTAL.

model/src/main/proto/oppia_logger.proto
Sarthak2601: done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: clubbed into one, we can handle packing and unpacking of data from controller and utility separately.

model/src/main/proto/oppia_logger.proto
Sarthak2601: done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Umm, I think these are values types and not the actual index values. We'll use these types to get the correct hint index for logging. Am I missing something?

model/src/main/proto/oppia_logger.proto
Sarthak2601: Removed event action and modified dependencies to depend directly on the context. One thing while implementing that I was kind of confused about was handling of cases where we don't log any context -- so I've created a null message for these type of cases.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/EventLogItemViewModel.kt
```diff
@@ -24,10 +24,7 @@ class EventLogItemViewModel(
 
   /** Returns the event log context in a human readable format. */
   fun formatContextString(): String =
-    eventLog.context.activityContextCase.name.capitalizeWords().substringBeforeLast(" ")
-
-  /** Returns the event log action name in a human readable format. */
-  fun formatActionNameString(): String = eventLog.actionName.name.capitalizeWords()
+    eventLog.context.activityContextCase.name.capitalizeWords()
```
BenHenning: Will this fit on one line with the fun declaration?

Also, does this have the correct behavior? I'd expect the enum names to be in UPPER_SNAKE_CASE so presumably capitalizeWords() won't have any effect.

app/src/main/res/layout/view_event_logs_event_log_item_view.xml
```diff
@@ -64,30 +64,13 @@
...
       android:text="@{viewModel.formatContextString()}"
       android:textColor="@color/oppia_primary_text_dark"
-      android:visibility="@{viewModel.eventLog.context.activityContextCase != EventLog.Context.ActivityContextCase.ACTIVITYCONTEXT_NOT_SET ? View.VISIBLE : View.GONE}"
-      app:layout_constraintBottom_toTopOf="@id/view_event_logs_action_name_text_view"
       app:layout_constraintEnd_toEndOf="parent"
       app:layout_constraintStart_toStartOf="parent"
       app:layout_constraintTop_toBottomOf="@id/event_log_time_and_priority_layout" />
 
-    <TextView
-      android:id="@+id/view_event_logs_action_name_text_view"
```
BenHenning, line 77: Is it maybe worth including the context string that's computed here so that it's easily referenceable when browsing Firebase logs?

app/src/sharedTest/java/org/oppia/android/app/devoptions/ViewEventLogsFragmentTest.kt
```diff
@@ -170,116 +169,6 @@ class ViewEventLogsFragmentTest {
     }
   }
 
-  @Test
-  fun testViewEventLogsFragment_checkRecentLogsAreDisplayedFirst() {
```
BenHenning: Should this & the next test instead be verifying a different view is first?

app/src/sharedTest/java/org/oppia/android/app/devoptions/ViewEventLogsFragmentTest.kt
```diff
@@ -509,32 +398,27 @@ class ViewEventLogsFragmentTest {
...
-      eventAction = EventAction.OPEN_HOME,
-      eventContext = null
+      eventContext = oppiaLogger.createOpenHomeContext()
     )
 
     oppiaLogger.logTransitionEvent(
       timestamp = TEST_TIMESTAMP + 20000,
-      eventAction = EventAction.OPEN_LESSONS_TAB,
-      eventContext = oppiaLogger.createTopicContext(TEST_TOPIC_ID)
+      eventContext = oppiaLogger.createOpenInfoTabContext(TEST_TOPIC_ID)
```
BenHenning: Should this be the lessons tab, instead (for parity with the previous implementation)?

domain/src/main/java/org/oppia/android/domain/oppialogger/OppiaLogger.kt
```diff
@@ -14,19 +13,17 @@ class OppiaLogger @Inject constructor(
   /** Logs transition events. See [AnalyticsController.logTransitionEvent] for more context. */
   fun logTransitionEvent(
     timestamp: Long,
-    eventAction: EventAction,
     eventContext: EventLog.Context?
```
BenHenning: Since you removed the tests checking for nullness, should a non-null context now be required here & elsewhere?

domain/src/main/java/org/oppia/android/domain/oppialogger/OppiaLogger.kt
```diff
@@ -126,26 +136,65 @@ class OppiaLogger @Inject constructor(
       .build()
   }
 
-  /** Returns the context of an event related to topic. */
-  fun createTopicContext(
+  /** Returns the context of an event related to opening of practice tab. */
```
BenHenning:
```suggestion
  /** Returns the context of the event indicating that the user opened the practice tab. */
```

Ditto elsewhere (KDoc phrasing).

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsController.kt
```diff
@@ -35,13 +34,11 @@ class AnalyticsController @Inject constructor(
    */
   fun logTransitionEvent(
     timestamp: Long,
-    eventAction: EventAction,
     eventContext: EventLog.Context?
```
BenHenning: Ditto here--can this now be non-null?

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
```diff
@@ -173,80 +173,128 @@ class OppiaLoggerTest {
...
+  fun testController_createOpenExplorationActivityContext_returnsCorrectExplorationContext() {
+    val eventContext = oppiaLogger.createOpenExplorationActivityContext(
       TEST_TOPIC_ID,
       TEST_STORY_ID,
       TEST_EXPLORATION_ID
     )
 
     assertThat(eventContext.activityContextCase).isEqualTo(
-      EventLog.Context.ActivityContextCase.EXPLORATION_CONTEXT
+      EventLog.Context.ActivityContextCase.OPEN_EXPLORATION_ACTIVITY
```
BenHenning: Optional nit: I suggest directly importing this & other cases (which can be done via the 'ctrl + enter' shortcut when your text cursor is on the enum name on this line) so that the isEqualTo can fit on one line--this can help with test readability.

model/src/main/proto/oppia_logger.proto
```diff
@@ -9,90 +9,97 @@ option java_multiple_files = true;
...
-  message ExplorationContext {
-    // The active topic ID when the event is logged.
-    string topic_id = 1;
-    // The active story ID when the event is logged.
-    string story_id = 2;
-    // The active exploration ID when the event is logged.
-    string exploration_id = 3;
-  }
+  // Representation of a null context which can be used to represent events which don't require
+  // any additional data to be recorded.
```
BenHenning: Alternatively, you could use a boolean and document that the value doesn't matter, only its presence (it's a bit of a hack, but we've used it elsewhere and I think I got it from a recommendation outside our team).

I generally use the value 'true' since it makes the code nicer ('true' helps communicate that the specific mode is "enabled").

utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseEventLogger.kt
```diff
@@ -27,7 +27,7 @@ class FirebaseEventLogger(
   /** Logs an event to Firebase Analytics with [NETWORK_USER_PROPERTY] and [COUNTRY_USER_PROPERTY]. */
   override fun logEvent(eventLog: EventLog) {
     bundle = eventBundleCreator.createEventBundle(eventLog)
-    firebaseAnalytics.logEvent(eventLog.actionName.toString(), bundle)
+    firebaseAnalytics.logEvent(eventLog.context.activityContextCase.toString(), bundle)
```
BenHenning: Should this be using the enum's ``name`` property for portability?

model/src/main/proto/oppia_logger.proto
BenHenning: Left a follow up comment separately for the missing context case, otherwise I think the changes look really good.

model/src/main/proto/oppia_logger.proto
BenHenning: I'm not sure that I follow--what's an example of what this value will be populated with? I'm not sure that I understand what a "value type" means in this context.

BenHenning (commented):
Thanks @Sarthak2601. Took another pass--PTAL. I think the most recent changes look **much** cleaner than the previous implementation--thanks for taking the time to clean up existing use cases!

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/EventLogItemViewModel.kt
Sarthak2601: The expected behaviour is `Upper Snake Case` from `UPPER_SNAKE_CASE`, right? If yes, then capitalizeWords() does work.

app/src/main/res/layout/view_event_logs_event_log_item_view.xml
Sarthak2601: So maybe have two different versions of the same context string? Currently we have the context string (`UPPER_SNAKE_CASE`) formatted to `Upper Snake Case`.

Screenshot_1644143676

What difference are you suggesting here?

app/src/sharedTest/java/org/oppia/android/app/devoptions/ViewEventLogsFragmentTest.kt
Sarthak2601: Removed these two by mistake, restored now -- thanks!

app/src/sharedTest/java/org/oppia/android/app/devoptions/ViewEventLogsFragmentTest.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/OppiaLogger.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/OppiaLogger.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsController.kt
Sarthak2601: done.

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
Sarthak2601: done!

model/src/main/proto/oppia_logger.proto
Sarthak2601: nice one, done -- thanks!

utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseEventLogger.kt
Sarthak2601: done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: By value type I meant the `IndexTypeCase` of HelpIndex, and by value I meant the integer value of `nextAvailableHintIndex`. Does that help?

BenHenning: HelpIndex provides the complete context: all hints that have been shown, and whether another is available to show (and similar for the solution since it's "built into" the structure). I thought we actually do want to know the index as well as the state which is why I thought HelpIndex would be an easier plug-and-play solution.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/EventLogItemViewModel.kt
BenHenning: Ah interesting. I didn't know it extract words from snake case.

app/src/main/res/layout/view_event_logs_event_log_item_view.xml
BenHenning: Ah, I don't think I fully understood that you were effectively turning two strings into one before. This change makes more sense to me now, and I think the UI per your image looks good.

BenHenning (commented):
Thanks @Sarthak2601. Just had a few remaining comments--PTAL.

model/src/main/proto/oppia_logger.proto
```diff
@@ -9,39 +9,94 @@ option java_multiple_files = true;
...
   // The timestamp of the event.
   int64 timestamp = 1;
-  // The action happening in the event.
-  EventAction action_name = 2;
   // The priority of the event.
-  Priority priority = 3;
+  Priority priority = 2;
   // The activity context of the event.
-  Context context = 4;
+  Context context = 3;
```
vinitamurthi: Lets not delete proto fields like this, instead you can reserve field number 2 and keep priority and context fields asi s 
i.e. 
```
int64 timestamp = 1;
reserved 2;
Priority priority = 3;
Context context = 4;
```
 (or you can keep the reserved keyword at the top of the message before all the other params)

See https://developers.google.com/protocol-buffers/docs/proto3#updating for details about how to update a proto message!

model/src/main/proto/oppia_logger.proto
```diff
@@ -9,39 +9,94 @@ option java_multiple_files = true;
...
-      // The event being logged is related to the topic activity.
-      TopicContext topic_context = 3;
-      // The event being logged is related to the story activity.
-      StoryContext story_context = 4;
-      // The event being logged is related to the concept card.
-      ConceptCardContext concept_card_context = 5;
-      // The event being logged is related to the revision card.
-      RevisionCardContext revision_card_context = 6;
-    }
-  }
```
vinitamurthi, line 35: Same comment here, you need to reserve fields 1 - 6, and set open_exploration_activity = 7;

model/src/main/proto/oppia_logger.proto
```diff
@@ -9,39 +9,94 @@ option java_multiple_files = true;
...
+      ExplorationContext open_exploration_activity = 1;
 
-  // Structure of an exploration context.
-  message ExplorationContext {
-    // The active topic ID when the event is logged.
-    string topic_id = 1;
-    // The active story ID when the event is logged.
-    string story_id = 2;
-    // The active exploration ID when the event is logged.
-    string exploration_id = 3;
```
vinitamurthi, line 44: Ditto here

model/src/main/proto/oppia_logger.proto
Sarthak2601: Done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: Insightful! done -- thanks!

model/src/main/proto/oppia_logger.proto
Sarthak2601: done.

model/src/main/proto/oppia_logger.proto
Sarthak2601: In this case, I've kept these three fields as is and added other fields from number 4, does that work?

BenHenning (approved):
Thanks @Sarthak2601. The latest changes LGTM!

BenHenning:
Though FYI it appears there are broken CI checks--please fix and if there are any major changes that need re-review please readd me.

rt4914 (approved):
LGTM, thanks.

Sarthak2601:
@vinitamurthi PTAL :)

vinitamurthi:
Was this closed accidentally?

vinitamurthi:
I have reopened this PR , there are some conflicts but otherwise LGTM!

vinitamurthi (approved):

Issue #3245, opened 2021-05-25
[A11y] Enable AccessibilityChecks[blocked on  #4991]


rt4914:
Enable AccessibilityChecks for espresso by removing `DisableAccessibilityChecks` annotation from all Espresso tests.

Reference: https://github.com/oppia/oppia-android/pull/3087 & #4133 

This PR should be done in multiple parts, ideally pick 3-5 files at a time.

Please use the linked PRs as a guide on how to solve the issue and how to structure your PR.

Make sure you add screenshot of passing espresso tests in your PR description once `DisableAccessibilityChecks` is removed.

Rohit1173:
I want to work on this issue

PR #4133, opened 2022-01-25
Fix #3362 Use OppiaTestRule in all Espresso Tests

### LINKED ISSUE #3362: Use `OppiaTestRule ` in all Espresso Tests
Use `OppiaTestRule` in all Espresso Tests

Basically every Espresso test file by default should inject this rule.

vrajdesai78:
## Explanation

Fixes #3362: Added OppiaTestRule in all Espresso Tests

## Essential Checklist

### DIFF
--- a/app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
@@ -73,6 +73,7 @@ import org.oppia.android.domain.platformparameter.PlatformParameterSingletonModu
 import org.oppia.android.domain.question.QuestionModule
 import org.oppia.android.domain.topic.PrimeTopicAssetsControllerModule
 import org.oppia.android.domain.workmanager.WorkManagerConfigurationModule
+import org.oppia.android.testing.DisableAccessibilityChecks
 import org.oppia.android.testing.OppiaTestRule
 import org.oppia.android.testing.TestLogReportingModule
 import org.oppia.android.testing.espresso.EditTextInputAction
@@ -100,7 +101,6 @@ import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
 import javax.inject.Inject
 import javax.inject.Singleton
-import org.oppia.android.testing.DisableAccessibilityChecks
 
 @RunWith(AndroidJUnit4::class)
 @LooperMode(LooperMode.Mode.PAUSED)
--- a/app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
@@ -100,6 +100,7 @@ import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
 import javax.inject.Inject
 import javax.inject.Singleton
+import org.oppia.android.testing.DisableAccessibilityChecks
 
 @RunWith(AndroidJUnit4::class)
 @LooperMode(LooperMode.Mode.PAUSED)
@@ -209,6 +210,7 @@ class PinPasswordActivityTest {
   }
 
   @Test
+  @DisableAccessibilityChecks
   fun testPinPassword_withAdmin_inputCorrectPin_opensHomeActivity() {
     ActivityScenario.launch<PinPasswordActivity>(
       PinPasswordActivity.createPinPasswordActivityIntent(
@@ -226,6 +228,7 @@ class PinPasswordActivityTest {
   }
 
   @Test
+  @DisableAccessibilityChecks
   fun testPinPassword_withUser_inputCorrectPin_opensHomeActivity() {
     ActivityScenario.launch<PinPasswordActivity>(
       PinPasswordActivity.createPinPasswordActivityIntent(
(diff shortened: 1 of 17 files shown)


vrajdesai78:
@rt4914 PTAL. Thanks

rt4914 (changes requested):
@vrajdesai78 Actually the usage of this file can fail some test cases but those test cases will be Espresso based and CI checks only Robolectric. 

So you will need to run all these files on Espresso tests and post screenshot of the result, if all the tests pass that's great. If something fails because of `accessibility` that means you will need to use `DisableAccessibilityChecks` for that particular test case.

vrajdesai78:
> @vrajdesai78 Actually the usage of this file can fail some test cases but those test cases will be Espresso based and CI checks only Robolectric. 
> 
> So you will need to run all these files on Espresso tests and post screenshot of the result, if all the tests pass that's great. If something fails because of `accessibility` that means you will need to use `DisableAccessibilityChecks` for that particular test case.
> 
> 

Ok I will look into it. Thanks

vrajdesai78:
I have run almost all test cases in which I have added OppiaTestRule for this issue. I have tried to run test cases from `ImageViewBindingAdaptersTest` but it's taking lots of time (more than 50 mins). Kindly find the screenshots of passed test cases below. 

margin_bind
Screenshot from 2022-01-28 19-12-28
Screenshot from 2022-01-28 19-14-25
Screenshot from 2022-01-28 19-16-05
This one test case is failing even without adding OppiaTestRule
Screenshot from 2022-01-28 20-08-42
Screenshot from 2022-01-28 20-12-11
Screenshot from 2022-01-28 20-13-20
Screenshot from 2022-01-28 20-33-28
Screenshot from 2022-01-28 20-34-27
Screenshot from 2022-01-28 20-35-19
Screenshot from 2022-01-28 20-36-33
Screenshot from 2022-01-28 22-25-24
Screenshot from 2022-01-28 22-33-27
Screenshot from 2022-01-28 22-34-11
stateAssembler
Screenshot from 2022-01-28 22-30-36

@rt4914 PTAL. Thanks

rt4914 (approved):
LGTM, thanks. Just file the two issue mentioned above.

vrajdesai78:
> LGTM, thanks. Just file the two issue mentioned above.

Thanks, Rajat. I have filed two issues #4148 and #4149 as we have discussed.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -207,6 +210,7 @@ class PinPasswordActivityTest {
   }
 
   @Test
+  @DisableAccessibilityChecks
```
BenHenning, line 213: For this & the other checks, can you provide some context as to why these need to be exempted?

BenHenning (changes requested):
Thanks @vrajdesai78. I had one file comment, but I'm also curious: is this actually address all #3362 TODOs? I saw one in StoryActivityTest that's corresponding to reenabling a disabled check rather than just adding the rule (so I'm actually wondering if this is fully addressing #3362 since it's actually disabling more tests for accessibility checks).

@rt4914 any thoughts here?

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
vrajdesai78: @BenHenning First I have run all expresso test without DisableAccessibilityChecks and after seeing the error (which is related to accessibility) I have added DisableAccessibilityChecks annotation

vrajdesai78: > @BenHenning  First I have run all expresso test without DisableAccessibilityChecks and after seeing the error (which is related to accessibility) I have added DisableAccessibilityChecks annotation.

vrajdesai78:
> @BenHenning First I have run all expresso test without DisableAccessibilityChecks and after seeing the error (which is related to accessibility) I have added DisableAccessibilityChecks annotation

@BenHenning PTAL. Thanks

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
BenHenning: This turns off the checks for accessibility, though, which we generally only ever want to do on a short-term and conditional basis (we shouldn't be broadly disabling accessibility checks without a clear need, and we should actively be trying to remove those disables since it means we aren't verifying accessibility correctness on those screens). It's not clear to me why these are correct to disable--it actually seems like we may want to wait to add OppiaTestRule to this & other files until we can fix the accessibility test issues in for those screens.

anandwana001 (approved):
My code owner file LGTM

rt4914:
> Thanks @vrajdesai78. I had one file comment, but I'm also curious: is this actually address all #3362 TODOs? I saw one in StoryActivityTest that's corresponding to reenabling a disabled check rather than just adding the rule (so I'm actually wondering if this is fully addressing #3362 since it's actually disabling more tests for accessibility checks).
> 
> @rt4914 any thoughts here?

@vrajdesai78 @BenHenning yes this is correct. We should first use `OppiaTestRule` in all `Espresso` tests and then we should focus on making sure that we can remove `DisableAccessibilityChecks` from all test cases. This is getting tracked here. https://github.com/oppia/oppia-android/issues/3245

BenHenning (approved):
Got it, thanks for clarifying @rt4914. In that case, the PR LGTM.

BenHenning:
Merging since it seems everything's been resolved.

PR #4136, opened 2022-01-27
Fix #4109 : Disable LDPI for oppia app

### LINKED ISSUE #4109: Disable LDPI for oppia app
LDPI is generally used by watches and we should be targeting MDPI and above only.

Disable LDPI for oppia android app.

For the solution, see: https://github.com/oppia/oppia-android/issues/4109#issuecomment-1100236130.

deepanshu731:
Fixes #4109 Disabled LDPI for oppia app as it was not needed.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,6 +6,32 @@
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
   <uses-permission android:name="android.permission.INTERNET" />
   <uses-permission android:name="android.permission.WAKE_LOCK" />
+  <compatible-screens>
+    <!-- all small size screens - except ldpi -->
+    <screen android:screenSize="small" android:screenDensity="mdpi" />
+    <screen android:screenSize="small" android:screenDensity="hdpi" />
+    <screen android:screenSize="small" android:screenDensity="xhdpi" />
+    <screen android:screenSize="small" android:screenDensity="xxhdpi" />
+    <screen android:screenSize="small" android:screenDensity="xxxhdpi" />
+    <!-- all normal size screens - except ldpi -->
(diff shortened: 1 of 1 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,32 @@
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
   <uses-permission android:name="android.permission.INTERNET" />
   <uses-permission android:name="android.permission.WAKE_LOCK" />
+  <compatible-screens>
```
BenHenning, line 9: Can you please point to the reference Android documentation for this as a comment? It'd be helpful for both the review & future reviewing.

I'm also thinking that we might need tests for this (to guard against unexpected screen sizes being added in the future), but it'll depend a bit on the documented Android behavior.

BenHenning (changes requested):
Thanks @deepanshu731. I took a first pass but I think that I need a bit more information before I can fully consider the changes--please take a look at my comment.

app/src/main/AndroidManifest.xml
deepanshu731: @BenHenning  https://developer.android.com/guide/topics/manifest/compatible-screens-element , this is the reference i used.

BenHenning:
Hmm. The warning on the dev article worries me. I also didn't see any way to configure this via AABs; it seems only controllable via the manifest. I'll need to think on this more before we proceed.

deepanshu731:
Any updates @BenHenning ?

BenHenning:
Apologies for the wait @deepanshu731. I'll follow up on this next week.

BenHenning:
I do have some thoughts here--I'm following up on #4109.

BenHenning:
The discussion is still ongoing in #4109; this isn't stale.

Issue #3258, opened 2021-05-27
Find a maintained replacement for CircularImageView


BenHenning:
The codebase currently depends on 2 different CircularImageView dependencies. One of them, https://github.com/sparrow007/CircularImageview, is no longer maintained and should be replaced with a dependency that's kept up-to-date. This will allow us to remove our custom fork of the library (which was needed per #2923 to remove the codebase's dependency on jcenter).

anandwana001:
Rajat and I discussed about using ShapeableImageView from material design library. 
https://developer.android.com/reference/com/google/android/material/imageview/ShapeableImageView
Need to test if it goes similar with our mocks.

BenHenning:
SGTM if that works.

AkinchanKushwaha:
Hi, Please check out https://github.com/hdodenhof/CircleImageView. It has 13.9k stars on GitHub.

anandwana001:
@AkinchanKushwaha Currently, we are using this library only, but we are looking forward to move to `ShapeableImageView `.

aadityaguptaa:
Can I work on this issue?

anandwana001:
Hi @aadityaguptaa I had assigned you this issue. 
In order to complete this issue please check the above comments and we can go for the best suitable solution.

aadityaguptaa:
@anandwana001 thank you, I am working on this, will give you updates in a couple of days

aadityaguptaa:
> @anandwana001 I am trying to change CircularImageView to ShapableImageView, but I am getting the error "This Feature requires ASM7" Any idea what's causing this error??
> 
> [![Screenshot-from-2021-10-20-19-28-15](https://camo.githubusercontent.com/6b59705f5ab688d36ba311a60e7c32f710be0ae889dfa4cdd8cffc10cc80fc2e/68747470733a2f2f692e6962622e636f2f594e42744d77622f53637265656e73686f742d66726f6d2d323032312d31302d32302d31392d32382d31352e706e67)](https://ibb.co/TwMHSdr)

okay nvm, I just had to rebuild the project

rt4914:
@aadityaguptaa Unassigning you from this issue because of inactivity, please re-assign yourself if you are currently working on this.

bhaktideshmukh:
Hey @rt4914 I would like to work upon this issue.Can you please assign it to me?

bkaur-bkj:
@bhaktideshmukh done

dmdbilal:
[...]

PR #4139, opened 2022-01-28
Fix part of #3258 : Replacing CircularImageView with ShapeableImageView

bhaktideshmukh:
## Explanation

 Fix part of #3258 : Replacing CircularImageView with ShapeableImageView.

## Screenshots

**Phone portrait mode**

| Before  | After|
|-------|---|
|||

**Phone landscape mode**

| Before  | After|
|-------|---|
|||

**Tablet portrait mode**

| Before  | After|
|-------|---|
|||

**Tablet landscape mode**

| Before  | After|
|-------|---|
|||

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #4155, opened 2022-02-04
Fix part of #3258 : Replacing CircularImageView with ShapeableImageView

bhaktideshmukh:
## Explanation

 Fix part of #3258 : Replacing CircularImageView with ShapeableImageView.

## Screenshots

**Phone portrait mode**

| Before  | After|
|-------|---|
|||

**Phone landscape mode**

| Before  | After|
|-------|---|
|||

**Tablet portrait mode**

| Before  | After|
|-------|---|
|||

**Tablet landscape mode**

| Before  | After|
|-------|---|
|||

**MDPI device and API 21** </br></br>

**Phone portrait mode**

| Before  | After|
|-------|---|
|||

**Phone landscape mode**

| Before  | After|
|-------|---|
|||

## Essential Checklist

## For UI-specific PRs only


bhaktideshmukh:
PTAL @BenHenning @rt4914 
Thanks.

app/src/main/res/values/styles.xml
```diff
@@ -459,4 +459,9 @@
     <item name="android:gravity">center_horizontal</item>
   </style>
 
+  <!-- ShapeableImageView -->
+  <style name="ShapeAppearanceOverlay.App.Circular" parent="">
```
BenHenning: I'm not sure on this name--aren't all of the uses of the view supposed to be circular? How does choosing this style vs. not change the view?

app/src/main/res/values/styles.xml
```diff
@@ -459,4 +459,9 @@
     <item name="android:gravity">center_horizontal</item>
   </style>
 
+  <!-- ShapeableImageView -->
```
BenHenning, line 468: Adding this comment as a placeholder to continue the discussion from https://github.com/oppia/oppia-android/pull/4139#discussion_r796232926.

app/src/main/res/values/styles.xml
```diff
@@ -459,4 +459,9 @@
...
+  <style name="ShapeAppearanceOverlay.App.Circular" parent="">
+    <item name="cornerSize">50%</item>
+  </style>
```
BenHenning, line 475: Another comment to continue the discussion from https://github.com/oppia/oppia-android/pull/4139#discussion_r796232690 since it hasn't been resolved yet.

app/src/main/res/layout-land/profile_edit_fragment.xml
```diff
@@ -33,7 +33,7 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-      <de.hdodenhof.circleimageview.CircleImageView
+      <com.google.android.material.imageview.ShapeableImageView
```
BenHenning, line 36: Continuing the discussion from https://github.com/oppia/oppia-android/pull/4139#discussion_r796231929.

I see that not all uses of ShapeableImageView specify an appearance overlay, but more do since my last review. Should all of them?

BenHenning (changes requested):
Thanks for the latest changes & recreating the PR @bhaktideshmukh. Took another pass and tried to retain my comments from the original PR--PTAL.

app/src/main/res/values/styles.xml
bhaktideshmukh: I took this document as reference for this property.
[](https://github.com/oppia/oppia-android/pull/4139#discussion_r796442630)https://howtodoandroid.com/shapeableimageview-material-components-android/

app/src/main/res/values/styles.xml
bhaktideshmukh: From your previous comment regarding omitting this parent part, I guess that's not possible since its the part of this property and omitting will give error.

app/src/main/res/layout-land/profile_edit_fragment.xml
bhaktideshmukh: Initially I didn't added appearance overlay property to all the files but now I do think it should be added to all.
What are your thoughts on it? 
Should I add this property to all files?

app/src/main/res/values/styles.xml
bhaktideshmukh: Actually initially I didn't added this style part but after building the view which should match the mock as circular was resulting as square. Hence I added this style part.

||

bhaktideshmukh:
PTAL @BenHenning 
Thanks

app/src/main/res/values/styles.xml
BenHenning: Thanks for the image. To my original question then, why do some views not specify this style but others do? Wouldn't we want all uses of ShapeableImageView currently used by Oppia use this style since they're all CircularImageView today?

app/src/main/res/values/styles.xml
BenHenning: Thanks. I suggest adding a comment pointing to that page for context, then, as it'll help others understand why it's correct if they come across this in the future (unfortunately, the proper name makes this a bit difficult to understand hence why the comment will have contextualize).

app/src/main/res/values/styles.xml
BenHenning: If a parent needs to be defined, then we ought to derive this from the main Oppia theme since that's also a material theme and may pre-define the cornerSize property.

app/src/main/res/layout-land/profile_edit_fragment.xml
BenHenning: Do the other views look correct? I'm actually a bit confused why they all look correct if only some have the property defined (since your image above suggests not defining this property results in a square image). Can you provide some context on why the views not using the property look correct (or if they don't)?

BenHenning:
Thanks @bhaktideshmukh. Left follow-up replies.

app/src/main/res/values/styles.xml
bhaktideshmukh: yes you were right some were having square view. 
I have fixed them now all the views are circular in shape.

But I have observed that in profile_progress_header,  profile_edit_fragment,  profile_list_profile_view shadow part is missing. Since the shadow attribute does not exist in ShapeableImageView , view is not matching the mock view.

**profile_progress_header** 

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

**add_profile_activity**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

**profile_list_profile_view**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

app/src/main/res/layout-land/profile_edit_fragment.xml
bhaktideshmukh: Done

app/src/main/res/values/styles.xml
bhaktideshmukh: https://howtodoandroid.com/shapeableimageview-material-components-android/  
From here I read that the key to the overlay behavior is the empty parent attribute. This is what causes the overlay behavior.

app/src/main/res/layout-land/profile_edit_fragment.xml
bhaktideshmukh: I have added the appearance overlay property for other views which were missing and now all are circular.

app/src/main/res/values/styles.xml
bhaktideshmukh: Done

bhaktideshmukh:
PTAL @BenHenning 
Thanks.

BenHenning:

app/src/main/res/layout/profile_chooser_add_view.xml
```diff
@@ -31,13 +31,15 @@
       app:layoutMarginStart="@{hasProfileEverBeenAddedValue ? @dimen/profile_chooser_add_view_margin_start_profile_already_added : @dimen/profile_view_not_added_margin}"
       app:layoutMarginTop="@{hasProfileEverBeenAddedValue ? @dimen/space_0dp : @dimen/profile_chooser_add_view_margin_top_profile_not_added}">
 
-      <de.hdodenhof.circleimageview.CircleImageView
+
```
BenHenning: Nit: please remove the extra newline.

app/src/main/res/layout/add_profile_activity.xml
```diff
@@ -38,19 +38,22 @@
...
+          <com.google.android.material.imageview.ShapeableImageView
             android:id="@+id/add_profile_activity_user_image_view"
             android:layout_width="120dp"
             android:layout_height="120dp"
             android:clickable="true"
             android:contentDescription="@string/current_profile_picture_content_description"
             android:focusable="true"
             app:add_shadow="true"
+            app:strokeColor="@color/avatar_border"
+            app:strokeWidth="1dp"
```
BenHenning: These properties weren't here before--did you mean to add them?

app/src/main/res/values/styles.xml
BenHenning: Nice catch--what will need to be changed to bring back the shadow?

app/src/main/res/values/styles.xml
BenHenning: Ah, I think that should be explicitly called out then in the comment above, and why it works (since otherwise it looks like a potential mistake that may be incorrectly changed in the future).

BenHenning (changes requested):
Thanks @bhaktideshmukh! Had a few follow-up comments--PTAL. It seems like the main gap at this point is the lack of shadows.

Could you please also sync with the latest develop and resolve the conflict, and then ensure that all CI checks are passing?

app/src/main/res/values/styles.xml
bhaktideshmukh: Done

app/src/main/res/layout/profile_chooser_add_view.xml
bhaktideshmukh: Done

app/src/main/res/layout/add_profile_activity.xml
bhaktideshmukh: Apologies , I forgot to stash these changes.
I have removed them.

app/src/main/res/values/styles.xml
bhaktideshmukh: > Nice catch--what will need to be changed to bring back the shadow?

I didn't found any solution , there is no such attribute in ShapeableImageView to add shadow.

bhaktideshmukh:
PTAL @BenHenning 
Thanks.

app/src/main/res/values/styles.xml
BenHenning: That's something that will need to be fixed before we can merge this PR--we don't want to regress the UI as part of this migration.

app/src/main/res/values/styles.xml
```diff
@@ -464,4 +464,12 @@
   <style name="TextViewCenterHorizontal" parent="TextView.Common1">
     <item name="android:gravity">center_horizontal</item>
   </style>
+
+  <!-- ShapeableImageView -->
+  <!-- context for ShapeableimageView https://howtodoandroid.com/shapeableimageview-material-components-android/
+       -->
+  <!-- Here key to the overlay behavior is the empty parent attribute. This is what causes the overlay behavior -->
```
BenHenning: Please combine these three comments into one and make sure to use full sentence structures (except for the first bit, that can just be 'ShapeableImageView').

The last bit could also probably be a bit clearer, e.g.: 'Note that the parent attribute is intentionally empty because if enables the overlay behavior for the shape due to ...' (note that the last bit needs to be filled in--the current text doesn't actually explain why this works).

Finally, please move the context line to the end (usually contexts are after the rest of the explanation to provide further reading).

BenHenning (commented):
Thanks @bhaktideshmukh! Just had a few follow-up comments--PTAL (I think the main gap at this point is fixing the shadow regression).

BenHenning:
Also @bhaktideshmukh please sync this with the latest develop branch & resolve any conflicts. After that, please make sure CI checks are passing.

app/src/main/res/values/styles.xml
bhaktideshmukh: Hey I tried searching about how we can add shadow to this view but didn't found anything.
So what shall I do now?

> That's something that will need to be fixed before we can merge this PR--we don't want to regress the UI as part of this migration.

I tried searching about how we can add shadow to this view but unfortunately nothing came up.
Now I don't know what I should do next.
Please suggest something regarding this?

app/src/main/res/values/styles.xml
bhaktideshmukh: Done

bhaktideshmukh:
PTAL @BenHenning 
Thanks.

app/src/main/res/values/styles.xml
```diff
@@ -465,10 +465,11 @@
...
 
-  <!-- ShapeableImageView -->
-  <!-- context for ShapeableimageView https://howtodoandroid.com/shapeableimageview-material-components-android/
-       -->
-  <!-- Here key to the overlay behavior is the empty parent attribute. This is what causes the overlay behavior -->
+  <!-- ShapeableImageView
+     Note that When defining a style to be used as a shapeAppearanceOverlay in shapeableImageView , we generally
+     set an empty parent by setting parent="". This is what causes the overlay behavior and also
+     shapeAppearanceOverlay stacks on top of shapeAppearance, so all values will be set in the shapeAppearance.
+     context for ShapeableImageView https://howtodoandroid.com/shapeableimageview-material-components-android/ -->
```
BenHenning: This still doesn't quite explain _why_ the parent bit works--that's the thing that's important to include, I think, since otherwise we won't understand how to debug this if it breaks in the future (or doesn't work for some specific situation).

app/src/main/res/values/styles.xml
BenHenning: I suggest two different avenues:
1. Checking how the old library implemented its shadow and see if we can take a similar approach here
2. Looking how people generally add shadows to overlays or shaped images in Android and see if any of those approaches can be adapted here

BenHenning (commented):
Thanks @bhaktideshmukh. Left follow-up comments.

Also, there's still a conflict--was this updated with the latest develop (from oppia/oppia-android, not your fork)?

app/src/main/res/values/styles.xml
bhaktideshmukh: I didn't get what exactly I should write , I found this much only in the documents related to ShapeableImageView regarding this empty parent attribute.

app/src/main/res/values/styles.xml
bhaktideshmukh: > I suggest two different avenues:
> 
>     1. Checking how the old library implemented its shadow and see if we can take a similar approach here
> 
>     2. Looking how people generally add shadows to overlays or shaped images in Android and see if any of those approaches can be adapted here

1. That's the main problem , In CircularImageView there was an inbuilt attribute of add_shadow to add the shadow where as in ShapeableImageView there no such attribute exists.
2. I tried to find it on Google but I couldn't find anything relevant. Are there any other places where I could look as well?

bhaktideshmukh:
> Thanks @bhaktideshmukh. Left follow-up comments.
> 
> Also, there's still a conflict--was this updated with the latest develop (from oppia/oppia-android, not your fork)?

Done.
Now all the tests are passing.

app/src/main/res/values/styles.xml
BenHenning: @bhaktideshmukh out of curiosity, have you tried setting the elevation on the image? Most material components respond to elevation and show shadows for those cases.

BenHenning:
@bhaktideshmukh I followed up on the discussion thread re: shadows. Also, glad to hear that the CI tests are now passing. :)

app/src/main/res/values/styles.xml
bhaktideshmukh: > @bhaktideshmukh out of curiosity, have you tried setting the elevation on the image? Most material components respond to elevation and show shadows for those cases.


Adding elevation on the image is giving view something like this

| |

Here elevation is by 1 dp.

bhaktideshmukh:
@BenHenning 

I tried adding strokeWidth and strokeColor to the view , it now little bit seems like shadow.
Please have a look:--


| |

app/src/main/res/values/styles.xml
BenHenning: Which view did you update the elevation for in that screenshot @bhaktideshmukh? It doesn't actually look like the shapeable image has elevation.

BenHenning:
> @BenHenning
> 
> I tried adding strokeWidth and strokeColor to the view , it now little bit seems like shadow. Please have a look:--
> 
> | |

Interesting. I think using elevation would still be preferred since it seems more "Material"-y, but this might be a good contingency if that can't be made to work. Could you add a 'before' screenshot for the exact same screen so we can do a before-and-after comparison? We'd need to tweak this one to be close the old one, but I think it's worth trying to get elevation working first.

app/src/main/res/values/styles.xml
bhaktideshmukh: After adding elevation :--

**profile_progress_header**

| Original  | Before elevation | After elevation |
|-------|---|----|
|||

**add_profile_activity**

| Original  | Before elevation| After elevation |
|-------|---|----|
|||

**profile_list_profile_view**

| Original  | Before elevation | After elevation |
|-------|---|----|
|||

bhaktideshmukh:
> > @BenHenning
> > I tried adding strokeWidth and strokeColor to the view , it now little bit seems like shadow. Please have a look:--
> > | |
> 
> Interesting. I think using elevation would still be preferred since it seems more "Material"-y, but this might be a good contingency if that can't be made to work. Could you add a 'before' screenshot for the exact same screen so we can do a before-and-after comparison? We'd need to tweak this one to be close the old one, but I think it's worth trying to get elevation working first.

Fine we will consider adding stroke width later in case setting elevation doesn't works.

app/src/main/res/layout/profile_list_profile_view.xml
```diff
@@ -25,6 +25,7 @@
       android:layout_marginTop="18dp"
       android:layout_marginBottom="18dp"
       android:background="@drawable/profile_image_shadow"
+      android:elevation="10dp"
```
BenHenning: Why this difference compared to the other views? Also, I'm wondering if this should be part of the common style in the same way the shape form is.

app/src/main/res/values/styles.xml
BenHenning: Thanks for the screenshots @bhaktideshmukh! I actually think elevation looks pretty good, and it's a really simple solution. However, I think the elevation you chose might bring on a slightly too strong shadow. Can you experimentally find an elevation value that puts the shadow visually closer to what we originally had?

BenHenning (commented):
Thanks @bhaktideshmukh! Just had a few follow-up comments--PTAL.

app/src/main/res/layout/profile_list_profile_view.xml
bhaktideshmukh: > Why this difference compared to the other views? Also, I'm wondering if this should be part of the common style in the same way the shape form is.

The difference was due to its size as it was smaller compared to the other ones.Although now I changed all of them to elevation of 8dp.
I tried to add this in common style but it is again resulting in invisible shadow.

app/src/main/res/values/styles.xml
bhaktideshmukh: > Thanks for the screenshots @bhaktideshmukh! I actually think elevation looks pretty good, and it's a really simple solution. However, I think the elevation you chose might bring on a slightly too strong shadow. Can you experimentally find an elevation value that puts the shadow visually closer to what we originally had?

I have reduced the elevation from 20dp to 8dp and now its more closer to the shadow we had in the original one.

app/src/main/res/layout/profile_list_profile_view.xml
BenHenning: Can you show what the new before & afters look like (the version without a shadow isn't needed, I think we're just interested now in the difference between the shadows) using the 8dp elevation?

Regarding the common style, can you provide a Gist with the styles change to provide context on what you tried? It'd be good to understand that. I'm not sure if this is a platform limitation, but gathering more data can be helpful.

bhaktideshmukh: > Why this difference compared to the other views? Also, I'm wondering if this should be part of the common style in the same way the shape form is.

I guess I misunderstood here something.
Can you please elaborate what exactly I m supposed to do regarding this common style?

bhaktideshmukh: > Can you show what the new before & afters look like (the version without a shadow isn't needed, I think we're just interested now in the difference between the shadows) using the 8dp elevation?
> 
> Regarding the common style, can you provide a Gist with the styles change to provide context on what you tried? It'd be good to understand that. I'm not sure if this is a platform limitation, but gathering more data can be helpful.


| Before (for 20dp elevation)  | After (for 8dp elevation) |
|-------|---|
|||

seanlip: @bhaktideshmukh See https://docs.github.com/en/get-started/writing-on-github/editing-and-sharing-content-with-gists/creating-gists for what a "Gist" is.

BenHenning: Sorry to clarify @bhaktideshmukh I'm interested in seeing the comparison of the current CircularImageView vs. ShapeableImageView with 8dp (since that's the main consideration). However, with the images you provided I can do a side-by-side with those two.

I think the 8dp shadow still looks a bit thicker than the previous version. Can you confirm and maybe try something like 4dp, instead? This is actually what I meant by 'experimentally finding the correct shadow': you should change the shadows and compare with the original images produced by CircularImageView until the shadow looks nearly the same.

bhaktideshmukh: > Sorry to clarify @bhaktideshmukh I'm interested in seeing the comparison of the current CircularImageView vs. ShapeableImageView with 8dp (since that's the main consideration). However, with the images you provided I can do a side-by-side with those two.
> 
> I think the 8dp shadow still looks a bit thicker than the previous version. Can you confirm and maybe try something like 4dp, instead? This is actually what I meant by 'experimentally finding the correct shadow': you should change the shadows and compare with the original images produced by CircularImageView until the shadow looks nearly the same.

Can you please have a look at the screenshots:--
I have added screenshots for different elevations -- 5dp,4dp,3dp.
I think 4 or 5 dp is giving closer view to the original CircularImageView.
Once you approve the final elevation I'll push the changes here.

**Elevation of 5dp**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

**Elevation of 4dp**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

**Elevation of 3dp**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

BenHenning:
Thanks @bhaktideshmukh. I think I lean toward the 4dp; it looks the closest to me (and it's conveniently divisible by 4 which is common in Material design). @rt4914 WDYT? PTAL: https://github.com/oppia/oppia-android/pull/4155#discussion_r819300366.

bhaktideshmukh:
> Thanks @bhaktideshmukh. I think I lean toward the 4dp; it looks the closest to me (and it's conveniently divisible by 4 which is common in Material design). @rt4914 WDYT? PTAL: [#4155 (comment)](https://github.com/oppia/oppia-android/pull/4155#discussion_r819300366).

I have pushed the changes -- reduced the elevation to 4dp.

bhaktideshmukh:
@BenHenning
I apologize , by mistake nav_header_navigation_drawer.xml file was left unchanged.
I replaced the CircularImageView with ShapeableImageView and now its matching the mock view.
Thanks.

**nav_header_navigation_drawer.xml**

| CircularImageView  | ShapeableImageView |
|-------|---|
|||

rt4914:
@BenHenning @bhaktideshmukh Definitely `4dp` is the most correct value. I agree with this.

bhaktideshmukh:
@BenHenning 
Gentle reminder to review this PR.
Thanks.

app/src/main/res/layout/nav_header_navigation_drawer.xml
```diff
@@ -24,9 +24,10 @@
     android:paddingBottom="@dimen/activity_vertical_margin"
     android:theme="@style/ThemeOverlay.AppCompat.Dark">
 
-    <de.hdodenhof.circleimageview.CircleImageView
+    <com.google.android.material.imageview.ShapeableImageView
       android:layout_width="64dp"
       android:layout_height="64dp"
+      app:shapeAppearanceOverlay="@style/ShapeAppearanceOverlay.App.Circular"
```
BenHenning: Should this also have elevation specified?

app/src/main/res/values/styles.xml
BenHenning: Sorry, should've followed up on this earlier. We generally shouldn't be making changes just because they work--it's really important to understand the 'why' (otherwise we may unintentionally bring extra consequences that we might miss adopting a solution that may only seem correct on the surface). I think we need to better understand why this works & reflect that here so that there isn't confusion.

I think the main gap is understanding why setting an empty parent works specifically, since that's actually rather surprising. I'm assuming it's preventing some sort of override that would otherwise prevent this behavior (but that would only be true if an overlay is the default, which I'm not sure). This uncertainty is actually I think it's important to fully understand the solution.

BenHenning (commented):
Thanks @bhaktideshmukh. Super sorry about the delayed review--I've been dealing with a super urgent matter for the last week.

Took another pass on the PR. I think code-wise it looks good (though I think one view might've been missed--I left a comment on this), but I want to make sure the solution is properly captured in your comment since we're not going to know as much about this approach as we do now in the future when we might go back and try to understand what's going on.

app/src/main/res/layout/nav_header_navigation_drawer.xml
bhaktideshmukh: No here elevation will not be added since we added it  only to the activities where shadow was visible in the mocks.

BenHenning:

I'll take a pass on Monday.In the meantime, I think your CI checks are failing @bhaktideshmukh. Can you please make sure that they're passing? I think your recent merge of develop may have led to a failure.

bhaktideshmukh:
> Hi. Really sorry for the review delay--I was still really slammed the last few weeks.
> 
> I'll take a pass on Monday.In the meantime, I think your CI checks are failing @bhaktideshmukh. Can you please make sure that they're passing? I think your recent merge of develop may have led to a failure.

I tried passing CI checks but I am still stuck on maven error.
I'll try it once again once review is done.

bhaktideshmukh:
@BenHenning 
All the checks are passing now.

app/src/main/res/values/styles.xml
```diff
@@ -465,11 +465,14 @@
...
-     shapeAppearanceOverlay stacks on top of shapeAppearance, so all values will be set in the shapeAppearance.
-     context for ShapeableImageView https://howtodoandroid.com/shapeableimageview-material-components-android/ -->
+  <!-- ShapeableImageView -->
+  <!-- Note that when defining a style to be used as a shapeAppearanceOverlay in shapeableImageView , we generally
+       set an empty parent by setting parent="".This is what causes the overlay behavior.
+       When we define a style,we would provide a parent style, and then we can override specific attributes to alter specific parts of that style.
+       When we specify no parent, then the style will overlay whatever theme is being applied. Applying explicit parent attribute instead of a empty
+       parent may cause overriding(other properties may get added i.e. color or theme used in activity).
+       Thus here in ShapeAppearanceOverlay we have defined a empty parent to allow the active theme behavior to apply.
+       https://blog.stylingandroid.com/material-shape-shapeappearance/ -->
```
BenHenning: So after re-reading the Blog, I'm starting to wonder if we're actually taking the correct approach. It seems setting ``parent=""`` only incidentally fixes the problem since the theme just happens to have the correct default corner family type. Could we perhaps not specify an empty parent and instead specify a corner family of rounded?

Per https://material.io/develop/android/theming/shape, https://medium.com/androiddevelopers/material-theming-with-mdc-shape-126c4e5cd7b4, and https://material.io/design/shape/about-shape.html#shape-customization-tool it seems like that should work. It'll also be easier to document since we can just say something like:

```
The corner size & family type are set up to ensure a circular shape is created for components with equal width & height (per https://material.io/design/shape/about-shape.html#shape-customization-tool).
```

BenHenning (commented):
Apologies again for the super late follow-up @bhaktideshmukh.

I left a new comment. I think that the difficulty in creating a clear explanation for the parent property is potentially indicative that the referenced blog author didn't clearly communicate why the code was correct. After doing some digging, I don't think it's necessary and that we can be more explicit in the style property definitions. PTAL at my latest comment.

app/src/main/res/values/styles.xml
bhaktideshmukh: I have pushed the changes:-- added cornerfamily and removed empty parent.

app/src/main/res/values/styles.xml
```diff
@@ -466,14 +466,11 @@
...
-       set an empty parent by setting parent="".This is what causes the overlay behavior.
-       When we define a style,we would provide a parent style, and then we can override specific attributes to alter specific parts of that style.
-       When we specify no parent, then the style will overlay whatever theme is being applied. Applying explicit parent attribute instead of a empty
-       parent may cause overriding(other properties may get added i.e. color or theme used in activity).
-       Thus here in ShapeAppearanceOverlay we have defined a empty parent to allow the active theme behavior to apply.
-       https://blog.stylingandroid.com/material-shape-shapeappearance/ -->
-  <style name="ShapeAppearanceOverlay.App.Circular" parent="">
+  <!-- The corner size & family type are set up to ensure a circular shape is created for components
+   with equal width & height.
+   (per https://material.io/design/shape/about-shape.html#shape-customization-tool). -->
```
BenHenning:
```suggestion
  <!-- The corner size & family type are set up to ensure a circular shape is created for components
   with equal width & height
   (per https://material.io/design/shape/about-shape.html#shape-customization-tool). -->
```

app/src/main/res/values/styles.xml
BenHenning: The latest wording addresses my concern on clarity (mostly because the latest code is more self-documenting).

app/src/main/res/values/styles.xml
BenHenning: Thanks @bhaktideshmukh! Can you confirm that all of the screens still look correct after this?

app/src/main/res/values/styles.xml
```diff
@@ -464,4 +464,13 @@
   <style name="TextViewCenterHorizontal" parent="TextView.Common1">
     <item name="android:gravity">center_horizontal</item>
   </style>
+
+  <!-- ShapeableImageView -->
+  <!-- The corner size & family type are set up to ensure a circular shape is created for components
+   with equal width & height.
+   (per https://material.io/design/shape/about-shape.html#shape-customization-tool). -->
+  <style name="ShapeAppearanceOverlay.App" >
```
BenHenning:
```suggestion
  <style name="ShapeAppearanceOverlay.App">
```

BenHenning: Given that we're determining this name, perhaps we can make it more descriptive. What about: ``ShapeableImageView.RoundedShape`` (or something similar)?

BenHenning (commented):
Thanks @bhaktideshmukh! I think the latest changes are much clearer. I have just a few follow-up comments, but otherwise the PR is pretty close to LGTM. PTAL.

app/src/main/res/values/styles.xml
bhaktideshmukh: Done

app/src/main/res/values/styles.xml
bhaktideshmukh: > Thanks @bhaktideshmukh! Can you confirm that all of the screens still look correct after this?

Yes. All the screens are matching the mocks now.

app/src/main/res/values/styles.xml
bhaktideshmukh: Done

bhaktideshmukh: > Given that we're determining this name, perhaps we can make it more descriptive. What about: `ShapeableImageView.RoundedShape` (or something similar)?

It seems like we need to stick to the 'ShapeAppearanceOverlay' name here but I added the 'RoundedShape' as you suggested.

BenHenning (approved):
Thanks @bhaktideshmukh! I appreciate your patience with my delayed reviews. I think the PR LGTM, and it's honestly a really nice improvement maintenance-wise. Thanks for doing this!

BenHenning:
Going ahead and merging this since no other changes are needed or requested.

PR #4172, opened 2022-02-11
Fix #4165 : Added e2e tests for TopicActivity

### LINKED ISSUE #4165: Add e2e tests for TopicActivity
Add End-to-End tests for the TopicActivity using Bazel + UiAutomator.

Test case should include the flow of
- Opening a topic.
- Exploration all the features of the topic.

To pick this issue you need:
- [Bazel setup](https://github.com/oppia/oppia-android/wiki/Oppia-Bazel-Setup-Instructions) 
- to be able to run a [E2E test](https://github.com/oppia/oppia-android/wiki/End-to-End-Testing-Guide#how-to-run-an-end-to-end-test)

References for writing E2E tests
- [End to End Testing Guide](https://github.com/oppia/oppia-android/wiki/End-to-End-Testing-Guide#how-to-run-an-end-to-end-test)
- [Android Instrumentation Tests, Bazel](https://docs.bazel.build/versions/5.0.0/android-instrumentation-test.html)

ayush0402:
## Explanation

Fixes #4165

https://user-images.githubusercontent.com/76056229/155803055-eb08dfb4-2980-497b-81b9-8b1006a10d49.mp4

## Essential Checklist


ayush0402:
@FareesHussain Can you take a look at the tests and suggest if I need to add/remove any tests?

Also, I tried adding a test for checking if show more actually expands the info description. I assumed that the text in the text view is actually changed so I tried writing this test assuming the before and after size of the text must be different...
```kotlin
@Test
  fun testTopicActivity_infoFragment_clickedSeeMore_descriptionExpanded() {
    navigateToTopicActivity()
    val initialDescription =
      device.findObjectByRes("topic_description_text_view").getText().toString().length
    device.findObjectByText("See More").click()
    device.waitForText("See Less")
    //Assert
    val finalDescription =
      device.findObjectByRes("topic_description_text_view").getText().toString().length
    val result = if ((finalDescription - initialDescription) > 0) true else false
    assertThat(result).isTrue()
  }
```
But after some debugging I realised that the size of the initial and after strings are just the same and it is using some other mechanism for actually expanding the text. So, I dropped this test for now.

Also, I had a doubt if my current way of checking if we have switched screen by checking for something supposed to be on the next screen is correct or is there a better way to do this?

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
```diff
@@ -0,0 +1,136 @@
...
+  @Test
+  fun testTopicActivity_infoFragment_clickSeeLess_showSeeMore() {
+    navigateToTopicActivity()
+    device.findObjectByText("See More").click()
+    device.waitForText("See Less")
+    device.findObjectByText("See Less").click()
+    // Assert See Less is visible
+    val seeMoreText = device.findObjectByText("See More")
+    assertThat(seeMoreText).isNotNull()
+  }
```
FareesHussain: @ayush0402 since we only need to check whether all the components of the TopicActivity are working we don't need multiple test cases for this.

The flow is exactly same as the video you sent in the chat. make sure to attach a video of the running tests.

FareesHussain (changes requested):
@ayush0402 PTAL left some comments

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Removed `testTopicActivity_infoFragment_clickSeeLess_showSeeMore` test. And added the video for current state of the tests to the PR explanation.

FareesHussain: include these back, but in the single testcase.

ayush0402: @FareesHussain PTAL if this is the correct implementation for the test. Also, updated the video.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    navigateToTopicActivity()
+    device.findObjectByText("See More").click()
+    device.waitForText("See Less")
+    device.findObjectByText("See Less").click()
+
+    device.findObjectByText("LESSONS").click()
+    device.findObjectByRes("expand_list_icon").click()
+    device.waitForRes("chapter_list_container")
+    device.findObjectByText("What is a Fraction?").click()
+    device.waitForText("CONTINUE")
```
FareesHussain: Continue button might get beyond the screen size. check if exploration specific resource id is visible or not

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    device.pressBack()
```
FareesHussain: After going back, add an assertion to check the visibliity of the icon displayed in the drop down.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    device.findObjectByText("PRACTICE").click()
+    checkSkillsForPracticeFragment()
+    device.findObjectByText("START").click()
+    device.waitForText("PRACTICE MODE")
+    device.pressBack()
+    device.findObjectByText("LEAVE").click()
+
+    device.findObjectByText("REVISION").click()
+    device.findObjectByText("Fractions of a group").click()
+    device.findObjectByText("RETURN TO TOPIC").click()
```
FareesHussain: Add a assertion for each of "Lessons", "practice", "Revision" to check if the user returns to the same tab in the topic

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByText
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.scrollRecyclerViewTextIntoView
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.startOppiaFromScratch
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForText
+
+/** Tests for TopicActivity. */
+class TopicActivityTest {
```
FareesHussain: I think this should be "TopicTest" instead.

Lets wait for ben's suggestion

FareesHussain (changes requested):
@ayush0402 Thanks

Left a few comments, rest looks good to me.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Ah, I get it now. I was under the impression that we should only have one assertion per test so was confused about asserting in this case. Will add these, Thanks.

FareesHussain:
@BenHenning despite having `tag = ["manual"]` for android_instrumentation_test in oppia_instrumentation_test

it is included in the computed tests
image

It seems the code for filtering instrumentation test targets have been removed in [this pr](https://github.com/oppia/oppia-android/pull/3757/files#diff-a4469a7f9299fffb1edc114d1b757e15c67a9fa9c2496dc9bb2d61bfe82fc1c9). IDK why all the test cases in ComputedAffectedTestsTest are passing.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Done.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Renamed to TopicTest for now. Will revert this if Ben suggests otherwise.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Done.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Done.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicActivityTest.kt
ayush0402: Can you explain this a bit more? If this means to use `waitForRes` instead then done.

ayush0402:
@FareesHussain PTAL. Updated the video as well.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
```diff
@@ -0,0 +1,84 @@
...
+
+  @Before
+  fun setUp() {
+    // Initialize UiDevice instance
+    device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+    device.startOppiaFromScratch()
+  }
+
+  @Test
+  fun testTopicActivity_exploredFully_finishedSuccessfully() {
```
BenHenning: It feels like each of these distinct blocks ought to be independent tests since they're verifying different aspects of the topic experience. This is a bit different than the exploration test since that needs to play through a whole exploration to verify that it can.

I think this might be slightly in contradiction with earlier feedback from @FareesHussain (Farees: please follow up if you disagree with this).

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByText
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.scrollRecyclerViewTextIntoView
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.startOppiaFromScratch
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForText
+
+/** Tests for TopicActivity. */
```
BenHenning: We shouldn't tie to specific activities at this level. Instead, suites contain categories of user pathways that are important to verify end-to-end. I suggest instead:

```suggestion
/** Tests for the topic viewer screen. */
```

BenHenning (changes requested):
Thanks @ayush0402! Left some initial comments--PTAL.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
FareesHussain: We had distinct blocks in exploration as well, each interaction had a different aspect. I believe it should be same here as well, each tab or component are independent. Only reason I can think of for having multiple tests is when we have interdependent topics.

I can't think of a reason for having independent tests.

FareesHussain (commented):
@BenHenning In case you've missed this https://github.com/oppia/oppia-android/pull/4172#issuecomment-1046023519

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
ayush0402: Done.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
BenHenning: Generally, we should have 1 test method per behavior that we're testing. However, this doesn't translate as cleanly for e2e tests since the covered behaviors are so high-level. Instead, I think we should define them as 1 test per single user flow that's being verified. In the case of explorations, we had 1 test to verify that we could play through and finish an exploration. Any other scenarios (such as verifying that hints work in various cases) would need to be in a separate test method rather than being added onto the existing test method.

It seems to me that this test, unlike the one for explorations, is actually verifying multiple scenarios and therefore would benefit from the cleanliness of having distinct test methods.

Note that I'm also thinking about how these suites are going to evolve over time (one possibility is that many of our QA tests may end up being built as UiAutomator tests so establishing a clear standard here is important).

WDYT?

BenHenning (commented):
Apologies for the delay; just followed up on the ongoing thread. PTAL @ayush0402 and @FareesHussain.

ayush0402:
@BenHenning PTAL if this way of writing tests having each fragment separated into a separate test method is good or not.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
```diff
@@ -0,0 +1,100 @@
...
+
+  @Before
+  fun setUp() {
+    // Initialize UiDevice instance
+    device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+    device.startOppiaFromScratch()
+  }
+
+  @Test
+  fun testTopicActivity_exploreInfoFragment_finishesSuccessfully() {
```
BenHenning: 'Fragment' is too implementation specific. This should be tied only to user behaviors, e.g. ``testTopicViewer_exploreInfoFragment_hasSummaryDetails`` or something similar. Ditto for the other tests.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.junit.Test
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.findObjectByText
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.scrollRecyclerViewTextIntoView
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.startOppiaFromScratch
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForRes
+import org.oppia.android.instrumentation.testing.EndToEndTestHelper.waitForText
+
+/** Tests for the topic viewer screen. */
+class TopicTest {
```
BenHenning: Perhaps instead: TopicViewerTest for consistency.

BenHenning (commented):
Thanks @ayush0402. Had a couple of follow-up thoughts--PTAL.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
ayush0402: Done.

ayush0402:
Delayed as this needs more discussion.

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
FareesHussain: suggesting 'Tab' instead of 'Fragment'
`testTopicViewer_exploreInfoTab_hasSummaryDetails`

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
FareesHussain: Sounds good, also had a discussion with Vinita and sean regarding this
@ayush0402 PTAL

instrumentation/src/javatests/org/oppia/android/instrumentation/topic/TopicTest.kt
ayush0402: Done

ayush0402:
@FareesHussain @BenHenning PTAL.

BenHenning:
Apologies. I will need to look at this tomorrow.

ayush0402:
@BenHenning Can you PTAL at this?
I am getting this error when trying to run these tests for checking progress using contentDescription.
Screenshot from 2022-05-02 20-46-59
(We want to add tests for orange color in circular progress bar after the lesson has been opened once but there seems not to be any direct method of testing this through UIAutomator other than the contentDescription.)
Can you suggest what going wrong with this approach?
Thanks.

BenHenning:
> @BenHenning Can you PTAL at this? I am getting this error when trying to run these tests for checking progress using contentDescription. Screenshot from 2022-05-02 20-46-59 (We want to add tests for orange color in circular progress bar after the lesson has been opened once but there seems not to be any direct method of testing this through UIAutomator other than the contentDescription.) Can you suggest what going wrong with this approach? Thanks.

@ayush0402 have you tried manually running through the steps of the test with TalkBack enabled and debug speech output? I'm curious to know whether you just stumbled upon an actual issue with the content description. If there's an inconsistency, then we'll need to figure out why. If not, we should fix the content description issue which should, in turn, fix the test.

app/src/main/java/org/oppia/android/app/topic/lessons/StorySummaryViewModel.kt
```diff
@@ -55,6 +66,11 @@ class StorySummaryViewModel(
     )
   }
 
+  private fun computeContentDescriptionText(): String {
+    return "$storyProgressPercentageText stories " +
+      "completed and $chaptersInProgress chapters in progress."
```
BenHenning, line 71: This needs to go in strings.xml so that it can be translated. There are other uses of variable interpolation in strings that can be referenced in the codebase.

Note that I don't think this would replace topic_story_progress_percentage since it seems like we also need to use a plural string in this case since 'chapter' can be plural.

app/src/main/java/org/oppia/android/app/topic/lessons/StorySummaryViewModel.kt
```diff
@@ -30,6 +35,12 @@ class StorySummaryViewModel(
   fun setStoryPercentage(storyPercentage: Int) {
     this.storyPercentage.set(storyPercentage)
     storyProgressPercentageText.set(computeStoryProgressPercentageText(storyPercentage))
+    contentDescriptionText.set(computeContentDescriptionText())
+  }
+
+  fun setChaptersInProgress(chaptersInProgress: Int) {
```
BenHenning, line 41: Could this & the above maybe be set at the same time? Otherwise there's the chance of the content description only being partly correct which is a bit awkward.

app/src/main/res/layout/topic_lessons_story_summary.xml
```diff
@@ -52,7 +52,7 @@
             android:id="@+id/story_progress_container"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:contentDescription="@{viewModel.storyProgressPercentageText}">
+            android:contentDescription="@{viewModel.contentDescriptionText}">
```
BenHenning, line 55: Since this is being changed, can you also add a Robolectric/Espresso test to verify that it's computed correctly? We shouldn't be solely relying on e2e tests for regression testing.

BenHenning (commented):
Thanks @ayush0402. Given the content description change I think there's a bit more that needs to be done here, but the e2e testing side LGTM.

Also, reminder to make sure that CI checks are fixed since they may be exposing other problems that could change the PR.

PR #4204, opened 2022-02-17
Fix #4006: Hide edit accounts option from administrator controls 

### LINKED ISSUE #4006: Hide Edit Accounts option from AdministratorControls

JishnuGoyal:
## Explanation

Fixes #4006

Removes general section in `Administrator Controls` by removing AdministratorControlsGeneralViewModel from the itemViewModelList. 

Instead of just commenting out `AdministratorControlsGeneralViewModel`, a  new platform parameter, `ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE` was introduced and used to gate it. 
Now, the Edit Accounts options can be enabled simply by changing the `ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE` to `true`.

Why this method has been used can be found in detail under the comments here: https://github.com/oppia/oppia-android/issues/4006#issuecomment-981965721

The screen looks like this now:
image

Device used: Emulator Pixel 4 API 28

## Essential Checklist


JishnuGoyal:
In AdministratorControlsActivityTest and OptionsFragmentTest, I didn't use test modules to force platform parameter values. Instead, a companion object was introduced into PlatformParameterModule which directly helped to force the values in tests.

This was because:
If we use test modules, it would be imperative to remove PlatformParameterModule::class from the TestApplicationComponent (otherwise duplicate binding error will occur).
But removing PlatformParameterModule from any one test will break other tests as observed already.

JishnuGoyal:
PTAL @BenHenning

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -56,4 +66,17 @@ class PlatformParameterModule {
...
+    private var enableLanguageSelectionUi = ENABLE_LANGUAGE_SELECTION_UI_DEFAULT_VALUE
+    private var enableEditAccountsOptionsUi = ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE
+
+    fun forceEnableLanguageSelectionUi(value: Boolean) {
+      enableLanguageSelectionUi = value
+    }
+
+    fun forceEnableEditAccountsOptionsUi(value: Boolean) {
+      enableEditAccountsOptionsUi = value
+    }
```
BenHenning: Hmm, I'm not opposed to this pattern but it should only be how the test version of this module works, not the production version (prefer not punching holes into public APIs for testing purposes--this can lead to fragile tests and even cause unintended situations when those APIs bleed into production code).

We also probably should have a dedicated test suite for this module to make sure that these setters do the correct thing regarding the injected parameter values (to assure correctness in tests).

BenHenning (commented):
Thanks @JishnuGoyal. I think this approach is interesting, and will probably be much cleaner. However, I had some thoughts--PTAL.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
JishnuGoyal: I was afraid of this issue too since that approach takes away a degree of abstraction. But I'm curious, would this new implementation work though? Since it makes things explicit. If yes, i could proceed to write tests for the same as mentioned... PTAL @BenHenning

JishnuGoyal:
PTAL @BenHenning

BenHenning:
Apologies @JishnuGoyal. I'll need to delay my review on this due to being a bit slammed this week due to a release we're preparing for later in the week.

JishnuGoyal:
@BenHenning Could this be reopened ? Or shall i make a new PR?
PTAL @BenHenning

seanlip:
@JishnuGoyal Feel free to reopen any PRs that were closed due to staleness. Make sure to resolve merge conflicts, too. Thanks!

JishnuGoyal:
PTAL @BenHenning

BenHenning:

I'll take a pass on Monday.In the meantime, you should update to the latest develop branch @JishnuGoyal and address any conflicts.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
BenHenning: @JishnuGoyal can you clarify your specific concern about the degree of abstraction being taken away? I'm not sure that I fully follow.

I think the latest changes have the same issue. PlatformParameterModule ought to always return the production representations of flags, whereas the TestPlatformParameterModule could have more flexibility in altering the parameter flags ahead of injection

BenHenning (commented):
Apologies again for the late reply @JishnuGoyal. I followed up on the open thread, and will be more responsive moving forward.

JishnuGoyal:
In the last few pushes, I have moved the companion object from `PlatformParameterModule` to the `TestPlatformParameterModule` so as to make sure that the apis affecting the values of the parameters do not bleed into production.

Please note that it was necessary to make a copy of PlatformParameters in TestPlatformParameterModule from PlatformParameterModule class, otherwise there is no way to affect these parameters in tests without breaking the other tests according to me.
Not copying the PlatformParameters again into TestPlatformParameterModule would mean that we will have to keep both PlatformParameterModule::class and TestPlatformParameterModule::class in the test -- this leads to missing bindings error

JishnuGoyal:
Further, the CI complains that EditAccountsOptionsUi platform parameter is not being provided in optionsFragmentTest.
There are two problems:

I can't find where in OptionsFragmentTest this platform parameter is being utilized. [Still, i have included this parameter explicitly in the test, yet this error]

There is no clear msg in CI why this is happening. Please give me a hint here.

PTAL @BenHenning 
Thanks!

BenHenning:
@JishnuGoyal is it possible the error refers to the other ``OptionsFragmentTest``? (Note that there are two, one under sharedTest/ and one under test/, both of which may need to be updated for the tests to pass as expected).

Also, I find the new solution **much** cleaner, though to clarify: do the tests actually pass? I'd normally expect that the feature overrides to need to happen before injection, but if we're not scoping the parameters that may not be the case here.

JishnuGoyal:
>  I find the new solution much cleaner, though to clarify: do the tests actually pass? I'd normally expect that the feature overrides to need to happen before injection, but if we're not scoping the parameters that may not be the case here.

Thank you so much @BenHenning .
Yes, the tests do pass, as the TestPlatformParameterModule.forceEditAccountsOptionsUI() and other functions that I created change the value of that flag and inject the same just before they are used. (Does this exactly answer what you asked?)

JishnuGoyal:
Finally, could this be merged yet? Asking because all the _Required_ tests have passed --
and i feel the TestPlatformParameterModule is not recognised by bazel in some cases because its visibility is limited to the testing versions of the app?

Or shall I make some bazel config changes in order to pass the tests?
PTAL @BenHenning 
Thanks!

BenHenning:
> > I find the new solution much cleaner, though to clarify: do the tests actually pass? I'd normally expect that the feature overrides to need to happen before injection, but if we're not scoping the parameters that may not be the case here.
> 
> Thank you so much @BenHenning . Yes, the tests do pass, as the TestPlatformParameterModule.forceEditAccountsOptionsUI() and other functions that I created change the value of that flag and inject the same just before they are used. (Does this exactly answer what you asked?)

Ah, sort of. Looking at the module again, this works because each time the parameter is injected it re-reads the static state. This is technically a violation of the platform parameter module contract as it takes measures to guarantee singleton state among parameters (since changing parameters mid-app could cause undefined behaviors). That being said, it's annoying to do this "properly" as it requires refactoring the injection point in the test to occur after the override (which I've done in a bunch of test suites now). I don't care to push that pattern here even though it may cause some headaches down the line. The preferred way to do this would be to introduce a JUnit rule that can properly set up the parameters ahead of test running (I've built something like this before, and it can be quite nice). This PR introduces the foundations for that work, so I think it'd be a good option for future work (I filed #4302 to jot down some ideas on what such an approach might look like).



> Finally, could this be merged yet? Asking because all the _Required_ tests have passed -- and i feel the TestPlatformParameterModule is not recognised by bazel in some cases because its visibility is limited to the testing versions of the app?
> 
> Or shall I make some bazel config changes in order to pass the tests? PTAL @BenHenning Thanks!

The PR still needs to be reviewed again (which I can do given the latest changes), but you should be ensuring every CI check is green in this case. The Bazel tests are a 'soft' required in that reviewers enforce they pass, but GitHub does not (it's been this way since they were introduced since they used to be unstable and slow, but I think they may be stable & fast enough now to make them required).

JishnuGoyal:
> error: /tmp/working16590447804925594998/_kotlinc/app-src_test_java_org_oppia_android_app_testing_options_OptionsFragmentTest_lib_kt_jvm/temp/stubs/org/oppia/android/app/testing/options/OptionsFragmentTest.java:141: error: cannot find symbol

From the CI^

I'm aware that the dependencies for OptionsFragmentTest and AdministratorControlsActivityTest are not resolved which is causing the CI to fail. Both these classes depend on TestPlatformParameterModule -- but I'm not able to find a Build.bazel file for these files so i could add this dependency. 

[The CI error also says `unresolved symbol - TestPlatformParameterModule`]

Could you please tell me where I could add the bazel dependencies?

I'll be addressing all the pending threads after my next push.
Thanks!

JishnuGoyal:
PTAL @BenHenning

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -184,8 +182,7 @@ class AdministratorControlsActivityTest {
 
   @Test
   fun testAdministratorControlsFragment_editAccountOptionsEnabled_generalOptionsIsDisplayed() {
-    PlatformParameterModule
-      .forceEnableEditAccountsOptionsUi(true, PlatformParameterForceMode.TESTING)
+    TestPlatformParameterModule.forceEnableEditAccountsOptionsUi(true)
```
BenHenning: Is this needed if it's being set in setUp()?

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -189,6 +191,9 @@ class OptionsFragmentTest {
 
   @Test
   fun testOptionFragment_notFromNavigationDrawer_navigationDrawerIsNotPresent() {
+    TestPlatformParameterModule.forceEnableEditAccountsOptionsUi(
+      ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE
```
BenHenning: Ditto here--it's already being done in setUp.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -337,8 +342,7 @@ class OptionsFragmentTest {
 
   @Test
   fun testOptionsFragment_featureEnabled_appLanguageOptionIsDisplayed() {
-    PlatformParameterModule
-      .forceEnableLanguageSelectionUi(true, PlatformParameterForceMode.TESTING)
+    TestPlatformParameterModule.forceEnableLanguageSelectionUi(true)
```
BenHenning: And here. DItto for all other similar cases in the PR.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
BenHenning: Latest changes seem really clean. Since I have no other concerns, resolving this.

BenHenning (commented):
Thanks @JishnuGoyal. Other than the failing CI & some nits, the PR more or less LGTM. PTAL at the nits.

> > error: /tmp/working16590447804925594998/_kotlinc/app-src_test_java_org_oppia_android_app_testing_options_OptionsFragmentTest_lib_kt_jvm/temp/stubs/org/oppia/android/app/testing/options/OptionsFragmentTest.java:141: error: cannot find symbol
> 
> From the CI^
> 
> I'm aware that the dependencies for OptionsFragmentTest and AdministratorControlsActivityTest are not resolved which is causing the CI to fail. Both these classes depend on TestPlatformParameterModule -- but I'm not able to find a Build.bazel file for these files so i could add this dependency.
> 
> [The CI error also says `unresolved symbol - TestPlatformParameterModule`]
> 
> Could you please tell me where I could add the bazel dependencies?
> 
> I'll be addressing all the pending threads after my next push. Thanks!

You can find the target corresponding to any test by querying it. For example:

```sh
bazel query app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```

Will probably return something like:

```
//app:src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```

Bazel creates 'file' targets for each file, but by simply removing the '.kt' we get the test target name:

```
//app:src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest
```

Which can be passed to ``bazel test`` in the normal way. Note that any test that is part of a top-level target (like ``//app:src/...``) has its shared deps in the respective folder's top-level ``TEST_DEPS`` list, so that's where you need to add new dependencies.

For tests that have their own targets, the query should return the exact target (which will look something like ``//app/src/...``) and will haves its own ``deps`` list to update.

I hope this helps.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
JishnuGoyal: done

JishnuGoyal:
Hello, @BenHenning 

While working on my Gsoc proposal, I came across  `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` which can be used to annotate methods (as well as fields and classes) to restrict their access to **_tests only._**
So I used the previous implementation where I had a companion object inside the `PlatformParameterModule`.

_Why I had to remove the `TestPlatformParameterModule` implementation:_

It required copying all the same feature flags from `PlatformParameterModule` into `TestPlatformParameterModule` --
This is a problem because
1) It broke the CI tests by creating unresolved cyclic dependencies
2) Code duplication


I have also verified that the methods inside `PlatformParameterModule`'s companion object aren't bleeding into production.

PTAL @BenHenning  and let me know how you think of this implementation.

JishnuGoyal:
> Ah, sort of. Looking at the module again, this works because each time the parameter is injected it re-reads the static state. This is technically a violation of the platform parameter module contract as it takes measures to guarantee singleton state among parameters (since changing parameters mid-app could cause undefined behaviors). That being said, it's annoying to do this "properly" as it requires refactoring the injection point in the test to occur after the override (which I've done in a bunch of test suites now). I don't care to push that pattern here even though it may cause some headaches down the line. The preferred way to do this would be to introduce a JUnit rule that can properly set up the parameters ahead of test running (I've built something like this before, and it can be quite nice). This PR introduces the foundations for that work, so I think it'd be a good option for future work (I filed https://github.com/oppia/oppia-android/issues/4302 to jot down some ideas on what such an approach might look like).


Thank you so much @BenHenning for all these insights and guiding me throughout this long PR. It goes without saying how much i got to learn.

I believe you would like this new approach which is not only similar to the previous one but also a little more cleaner. (the new issues you have linked to this PR earlier won't be affected according to me) 

PTAL @BenHenning  and let me know if any further changes are needed from my end!

JishnuGoyal:
Gentle reminder to PTAL @BenHenning

BenHenning:
> Hello, @BenHenning
> 
> While working on my Gsoc proposal, I came across `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` which can be used to annotate methods (as well as fields and classes) to restrict their access to **_tests only._** So I used the previous implementation where I had a companion object inside the `PlatformParameterModule`.
> 
> _Why I had to remove the `TestPlatformParameterModule` implementation:_
> 
> It required copying all the same feature flags from `PlatformParameterModule` into `TestPlatformParameterModule` -- This is a problem because
> 
> 1. It broke the CI tests by creating unresolved cyclic dependencies
> 2. Code duplication
> 
> I have also verified that the methods inside `PlatformParameterModule`'s companion object aren't bleeding into production.
> 
> PTAL @BenHenning and let me know how you think of this implementation.

Apologies for the delayed review @JishnuGoyal. I meant to reply to this last week.

What issue did you run into with (1)? I wouldn't expect there to be a cycle here.

Re: (2), this is actually by design. It's a best practice to duplicate code for testing purposes if it either simplifies the expression/maintenance of the tests, or reduces the amount of test functionality that's exposed to production (which is this case). I think we *do* want to use a test module here, so I want to better understand why that doesn't work.

JishnuGoyal:
> > Hello, @BenHenning
> > While working on my Gsoc proposal, I came across `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` which can be used to annotate methods (as well as fields and classes) to restrict their access to **_tests only._** So I used the previous implementation where I had a companion object inside the `PlatformParameterModule`.
> > _Why I had to remove the `TestPlatformParameterModule` implementation:_
> > It required copying all the same feature flags from `PlatformParameterModule` into `TestPlatformParameterModule` -- This is a problem because
> > 
> > 1. It broke the CI tests by creating unresolved cyclic dependencies
> > 2. Code duplication
> > 
> > I have also verified that the methods inside `PlatformParameterModule`'s companion object aren't bleeding into production.
> > PTAL @BenHenning and let me know how you think of this implementation.
> 
> Apologies for the delayed review @JishnuGoyal. I meant to reply to this last week.
> 
> What issue did you run into with (1)? I wouldn't expect there to be a cycle here.
> 
> Re: (2), this is actually by design. It's a best practice to duplicate code for testing purposes if it either simplifies the expression/maintenance of the tests, or reduces the amount of test functionality that's exposed to production (which is this case). I think we _do_ want to use a test module here, so I want to better understand why that doesn't work.

I will have to check again why the CI broke by pushing that old code again which might be a long process... Meanwhile could you tell why this approach is problematic? I feel this is much cleaner too and is being used in `ExplorationCheckpointController` as well. In fact I had to come up with this one _because_ the older approach was saying something related to 'TestPlatformParameterModule unrecognised' and breaking the CI as far as I remember..

JishnuGoyal:
In my opinion, this approach would be the best solution because it takes care of all the current requirements -- 
better testing of the platform parameters as well as no bleeding of the apis into production. Please let me know your views too.
PTAL @BenHenning

BenHenning:
> > > Hello, @BenHenning
> > > While working on my Gsoc proposal, I came across `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` which can be used to annotate methods (as well as fields and classes) to restrict their access to **_tests only._** So I used the previous implementation where I had a companion object inside the `PlatformParameterModule`.
> > > _Why I had to remove the `TestPlatformParameterModule` implementation:_
> > > It required copying all the same feature flags from `PlatformParameterModule` into `TestPlatformParameterModule` -- This is a problem because
> > > 
> > > 1. It broke the CI tests by creating unresolved cyclic dependencies
> > > 2. Code duplication
> > > 
> > > I have also verified that the methods inside `PlatformParameterModule`'s companion object aren't bleeding into production.
> > > PTAL @BenHenning and let me know how you think of this implementation.
> > 
> > 
> > Apologies for the delayed review @JishnuGoyal. I meant to reply to this last week.
> > What issue did you run into with (1)? I wouldn't expect there to be a cycle here.
> > Re: (2), this is actually by design. It's a best practice to duplicate code for testing purposes if it either simplifies the expression/maintenance of the tests, or reduces the amount of test functionality that's exposed to production (which is this case). I think we _do_ want to use a test module here, so I want to better understand why that doesn't work.
> 
> I will have to check again why the CI broke by pushing that old code again which might be a long process... Meanwhile could you tell why this approach is problematic? I feel this is much cleaner too and is being used in `ExplorationCheckpointController` as well. In fact I had to come up with this one _because_ the older approach was saying something related to 'TestPlatformParameterModule unrecognised' and breaking the CI as far as I remember..

As mentioned, I'd need to see the error to better understand the problem.

For why it's preferable to split out the modules, technically the current implementation *is* bleeding into production code (the ``VisbileForTesting`` annotation doesn't result in a compile-time removal of the code, it just provides context for static analysis tools to issue errors if the contract is ever violated). I don't think Bazel will actually enforce this contract. While we could somewhat easily build our own verification of it, it still has the unfortunate side effect of needing to be stripped out by Proguard/R8 to avoid paying a production penalty for the extra methods.

The other aspect which I didn't mention is that we're likely going to move to a more generic version of this in the future where we use test-only annotations for overriding parameter values. I don't think that solution will work with ``VisibleForTesting`` since test utilities aren't actually tests themselves, so we would need a separate module for the test versions (though the point is slightly moot since I think we actually want to mock out the parameters from underneath the module and converge with one module). The other consideration is that not all parameters are the same in that some are local to the binary and others may eventually connect to remote services. This makes the overriding less obvious in test cases.

All that being said, I think that I am actually okay with this approach as a medium-term solution even though it sort of violates the normal conventions of not exposing production things for testing purposes. It will actually be less work in the long run for us since you're handling the convergence on a single module, so all we need to do in the future is remove these extra APIs and replace them with a more test-friendly (and production-independent) API.

JishnuGoyal:
https://github.com/oppia/oppia-android/runs/6343605385?check_suite_focus=true 

This is the error I was talking about. Although i think it might be related to an un-added Build.bazel file. If so ,where exactly do i add the same?
PTAL @BenHenning 
Thanks!

BenHenning:
Reopening since this was closed due to my inactivity, not @JishnuGoyal's.

BenHenning:
@JishnuGoyal I believe the problem is that you're adding a new module dependency on ``TestPlatformParameterModule`` in both ``OptionsFragmentTest``s, but haven't updated their dependencies to include the library that provides ``TestPlatformParameterModule``. Both tests use this shared test deps list: https://github.com/oppia/oppia-android/blob/107c6ed3955e3b93f745edad836e4cac962b17e5/app/BUILD.bazel#L852
However, they don't include a reference to the library providing ``TestPlatformParameterModule``: https://github.com/oppia/oppia-android/blob/107c6ed3955e3b93f745edad836e4cac962b17e5/testing/src/main/java/org/oppia/android/testing/platformparameter/BUILD.bazel#L25
It looks like the deps list needs to be updated.

---

Note that I found this by looking at the error, seeing that it was Dagger generation that was failing, then checking the changes to the updated test, and, since a new dependency was being added, checked to see if it was in the deps list for that test.

BenHenning:
@JishnuGoyal do you have a plan for wrapping this up? It's actually come up in some recent user feedback, so I think it'd be nice to finish it ahead of our upcoming releases to help mitigate the user confusion that's coming from it being in admin controls.

JishnuGoyal:
PTAL @BenHenning

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
```diff
@@ -71,9 +74,15 @@ class AdministratorControlsViewModel @Inject constructor(
...
-    val itemViewModelList: MutableList<AdministratorControlsItemViewModel> = mutableListOf(
-      AdministratorControlsGeneralViewModel()
-    )
+
+    val itemViewModelList = mutableListOf<AdministratorControlsItemViewModel>()
+
+    if (enableEditAccountsOptionsUi.value) {
+      itemViewModelList.add(
+        AdministratorControlsGeneralViewModel()
+      )
```
BenHenning: Nit: this can fit on one line.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileAndDeviceIdFragmentTest.kt
```diff
@@ -162,6 +153,8 @@ class ProfileAndDeviceIdFragmentTest {
 
   @Before
   fun setUp() {
+    TestPlatformParameterModule.forceEnableEditAccountsOptionsUi(true)
```
BenHenning, line 156: Does this one need to be enabled for this test suite? It seems unrelated.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -141,7 +128,10 @@ class OptionsFragmentTest {
 
   @Before
   fun setUp() {
-    TestModule.forceEnableLanguageSelectionUi = true
+    TestPlatformParameterModule.forceEnableLanguageSelectionUi(true)
+    TestPlatformParameterModule.forceEnableEditAccountsOptionsUi(
+      ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE
```
BenHenning: Perhaps this should be fixed to be the same value everytime?

BenHenning (commented):
Nice @JishnuGoyal! This is looking really nice. I just had a few nit comments, otherwise the PR is really close to 'LGTM'.

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileAndDeviceIdFragmentTest.kt
JishnuGoyal: removed in last commit

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsViewModel.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
JishnuGoyal: removing edit accounts force method since it isn't needed

JishnuGoyal:
PTAL @BenHenning the failing tests seems a flaky

BenHenning (approved):
Thanks @JishnuGoyal. No further comments from me--this LGTM!

I'm re-running the failing Bazel test shard. It sometimes fails when setting up the Android environment unfortunately.

I believe you still need approval from @rt4914 & @anandwana001 so I'm assigning this over to them.

BenHenning:
@rt4914 @anandwana001 PTAL for codeowners.

BenHenning:
(Removing Akshay per recent codeowner changes).

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
```diff
@@ -193,6 +195,68 @@ class AdministratorControlsActivityTest {
     }
   }
 
+  @Test
+  fun testAdministratorControlsFragment_editAccountOptionsEnabled_generalOptionsIsDisplayed() {
+
```
rt4914: Nit: This blank line can be removed.

rt4914 (approved):
One minor non-blocking suggestions. Thanks LGTM

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
JishnuGoyal: done, thanks

JishnuGoyal:
PTAL @BenHenning

app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivityTest.kt
BenHenning: Looks to be done now. Marking as resolved.

BenHenning:
Thanks @JishnuGoyal! Really glad to see this get wrapped up. Merging now!

PR #4217, opened 2022-02-22
Fix #3095: Use protos with intent extras in RecentlyPlayedActivity

### LINKED ISSUE #3095: Use protos with intent extras in RecenlyPlayedActivity
As per [this](https://github.com/oppia/oppia-android/pull/3065#discussion_r610915653) comment and approach we should strive towards using protos in transferring information between activties.

rishidyno:
## Explanation

Fixes  #3095

## Essential Checklist


rishidyno:
@rt4914 @BenHenning  @anandwana001 @vinitamurthi PTAL.

 KDoc missing for files:
/home/runner/work/oppia-android/oppia-android/app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt:10
getting this error, but kDok is already added for `ActivityRouter`.

Also, do I need to write a test file for `ActivityRouter`, or should I add it to `test_file_exemption`?

model/src/main/proto/route.proto
```diff
@@ -0,0 +1,21 @@
...
+
+// Corresponds to intent extras of destination activity.
+message DestinationScreen {
+  oneof destination_screen {
+    RecentlyPlayedActivityIntentExtras recently_played_activity_intent_extras = 1;
+  }
+}
+
+// Intent extras for RecentlyPlayedActivity.
+message RecentlyPlayedActivityIntentExtras {
```
BenHenning: Prefer RecentlyPlayedActivityParams as the nomenclature. Also, these can go into arguments.proto rather than introducing a new proto file (that contains both fragment and activity protos).

app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt
```diff
@@ -0,0 +1,37 @@
+package org.oppia.android.app.activity
+
+import androidx.appcompat.app.AppCompatActivity
+import org.oppia.android.app.home.recentlyplayed.RecentlyPlayedActivity
+import org.oppia.android.app.model.DestinationScreen
+import org.oppia.android.app.model.RecentlyPlayedActivityIntentExtras
+import org.oppia.android.util.logging.ConsoleLogger
+import javax.inject.Inject
+
+class ActivityRouter @Inject constructor(
```
BenHenning: A pattern like this could work, but we'd need to take care to make sure it doesn't depend directly on the activity classes it can route to (otherwise we'll end up with a circular dependency when trying to split this out in Bazel).

To do that, I think we'd need to create a Dagger multibinding map that maps DestinationScreen enums to a new Route interface that returns an intent given a params proto (which can be a generic proto type, i.e. ``T: MessageLite``). That would then enable this router to be fully generic since it would simply inject the map of routes, retrieve the one corresponding to the destination screen, create the intent, pass along the parameters, and route to the resulting intent. To add a new route, one only needs to add a new Dagger binding (which becomes fully plug-and-play when we introduce Dagger Hilt support in the future).

See https://dagger.dev/dev-guide/multibindings.html for how to set up a multibound map with enum keys.

BenHenning (changes requested):
Thanks @rishidyno. This is an interesting approach--I've left some thoughts. Also:

> @rt4914 @BenHenning @anandwana001 @vinitamurthi PTAL.
> 
> KDoc missing for files: /home/runner/work/oppia-android/oppia-android/app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt:10 getting this error, but kDok is already added for `ActivityRouter`.
> 
> Also, do I need to write a test file for `ActivityRouter`, or should I add it to `test_file_exemption`?

ActivityRouter is missing a top-level class KDoc which is also required.

Re: tests, yes you should add them. All new files should have tests unless there are special circumstances to warrant them not having them (such as being an interface with no logic or specific implementations).

rt4914:
@rishidyno Un-assigining until Dagger-multimodule solution is applied and also approach is finalised.

app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt
rishidyno: @BenHenning Do we have similar implementation for dagger multibinding in our application for architectural reference? As, I am new to dagger, will need to do some digging before further progress. Thanks.

rishidyno: @BenHenning Do we need custom map key or should we use `@StringKey()` for different routes.

app/src/main/java/org/oppia/android/app/activity/Route/ActivityRouter.kt
```diff
@@ -1,9 +1,9 @@
-package org.oppia.android.app.activity
+package org.oppia.android.app.activity.Route
```
BenHenning: Per https://developer.android.com/kotlin/style-guide#package_names package names are always lowercase.

app/src/main/java/org/oppia/android/app/activity/ActivityRouter.kt
BenHenning: We don't have an existing example with custom keys @rishidyno, but the Dagger documentation reference above has an example.

Also, I think this is a case where a custom key is needed because we'd want to bind to the enum value generated by proto to register the activity's route implementation.

rishidyno:
I will continue this pr after 20th April.

PR #3852, opened 2021-09-28
Fix #1390: Show terms of service & privacy policy notices.

veena14cs:
## Explanation
Fix #1390

This PR fixes [#1390](https://github.com/oppia/oppia-android/issues/1390). This PR introduces Privacy Policy and Terms of Service in the app. The PR contains Policies page through onboarding flow and FAQ/Help screens.

This PR also fixes [#1022](https://github.com/oppia/oppia-android/issues/1022). This PR introduces two new custom tags `<oppia-ul>` and `<oppia-ol>` to support html `<ul>` and `<ol>` tags respectively. To addon the new custom list tags also supports lower versions devices and nested list tags. 

This PR also fixes [#2193](https://github.com/oppia/oppia-android/issues/2193) RTL issues in Bullet points and fixes the crashes in the issue [#2391](https://github.com/oppia/oppia-android/issues/2391)

## Mock Links

Onboarding 
[Mobile Landscape](https://xd.adobe.com/view/ee9e607b-dbd6-4372-48dc-b687d32af3da-98af/screen/fa730345-6bb7-4740-a967-2618919d7f48/)
[Tablet](https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/ed85c1d0-4e3f-47cf-a2f7-4b0779a7c8bc/specs/)
FAQ and Help
[Mobile Landscape](https://xd.adobe.com/view/ee9e607b-dbd6-4372-48dc-b687d32af3da-98af/screen/58e68418-f8f5-4f94-b8b0-1d9cac0c2f12/)
[Tablet](https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/2c17acd0-5d5e-4cb2-be6b-1bf7af8bdf5a/specs/)

## Document Reference
https://docs.google.com/document/d/1QoW2S-HGNrQ_6rodOByjGPd8e_w80PHKE3PJ82t1-hE/edit?usp=sharing

## Screenshots
**Mobile Portrait on Nexus 6**

Screenshot_1634630041.....Screenshot_1634949081...Screenshot_1634630033......Screenshot_1634949088

Screenshot_1637266595.....Screenshot_1637266617

Screenshot_1637266577.....Screenshot_1637266625

Screenshot_1634630045.....Screenshot_1634949092

Screenshot_1634948852......Screenshot_1634949016

Screenshot_1634948847....Screenshot_1634949022

Screenshot_1634948856.....Screenshot_1634949010

[...]

PR #4218, opened 2022-02-24
Fix #3906: Incorrect ouput for rich text in hints and solution [Blocked: #3852]

### LINKED ISSUE #3906: [A11Y] Incorrect output for rich-text in Hints and Solution
Incorrect output for rich-text in Hints and Solution.

vrajdesai78:
## Explanation
Fix #3906:  Parsed HTML to normal text while setting Content Description for Hints and Solution. 

## Essential Checklist

https://user-images.githubusercontent.com/43074241/155514660-2daadc00-a73d-4d51-a702-b52ed341d09a.mp4

### DIFF
--- a/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
+++ b/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
@@ -1,8 +1,8 @@
 package org.oppia.android.app.hintsandsolution
 
-import androidx.core.text.HtmlCompat
 import androidx.databinding.ObservableField
 import androidx.lifecycle.ViewModel
+import javax.inject.Inject
 import org.oppia.android.R
 import org.oppia.android.app.fragment.FragmentScope
 import org.oppia.android.app.model.HelpIndex
@@ -13,7 +13,9 @@ import org.oppia.android.app.translation.AppLanguageResourceHandler
 import org.oppia.android.domain.hintsandsolution.isHintRevealed
 import org.oppia.android.domain.hintsandsolution.isSolutionRevealed
 import org.oppia.android.domain.translation.TranslationController
-import javax.inject.Inject
+import org.oppia.android.util.gcsresource.DefaultResourceBucketName
+import org.oppia.android.util.parser.html.ExplorationHtmlParserEntityType
+import org.oppia.android.util.parser.html.HtmlParser
 
 /**
  * RecyclerView items are 2 times of (No. of Hints + Solution),
@@ -28,7 +30,10 @@ private const val DEFAULT_HINT_AND_SOLUTION_SUMMARY = ""
 @FragmentScope
 class HintsViewModel @Inject constructor(
   private val resourceHandler: AppLanguageResourceHandler,
-  private val translationController: TranslationController
+  private val translationController: TranslationController,
+  private val htmlParserFactory: HtmlParser.Factory,
+  @DefaultResourceBucketName private val resourceBucketName: String,
+  @ExplorationHtmlParserEntityType private val entityType: String
 ) : HintsAndSolutionItemViewModel() {
 
   val newAvailableHintIndex = ObservableField<Int>(-1)
@@ -97,17 +102,34 @@ class HintsViewModel @Inject constructor(
   }
 
   fun computeHintListDropDownIconContentDescription(): String {
+    val hintsViewModel = HintsViewModel(
+      resourceHandler,
+      translationController,
+      htmlParserFactory,
+      resourceBucketName,
+      entityType
+    )
     return resourceHandler.getStringInLocaleWithWrapping(
       R.string.show_hide_hint_list,
-      HtmlCompat.fromHtml(
-        hintsAndSolutionSummary.get() ?: DEFAULT_HINT_AND_SOLUTION_SUMMARY,
-        HtmlCompat.FROM_HTML_MODE_LEGACY
+      htmlParserFactory.create(
+        resourceBucketName,
+        entityType,
+        hintsViewModel.explorationId.get()!!,
+        /* imageCenterAlign= */ true
+      ).parseOppiaHtml(
+        hintsAndSolutionSummary.get() ?: DEFAULT_HINT_AND_SOLUTION_SUMMARY
       )
     )
   }
 
   private fun addHintToList(hintIndex: Int, hint: Hint) {
-    val hintsViewModel = HintsViewModel(resourceHandler, translationController)
+    val hintsViewModel = HintsViewModel(
+      resourceHandler,
+      translationController,
+      htmlParserFactory,
+      resourceBucketName,
+      entityType
+    )
     hintsViewModel.title.set(hint.hintContent.contentId)
     val hintContentHtml =
       translationController.extractString(hint.hintContent, writtenTranslationContext)
--- a/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
+++ b/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
@@ -1,5 +1,6 @@
 package org.oppia.android.app.hintsandsolution
 
+import androidx.core.text.HtmlCompat
 import androidx.databinding.ObservableField
 import androidx.lifecycle.ViewModel
 import org.oppia.android.R
@@ -98,7 +99,10 @@ class HintsViewModel @Inject constructor(
   fun computeHintListDropDownIconContentDescription(): String {
     return resourceHandler.getStringInLocaleWithWrapping(
       R.string.show_hide_hint_list,
-      hintsAndSolutionSummary.get() ?: DEFAULT_HINT_AND_SOLUTION_SUMMARY
+      HtmlCompat.fromHtml(
+        hintsAndSolutionSummary.get() ?: DEFAULT_HINT_AND_SOLUTION_SUMMARY,
+        HtmlCompat.FROM_HTML_MODE_LEGACY
+      )
     )
   }
 
--- a/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
+++ b/app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
@@ -2,7 +2,6 @@ package org.oppia.android.app.hintsandsolution
 
 import androidx.databinding.ObservableField
 import androidx.lifecycle.ViewModel
-import javax.inject.Inject
 import org.oppia.android.R
 import org.oppia.android.app.fragment.FragmentScope
 import org.oppia.android.app.model.HelpIndex
@@ -16,6 +15,7 @@ import org.oppia.android.domain.translation.TranslationController
 import org.oppia.android.util.gcsresource.DefaultResourceBucketName
 import org.oppia.android.util.parser.html.ExplorationHtmlParserEntityType
 import org.oppia.android.util.parser.html.HtmlParser
+import javax.inject.Inject
 
 /**
  * RecyclerView items are 2 times of (No. of Hints + Solution),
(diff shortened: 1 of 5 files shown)


vrajdesai78:
@BenHenning I figured out that there were only a few cases (I am able to find only 1) where HTML is not parsing properly for content description before changing this code. How can I approach to write a test case for this issue?

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
```diff
@@ -98,7 +99,10 @@ class HintsViewModel @Inject constructor(
   fun computeHintListDropDownIconContentDescription(): String {
     return resourceHandler.getStringInLocaleWithWrapping(
       R.string.show_hide_hint_list,
-      hintsAndSolutionSummary.get() ?: DEFAULT_HINT_AND_SOLUTION_SUMMARY
+      HtmlCompat.fromHtml(
```
BenHenning: We should only be using our own HtmlParser when parsing HTML to ensure consistency and support for custom tags.

I think this will require doing something similar to the parseOppiaHtml call in HintsAndSolutionDialogFragmentPresenter, though you'll need to update parseOppiaHtml itself to support not taking a TextView (for cases when we just want to parse into spannables without tweaking the textview itself).

BenHenning (changes requested):
Apologies for the delayed review. Left one comment--PTAL.

BenHenning:
> @BenHenning I figured out that there were only a few cases (I am able to find only 1) where HTML is not parsing properly for content description before changing this code. How can I approach to write a test case for this issue?

@vrajdesai78 you would need to write a test in the suite corresponding to the view model (it may actually be a test suite for a fragment or activity that indirectly uses the view model), load the content that fails, and verify that it now has the correct content description. You should verify that the test also fails without the fix in place to make sure that it's correctly passing when it passes with the fix.

rt4914:
@vrajdesai78 As per our discussion earlier the most important thing in this PR is writing test which I am not sure about. So once you have written it as per Ben's comment please let me know and I will do a pass on it.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
vrajdesai78: I have updated parseOppiaHtml and related code.

vrajdesai78:
> > @BenHenning I figured out that there were only a few cases (I am able to find only 1) where HTML is not parsing properly for content description before changing this code. How can I approach to write a test case for this issue?
> 
> @vrajdesai78 you would need to write a test in the suite corresponding to the view model (it may actually be a test suite for a fragment or activity that indirectly uses the view model), load the content that fails, and verify that it now has the correct content description. You should verify that the test also fails without the fix in place to make sure that it's correctly passing when it passes with the fix.

I have added a test case `testStateFragment_showHintsAndSolution_hasCorrectContentDescription` in StateFragmentTest. In this test I am checking that hint has correct content description.

Test case failing if HTML is not parsed for content description (Both espresso and robolectric).
Screenshot from 2022-02-26 21-09-58

Test case passing after parsing HTML properly (On both espresso and robolectric).
Screenshot from 2022-02-27 00-12-33

@BenHenning PTAL.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
```diff
@@ -28,7 +30,10 @@ private const val DEFAULT_HINT_AND_SOLUTION_SUMMARY = ""
 @FragmentScope
 class HintsViewModel @Inject constructor(
   private val resourceHandler: AppLanguageResourceHandler,
-  private val translationController: TranslationController
+  private val translationController: TranslationController,
+  private val htmlParserFactory: HtmlParser.Factory,
+  @DefaultResourceBucketName private val resourceBucketName: String,
+  @ExplorationHtmlParserEntityType private val entityType: String
```
BenHenning, line 36: I think this changes based on whether an exploration or question is being loaded. We probably need to pass in the entity type from the fragment (which I expect we must do for the content itself since hints can contain images).

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -1116,6 +1117,36 @@ class StateFragmentTest {
     }
   }
 
+  @Test
+  fun testStateFragment_showHintsAndSolution_hasCorrectContentDescription() {
```
BenHenning: This test probably should be in StateFragmentLocalTest (we generally don't test hints here because we sometimes would need to wait seconds for them to show up).

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -196,16 +196,22 @@ class CustomHtmlContentHandler private constructor(
...
       // tracking.
+      if (imageRetriever != null) {
+        return HtmlCompat.fromHtml(
+          "<init-custom-handler/>$html",
+          HtmlCompat.FROM_HTML_MODE_LEGACY,
+          imageRetriever,
+          CustomHtmlContentHandler(customTagHandlers, imageRetriever),
+        ) as Spannable
+      }
       return HtmlCompat.fromHtml(
```
BenHenning: Could you consolidate this and the above into one call, instead?

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -196,16 +196,22 @@ class CustomHtmlContentHandler private constructor(
      */
     fun <T> fromHtml(
       html: String,
-      imageRetriever: T,
+      imageRetriever: T?,
```
BenHenning: I think it's fine to have a nullable property for a helper, but for the public API I suggest making things more explicit by having 1 version that takes a non-null image retriever and another which takes nothing but the HTML.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -46,24 +46,16 @@ class HtmlParser private constructor(
    */
   fun parseOppiaHtml(
     rawString: String,
-    htmlContentTextView: TextView,
+    htmlContentTextView: TextView? = null,
```
BenHenning: Ditto here--we ought to have two versions of the function to keep things clearer. I suggest using a private helper to consolidate the two (and that can certainly use a nullable parameter to make things simpler).

BenHenning (commented):
Thanks @vrajdesai78. This is definitely on the right track. Besides the comments I left, please also:
- Add tests for the content handler and html parser changes
- Make sure all CI checks are passing (a bunch are currently failing)

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
vrajdesai78: @BenHenning Done

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
vrajdesai78: @BenHenning Done

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
vrajdesai78: Done

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
vrajdesai78: I have created another version of parseOppiaHtml which takes only rawString as a parameter, but I didn't how can I use a private helper.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
vrajdesai78: I am not able to understand that how can I use entity type from fragment. Can you elaborate a little bit. Thanks @BenHenning

vrajdesai78:
@BenHenning can you give me a hint that why CI checks are failing and how can I debug it. Thanks

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -1306,6 +1307,31 @@ class StateFragmentLocalTest {
...
+      submitFractionAnswer(answerText = "1/4")
+
+      // Reveal the hint.
+      openHintsAndSolutionsDialog()
+      pressRevealHintButton(hintPosition = 0)
+
+      // checking content description of hint
+      onView(withId(R.id.hint_list_drop_down_icon)).check(
+        matches(
+          ViewMatchers.withContentDescription(
```
BenHenning, line 1326: Nit: this should be directly imported (like other Espresso functions).

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -1306,6 +1307,31 @@ class StateFragmentLocalTest {
...
+      clickContinueButton()
+      // Submit two incorrect answers.
+      submitFractionAnswer(answerText = "1/3")
+      submitFractionAnswer(answerText = "1/4")
+
+      // Reveal the hint.
+      openHintsAndSolutionsDialog()
+      pressRevealHintButton(hintPosition = 0)
+
+      // checking content description of hint
```
BenHenning, line 1323:
```suggestion
      // Ensure the hint description is correct and doesn't contain any HTML.
```

Fixes capitalization & punctuation, but it also rewords to be a bit clearer.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -208,6 +208,12 @@ class CustomHtmlContentHandler private constructor(
         CustomHtmlContentHandler(customTagHandlers, imageRetriever),
       ) as Spannable
     }
+
+    fun fromHtml(html: String): Spannable {
+      return HtmlCompat.fromHtml(
+        "<init-custom-handler/>$html", HtmlCompat.FROM_HTML_MODE_LEGACY,
```
BenHenning, line 214: This should still be using CustomContentHandler.

It might actually be worth waiting until #3852 is merged since it's introducing a very similar function with tests which might make this a lot easier and avoid redundant work.

BenHenning (commented):
Thanks @vrajdesai78. After looking at this again, I think it might actually be worth waiting until #3852 is finished since it's introducing the exact HtmlParser functionality that you need (i.e. parsing HTML without including an image provider).

BenHenning:
> @BenHenning can you give me a hint that why CI checks are failing and how can I debug it. Thanks

@vrajdesai78 these are Bazel failures that are coming up because you're changing a view model to depend on HtmlParser, and the view_models library doesn't include the needed HtmlParser dependency in its deps list:

https://github.com/oppia/oppia-android/blob/2004db709549619cca811142b627654cc3e29a92/app/BUILD.bazel#L632

The deps list will need to be updated to include it.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
vrajdesai78: Ok, then I will wait for #3852 to get merged and after that I will do required changes.

vrajdesai78:
Keep this PR open

vrajdesai78:
Keep this PR open

PR #4219, opened 2022-02-24
Fix part of #3245 : Enabling AccessibilityChecks for StoryFragmentTest

bhaktideshmukh:
## Explanation

Fix part of #3245  : Enable AccessibilityChecks for StoryFragmentTest
Removed the @DisableAccessibilityChecks Annotation from the StoryFragmentTest

After removing @DisableAccessibilityChecks Annotation two tests were failing :--
1. testStoryFragment_clickOnToolbarNavigationButton_closeActivity
2. testStoryFragment_changeConfiguration_explorationStartCorrectly()

After replacing ClickableSpan with URLSpan for better accessibility and fixing minimum width height for image view , both the tests are passing successfully.

Screenshot from 2022-02-24 22-35-50

Screenshot from 2022-02-24 22-37-28

**Screenshots**

| Before  | After|
|-------|---|
|||

| Before  | After|
|-------|---|
|||

## Essential Checklist

## For UI-specific PRs only


Rohit1173:
Once check the Logs of the mentioned tests before making changes i.e., just remove the annotation and run the tests
From that, I don't think we should make changes in the xml file, as the output in the log just suggests `URLSpan should be used in place of ClickableSpan for improved accessibility `.

bhaktideshmukh:
> Once check the Logs of the mentioned tests before making changes i.e., just remove the annotation and run the tests From that, I don't think we should make changes in the xml file, as the output in the log just suggests `URLSpan should be used in place of ClickableSpan for improved accessibility `.

2 accessibility errors were found:--
first was to replace ClickableSpan with URLSpan for improved accessibility.
Second was view falls below the minimum recommended size for touch targets and thus I made the changes in xml files.

Screenshot from 2022-02-26 15-16-10

bhaktideshmukh:
PTAL @Rohit1173 
Thanks.

Rohit1173:
Ok , I have checked and yes there were 2 different for 2 tests.
So as there were UI changes in the PR you need to put the Before and After Screenshots of that particular screen on which you have made changes
Other than that, everything seems fine.
Thanks.

app/src/main/java/org/oppia/android/app/story/StoryFragmentPresenter.kt
```diff
@@ -183,7 +183,8 @@ class StoryFragmentPresenter @Inject constructor(
               storyItemViewModel.missingPrerequisiteChapter.name
             )
             val chapterLockedSpannable = SpannableString(missingPrerequisiteSummary)
-            val clickableSpan = object : ClickableSpan() {
+            val url = "https://developer.android.com"
```
vrajdesai78, line 186: Why you have added this specific URL for urlspan

bhaktideshmukh:
> Ok , I have checked and yes there were 2 different for 2 tests. So as there were UI changes in the PR you need to put the Before and After Screenshots of that particular screen on which you have made changes Other than that, everything seems fine. Thanks.

Done!!

app/src/main/java/org/oppia/android/app/story/StoryFragmentPresenter.kt
bhaktideshmukh: > Why you have added this specific URL for urlspan

It was mandatory to pass the parameter url in URLSpan() and it cannot be kept null hence I added that.

vrajdesai78: Ok, so instead of passing this URL you can pass an empty string

bhaktideshmukh: > Ok, so instead of passing this URL you can pass an empty string

Keeping string as empty is again giving accessibility error stating -- URLSpan should not contain relative links.

app/src/main/res/layout/story_chapter_view.xml
```diff
@@ -108,6 +108,8 @@
             android:layout_marginTop="8dp"
             android:layout_marginEnd="8dp"
             android:layout_marginBottom="4dp"
```
rt4914: All all the views where you have used `minHeight` / `minWidth`, if that view contains "margin", shift that to padding.
For example for this view you can remove
```
            android:layout_marginStart="8dp"
            android:layout_marginTop="8dp"
```

and replace with
```
            android:paddingStart="8dp"
            android:paddingTop="8dp"
            android:paddingEnd="8dp"
            android:paddingBottom="4dp"
```

app/src/main/java/org/oppia/android/app/story/StoryFragmentPresenter.kt
rt4914: @BenHenning Any thoughts on this?

rt4914 (changes requested):
@bhaktideshmukh PTAL

Also can you confirm that the click functionality is still working correctly? i.e. if user clicks on the text (URLSpan) it does work as expected.

rt4914:
@BenHenning PTAL at https://github.com/oppia/oppia-android/pull/4219/files#r815306442

bhaktideshmukh:
> @bhaktideshmukh PTAL
> 
> Also can you confirm that the click functionality is still working correctly? i.e. if user clicks on the text (URLSpan) it does work as expected.

Yes it is working as expected.

app/src/main/res/layout/story_chapter_view.xml
bhaktideshmukh: Done

app/src/main/java/org/oppia/android/app/story/StoryFragmentPresenter.kt
BenHenning: We shouldn't be defaulting a URL in production code like this. @bhaktideshmukh can you clarify why a URLSpan is the correct solution here vs. using a ClickableSpan? We use the latter throughout the app, so I think we need to understand how to make ClickableSpan TalkBack compatible rather than replacing it with something else that's meant for a different purpose.

BenHenning (changes requested):
Followed up on one comment thread.

app/src/main/java/org/oppia/android/app/story/StoryFragmentPresenter.kt
bhaktideshmukh: > We shouldn't be defaulting a URL in production code like this. @bhaktideshmukh can you clarify why a URLSpan is the correct solution here vs. using a ClickableSpan? We use the latter throughout the app, so I think we need to understand how to make ClickableSpan TalkBack compatible rather than replacing it with something else that's meant for a different purpose.

When in accessibility mode talkback won't read spannable string as a clickable object,therefore the user won't know there's a link to click on. maybe that's why URLSpan is preferred over ClickableSpan.

I tried one solution for handling accessibility for ClickableSpan :--
added 'ViewCompat.enableAccessibleClickableSpanSupport(view)' in StoryFragmentPresenter.kt file but it didn't worked.

Now what should be done next?

BenHenning: This is actually documented in detail in https://support.google.com/accessibility/android/answer/6378148?hl=en.

Fixing this properly (i.e. using a URLSpan with an IntentFilter is going to be a lot more work than this PR is currently doing), and honestly probably needs an investigation by itself to even see what the a11y experience would be for those users before we spend the time supporting them.

At a high-level, this would require implementing a generic router to receive intent filters for all ClickableSpans, and find a way to call back into the correct place to handle the clicks. I'm not actually 100% sure how this would work since the intent filter needs to point to an activity, broadcast receiver, or service, none of which individually would allow us to solve this generically since they don't have access to the callbacks to handle the logic (some one-off cases like the one fixed by this PR could maybe work by intenting back to the same activity, but we'd need to be careful about manually handling the intent and this doesn't solve the generic cases like concept cards).

@rt4914 this is actually a potentially large project--what are your thoughts? I'm not actually sure we have any way to proceed with ClickableSpan, but I'm also not sure how to make this work for URLSpan based on the above issues.

rt4914: @BenHenning Apologies. This was my mistake that I assumed that this solution can work directly. Clearly it needs more investigation before finalising the solution. I will close this PR for now and update the issue with further details and the maybe focus on a demo or take help from @vrajdesai78 and @bhaktideshmukh here.

rt4914:
@bhaktideshmukh Closing this w.r.t [this comment](https://github.com/oppia/oppia-android/pull/4219#discussion_r820991182) and issue #4235 

You can still work on it but as per the issue description.

Issue #4236, opened 2022-03-08
Add remaining tests for AsyncResultSubject.


BenHenning:
### Is your feature request related to a problem? Please describe.

The class [AsyncResultSubject](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/testing/src/main/java/org/oppia/android/testing/data/AsyncResultSubject.kt#L28) (in `org.oppia.android.testing.data`) provides custom Truth subjects for verifying properties of `AsyncResult`. While a good number of tests are already written in [AsyncResultSubjectTest](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/testing/src/test/java/org/oppia/android/testing/data/AsyncResultSubjectTest.kt#L14), not all methods and edge cases are covered. This results in partial test coverage and missed validation of important behaviors.

### Describe the solution you'd like

We should expand `AsyncResultSubjectTest` to ensure **all public methods in `AsyncResultSubject` are thoroughly tested**, covering both success and failure paths. This will help guarantee correctness, improve reliability, and prevent regressions in future changes.

### Tasks / Checklist

Add new tests for the following scenarios:

1. **`isSuccessThat()`**: Already partially tested with strings, but add more tests for other types (e.g., nullable values, mismatched casts).
2. **`isComparableSuccessThat()`**: Verify with comparable types (e.g., `Int`, `String`) including correct ordering and type mismatch failures.
3. **`isProtoSuccessThat()`**: Add tests using a simple `MessageLite` proto to confirm correct proto assertions.
4. **`isFailureThat()`**: More validation on error chains, custom exceptions, and verifying root causes.
5.  **`isNewerOrSameAgeAs()` and `isOlderThan()`**: Add tests to validate correct age comparison between different `AsyncResult` instances.
6.  **`hasSameEffectiveValueAs()`**: Add more edge case tests (e.g., comparing pending vs failure, null vs non-null values).
7.  **Type mismatch scenarios**: Expand coverage to ensure incorrect type extraction (`extractSuccessValue`, `isIntSuccessThat`, etc.) throws the expected assertion error.

### Acceptance Criteria

* Each of the above methods should have **at least one dedicated test** verifying correct behavior.
* Negative test cases (incorrect types, mismatched expectations) should also be included.
* Ensure full branch and statement coverage for `AsyncResultSubject`.
[...]

PR #4237, opened 2022-03-08
Fix #3813, #92, part of #4044: Refactor AsyncResult into a sealed class

### LINKED ISSUE #92: Introduce custom Truth subject for AsyncResult
If AsyncResult from #87 sticks around long-term, we should introduce a custom Truth subject to simplify testing these results. Right now, testing them is a bit awkward:

```kotlin
// Pending check
assertThat(result.isPending()).isTrue()

// Success check
assertThat(result.isSuccess()).isTrue()
assertThat(result.getOrThrow()).isEqualTo(expected)

// Failure check
assertThat(result.isFailure()).isTrue()
assertThat(result.getFailureOrNull()).hasCauseThat()...
```

These could be largely simplified using a custom subject:

```kotlin
// Pending check
assertThat(result).isPending()

// Success check (two options depending on situation)
assertThat(result).isSuccess()
assertThat(result).hasValue(expected) // Also checks isSuccess()

// Failure check (four options depending on situation)
assertThat(result).isFailure()
assertThat(result).withFailure().isInstanceOf(...) // also checks isFailure()
assertThat(result).withFailure().hasCauseThat()... // also checks isFailure()
assertThat(result).withFailure().hasMessageThat()... // also checks isFailure()

// Completion check
assertThat(result).isCompleted()
```

### LINKED ISSUE #3813: Migrate all data provider tests over to using DataProviderTestMonitor
#3794 is introducing a new DataProviderTestMonitor test utility that cuts down a lot of the boilerplate needed to test data providers. All existing tests that don't use this monitor for testing data providers should be updated to use it. #3794 also includes two new test suites which make extensive use of the utility (LocaleControllerTest & TranslationControllerTest).

BenHenning:
## Explanation
Fix #3813
Fix #92
Fix part of #4044 (see below for how this relates to the broader math expressions project)

This PR refactors ``AsyncResult`` to be a sealed class rather than using enums which has a distinct advantage: the strong typing guarantees from a sealed class ensure the presence of certain data is guaranteed (such as the ``Throwable`` from the failing ``AsyncResult``). However, this refactor is very far-reaching since essentially every usage of ``AsyncResult`` needed to be updated (though not necessarily just references).

### High-level ``AsyncResult`` pattern changes
Here are comparisons in how the common operations have changed:

Before:
```kotlin
val result: AsyncResult<String> = fetchOrReceiveSomeResult()

// Construction.
val pending = AsyncResult.pending<String>()
val failure = AsyncResult.failure<String>(IllegalStateException("Failure reason"))
val success = AsyncResult.success("success value")

// Check type.
val isPending = result.isPending()
val isFailure = result.isFailure()
val isSuccess = result.isSuccess()
val isCompleted = result.isCompleted()

// Retrieve success values.
val successValue1 = result.getOrDefault("default string")
val successValue2 = result.getOrThrow()

// Retrieve failure reason.
val failureReason = result.getErrorOrNull()

// Generic type check.
when (result) {
  // We typically ignore the pending case.
  result.isFailure() -> { /* Do something with result.getErrorOrNull() */ }
  result.isSuccess() -> { /* Do something with result.getOrDefault() or result.getOrThrow() */ }
  // The 'else' case is possible, and often not implemented.
}
```

After:
```kotlin
val result: AsyncResult<String> = fetchOrReceiveSomeResult()

// Construction.
val pending = AsyncResult.Pending<String>()
val failure = AsyncResult.Failure<String>(IllegalStateException("Failure reason"))
val success = AsyncResult.Success("success value")

// Check type.
val isPending = result is AsyncResult.Pending
val isFailure = result is AsyncResult.Failure
val isSuccess = result is AsyncResult.Success
val isCompleted1 = result !is AsyncResult.Pending
val isCompleted2 = result is AsyncResult.Failure || result is AsyncResult.Success

// Retrieve success values.
val success1 = success.value // Type is 'String'
val success2 = (result as? AsyncResult.Success)?.value // Type is 'String?' conditioned on if the result is a success.

// Retrieve failure reason.
val error1 = failure.error // Type is 'Throwable'
val error2 = (result as? AsyncResult.Failure)?.error // Type is 'Throwable?' conditioned on if the result is a failure.

// Generic type check.
when (result) {
  is AsyncResult.Pending { /* Do something in the pending case, such as showing a loading indicator. */ }
  is AsyncResult.Failure { /* Do something with result.error which is never null. */ }
  is AsyncResult.Success { /* Do something with result.value which is always defined. */ }
  // 'else' can't exist since all cases are exhausted, and Kotlin will eventually produce an error if sealed class 'when's aren't exhaustive.
}

// Note 'getOrThrow' is no longer needed since 'value' can only be retrieved if the result is confirmed to be a success.
// Similarly, 'getErrorOrNull' isn't needed since the error can similarly only be extracted from a Failure result.
// 'getOrDefault' is still possible using patterns as so (where the second is preferred to ensure all cases are considered):
val defaultExample1 = if (result is AsyncResult.Success) result.value else default
val defaultExample2 = when (result) {
  // Or, split out 'failure' to log an error message.
  is AsyncResult.Pending, is AsyncResult.Failure -> default
  is AsyncResult.Success -> result.value
}

// Note that result transformations haven't changed at all.
```

The main advantages of the new approach:
- All type cases must always be considered when using 'when' with the result (forcing consideration of both error and pending cases--this is useful since we often ignore pending cases today despite them existing and occurring)
- There's no longer any question about the presence of success values or failure error throwables (they are always present if the corresponding result is the correct type) which results in safer, more readable, and shorter code
- The implementation properly allows for null success values (see 'rationale' section below)

The only disadvantage is that some generic cases will conflate ``AsyncResult``'s sub-types, such as this case:

```kotlin
val results1 = mutableListOf(AsyncResult.Pending())
// results1 is of type MutableList<AsyncResult.Pending>, but this can be fixed by being explicit:
val results2 = mutableListOf<AsyncResult<String>>(AsyncResult.Pending())
// Or:
val results3: MutableList<AsyncResult<String>> = mutableListOf(AsyncResult.Pending())
```

### Rationale for why this PR is needed & connection to algebraic expressions
#4239 introduces changes to ``QuestionAssessmentProgressController`` and ``ExplorationProgressController`` such that data providers with null payloads are transformed into others. However, that caused an issue when transforming results since 'null' can mean either the absence of 'success' (i.e. a failure) or a successful null value, and ``AsyncResult`` couldn't tell the difference between these without leveraging a sealed class.

See #4239 for why it's necessary for the broader algebraic expressions work (which is the same justification as why this PR is). This PR exists to make #4239 smaller since it brought in some additional refactoring & cleanup beyond the minimal migration to a sealed class. Furthermore, this included in the algebraic project chain for the same reason as #4239: to simplify merging and reviews.

### Tests migration and ``DataProviderMonitorFactory``
Rather than updating all tests which verified ``DataProvider`` results to use the new ``AsyncResultSubject``, it seemed nicer to migrate them over to ``DataProviderMonitorFactory`` (which effectively fully fixed #3813). This resulted in a nice simplification in a lot of test suites by eliminating Mockito usage entirely (except in the monitor's implementation). I also replaced all ``toLiveData``-esque calls to use ``DataProvider``s, instead (which required changing a bunch of controllers over to using ``DataProvider``s instead of ``LiveData`` (which is the preferred approach since ``LiveData`` should never be used in the domain layer to ensure good threading cohesion).

### Test exemptions & changes
Note that there's a new ``AsyncResultSubject`` being introduced here to simplify testing (which consequently fixes #92), but tests aren't added for this subject for similar reasons to past algebraic expression PRs. #4236 was filed to track adding tests for this subject in the future. All existing cases that weren't migrated as part of the monitor factory refactor were moved over to using this new subject (except those explicitly testing ``AsyncResult``).

A bunch of tests have been removed since they are no longer possible due to compile-time guarantees about the value of ``AsyncResult`` in different scenarios.

``StateFragmentTest`` was updated to account for the content description for incorrect answers to not be computed correctly (see below section for details).

### Changes to ``StateRecyclerViewAssembler`` & content descriptions
There was a bug in ``StateRecyclerViewAssembler`` where an answer wouldn't be indicated as incorrect correctly in explorations. I fixed this by introducing a more explicit 'correct_answer' signal in the answer response proto. This fix was needed since one of ``StateFragmentTest``'s tests started failing when another issue was fixed in ``ExplorationProgressControllerTest``: in an old PR, one of the notifications for the current state ``DataProvider`` was dropped mid-answer submission. Restoring this notification triggered the test to correctly fail (due to it actually hitting the bug in the assembler; before, the content description wasn't being correctly defined in that test and it was falsely passing). The fix to the assmbler caused the test to start passing (but another to fail due to it not being supported on Robolectric; it's been subsequently ignored like others of its kind).

The new fix is intentionally designed to be robust against cases when an answer doesn't have a correct label (since that was assumed to be present before).

## Essential Checklist

## For UI-specific PRs only
N/A -- This should be a non-side effect change. Despite the UI changes, it shouldn't actually change user behaviors (possibly with the exception of answer notification being a bit faster which led to the UI test fix that was explained above) with one primary exception: this fixes the content description computation for incorrect answers. **Reviewers:** please let me know if you'd like this demonstrated, and how. I also didn't explicitly verify that affected shared tests pass on Espresso, but I highly doubt they wouldn't (or at least, would change their current pass/fail status) given that their Robolectric counterparts pass and this is a non-functional change.

I did not verify StateFragmentTest on Espresso since it's currently broken (see #4238). This is fixed in the next PR where the test suite is verified to pass.


BenHenning:
@vinitamurthi & @rt4914 PTAL for codeowners (**Rajat**: please look at the accessibility bits in the PR description & code specifically in case you have any additional concerns and requests).

@anandwana001 PTAL as part of the broader algebraic expressions project.

vinitamurthi (approved):
LGTM as codeowner

anandwana001 (approved):
LGTM

BenHenning:
So unexpectedly, this PR introduced a regression that broke exploration routing due to a change in behavior on how ``AsyncResult``s are compared for equality. Fortunately, it was easy to fix this once I understood what the problem is. However, it was unexpected for this refactor to have a user impact.

@anandwana001 could you PTAL at https://github.com/oppia/oppia-android/pull/4237/commits/06b51d3e65a78090a3c53c42b39f1bd948916f0c (the fix commit) as a sanity check? It'd be good to have another pair of eyes on this given that it broke a bunch of data provider subscriptions before the fix was in place.

anandwana001 (approved):
LGTM

BenHenning:
Thanks @anandwana001!

rt4914 (approved):
LGTM, thanks.

PR #4230, opened 2022-03-02
Fix part of #3861: Submit-time errors permanently break the answer submission flow

anandwana001:
## Explanation

Fix a part of #3861 

### My Observation while fixing this issue
1. In `FractionInteractionViewModel` using text watcher we keep track of real-time errors at function `checkPendingAnswerError`
2. On submit button click, we again come to `checkPendingAnswerError` with a submit-time check.
3. In function `getSubmitTimeError` when we call `parseFraction` using the input string `1/` it returns null.

### Cause
1. `parseFraction` returns null, which make `pendingAnswerError` null.
2. Now, when we try to get the pending answer it creates a new user instance for us, which only consists of `contentDescription`.
```proto
private UserAnswer() {
    contentDescription_ = "";
  }
```
3. Now, with this incomplete proto, we try to access `answer`, and `writtenTranslationContext` from our controller, which leads to falling under the `finally` case in try/catch handle where we use an uninitialized variable to access.
https://github.com/oppia/oppia-android/blob/dd024327d32d4c8edcbf12fe31d9df97f7f71a4f/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt#L180

### Fix
1. Add an additional check `normalized.endsWith("/") -> FractionParsingError.INVALID_FORMAT` to catch on real-time.
It should have caught on real-time so that submit button doesn't get active.
2. Do we need some updates in UserAnswer proto default creation?
3. Initialising the variable inside the try/catch will again lead to crashes in the future, need some better understanding here.
https://github.com/oppia/oppia-android/blob/dd024327d32d4c8edcbf12fe31d9df97f7f71a4f/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt#L174

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
[...]

Issue #3467, opened 2021-07-13
Update the mechanism to save checkpoints in the domain layer.


MaskedCarrot:
The current mechanisms for saving checkpoints saves a new checkpoint for every user interaction. But due to the current locking mechanism in `ExplorationProgressController`, we cannot ensure that things in the exploration only change when all save operations are complete. Therefore there can be a potential race condition here that if the user exits the exploration before the last save operation could complete, in that case we might exit with the wrong save operation result.  

Ideally, we should ensure eventual consistency (i.e. regardless of the order of operations, the final outcome is still the correct one). Therefore the mechanism to save checkpoints has to be updated to eliminate any potential race conditions that can arise while saving checkpoints.

For more context see [comment](https://github.com/oppia/oppia-android/pull/3408#discussion_r668182763)

MaskedCarrot:
cc: @BenHenning @aggarwalpulkit596

BenHenning:
I'm wondering if more work is actually needed here now that you're using a DataProvider to deliver the results. It should be eventually consistent and the data race window is pretty slim I think. It might be worth just keeping this around as a reminder to mentally audit the data flow for checkpointing to make sure everything is sound.

PR #4239, opened 2022-03-10
Fix #3622, #4238, #3861, part of #4044: Fix state player deadlock by migrating progress controllers over to a command-queue structure

### LINKED ISSUE #3622: Update internal locking of Exploration Progress Controller
Update the internal locking of Exploration Progress controller to use something like an in-memory
  blocking cache to simplify state locking. However, doing this correctly requires a fix in
  MediatorLiveData to avoid unexpected cancellations in chained cross-scope coroutines. Note that this is also essential to ensure post-load operations can be queued before load completes to avoid cases in tests where the exploration load operation needs to be fully finished before performing a post-load operation. The current state of the controller is leaking this
  implementation detail to tests.

### LINKED ISSUE #3861: Submit-time errors permanently break the answer submission flow
**Describe the bug**
When a submit-time error is encountered, attempting to submit another answer (regardless of correctness) causes an error to be logged in logcat and the submit button stops working (the entire exploration experience goes into a bad state).

**To Reproduce**
1. Open the prototype exploration
2. Navigate to the fraction input state (second state)
3. Input "1/" and click submit
4. Observe the submit-time error
5. Now, try to input "1/2"
6. Observe that the submit button goes gray, no feedback is given, and the exploration is now broken (including back navigation)

Note that navigating away & back to the lesson (e.g. via checkpointing) fixes the issue & is probably a reasonable mitigation in the meantime.

**Expected behavior**
Submit-time errors should not break the lesson experience.

**Demonstration**
None--this is fairly easy to repro so no demo was added.

**Environment**
- Android or SDK version (e.g. Android 5 or SDK 21): SDK 27
- App version (you can get this through system app settings or via the admin controls menu in-app): 0.6-dev-7ff1b2bbd2

**Additional context**
This was found during alpha MR3 development, but it's not specifically tied to the MR3 release.

### LINKED ISSUE #4238: StateFragmentTest is flaky due to ExplorationProgressController deadlock
I noticed while developing #2173. StateFragmentTest can't succeed a single run on Espresso today and it's due to a deadlock in ExplorationProgressController that was caused with the introduction of the domain-side hint handler. The deadlock can occur since ExplorationProgressController calls down into the hint handler which can synchronously call back to ExplorationProgressController. Since both the handler and controller have internal locks, this results in a deadlock. The same scenario can occur for questions since it has the same wiring.

The solution to this is to move away from using locks (e.g. by leveraging Kotlin actors and/or channels/flows). Unfortunately, it requires essentially completely rewriting the core data flow for both explorations and questions.

BenHenning:
## Explanation

Fix #3622
Fix #4238
Fix #3861
Fix part of #4044 (see below for how this relates to the broader math expressions project)

This PR fixes a deadlock that's possible in both ``QuestionAssessmentProgressController`` and ``ExplorationProgressController`` wherein the internal locks used to synchronize session state can conflate with the locks of the controllers' ``HintHandler``s (which calls back into the controller within the lock). While this seems really difficult to hit when playing a training session or exploration, ``StateFragmentTest`` seems to very consistently hit the deadlock after only 10-15 tests (so it essentially can't be run without the fixes introduced here).

While I could've tried rearranging the locks to be more cooperative, it's actually difficult to guarantee that deadlocks won't happen without rebuilding the internal locking mechanism for controller data. I chose the latter by introducing a command queue-esque pattern using a Kotlin actor with ``StateFlow``s for pushing state back to callers (see https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html; actors are a powerful way to leverage the lightweight nature of coroutines to manage cross-thread state machines). To simplify the implementation, this PR also introduces conversion methods from ``StateFlow``s to ``DataProvider``s (I could see us leveraging these flows more in the future, maybe as a long-term replacement to ``DataProvider``s with some refinement).

One side effect of using a command queue to manage what is effectively a session state is that state can now fairly easily leak between sessions. Efforts have been taken to minimize the chance of this happening (mainly ignoring events that arrive from a previously ended session), but a better long-term solution would be to create new session controllers for each session (which should be something possible with custom scopes once we integrate Dagger Hilt in #1720).

Due to a general push toward using ``DataProvider``s much more heavily in these controllers rather than ``LiveData``, ``AsyncResult`` needed to be refactored in order for some of the internal transformed chains in the controllers to work correctly (which is where #4237 comes in).

This also permanently fixes #3861 and other potential errors that could happen by ensuring that the provider state can continue even if internal errors are encountered rather than entering a permanently broken state (#4230 mitigates the immediate repro steps from the bug by fixing the underlying issue that caused the exception, whereas this PR focuses on preventing those scenarios from causing the bug in question).

### Some explanations for the threading approach
I was originally going to make hint handler also a command queue, but it became a bit complicated. To keep things simpler, I removed the threading restricting on it so that it can only be accessed on a single thread (similarly to the internal progress objects used by both controllers) which helps avoid needing a lock without worrying about data races. Note that this command queue style implementation of ``HintHandler`` is available to view in the past commits in this PR if it's interesting for future development (since it's a simpler command queue than the ones used by the controllers due to having an inherently simpler internal state machine).

This PR demonstrates a general tendency to move the app toward a 'notify-and-pull' model (e.g. ``DataProvider``s) vs. a push-subscribe pattern. The former is easier to synchronize, parallelize, ensure lifecycle safety, and is often more performant since operations only need to execute when their results are needed.

Note that both controllers' internal computation behaviors are completely different with this new approach. Previously, the internal progress state machine would only be fully processed when the current ephemeral state/question was retrieved from the UI layer (i.e. lazily), but now it's computed shortly after the change (i.e. eagerly) since that works a bit better with the ``Flow``s being used for internal cross-thread communication. This technically means the controllers are slightly less efficient since execution may happen in cases when the state/question is no longer being viewed by the user, but it should realistically make a small difference since we have to thread hop for the command queue, anyway, and computing the ephemeral state/question should be cheap. This has the benefit of a potentially snappier load experience in the frontend, though, since the ephemeral state/question will be available sooner.

All operations in the controllers now execute their operations regardless of whether they're observed (which should better match callers' expectations).

Note that the internal state machine complexity can probably be drastically simplified now that the command queue can also act as a state machine, but I didn't want to risk regressions by removing it (I was trying to keep the logical flow as similar as possible to the existing behavior).

### Updates to dependencies

The coroutines core dependency needed to be updated since the older version didn't include ``StateFlow`` which was necessary for providing a sync-back mechanism for the controller command queues.

### Stability checks

I verified that both ``QuestionPlayerActivity`` and ``StateFragmentTest`` pass on Espresso (where the latter deadlocked prior to this PR); see the screenshots below. Further, I tested ``ExplorationProgressControllerTest`` and ``QuestionAssessmentProgressControllerTest`` 100x times to verify both are not flaky (since some flakes were discovered during development). Both test suites passed all 100 runs.

### Bazel app module tests
This PR introduces support for dedicated app module tests in Bazel (rather than bundling them as part of the broader list of app module tests). This required a few changes in .bzl and .bazel files, but the change is otherwise straightforward and extremely beneficial for team members that rely on the Android Studio with Bazel plugin for regular development.

### Miscellaneous
I tried introducing a Bazel emulator test while developing this (as part of #59) but it ran into a desugar issue with Mockito and Robolectric (due to them both being pulled into the test binary that's being built to send to the device) which AGP apparently works around. See https://github.com/bazelbuild/bazel/issues/13684 for a Bazel issue tracking the same issue.

This fixes #3622 since it eliminates the lock entirely in favor of using actors for safe cross-thread communication. The ``MediatorLiveData`` specific part of this issue's TODO comment isn't relevant anymore, either, since we replaced ``MediatorLiveData`` with a custom notifying ``LiveData`` within ``DataProviders`` a while back.

This fixes #4238 fully by ensuring any exceptions caused within the state flow of an exploration or questions session are captured and logged, but don't actually break the internal state machine of the session. This issue was discovered during the development of #2173 due to an exception being thrown in the classifier. While that issue was fixed, a test couldn't actually be added for the #4238 fix since it's difficult to intentionally trigger exceptions during answer classification.

### A note on checkpoints in ``ExplorationProgressController``
The checkpointing behavior in ``ExplorationProgressController`` became more complex as a result of these changes. In particular, checkpointing behaves as follows:
- Something changes in the play session
- A checkpoint recomputation is requested (which results in a new checkpoint being computed and then saved to disk)
- The checkpoint is 'processed' by updating the play state of the exploration (i.e. in ``StoryProgressController``)
- The ephemeral state is recomputed since it contains the checkpoint and the checkpoint has changed

Each of these more or less require a command queue "hop" which means other operations can occur between them. As a result, #3467 can still occur since an exploration can be finished before the process happens (though the behavior has changed: the progress simply won't be processed which means an exploration might not be indicated as played, but the checkpointing should almost always be saved before the exploration session end is processed). Fixing this isn't straightforward since the exploration's play state can't be changed until the checkpoint is confirmed as saved which either requires updating ``StoryProgressController`` to also account for the presence of checkpoints, or to use more flows internally to combine the save-and-process operation into one in a way that doesn't lock up ``ExplorationProgressController`` (I'm currently not exactly sure how one would do this, but I haven't thought about it in depth yet).

## Essential Checklist

## For UI-specific PRs only
This PR does not intentionally change and user-perceivable behaviors in the questions or state players except potentially fixing deadlocks/ANRs that could occur while playing. There could be regressions in the players, and they may potentially be more performant after this change, but neither can easily be demonstrated via videos or screenshots. There are no accessibility behavior changes to demonstrate, either.

However, two key tests have been run and verified as passing on Espresso to demonstrate that the deadlock has been fixed. Specifically:

| QuestionPlayerActivityTest | StateFragmentTest |
|------|------|
| deadlock_fix_pr_question_player_activity_test_passing | deadlock_fix_pr_state_fragment_test_passing |


BenHenning:
@vinitamurthi PTAL for codeowners.

@anandwana001 PTAL as part of the broader algebraic expressions project.

anandwana001 (approved):
LGTM

BenHenning:
@anandwana001 could you PTAL at d18c739b1b09fc8fb02adbb85bca69117d887090? It'd be good to get a second pair of eyes on it.

It's a non-trivial change that was needed to fix a series of issues I noticed with explorations. In particular, if the user navigated quickly in and out of an exploration then it could cause the player to destabilize, and potentially crash. It's due to state leaking via the data provider channels, and the fix is to make sure all data providers for both exploration and question controllers are kept fully isolated between sessions (in the same way as the controllers' internal state). This is much more fault tolerant than even our previous controller implementations, and surprisingly all of the tests passed without changes (including the UI tests via Espresso), so it seems like a net simplification and improvement in robustness.

Please let me know if you have any thoughts or concerns.

Edit: Please also look at 40d0a05b1c04390e4410c6f2db2a86e016b56a9a since I realized downstream that the actor itself needs to also be isolated since some events can arrive post-session (such as hint updates), and now that we're no longer filtering on session ID (by design) we need to make sure that such events can't unintentionally be executed for the next session. This is the 'final piece' of isolating the sessions, so it should now be impossible for state to be conflated.

anandwana001 (approved):
Mentioned commit LGTM, no thoughts on these as such.

BenHenning:
Going ahead and merging this per prior discussion with @vinitamurthi since it's high priority for the upcoming alpha release (#4265 is tracking reviewing this as a follow-up after merging).

PR #4256, opened 2022-03-18
Fix #4255: Add support for sharding app module Gradle tests in CI

### LINKED ISSUE #4255: Gradle app module tests frequently fail in CI
This has been an ongoing problem for a few weeks now. From investigation, it seems that Gradle app module tests have reached a point where they consume too much memory and can sometimes be killed by the host CI environment, causing the whole suite to fail. Given that the run takes 40-60 minutes to run, this is really painful (especially when one needs to retry the workflow several times to reach a passing state).

BenHenning:
## Explanation

Fixes #4255

This fixes the recent Gradle flakes we've seen by sharding app module tests into 4 separate runners.

### Introduction of Gradle commands to facilitate sharding
Gradle doesn't have sharding built in, so I needed to take a hacky approach by excluding files from other shards when executing a particular shard (leaving only the tests corresponding to that shard). Since Gradle doesn't expect its source sets to change between builds, a clean build must separate running the shards. Running a shard is as simple as:

```bash
./gradlew :app:testDebugUnitTest --shard0
```

The current shards and their respective tests can be listed using:

```bash
./gradlew :app:testDebugUnitTest --list-shards
```

(Note that some of the tests that are excluded for localization are also included in the list but won't actually be executed).

Another requirement is ensuring that test shard bucketing is deterministic and consistent between runs and environments. To this end, it made sense to use the hash of test files' paths relative to the app module project directory. However, simply using the hash to bucket the tests (such as might happen in a hash table) resulted in a less even distribution of tests (probably due to Java's string ``hashCode`` function being fairly weak for uniqueness), so I decided to instead use the index as a seed of a PRNG. This resulted in a better distribution of tests among shards, and determinism.

### CI changes & verifying that it works
The shards are fixed rather than computed like Bazel, so 4 new workflows were added to separately execute each shard. The results are then combined together in the same way as Bazel, and the result of that workflow is what blocks PR submission. I reused the existing app module workflow for the latter part, so we don't need to change the CI checks requirements (which then means no one needs to update their branches for submission).

See [this workflow](https://github.com/oppia/oppia-android/runs/5606675818?check_suite_focus=true) for an example failure, and [this workflow](https://github.com/oppia/oppia-android/runs/5606875428?check_suite_focus=true) to see that it the required check fails due to the shard failing.

Each shard uploads its test results separately (as demonstrated from [this run](https://github.com/oppia/oppia-android/actions/runs/2006558741)).

The sharding also significantly speeds up running the app module tests (to the point where they're now probably faster than their Bazel counterparts), so this should be a nice qualify-of-life improvement beyond just the flake going away.

## Essential Checklist

## For UI-specific PRs only

N/A -- This is affecting only build infrastructure.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,12 +97,13 @@ jobs:
           name: testing reports
           path: testing/build/reports
 
-  app_tests:
-    name: App Module Robolectric Tests
-    runs-on: ${{ matrix.os }}
+  run_app_module_test:
+    name: Run app module Robolectric test shard
+    runs-on: ubuntu-18.04
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-18.04]
+        shard: [shard0, shard0, shard0, shard0]
     steps:
       - uses: actions/checkout@v2
 
@@ -136,3 +137,16 @@ jobs:
         with:
           name: app reports
           path: app/build/reports
+
+  app_tests:
+    name: App Module Robolectric Tests
+    needs: run_app_module_test
+    if: ${{ always() }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+    steps:
+      - name: Check tests passed (for tests that ran)
+        if: ${{ needs.run_app_module_test.result != 'success' }}
+        run: exit 1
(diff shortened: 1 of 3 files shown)


BenHenning:
@anandwana001 PTAL.

@vinitamurthi if you have time, feel free to review as well. I'm happy to merge with either of your approvals.

anandwana001 (approved):
LGTM

BenHenning:
Thanks @anandwana001! Going ahead and merging this since it should be a really big help on some of the large PRs currently pending merging.

PR #4248, opened 2022-03-14
TODO: Learner analytics impl

BenHenning:
## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [ ] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [ ] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [ ] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [ ] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #4247, opened 2022-03-14
TODO: Learner analytics integration

BenHenning:
## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [ ] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [ ] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [ ] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [ ] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #4115, opened 2022-01-19
Fix part of #4064: Learner Analytics platform parameter

Sarthak2601:
## Explanation

Fixes part of #4064 
This PR gates Learner Study Analytics behind a feature flag. This means that Profile and Devices Id screen and learner study related analytics logging would only occur if this flag is set to `true`. 

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #4267, opened 2022-03-28
Fix #4249, part of #4064: Domain components for learner analytics

### LINKED ISSUE #4249: Update SyncStatusManagerImpl and LoggingIdentifierController to use StateFlow instead of AtomicReference
SyncStatusManagerImpl and LoggingIdentifierController (introduced as part of fixing #4064) need to track thread-safe state. Doing this correctly without running into ordering problems in its fake requires leveraging a StateFlow property (which can't be introduced until coroutines are updated which is being done as part of #4239.

BenHenning:
## Explanation

Fix #4249
Fix part of #4064

Introduces the domain utilities necessary for logging learner analytics, but doesn't make them available for actual usage yet (that is being done in #4269 to keep this PR smaller & more focused).

Some notes on the history of this PR:
- This PR is a rebase of #4253 to remove dependencies on #2173 (which has since been merged into developer) for a much cleaner history.
- This PR is pulling out elements from #4118, #4247, and #4248 which contained completed work by @Sarthak2601.
- This PR extracts just the 'domain' pieces from the above, and changes a bunch about its architecture & adds tests. This PR should have little to no impact on the behavior of the app since the new logging functionality isn't being used yet, or is even accessible to broad components in the app.
- This is starting as 'pt4' since it's  continuing the work introduced in: #4114, #4115, and #4116.

This PR makes a number of changes over the original design document and implementation, the most noteworthy being:
- This PR organizes learner analytics logging into its own logger (and makes changes to event bundle creation & the generic ``OppiaLogger``). I think that we should move toward this pattern generally in the future rather than continuing with a generic ``OppiaLogger`` as it seems to help keep things much more focused. Existing logging should not be affected.
- The notion of a device ID has been dropped as there's no reliable way retrieve such an ID (see https://developer.android.com/training/articles/user-data-ids). Instead, we're using a per-device ID (by leveraging ``PersistentCacheStore``), and have confirmed with study partners that this is workable.
- The logging logic for the new logs was rearranged such that all new analytics logs will be logged for everyone, but the user and installation-tied IDs won't be logged in such cases (since they are more sensitive data). These events are generally useful for the platform, so we shouldn't restrict them as such.
- Learner ID generation for profiles only occurs if the experiment is enabled, and otherwise stays empty. We may add future cleanup code to ensure it's erased across studies, but this at least lays the initial groundwork to keep such IDs separate when they aren't needed.

For a high-level on the design, please refer to [this design document](https://docs.google.com/document/d/1c8lpH-IUvoU1t4LUoYNqNilP2e9yCnzGnSSG0yBxBrY/edit).

Other noteworthy design choices:
- ``DebugEventLogger`` was updated to call through to the real logger (as it makes event verification simpler in developer builds; normally analytics is off so this won't have any effects for the broader team)
- Both ``DebugEventLogger`` and ``FakeEventLogger`` were updated to be thread-safe
- Some extended functionality was added to ``FakeEventLogger``
- ``LearnerAnalyticsLogger`` is designed a bit differently compared to other domain classes in that it actually provides session-specific objects to the application-wide singleton graph (which is needed for logging certain situations, such as the user playing/stopping audio during a play session)
- ``LoggingIdentifierController`` makes use of a lazy retrieval for session ID now (which is fine because it's guaranteed to compute exactly one initial ID)
- ``StateFlow`` is used for easier cross-thread communication, including to expose internal asynchronous state across domain components (the only way we had to do this before was ``Deferred``, and that can be clunky; the new approach is much cleaner)
- An ``EventLogSubject`` was introduced to make testing event logs easier. It's used extensively in tests for this feature, but most existing use cases weren't migrated. #4272 is tracking adding tests for this subject (hence the test file exemption).
- There were TODOs introduced on #4064 to provide explicit clarity to reviewers on what needs to be changed in later PRs (as there's some things being introduced before the final PR that aren't actually used yet to help break up the project)
- Multiple test suites verify behaviors with and without the feature enabled to be very explicit about what behavior occurs when
- ``EventBundleCreatorTest`` in particular has very strict tests to ensure that sensitive IDs are logged exactly when expected (initially, never since they aren't turned on in this PR; this is fixed in a later PR)
- ``ExplorationDataController`` was updated to introduce new play entrypoints, but these aren't "interesting" yet as the underlying ``ExplorationProgressController`` changes are coming in a later PR. Further, testing coverage technically removes checking ``playExploration``, but it'll be removed (and it's technically tested through the other functions since they call through).
- A new ``ClipboardManager`` was introduced with the specific design of not allowing the broad app access to clipboard information from other apps. Instead, it provides an interface to confirm whether the app's known clipboard has been kept. A regex content check was added to ensure developers never use the clipboard service directly and instead use this manager.
- ``PersistentCacheStore`` was updated to include a new ``primeInMemoryAndDiskCacheAsync`` function which works more predictably for initialization than ``primeInMemoryCacheAsync`` (formerly ``primeCacheAsync``). In particular, ``primeInMemoryCacheAsync`` is better for ensuring that the cache will quickly be read once it needs to be (and, if it isn't, will default in the same way the cache store normally defaults). However, there are cases when the app wants to change the default values such that: (1) the normal default is never used, (2) the default has to be computed and isn't cheap, and (3) it should never compute that default again once saved on disk. ``primeInMemoryAndDiskCacheAsync`` makes these assurances which, in turn, makes the installation ID cache store even possible without potential race conditions or breaking Dagger's cheap-initialization best practice. 

Test exemptions: all exemptions are annotations or interfaces except ``EventBundleCreatorTest`` (which is explained above).

## Essential Checklist

## For UI-specific PRs only

N/A -- This PR doesn't make UI changes, and existing flows shouldn't be affected.


BenHenning:
PTAL @anandwana001 at the whole PR (as you'll be the sole reviewer for this change).

I'm also planning to defer @vinitamurthi's review per earlier conversations.

anandwana001 (changes requested):
Overall LGTM

anandwana001 (approved):

anandwana001 (approved):

BenHenning:
Deferring @rt4914's review (tracked in #4266) to help get this merged sooner, and since @anandwana001 has reviewed the whole PR.

rt4914:
Approved. I have checked only for code-owner files.

PR #4279, opened 2022-03-31

translatewiki:
Translation updates


BenHenning:
The strings look good, but I need to double check the FAQ (& general app layout) to make sure string interpolation is working correctly, and that the little bit of embedded HTML wasn't mangled.

BenHenning:
Other than the unrelated Gradle CI failure & some missing strings, the translations seem to be working. They're generally longer than other languages, so they're revealing some sizing issues throughout the app but they should be sufficient for the upcoming study.

We may need to consider specifying character limits in the future, as well, if that's an option.

BenHenning (approved):

PR #4327, opened 2022-04-27
Fixes #4326: Performance Metrics Platform Params

Sarthak2601:
## Explanation

Fixes #4326
This PR introduces the following: 

- [x] Flag for feature gating Performance Metric Collection
- [x] Flag for configurable time interval for log uploading
- [x] Flag for configurable time interval for frequent metric collection
- [x] Flag for configurable time interval for normal metric collection

Here Flag = Platform Parameter

This flag would be used at numerous places like work manager configurations to provide values for necessary further action. 

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

PR #4336, opened 2022-04-30
Fixes #4328: Proto definitions for performance metrics collection

### LINKED ISSUE #4328: Add Proto Definitions for Performance Metrics collection
This issue tracks the implementation of protocol buffer definitions for the process of Performance Metrics collection. These definitions would play a key role while logging these metrics. We'll be defining the following metrics: 

- Timestamp
- Priority
- a boolean to know whether app is in foreground or not
- Memory tier of the device [ =<1GB, =<2GB && >1GB, >2GB ]
- Storage tier of the device [ =<5GB, =<20GB && >5GB, >20GB ]
- Network type of the device [ none, wifi, cellular ]
- Current Activity.

Sarthak2601:
## Explanation

Fixes #4328
Protocol buffer definitions for the process of Performance Metrics collection. Every metric log would contain the following: 

- Timestamp
- Priority
- a boolean to know whether app is in foreground or not
- Memory tier of the device [ <2GB, =<3GB && >=2GB, >3GB ]
- Storage tier of the device [ <32GB, =<64GB && >=32GB, >64GB ]
- Network type of the device [ none, wifi, cellular ]
- Current Activity.

## Essential Checklist


model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Structure for an oppia metric log.
+message OppiaMetricLog {
+  // The timestamp of the event.
+  // unit: long/millis, timezone: UTC
```
BenHenning: Should be one one line as both lines are documenting the same field.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    LOW_MEMORY_TIER = 1; // =<1GB
+
+    // The memory tier where the amount of memory is less than or equal to 2GB and more than 1GB.
+    MEDIUM_MEMORY_TIER = 2; // =<2GB && >1GB
+
+    // The memory tier where the amount of memory is greater than 2GB.
+    HIGH_MEMORY_TIER = 3; // >2GB
+  }
+
+  // Supported storage tiers for indicating the total storage space on user's device.
```
BenHenning, line 152: Is there any external reference documentation that we can point to that supports the size designations for each of these size tiers?

Ditto for MemoryTier.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    // size exceeds a certain limit and no low priority logs exist.
+    MEDIUM_PRIORITY = 2;
+
+    // The priority of events whose logs should not be removed from the storage on most occasions
+    // but can be removed if they're the only ones there and size limit exceeds.
+    HIGH_PRIORITY = 3;
+  }
+
+  // Supported activities for indicating the current activity where the event logging is occuring.
+  enum CurrentActivity {
```
BenHenning: Perhaps instead this should be 'screen' instead of 'Activity'? The latter would require tracing to implement generically, or we'll have to remember to update this enum (which is a bit error-prone). We could go with the tracing approach in the future, but for now maybe we can just abstract it to screens (even though generally screens->activities will be a 1:1 relationship, the benefit is not needing to necessarily know the activity at the domain level, just which screen it corresponds to).

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    // The undefined priority of an event.
+    PRIORITY_UNSPECIFIED = 0;
+
+    // The priority of events whose logs will be the first to be removed from the storage if the
+    // size exceeds a certain limit
+    LOW_PRIORITY = 1;
+
+    // The priority of events whose logs will be the ones to be removed from the storage if the
+    // size exceeds a certain limit and no low priority logs exist.
+    MEDIUM_PRIORITY = 2;
```
BenHenning, line 119: Why do we need a medium priority? This isn't needed for the other event logs.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    int64 bytes_received = 1;
+
+    // Number of bytes transmitted by the application from the device during the bucket's time
+    // interval.
+    int64 bytes_transmitted = 2;
+  }
+
+  // Structure of the cpu usage package.
+  message CpuUsage{
+    // Amount of cpu used by the application.
```
BenHenning, line 103: What is the unit for this field?

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    // Number of bytes received by the application on the device during the bucket's time interval.
+    int64 bytes_received = 1;
+
+    // Number of bytes transmitted by the application from the device during the bucket's time
+    // interval.
+    int64 bytes_transmitted = 2;
+  }
+
+  // Structure of the cpu usage package.
+  message CpuUsage{
```
BenHenning: Nit: space before ``{``.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+  message NetworkUsage{
+    // Number of bytes received by the application on the device during the bucket's time interval.
+    int64 bytes_received = 1;
+
+    // Number of bytes transmitted by the application from the device during the bucket's time
+    // interval.
+    int64 bytes_transmitted = 2;
+  }
+
+  // Structure of the cpu usage package.
```
BenHenning: 'Package' isn't really clear. If this is a type of metric, perhaps the documentation & the proto name should end with 'Metric'.

Ditto for all others that are types of metrics.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+
+  // Structure of the memory usage package.
+  message MemoryUsage {
+    // Amount of memory used by the application on the device in bytes.
+    int64 total_pss_bytes = 1;
+  }
+
+  // Structure of the network usage package.
+  message NetworkUsage{
+    // Number of bytes received by the application on the device during the bucket's time interval.
```
BenHenning: Is 'bucket time interval' defined anywhere in this proto file? Otherwise, it's potentially really unclear.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+
+  // Structure of the startup latency package.
+  message StartupLatency {
+    // Amount of latency on application startup in millis.
+    int64 startup_latency_millis = 1;
+  }
+
+  // Structure of the memory usage package.
+  message MemoryUsage {
+    // Amount of memory used by the application on the device in bytes.
```
BenHenning, line 78: Suggest pointing to external documentation to provide a definition for 'pss' since it's not self-evident.

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+
+  // Structure of the storage usage package.
+  message StorageUsage {
+    // Amount of storage usage by the app on user's device in bytes.
+    int64 storage_usage_bytes = 1;
+  }
+
+  // Structure of the startup latency package.
+  message StartupLatency {
+    // Amount of latency on application startup in millis.
```
BenHenning: What is latency in this context?

Perhaps instead: ``Number of milliseconds required to start up the application from a cold start.``

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+
+  // The current network type of the device.
+  NetworkType network_type = 6;
+
+  // The current activity where the event is getting logged.
+  CurrentActivity current_activity = 7;
+
+  // Structure of the apk size package.
+  message ApkSize {
+    // Size of the android package file of the app on the device in bytes.
```
BenHenning: Can be simplified I think: ``Size of the app's installed APK file, in bytes.``

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+    int64 apk_size_bytes = 1;
+  }
+
+  // Structure of the storage usage package.
+  message StorageUsage {
+    // Amount of storage usage by the app on user's device in bytes.
```
BenHenning: Please document whether this includes the APK size and app cache (since these are normal questions to ask, I think, given the description).

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+// Structure for an oppia metric log.
+message OppiaMetricLog {
+  // The timestamp of the event.
+  // unit: long/millis, timezone: UTC
+  int64 timestamp_millis = 1;
+
+  // The priority of the event.
+  Priority priority = 2;
+
+  // Indicates whether the application is in foreground.
```
BenHenning:
```suggestion
  // Indicates whether the application is in the foreground.
```

model/src/main/proto/performance_metrics.proto
```diff
@@ -0,0 +1,149 @@
...
+  StorageTier storage_tier = 5;
+
+  // The current network type of the device.
+  NetworkType network_type = 6;
+
+  // The current activity where the event is getting logged.
+  CurrentActivity current_activity = 7;
+
+  // Structure of the apk size package.
+  message ApkSize {
```
BenHenning: This & other metrics aren't actually referenced in ``OppiaMetricLog``--are they missing?

BenHenning (changes requested):
Thanks @Sarthak2601. Mostly nits, but I had a few clarifying questions as well.

model/src/main/proto/performance_metrics.proto
Sarthak2601: According to our initial discussions, we agreed upon the 3 priority scale -- low, medium and high with:
- startup metrics: low
- ui specific metrics: medium
- periodic metrics: high

Should we make changes to this?

model/src/main/proto/performance_metrics.proto
Sarthak2601: CPU usage is going to be in percentages, so unit-less -- does that make sense?

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: thanks for flagging -- added.

model/src/main/proto/performance_metrics.proto
Sarthak2601: added definition to make it clearer -- does it help?

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: There is no clear-cut documentation that supports the size designations for each of these size tiers. It was something that we came up with during our initial discussions and thought was apt. Do you have any reference doc suggestions regarding this?

model/src/main/proto/performance_metrics.proto
```diff
@@ -26,45 +25,78 @@ message OppiaMetricLog {
...
-  CurrentActivity current_activity = 7;
+  // The metric log of the event.
+  MetricLog metric_log = 7;
 
-  // Structure of the apk size package.
+  // The current screen where the event is getting logged.
+  CurrentScreen current_screen = 8;
+
+  // Structure of a metric log.
+  message MetricLog {
```
BenHenning: Why have a separate message type for this vs. just having the one-of in the top-level event?

model/src/main/proto/performance_metrics.proto
```diff
@@ -26,45 +25,78 @@ message OppiaMetricLog {
...
     int64 total_pss_bytes = 1;
   }
 
-  // Structure of the network usage package.
-  message NetworkUsage{
-    // Number of bytes received by the application on the device during the bucket's time interval.
+  // Structure of the network usage metric.
+  message NetworkUsage {
+    // Usage data is collected in discrete bins of time called 'Buckets'. This fields denotes the
+    // number of bytes received by the application on the device during this bucket's time interval.
```
BenHenning: How long is a bucket? Also, shouldn't this be logged at the end of the bucket's timeframe (i.e. indicating that bytes were received, not are currently being received)?

model/src/main/proto/performance_metrics.proto
```diff
@@ -26,45 +25,78 @@ message OppiaMetricLog {
...
     int64 bytes_received = 1;
 
-    // Number of bytes transmitted by the application from the device during the bucket's time
+    // Usage data is collected in discrete bins of time called 'Buckets'. This fields denotes the
+    // number of bytes transmitted by the application on the device during this bucket's time
     // interval.
     int64 bytes_transmitted = 2;
```
BenHenning: Transmission can mean both sending and receiving. Perhaps be explicit here & in the documentation by indicating 'sent' (to contrast with 'received').

model/src/main/proto/performance_metrics.proto
BenHenning: If it's a percentage, then it's presumably a ratio. What's being divided to compute it? Overall, I'm trying to dig into the context of what this value means since it's not clear from current documentation how to actually use or populate it.

model/src/main/proto/performance_metrics.proto
BenHenning: I don't think it's fully clarified yet, but see my newest comment.

model/src/main/proto/performance_metrics.proto
BenHenning: Hmm, I think it's fine to keep this given that context.

model/src/main/proto/performance_metrics.proto
BenHenning: I don't, but perhaps we should be doing a brief investigation of the distribution of available storage on Android devices throughout the world to get an idea on what these tiers should be (versus arbitrarily setting them).

model/src/main/proto/performance_metrics.proto
BenHenning: This doesn't seem resolved yet--I was also suggesting renaming the proto messages to more clearly indicate that they correspond to a measurable metric. Do you have any concerns with that?

BenHenning (changes requested):
Thanks @Sarthak2601. I took another pass--PTAL.

Also, #4327 has been merged in, so I think this PR needs to be brought up-to-date with develop.

model/src/main/proto/performance_metrics.proto
Sarthak2601: I preferred creating a separate message type because it becomes easier to pipe, pack and unpack data from the logging site to OppiaLogger (+Controller) and then finally to the utility. Just like it has been done in case of 'Context' for the analytics event logs. If we have a top level one-of, we'll always have to transport the entire metric log from one place to another and won't be able to safely transport packaged individual values (like in case of network usage) together. Does that make sense?

model/src/main/proto/performance_metrics.proto
Sarthak2601: A bucket is as long as we want it to be. We can customise the start and end time of a query to get network usage in between those timestamps. And yes, it should be logged at the end of the bucket's timeframe -- modified the documentation to make it clearer -- does that help?

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: nope, no concerns -- done!

model/src/main/proto/performance_metrics.proto
BenHenning: I think that's fine, but perhaps this should just be 'LoggableMetric' instead, as it's not a full log but rather a constituent property of a log object.

model/src/main/proto/performance_metrics.proto
Sarthak2601: I did a brief research on this topic and unfortunately, I couldn't find anything supporting the classification of android devices on the basis of storage/memory space. Perhaps it is because of the rapidly changing storage and memory space configurations in newer android devices. It'll be really helpful if you could suggest something here :)

BenHenning: https://www.counterpointresearch.com/average-smartphone-nand-flash-capacity-crossed-100gb-2020/ is the closest to historical data that I can find, unfortunately, it's only averages. It does provide data back a few years, so perhaps we could cross reference that with historical usage data to get a rough idea on distribution? https://www.counterpointresearch.com/smartphone-storage-capacity-zooms-increased-demand/ might also help since it provides a brand-based breakdown.

We could also look at most used devices in India, Kenya, Ghana, and Brazil to get a breadth of devices and their storage capacity offerings.

This isn't going to be very good math, but it'll probably provide enough basis for estimating storage classes.

model/src/main/proto/performance_metrics.proto
Sarthak2601: done.

model/src/main/proto/performance_metrics.proto
Sarthak2601: I've created a separate issue that tracks the implementation of CPU usage performance metrics. I've added a todo here to ensure correct proto message + comment goes in when we merge all our PRs.

model/src/main/proto/performance_metrics.proto
Sarthak2601: Taking into account the data Isalo provided, I've modified the memory and storage tiers to be somewhat like:  

1. Memory Tiers 
- Low --> <2GB
- Medium --> <=3GB && >=2GB
- High --> >3GB

2. Storage Tiers 
- Low --> <32GB
- Medium --> <=64GB && >=32GB
- High --> >64GB

I've taken the device market share + device model specs (memory and storage) into account to come up with this. Does this suffice?

BenHenning: Yup this seems quite sensible--thanks!

model/src/main/proto/performance_metrics.proto
BenHenning: I think it's reasonably clear now--thanks!

BenHenning (approved):
Thanks @Sarthak2601! I have nothing else to add--the PR LGTM!

PR #4392, opened 2022-06-13
Fix #2581: Marquee auto restart issue

### LINKED ISSUE #2581: Marquee restarts automatically in certain scenarios.
**Describe the bug**
In ideal case, Marquee should only start when user click on toolbar but Marquee restarts automatically in certain scenarios after Marquee stops for the first time.

**To Reproduce**

General Steps for all below mentioned cases:
1) Go to The Meaning of Equal Parts exploration
2) Go to Second state. Note that the title is not moving.
3) Click on Toolbar and Marquee starts moving.

Case 1:
4) Once marquee stops, pull down notification bar of your mobile from top to bottom and then pull up, you will see Marquee moving again without clicking the toolbar.

Case 2:
4) Once marquee stops, click on EditText, Marquee starts moving again. Now wait for it to stop don't do anything.
Also after the marquee stops, then press back button to close the keyboard and you will see Marquee starts moving again

Case 3:
4) Once marquee stops, click on hint button, Marquee starts moving again.

Case 4:
4) Once marquee stops, click on close button at top-left, Marquee starts moving again.

Case 5:
4) Once marquee stops, on clicking it again, Marquee won't start no matter how many times user clicks on the toolbar.

**Expected behavior**
Marquess effect should start only on click on the toolbar and not automatically & it should starts every time user clicks on the toolbar after marquee stops moving.

For reference please refer to this issue: #1770

KevinGitonga:
## Explanation
  Fix #2581
  This fix introduces a custom View "MarqueeView" which acts as parent view for the textview in the toolbar this allows the MarqueeView to manipulate the child TextView and allow's it to marquee through translation animation if the text is long enough. It also solves the issue which we had which of text animating over the back navigation icon by attaching the textview inside a scrollview which allows it disappear cleanly without overlapping.

## Essential Checklist

## For UI-specific PRs only

## Before changes demonstration
https://user-images.githubusercontent.com/20886444/173300328-75af184a-d528-4a2b-8ae9-15ec942e577a.mp4

## After changes demonstration
https://user-images.githubusercontent.com/20886444/185438100-fcc50852-e8ca-4828-9d2f-26697ff40646.mp4

## After changes RTL demonstration
https://user-images.githubusercontent.com/20886444/185930082-cf408e7c-2b32-479b-9312-e6cfe5c7e8f0.mp4


app/src/main/java/org/oppia/android/app/devoptions/markchapterscompleted/MarkChaptersCompletedFragmentPresenter.kt
```diff
@@ -78,6 +78,14 @@ class MarkChaptersCompletedFragmentPresenter @Inject constructor(
...
           }
         }
+      } else if (binding.isAllChecked == true) {
+        binding.isAllChecked = false
+        getMarkChaptersCompletedViewModel().getItemList().forEach { viewModel ->
+          if (viewModel is ChapterSummaryViewModel) {
+            if (!viewModel.checkIfChapterIsCompleted())
+              chapterUnselected(viewModel.chapterIndex, viewModel.nextStoryIndex)
+          }
+        }
```
rt4914: This branch contains code of your previous branch too. 
I suggest two solutions:
1. Either undo-the-changes in this branch by checking at the `Files Changed` tab on https://github.com/oppia/oppia-android/pull/4392
2. Or, create a new branch+PR for this issue and in this case the branch should be created from `develop`

app/src/main/java/org/oppia/android/app/devoptions/markstoriescompleted/MarkStoriesCompletedFragmentPresenter.kt
```diff
@@ -74,6 +77,10 @@ class MarkStoriesCompletedFragmentPresenter @Inject constructor(
           if (!viewModel.isCompleted)
             storySelected(viewModel.storySummary.storyId)
         }
+      } else {
+        getMarkStoriesCompletedViewModel().getStorySummaryMap().values.forEach { viewModel ->
+          if (!viewModel.isCompleted) storyUnselected(viewModel.storySummary.storyId)
+        }
```
rt4914: Undo changes in this file.

app/src/main/java/org/oppia/android/app/devoptions/marktopicscompleted/MarkTopicsCompletedFragmentPresenter.kt
```diff
@@ -64,15 +64,18 @@ class MarkTopicsCompletedFragmentPresenter @Inject constructor(
...
 
     binding.markTopicsCompletedAllCheckBox.setOnCheckedChangeListener { _, isChecked ->
       if (isChecked) {
         getMarkTopicsCompletedViewModel().getTopicList().forEach { viewModel ->
           if (!viewModel.isCompleted) topicSelected(viewModel.topic.topicId)
         }
+      } else {
+        getMarkTopicsCompletedViewModel().getTopicList().forEach { viewModel ->
+          if (!viewModel.isCompleted) topicUnselected(viewModel.topic.topicId)
+        }
```
rt4914: Undo changes in this file.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -61,6 +63,12 @@ class TopicFragmentPresenter @Inject constructor(
       binding.topicToolbarTitle.isSelected = true
     }
 
+    lifecycleSafeTimerFactory.createTimer(25000).observe(fragment.viewLifecycleOwner) {
```
rt4914: What does this `25000` timer exactly do? Can you please explain this how this fixes the issue.

rt4914 (dismissed):
@KevinGitonga Suggested changes.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
KevinGitonga: Hi Rajat this fix unselects the selected marquee textview exactly 25 seconds later after it's selected, this is on assumption that the marquue loop has ended or is close to ending this prevents repeat looping incase of other scenarios.

app/src/main/java/org/oppia/android/app/devoptions/markchapterscompleted/MarkChaptersCompletedFragmentPresenter.kt
KevinGitonga: Update applied.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
BenHenning: Does unselecting it cause the marqueeing to stop? If so, could we conditionally unselect it once it's finished rather than using 25 seconds? I'm concerned that '25' is a magic number and might not apply correctly to all situations.

Also, does re-tapping on the marquee correctly restart it once it's been deselected?

BenHenning:
Thanks @KevinGitonga. Left one comment--please also add tests once we land on a final solution.

Note that in the future I suggest only sending a PR for review once it's completed from your perspective (which includes adding tests). If you'd like first thoughts on a PR as you're finishing (e.g. before tests are written), I suggest marking the PR as a 'draft' and explicitly mentioning when assigning reviewers that you'd like their initial thoughts. This helps better communicate expectations and ensure that reviewers are giving you the type of feedback that you're looking for.

app/src/main/java/org/oppia/android/app/devoptions/marktopicscompleted/MarkTopicsCompletedFragmentPresenter.kt
KevinGitonga: Changes Applied

app/src/main/java/org/oppia/android/app/devoptions/markstoriescompleted/MarkStoriesCompletedFragmentPresenter.kt
KevinGitonga: Changes Applied

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
KevinGitonga: Hi Ben unselecting it sure does stop the marqueeing, how do we know that it has finished marqueeing? Yes works correctly however if scenario occurs earlier than timer this does not solve the issue.

KevinGitonga: @BenHenning How do you suggest we conditionally unselect it once it's finished.

BenHenning: I think I mentioned this in a chat or in email, but we might actually need to look into extending TextView and forming our own basic marquee functionality in order to handle this properly (since my searches also didn't yield any way to actually detect if the marquee finished). This is the main reason why I suspect that this issue might be a bit harder compared to some of the others you've been looking at--it's not actually obvious what the solution is here.

KevinGitonga: Hi @BenHenning myself too, i haven't found any solid docs on ways to detect this. Check out this suggestion to use reflection
https://stackoverflow.com/questions/12392540/fire-event-when-marquee-completes do you think it can work. I have used reflection to try get fields for testing and it didn't work successfully for me.

BenHenning: @KevinGitonga reflection is definitely not ideal (besides the performance implications, it's also tricky to get working correctly with Proguard and is broadly considered an antipattern in production applications).

I looked more deeply into the marquee implementation within TextView and am more confident that it shouldn't be too difficult for us to recreate its behavior. I actually don't think we even need to bother with extending TextView.

I think we can introduce our own text marquee logic that just translates the TextView's text over time until the full text has been translated (i.e. the text's width) at which point it is restarted. Android's marquee implementation does a bit more then this (to account for images within the text view, and fading), but we don't need that complexity. It's also important to note that the marquee seems to update no faster than 30 frames per second which is probably for performance reasons, so we should probably replicate that.

Let me know if you have any specific concerns with approach the solution in this way.

KevinGitonga: @BenHenning I have noticed that using translation animation to move text from RTL may be tricky since text will just go over the back navigation not as in Marquee where it disappears before it.

app/src/main/java/org/oppia/android/app/customview/MarqueeView.kt
```diff
@@ -0,0 +1,236 @@
...
+import android.widget.FrameLayout
+import android.widget.LinearLayout
+import android.widget.ScrollView
+import android.widget.TextView
+import java.lang.RuntimeException
+import kotlin.math.abs
+
+/**
+ * Provides a simple marquee effect for a single [android.widget.TextView].
+ * Reference: https://github.com/ened/Android-MarqueeView/blob/master/library/src/main/java/asia/ivity/android/marqueeview/MarqueeView.java
```
BenHenning: We can't just copy an existing view (or any code) from a third-party. If we want to use MarqueeView but need to change it, then we should fork the original repo, make the changes we need, and add it as a dependency (using jitpack for simplicity). We actually do this for other situations, such as KotliTeX and AndroidSVG.

BenHenning (changes requested):
Thanks @KevinGitonga! I left one primary comment about the way we're incorporating MarqueeView--that definitely will need updating. Also: is your PR description accurate? It doesn't seem to be discussing the new dependency that we're pulling in, or why it's needed. Please update it accordingly.

Once the comment is addressed & description updated, I'll take a full pass on the PR.

app/src/main/java/org/oppia/android/app/customview/MarqueeView.kt
KevinGitonga: Makes sense i will need to fork this and update the library. Is it okay if i fork to my account or it will need to be under the Oppia Project.

seanlip: I have made a fork here which you can use: https://github.com/oppia/Android-MarqueeView

WORKSPACE
```diff
@@ -139,6 +139,14 @@ git_repository(
     shallow_since = "1647554845 -0700",
 )
 
+# A custom fork of Android-MarqueeView that is updated with latest dependencies compatible with Oppia and also
+# min target SDK version set to be compatible with Oppia.
+git_repository(
+    name = "marqueeview",
+    commit = "472f8d63faf6491da0f5a04cafbad3f70878fb81",
+    remote = "https://github.com/KevinGitonga/Android-MarqueeView",
```
BenHenning: Reminder to update this to:
- Point to Oppia's fork, instead
- Point to the correct commit (once all of your changes are merged)
- Add the recommended shallow_since property per Bazel's recommendation when you try building (it's a warning that shows up at the beginning of the build log)

app/build.gradle
```diff
@@ -178,6 +178,7 @@ dependencies {
       'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1',
       'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1',
       'org.mockito:mockito-core:2.7.22',
+      'com.github.KevinGitonga:Android-MarqueeView:472f8d63fa',
```
BenHenning: Shouldn't this be ``"//third_party:asia_ivity_android_marqueeview",`` instead?

app/src/main/res/layout-sw600dp/story_fragment.xml
```diff
@@ -36,13 +37,20 @@
         app:navigationContentDescription="@string/navigate_up"
         app:navigationIcon="?attr/homeAsUpIndicator">
 
+        <asia.ivity.android.marqueeview.MarqueeView
+          android:id="@+id/marquee_view"
+          android:layout_width="match_parent"
+          android:layout_height="wrap_content"
+          marquee:speed ="10"
```
BenHenning: Nit: extra space before the ``=``. Ditto in several of the other changed layout files.

app/src/main/res/layout-sw600dp/story_fragment.xml
```diff
@@ -36,13 +37,20 @@
...
+          marquee:pause="500">
         <TextView
```
BenHenning: The indentation seems off here now that this is a child of the marquee view--please fix. Ditto for other layout files.

app/src/main/res/layout-sw600dp/story_fragment.xml
```diff
@@ -36,13 +37,20 @@
         app:navigationContentDescription="@string/navigate_up"
         app:navigationIcon="?attr/homeAsUpIndicator">
 
+        <asia.ivity.android.marqueeview.MarqueeView
+          android:id="@+id/marquee_view"
```
BenHenning: Since IDs are global, all IDs should be specific to their specific fragment or custom view. In this case, I suggest: ``story_toolbar_title_marquee`` ('view' is redundant). Ditto everywhere else.

app/src/main/res/layout-sw600dp/story_fragment.xml
BenHenning: If it's not too much effort, could we maybe rename this & the next custom attributes in the base library to be something more understandable? Specifically:
- 'speed' of what?
- What are the units for 'speed' and 'pause'?
- Is 'pause' really just a 'delay'?

Including these details in the attribute names should help with clarity.

app/src/main/res/layout/exploration_activity.xml
```diff
@@ -38,13 +38,21 @@
...
+            app:pause="500">
           <TextView
             android:id="@+id/exploration_toolbar_title"
             style="@style/MarqueeToolbarTitleTextView"
-            android:layout_width="0dp"
+            android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_marginEnd="20dp"
-            android:layout_weight="1" />
+            />
```
BenHenning: This should be on the line above.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -225,36 +224,6 @@ class StoryFragmentTest {
     }
   }
 
-  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
-  // its view
-  fun testStoryFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly() {
```
BenHenning: We should still be testing that the marquee works as expected. These tests should be updated rather than removed. Ditto for all other test suites & affected screens in this PR.

BenHenning (changes requested):
Thanks @KevinGitonga. Took a pass on the PR.

Regarding speed, is that based on pixels, dp, or sp? I'd assuming it should be sp since otherwise the speed will be different based on text scaling & display density which doesn't sound correct.

Also, should we maybe use a consistent pause & speed value across all marquee views (e.g. by introducing a shared value)?

KevinGitonga:
> Thanks @KevinGitonga. Took a pass on the PR.
> 
> Regarding speed, is that based on pixels, dp, or sp? I'd assuming it should be sp since otherwise the speed will be different based on text scaling & display density which doesn't sound correct.
> 
> Also, should we maybe use a consistent pause & speed value across all marquee views (e.g. by introducing a shared value)?

Thanks @BenHenning it is based on pixels. Great concerns should we convert the pixels to sp and use this instead. I will check behavior with this. Although the it might be hard for the user to notice the difference in speed. I have checked on tablet emulator and haven't noticed any changes.

WORKSPACE
KevinGitonga: Noted..

app/build.gradle
KevinGitonga: This is for gradle dependency. Doesn't resolve this when building.

app/src/main/res/layout-sw600dp/story_fragment.xml
KevinGitonga: Noted.

app/src/main/res/layout-sw600dp/story_fragment.xml
KevinGitonga: Noted.

app/src/main/res/layout-sw600dp/story_fragment.xml
KevinGitonga: Missed this the rest are named with their corresponding Fragment e.g. exploration_marquee_view, license_text_viewer_activity_marquee_view.

app/src/main/res/layout-sw600dp/story_fragment.xml
KevinGitonga: Noted.

app/src/main/res/layout/exploration_activity.xml
KevinGitonga: Noted.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
KevinGitonga: Will revert the tests for this PR and find out the exact reason why they were failing.

app/src/main/res/layout-sw600dp/story_fragment.xml
BenHenning: Still unfixed I believe.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -323,8 +322,7 @@ class ExplorationActivityTest {
       explorationActivityTestRule.activity.findViewById(R.id.exploration_toolbar_title)
     ViewCompat.setLayoutDirection(explorationToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
 
-    onView(withId(R.id.exploration_toolbar_title)).perform(click())
-    assertThat(explorationToolbarTitle.ellipsize).isEqualTo(TextUtils.TruncateAt.MARQUEE)
```
BenHenning, line 327: For this & the other tests, could you maybe verify that the marquee view is the correct view type (i.e. using ``isInstanceOf``)? It's not a great way to verify "correctness," but I think it's the only way currently-feasible way to verify that things are behaving/hooked up correctly.

WORKSPACE
```diff
@@ -139,6 +139,14 @@ git_repository(
     shallow_since = "1647554845 -0700",
 )
 
+# A custom fork of Android-MarqueeView that is updated with latest dependencies compatible with Oppia and also
+# min target SDK version set to be compatible with Oppia.
+git_repository(
+    name = "marqueeview",
```
BenHenning, line 145: Per earlier comment, reminder to add the shallow_since (you should be able to find the recommended line to add from the build logs run in CI).

app/build.gradle
BenHenning: Ah oops, not sure how I made that mistake. Thanks. :)

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
BenHenning: While Rajat originally opened this thread, closing it now since we've implemented an entirely different solution that doesn't have the same issue originally identified.

BenHenning (commented):
Thanks @KevinGitonga. Just had a few follow-up comments--PTAL.

app/src/main/res/layout-sw600dp/story_fragment.xml
KevinGitonga: Noted will try with AS inbuilt reformatter.

WORKSPACE
KevinGitonga: Noted.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
KevinGitonga: Noted makes sense.

KevinGitonga:
@BenHenning have added tests and shallow_since config PTAL.

BenHenning (approved):
Thanks @KevinGitonga! I think this has turned into quite a nice solution. LGTM.

BenHenning:
@rt4914 since you requested changes, could you please take another look?

BenHenning:
Actually, dismissing @rt4914's review since his original concerns seem addressed and he's on holiday at the moment.

Rajat: if you have any follow-up concerns, please be sure to let @KevinGitonga know.

Going ahead and merging this.

PR #4408, opened 2022-06-30

translatewiki:
Translation updates


BenHenning (approved):
Nice--I think this adds all of the expressions & equations translations for pt-BR + fixes some of the fuzzy strings. It also reveals that we have some test & infrastructural strings that shouldn't be translated.

PR #4417, opened 2022-07-03
Fix #4300, #2432: Introduce beta & general availability update notices and build flavors

### LINKED ISSUE #2432: Create multiple build flavors in Bazel
While we don't yet have compile-time feature gating, we should be introducing different flavors of the app in Bazel builds. Specifically, we should have:
- oppia_dev: a developer-only build with native multidex and no proguard optimizations
- oppia_prod: a production-facing build with legacy multidex and with proguard optimizations including 3-pass optimization (this needs to be verified as working since we've run into dex limit issues with databinding)

Longer-term, we will likely want alpha/beta/ga versions of the prod build. Ideally, these would also be Android app bundles rather than APKs, but that's blocked on https://github.com/bazelbuild/bazel/issues/11497.

### LINKED ISSUE #4300: Introduce beta notice
Users need to be aware of when they're changing to a beta version of the app.

See [PRD](https://docs.google.com/document/d/1vnD20MuZ8JVr6ZuA40p5c2GRiJrCpbodMghqHJSeMHg/edit?usp=sharing) and [mocks](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/d92286cb-e0d8-4687-b6a3-6448c44dbe00/) (see the 'Splash Page w/ Beta notice'). Users should see a notice when changing to the beta channel (from either alpha or GA), and only during app startup. If they select "Don't show again" then the notice should not be shown again _unless_ they change from & back to the beta channel (then the app should reset the earlier preference).

The preference is per-app, not per-user.

**Describe alternatives you've considered**
Lots of alternatives, including using an icon, title change, splash text + waiting until the user clicks on it, extra bar added to the profile page & home screen, and others. The solution the team landed on was the simplest approach (technically) that should minimize user confusion and avoid accessibility problems.

**Additional context**
None.

BenHenning:
## Explanation
Fixes #4300
Fixes #2432

This PR introduces support for showing notices when users change from certain flavors of the app to the beta and/or GA (general availability) flavors (both of which are newly introduced in this PR).

The user behavior is as follows:
- Changing from any flavor _to_ beta will result in the beta notice showing for the user the *first* time they open the app.
- Changing from alpha or beta flavors _to_ the GA flavor will result in the GA notice being shown.
- The user will always see the notice after the first open following one of these transitions, and in no other cases (including the first open for beta).
- The user will not see one or both notices if they select the "never show again" checkbox(es) prior to dismissing the notice(s).
- The app deprecation notice will always take logical precedence over other notices (i.e. it will always show in cases when the app is deprecated regardless of flavor transitions).

In addition to the above, text was added to the splash screen for the developer, alpha, and beta flavors of the app to make it a bit more explicit to the user. The alpha and beta splash experiences have been configured to wait 2 seconds so that the user has a moment to see the actual text before the app finishes loading.

Note also that existing alpha users will not see the beta notice after installing the first beta version (since no flavor has been recorded). While this is technically fixable, it doesn't seem worthwhile given the current small user audience that would be affected. Subsequent transitions from alpha to beta will work as expected (and lead to the beta notice being shown).

As a side note: #2432 is fixed as of this PR since the final planned build flavors are being introduced here. While more flavors may be added in the future, it seems sufficient to consider "build flavors are introduced" to be done now that there are developer, alpha, beta, and GA flavors of the app.

Technical things of note:
- The splash screen now has a contingency fallback of 5 seconds in case the startup state doesn't ever load/the app gets stuck (it will fall back to a default state).
- The transitions required a new BuildFlavor proto to be included and bound to Dagger. For tests and the e2e test application component, a new testing-specific flavor has been introduced (though it's never expected to correspond to a distributable build of the app).
- The existing 'deprecation' package under the app module has been repurposed to be the location for all notices in the app (including the two new ones). We'll likely add more here in the future, as well (such as the GA deprecation notices).
- The new beta and GA flavors of the app have been added to the build_tests CI workflow--you can see the artifacts of the workflow in this PR or on develop after the PR is merged for downloads to the corresponding built AABs.
- The Kenya-specific alpha version of the app has the same 'alpha' build flavor as the default alpha version of the app since there doesn't seem to be any reason why they should differ here.
- Some places were updated to use Application instead of Context to make things a bit tighter in expectation for *what* type of context is available to use (for cases when casting to injector providers is used). This helps to avoid a particularly challenging bug that I ran into during test development wherein the wrong context was being used after creating a new application (since that application's application context was pointing to the original application). This means that ``Application.getApplicationContext()`` may actually not return itself in instrumentation cases. The changes help avoid this leading to actual issues in tests (but they should have no impact on the app's behavior in production environments).
- PersistentCacheStore had a few significant changes to better facilitate the needs of this PR's domain changes:
  - Changes in the store are only published to its in-memory cache (and subsequently notified to subscribers of the store) if the cache store actually changes.
  - The store's existing two priming functions were combined into a single new one with more flexibility as it can be used to tweak both changing the on-disk cache and publishing the in-memory one. The documentation around priming has also been significantly improved in terms of detail as I now think it's a legitimate use case that we'll want to use in more places in the future.
- InMemoryBlockingCache was updated to include publishing before-and-after values to its registered observer callback to avoid over-notifying for alterations to the cache that don't logically change it (this can actually lead to subtle differences in behavior that may not always be expected, so the change should improve robustness of the cache).

Details regarding test changes & exemptions:
- SplashActivityTest and AppStartupStateControllerTest were updated to enable parameterized testing since they have to test many different cases of build flavor transitions.
- The only test exemptions were: renames (deprecation notice classes), files that can't be tested (applications, application components, and listeners), and files that are not conventionally tested (test activities and fragment presenters).
- A new BUILD file was added for PersistentCacheStoreTest (which made development a bit easier when verifying some of the changes that ended up breaking the test suite).
- One of SplashActivityTest's error case tests were removed since it now seems impossible to trigger a situation that could actually cause it. Similarly, the pending case with 5-second timeout also seems impossible to test, though it's theoretically possible to occur if something catastrophically goes wrong in a production environment with a real clock.

## Essential Checklist

## For UI-specific PRs only
This PR absolutely should include screenshots & a demo video, but for the sake of optimizing time this will be done post-merge.


BenHenning:
Remaining TODOs before merge:
- [x] Finish description (enough--can be detailed later)
- [x] Review code
- [x] Bring up-to-date with develop

app/src/main/java/org/oppia/android/app/application/alpha/BUILD.bazel
```diff
@@ -11,8 +11,12 @@ kt_android_library(
         "AlphaApplicationComponent.kt",
         "AlphaOppiaApplication.kt",
     ],
-    visibility = ["//:oppia_binary_visibility"],
+    visibility = [
+        "//:oppia_binary_visibility",
+        "//app/src/test/java/org/oppia/android/app/application/alpha:__pkg__",
```
BenHenning: Check if this is actually needed.

utility/src/main/java/org/oppia/android/util/data/AsyncDataSubscriptionManager.kt
```diff
@@ -82,7 +82,9 @@ class AsyncDataSubscriptionManager @Inject constructor(
    */
   suspend fun notifyChange(id: Any) {
     // First, retrieve subscribers & the closure of child IDs to notify.
-    val subscriptions = subscriptionLock.withLock { computeSubscriptionClosure(id) }
+    val subscriptions = subscriptionLock.withLock {
+      computeSubscriptionClosure(id)
+    }
```
BenHenning: Revert.

utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseEventLogger.kt
```diff
@@ -69,9 +69,7 @@ class FirebaseEventLogger private constructor(
     private val networkConnectionUtil: NetworkConnectionUtil,
     private val eventBundleCreator: EventBundleCreator
   ) {
-    private val firebaseAnalytics by lazy {
-      FirebaseAnalytics.getInstance(context.applicationContext)
-    }
+    private val firebaseAnalytics by lazy { FirebaseAnalytics.getInstance(context) }
```
BenHenning: Inject Application, instead?

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
```diff
@@ -282,7 +284,10 @@ class ExplorationCheckpointController @Inject constructor(
       cacheStore
     }
 
-    cacheStore.primeInMemoryCacheAsync().invokeOnCompletion { throwable ->
+    cacheStore.primeInMemoryAndDiskCacheAsync(
+      updateMode = UpdateMode.UPDATE_IF_NEW_CACHE,
+      publishMode = PublishMode.DO_NOT_PUBLISH_TO_IN_MEMORY_CACHE
```
BenHenning: Should be publish.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -129,7 +131,10 @@ class ProfileManagementController @Inject constructor(
 
   // TODO(#272): Remove init block when storeDataAsync is fixed
   init {
-    profileDataStore.primeInMemoryCacheAsync().invokeOnCompletion {
+    profileDataStore.primeInMemoryAndDiskCacheAsync(
+      updateMode = UpdateMode.UPDATE_IF_NEW_CACHE,
+      publishMode = PublishMode.DO_NOT_PUBLISH_TO_IN_MEMORY_CACHE
```
BenHenning: Should be publish.

domain/src/main/java/org/oppia/android/domain/topic/StoryProgressController.kt
```diff
@@ -371,8 +373,11 @@ class StoryProgressController @Inject constructor(
       cacheStore
     }
 
-    cacheStore.primeInMemoryCacheAsync().invokeOnCompletion {
-      it?.let { it ->
+    cacheStore.primeInMemoryAndDiskCacheAsync(
+      updateMode = UpdateMode.UPDATE_IF_NEW_CACHE,
+      publishMode = PublishMode.DO_NOT_PUBLISH_TO_IN_MEMORY_CACHE
```
BenHenning: Should be publish.

BenHenning (commented):
Completed self-review.

app/src/main/java/org/oppia/android/app/application/alpha/BUILD.bazel
BenHenning: Done.

utility/src/main/java/org/oppia/android/util/data/AsyncDataSubscriptionManager.kt
BenHenning: Done.

utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseEventLogger.kt
BenHenning: Done.

domain/src/main/java/org/oppia/android/domain/exploration/lightweightcheckpointing/ExplorationCheckpointController.kt
BenHenning: Done.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
BenHenning: Done.

domain/src/main/java/org/oppia/android/domain/topic/StoryProgressController.kt
BenHenning: Done.

BenHenning:
Due to the significant time sensitivity of this PR, I'm force-merging it without review. I've self-reviewed it and plan to ask for a separate reviewer to take a look post-merge (#4567 is tracking this).

PR #4426, opened 2022-07-11
Fix #4438: Add profile login counter

### LINKED ISSUE #4438: Add profile login counter
This issue is a part of the GSoC project: Interactive Onboarding Flow.
We are planning to introduce certain onboarding flows in order to make the experience of a new user effortless in terms of having to decide which button to click next; and to make sure they are able to start learning quickly and take advantage of all the features that the app has to offer.

To make sure we do not throw too much information at the user at once, we check how many times this user has logged into the app and then decide what to show to them, thus streamlining how much information is being given to them and when.

JishnuGoyal:
## Explanation

Fixes #4438

This PR is a part of the GSoC project: Interactive Onboarding Flow which introduces a new field into the profile proto called number_of_logins. This is a variable which will be used to count the number of logins made by a user, based on which specific UI elements and spotlights could be shown to them.

The PR affects the ProfileManagementController -- we introduce a method called incrementLoginCount which is called when the user logs into their profile.

All the new changes are backed up by appropriate tests which check that the functionality works as expected.

## Essential Checklist

### DIFF
--- a/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
+++ b/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
@@ -26,6 +26,7 @@ import org.oppia.android.util.data.DataProvider
 import org.oppia.android.util.data.DataProviders
 import org.oppia.android.util.data.DataProviders.Companion.transform
 import org.oppia.android.util.data.DataProviders.Companion.transformAsync
+import org.oppia.android.util.data.DataProviders.Companion.transformNested
 import org.oppia.android.util.locale.OppiaLocale
 import org.oppia.android.util.platformparameter.LearnerStudyAnalytics
 import org.oppia.android.util.platformparameter.PlatformParameterValue
@@ -607,19 +608,47 @@ class ProfileManagementController @Inject constructor(
   }
 
   /**
-   * Log in to the user's Profile by setting the current profile Id and updating profile's last logged in time.
+   * Log in to the user's Profile by setting the current profile Id, updating profile's last logged in
+   * time and updating the total number of logins for the current profile Id.
    *
    * @param profileId the ID corresponding to the profile being logged into.
    * @return a [DataProvider] that indicates the success/failure of this login operation.
    */
   fun loginToProfile(profileId: ProfileId): DataProvider<Any?> {
-    return setCurrentProfileId(profileId).transformAsync(LOGIN_TO_PROFILE_PROVIDER_ID) {
-      return@transformAsync getDeferredResult(
+    return setCurrentProfileId(profileId).transformNested(LOGIN_TO_PROFILE_PROVIDER_ID) {
+      return@transformNested getDeferredResult(
         profileId,
         null,
-        updateLastLoggedInAsync(profileId)
-      )
+        incrementLoginCountAsync(profileId)
+      ).transformAsync {
+        return@transformAsync getDeferredResult(
+          profileId,
+          null,
+          updateLastLoggedInAsync(profileId)
+        )
+      }
     }
+  }
+
+  private fun incrementLoginCountAsync(profileId: ProfileId): Deferred<ProfileActionStatus> {
+//    val deferred =
+    return profileDataStore.storeDataWithCustomChannelAsync(updateInMemoryCache = true) {
+      val profile =
+        it.profilesMap[profileId.internalId] ?: return@storeDataWithCustomChannelAsync Pair(
+          it,
+          ProfileActionStatus.PROFILE_NOT_FOUND
+        )
+      val currentNumberOfLogins = profile.numberOfLogins
+      val updatedProfile = profile.toBuilder().setNumberOfLogins(currentNumberOfLogins + 1).build()
+      val profileDatabaseBuilder = it.toBuilder().putProfiles(
+        profileId.internalId,
+        updatedProfile
+      )
+      Pair(profileDatabaseBuilder.build(), ProfileActionStatus.SUCCESS)
+    }
+//    return dataProviders.createInMemoryDataProviderAsync(UPDATE_NUMBER_OF_LOGINS_PROVIDER_ID) {
+//      return@createInMemoryDataProviderAsync getDeferredResult(profileId, null, deferred)
+//    }
   }
 
   private fun setCurrentProfileId(profileId: ProfileId): DataProvider<Any?> {
--- a/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
+++ b/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
@@ -7,10 +7,6 @@ import android.media.ThumbnailUtils
 import android.net.Uri
 import android.provider.MediaStore
 import androidx.exifinterface.media.ExifInterface
-import java.io.File
-import java.io.FileOutputStream
-import javax.inject.Inject
-import javax.inject.Singleton
 import kotlinx.coroutines.Deferred
 import org.oppia.android.app.model.AppLanguage
 import org.oppia.android.app.model.AudioLanguage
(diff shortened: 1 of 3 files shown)


JishnuGoyal:
PTAL @BenHenning --  is the failing test a flake? Could you shed some light on why the test might be failing, since it doesn't show any error.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -607,19 +609,47 @@ class ProfileManagementController @Inject constructor(
...
    */
   fun loginToProfile(profileId: ProfileId): DataProvider<Any?> {
-    return setCurrentProfileId(profileId).transformAsync(LOGIN_TO_PROFILE_PROVIDER_ID) {
-      return@transformAsync getDeferredResult(
+    return setCurrentProfileId(profileId).transformNested(LOGIN_TO_PROFILE_PROVIDER_ID) {
+      return@transformNested getDeferredResult(
         profileId,
         null,
-        updateLastLoggedInAsync(profileId)
+        incrementLoginCountAsync(profileId)
```
BenHenning: Could we do this at the same time as updating the last logged in time? It seems like that would be simpler, and that they both are likely to always be updated at the same time.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -521,7 +521,7 @@ class ProfileManagementControllerTest {
   }
 
   @Test
-  fun testLoginToProfile_addProfiles_loginToProfile_checkGetProfileIdAndLoginTimestampIsCorrect() {
+  fun testLoginProfile_addProfiles_loginProfile_checkProfileId_Timestamp_NumberOfLoginsCorrect() {
```
```suggestion
  fun testLoginProfile_addedProfile_profileIdTimestampAndNumberOfLoginsCorrectlyDefaulted() {
```

Alternative that might be a bit clearer. Please also note the capitalization convention, and that 'login/logout' are nouns and 'log in/log out' are the verb forms.

Ditto below.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -532,6 +532,32 @@ class ProfileManagementControllerTest {
     val profile = monitorFactory.waitForNextSuccessfulResult(profileProvider)
     assertThat(profileManagementController.getCurrentProfileId().internalId).isEqualTo(2)
     assertThat(profile.lastLoggedInTimestampMs).isNotEqualTo(0)
+    assertThat(profile.numberOfLogins).isNotEqualTo(0)
```
BenHenning: Shouldn't numberOfLogins be one if the profile was logged in?

I'd think we'd want to check the profile before and after it's logged in to verify the initial login count state, and updated login count state.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -532,6 +532,32 @@ class ProfileManagementControllerTest {
...
+  }
+
+  @Test
+  fun testLoginToProfile_addProfile_loginToProfile_logoutProfile_checkNumberOfLoginsIsCorrect() {
```
BenHenning: This is potentially verifying more than one thing. Per my comment above, perhaps all we need to check here is that logging in, then out, then in results in a login count of 2? '0' and '1' cases can be tested in separate cases per the above.

That ensures the one-behavior-per-test convention is kept.

model/src/main/proto/profile.proto
```diff
@@ -70,6 +70,9 @@ message Profile {
   // are enabled as part of a user study. In non-user study cases this is expected to either be
   // empty or a meaningless value.
   string learner_id = 13;
+
+  // Represents the number of times a user has logged into their profile
```
BenHenning:
```suggestion
  // Represents the number of times a user has successfully logged into their profile.
```

Nit: punctuation. Also, it might be nice to clarify 'successfully' here to avoid any *potential* misinterpretation in the future.

BenHenning (changes requested):
Thanks @JishnuGoyal! Took a first pass. Note that CI seems to be failing--please make sure to fix these before sending the PR back for full review.

BenHenning:
Also @JishnuGoyal please note how I've edited your PR title & description for representing the fixed issue. Please follow this convention in future PRs.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
JishnuGoyal: thanks for the tip, done

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
JishnuGoyal: noted, thanks!

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
JishnuGoyal: This is how one-behaviour-per-test convention is being tried to follow in my latest commit: 

1) Check no_of_logins = 0 when profile is just created: **testAddProfile_addProfile_checkProfileIsAdded**
2) Check no_of_logins = 1 after first login: **testLoginProfile_addedProfile_profileIdTimestampAndNumberOfLoginsCorrectlyDefaulted**

3) Check no_of_logins after two logins: **testLoginToProfile_addProfile_loginToProfileTwice_checkNumberOfLoginsIsTwo**

JishnuGoyal: please let me know I still need to add more tests here

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
JishnuGoyal: done

model/src/main/proto/profile.proto
JishnuGoyal: done

JishnuGoyal:
image

This is the error. Could you shed some light on why this might be occurring?
PTAL @BenHenning

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -521,7 +521,7 @@ class ProfileManagementControllerTest {
   }
 
   @Test
-  fun testLoginToProfile_addProfiles_loginToProfile_checkGetProfileIdAndLoginTimestampIsCorrect() {
+  fun testLoginProfile_addedProfile_profileIdTimestampAndNumberOfLoginsCorrectlyDefaulted() {
```
BenHenning: Is this actually defaulted? It seems more like it's being updated because the account was logged in.

Also, it occurs to me the test is wrong since the last logged in time isn't changing. Perhaps we could advance the fake system clock prior to logging in so that there's a non-zero last logged in time to verify against?

model/src/main/proto/profile.proto
```diff
@@ -70,6 +70,9 @@ message Profile {
   // are enabled as part of a user study. In non-user study cases this is expected to either be
   // empty or a meaningless value.
   string learner_id = 13;
+
+  // Represents the number of times a user has successfully logged into their profile
```
BenHenning:
```suggestion
  // Represents the number of times a user has successfully logged into their profile.
```

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: I think the latest state is now clear, though I suggested a different name for 'defaulted' (since my understanding before might've been incorrect given the confusion of when the login count was actually meant to be incremented in each test case).

BenHenning (commented):
Thanks @JishnuGoyal! I think you missed part of my 'punctuation' comment in the proto file, but I left a new one.

Otherwise, I have mostly nits left--PTAL.

model/src/main/proto/profile.proto
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
JishnuGoyal: The last logged in time is actually non zero. Could you recheck if I missed something?
Also this test check the profile data after logging in once, so I renamed the test. PTAL.

JishnuGoyal:
PTAL @BenHenning

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: As discussed during our meeting, I apparently misread this. Thanks for the correction. :)

BenHenning (approved):
Thanks @JishnuGoyal. This LGTM!

PR #4462, opened 2022-07-23
Fixes #4340: Introduces Performance Metrics logging

### LINKED ISSUE #4340: Add performance metrics logging during critical user journeys
This issue tracks the implementation of metric logging during critical user journeys in the application.

Sarthak2601:
## Explanation

Fixes #4340 
Introduces performance metrics logging at application startup and via background workers. 

Things to note: 

1. Implementation of a ScreenName for each activity
2. Tests for verifying that every activity's intent has the screen name value. 
3. Application startup logging at one place:
I've moved apk size logging and storage usage logging from PerformanceMetricsLogger to the new ActivityLifecycleObserver. This means that the new observer is logging all app startup metrics from one place. 

## Essential Checklist


domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,79 @@
...
+import android.app.Application
+import android.os.Bundle
+import org.oppia.android.domain.oppialogger.ApplicationStartupListener
+import org.oppia.android.util.system.OppiaClock
+import javax.inject.Inject
+import javax.inject.Singleton
+
+private const val APPLICATION_STARTUP_SCREEN = "application_startup_screen"
+
+/** Observer that observes activity lifecycle. */
```
BenHenning: Perhaps clarify that this observer is meant to log analytics events for given activity lifecycle events? The current doc comment is a bit too non-specific.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,79 @@
...
+    if (!isStartupLatencyLogged) {
+      performanceMetricsLogger.logStartupLatency(
+        initialTimestamp,
+        activity.javaClass.simpleName
+      )
+      isStartupLatencyLogged = true
+    }
+    performanceMetricsLogger.logMemoryUsage(
+      activity.javaClass.simpleName
+    )
```
BenHenning: Can this fit on one line?

Also, unfortunately I don't think this is going to work. For most activities, this name is going to be obfuscated after we run Proguard on production builds (you can probably observe this by building a production flavor of the app, e.g. ``bazel build //:oppia_alpha``).

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,79 @@
...
+    application.registerActivityLifecycleCallbacks(this)
+    performanceMetricsLogger.logApkSize(APPLICATION_STARTUP_SCREEN)
+    performanceMetricsLogger.logStorageUsage(APPLICATION_STARTUP_SCREEN)
+  }
+
+  override fun onActivityResumed(activity: Activity) {
+    currentScreen = activity.javaClass.simpleName
+    if (!isStartupLatencyLogged) {
+      performanceMetricsLogger.logStartupLatency(
+        initialTimestamp,
```
BenHenning: Should we maybe be logging the delta (i.e. the number of milliseconds to startup) instead of the timestamp?

scripts/assets/test_file_exemptions.textproto
```diff
@@ -815,4 +815,4 @@ exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/perfor
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/performancemetrics/PerformanceMetricsEventLogger.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/performancemetrics/PerformanceMetricsLogScheduler.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorkerFactory.kt"
-exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/PerformanceMetricsLoggerModule.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverModule.kt"
```
BenHenning: We should be testing the module, but I think I left this as a comment in the previous PR.

BenHenning (changes requested):
Thanks @Sarthak2601. I took an initial pass and had some thoughts--PTAL. I think these comments may result in some broader changes, so I'll take a more thorough review in my next pass once these ones are addressed.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: The delta was actually getting logged in the logger directly, however I think that might lead to less-accurate readings than the ones we'll get from the observer. Moved delta calculation here.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: shifted this to a new approach that we discussed in chat.

scripts/assets/test_file_exemptions.textproto
Sarthak2601: Will fix it in the previous PR first and it'll reflect here.

app/src/main/java/org/oppia/android/app/profile/AddProfileActivity.kt
```diff
@@ -26,6 +26,7 @@ class AddProfileActivity : InjectableAppCompatActivity() {
     super.onCreate(savedInstanceState)
     (activityComponent as ActivityComponentImpl).inject(this)
     addProfileFragmentPresenter.handleOnCreate()
+    intent = addProfileFragmentPresenter.getCurrentAppScreenNameIntent()
```
BenHenning: We should be adding the value during intent creation, not overriding it here (as that may have unexpected behaviors--we rely on the intent being present in a lot of cases.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.content.Intent
+import org.oppia.android.app.model.CurrentAppScreen
+import org.oppia.android.app.model.ScreenName
+import org.oppia.android.util.extensions.putProtoExtra
+import javax.inject.Inject
+
+private const val CURRENT_APP_SCREEN_NAME_KEY = "key"
+
+/** Utility to provide utilities related to wrapping the current app-screen name in an intent. */
+class CurrentAppScreenNameWrapper @Inject constructor() {
```
BenHenning: Perhaps: CurrentAppScreenNameIntentDecorator?

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
```diff
@@ -0,0 +1,24 @@
...
+
+  /** Returns an intent packed with [CurrentAppScreen]. */
+  fun getCurrentAppScreenNameIntent(screenName: ScreenName): Intent = Intent().apply {
```
BenHenning: Per earlier suggestions, this could be: decorateWithScreenName

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
```diff
@@ -0,0 +1,24 @@
...
+    this.putProtoExtra(
+      CURRENT_APP_SCREEN_NAME_KEY,
+      CurrentAppScreen.newBuilder().setScreenName(screenName).build()
+    )
+  }
+
+  /** Returns the intent-key that'll be used to retrieve [ScreenName]. */
+  fun getCurrentAppScreenNameIntentKey(): String = CURRENT_APP_SCREEN_NAME_KEY
```
BenHenning: You could add a getter to retrieve the screen name for a given intent and keep the key private. That might actually be a bit cleaner.

app/src/main/java/org/oppia/android/app/profile/AddProfileActivity.kt
rt4914: +1

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivity.kt
```diff
@@ -24,5 +24,6 @@ class ProfileChooserActivity : InjectableAppCompatActivity() {
     super.onCreate(savedInstanceState)
     (activityComponent as ActivityComponentImpl).inject(this)
     profileChooserActivityPresenter.handleOnCreate()
+    intent = profileChooserActivityPresenter.getCurrentAppScreenNameIntent()
```
rt4914: Same a another comment thread by Ben.

rt4914 (changes requested):
@Sarthak2601 PTAL

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
Sarthak2601: But even if we go that way, we'd still need the key to be public for `putProtoExtra` right?

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
Sarthak2601: done.

app/src/main/java/org/oppia/android/app/profile/AddProfileActivity.kt
Sarthak2601: noted, done.

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivity.kt
Sarthak2601: done.

Sarthak2601:
Note: I'm waiting on one of the comments regarding intent key to be resolved, post which I'll implement this approach for the rest of the implementation. 

cc: @BenHenning

app/src/main/java/org/oppia/android/app/profile/AddProfileActivity.kt
```diff
@@ -18,6 +21,10 @@ class AddProfileActivity : InjectableAppCompatActivity() {
     fun createAddProfileActivityIntent(context: Context, colorRgb: Int): Intent {
       val intent = Intent(context, AddProfileActivity::class.java)
       intent.putExtra(ADD_PROFILE_COLOR_RGB_EXTRA_KEY, colorRgb)
+      intent.putProtoExtra(
+        CurrentAppScreenNameIntentDecorator.getCurrentAppScreenNameIntentKey(),
+        CurrentAppScreenNameIntentDecorator.decorateWithScreenName(ScreenName.ADD_PROFILE_ACTIVITY)
```
BenHenning: Why not have something like ``currentAppscreenNameDecorator.decorateWithScreenName(ScreenName.ADD_PROFILE_ACTIVITY)``? Then you don't need to expose the key.

You could also use an instance extension function, but they're a bit trickier. Example:

```kotlin
class CurrentAppSCreenNameIntentDecorator {
  fun Intent.decorateWithScreenName(screenName: ScreenName) {...}
}

...

currentAppScreenNameIntentDecorator.run {
  intent.decorateWithScreenName(ScreenName.ADD_PROFILE_ACTIVITY)
}
```

Or, if the extension can go in the decorator's companion object, then you could just import the extension and call it (no ``run`` needed).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
```diff
@@ -103,8 +102,13 @@ class ActivityLifecycleObserverTest {
   @Test
   fun testObserver_onFirstActivityResume_verifyCurrentScreenReturnsCorrectValue() {
     val activity = Robolectric.buildActivity(Activity::class.java).get()
-    activity.intent = currentAppScreenNameWrapper
-      .getCurrentAppScreenNameIntent(ScreenName.ACTIVITY_NAME_UNSPECIFIED)
+    activity.intent = Intent().apply {
+      this.putProtoExtra(
+        CurrentAppScreenNameIntentDecorator.getCurrentAppScreenNameIntentKey(),
+        CurrentAppScreenNameIntentDecorator
+          .decorateWithScreenName(ScreenName.ACTIVITY_NAME_UNSPECIFIED)
```
BenHenning: Is this actually needed? I'd expect retrieval to default to unspecified if the screen name isn't present.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -53,7 +52,7 @@ class ActivityLifecycleObserver @Inject constructor(
 
   override fun onActivityResumed(activity: Activity) {
     val currentAppScreen = activity.intent.getProtoExtra(
```
BenHenning: I think you could have a simple getter on the decorator which injects the current activity, retrieves the intent, and then retrieves the screen name from that intent. That would simplify retrieval.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/PerformanceMetricsController.kt
```diff
@@ -46,7 +46,7 @@ class PerformanceMetricsController @Inject constructor(
    */
   fun logPerformanceMetricsEvent(
     timestamp: Long,
-    currentScreen: OppiaMetricLog.CurrentScreen,
+    currentScreen: String,
```
BenHenning: Why this change?

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
```diff
@@ -8,11 +8,15 @@ import com.google.common.util.concurrent.SettableFuture
...
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
+import org.oppia.android.domain.oppialogger.analytics.ActivityLifecycleObserver
+import org.oppia.android.domain.oppialogger.analytics.PerformanceMetricsLogger
 import org.oppia.android.domain.util.getStringFromData
 import org.oppia.android.util.logging.ConsoleLogger
 import org.oppia.android.util.threading.BackgroundDispatcher
 import javax.inject.Inject
 
+private const val BACKGROUND_WORKER_SCREEN = "background_worker_screen"
```
BenHenning: Could you dedicate one of the CurrentScreen enum constants for this case to avoid needing to use strings everywhere? Stronger types are preferred where possible.

model/src/main/proto/performance_metrics.proto
```diff
@@ -29,7 +29,7 @@ message OppiaMetricLog {
   LoggableMetric loggable_metric = 7;
 
   // The current screen where the event is getting logged.
-  CurrentScreen current_screen = 8;
+  string current_screen = 8;
```
BenHenning: Could we use ScreenName or CurrentAppScreen here, instead?

model/src/main/proto/screens.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+enum ScreenName {
```
BenHenning, line 9: This & its constants should be documented.

model/src/main/proto/screens.proto
```diff
@@ -0,0 +1,18 @@
...
+  ACTIVITY_NAME_UNSPECIFIED = 0;
```
BenHenning: Should be SCREEN_NAME_UNSPECIFIED (per convention).

model/src/main/proto/screens.proto
```diff
@@ -0,0 +1,18 @@
...
+  SPLASH_ACTIVITY = 1;
+  PROFILE_CHOOSER_ACTIVITY = 2;
+  ADD_PROFILE_ACTIVITY = 3;
+  HOME_ACTIVITY = 4;
+}
+
+message CurrentAppScreen {
```
BenHenning, line 158: This & the field member should be documented.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
BenHenning: Not if we take in the intent object (since then the key is only needed by the decorator to add/retrieve it).

BenHenning (changes requested):
Thanks @Sarthak2601! Took a full review--PTAL.

rt4914 (approved):
LGTM for my code-owners.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: I've introduced an extension function which will act as a getter. For consistency between screen name packing and retrieval, I've opted to use the intent instead of injecting the activity. Does that make sense?

app/src/main/java/org/oppia/android/app/profile/AddProfileActivity.kt
Sarthak2601: Thanks! 
I've implemented the extension approach, PTAL :)

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/PerformanceMetricsController.kt
Sarthak2601: Modified this to take in ScreenName.

model/src/main/proto/performance_metrics.proto
Sarthak2601: Done.

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameWrapper.kt
Sarthak2601: good point, done.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: Yes, but since we're decorating the activity's intent with a screen name -- like in case of an actual activity, we need to set an initial value. The getter will default to an unspecified value but I don't think the setter should take null as a value as of now as it'll defeat the entire purpose of adding screen name to the activity. That being said, do you think we should support a null case for the setter too? 

However, since it'll return an unspecified value even when null, I've modified the case to contain home_activity as the screen name.

app/src/main/java/org/oppia/android/app/help/HelpActivity.kt
```diff
@@ -82,6 +84,7 @@ class HelpActivity :
       val intent = Intent(context, HelpActivity::class.java)
       intent.putExtra(NAVIGATION_PROFILE_ID_ARGUMENT_KEY, profileId)
       intent.putExtra(BOOL_IS_FROM_NAVIGATION_DRAWER_EXTRA_KEY, isFromNavigationDrawer)
+      intent.decorateWithScreenName(HELP_ACTIVITY)
```
BenHenning, line 101: I'm starting to wonder for the generic perspective: there are three cases where this can break-down as code changes or is added:
1. An activity uses the wrong enum constant or doesn't include it
2. A new constant is added
3. A new activity is added without a corresponding constant

I think (2) is handled by using an exhaustive ``when`` for the screen name constant, and (3) can be handled by introducing a new static check which verifies that all activities in the manifest have a corresponding screen name.

(1) is trickier, though. My initial thought is having a test for each activity to verify that its screen name is correct, but that's...messy. It's also difficult to ensure that the test itself is added when a new activity is added. We are planning to move all activity intent creation over to a central activity router in the future--that might be a better place to perform this verification but it's outside the scope of your PR.

I think it's essential that all three are properly addressed since otherwise we can easily miss new activities correctly hooking into the analytics system. Do you have any thoughts on these (particularly (1))?

app/src/main/java/org/oppia/android/app/profileprogress/ProfilePictureActivity.kt
```diff
@@ -27,11 +29,12 @@ class ProfilePictureActivity : InjectableAppCompatActivity() {
...
-      val intent = Intent(context, ProfilePictureActivity::class.java)
-      intent.putExtra(
-        PROFILE_PICTURE_ACTIVITY_PROFILE_ID_KEY, internalProfileId
-      )
-      return intent
+      return Intent(context, ProfilePictureActivity::class.java).apply {
+        decorateWithScreenName(PROFILE_PICTURE_ACTIVITY)
+        putExtra(
+          PROFILE_PICTURE_ACTIVITY_PROFILE_ID_KEY, internalProfileId
+        )
```
BenHenning: Nit: can fit on one line (small existing code cleanup), but optional. Just seems nice to do since this code is changing, anyway.

app/src/sharedTest/java/org/oppia/android/app/mydownloads/MyDownloadsActivityTest.kt
```diff
@@ -117,6 +119,16 @@ class MyDownloadsActivityTest {
...
   }
 
+  @Test
+  fun testActivity_getIntent_verifyScreenNameInIntent() {
+    val screenName = MyDownloadsActivity.createMyDownloadsActivityIntent(
+      ApplicationProvider.getApplicationContext(),
+      1
+    ).getCurrentAppScreenName()
+
+    assertThat(screenName).isEqualTo(ScreenName.MY_DOWNLOAD_ACTIVITY)
```
BenHenning: Should the constant be MY_DOWNLOADS_ACTIVITY, instead?

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
```diff
@@ -144,6 +146,13 @@ class ResumeLessonActivityTest {
     ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
   }
 
+  @Test
+  fun testActivity_getIntent_verifyScreenNameInIntent() {
```
BenHenning: For this & the other test names, perhaps ``testActivity_createIntent_verifyScreenNameInIntent`` would be clearer?

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
```diff
@@ -102,7 +102,7 @@ class MetricLogSchedulingWorker private constructor(
   private fun schedulePeriodicUiMetricLogging(): Result {
     return try {
       val currentScreen = activityLifecycleObserver.getCurrentScreen()
-      if (currentScreen != null) {
+      if (currentScreen != SCREEN_NAME_UNSPECIFIED) {
         performanceMetricsLogger.logMemoryUsage(currentScreen)
       } else {
         performanceMetricsLogger.logMemoryUsage(BACKGROUND_WORKER_SCREEN)
```
BenHenning: Doesn't the lifecycle observer already know if it's in the background? It may be cleaner just to have the observer return the exact correct screen (including background) rather than needing to check it like this (plus, unspecified might actually indicate an incorrectly configured activity which is not necessarily in the background).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
```diff
@@ -102,17 +101,11 @@ class ActivityLifecycleObserverTest {
   @Test
   fun testObserver_onFirstActivityResume_verifyCurrentScreenReturnsCorrectValue() {
     val activity = Robolectric.buildActivity(Activity::class.java).get()
```
BenHenning: Perhaps use TestActivity, instead? Then you could use the standard AndroidX test utilities (like launching via an ActivityScenario) and avoid the direct Robolectric dependency.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
```diff
@@ -102,17 +101,11 @@ class ActivityLifecycleObserverTest {
   @Test
   fun testObserver_onFirstActivityResume_verifyCurrentScreenReturnsCorrectValue() {
```
BenHenning, line 130: I suggest also adding tests for an unknown activity producing an unspecified value.

model/src/main/proto/screens.proto
```diff
@@ -5,14 +5,154 @@ package model;
...
+  // Screen name value for the scenario when the force network type activity is visible to the user.
+  FORCE_NETWORK_TYPE_ACTIVITY = 45;
+
+  // Screen name value for the scenario when the admin pin activity is visible to the user.
+  ADMIN_PIN_ACTIVITY = 46;
 }
 
+// Defines the current visible ui screen of the application.
 message CurrentAppScreen {
+  //S creen name value depicting the name of the currently visible ui screen.
```
BenHenning: Space is in the wrong place--please fix.

model/src/main/proto/screens.proto
```diff
@@ -5,14 +5,154 @@ package model;
...
+  ADMINISTRATOR_CONTROLS_ACTIVITY = 7;
+
+  // Screen name value for the scenario when the profile and device id activity is visible to the user.
+  PROFILE_AND_DEVICE_ID_ACTIVITY = 8;
+
+  // Screen name value for the scenario when the completed story list activity is visible to the user.
+  COMPLETED_STORY_LIST_ACTIVITY = 9;
+
+  // Screen name value for the scenario when the faq single activity is visible to the user.
+  FAQ_SINGLE_ACTIVITY= 10;
```
BenHenning: Nit: need space before ``=``.

model/src/main/proto/screens.proto
```diff
@@ -5,14 +5,154 @@ package model;
...
+  MARK_STORIES_COMPLETED_ACTIVITY = 44;
+
+  // Screen name value for the scenario when the force network type activity is visible to the user.
+  FORCE_NETWORK_TYPE_ACTIVITY = 45;
+
+  // Screen name value for the scenario when the admin pin activity is visible to the user.
+  ADMIN_PIN_ACTIVITY = 46;
 }
 
+// Defines the current visible ui screen of the application.
```
BenHenning: Nit: "UI" (capitalization). Ditto below.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -1,17 +1,31 @@
...
+  }
 
-  /** Returns an intent packed with [CurrentAppScreen]. */
-  fun decorateWithScreenName(screenName: ScreenName): CurrentAppScreen =
-    CurrentAppScreen.newBuilder().setScreenName(screenName).build()
-
-  /** Returns the intent-key that'll be used to retrieve [ScreenName]. */
-  fun getCurrentAppScreenNameIntentKey(): String = CURRENT_APP_SCREEN_NAME_KEY
+  /** Returns [ScreenName] after unpacking intent. */
+  fun Intent.getCurrentAppScreenName(): ScreenName {
```
BenHenning: Suggest: ``extractCurrentAppScreenName``, instead, since this is doing more than trivial 'getting'.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
```diff
@@ -203,7 +202,7 @@ class EventBundleCreatorTest {
     assertThat(bundle).string("memory_tier").isEqualTo("high_memory")
     assertThat(bundle).string("storage_tier").isEqualTo("high_storage")
     assertThat(bundle).string("network_type").isEqualTo("wifi")
-    assertThat(bundle).string("current_screen").isEqualTo("home_screen")
+    assertThat(bundle).string("current_screen").isEqualTo("test_screen_unspecified")
```
BenHenning: Is this correct? Shouldn't it be "screen_name_unspecified"?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
BenHenning: Yep, the new approach looks good. Thanks!

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
BenHenning: I'm not sure that I understand the 'null' part of the argument. If the proto is absent from the intent, the decorator will extract a default proto instance (which will always correspond to ACTIVITY_NAME_UNSPECIFIED).

I think having a specific name is better here, but perhaps it's worth testing this default scenario in the decorator's test suite.

scripts/assets/test_file_exemptions.textproto
BenHenning: Reminder to remove this once it's addressed and merged into this PR.

app/src/main/java/org/oppia/android/app/help/HelpActivity.kt
BenHenning: Another thought I had on potentially addressing (1): you could add a regex check to verify that all *ActivityTest.kt files include a test called "testActivity_getIntent_verifyScreenNameInIntent". It's a bit hacky, but it's a reasonable stopgap, I think, even if it requires a human to verify it.

If we go this route, I think it's worth filing a new issue and adding a TODO to move the check over something more robust and what that might be (probably when we have the activity router implementation).

BenHenning (changes requested):
THanks @Sarthak2601! Took another pass. The PR is pretty close to LGTM, but I had one high-level comment. Beyond that, please also make sure all CI checks are passing before sending this back for review.

app/src/sharedTest/java/org/oppia/android/app/resumelesson/ResumeLessonActivityTest.kt
Sarthak2601: done.

app/src/sharedTest/java/org/oppia/android/app/mydownloads/MyDownloadsActivityTest.kt
Sarthak2601: done.

app/src/main/java/org/oppia/android/app/profileprogress/ProfilePictureActivity.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

model/src/main/proto/screens.proto
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
Sarthak2601: done.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: done.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: done.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: done.

scripts/assets/test_file_exemptions.textproto
Sarthak2601: done.

app/src/main/java/org/oppia/android/app/help/HelpActivity.kt
Sarthak2601: Yes, these are a few things which might come up in the future. I've introduced the exhaustive `when` along with 2 regex checks to verify: 
- Usage of ScreenName in all activities 
- Usage of "testActivity..." in all activityTests. 

I've also added tests to verify regex changes.

Sarthak2601: Filed #4557 as well.

app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileListActivityTest.kt
```diff
@@ -105,14 +107,21 @@ class ProfileListActivityTest {
...
+  }
+
   @Test
   fun testProfileListActivity_hasCorrectActivityLabel() {
     activityTestRule.launchActivity(createProfileListActivityIntent())
     val title = activityTestRule.activity.title
 
     // Verify that the activity label is correct as a proxy to verify TalkBack will announce the
     // correct string when it's read out.
-    Truth.assertThat(title).isEqualTo(context.getString(R.string.profile_list_activity_title))
```
BenHenning, line 117: Thanks for cleaning up this existing case!

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
```diff
@@ -219,6 +223,15 @@ class QuestionPlayerActivityTest {
     testCoroutineDispatchers.unregisterIdlingResource()
   }
 
+  @Test
+  fun testActivity_createIntent_verifyScreenNameInIntent() {
+    val currentScreenName = QuestionPlayerActivity.createQuestionPlayerActivityIntent(
+      context, ArrayList(SKILL_ID_LIST), profileId
```
BenHenning, line 231:
```suggestion
      context, listOf(SKILL_ID_LIST), profileId
```

Nit--more Kotlin-idiomatic.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,81 @@
...
+   * Returns a boolean flag that makes sure that startup latency is logged only once in the entire
+   * application lifecycle.
+   */
+  private var isStartupLatencyLogged: Boolean = false
+
+  private var currentScreen: ScreenName = SCREEN_NAME_UNSPECIFIED
+
+  /**
+   * Returns the current active UI screen that's visible to the user. If UI is in-active or the app
+   * is in background, this value returns [SCREEN_NAME_UNSPECIFIED].
```
BenHenning:
```suggestion
   * Returns the current active UI screen that's visible to the user.
   *
   * If the UI is inactive or the app backgrounded, [SCREEN_NAME_UNSPECIFIED] is returned.
```

BenHenning: I'm not actually sure about this policy. I feel like we should reserve UNSPECIFIED to truly be the cases where we don't know. In this case, we actually do--the case is that the app is backgrounded. I think it'd be cleaner and more robust to have a dedicated constant for the background case.

Edit: actually I'm now unsure about this. It seems there is a background screen actually--does this documentation need to be updated?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,81 @@
...
+   */
+  fun getCurrentScreen(): ScreenName = currentScreen
+
+  override fun onCreate() {
+    initialTimestamp = oppiaClock.getCurrentTimeMs()
+    application.registerActivityLifecycleCallbacks(this)
+    performanceMetricsLogger.logApkSize(currentScreen)
+    performanceMetricsLogger.logStorageUsage(currentScreen)
```
BenHenning: Is there any issue with these being called on the main thread?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,81 @@
...
+  private var currentScreen: ScreenName = SCREEN_NAME_UNSPECIFIED
+
+  /**
+   * Returns the current active UI screen that's visible to the user. If UI is in-active or the app
+   * is in background, this value returns [SCREEN_NAME_UNSPECIFIED].
+   */
+  fun getCurrentScreen(): ScreenName = currentScreen
+
+  override fun onCreate() {
+    initialTimestamp = oppiaClock.getCurrentTimeMs()
```
BenHenning: Perhaps: ``appStartTimeMillis`` might be a bit clearer?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
```diff
@@ -0,0 +1,81 @@
...
+  override fun onActivityStarted(activity: Activity) {}
+
+  override fun onActivityStopped(activity: Activity) {}
+
+  override fun onActivitySaveInstanceState(activity: Activity, bundle: Bundle) {}
+
+  override fun onActivityDestroyed(activity: Activity) {}
+
+  private fun getStartupLatency(initialTimestamp: Long): Long =
+    oppiaClock.getCurrentTimeMs() - initialTimestamp
```
BenHenning:
```suggestion
  private fun getStartupLatencyMillis(initialTimestampMillis: Long): Long =
    oppiaClock.getCurrentTimeMs() - initialTimestampMillis
```

(Just to make sure there's a unit on time values/methods).

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
```diff
@@ -74,7 +79,8 @@ class MetricLogSchedulingWorker private constructor(
 
   private fun schedulePeriodicBackgroundMetricLogging(): Result {
     return try {
-      // TODO(#4340): Add functionality to log cpu and network usage performance metrics.
+      performanceMetricsLogger.logNetworkUsage(BACKGROUND_SCREEN)
+      // TODO(#4340): Add functionality to log cpu usage performance metrics.
```
BenHenning: The PR is marking this issue as fixed, but per this it's not being fully resolved. Suggest updating the PR or filing a new issue & updating this TODO.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
```diff
@@ -0,0 +1,318 @@
...
+      val currentScreen = activityLifecycleObserver.getCurrentScreen()
+      val startupLatencyEvents =
+        fakePerformanceMetricsEventLogger.getMostRecentPerformanceMetricsEvents(3)
+      val startupLatencyEvent = startupLatencyEvents[1]
+
+      assertThat(startupLatencyEvent.loggableMetric.loggableMetricTypeCase).isEqualTo(
+        STARTUP_LATENCY_METRIC
+      )
+      assertThat(startupLatencyEvent.timestampMillis).isEqualTo(TEST_TIMESTAMP_IN_MILLIS_TWO)
+      assertThat(startupLatencyEvent.currentScreen).isEqualTo(currentScreen)
```
BenHenning: Perhaps verify the actual screen--ditto elsewhere. Otherwise, we're only testing consistency and not direct state.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
```diff
@@ -0,0 +1,318 @@
...
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = ActivityLifecycleObserverTest.TestApplication::class)
+class ActivityLifecycleObserverTest {
+
+  @get:Rule
+  var activityRule =
+    ActivityScenarioRule<TextInputActionTestActivity>(
+      TextInputActionTestActivity.createIntent(ApplicationProvider.getApplicationContext()).apply {
+        decorateWithScreenName(ScreenName.BACKGROUND_SCREEN)
```
BenHenning: Should this maybe be a non-background screen? It's conceptually invalid for BACKGROUND_SCREEN to ever be used with an activity, and it sort of makes the tests below more confusing when trying to determine which tests are verifying a foreground vs. background screen state.

scripts/BUILD.bazel
```diff
@@ -61,6 +61,7 @@ REGEX_PATTERN_CHECK_ASSETS = generate_regex_assets_list_from_text_protos(
     name = "regex_asset_files",
     file_content_validation_file_names = ["file_content_validation_checks"],
     filepath_pattern_validation_file_names = ["filename_pattern_validation_checks"],
+    screen_name_presence_validation_file_names = ["screen_name_presence_validation_checks"],
```
BenHenning: Why not include this in the existing file_content_validation_checks.textproto? That seems simpler since the checks seem the same as the generic regex ones unless I'm missing something.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import android.content.Intent
+import org.oppia.android.app.model.CurrentAppScreen
+import org.oppia.android.app.model.ScreenName
+import org.oppia.android.util.extensions.getProtoExtra
+import org.oppia.android.util.extensions.putProtoExtra
+
+private var CURRENT_APP_SCREEN_NAME_KEY = "CURRENT_APP_SCREEN_NAME_KEY"
+
+/** Utility to provide utilities related to wrapping the current app-screen name in an intent. */
```
BenHenning:
```suggestion
/** Decorator for wrapping an activity's [ScreenName] within its intent. */
```

Suggested changes for clarity and conciseness.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+object CurrentAppScreenNameIntentDecorator {
+  /**
+   * Packs the intent with a [CurrentAppScreen] proto object that sets [screenName] as the current
+   * screen.
```
BenHenning, line 15:
```suggestion
   * screen.
   *
   * [extractCurrentAppScreenName] should be used for retrieving the screen name later.
```

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+   */
+  fun Intent.decorateWithScreenName(screenName: ScreenName?) {
+    if (screenName != null) {
+      this.putProtoExtra(
```
BenHenning: Nit here & below: omit 'this' since it isn't needed.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+
+private var CURRENT_APP_SCREEN_NAME_KEY = "CURRENT_APP_SCREEN_NAME_KEY"
+
+/** Utility to provide utilities related to wrapping the current app-screen name in an intent. */
+object CurrentAppScreenNameIntentDecorator {
+  /**
+   * Packs the intent with a [CurrentAppScreen] proto object that sets [screenName] as the current
+   * screen.
+   */
+  fun Intent.decorateWithScreenName(screenName: ScreenName?) {
```
BenHenning: Why allow a nullable ``ScreenName``? We usually don't use nullable protos since they have reasonable default states.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+    if (screenName != null) {
+      this.putProtoExtra(
+        CURRENT_APP_SCREEN_NAME_KEY,
+        CurrentAppScreen.newBuilder().setScreenName(screenName).build()
+      )
+    }
+  }
+
+  /** Returns [ScreenName] after unpacking intent. */
```
BenHenning:
```suggestion
  /** Returns the [ScreenName] packed in the intent via [decorateWithScreenName]. */
```

Also for clarity since it's nice to tie the read & write methods together in documentation.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
```diff
@@ -552,10 +553,55 @@ class EventBundleCreator @Inject constructor(
     OppiaMetricLog.NetworkType.UNRECOGNIZED -> "unknown_network_type"
   }
 
-  private fun OppiaMetricLog.CurrentScreen.toAnalyticsName() = when (this) {
-    OppiaMetricLog.CurrentScreen.SCREEN_UNSPECIFIED -> "unspecified_current_screen"
-    OppiaMetricLog.CurrentScreen.HOME_SCREEN -> "home_screen"
-    OppiaMetricLog.CurrentScreen.UNRECOGNIZED -> "unknown_screen_name"
-    // TODO(#4340): Add support for all screens which are going to be used for metric logging.
+  private fun ScreenName.toAnalyticsName() = when (this) {
```
BenHenning, line 563: Are there tests for verifying each of these translate correctly? If not, this is a good potential use for a parameterized test to keep things relatively simple. We do enum conversion in a few other parameterized tests elsewhere in the codebase that you could reference.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
+package org.oppia.android.util.logging
+
+import android.content.Intent
+import org.oppia.android.app.model.CurrentAppScreen
+import org.oppia.android.app.model.ScreenName
+import org.oppia.android.util.extensions.getProtoExtra
+import org.oppia.android.util.extensions.putProtoExtra
+
+private var CURRENT_APP_SCREEN_NAME_KEY = "CURRENT_APP_SCREEN_NAME_KEY"
```
BenHenning: This can be a ``const val`` I believe.

BenHenning (commented):
Thanks @Sarthak2601! Took another pass. I think I essentially only have nits left, minus the bit about the regex checks (which is more about simplifying things). I think this PR is very close to being wrapped up.

Also, I think you have lint tests failing per CI. I suggest addressing that failure before sending the PR back.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: Yes, I'll update the documentation.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: removed.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

app/src/sharedTest/java/org/oppia/android/app/settings/profile/ProfileListActivityTest.kt
Sarthak2601: no problem!

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: done.

Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: done.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: true. I've changed this to _`HOME_ACTIVITY`_ -- does that work?

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
Sarthak2601: agreed -- done!

domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorker.kt
Sarthak2601: #4466 already corresponds to this -- I've updated the todo.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
Sarthak2601: done.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserver.kt
Sarthak2601: modified, PTAL.

scripts/BUILD.bazel
Sarthak2601: done.

app/src/sharedTest/java/org/oppia/android/app/topic/questionplayer/QuestionPlayerActivityTest.kt
BenHenning: Ah, it wasn't done but it appears that's because an ``ArrayList`` is required here.

scripts/assets/file_content_validation_checks.textproto
```diff
@@ -371,4 +373,25 @@ file_content_checks {
   prohibited_content_regex: "@color/(?!color_def_).+"
   failure_message: "Only colors from color_defs.xml may be used in color_palette.xml."
 }
-
+file_content_checks {
+  file_path_regex: ".+?ActivityTest.kt"
+  failure_message: "Screen Name test not added."
```
BenHenning: For this & below, could you revise these failure messages to also provide some instruction for how to fix the issue? See other failure messages in this file for references.

scripts/assets/file_content_validation_checks.textproto
```diff
@@ -371,4 +373,25 @@ file_content_checks {
...
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/markchapterscompleted/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/marktopicscompleted/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/forcenetworktype/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/devoptions/markstoriescompleted/testing.+?\\.kt"
+  exempted_file_patterns: "app/src/main/java/org/oppia/android/app/player/state/testing.+?\\.kt"
+  exempted_file_patterns: "testing.+?\\.kt"
+  required_content_regex: "org\\.oppia\\.android\\.app\\.model\\.ScreenName"
```
BenHenning: Could we perhaps match against passing a ScreenName into the needed function? Both can technically be "faked" but this one I'm concerned will have more false positives.

scripts/src/java/org/oppia/android/scripts/proto/file_content_validation_checks.proto
```diff
@@ -27,4 +27,7 @@ message FileContentCheck {
 
   // Regex patterns for all file/file paths that should be exempted for this check.
   repeated string exempted_file_patterns = 5;
+
+  // Regex which should be contained in the file.
```
BenHenning:
```suggestion
  // Regex which must be contained in the file.
```

scripts/src/java/org/oppia/android/scripts/regex/RegexPatternValidationCheck.kt
```diff
@@ -226,9 +246,17 @@ private data class MatchableFileContentCheck(
...
+      return lines.withIndex().filter { (_, line) ->
+        prohibitedContentRegex.containsMatchIn(line)
+      }.map { (index, _) -> index }
+    else return emptyList()
+  }
+
+  /** Returns a boolean indicating whether the file contains the required content or not. */
+  fun doesFileContainRequiredRegex(fileContent: String): Boolean {
+    if (requiredContentRegex != null) return requiredContentRegex.containsMatchIn(fileContent)
+    else return true
```
BenHenning:
```suggestion
    requiredContentRegex?.containsMatchIn(fileContent) ?: true
```

More concise & idiomatic.

scripts/src/java/org/oppia/android/scripts/regex/RegexPatternValidationCheck.kt
```diff
@@ -239,9 +267,15 @@ private data class MatchableFileContentCheck(
   companion object {
     /** Returns a new [MatchableFileContentCheck] based on the specified [FileContentCheck]. */
     fun createFrom(fileContentCheck: FileContentCheck): MatchableFileContentCheck {
+      val prohibitedContentRegex = if (fileContentCheck.prohibitedContentRegex == "") null
+      else fileContentCheck.prohibitedContentRegex.toRegex()
```
BenHenning:
```suggestion
      val prohibitedContentRegex =
        fileContentCheck.prohibitedContentRegex.takeIf(String::isNotEmpty)?.toRegex()
```

Bit more idiomatic and possibly easier to read. Ditto for the other one.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -157,6 +157,8 @@ class RegexPatternValidationCheckTest {
   private val wikiReferenceNote =
     "Refer to https://github.com/oppia/oppia-android/wiki/Static-Analysis-Checks" +
       "#regexpatternvalidation-check for more details on how to fix this."
+  private val doesNotContainScreenName = "Screen Name not added."
```
BenHenning: Nit: move this & the other above the note, and rename to match the other error message field names.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -207,7 +211,8 @@ class RegexPatternValidationCheckTest {
       """
       File name/path violation: $activitiesPlacementErrorMessage
       - data/src/main/TestActivity.kt
-
+      
+      data/src/main/TestActivity.kt: $doesNotContainScreenName
```
BenHenning: For this & other tests, can we update these to avoid including these extra errors? They seem to be "bleeding into" some of these other tests, and they shouldn't be since tests should be verifying isolated behaviors.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
```diff
@@ -139,6 +145,12 @@ class EventBundleCreatorTest {
   @Inject lateinit var context: Context
   @Inject lateinit var eventBundleCreator: EventBundleCreator
 
+  @Parameter
+  lateinit var screenNameCase: String
+
+  @Parameter
+  lateinit var screenNameCaseValue: String
```
BenHenning:
```suggestion
  @Parameter lateinit var name: String
  @Parameter lateinit var expNameStr: String

  private val screenName by lazy { ScreenName.valueOf(name) }
```

Kotlin actually has a built-in way to convert strings to enums, so long as the name constant perfectly matches. Also, the shorter parameter names can help cut down on how long the iterations are.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
```diff
@@ -1460,6 +1472,262 @@ class EventBundleCreatorTest {
     assertThat(bundle).string("install_id").isEqualTo(TEST_INSTALLATION_ID)
   }
 
+  @Test
+  @RunParameterized(
+    Iteration("forHomeScreen", "screenNameCase=HOME_ACTIVITY", "screenNameCaseValue=home_activity"),
```
BenHenning:
```suggestion
    Iteration("home", "name=HOME_ACTIVITY", "expStrName=home_activity"),
```

Per my suggestions above. Here, we can fit it easily. Hopefully most can fit on one line now with this approach which should significantly reduce how much space needs to be dedicated for the test.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+  fun Intent.decorateWithScreenName(screenName: ScreenName) {
+    putProtoExtra(
+      CURRENT_APP_SCREEN_NAME_KEY,
+      CurrentAppScreen.newBuilder().setScreenName(screenName).build()
+    )
+  }
+
+  /** Returns the [ScreenName] packed in the intent via [decorateWithScreenName]. */
+  fun Intent.extractCurrentAppScreenName(): ScreenName {
+    return this.getProtoExtra(
```
BenHenning: Ditto here ('this' not needed).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ActivityLifecycleObserverTest.kt
BenHenning: Sure--that seems sensible. Thanks!

BenHenning (changes requested):
Thanks @Sarthak2601! This is looking really good. Just had a few comments mainly on new code--PTAL.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
Sarthak2601: done.

utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt
Sarthak2601: done.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
Sarthak2601: done.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
Sarthak2601: done.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
Sarthak2601: done.

scripts/src/java/org/oppia/android/scripts/regex/RegexPatternValidationCheck.kt
Sarthak2601: done.

scripts/src/java/org/oppia/android/scripts/regex/RegexPatternValidationCheck.kt
Sarthak2601: done.

scripts/src/java/org/oppia/android/scripts/proto/file_content_validation_checks.proto
Sarthak2601: done.

scripts/assets/file_content_validation_checks.textproto
Sarthak2601: done.

scripts/assets/file_content_validation_checks.textproto
Sarthak2601: I'm not sure if we can test for the exact ScreenName value (since all enum values are different and imported directly) but I've modified the required_content_regex to match for `decorateWithScreenName(`. This ensures that the screen name is added since the app won't build without a screen name being checked into this function. Does that work?

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -207,7 +219,7 @@ class RegexPatternValidationCheckTest {
       """
       File name/path violation: $activitiesPlacementErrorMessage
       - data/src/main/TestActivity.kt
-
+
```
BenHenning: Nit: remove extra spacing.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -1710,10 +1724,11 @@ class RegexPatternValidationCheckTest {
 
   @Test
   fun testFilenameAndContent_useProhibitedFileName_useProhibitedFileContent_multipleFailures() {
+    val acceptedContent = "decorateWithScreenName(TEST_ACTIVITY)"
```
BenHenning: Here & elsewhere: maybe 'requiredContent'?

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -2159,6 +2174,70 @@ class RegexPatternValidationCheckTest {
...
+      .isEqualTo(
+        """
+        $stringFilePath: $screenNameNotPresentErrorMessage
+        $wikiReferenceNote
+        """.trimIndent()
+      )
+  }
+
+  @Test
+  fun testScreenNamePresence_activityFileWithAcceptedContent_screenNameIsPresent() {
```
BenHenning: Ditto here: perhaps 'required'? Or maybe better: specify what that actual content is (like you do for the activitytest test below).

scripts/assets/file_content_validation_checks.textproto
BenHenning: Yeah, this seems more robust than the previous pattern. Thanks!

BenHenning (approved):
Thanks @Sarthak2601! Only have nits left, otherwise the PR LGTM. Feel free to self-resolve the remaining comments once addressed.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
Sarthak2601: done.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
Sarthak2601: done.

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
Sarthak2601: done.

PR #4468, opened 2022-07-26
Fixes #2627: Improve accessibility for Pin Verification Screen

### LINKED ISSUE #2627: [A11y] Pin Verification Screen
## Current Output

https://user-images.githubusercontent.com/9396084/107003150-543c2700-67b2-11eb-97f0-636b2bd4db10.mp4

## Issues Identified
- [ ] If there is incorrect PIN the talkback does not shift to error and does not say that it was incorrect therefore the learner won't know what went wrong. Ideally as soon as error is show it should get highlighted. (For reference: In this [video](https://github.com/oppia/oppia-android/issues/2616) as soon as we enter incorrect answer the talkback automatically focuses on error.)
- [ ] If PIN area is selected, on swipe to next talkback shifts to **FORGOT MY PIN** but instead it should shift to **Show/Hide** Icon.
- [ ] Output for **Show/Hide View** is following: 
(a) Password [shown/hidden] icon
(b.) [SHOW/HIDE]
(c.) Double-tap to activate

Note: To understand the above issue(s) completely, it is recommended that you setup Talkback and play with the app keeping it on and that will give you better context.

**Accessibility Guide**: https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide

vrajdesai78:
## Explanation
Fixes #2627: Improve accessibility for Pin Verification screen. I have removed seperate textview which is used to show error text. Now, error is set directly to PinPassPasswordInputEditText using onTextChanged extension function. 

## Essential Checklist

### Before

https://user-images.githubusercontent.com/9396084/107003150-543c2700-67b2-11eb-97f0-636b2bd4db10.mp4

### After

https://user-images.githubusercontent.com/43074241/180941747-b5dad9b0-2ea4-45c2-9003-a4dc1ac56ff9.mp4

### Passing updated testcases
Screenshot from 2022-07-29 23-49-25
Screenshot from 2022-07-29 23-52-30
Screenshot from 2022-07-29 23-55-13


rt4914 (changes requested):
From Code perspective and implementation its really good.

Not sure why tests are failing, can you run End-to-End and other robolectric tests and check.

Also, maybe just merge with latest develop.

vrajdesai78:
Updated video based on new changes

rt4914 (changes requested):
Please remove all `DisableAccessibilityChecks ` and corresponding TODO's for #3245 

As this PR is fixing all of them too.

vrajdesai78:
All robolectric tests for PinPasswordActivityTest are passing locally even without disabling Accessibility checks. 
Screenshot from 2022-08-04 00-33-55

vrajdesai78:
Unable to run robolectric tests locally. I have run espresso tests and they are passing as expected. I have tried to run all the robolectric tests of PinPasswordActivityTest but it's just loading. I have kept tests running for more than 30 mins but still not even one test is getting passed or failed.

Screenshot from 2022-08-05 14-57-09

BenHenning:
@vrajdesai78 are you able to run any app module Robolectric tests? If so, does the PinPasswordActivityTest pass for you on Robolectric for the develop branch (i.e. without your changes in this PR)? These details may help provide some clarity into what's happening.

vrajdesai78:
I have tried with develop branch. All test cases are passing in develop branch.

rt4914 (changes requested):
I have been trying to check robolectric tests for almost a dozen hours now and these are my findings.

1. `testPinPassword_withUser_forgot_inputAdminPinAndNewPinAndOldPin_wrongPinError` this test is incorrect as `pin_password_incorrect_pin` is checked incorrectly.
2. The following four tests are unable to finish on Robolectric:
Screenshot 2022-08-11 at 2 05 48 PM


3. In all these 4 tests if we replace `hasErrorText(context.resources.getString(R.string.pin_password_incorrect_pin))` with `hasNoErrorText` all tests pass which means that there is definitely some issue as the tests are not visible.

4. Also, where-ever we are trying to append the text to the input-pin TextLayoutInput, we should use

```
 onView(
        allOf(
          withId(R.id.pin_password_input_pin_edit_text),
          isDescendantOfA(withId(R.id.pin_password_input_pin))
        )
      )
```
and not 
`onView(withId(R.id.pin_password_input_pin_edit_text))`


Try to work around these cases.

Maybe directly add `@Ignore` to above mentioned 4 tests and run all robolectirc tests without any change and notice they all pass. And now start debugging the remaining 4 tests one by one.

rt4914:
I have made few changes to test cases (which are incorrect changes too) just to show that all tests are either passing or failing.
The biggest culprit is `testCoroutineDispatchers.runCurrent()` at some places its blocking the test completely.

## Output
Screenshot 2022-08-11 at 2 45 29 PM

## Corresponding Code (definitely contains some incorrect code)

https://gist.github.com/rt4914/2982b2a4c0bd13b213959902851040b0

rt4914:
@BenHenning Can you have a look at my last two comments here. Tests are not finishing in certain cases.

BenHenning:
@rt4914 how consistently does ``runCurrent`` block tests? Have you been able to look at all into what it's doing when blocking?

vrajdesai78:
> @rt4914 how consistently does `runCurrent` block tests? Have you been able to look at all into what it's doing when blocking?

@BenHenning I have tried with multiple scenarios and found that closeSoftKeyboard() function is the culprit which blocking our tests from running. Now, the test are running (without any blockage) but they are failing as in robolectric it is not able to check error text using hasErrorText(). We have used `hasErrorText()` in other tests but in that we are checking it once submit button is clicked but here we don't have any submit button that's why test is failing.

BenHenning:
> > @rt4914 how consistently does `runCurrent` block tests? Have you been able to look at all into what it's doing when blocking?
> 
> @BenHenning I have tried with multiple scenarios and found that closeSoftKeyboard() function is the culprit which blocking our tests from running. Now, the test are running (without any blockage) but they are failing as in robolectric it is not able to check error text using hasErrorText(). We have used `hasErrorText()` in other tests but in that we are checking it once submit button is clicked but here we don't have any submit button that's why test is failing.

@vrajdesai78 it doesn't make sense to me that a submit button would affect anything except maybe somehow causing closeSoftKeyboard to work, or triggering a later task that can be synchronized with TestCoroutineDispatchers. What I think you need to check next is what is actually happening when the error text should be set relative to each step in the test (which will tell if it's being set correctly before it's asserted in the scenario where it fails). If it isn't being set in time, you can look into why (if it's an async thing, you should be able to use TestCoroutineDispatchers). If it is being set in time, you'll need to investigate within Espresso and Robolectric why the assertion is failing even though the state is correct.

vrajdesai78:
Keep this PR Open

vrajdesai78:
I have created a debugging doc for current failing robolectric test cases. You can find it [here](https://docs.google.com/document/d/13Cu3DqJ-7qP7SQsdHCs1nVWWZBP2l10_0_urW4kczGY/edit?usp=sharing) @BenHenning @rt4914 can you help me out with this

app/src/main/res/layout/pin_password_activity.xml
```diff
@@ -106,9 +106,10 @@
               android:id="@+id/pin_password_input_pin"
               style="@style/TextInputLayout"
               android:layout_width="0dp"
-              android:layout_height="56dp"
+              android:layout_height="wrap_content"
               android:layout_marginTop="2dp"
               android:hint="@{viewModel.profile.isAdmin ? @string/input_pin_password_as_admin : @string/input_pin_password_as_user}"
+              app:errorMessage="@{viewModel.errorMessage}"
```
BenHenning, line 112: Ah, does TextInputLayout handle the accessibility case for the error field automatically?

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -943,9 +943,9 @@ class PinPasswordActivityTest {
         closeSoftKeyboard()
       )
       onView(isRoot()).perform(orientationLandscape())
-      onView(withText(context.getString(R.string.pin_password_incorrect_pin))).check(
+      onView(withId(R.id.pin_password_input_pin)).check(
         matches(
-          isDisplayed()
+          hasErrorText(context.resources.getString(R.string.pin_password_incorrect_pin))
```
BenHenning, line 966: I think it may also be sensible to test the case(s) for when this error is empty, too.

BenHenning (changes requested):
Thanks @vrajdesai78. I think technically, this PR looks quite nice. Just had a couple of thoughts.

Re: the test issue, I don't have comment access on your doc so posting my suggestion here. Have you tried adding ``testCoroutineDispatchers.runCurrent()`` before the assertion to see if it's a timing issue?

app/src/main/res/layout/pin_password_activity.xml
vrajdesai78: Yes, I think so because in other cases also we are setting error message to TextInputLayout so I think that it will handle accessibility too

vrajdesai78:
> Thanks @vrajdesai78. I think technically, this PR looks quite nice. Just had a couple of thoughts.
> 
> Re: the test issue, I don't have comment access on your doc so posting my suggestion here. Have you tried adding ``testCoroutineDispatchers.runCurrent()`` before the assertion to see if it's a timing issue?

Yes, I have tested with testCouroutineDispatchers.runCurrent() but still it's giving me the same error, infact if I use it before closeSoftKeyboard() it is stopping test from executing further. So, here testCouroutineDispatchers.runCurrent() won't help. I have also tried to add submit button and tried to set error message onclick of submit button but still test is failing. I think there might be some issue with the way I am setting error message.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -224,9 +224,9 @@ class PinPasswordActivityTest {
       closeSoftKeyboard()
       onView(withId(R.id.pin_password_input_pin_edit_text)).perform(closeSoftKeyboard())
         .perform(editTextInputAction.appendText("54321"), closeSoftKeyboard())
-      onView(withText(context.getString(R.string.pin_password_incorrect_pin))).check(
```
BenHenning, line 245: Out of curiosity, does this old check work if put back? We should be using the text input layout version, but I'm curious.

Also, have you checked via step debugging whether the error actually is set by the time it's checking for it to be present?

Finally, have you verified the behavior on both Robolectric and Espresso? Is it failing in the same way on both?

BenHenning:
I left a follow up comment with a few suggestions to aid in investigating @vrajdesai78. I've checked your code again and looked at TextInputLayout and nothing is jumping out to me yet, so I think we need to collect more data here.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
vrajdesai78: > Out of curiosity, does this old check work if put back? We should be using the text input layout version, but I'm curious.
> 
> Also, have you checked via step debugging whether the error actually is set by the time it's checking for it to be present?
> 
> Finally, have you verified the behavior on both Robolectric and Espresso? Is it failing in the same way on both?

Tests are passing in espresso and failing in robolectric. I have tried to check whether TextInputLayout has error text or not by asserting the TextInputLayout and in that TextInputLayout.error is returning null.

vrajdesai78:
```  
@Test 
  fun testPinPassword_withAdmin_inputWrongPin_incorrectPinShows() {
    ActivityScenario.launch<PinPasswordActivity>(
      PinPasswordActivity.createPinPasswordActivityIntent(
        context = context,
        adminPin = adminPin,
        profileId = adminId
      )
    ).use {
      closeSoftKeyboard()
      onView(withId(R.id.pin_password_input_pin_edit_text)).perform(closeSoftKeyboard())
        .perform(editTextInputAction.appendText("54321"), closeSoftKeyboard())
      onView(withText(R.string.pin_password_incorrect_pin)).check(
        matches(isDisplayed())
      )
    }
  } 
```

I have tried with above test and it's passing in espresso and failing in robolectric. Current version of test i.e. checking hasErrorText is also working on espresso but failing in robolectric.

BenHenning:
@vrajdesai78 what do you mean by the current version of the test is failing on Robolectric? That doesn't seem to be the case given that our CI checks are passing. Am I misunderstanding? You'll also need to step debug through the assertion pipeline to figure out why it isn't matching. What's likely happening is the view doesn't have the expected state when we think it does. If that's the case, you'll need to step debug to see why it isn't being set on time, and when it actually is being set.

vrajdesai78:
> @vrajdesai78 what do you mean by the current version of the test is failing on Robolectric? That doesn't seem to be the case given that our CI checks are passing. Am I misunderstanding? You'll also need to step debug through the assertion pipeline to figure out why it isn't matching. What's likely happening is the view doesn't have the expected state when we think it does. If that's the case, you'll need to step debug to see why it isn't being set on time, and when it actually is being set.

@BenHenning I have tried to use `testCoroutineDispatchers.advanceUntilIdle()` when we enter correct pin and it opens HomeActivity, and suprisingly I found `testPinPassword_withUser_inputCorrectPin_opensHomeActivity` test is running infinite if I use `testCoroutineDispatchers.advanceUntilIdle()`. So, I am thinking that may be advaneUntilIdle is not working properly in `PinPasswordActivityTest` and we can't conclude anything from that. I have also tried to use `testCoroutineDispatchers.advanceTimeBy(50000)` but still the test is failing. 

So, now what would you suggest, how can I debug this more.

BenHenning:
As mentioned in chat, I think you'll need to investigate *why* advanceUntilIdle doesn't end as that will provide context on the original failure @vrajdesai78.

vrajdesai78:
image
As @BenHenning suggested, I have tried to debug ErrorTextMatcher and found that `TextInputLayout.error` is null as shown above. I have tried to check it with different tests, and for other tests `TextInputLayout.error` is returning error properly. 

@BenHenning can you suggest me next steps

BenHenning:
> image As @BenHenning suggested, I have tried to debug ErrorTextMatcher and found that `TextInputLayout.error` is null as shown above. I have tried to check it with different tests, and for other tests `TextInputLayout.error` is returning error properly.
> 
> @BenHenning can you suggest me next steps

@vrajdesai78 just to check, what is the value of just the error (without the 'toString' bit)? I'm curious why it's resolving to ``"null"`` rather than ``null`` (i.e. that it's a string).

This does provide us with context, though: the error isn't being defined by the time we try to check for its presence. You'll need to step through the production code & test at the same time to figure out how far into the data pipeline the test gets before it verifies the error (evidently, it's not getting all the way to setting the error).

Apologies for being so delayed on this. I will keep this assigned to me and try to investigate it later tonight my time.

vrajdesai78:
@BenHenning updated [debugging doc](https://docs.google.com/document/d/13Cu3DqJ-7qP7SQsdHCs1nVWWZBP2l10_0_urW4kczGY/edit?usp=sharing). I found that pin is not entering in `pin_password_input_pin_edit_text` which is the case of failing test case.

rt4914 (approved):
In this I don't have anything to block on except for test cases. But relying on @BenHenning for that part of PR.

@vrajdesai78 Approving this for now to unblock you but still assign it back to me once approved by @BenHenning 
(You can merge it, I will still do the post review).

vrajdesai78:
> @BenHenning updated [debugging doc](https://docs.google.com/document/d/13Cu3DqJ-7qP7SQsdHCs1nVWWZBP2l10_0_urW4kczGY/edit?usp=sharing). I found that pin is not entering in `pin_password_input_pin_edit_text` which is the case of failing test case.

Added new findings in [this](https://docs.google.com/document/d/13Cu3DqJ-7qP7SQsdHCs1nVWWZBP2l10_0_urW4kczGY/edit?usp=sharing) debugging doc.

BenHenning:
Leaving an update to capture some of the findings from my investigation:
- I'm finding that both runCurrent and advanceUntilIdle stall in a livelock, and I've isolated this to Robolectric's ShadowLooper.idle().
- idle() hangs after the TextInputLayout's error text is caused.
- Upon analyzing the error text, it seems to be continuously recomputing the error text icon (even though there isn't one), and is causing the error text layout to continually be remeasured.
- ShaderLooper.idle() is known to have issues with animations specifically.
- I tried to disable animations programmatically, but this is a bit tricky to do in Robolectric and didn't seem to make a difference.
- The solution listed in https://github.com/robolectric/robolectric/issues/6158 was actually reasonable (or, at least, it let us get past the problematic code), but it causes Robolectric's ActivityController to have an issue, instead (which also happens with a vanilla advanceUntilIdle() call without the delay, and without the breaking error text above).

It seems very likely this is a problem with both Robolectric and TextInputLayout (the former doesn't handle animations well, and the latter seems to be doing potentially problematic things). I still have no idea how to work around this problem as it seems like it may not actually be an issue with animations. I think one part that especially confuses me is why real-time & delayed callback time causes it to advance past the measure, but zero-time delays does not.

BenHenning:
I just found a really important detail: when investigating Robolectric's LooperDelegatingScheduler I found that advanceTo() is being called with increasing long time intervals. The scheduler actually is correctly running, it just reaches a point where the delay in running takes so long that it looks like a process stall. Not yet sure *why* this is happening, though.

Edit: Ah, this is because advanceUntilIdle() attempts to execute all tasks. However, if there is no limit then we'll execute ad infinitum (though TestCoroutineDispatchers' timeout should kick in at some point). This might help me root cause what's actually scheduling those tasks.

Edit 2: Welp, looks like it might be the EditText blink animation that stalls advanceUntilIdle(). I doubt this is the issue causing runCurrent() to stall when error text is set, but I want to at least find a way to disable to blink animation.

Edit 3: Confirming that disabling the blinking cursor allows advanceUntilidle() to not stall, but it's unrelated (at least partly) to runCurrent() stalling when enabling error text.

Edit 4: It seems that the setPostFrameCallbackDelay() + advanceUntilIdle() approach for 'fixing' the error text isn't actually doing what I thought it would. It somehow causes the Robolectric main looper message queue to break & stop executing things which is why Espresso ends up hanging during view assertion time.

Edit 5: Okay, so it seems like the post frame callback delay doesn't actually do anything to fix the issue (it just caused the message queue to enter a strangely invalid blocking state that, when unblocked, triggers the livelock).

BenHenning:
So I got the test passing by moving the TextInputLayout outside ConstraintLayout. It seems that ConstraintLayout & TextInputLayout are causing a bug in one of them (can't be sure which). This happens by ConstraintLayout re-measuring TextInputLayout several times in such a way where TextInputLayout requests a re-layout. The combination of these behaviors seems to cause a constant back-and-forth measure process that never ends in Robolectric (when observing TextInputLayout's measures, it switches between 3-4 different widths, one of which is 0 and causes the state to revert back).

I'm not yet sure what in ConstraintLayout is actually causing this, or what we can do about it.

Edit: Welp, I found a fix. I noticed that the TextInputLayout is being set with a 0dp width which I think is triggering the bad case that breaks it in Robolectric. I suspect this is just an invalid state to put the layout in, and is just causing an inefficiency in regular runtime that's not noticed (though it's possible this is still only observable in Robolectric). I'm going to investigate to see if it's okay not to use 0dp here, and possibly prohibiting it to avoid others hitting this rather difficult-to-investigate problem.

Edit 2: New theory: add_profile_activity also has a 0dp-width TextInputLayout inside a ConstraintLayout, and it also verifies TextInputLayout error text successfully. However, a major difference is that in this case we're actually nesting ConstraintLayouts with both the TextInputLayout having a 0dp (match_constraint) width, as well as its parent ConstraintLayout.

Edit 3: Confirming that changing the parent ConstraintLayout's width to not be 0dp also fixes the issue. Gotta think about how to check for this via tests/checks.

BenHenning:
Not stale, I'm still working on this.

BenHenning:
I think you can resolve this now @vrajdesai78 by changing pin_password_activity such that it doesn't have nested ConstraintLayouts. That will avoid the problem.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
vrajdesai78: Added a testcase to check that error is empty on correct pin.

vrajdesai78:
@BenHenning I have fixed the failing test cases as per you suggestion by setting width to match_parent instead of 0dp. Also, added a new test to check that error is not displayed when correct pin is entered. Thanks

app/src/main/res/layout/pin_password_activity.xml
```diff
@@ -95,7 +95,7 @@
             app:layout_constraintTop_toBottomOf="@id/hello_text_view" />
 
           <androidx.constraintlayout.widget.ConstraintLayout
-            android:layout_width="@dimen/pin_password_activity_input_with_icon_width"
+            android:layout_width="match_parent"
```
BenHenning: Is this change actually correct to make? A solution that doesn't require changing dimensions would be collapsing the nested ConstraintLayouts (which seems like the more serious issue here as it is both a performance issue and complex bugs like the one this PR exposed).

BenHenning (commented):
Thanks @vrajdesai78. Just had one follow-up comment about the change to pin_password_activity--PTAL.

app/src/main/res/layout/pin_password_activity.xml
vrajdesai78: I have reverted back to dimen and inside dimens.xml I have used -1px which is equivalent to match parent. I have referred this [stackoverflow](https://stackoverflow.com/a/37971851/12314934) answer. I have checked UI and I haven't found any issue with the current changes.

app/src/main/res/values-sw600dp-land/dimens.xml
```diff
@@ -340,7 +340,7 @@
   <dimen name="pin_password_activity_padding_top">48dp</dimen>
   <dimen name="pin_password_activity_hello_text_view_margin_start">0dp</dimen>
   <dimen name="pin_password_activity_hello_text_view_margin_end">0dp</dimen>
-  <dimen name="pin_password_activity_input_with_icon_width">0dp</dimen>
+  <dimen name="pin_password_activity_input_with_icon_width">-1px</dimen>
```
BenHenning, line 344: For this & the other files, please include a comment explaining what this means (since it's otherwise a bit of a code smell since negative width doesn't make sense).

BenHenning (commented):
Thanks @vrajdesai78. Just had one follow-up nit otherwise the change LGTM.

app/src/main/res/values-sw600dp-land/dimens.xml
vrajdesai78: Added a comment, thanks

BenHenning (approved):
Thanks @vrajdesai78! LGTM.

Issue #264, opened 2019-10-25
Move custom html parsing & image handling logic to custom view with attributes and a custom binding adapter


BenHenning:
Right now we need to manually set up a TextView for custom Oppia HTML/image support via an HTML processor utility. However, this could happen automatically with custom attributes & binding adapters. All that's needed is the entity ID (e.g. exploration ID), entity type (e.g. 'exploration'), and the HTML to handle. We would need a way to tie the view to Dagger to inject the HtmlParser, however that probably can be done by introducing a ViewScope/ViewComponent.

This would make working with Oppia-customized HTML easier across the app.

EskAlam06:
it's a good enhancement issue , should remain open

PR #4489, opened 2022-08-03
Fix time ago plurals

### LINKED ISSUE #3841: 'time_ago' & quantity plurals should be combined
After #3795, TextViewBindingAdapters combines two strings (time_ago & plurals) into a single one. This ought to just be different plural options so that translators have better context when translating the full conceptual string.

mbobiosio:
## Explanation

Fixes #3841 

This pull request provides an all kotlin approach to handling timeAgo UI updates.
I have created an extension gets a long timestamp and returns a string e.g just now, a minute ago, an hour ago, 3 days go

Usage: I have also updated BindingAdapter for TextView. This textview simply parses the string result of the extension above

## Essential Checklist

## For UI-specific PRs only

### DIFF
--- a/app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
+++ b/app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
@@ -27,6 +27,7 @@ fun setProfileDataText(textView: TextView, timestamp: Long) {
   )
 }
 
+// TODO: Move this to a common place.
 fun getResourceHandler(view: View): AppLanguageResourceHandler {
   val provider =
     getAttachedActivity(view) as AppLanguageActivityInjectorProvider
--- a/app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
+++ b/app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
@@ -1,0 +1,49 @@
+package org.oppia.android.app.databinding
+
+import android.app.Activity
+import android.content.ContextWrapper
+import android.view.View
+import android.widget.TextView
+import androidx.databinding.BindingAdapter
+import org.oppia.android.R
+import org.oppia.android.app.translation.AppLanguageActivityInjectorProvider
+import org.oppia.android.app.translation.AppLanguageResourceHandler
+import org.oppia.android.app.utility.datetime.timeAgo
+
+@BindingAdapter("profile:lastVisited")
+fun TextView.setProfileLastVisitedText(timestamp: Long?) {
+  timestamp?.let {
+    text = it.timeAgo(this)
+  }
+}
+
+@BindingAdapter("profile:created")
+fun setProfileDataText(textView: TextView, timestamp: Long) {
+  val resourceHandler = getResourceHandler(textView)
+  val time = resourceHandler.computeDateString(timestamp)
+  textView.text = resourceHandler.getStringInLocaleWithWrapping(
+    R.string.profile_edit_created,
+    time
+  )
+}
+
+fun getResourceHandler(view: View): AppLanguageResourceHandler {
+  val provider =
+    getAttachedActivity(view) as AppLanguageActivityInjectorProvider
+  return provider.getAppLanguageActivityInjector().getAppLanguageResourceHandler()
+}
+
+private fun getAttachedActivity(view: View): Activity {
+  var context = view.context
+  while (context != null && context !is Activity) {
+    check(context is ContextWrapper) {
+      (
+        "Encountered context in view (" + view + ") that doesn't wrap a parent context: " +
+          context
+        )
+    }
+    context = context.baseContext
+  }
+  checkNotNull(context) { "Failed to find base Activity for view: $view" }
+  return context as Activity
+}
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapters.java
```diff
@@ -56,7 +65,9 @@ public static void setDrawableTopCompat(
     );
   }
 
-  /** Binds an AndroidX KitKat-compatible drawable end to the specified text view. */
+  /**
+   * Binds an AndroidX KitKat-compatible drawable end to the specified text view.
+   */
```
BenHenning, line 70: The previous formatting was actually correct (see https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line). Ditto elsewhere in the PR.

app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+import android.view.View
+import android.widget.TextView
+import androidx.databinding.BindingAdapter
+import org.oppia.android.R
+import org.oppia.android.app.translation.AppLanguageActivityInjectorProvider
+import org.oppia.android.app.translation.AppLanguageResourceHandler
+import org.oppia.android.app.utility.datetime.timeAgo
+
+@BindingAdapter("profile:lastVisited")
+fun TextView.setProfileLastVisitedText(timestamp: Long?) {
```
BenHenning, line 14: How many views need to bind this? If it's not many, I wonder if it would be better to create a custom view, instead (which may be a bit cleaner since then you don't need to create a custom binding adapter).

This also might be a good way to establish the start to a new pattern where we have a centralize TextView that handles common Oppia-specific operations (such as HTML binding--see #264).

app/src/main/java/org/oppia/android/app/settings/profile/ProfileEditFragmentPresenter.kt
```diff
@@ -132,22 +135,25 @@ class ProfileEditFragmentPresenter @Inject constructor(
...
-              fragment.startActivity(intent)
-            } else {
-              val intent = Intent(fragment.requireContext(), ProfileListActivity::class.java)
-              intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
-              fragment.startActivity(intent)
-            }
+        fragment
+      ) {
+        if (it is AsyncResult.Success) {
+          fragment.requireContext().toast(
```
BenHenning, line 141: We don't actually use toasts since they have a few issues that are hard to work around. Specifically: they can't be restyled easily (since they're a system UI), and they won't show if the user sets certain notification settings (such as do-not-disturb, if I recall correctly). Alternatively, we can show a snackbar (which we don't use yet, but it's something we could discuss with our design team), or a dialog (which is mainly what we use now).

app/src/main/java/org/oppia/android/app/utility/datetime/DateTimeUtil.kt
```diff
@@ -31,3 +34,54 @@ class DateTimeUtil @Inject constructor(
...
+const val DAY = 24 * HOUR
+const val MONTH = 30 * DAY
+const val YEAR = 12 * MONTH
+
+private fun currentDate() = Calendar.getInstance().timeInMillis
+
+/**
+* Returns the readable string of the duration from the provided time in [Long].
+*/
+fun Long.timeAgo(view: View): String {
```
BenHenning, line 50: Rather than this being a generic Long extension, perhaps it could be part of DateTimeUtil itself? Then we wouldn't need to pass in a view.

app/src/main/java/org/oppia/android/app/utility/datetime/DateTimeUtil.kt
```diff
@@ -31,3 +34,54 @@ class DateTimeUtil @Inject constructor(
...
+  val time = this
+  val now = currentDate()
+  val diff = (now - time) / 1000
+
+  val resourceHandler = getResourceHandler(view)
+
+  return when {
+    diff < MINUTE -> resourceHandler.getStringInLocale(R.string.just_now)
```
BenHenning, line 58: I think fixing this string actually involves introducing an Android quantity string with formatting that combines all of the different cases into one. This seems like the ideal use case for quantity strings as it will provide a full context to translators such that they can specialize the wording to better fit different languages.

BenHenning (commented):
Hi @mbobiosio. Took a quick pass on this PR, as well, and had some high-level design feedback--PTAL. Also, is this PR being conflated with your other? It seems like there's a lot of code shared between the two.

app/src/main/java/org/oppia/android/app/databinding/TextViewBindingAdapter.kt
mbobiosio: There's three layouts that rely on this. I thought of a custom view but it sounded too expensive for the operation.

app/src/main/java/org/oppia/android/app/settings/profile/ProfileEditFragmentPresenter.kt
mbobiosio: Oh no. This wasn't really supposed to be here. I'll send you a chat if that's okay

PR #4498, opened 2022-08-08
 Fix #4451: Add domain level functionality for spotlights 

### LINKED ISSUE #4451: Add domain level functionality for spotlights
As a part of the GSoC project: Interactive Onboarding Flow which focusses on making the initial user experience more engaging and smooth, the plan is to introduce spotlights into the app. 

As an initial part of this plan, implement the domain level functionality for spotlights. This should cover all model classes and spotlight controllers necessary.

JishnuGoyal:
## Explanation
Fixes #4451

This PR lays the domain-level foundation for "Spotlights". These are a part of the Interactive On-boarding Flow (GSoC) project which focuses on making the initial user experience easier and interactive, and aims to make the user start learning effectively as soon as possible.

`spotlight.proto` is a model class that holds the information about which spotlights have/have not been seen yet by the user.

`SpotlightStateController` is an injectable class which controls the domain level functionality needed to make changes to the above model class. Essentially, this class can mark spotlights "seen" and retrieve the spotlight states (where states determine whether they are seen/unseen).

The entire Spotlight experience can be turned on or off just by setting the platform parameter flag called `EnableSpotlightUi` to true or false.

This PR is backed by necessary tests to check if the domain layer for spotlights works correctly or not.

## Essential Checklist


JishnuGoyal:
Please take an initial pass at this 
PTAL @BenHenning

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+          }
+          Spotlight.FeatureCase.VOICEOVER_PLAY_ICON -> {
+            it.voiceoverPlayIcon
+          }
+          Spotlight.FeatureCase.VOICEOVER_LANGUAGE_ICON -> {
+            it.voiceoverLanguageIcon
+          }
+          Spotlight.FeatureCase.FEATURE_NOT_SET -> {
+            return@transformAsync AsyncResult.Failure(
+              SpotlightFeatureNotFoundException("Spotlight feature was not found")
```
BenHenning: I don't mind the custom exception, but perhaps we could just use an IllegalArgumentException or IllegalStateException here?

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+        AsyncResult.Success(viewState)
+      }
+  }
+
+  private fun recordSpotlightStateAsync(
+    profileId: ProfileId,
+    feature: Spotlight.FeatureCase,
+    viewState: SpotlightViewState
+  ): Deferred<Any> {
+    return retrieveCacheStore(profileId).storeDataWithCustomChannelAsync(
```
BenHenning: Why is a custom channel needed here?

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+    } else {
+      val cacheStore =
+        cacheStoreFactory.createPerProfile(
+          CACHE_NAME,
+          SpotlightStateDatabase.getDefaultInstance(),
+          profileId
+        )
+      cacheStoreMap[profileId] = cacheStore
+      cacheStore
+    }
```
BenHenning: I suspect ``getOrPut`` would be a more idiomatic way to do this. Consider:

```suggestion
    val cacheStore = cacheStoreMap.getOrPut(profileId) {
        cacheStoreFactory.createPerProfile(
          CACHE_NAME,
          SpotlightStateDatabase.getDefaultInstance(),
          profileId
        )
    }
```

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+   */
+  fun retrieveSpotlightViewState(
+    profileId: ProfileId,
+    feature: Spotlight.FeatureCase,
+  ): DataProvider<SpotlightViewState> {
+    return retrieveCacheStore(profileId)
+      .transformAsync(
+        RETRIEVE_SPOTLIGHT_CHECKPOINT_DATA_PROVIDER_ID
+      ) {
+
```
BenHenning: Nit: remove blank newline for consistency.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+        val viewState = when (feature) {
+          Spotlight.FeatureCase.ONBOARDING_NEXT_BUTTON -> {
+            it.onboardingNextButton
+          }
```
BenHenning: I suggest directly importing the feature case so that this can neatly fit on one line. E.g.:

```suggestion
          ONBOARDING_NEXT_BUTTON -> it.onboardingNextButton
```

Ditto for the other cases, and for the setter cases below.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+  ): Deferred<Any> {
+    return retrieveCacheStore(profileId).storeDataWithCustomChannelAsync(
+      updateInMemoryCache = true
+    ) {
+      val spotlightStateDatabaseBuilder = it.toBuilder()
+
+      val spotlight = when (feature) {
+        Spotlight.FeatureCase.ONBOARDING_NEXT_BUTTON -> {
+          spotlightStateDatabaseBuilder.setOnboardingNextButton(viewState)
+        }
```
BenHenning: Per my comment above, if a custom channel isn't needed then you can probably simplify this by doing something like:

```kotlin
spotlightState.toBuilder().run { builder ->
  when (feature) {
    ONBOARDING_NEXT_BUTTON -> builder.onboardingNextButton = viewState
    ...
  }
}.build()
```

The above is a bit more compacted and still requires an exhaustive ``when`` statement.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+      return@createInMemoryDataProviderAsync AsyncResult.Success(deferred.await())
+    }
+  }
+
+  /**
+   * Retrieves the current [SpotlightViewState] of a spotlit feature for a given profile.
+   *
+   * @param profileId the ID of the profile that will be viewing the spotlight
+   * @param feature the spotlight feature to be spotlit
+   *
```
BenHenning: Nit: remove extra line.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -0,0 +1,182 @@
...
+  private val dataProviders: DataProviders,
+) {
+  // thrown when spotlight feature is not set while retrieving or marking spotlight view states
+  class SpotlightFeatureNotFoundException(message: String) : Exception(message)
+
+  private val cacheStoreMap =
+    mutableMapOf<ProfileId, PersistentCacheStore<SpotlightStateDatabase>>()
+
+  /**
+   * Marks the [SpotlightViewState] of a spotlit feature for a given profile as seen
```
BenHenning:
```suggestion
   * Marks the [SpotlightViewState] of a spotlit feature for a given profile as seen.
```

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Superclass for all the SpotlightViewStates
```
BenHenning: Not technically a class, and this documentation could be slightly confusing. Consider, instead, explaining what a spotlight is (generally) and what this proto represents with respect to those.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+message Spotlight {
+  oneof feature {
```
BenHenning, line 12: This should be documented, too (see other oneof protos for references).

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+    // feature for onboarding screen next button
```
BenHenning: This should be a complete sentence. Furthermore, I think you could make it a bit clearer. Consider:

```suggestion
    // Corresponds to the onboarding screen's next button.
```

Ditto for the others below.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+
+    // feature for voice over icon in exploration screen
+    SpotlightViewState voiceover_play_icon = 7;
+
+    // feature for voice over language options in audio player
+    SpotlightViewState voiceover_language_icon = 8;
+  }
+}
+
+// Possible view States for a spotlight
```
BenHenning:
```suggestion
// Possible view states for a spotlight.
```

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+enum SpotlightViewState {
+  SPOTLIGHT_VIEW_STATE_UNSPECIFIED = 0;
+  SPOTLIGHT_SEEN = 1;
+  SPOTLIGHT_NOT_SEEN = 2;
```
BenHenning: These should be documented as well.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+  SpotlightViewState promoted_stories = 5;
+  SpotlightViewState lessons_back_button = 6;
+  SpotlightViewState voiceover_play_icon = 7;
+  SpotlightViewState voiceover_language_icon = 8;
+}
+
+
+
+
+
```
BenHenning: Please remove extra newlines--only one EOF newline should end the file.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+// Top level proto for storing SpotlightsViewStates
+message SpotlightStateDatabase {
+  SpotlightViewState onboarding_next_button = 1;
+  SpotlightViewState topic_lesson_tab = 2;
+  SpotlightViewState topic_revision_tab = 3;
+  SpotlightViewState first_chapter = 4;
+  SpotlightViewState promoted_stories = 5;
+  SpotlightViewState lessons_back_button = 6;
+  SpotlightViewState voiceover_play_icon = 7;
+  SpotlightViewState voiceover_language_icon = 8;
```
BenHenning: These should be documented.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Possible view States for a spotlight
+enum SpotlightViewState {
+  SPOTLIGHT_VIEW_STATE_UNSPECIFIED = 0;
+  SPOTLIGHT_SEEN = 1;
+  SPOTLIGHT_NOT_SEEN = 2;
+}
+
+// Top level proto for storing SpotlightsViewStates
```
BenHenning: I think this needs more detail, similar to my comment above for Spotlight.

model/src/main/proto/spotlight.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
+
+package model;
```
BenHenning, line 3: FYI that this proto will need to be added to the model module's BUILD.bazel file so that it can be included in Bazel builds of the app in later PRs.

BenHenning (changes requested):
Thanks @JishnuGoyal! Took a first pass review--PTAL.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done, PTAL and lmk if changes are needed.

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done -- PTAL

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

JishnuGoyal:
@BenHenning Please check if this is on the right track; ill be adding tests in my upcoming commit. PTAL

domain/src/main/java/org/oppia/android/domain/spotlight/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+        "//utility/src/main/java/org/oppia/android/util/data:async_result",
+        "//utility/src/main/java/org/oppia/android/util/data:data_provider",
+        "//utility/src/main/java/org/oppia/android/util/data:data_providers",
+        "//utility/src/main/java/org/oppia/android/util/locale:oppia_locale",
+        "//utility/src/main/java/org/oppia/android/util/profile:directory_management_util",
+        "//utility/src/main/java/org/oppia/android/util/system:oppia_clock",
+    ],
+)
+
+dagger_rules()
```
BenHenning: Please add EOF newline.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -58,10 +66,8 @@ class SpotlightStateController @Inject constructor(
 
   /**
    * Retrieves the current [SpotlightViewState] of a spotlit feature for a given profile.
-   *
```
BenHenning: This one should be kept--there's always a blank doc line between the summary and the at-clauses.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -158,12 +129,13 @@ class SpotlightStateController @Inject constructor(
     val cacheStore = if (profileId in cacheStoreMap) {
       cacheStoreMap[profileId]!!
     } else {
-      val cacheStore =
+      val cacheStore = cacheStoreMap.getOrPut(profileId) {
```
BenHenning: Now that this is using getOrPut, you can just return it. The code on lines 139 and 140 are no longer needed.

model/src/main/proto/spotlight.proto
```diff
@@ -5,55 +5,72 @@ package model;
...
+// Possible view states for a spotlight.
 enum SpotlightViewState {
+  // This is the default state for a spotlight in the beginning. In cases where the app won't be
+  // able to load a spotlight state for a feature, this state shall be reported.
   SPOTLIGHT_VIEW_STATE_UNSPECIFIED = 0;
+
+  // This corresponds to a state when the spotlight for a feature has been seen by the user.
   SPOTLIGHT_SEEN = 1;
+
+  // This corresponds to a state when the spotlight for a feature has been not seen by the user.
```
BenHenning, line 48: Is this when the spotlight hasn't been seen, or hasn't been dismissed? I think it's worth clarifying here so that readers know how to associate this documentation with what they might observe when using the app.

model/src/main/proto/spotlight.proto
```diff
@@ -5,55 +5,72 @@ package model;
 option java_package = "org.oppia.android.app.model";
 option java_multiple_files = true;
 
-// Superclass for all the SpotlightViewStates
+// Spotlights are visual elements that are shown during the initial user onboarding experience of
+// the app that highlight certain UI elements and show how/what they should be used for.
 message Spotlight {
   oneof feature {
```
BenHenning, line 12: The oneof itself should also be documented since it also corresponds to generated code.

domain/src/main/java/org/oppia/android/domain/spotlight/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+        "//domain/src/main/java/org/oppia/android/domain/oppialogger/analytics:learner_analytics_logger",
+        "//domain/src/main/java/org/oppia/android/domain/oppialogger/exceptions:controller",
+        "//model/src/main/proto:spotlight_java_proto_lite",
+        "//third_party:javax_inject_javax_inject",
+        "//utility/src/main/java/org/oppia/android/util/data:async_result",
+        "//utility/src/main/java/org/oppia/android/util/data:data_provider",
+        "//utility/src/main/java/org/oppia/android/util/data:data_providers",
+        "//utility/src/main/java/org/oppia/android/util/locale:oppia_locale",
+        "//utility/src/main/java/org/oppia/android/util/profile:directory_management_util",
+        "//utility/src/main/java/org/oppia/android/util/system:oppia_clock",
```
BenHenning: Are all of these dependencies needed? It definitely looks like a bunch of them aren't being used by the controller implementation and can probably be removed.

BenHenning (changes requested):
Thanks @JishnuGoyal! I think this is much clearer now! There's one comment you didn't respond to earlier, but otherwise I only had a few more follow-ups. Please go ahead and finish up this PR and send it back when you're ready.

Note also that it appears there are some CI checks failing, as well. I noticed the lint checks are failing in particular--have you run setup.sh locally for your repository? It adds a pre-push hook to check for common lint errors before pushing and helps reviewers avoid needing to look at various nits.

model/src/main/proto/spotlight.proto
JishnuGoyal: I'm  removing the spotlight state unspecified state as this is currently of no use, and will only confuse developers. Since our current fallback mechanism when the spotlight state can't be retrieved is to show the spotlight again, it is fine to have the default state as "not seen". Per your question, we do not have the "dismissed" state in the app currently.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: I have made the exception inherit from an IllegalArgumentException -- do I remove the custom exception completely and use the above?

JishnuGoyal:
PTAL @BenHenning for the newly added tests

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -97,6 +101,13 @@ class SpotlightStateController @Inject constructor(
       }
   }
 
+  private fun spotlightStateRetrieverHelper(spotlightViewState: SpotlightViewState):
+    SpotlightViewState {
+      return if (spotlightViewState == SpotlightViewState.SPOTLIGHT_VIEW_STATE_UNSPECIFIED) {
+        SpotlightViewState.SPOTLIGHT_NOT_SEEN
+      } else spotlightViewState
+    }
```
BenHenning:
```suggestion
  private fun spotlightStateRetrieverHelper(
    spotlightViewState: SpotlightViewState
  ): SpotlightViewState {
    return if (spotlightViewState == SpotlightViewState.SPOTLIGHT_VIEW_STATE_UNSPECIFIED) {
      SpotlightViewState.SPOTLIGHT_NOT_SEEN
    } else spotlightViewState
  }
```

Breaking in this way avoids the extra awkward indentation being needed.

BenHenning: Perhaps you could do an extension here, but more importantly the name could be more clear. Consider:

```kotlin
private fun SpotlightViewState.ensureSeenOrNotSeen(): SpotlightViewState =
  if (this == SPOTLIGHT_VIEW_STATE_UNSPECIFIED) SPOTLIGHT_NOT_SEEN else this
```

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -79,14 +83,14 @@ class SpotlightStateController @Inject constructor(
...
         val viewState = when (feature) {
-          ONBOARDING_NEXT_BUTTON -> it.onboardingNextButton
-          TOPIC_LESSON_TAB -> it.topicLessonTab
-          TOPIC_REVISION_TAB -> it.topicRevisionTab
-          FIRST_CHAPTER -> it.firstChapter
-          PROMOTED_STORIES -> it.promotedStories
-          LESSONS_BACK_BUTTON -> it.lessonsBackButton
-          VOICEOVER_PLAY_ICON -> it.voiceoverPlayIcon
-          VOICEOVER_LANGUAGE_ICON -> it.voiceoverLanguageIcon
+          ONBOARDING_NEXT_BUTTON -> spotlightStateRetrieverHelper(it.onboardingNextButton)
```
BenHenning: I suggest making the change down when viewState is being returned so that it only needs to happen in one place.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+@Config(application = SpotlightStateControllerTest.TestApplication::class)
+class SpotlightStateControllerTest {
+
+  @Inject
+  lateinit var spotlightStateController: SpotlightStateController
+
+  @Inject
+  lateinit var dataProviderTestMonitor: DataProviderTestMonitor.Factory
+
+  private val profileId = ProfileId.newBuilder().setInternalId(0).build()
```
BenHenning: Suggest naming this ``profileId0`` to differentiate with the one below.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testMarkSpotlightState_validFeature_notYetMarked_returnsSuccess() {
+    val markSpotlightProvider =
+      spotlightStateController.markSpotlightViewed(profileId, FIRST_CHAPTER)
+    dataProviderTestMonitor.waitForNextSuccessfulResult(markSpotlightProvider)
```
BenHenning: The ``waitFor*`` methods are more for operational parts of tests rather than explicit verifications. We have different methods for that. This test would be a bit more idiomatic if it were structured like:

```suggestion
    val markSpotlightProvider =
      spotlightStateController.markSpotlightViewed(profileId, FIRST_CHAPTER)
    val monitor = monitorFactory.createMonitor(markSpotlightProvider)
    testCoroutineDispatchers.runCurrent()
    
    monitor.ensureNextResultIsSuccess()
```

Here we're actually verifying that the provider works as expected (i.e. that the result is only available after the background processing finishes).

Ditto for the other tests below (prerequisite state like an already-marked spotlight should continue to use the waitFor* methods).

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  @Test
+  fun testMarkSpotlightView_invalidFeature_returnsFailure() {
+    val invalidFeature = Spotlight.FeatureCase.FEATURE_NOT_SET
+    val markSpotlightProvider =
+      spotlightStateController.markSpotlightViewed(profileId, invalidFeature)
+    dataProviderTestMonitor.waitForNextFailureResult(markSpotlightProvider)
```
BenHenning: This should verify that the correct failure is being returned (e.g. by checking the instance type & exception cause method).

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+  @Test
+  fun testRetrieveSpotlightViewState_firstChapter_notMarked_returnsSpotlightStateNotSeen() {
+    val retrieveSpotlightStateProvider =
+      spotlightStateController.retrieveSpotlightViewState(profileId, FIRST_CHAPTER)
+    val result = dataProviderTestMonitor.waitForNextSuccessfulResult(retrieveSpotlightStateProvider)
+    assertEquals(result, SpotlightViewState.SPOTLIGHT_NOT_SEEN)
+  }
+
+  @Test
+  fun testRetrieveSpotlightViewState_firstChapter_marked_returnsSpotlightStateSeen() {
```
BenHenning, line 142: Perhaps a test verifying that marking a _different_ spotlight does not cause one to be marked as seen) would also be nice to add.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+  @Test
+  fun testMarkSpotlightView_invalidFeature_returnsFailure() {
+    val invalidFeature = Spotlight.FeatureCase.FEATURE_NOT_SET
+    val markSpotlightProvider =
+      spotlightStateController.markSpotlightViewed(profileId, invalidFeature)
+    dataProviderTestMonitor.waitForNextFailureResult(markSpotlightProvider)
+  }
+
+  @Test
+  fun testMarkSpotlightState_validFeature_differentProfile_returnsSuccess() {
```
BenHenning, line 125: Suggest also verifying that saved state is unique across by profiles by adding a test that marks a spotlight as seen in one profile and then fetch it in a second (to demonstrate that it's unseen).

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+  fun testRetrieveSpotlightViewState_voiceoverPlayIcon_marked_returnsSpotlightStateSeen() {
+    markSpotlightSeen(VOICEOVER_PLAY_ICON)
+    val retrieveSpotlightStateProvider =
+      spotlightStateController.retrieveSpotlightViewState(profileId, VOICEOVER_PLAY_ICON)
+    val result = dataProviderTestMonitor.waitForNextSuccessfulResult(retrieveSpotlightStateProvider)
+    assertEquals(result, SpotlightViewState.SPOTLIGHT_SEEN)
+  }
+
+  @Test
+  fun testRetrieveSpotlightViewState_voiceoverLanguageIcon_notMarked_returnsSpotlightStateNotSeen() { // ktlint-disable max-line-length
```
BenHenning:
```suggestion
  fun testRetrieveSpotlightViewState_voiceoverLangIcon_notMarked_returnsSpotlightStateNotSeen() {
```

Prefer not having to disable the max-line length check. :)

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+    )
+
+    val spotlightStateProvider =
+      spotlightStateController.retrieveSpotlightViewState(profileId1, FIRST_CHAPTER)
+    val result = dataProviderTestMonitor.waitForNextSuccessfulResult(spotlightStateProvider)
+    assertEquals(result, SpotlightViewState.SPOTLIGHT_SEEN)
+  }
+
+  @Test
+  fun testRetrieveSpotlightViewState_validFeature_notMarked_differentProfile_returnsSpotlightNotSeen() { // ktlint-disable max-line-length
```
BenHenning:
```suggestion
  fun testRetrieveSpotlightViewState_validFeature_notMarked_diffProfile_returnsSpotlightNotSeen() {
```

Ditto here.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+      spotlightStateController.markSpotlightViewed(
+        profileId,
+        spotlightFeature
+      )
+    )
+  }
+
+  private fun setUpTestApplicationComponent() {
+    ApplicationProvider.getApplicationContext<TestApplication>()
+      .inject(this)
```
BenHenning:
```suggestion
    ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
```

Can fit on one line.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -0,0 +1,388 @@
...
+    @Provides
+    @LoadLessonProtosFromAssets
+    fun provideLoadLessonProtosFromAssets(testEnvironmentConfig: TestEnvironmentConfig): Boolean =
+      testEnvironmentConfig.isUsingBazel()
+
+    @Provides
+    @LearnerStudyAnalytics
+    fun provideLearnerStudyAnalytics(): PlatformParameterValue<Boolean> {
+      // Enable the study by default in tests.
+      return PlatformParameterValue.createDefaultParameter(defaultValue = true)
```
BenHenning: Is this necessary? The parameter doesn't seem to actually affect the controller from what I can see. Could we just depend on the test parameter module to simplify things a bit?

BenHenning (commented):
Thanks @JishnuGoyal! I think the tests look pretty good. I have a bunch of small follow-up improvement suggestions. I think after all my comments have been addressed that the PR will be quite close to done.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: so elegant!❤

JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: Thanks for the tip, noted! Done.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/BUILD.bazel
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done, please check if the updated doc looks fine

model/src/main/proto/spotlight.proto
JishnuGoyal: done - please check if the doc looks fine

domain/src/main/java/org/oppia/android/domain/spotlight/BUILD.bazel
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: I'm facing a problem doing this. Pushed my latest code -- PTAL;  P.S. the markSpotlightSeen() function has a try catch block at the end.

JishnuGoyal:
PTAL @BenHenning , thanks!

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
```diff
@@ -137,19 +136,14 @@ class SpotlightStateController @Inject constructor(
...
-      cacheStoreMap[profileId] = cacheStore
-      cacheStore
+    val cacheStore = cacheStoreMap.getOrPut(profileId) {
+      cacheStoreFactory.createPerProfile(
+        CACHE_NAME,
+        SpotlightStateDatabase.getDefaultInstance(),
+        profileId
+      )
     }
+    cacheStoreMap[profileId] = cacheStore
```
BenHenning: I don't think you need this line, either, since ``getOrPut`` ensures that it's in the map.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -93,37 +96,47 @@ class SpotlightStateControllerTest {
...
   }
 
   @Test
   fun testMarkSpotlightView_invalidFeature_returnsFailure() {
     val invalidFeature = Spotlight.FeatureCase.FEATURE_NOT_SET
     val markSpotlightProvider =
-      spotlightStateController.markSpotlightViewed(profileId, invalidFeature)
-    dataProviderTestMonitor.waitForNextFailureResult(markSpotlightProvider)
+      spotlightStateController.markSpotlightViewed(profileId0, invalidFeature)
+    assertThrows(SpotlightStateController.SpotlightFeatureNotFoundException::class) {
```
BenHenning: This doesn't look right--we should verify the except returned by ``ensureNextResultIsFailing`` is the correct exception with the correct cause message. There shouldn't be an actual exception thrown here.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -234,15 +247,15 @@ class SpotlightStateControllerTest {
   fun testRetrieveSpotlightViewState_voiceoverPlayIcon_marked_returnsSpotlightStateSeen() {
     markSpotlightSeen(VOICEOVER_PLAY_ICON)
     val retrieveSpotlightStateProvider =
-      spotlightStateController.retrieveSpotlightViewState(profileId, VOICEOVER_PLAY_ICON)
+      spotlightStateController.retrieveSpotlightViewState(profileId0, VOICEOVER_PLAY_ICON)
     val result = dataProviderTestMonitor.waitForNextSuccessfulResult(retrieveSpotlightStateProvider)
     assertEquals(result, SpotlightViewState.SPOTLIGHT_SEEN)
```
BenHenning: Oh I missed this before--this should use Truth not JUnit assertions. Ditto for everywhere else in the test suite.

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
```diff
@@ -260,8 +273,10 @@ class SpotlightStateControllerTest {
   fun testRetrieveSpotlightViewState_invalidFeature_returnsFailure() {
     val invalidFeature = Spotlight.FeatureCase.FEATURE_NOT_SET
     val spotlightStateProvider =
-      spotlightStateController.retrieveSpotlightViewState(profileId, invalidFeature)
-    dataProviderTestMonitor.waitForNextFailureResult(spotlightStateProvider)
+      spotlightStateController.retrieveSpotlightViewState(profileId0, invalidFeature)
+    val monitor = dataProviderTestMonitor.createMonitor(spotlightStateProvider)
+    testCoroutineDispatchers.runCurrent()
+    monitor.ensureNextResultIsFailing()
```
BenHenning: We should verify the specific error causing the failure. Ditto for other such cases.

model/src/main/proto/spotlight.proto
```diff
@@ -8,6 +8,7 @@ option java_multiple_files = true;
 // Spotlights are visual elements that are shown during the initial user onboarding experience of
 // the app that highlight certain UI elements and show how/what they should be used for.
 message Spotlight {
+  // Determines the UI element being spotlit
```
BenHenning:
```suggestion
  // Determines the UI element being spotlit.
```

model/src/main/proto/spotlight.proto
BenHenning: Unless I'm mistaken, it doesn't seem like the doc line has been updated. Am I missing something?

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
BenHenning: I left a new comment in the latest review pass--I think you might be checking it incorrectly (see that comment--we shouldn't be using ``assertThrows`` here). I also suggest looking at other tests that use the monitor for testing failures; they'll probably be really helpful as references.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
BenHenning: I don't mind either way; I think it's fine to use a custom exception.

BenHenning (commented):
Thanks @JishnuGoyal. Took another pass--PTAL.

model/src/main/proto/spotlight.proto
JishnuGoyal: We do not have the dismiss feature (which was something that was being planned before the full product planning). So this is the state when the feature has not yet been seen by the user.

domain/src/main/java/org/oppia/android/domain/spotlight/SpotlightStateController.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done, thanks this helps!

domain/src/test/java/org/oppia/android/domain/spotlight/SpotlightStateControllerTest.kt
JishnuGoyal: done

model/src/main/proto/spotlight.proto
JishnuGoyal: done

JishnuGoyal:
@BenHenning PTAL please take another pass at this, thanks.

BenHenning (commented):
Thanks @JishnuGoyal. Other than the CI failures, the PR LGTM. Please assign back once they're fixed.

JishnuGoyal:
PTAL @BenHenning

BenHenning (approved):
Thanks @JishnuGoyal. Everything LGTM here.

PR #4500, opened 2022-08-15
Fix #4492 & 3537: Improved hints and solution UI UX

### LINKED ISSUE #3537: Merge hints_summary.xml into single xml file
Currently there are 4 versions of `hints_summary.xml` file merge it into single xml file.

We can use https://text-compare.com/ to compare two versions of this file and for all the differences we can create variables in `dimens.xml` file and use it accordingly.

It can happen that it becomes really hard to merge all 4 files into single file, in such cases prefer code-readability over single file. Therefore it can be fine if in the end the code still contains 2 or 3 files.

**Note**: **In PR, make sure you add before and after screenshot of mobile-portrait, mobile-landscape, tablet-portrait and tablet-landscape for comparison and make sure that there is not difference between before and after UI.**

### LINKED ISSUE #4492: Improve presentation of hints for the student.
The current hints design is not intuitive, and students don't intuitively understand how to use it. 

Here are @mschanteltc's updated designs for the hints screen. Generally, there are 3 states of the Hint Card -- Unrevealed, Expanded, and Collapsed. (Note: open the links in an incognito window to play the interactive prototype.)

- [Hint 1](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/619724cc-0ecf-4c32-a452-432fded803f5)
    - Note: For all hints, when the user hits "Reveal Hint," it will change from the Unrevealed state to the Expanded state.
- [Hint 2](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/5311124a-b143-4916-aa46-6e318aca5650) 
    -  **IMPORTANT**: Please note that there is a design limitation of this prototype which is there because implementing it in the prototype would take too long. Specifically: within the prototype, when Hint 2 is revealed and Hint 1 is expanded afterwards, Hint 2 returns back to "Reveal Hint." This should not happen. If the hint is revealed, it can never go back to the Unrevealed state.
- [Solution](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/545c6f32-2fec-4a8b-bb53-cecfb8e52f31)
    - **IMPORTANT**: Please note that there is a design limitation of this prototype -- when a Hint Card is opened, it currently overlaps any Hint Cards beneath it. This is **not** supposed to happen. Instead, all cards below the expanded hint card should transition downwards. This can be seen in the Hint 2 mock above, when Hint 1 is expanded. 

Overall note (**IMPORTANT**): In the prototypes above, the hint/solution cards cover the "Return to Lesson" button when expanded. This **should not** happen, and it only works the way it currently does because of the limited time available to implement the prototype. Instead, in all cases, the "Return to Lesson" button should transition downwards ([such as when the Hint 1 Card is expanded in this prototype](https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/5311124a-b143-4916-aa46-6e318aca5650/)).

Notes on the design:
- It's fine for multiple hint cards to be expanded at the same time. This is because students might want to refer to other hints while reading a given hint, in order to get the necessary context.
- For "Reveal solution", it's fine to keep the confirmation dialog. But if dropping it is easier, we can do that too. Let's do what's technically easiest here.
- Hints should appear one by one after a certain amount of time has elapsed. When the user is in the Hint Modal, the timer should not be counting. This is because there should be at least some time which the learner can use to think about and apply the suggestions in the hint before we prompt them with a follow-up hint.

**Additional context**

Here is what the hints screen currently looks like. The dropdown arrow is not clear and there are too many clicks needed to get to the hint. Students also struggle to exit the hints page.
image7

rt4914:
## Explanation
Fixes #4492 by introducing a complete revamp of the hint & solution management flow in the app (see designs below).

Fixes #3537 by merging the hint summary layouts.

Note that one part of #4492 isn't implemented: pausing the hint timer while hints are being viewed. Ensuring this is properly implemented is being tracked as part of #4510.

Two new test exemptions were added: one for a new view model file (which seems sensible since most of the changes in this PR are aesthetic and not functional), and a second for a new hints configuration module that's specific to the alpha user study build of the app (which is temporary and we generally don't test thoroughly, anyway).

**Designs**
https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/619724cc-0ecf-4c32-a452-432fded803f5/specs/
https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/5311124a-b143-4916-aa46-6e318aca5650/specs/
https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/545c6f32-2fec-4a8b-bb53-cecfb8e52f31/specs/

## Essential Checklist

## For UI-specific PRs only

## Return to Lesson button

**NOTE**: This button was created later on that's why not all screenshots contain this button but the below screenshot should suffice. It is fully functional too.

   

 

## Dark Mode (Not Implemented)

Do not have designs for dark mode

## Accessibility Output

There are few issues but not solving them right now.

## Video Output

https://user-images.githubusercontent.com/9396084/184695958-d23f1d77-a2f0-43bd-848a-43483f11fb04.mp4


app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragment.kt
```diff
@@ -90,13 +90,10 @@ class HintsAndSolutionDialogFragment :
...
     savedInstanceState: Bundle?
-  ): View? {
+  ): View {
     if (savedInstanceState != null) {
-      currentExpandedHintListIndex =
-        savedInstanceState.getInt(CURRENT_EXPANDED_LIST_INDEX_SAVED_KEY, -1)
-      if (currentExpandedHintListIndex == -1) {
-        currentExpandedHintListIndex = null
-      }
+      expandedItemsList =
```
BenHenning, line 95: Noting that it would be preferable to use a proto here to avoid passing around mutable state.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
```diff
@@ -328,46 +290,34 @@ class HintsAndSolutionDialogFragmentPresenter @Inject constructor(
...
+    binding.returnToLessonButton.setOnClickListener {
+      (fragment.requireActivity() as? HintsAndSolutionListener)?.dismiss()
+    }
+  }
+
   private fun handleAllHintsExhausted(allHintsExhausted: Boolean) {
-    if (itemList[itemList.size - RECYCLERVIEW_INDEX_CORRECTION_MULTIPLIER] is SolutionViewModel) {
-      val solutionViewModel =
-        itemList[itemList.size - RECYCLERVIEW_INDEX_CORRECTION_MULTIPLIER] as SolutionViewModel
+    if (itemList[itemList.size - 2] is SolutionViewModel) {
```
BenHenning, line 317: This seems a bit like a magic value--why the '2'?

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
```diff
@@ -238,66 +227,39 @@ class HintsAndSolutionDialogFragmentPresenter @Inject constructor(
...
-          bindingAdapter.notifyItemChanged(currentExpandedHintListIndex!!)
-        } else {
-          previousIndex?.let {
-            bindingAdapter.notifyItemChanged(it)
-          }
-          currentExpandedHintListIndex?.let {
-            bindingAdapter.notifyItemChanged(it)
-          }
-        }
+        expandOrCollapseItem(position)
```
BenHenning, line 238: For this & above: nice simplification!

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
```diff
@@ -328,46 +290,34 @@ class HintsAndSolutionDialogFragmentPresenter @Inject constructor(
...
     }
   }
 
+  private fun bindReturnToLessonViewModel(
+    binding: ReturnToLessonButtonItemBinding,
+    returnToLessonViewModel: ReturnToLessonViewModel
+  ) {
+    binding.buttonViewModel = returnToLessonViewModel
+
+    binding.returnToLessonButton.setOnClickListener {
```
BenHenning, line 309: Could be moved to the view model for slightly simpler binding.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
```diff
@@ -63,15 +56,13 @@ class HintsViewModel @Inject constructor(
     for (index in hintList.indices) {
       if (itemList.isEmpty()) {
         addHintToList(index, hintList[index])
-      } else if (itemList.size > 1) {
+      } else if (itemList.size > 0) {
         val isLastHintRevealed =
-          (itemList[itemList.size - RECYCLERVIEW_INDEX_CORRECTION_MULTIPLIER] as HintsViewModel)
+          (itemList[itemList.size - 1] as HintsViewModel)
```
BenHenning: Could use itemList.last() here & below.

app/src/main/res/layout/return_to_lesson_button_item.xml
```diff
@@ -0,0 +1,32 @@
...
+  <data>
+
+    <import type="android.view.View" />
+
+    <variable
+      name="buttonViewModel"
+      type="org.oppia.android.app.hintsandsolution.ReturnToLessonViewModel" />
+  </data>
+
+  <androidx.constraintlayout.widget.ConstraintLayout
```
BenHenning: Probably don't need this since there's just one element. Also, we aren't setting a width/height on the button.

app/src/main/res/values/color_palette.xml
```diff
@@ -67,4 +67,9 @@
   <color name="color_palette_checkbox_selected_state_color">@color/color_def_checkbox_green</color>
   <color name="color_palette_mark_completed_text_color">@color/color_def_white</color>
   <color name="color_palette_mark_completed_all_text_color">@color/color_def_white</color>
+  <color name="color_palette_hints_and_solution_summary_card_border_color">@color/color_def_accessible_light_grey_2</color>
```
BenHenning, line 70: These colors shouldn't be specifically tied to hints and solutions, but instead tied to the types of UI elements being colorized.

app/src/main/res/values/strings.xml
```diff
@@ -433,8 +433,9 @@
   <string name="show_hints_and_solution">Show hints and solution</string>
   <string name="hints_andSolution_close_icon_description">Navigate up</string>
   <string name="hints_toolbar_title">Hints</string>
+  <string name="show_hint">Show Hint</string>
+  <string name="show_solution">Show</string>
   <string name="reveal_solution">Reveal Solution</string>
-  <string name="reveal_hint">Reveal Hint</string>
```
BenHenning, line 437: We should probably keep the old name so that we can keep translations.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -1344,6 +1338,7 @@ class StateFragmentLocalTest {
     appStringIetfTag = "ar",
     appStringAndroidLanguageId = "ar"
   )
+  @Ignore // TODO(#4492): Update the arabic string for show hint button.
```
BenHenning: Reminder to fix (which I think can be if translations are restored).

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -1729,14 +1724,27 @@ class StateFragmentLocalTest {
...
+  @Test
+  fun testStateFragment_openHintsAndSolution_checkReturnToLessonButtonIsVisible() {
+    launchForExploration(FRACTIONS_EXPLORATION_ID_1).use {
+      startPlayingExploration()
+      playThroughFractionsState1()
+      submitTwoWrongAnswersForFractionsState2()
+
+      openHintsAndSolutionsDialog()
+      onView(allOf(withId(R.id.return_to_lesson_button), isDisplayed()))
+        .inRoot(isDialog())
```
BenHenning: Nit: can fit on one line?

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
```diff
@@ -1729,14 +1724,27 @@ class StateFragmentLocalTest {
...
+      pressShowHintButton(hintPosition = 0)
       closeHintsAndSolutionsDialog()
 
       onView(withId(R.id.hint_bulb))
         .check(matches(withDrawable(R.drawable.ic_keyboard_arrow_right_white_48)))
     }
   }
 
+  @Test
+  fun testStateFragment_openHintsAndSolution_checkReturnToLessonButtonIsVisible() {
```
BenHenning, line 1732: What about a test verifying clicking on the return to lesson button?

BenHenning, line 1732: Might be worth adding these new tests to questions' local tests, too.

BenHenning (commented):
Thanks for much @rt4914! Really well done implementation. Just had a few comments that I plan to follow up on directly.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
rt4914: `2` because there is a button `Return to Lesson` at the bottom of the list, the index for which will be `itemList.size - 1` and therefore the index for solution will be `itemList.size - 2`

Added following comment above this code
```
    // The last item of the list is ReturnToLessonViewModel and therefore second last item is
    // SolutionViewModel as a result subtracting 2 from itemList size.
```

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsAndSolutionDialogFragmentPresenter.kt
rt4914: Thanks.

app/src/main/java/org/oppia/android/app/hintsandsolution/HintsViewModel.kt
rt4914: Updated at two places in this file and checked in presenter file too.

app/src/main/res/layout/return_to_lesson_button_item.xml
rt4914: Updated and removed `ConstraintLayout`

app/src/main/res/values/strings.xml
rt4914: Restored all `reveal_hint` strings.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
rt4914: Removed as translations have been restored.

app/src/test/java/org/oppia/android/app/player/state/StateFragmentLocalTest.kt
rt4914: Done

app/src/main/res/values/strings.xml
```diff
@@ -433,9 +433,9 @@
   <string name="show_hints_and_solution">Show hints and solution</string>
   <string name="hints_andSolution_close_icon_description">Navigate up</string>
   <string name="hints_toolbar_title">Hints</string>
-  <string name="show_hint">Show Hint</string>
   <string name="show_solution">Show</string>
   <string name="reveal_solution">Reveal Solution</string>
+  <string name="reveal_hint">Reveal Hint</string>
```
BenHenning: We can still change the text to Show Hint without renaming the string (per the mocks).

BenHenning:
FWIW this seems like a complete fix of #4492 as far as I can tell, though some investigation may be needed with regards to the hint timing logic.

BenHenning:
Confirming that the current logic will NOT pause the timer while the hints are being viewed (I think this is a change in behavior from the old hints implementation). I think fixing it will be straightforward, but I'll follow up soon.

BenHenning:
Okay, this is very much not going to be trivial, so I think we may need to skip it for this release. Will follow up with Sean on product priority here.

The main issue is that hints are handled in the domain layer entirely separately from the frontend lifecycle. We almost certainly would need to either somehow provide signal from the frontend when hints are being shown (and not mess this up since it'll break the hints timer), or manage hint state based on the observation of its results (similar to how ExplorationProgressController used to work), but this is difficult and not at all straightforward since hint behaviors are observed *through* ExplorationProgressController. I suspect we may need the DataProvider dynamic transformation that I'm working on for the beta release in order to even start tackling this issue.

app/src/main/res/values/strings.xml
BenHenning: Resolving this in my latest changes.

seanlip:
I would say, just make the timer for showing the next hint long enough so that it's unlikely that a student will get another hint while the current hint is showing. 2 or 2.5 minutes should be reasonable.

(Note: this is still a stopgap and we should fix it properly in the long term.)

BenHenning:
> I would say, just make the timer for showing the next hint long enough so that it's unlikely that a student will get another hint while the current hint is showing. 2 or 2.5 minutes should be reasonable.
> 
> (Note: this is still a stopgap and we should fix it properly in the long term.)

@seanlip fortunately we can very easily tweak this specifically for the study build. We currently have three hint delays which I think should be adjusted as such:
- Delay for initial hint: 60 seconds (this should be kept the same since the issue won't affect it)
- Delay for subsequent hints if the user doesn't try to answer: 30 seconds (bump to 2.5 minutes)
- Delay for subsequent hints if the user submits a wrong answer: 10 seconds (keep since the user is more likely on the card at this point, and we want them to see the hint quickly)

Does this seem sensible?

seanlip:
Yes, seems reasonable. I slightly lean towards making the third one 15 seconds instead of 10, to give the learner some time to read the wrong answer feedback.

BenHenning (approved):
Approving this with some open comments. https://github.com/oppia/oppia-android/issues/4510 is tracking addressing those comments post-merge. Plus, an additional review should take place for the changes I contributed directly to the PR (since they haven't yet been reviewed).

Issue #4235, opened 2022-03-07
[A11Y] Fix accessibility issues for ClickableSpan


rt4914:
### Is your feature request related to a problem? Please describe.

Currently, Oppia Android relies on ClickableSpan in multiple places (e.g., Concept Cards, Revision Cards, Exploration, Question Player, etc.) for handling in-text interactions within HTML content parsed by [HtmlParser](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt#L24) . However, `ClickableSpan` has known accessibility issues and does not work reliably with screen readers like TalkBack.

This has caused **A11Y tests in `StoryFragmentTest` to fail**, and Android Studio itself recommends migrating to `URLSpan`.
However, `ClickableSpan` has **known accessibility issues**:
- Screen readers like TalkBack do not always announce the clickable element.
- Users are sometimes not given the correct actionable context (“double tap to activate link” is missing).
- A11Y tests in `StoryFragmentTest` are failing due to this issue.
- Android Studio itself recommends migrating to `URLSpan` instead of `ClickableSpan`.

### Repro Steps
1. Enable **TalkBack** on an Android device/emulator.
2. Open Oppia and navigate to a `StoryFragment`:
   - From the **Home screen**, select a **Topic**.
   - Inside the topic page, choose a **Story**.
   - This opens the **StoryFragment**, which shows a list of chapters.
   - Tap a chapter to open its Concept Card or exploration.
3. Locate text that contains a clickable link (e.g., in a Concept Card).
4. Try to explore the link text using TalkBack.

**Observed behavior:**  
- TalkBack does not always announce the clickable link properly.
- Sometimes no actionable context is provided (e.g., it just reads the text without “link” or “double tap to activate”).

**Expected behavior:**  
- TalkBack should consistently announce the link as **clickable**.
- It should provide actionable feedback like “double tap to activate link”.
- The behavior should be consistent across all clickable spans (Concept Cards, Policy links, Revision Cards, Exploration, Question Player, etc.).

### Proposed Solution

We need a **generic and robust replacement** for `ClickableSpan` that ensures accessibility while preserving Oppia’s routing logic.

#### 1. Investigation Phase
* Check TalkBack behavior between ClickableSpan vs. URLSpan. 
* Prototype using URLSpan backed by IntentFilter to evaluate screen reader compatibility. 
[...]

PR #4554, opened 2022-09-05
Fix #4450: Add next and previous card options on revision screen

### LINKED ISSUE #4450: Add next and previous card options on revision screen
On the revision screen, the users need to exit to the topic page and re-select the next revision card.

Add option to navigate to the next and previous cards from within the revision screen. In order for the users to better understand  the usage of the revision and lesson screen (and to differentiate between the two), change the color of the revision screen headers to orange.

JishnuGoyal:
## Explanation

Fixes #4450

This PR creates navigation cards on the revision card screen so that the users can easily view the next revision item without having to return to the topic screen, thus fixing the issue #4450 as a part of the Interactive Onboarding Flow (GSoC) project. The current implementation makes sure that the functionality can be viewed correctly on a mobile screen.
This PR is backed up by unit tests.

Link to the original mock: https://xd.adobe.com/view/3dca36c2-5115-419c-b25e-0f10526b077c-6899/screen/9c633b9e-aae5-428d-a037-efcc0338c4df/specs/ 

## Essential Checklist

## For UI-specific PRs only

Mobile (Portrait)
1. When next and prev both are available
image

2. When only next is available
image

3. When only previous is available
image

Mobile (Landscape)
image

Tablet (Portrait)  - prev button only
image

Tablet (Portrait) - both buttons
image

Tablet (portrait) - next button only
image

Tablet (Landscape) 
image

RTL:
image

RTL (land):
image

Confirming that the Espresso tests are passing locally:
image

Functionality demo:
https://user-images.githubusercontent.com/64526117/192758757-37e18649-2a5f-48dd-8917-19d5c15b143d.mp4

TalkBack:https://user-images.githubusercontent.com/64526117/192503526-8ef9100d-083e-40ef-ad17-04f857955354.mp4


JishnuGoyal:
PTAL @BenHenning please take an initial pass -- does the general implementation look correct?

BenHenning:
> PTAL @BenHenning please take an initial pass -- does the general implementation look correct?

Hi @JishnuGoyal. I didn't review it in detail (i.e. I have a bunch of more detailed feedback that I can give), but I think it looks solid at a high-level. I do have some high-level feedback:
- It looks like you had extra strings pulled in--please make sure to fix this.
- I suggest controlling the visibility of the tiles via databinding instead of directly in Kotlin. It's a bit more idiomatic for how we usually do things, and I suspect it may simplify things in the binding flow a bit.
- Consider combining the next/previous state into a single LiveData with an object that can represent the four different possible states: no previous/next, one previous no next, no previous one next, one previous one next. A sealed object would be a really nice fit here, but I don't think that will work in databinding so maybe just a simple, single data class or perhaps a proto oneof would work best.
- Before sending this for review, I also suggest: adding missing tests, fixing CI checks, adding tablet/portrait screenshots to the PR description, demonstrate the a11y flow via a video in the PR description, and adding RTL screenshots to the PR description.

JishnuGoyal:
PTAL @BenHenning , thanks!

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
```diff
@@ -29,6 +37,15 @@ class RevisionCardActivity :
...
     }
+    supportActionBar?.setBackgroundDrawable(
+      ResourcesCompat.getDrawable(
+        resources,
+        R.color.color_def_oppia_brown_dark,
+        theme
+      )
+    )
+    window.statusBarColor =
+      ResourcesCompat.getColor(resources, R.color.color_def_oppia_brown_dark2, theme)
```
BenHenning: Why is this change needed? We generally don't set colors like this since we need to make sure things are wired properly for dark mode.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
```diff
@@ -2,22 +2,30 @@ package org.oppia.android.app.topic.revisioncard
...
-  InjectableAppCompatActivity(), ReturnToTopicClickListener, ConceptCardListener {
+  InjectableAppCompatActivity(),
+  ReturnToTopicClickListener,
+  ConceptCardListener,
+  RouteToRevisionCardListener {
 
   @Inject
   lateinit var revisionCardActivityPresenter: RevisionCardActivityPresenter
 
+  @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
```
BenHenning: This implies there is code below that won't work on KitKat--what is it? Normally we need to get things behind API checks for this reason.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -79,6 +86,45 @@ class RevisionCardFragmentPresenter @Inject constructor(
     return binding.root
   }
 
+  private fun setupRevisionNavigationCards(
```
BenHenning:
```suggestion
  private fun setUpRevisionNavigationCards(
```

Nit: verb vs. noun form.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
```diff
@@ -32,6 +34,53 @@ class RevisionCardViewModel @Inject constructor(
...
+  }
+
+  val previousSubtopicLiveData: LiveData<EphemeralSubtopic> by lazy {
+    Transformations.map(topicLiveData, ::processPreviousSubtopicData)
+  }
+
+  private fun processPreviousSubtopicData(
+    topicLiveData: AsyncResult<EphemeralTopic>
+  ): EphemeralSubtopic? {
+    return if (subtopicId == 0) EphemeralSubtopic.getDefaultInstance()
```
BenHenning: Is this check actually needed? It seems like the code below will result in a default instance being returned, anyway.

app/src/main/res/values/strings.xml
```diff
@@ -571,4 +571,5 @@
   <string name="profile_edit_fragment_test_activity_label">Profile Edit Fragment Test Activity</string>
   <string name="administrator_controls_fragment_test_activity_label">Administrator Controls Fragment Test Activity</string>
   <string name="select_a_topic_to_start">Select a Topic to Start</string>
+  <string name="continue_studying">Continue Studying</string>
```
BenHenning: This string name should describe what the button is/for rather than reflect its contents (as that's more useful for identifying it, helps with the fact that string names are global identifiers in the app, and future proofs the string against its label changing in the future).

app/src/main/res/values/color_defs.xml
```diff
@@ -6,6 +6,7 @@
   <color name="color_def_oppia_light_green">#F0FFFF</color>
   <color name="color_def_oppia_light_yellow">#FFFFF0</color>
   <color name="color_def_oppia_brown_dark">#BE563C</color>
+  <color name="color_def_oppia_brown_dark2">#882814</color>
```
BenHenning: According to https://colors.artyclick.com/color-name-finder/ this is called "reddish_brown", so maybe we could use that name, instead?

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
-        SUBTOPIC_TOPIC_ID
+        SUBTOPIC_TOPIC_ID_0
       )
     ).use {
       testCoroutineDispatchers.runCurrent()
 
-      onView(withId(R.id.revision_card_return_button))
-        .check(matches(withText(R.string.return_to_topic)))
+      onView(withId(R.id.previous_navigation_card))
+        .check(matches(not((isDisplayed()))))
```
BenHenning: Nit: can fit on one line.

Ditto elsewhere (where relevant).

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
         FRACTIONS_TOPIC_ID,
-        SUBTOPIC_TOPIC_ID
+        SUBTOPIC_TOPIC_ID_0
       )
     ).use {
       testCoroutineDispatchers.runCurrent()
 
-      onView(withId(R.id.revision_card_return_button))
-        .check(matches(withText(R.string.return_to_topic)))
+      onView(withId(R.id.previous_navigation_card))
```
BenHenning: Should the next one be displayed? Suggest adding a check for it and renaming the test to be "checkOnlyPrevNavCardIsNotDisplayed".

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
+        profileId.internalId,
+        FRACTIONS_TOPIC_ID,
+        SUBTOPIC_TOPIC_ID_3
+      )
+    ).use {
       testCoroutineDispatchers.runCurrent()
 
-      onView(withId(R.id.revision_card_toolbar_title))
-        .check(matches(withText("What is a Fraction?")))
+      onView(withId(R.id.next_navigation_card))
```
BenHenning: Ditto here, but for the previous nav button.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
-        SUBTOPIC_TOPIC_ID_2
+        SUBTOPIC_TOPIC_ID_1
+      )
+    ).use {
+      testCoroutineDispatchers.runCurrent()
+
+      onView(withId(R.id.previous_navigation_card)).perform(click())
+      testCoroutineDispatchers.runCurrent()
+      intended(hasComponent(RevisionCardActivity::class.java.name))
+      intended(hasExtra(SUBTOPIC_ID_EXTRA_KEY, SUBTOPIC_TOPIC_ID_1 - 1))
```
BenHenning: For this & the next test, suggest using the direct subtopic ID constant rather than using math (to make it more clear to the reader) _which_ card is being opened.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
+
+      onView(withId(R.id.next_navigation_card)).perform(click())
+      testCoroutineDispatchers.runCurrent()
+      intended(hasComponent(RevisionCardActivity::class.java.name))
+      intended(hasExtra(SUBTOPIC_ID_EXTRA_KEY, SUBTOPIC_TOPIC_ID_1 + 1))
+    }
+  }
+
+  @Test
+  fun testRevisionCardTestActivity_configurationChange_toolbarTitle_fractionSubtopicId1_isDisplayedCorrectly() { // ktlint-disable max-line-length
```
BenHenning, line 422: For this & next test: please rename so as to not need to disable the max-line-length ktlint check.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -279,79 +284,134 @@ class RevisionCardFragmentTest {
...
-      onView(withId(R.id.revision_card_explanation_text))
-        .check(matches(withText(containsString("Description of subtopic is here."))))
+      onView(withId(R.id.revision_card_toolbar_title))
+        .check(matches(withText("What is a Fraction?")))
     }
   }
 
   @Test
-  fun testRevisionCardTestActivity_configurationChange_fractionSubtopicId1_checkReturnToTopicButtonIsDisplayedSuccessfully() { // ktlint-disable max-line-length
+  fun testRevisionCardTestActivity_configurationChange_fractionSubtopicId2_checkExplanationAreDisplayedSuccessfully() { // ktlint-disable max-line-length
```
BenHenning, line 443:
```suggestion
  fun testRevisionCardTestActivity_configurationChange_fractionSubtopicId2_checkExplanationIsDisplayedSuccessfully() { // ktlint-disable max-line-length
```

Nit: wrong verb tense.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -54,8 +54,10 @@ const val TEST_TOPIC_ID_0 = "test_topic_id_0"
 const val TEST_TOPIC_ID_1 = "test_topic_id_1"
 const val TEST_TOPIC_ID_2 = "test_topic_id_2"
 const val FRACTIONS_TOPIC_ID = "GJ2rLXRKD5hw"
-const val SUBTOPIC_TOPIC_ID = 1
+const val SUBTOPIC_TOPIC_ID_0 = 0
+const val SUBTOPIC_TOPIC_ID_1 = 1
 const val SUBTOPIC_TOPIC_ID_2 = 2
+const val SUBTOPIC_TOPIC_ID_3 = 3
```
BenHenning: Are these specific to fractions? I suggest renaming them to be clearer on which topic they correspond to since they're currently quite general in their naming.

BenHenning (changes requested):
Thanks @JishnuGoyal! Some top-level thoughts:
- Could you include two brief videos in the PR description showing you playing through and clicking the next/previous buttons with and without TalkBack (i.e. one video for TalkBack, one for not)? I think seeing how this will actually behave will help better contextualize your screenshots.
- Why are the colors different between the tablet & non-tablet screenshots? Something seems wrong there.
- For RTL, it seems like the header text isn't being reversed to the other side of the layout--is the wrong style being used for that TextView?
- Please include screenshots for RevisionCardFragmentTest passing on Espresso since it's been changed.
- It would be nice, I think, to include at least one screenshot of the new UI with dark mode just to see how it looks. While we're not factoring in dark mode fully for each individual PR, there are some color changes here that I'm wondering about.
- The header looks a bit weird to me both for the "next button only" case, and on tablet in general. For the latter, it probably needs to be bigger. For the former, maybe a bigger header would help but I'm actually wondering if perhaps the button needs to be centered. I suggest emailing @mschanteltc and showing her the screenshots from this PR to get her thoughts on what we might be able to tweak to help with how it looks.
- CI checks are failing--PTAL.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: these tests are not a part of this PR, but there might be changes due to refactoring of a variable name. I suggest doing this in a separate issue. I will be reverting the refactoring though.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: these tests were not changed/created during this PR, though a variable name was refactored. I suggest getting this changed on a separate issue as I'm working on a limited time availability.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
JishnuGoyal: this is needed because the proposed plan was to color revision screens in orange so as to let the users know that they are a separate feature than the lessons screens.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
JishnuGoyal: continuing on the last review comment, it seems that changing colors is not possible below lollipop (see: https://stackoverflow.com/questions/33678266/changing-the-status-bar-color-on-android-kitkat)

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/main/res/values/strings.xml
JishnuGoyal: done

app/src/main/res/values/color_defs.xml
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
JishnuGoyal: done.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

mschanteltc:
Thank you Jishnu. I agree with Ben that when "only next is available," the current formatting looks a bit weird. Here are my suggestions for improvement: 

**All Layouts**
Create a gray background (#F2F2F2) for the "Continue Studying" section. 

**Mobile (Portrait)**
"Next only": Right-align the "Continue Studying" text and right-align its textbox with respect to its container. Right-align the Subtopic card (which is already done).

**Mobile (Landscape) and Tablet (Portrait, Landscape)**
Container width: Let's make the max container width that holds the heading and cards 480px. That way, elements in this section look more associated with each other. (Please correct me if I am using the correct terminology). I added links to the proposed designs and an image below. Let me know what you think!

Mobile, Portrait [[Prev and Next](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/e953d1cc-a0d6-4c25-890e-644af427befe/)] [[Next Only](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/493f3044-e6b4-4b75-bc2d-b8822f2121a3)]
Mobile, Landscape [[Prev and Next](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/aeced468-950a-4a26-abda-e25f0ff7896e)] [[Next Only](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/89f2160b-c9be-4c5d-8ce7-e82d60cff056)]
Tablet, Portrait [[Prev and Next](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/ff76509d-20e8-459c-ade7-b990f066e470)] [[Next Only](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/a92b7df9-f11e-46bd-8765-629742fc244e)]
Tablet, Landscape [[Prev and Next](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/f76e6dfa-8dd5-43a4-a33c-d7301b84b872)] [[Next Only](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/21ae7c35-4ac6-408c-bfc5-3f22ec8f51c6)]

TL Revision Card

TL Revision Card (Next Only)

JishnuGoyal:
Few parts that are left: 
1) Respond to Chantel's suggestions
2) Update the screenshots (mobile + tablet) with the latest screenshots
3) Add a video playing through the functionality introduced
4) Add todos in the code where the disableAccessibilityChecks annotation has been added
5) One test probably needs to be added to check if the header aligns itself to the right when only next revision card is available. 

PTAL @BenHenning, thanks.

JishnuGoyal:
Thank you so much @mschanteltc for your valuable suggestions. I find the new design ideas really nice!

@BenHenning I think the technical problem before we implement @mschanteltc's suggestions is that in the tablet mode, the width of the screen keeps changing depending upon the amount of content that is shown. The problem with navigation cards placement itself then boils down to the issue that a constant width for tablets isn't set. I think this problem deserves a separate issue. 

For this PR, i plan to fix the header alignment as suggested by Chantel, by aligning it to the right when only the next navigation card is visible.

Please let me know what are your thoughts @BenHenning @mschanteltc.

Thanks

app/src/main/java/org/oppia/android/app/topic/RouteToRevisionCardListener.kt
```diff
@@ -2,5 +2,10 @@ package org.oppia.android.app.topic
 
 /** Listener for when an [TopicActivity] should route to a [RevisionCardFragment]. */
 interface RouteToRevisionCardListener {
-  fun routeToRevisionCard(internalProfileId: Int, topicId: String, subtopicId: Int)
+  fun routeToRevisionCard(
+    internalProfileId: Int,
+    topicId: String,
+    subtopicId: Int,
+    subtopicListSize: Int = -1
```
BenHenning: We probably shouldn't be defaulting this since -1 is never a correct value to use. Is this default necessary?

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionFragmentPresenter.kt
```diff
@@ -55,11 +56,20 @@ class TopicRevisionFragmentPresenter @Inject constructor(
...
     return binding.root
   }
 
   override fun onTopicRevisionSummaryClicked(subtopic: Subtopic) {
-    routeToReviewListener.routeToRevisionCard(internalProfileId, topicId, subtopic.subtopicId)
+    routeToReviewListener.routeToRevisionCard(
+      internalProfileId,
+      topicId,
+      subtopic.subtopicId,
+      subtopicListSize
```
BenHenning: If we switch to nullable, we should make this a ``checkNotNull`` with an error message since the value must be defined for things to work correctly.

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionViewModel.kt
```diff
@@ -27,7 +27,7 @@ class TopicRevisionViewModel @Inject constructor(
 ) : ObservableViewModel() {
   private lateinit var profileId: ProfileId
   private lateinit var topicId: String
-  private val subtopicList: MutableList<TopicRevisionItemViewModel> = ArrayList()
+  val subtopicList: MutableList<TopicRevisionItemViewModel> = ArrayList()
```
BenHenning: Does the list need to be public? If so, does it need to be mutable?

If we don't need the mutable, you could hide this such as:

```kotlin
private val changeableSubtopicList = mutableListOf<TopicRevisionItemViewModel>()
val subtopicList: List<TopicRevisionItemViewModel>
  get() = changeableSubtopicList
```

That being said, do we actually even need to use a mutable list here? The subtopic list shouldn't ever actually be changing.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
```diff
@@ -62,30 +69,39 @@ class RevisionCardActivity :
...
+    internal const val SUBTOPIC_LIST_SIZE_EXTRA_KEY = "RevisionCardActivity.subtopic_list_size"
 
     /** Returns a new [Intent] to route to [RevisionCardActivity]. */
     fun createRevisionCardActivityIntent(
       context: Context,
       internalProfileId: Int,
       topicId: String,
-      subtopicId: Int
+      subtopicId: Int,
+      subtopicListSize: Int = -1
```
BenHenning: Ditto here--prefer to not default it.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -283,95 +295,102 @@ class RevisionCardFragmentTest {
     }
   }
 
+  @DisableAccessibilityChecks
```
BenHenning, line 318: Why is this needed? Generally, we prefer not to disable accessibility checks.

Ditto for the others.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
```diff
@@ -32,8 +33,48 @@ class RevisionCardViewModel @Inject constructor(
...
+    Transformations.map(topicLiveData, ::processNextSubtopicData)
+  }
+
+  val previousSubtopicLiveData: LiveData<EphemeralSubtopic> by lazy {
+    Transformations.map(topicLiveData, ::processPreviousSubtopicData)
+  }
+
+  private fun processPreviousSubtopicData(
+    topicLiveData: AsyncResult<EphemeralTopic>
+  ): EphemeralSubtopic? {
```
BenHenning: For this & the next: do they ever return null? It doesn't seem like it.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -79,6 +86,60 @@ class RevisionCardFragmentPresenter @Inject constructor(
     return binding.root
   }
 
+  private fun setUpRevisionNavigationCards(
+    binding: RevisionCardFragmentBinding,
+    topicId: String,
+    subtopicId: Int,
+    subtopicListSize: Int
+  ) {
+
```
BenHenning: Nit: remove extra newline for style consistency.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -79,6 +86,60 @@ class RevisionCardFragmentPresenter @Inject constructor(
...
+    binding.prevSubtopicImageView.setEntityType(entityType)
+    binding.nextSubtopicImageView.setEntityType(entityType)
+
+    if (subtopicId == 1) {
+      binding.previousNavigationCard.visibility = View.INVISIBLE
+    } else {
+      getReviewCardViewModel().previousSubtopicLiveData.observe(fragment) { previousSubtopic ->
```
BenHenning: Why can't these changes be done via databinding? I'd imagine the image view would be set something like:

```xml
app:lessonThumbnail="@{viewModel.previousSubtopic.subtopic.subtopicThumbnail}"
```

And the text:

```xml
app:text="@{viewModel.computeSubtopicTitleText()}"
```

Where ``computeSubtopicTitleText`` does the same translation extra string as below. We do this in other view models as well.

This allows you to manage everything through databinding, and may somewhat minimize code duplication between the previous/next subtopic cards (depending on how you implement it).

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -79,6 +86,60 @@ class RevisionCardFragmentPresenter @Inject constructor(
...
 
+  private fun setUpRevisionNavigationCards(
+    binding: RevisionCardFragmentBinding,
+    topicId: String,
+    subtopicId: Int,
+    subtopicListSize: Int
+  ) {
+
+    binding.prevSubtopicImageView.setEntityType(entityType)
+    binding.nextSubtopicImageView.setEntityType(entityType)
```
BenHenning: These can be set via databinding.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -79,6 +86,60 @@ class RevisionCardFragmentPresenter @Inject constructor(
...
+
+    if (subtopicId == 1) {
+      binding.previousNavigationCard.visibility = View.INVISIBLE
```
BenHenning: Suggest instead making this a bit more explicit, such as:

```kotlin
binding.previousNavigationCard.visibility = if (subtopicId == 1) View.INVISIBLE else View.VISIBLE
```

This makes it specifically clear when the card is visible vs. not visible.

That being said, why are we using INVISIBLE instead of GONE? If these will never be visible in this situation, it doesn't seem to make sense to pay the extra measuring/layout cost since the view shouldn't be showing up.

Ditto for the next card below.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
BenHenning: In that case, you need to move that bit of code to a method that uses ``@TargetApi``, and then conditionally call it using an if-check that checks the current build version. That way the app can still run on KitKat--it'll just bypass this extra logic.

That being said, what _will_ the behavior be on KitKat if we can't run the status bar color code? Also, do we need to change the status bar color? I feel like we should be doing whatever we do on other screens for this rather than deviating.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
BenHenning: I think this may still be unresolved.

BenHenning (changes requested):
Thanks @JishnuGoyal. Took another pass.

Please make sure all remaining items from your list are addressed before sending this back, in addition to the comments I've left. I agree that it's reasonable not to fix the layout spacing issues for the cards since it's an existing problem, but please do file a new issue and add TODOs in code to fix that.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
BenHenning: But we don't do that for concept cards (which also use orange)--do we actually need it?

app/src/main/java/org/oppia/android/app/topic/RouteToRevisionCardListener.kt
JishnuGoyal: removed

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionViewModel.kt
JishnuGoyal: rolled back these changes - now this is private

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionFragmentPresenter.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: This is needed because work on urlspan (#4235) is still under progress. Adding a new issue for this: #4631

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -283,95 +295,107 @@ class RevisionCardFragmentTest {
...
-
       onView(withId(R.id.previous_navigation_card)).perform(click())
       testCoroutineDispatchers.runCurrent()
       intended(hasComponent(RevisionCardActivity::class.java.name))
-      intended(hasExtra(SUBTOPIC_ID_EXTRA_KEY, SUBTOPIC_TOPIC_ID_1 - 1))
+      intended(hasExtra(SUBTOPIC_ID_EXTRA_KEY, FRACTIONS_SUBTOPIC_TOPIC_ID_0))
     }
   }
 
+//  TODO(#4631): Remove this once #4235 is resolved.
```
BenHenning: Please fix spacing for this comment. Ditto elsewhere in this test.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
```diff
@@ -133,6 +132,12 @@ import org.robolectric.annotation.LooperMode
 import javax.inject.Inject
 import javax.inject.Singleton
 
+const val FRACTIONS_SUBTOPIC_TOPIC_ID_0 = 1
+const val FRACTIONS_SUBTOPIC_TOPIC_ID_1 = 2
+const val FRACTIONS_SUBTOPIC_TOPIC_ID_2 = 3
+const val FRACTIONS_SUBTOPIC_TOPIC_ID_3 = 4
+const val FRACTIONS_SUBTOPIC_LIST_SIZE = 4
```
BenHenning: Can these be private?

app/src/main/res/values/themes.xml
```diff
@@ -18,6 +18,14 @@
     <item name="android:windowAnimationStyle">@style/Animation.Activity</item>
   </style>
 
+  <style name="RevisionCardActivityTheme" parent="Theme.MaterialComponents.Light.NoActionBar.Bridge">
```
BenHenning, line 21: Shouldn't this theme inherit from the one that the activity was using before (OppiaThemeWithoutActionBar)?

app/src/main/res/values/themes.xml
```diff
@@ -18,6 +18,14 @@
...
+    <item name="colorPrimaryDark">@color/component_color_concept_card_fragment_status_bar_color</item>
+    <item name="windowActionBar">false</item>
+    <item name="windowNoTitle">true</item>
+    <item name="android:editTextStyle">@style/InputInteractionEditText</item>
+    <item name="android:windowAnimationStyle">@style/Animation.Activity</item>
```
BenHenning: Why are these needed?

app/src/main/res/layout/revision_card_fragment.xml
```diff
@@ -28,9 +28,12 @@
         style="@style/Body"
         android:layout_marginStart="@dimen/revision_card_fragment_layout_margin_start"
         android:layout_marginTop="@dimen/revision_card_fragment_layout_text_margin_top"
-        android:layout_marginEnd="@dimen/revision_card_fragment_layout_margin_end" />
+        android:layout_marginEnd="@dimen/revision_card_fragment_layout_margin_end"
+        android:minWidth="48dp"
+        android:minHeight="48dp" />
 
       <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/navigation_card_container"
```
BenHenning: Given that IDs are global, could we make this more specific?

app/src/main/res/layout/revision_card_activity.xml
```diff
@@ -22,7 +22,7 @@
         android:id="@+id/revision_card_toolbar"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:background="@color/component_color_revision_card_activity_toolbar_color"
+        android:background="@color/component_color_concept_card_fragment_toolbar_color"
```
BenHenning, line 25: We should change component_color_revision_card_activity_toolbar_color instead of using concept card's here (since we're effectively changing the revision title color in this PR).

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
```diff
@@ -114,4 +129,31 @@ class RevisionCardViewModel @Inject constructor(
...
   }
+
+  class Factory @Inject constructor(
+    private val activity: AppCompatActivity,
+    private val topicController: TopicController,
+    private val oppiaLogger: OppiaLogger,
+    @TopicHtmlParserEntityType private val entityType: String,
+    private val translationController: TranslationController
+  ) {
+    fun create(
```
BenHenning, line 163: This & all methods above, and for the Factory class: we need KDocs here.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
```diff
@@ -53,76 +48,40 @@ class RevisionCardFragmentPresenter @Inject constructor(
...
-    viewModel.initialize(topicId, subtopicId, profileId)
     logRevisionCardEvent(topicId, subtopicId)
 
     binding.let {
       it.viewModel = viewModel
       it.lifecycleOwner = fragment
     }
 
-    setupRevisionNavigationCards(binding, topicId, subtopicId)
+//    setUpRevisionNavigationCards(binding, topicId, subtopicId, subtopicListSize)
```
BenHenning: Remove.

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionFragmentPresenter.kt
```diff
@@ -55,11 +57,22 @@ class TopicRevisionFragmentPresenter @Inject constructor(
...
 
   override fun onTopicRevisionSummaryClicked(subtopic: Subtopic) {
-    routeToReviewListener.routeToRevisionCard(internalProfileId, topicId, subtopic.subtopicId)
+    routeToReviewListener.routeToRevisionCard(
+      internalProfileId,
+      topicId,
+      subtopic.subtopicId,
+      checkNotNull(subtopicListSize) {
+        throw IllegalArgumentException("Subtopic list size not found.")
+      }
```
BenHenning:
```suggestion
      checkNotNull(subtopicListSize) { "Subtopic list size not found." }
```

No need for the extra throw.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
BenHenning: Still partially unresolved, I think--can we remove the target annotation now?

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt
BenHenning: This has been resolved.

BenHenning (commented):
Thanks @JishnuGoyal. Mostly LGTM. Just had a few nits left. Please also wait for CI to finish & fix all failures before sending back.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentTest.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardFragmentPresenter.kt
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revision/TopicRevisionFragmentPresenter.kt
JishnuGoyal: done

JishnuGoyal: done

app/src/main/res/layout/revision_card_activity.xml
JishnuGoyal: done

app/src/main/res/values/themes.xml
JishnuGoyal: I notice doing this makes status bar color green instead of orange, even after overriding colorPrimaryDark

app/src/main/res/values/themes.xml
JishnuGoyal: removed

app/src/main/res/layout/revision_card_fragment.xml
JishnuGoyal: done

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
JishnuGoyal: done -- PTAL.

JishnuGoyal:
PTAL @BenHenning

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
BenHenning: All other public fields & methods need KDocs, too.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
```diff
@@ -130,13 +130,15 @@ class RevisionCardViewModel private constructor(
     }
   }
 
+  /** Factory method to initialize the fields of [RevisionCardViewModel]. */
```
BenHenning:
```suggestion
  /** Factory to create new [RevisionCardViewModel]s. */
```

app/src/main/res/values/themes.xml
BenHenning: Hmm odd. I'm not keen on the disparity here, but debugging these styles can be really time consuming so I think we can live with this.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
```diff
@@ -134,6 +134,7 @@ class RevisionCardActivityTest {
   lateinit var monitorFactory: DataProviderTestMonitor.Factory
 
   private val profileId = ProfileId.newBuilder().apply { internalId = 1 }.build()
+  private val fractionsSubtopicListSize = 4
```
BenHenning: Suggest making a constant at the top for the test (in the same way as RevisionCardFragmentTest).

BenHenning (commented):
Thanks @JishnuGoyal. Just a few small things left (I think one of my past KDoc comments was a bit unclear, but I've clarified it now).

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
BenHenning: To clarify, please document all non-private members and all properties in the view model's constructor.

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
JishnuGoyal: done

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
JishnuGoyal: done

JishnuGoyal:
PTAL @BenHenning

app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardViewModel.kt
BenHenning: Please make sure to read these comments, fully. You didn't document the properties or the fields, only the methods.

BenHenning (commented):
@JishnuGoyal I can't approve this PR yet because one of the comments wasn't fully addressed. PTAL.

BenHenning (approved):
Thanks @JishnuGoyal! PR LGTM.

Issue #4562, opened 2022-09-07
Add comprehensive internationalization and localization validation checks


BenHenning:
TODO: fill in this description from my original WIP TDD.

> (From #4992) We've had past strings break due to having incorrect format specifiers and some HTML formatting problems (though Translatewiki mostly covers those problems). We still need much more robust checks for strings--these haven't been implemented yet.

-----

Notes from @seanlip: We should implement at least the checks that we do on Web. We have found breaking changes due to incorrectly-formatted translatewiki strings that result in last-minute release delays (see e.g. #5034) so this issue is bumped to high priority and is now being treated as a bug. Here are the checks that we currently do on Web (in some case, an analogous check is listed below since the translation file formats are different):

- [ ] There is at least one key in the English resource file (app/src/main/res/values/strings.xml).
- [ ] All resources are correctly formatted (in terms of the XML tag structure).
- [ ] The keys in the qq file (see #3660) exactly match those in the English file (nothing is missing)
- For all other resource files in app/src/main/res/values-* (besides the qq file):
   - [ ] **[Most important -- tackle this first]** HTML tags in translations are correctly preserved. 
       - Rough algorithm: look for escaped HTML tags (since direct HTML and CDATA are both now prohibited); remove anything apart from `<...>`; compare both sets of tags. See https://github.com/oppia/oppia/blob/develop/core/controllers/base_test.py#L1045 for reference. Note that we could make this stronger by requiring that e.g. ol/li tags appear in the same order as the original, though we might not want to specify that the full set of tags has the same order due to differences in word order between different languages.
   - [ ] Any `%` strings in the English resource file (e.g. `%1$s`, `%s`) are correctly preserved.
   - [ ] The keys are a subset of the keys in the English resource file.
   - ~~All keys in each file are sorted alphabetically (not strictly needed, but helps with eyeballing and manual investigation of issues)~~ -- EDIT: this might actually not be desired, see https://github.com/oppia/oppia-android/pull/4823/files. Removing the checkbox associated with it for now.

PR #4563, opened 2022-09-07
Fix part of #4562: Add string consistency tool & CI check

BenHenning:
## Explanation
Fixes part of #4562.

This PR introduces two new scripts:
- One for computing how many strings are not yet translated for Arabic, Brazilian Portuguese, and Swahili (as compared to the base English strings)
- One for verifying newline consistency between the English & non-English strings (based on number of lines)

The latter was also added as one of the jobs to be run during the static checks CI workflow, and is specifically useful to ensure newlines weren't inadvertently added by translators (see the updated translation strings for an idea on how often this has happened). Note that, as part of fixing these strings, the source string for lets_get_started was updated to no longer include a newline. This seems reasonable given that we should never use newlines for spacing/styling unless it's for logical splits (like paragraphs). However, I can't verify whether actual style changes are needed since this string is used as part of the unreleased app walkthrough feature. I expect that we'll re-audit the UI of that feature when we revisit it.

This PR also includes a major performance fix for RepositoryFile that will benefit all checks that use it. Kotlin's built-in file tree walker is built to implicitly follow symlinks on the filesystem. For Bazel builds, this expands the codebase to >1M files (since directory filtering happens *after* the files are known). Since scripts don't actually need to follow symlinks, the utility was updated to specifically not follow them (using Java NIO's file tree walk routine) which led to an observed ~10x performance improvement. Further optimizations could be done by building our own tree-walker and filtering directories during search (which will probably be necessary if we ever _do_ need to follow symlinks, but it probably won't be needed for a long time since the current performance is quite solid).

The former script is useful when manually auditing the translation progress using the codebase as the source of truth rather than Translatewiki. For now, it's been helpful in beta MR1 planning but longer-term it may also be useful as a badge on the repo's README.

Finally, the Kotlin stdlib dependency was updated to point to jdk8 instead of jdk7 (since a specific feature was needed from the former, and the codebase can rely on higher Java SDK versions than 7 given that the minimum Java version is 1.8).

## Essential Checklist

## For UI-specific PRs only
N/A -- this is a developer infrastructure-only PR, except for all of the string newline fixes (which just result in a more correct UI for non-English), and the changed English string (which is inaccessible to demonstrate). Screenshots could be uploaded, but the newlines need to be removed regardless so it doesn't seem specifically useful here (plus those languages may not be obvious to verify for reviewers).


BenHenning:
This should be ready to merge once #4417 is merged and this PR brought up-to-date with develop (I self reviewed it and didn't find anything to change).

BenHenning:
Since this PR is part of broader urgent work for the Beta MR1 release, I'm force-merging this without review. I've self-reviewed it and found no major issues. Furthermore, #4567 is tracking ensuring that this does get reviewed by someone else later after it's been merged.

PR #4629, opened 2022-09-28
Fix #1801: Audio player running after closing 

### LINKED ISSUE #1801: Audio player running after closing
**Describe the bug**
Alpha release apk

**To Reproduce**
1. Start an exploration
2. Play audio - listen completely
3. Answer correct - continue
4. Next state
5. Click on audio icon in toolbar - this close the player - hide the audio player from screen
6. You can hear the voiceover now also

[Video File](https://drive.google.com/file/d/1hOncsuKvQwQyA0dCyuoVj4P6H60wJoqg/view?usp=sharing) describing the issue.

**Expected behavior**
After closing the audio player, player should not run

**Screenshots**
If applicable, add screenshots or videos to help explain your problem.

**Device**
- Redmi note 5 pro

**Additional context**
Note that a related issue is #2893 which is caused by the same synchronization issues in the audio player.

KevinGitonga:
## Explanation

Fix #1801
This fix adds check in the implemented playStatusLiveData to complete audioPause request if any is pending while the playerStatus keeps changing. When user clicks on pause audio icon and the UI disappears there is a delay in audio actually pausing. Adding a check in playStatusLiveData completes pause pending request fast enough to prevent it from playing.

## Essential Checklist

## For UI-specific PRs only

## Before changes demo
https://user-images.githubusercontent.com/20886444/192735111-6baf46f8-3531-49f8-83d6-98277dc74c6d.mp4

## After changes demo
https://user-images.githubusercontent.com/20886444/192735968-13b7e2b5-e296-4978-9194-4df5b807897b.mp4


app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt
```diff
@@ -266,6 +272,7 @@ class AudioFragmentPresenter @Inject constructor(
   }
 
   private fun hideAudioFragment() {
+    isPauseAudioRequestPending = true
```
BenHenning: I'm not sure that I understand the flow fully. The ``(fragment as AudioUiManager).pauseAudio()`` line below seems to just call through to ``pauseAudio`` above, so it's a bit unclear to me why we only set this to true here but not in other cases when a pause seems to be requested (such as when 'pauseAudio' itself is clicked). Am I maybe missing something?

BenHenning (commented):
Thanks @KevinGitonga. I can see this conceptually working (though per the videos it seems like it does functionally work), but I think some clarity is needed into why this particular approach works since it doesn't seem to exactly be "throttle control" for pausing per my comment.

Once we finalize the solution, I think you can go ahead with writing tests. That being said, until then you should consider marking this PR as 'Draft' since it isn't technically complete yet (and thus not ready for a full review).

app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt
KevinGitonga: @BenHenning updating the status of "isPauseAudioRequestPending" inside pauseAudio works the same. What i wanted was to update this early enough when the process of hiding the audioPlayer Fragment has started this is on "hideAudioFragment". This seems to fix the issue according to what i found out.

app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt
```diff
@@ -101,6 +102,9 @@ class AudioFragmentPresenter @Inject constructor(
       Observer {
         prepared = it != UiAudioPlayStatus.LOADING && it != UiAudioPlayStatus.FAILED
         binding.audioProgressSeekBar.isEnabled = prepared
+        if (prepared && isPauseAudioRequestPending) {
+          pauseAudio()
+        }
```
BenHenning, line 122: Per the other conversation thread, is this the part that actually fixes it? I'm actually a bit confused here--when would this ever be called such that pauseAudio() doesn't correctly pause things? I think it'd be helpful to see a flow diagram of what's actually happening during the bug to better understand why this works.

KevinGitonga: Actually after some more checking out i figured out that, when user is proceeding to next section by clicking on continue. And clicks on pause very fast. Pause will not occur for audio since it is yet to be prepared so on "pauseAudio" the check for prepared is false hence audio continues to play since it is not paused in viewModel. Calling pauseAudio in audioStatusObserver and adding check if we have pending pause request enables the pending request to execute once the player is prepared. Will share more details on flow for this as i improve on the fix.

KevinGitonga: On this, the root cause is when the user clicks on pause button quickly after continuing to new topic, the audio player is never prepared so it won't be actually paused. The solution is as stated checking on the Audio player lifecycle observer method and calling pause once the status changes to prepared and we have a pending pauseAudioRequest pauseAudio will be called to serve this request and audio will not continue playing. Audio is not prepared because if an else check is added on pauseAudio and logs added here this is invoked. Since it is clear what is the cause i do not see reason to demo this in flow chart. The reason for calling pause from observer is because if we force audioPlayer pause if not prepared it causes an exception therefore we have to wait until it is prepared which can be done on audioPlayerStatus observer method.

KevinGitonga: The reason for "isPauseAudioRequestPending" is to track if we have a pause request. When a user clicks pause fast after continuing to the next topic pauseAudio will be called will not be send to viewModel since the player is yet to be prepared. When the status changes in observer and we have a pending pause request it will be serviced directly from here and the audioPlayer should not proceed with playing audio while the UI has already been minimized. @BenHenning i hope this comments provide enough context on the issue.

BenHenning: I see. I wonder if this can be simplified a bit, but I actually think the first thing to do is to implement tests that properly fail without the fix, and pass with the fix, to demonstrate that it works as expected. We can then see if any parts can be simplified further.

I actually think this approach is mainly a stopgap to work around a design flaw in the AudioController (it doesn't actually manage 'requests' quite in the way you described, so we're essentially dealing with a data race here). I think this is a reasonable stopgap until we rewrite the AudioController implementation (which we're planning to do in the future).

KevinGitonga: Not quite sure if this is a data race situation, it might be because even if user waits a little bit before pressing pause the audio will continue playing. I will look at how to write tests for this.

KevinGitonga:
Hi @BenHenning yet to assert this in tests as it seems complex on grounds that the AudioPlayer is restarted when the user clicks on continue button to proceed to the next lesson in the Exploration, Also not able to find a suitable way to invoke ShadowMediaPlayer for this context. The cause of Audio Playing after the user clicks the pause button while playing on a given exploration might be caused by the below issue [1]. 

[1]. After observing my logs data the audio player continues to play since the audio player is not in a “prepared” state. This state is checked as noted [here](https://github.com/oppia/oppia-android/blob/5c6af2c2d551545289c93c34098a8de435281cb4/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L228) before actually performing a pause through the audioPlayerController in AudioViewModel [here](https://github.com/oppia/oppia-android/blob/091b45a1885562752f6d25da8e5494e4a331bdb4/app/src/main/java/org/oppia/android/app/player/audio/AudioViewModel.kt#L144). The pause will be skipped as the player is not in a “prepared” state hence why audio will continue to play once the state changes to prepared [here](https://github.com/oppia/oppia-android/blob/5c6af2c2d551545289c93c34098a8de435281cb4/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L107) which is the expected behavior. This is despite the player ui fragment being hidden.

The solution which works for this issue is to insert an if statement to check if a previous pause request exists as the state changes inside of [this](https://github.com/KevinGitonga/oppia-android/blob/d67eca9215e7af37c54d6345ce18252c9a2e0cae/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L114) observer directly and if at the moment we have a pending request which was not executed when pauseAudio was initially invoked [here](https://github.com/KevinGitonga/oppia-android/blob/d67eca9215e7af37c54d6345ce18252c9a2e0cae/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L234). This will work if only the state is “prepared” and we have a “pending” request to pause the audio player. Kindly PTAL and advice your thoughts on this.

BenHenning:
Thanks for the write-up @KevinGitonga. As discussed during the meeting earlier, I think the next steps here are carefully documenting the solution & why the code works/shouldn't change if we can't actually verify this in a test.

KevinGitonga:
Thanks @BenHenning 

After observing my logs data the audio player continues to play since the audio player is not in a “prepared” state. This state is checked as noted [here](https://github.com/oppia/oppia-android/blob/5c6af2c2d551545289c93c34098a8de435281cb4/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L228) before actually performing a pause through the audioPlayerController in AudioViewModel [here](https://github.com/oppia/oppia-android/blob/091b45a1885562752f6d25da8e5494e4a331bdb4/app/src/main/java/org/oppia/android/app/player/audio/AudioViewModel.kt#L144). The pause will be skipped as the player is not in a “prepared” state hence why audio will continue to play once the state changes to prepared [here](https://github.com/oppia/oppia-android/blob/5c6af2c2d551545289c93c34098a8de435281cb4/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L107) which is the expected behavior. This is despite the player view fragment being hidden.

The **solution** which works for this issue is to insert an if statement to check if a previous pause request exists as the state changes inside of [this](https://github.com/KevinGitonga/oppia-android/blob/d67eca9215e7af37c54d6345ce18252c9a2e0cae/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L114) observer directly and if at the moment we have a pending request which was not executed when pauseAudio was initially invoked [here](https://github.com/KevinGitonga/oppia-android/blob/d67eca9215e7af37c54d6345ce18252c9a2e0cae/app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt#L234). This will work if only the state is “prepared” and we have a “pending” request to pause the audio player.

The main reasons why this fix cannot be checked through tests in platforms available is because of system constraint's in testing audio. I could also not come up with a suitable way to test this after researching how to undertake this. For this scenario we would require to listen the state of audio or check the state of the player buttons which would not be applicable since the audio player view itself has already been hidden from the UI as part of the pause process.

I have added comments and links to this comments and the issue itself in the code to assist other developers understand the reasons for the checks.

app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt
```diff
@@ -109,6 +111,16 @@ class AudioFragmentPresenter @Inject constructor(
       Observer {
         prepared = it != UiAudioPlayStatus.LOADING && it != UiAudioPlayStatus.FAILED
         binding.audioProgressSeekBar.isEnabled = prepared
+
+        // This check will execute any pending pause request that causes
+        // issue with audio not being paused as user navigates through
+        // lessons in a topic check below issue for more details
+        // https://github.com/oppia/oppia-android/issues/1801
+        // Link to notes on why this fix works can be found on link below
+        // https://github.com/oppia/oppia-android/pull/4629#issuecomment-1410005186
```
BenHenning:
```suggestion
        // This check will execute any pending pause request that causes
        // issues with audio not being paused as the user navigates through
        // lessons in a topic. Check #1801 for more details, and specifically
        // https://github.com/oppia/oppia-android/pull/4629#issuecomment-1410005186
        // for notes on why this fix works can be found on link below.
```

Some refinement. Also, please make sure this is line-wrapped to 100 characters (it seems like it might be able to be a bit more compact at a glance).

BenHenning (commented):
Thanks @KevinGitonga! Left one follow-up comment, otherwise the PR LGTM. PTAL.

app/src/main/java/org/oppia/android/app/player/audio/AudioFragmentPresenter.kt
BenHenning: This seems resolved in the latest changes.

BenHenning (approved):
Thanks @KevinGitonga. I think you missed responding to the comment thread, but otherwise the PR LGTM.

PR #4636, opened 2022-10-02
Fixes #3169: Item selection interaction text should change on selection

### LINKED ISSUE #3169: ItemSelectionInteraction text change on selection
**EDIT** (by @seanlip): See updated description here -- https://github.com/oppia/oppia-android/issues/3169#issuecomment-1225305069

----------------------

**Previous description**

[Mock1](https://xd.adobe.com/view/dfe8ccd4-39eb-414d-a0b2-b5674357f966-8abb/screen/722a0d0b-737d-4f07-95aa-0cfa7db06530/)

As per Mock1, by default the text is `Please select atleast one choice.`. If atleast one option is selected we hide this text and if the option is deselected then we show this text again.

Add this functionality and its relevant test.

## Expected Cases:

1. No item selected - Default Text - `Please select one or more choices.`
2. If total selection count is atleast 1 and less than allowable selections, change text to `You may select more choices.`
3. If total selection count is equal to allowable selections, change text to `No more than 4 choices may be selected.`

You can experiment similar cases on [What is a Fraction?](https://www.oppia.org/explore/umPkwp0L1M0-) . You will need to answer a lot of states before you will reach a state where you will find this interaction where you can select multiple options.

Akshatkamboj14:
## Explanation
Fixes #3169

## Essential Checklist

## For UI-specific PRs only


Akshatkamboj14:
Hey @BenHenning as you suggested on the Chat I have added the list of observables boolean and I have initialized it also now I need to know how to disable it or enable it. Can you help me more with this?

BenHenning:
Sorry, will need to look at this tomorrow.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -56,6 +57,8 @@ class SelectionInteractionViewModel private constructor(
     computeChoiceItems(choiceSubtitledHtmls, hasConversationView, this)
 
   private val isAnswerAvailable = ObservableField(false)
+  val selectedItemText = ObservableField("Please select all correct choices")
+  val selectedItemsList: List<ObservableBoolean>? = null
```
BenHenning: So what we want to model here isn't selected items, but rather selectableItems. This view model should set all checkboxes as enabled by default (true values), and then disable all unselected items once the max number is selected (setting just those ones to false).

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
```diff
@@ -20,5 +20,8 @@ class SelectionInteractionContentViewModel(
     if (isCurrentlySelected != shouldNowBeSelected) {
       isAnswerSelected.set(shouldNowBeSelected)
     }
+    selectionInteractionViewModel.selectedItemsList?.forEach {
```
BenHenning: We need to pass in the observable boolean corresponding to this specific checkbox from SelectionInteractionContentViewModel's constructor.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
BenHenning: This also needs to be initialized to be a list of observable booleans for each possible choice, e.g. something like:

```kotlin
val selectedItemsList by lazy { List(choiceItems.size) { ObservableBoolean() } }
```

BenHenning (commented):
I think this is on the right track @Akshatkamboj14. I've left some follow-up suggestions to help guide the solution.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
Akshatkamboj14: @BenHenning, How can we do that here can you show me some example?

BenHenning: I'm not sure if there is an exact example @Akshatkamboj14. In this case, I think you need to think through the logic of how this works and try to come up with something that makes sense to you. I don't think this is a case were can just copy and paste from another location in code.

Akshatkamboj14: Okay, @BenHenning I'll think of something.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
```diff
@@ -21,4 +21,21 @@ class SelectionInteractionContentViewModel(
       isAnswerSelected.set(shouldNowBeSelected)
     }
   }
+
```
BenHenning: You should pass in the ObservableBoolean corresponding to this specific checkbox via the view model’s constructor, then you can bind directly to that observable and don’t need this function at all.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -148,6 +153,18 @@ class SelectionInteractionViewModel private constructor(
     }
   }
 
+  private fun updateSelectionText() {
```
BenHenning, line 164: This function also needs to update the enabled states of each checkbox based on whether more options can be selected (it needs to disable unselected checkboxes when the limit is reached, and enable all checkboxes if the limit is not reached).

BenHenning (commented):
Thanks @Akshatkamboj14. I think this is on the right track, but it requires some changes. PTAL at my comments.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
Akshatkamboj14: hey @BenHenning PTAL, I have tested my code in the suggested prototype exploration, and its working.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: Okay

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -148,6 +153,18 @@ class SelectionInteractionViewModel private constructor(
...
+    if (selectedItems.size < maxAllowableSelectionCount) {
+      selectedItemText.set("You may select more choices")
```
BenHenning: This & the others should be strings that are loaded from the resources file, rather than hardcoded (so that they may be translated).

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -45,17 +46,19 @@ class SelectionInteractionViewModel private constructor(
...
       ?: minAllowableSelectionCount
   }
-  private val selectedItems: MutableList<Int> = mutableListOf()
+  val selectedItems: MutableList<Int> = mutableListOf()
   val choiceItems: ObservableList<SelectionInteractionContentViewModel> =
     computeChoiceItems(choiceSubtitledHtmls, hasConversationView, this)
 
   private val isAnswerAvailable = ObservableField(false)
+  val selectedItemText = ObservableField("Please select all correct choices")
+  val enabledItemsList by lazy { List(choiceItems.size) { ObservableBoolean() } }
```
BenHenning: Shouldn't these be default enabled, e.g. by initializing with ``ObservableBoolean(true)``?

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -45,17 +46,19 @@ class SelectionInteractionViewModel private constructor(
...
     interaction.customizationArgsMap["maxAllowableSelectionCount"]?.signedInt
       ?: minAllowableSelectionCount
   }
-  private val selectedItems: MutableList<Int> = mutableListOf()
+  val selectedItems: MutableList<Int> = mutableListOf()
   val choiceItems: ObservableList<SelectionInteractionContentViewModel> =
     computeChoiceItems(choiceSubtitledHtmls, hasConversationView, this)
 
   private val isAnswerAvailable = ObservableField(false)
+  val selectedItemText = ObservableField("Please select all correct choices")
```
BenHenning: This should use a string ID rather than a hardcoded, in-text string.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
BenHenning: Thanks for checking @Akshatkamboj14. You should still change the model to take in the observable boolean via its constructor, though, as that better follows the conventions used in other view models and creates a better separation of concerns.

BenHenning (changes requested):
Thanks @Akshatkamboj14. Besides the latest comments, you should also add tests to verify that:
- Each of the strings shows up at the correct, expected time
- The checkboxes are enabled/disabled when expected

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: Okay @BenHenning

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: Okay @BenHenning but I am facing issues in the 3 more where the `maxAllowableSelectionCount ` is used, I am not getting how we can use this in `strings.xml`.

BenHenning: You can pass variables into strings by adding a '%s' to the string, and then calling a format method to pass the value into the string. For a reference, see:

https://github.com/oppia/oppia-android/blob/798f08b881f1cc84556d0c30dbb6a1b79b4bbb75/app/src/main/res/values/strings.xml#L17

and

https://github.com/oppia/oppia-android/blob/091b45a1885562752f6d25da8e5494e4a331bdb4/app/src/main/java/org/oppia/android/app/player/audio/AudioViewModel.kt#L149

Akshatkamboj14: @BenHenning I am getting an error in `resourceHandler` and I have also tried to google it also it's saying that I have to take context but that is also not working in my case.

BenHenning: @Akshatkamboj14 I need more information to be able to help:
- What specific error did you run into?
- What code is causing it?
- What have you tried when attempting to fix it?
- What do you think the problem might be?
- How would you proceed next to solve it?

Akshatkamboj14: @BenHenning in the `getString` method I guess I have to get the `context` reference to call the `getString` method and I have tried `this.getString()`, `writtenTranslationContext.getString`, `context.getString()`, `Context.getString()`  and [https://github.com/oppia/oppia-android/pull/4636#discussion_r1007630463](url) the one you have mentioned in the above comments, but they all are not working and I am getting this error,

image

BenHenning: @Akshatkamboj14 have you looked at other view models to see how they are translated? You should also provide more specifics than "doesn't work" as that doesn't provide much context--is it always a symbol error?

Akshatkamboj14: Hey @BenHenning I have looked and most of them are using `resourceHandler` as their context but in the `SelectionInteractionViewModel` there is no parameter of `resourceHandler `.

image

image

image

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: Okay Ben

BenHenning: I think these are false at the moment (ObservableBoolean is false by default)--is that correct?

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
BenHenning: This doesn't seem done yet--am I missing something?

app/src/main/res/values/strings.xml
```diff
@@ -574,4 +574,7 @@
   <string name="administrator_controls_fragment_test_activity_label">Administrator Controls Fragment Test Activity</string>
   <string name="select_a_topic_to_start">Select a Topic to Start</string>
   <string name="revision_navigation_cards_header">Continue Studying</string>
+  <string name="please_select_all_correct_choices">Please select all correct choices</string>
+  <string name="you_may_select_more_choices">You may select more choices</string>
+  <string name="no_more_than_choices_may_be_selected">No more than %s choices may be selected.</string>
```
BenHenning: Are these exact copies from Oppia web? The last one has punctuation but the first two don't which seems like an inconsistency.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
BenHenning: This still seems unresolved, I think.

BenHenning (commented):
Thanks @Akshatkamboj14. I think some of my existing comments weren't replied to, or haven't fully been addressed yet. Did you mean to reply? PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
Akshatkamboj14: done.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: done.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: done.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
```diff
@@ -45,17 +49,28 @@ class SelectionInteractionViewModel private constructor(
...
+  val selectedItems: MutableList<Int> = mutableListOf()
+  private val enabledItemsList by lazy {
+    List(
+      choiceSubtitledHtmls.size
+    ) {
+      ObservableBoolean(
+        true
+      )
+    }
+  }
```
BenHenning:
```suggestion
  private val enabledItemsList by lazy {
    List(choiceSubtitledHtmls.size) {
      ObservableBoolean(true)
    }
  }
```

Can be condensed.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
```diff
@@ -20,5 +21,13 @@ class SelectionInteractionContentViewModel(
...
     }
+    if (
+      selectionInteractionViewModel.selectedItems.size ==
+      selectionInteractionViewModel.maxAllowableSelectionCount
+    ) {
+      enabledItem.set(isAnswerSelected.get())
+    } else {
+      enabledItem.set(true)
+    }
   }
```
BenHenning, line 32: This part should be happening in SelectionInteractionViewModel, not here.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
BenHenning: Not done yet--all that's being changed is the selection item next, not the actual checkbox enabled/disabled states. I left another comment in SelectionInteractionContentViewModel as well, PTAL.

BenHenning (commented):
Thanks @Akshatkamboj14.

Does this solution actually work? It seems like it might not correctly enable/disable items if you reach the selection limit, then deselect items, then reach the limit again. I think the solution also isn't quite matching what we discussed, so I've left some follow-up comments--PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
Akshatkamboj14: I think Ben, that it should be here because `SelectionInteractionViewModel` has a different XML file that implements the textView that should be changed according to the selection that the user made i.e `selection_iteraction_item.xml` and `SelectionInteractionContentViewModel` has XML `multiple_choice_interaction_items.xml` which implements Radio Buttons.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: Okay, ben.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionViewModel.kt
Akshatkamboj14: I think ben this function is only for the text as I mentioned above that this viewmodel's Xml is only implementing the textView and the radioButton is implemented by `multiple_choice_interaction_items.xml` which is part of `SelectionInteractionContentViewModel` and Ben now the code is working fine Ig.

Akshatkamboj14:
> Thanks @Akshatkamboj14.
> 
> Does this solution actually work? It seems like it might not correctly enable/disable items if you reach the selection limit, then deselect items, then reach the limit again. I think the solution also isn't quite matching what we discussed, so I've left some follow-up comments--PTAL.

@BenHenning I have tried this code on my emulator and this is working fine I tried reaching the limit and it disabled the rest of the checkboxes that were not selected and when deselecting any of the checkboxes it then again I can select the checkbox.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/SelectionInteractionContentViewModel.kt
BenHenning: To clarify: SelectionInteractionViewModel is responsible for managing the states of all checkboxes/radio buttons being shown. It's also a dependency inversion to have this model depend on SelectionInteractionViewModel as SelectionInteractionViewModel creates instances of this model (I'm aware that it's an existing dependency, but it ought not to be so we shouldn't continue to propagate it). Since SelectionInteractionViewModel already manages checkbox checked states, it makes sense for it to also manage enabled states.

I agree that this view model should have the ObservableBoolean for enabled that's bound via XML, but doing it in this way (calling into SelectionInteractionViewModel) is where the issue is.

BenHenning (commented):
Left one follow-up comment @Akshatkamboj14 PTAL.

Akshatkamboj14:
So Ben, how can we work this out in this view model as you have mentioned SelectionInteractionContentViewModel is bounded with multiple_choice_interaction_item and we are passing ObservableBoolean also in its constructor, so what can we do to avoid calling SelectionInteractionViewModel?

BenHenning:
Since the branch was force pushed, a new version of the PR will need to be opened.

BenHenning:
> So Ben, how can we work this out in this view model as you have mentioned SelectionInteractionContentViewModel is bounded with multiple_choice_interaction_item and we are passing ObservableBoolean also in its constructor, so what can we do to avoid calling SelectionInteractionViewModel?

@Akshatkamboj14 we should only change ObservableBoolean in SelectionInteractionViewModel which will avoid needing to call up to SelectionInteractionViewModel from the content view models. They only need SelectionInteractionViewModel to decide whether to set their ObservableBoolean to true/false; we can perform that logic in SelectionInteractionViewModel, instead.

PR #4653, opened 2022-10-13
Fix #4644 : Fixes not building for m1 Mac

### LINKED ISSUE #4644: Does not build with the default setup on M1.
The project does not build with the default setup.
Running macOS Monterey 12.1 M1 building for pixel3a API 29

Following the setup for mac [here](https://github.com/oppia/oppia-android/wiki/Contributing-to-Oppia-android)
Install oppia-android and Run the app from Android Studio.

`Execution failed for task ':model:generateProto'.
> Could not resolve all files for configuration ':model:protobufToolsLocator_protoc'.
   > Could not find protoc-3.8.0-osx-aarch_64.exe (com.google.protobuf:protoc:3.8.0).
     Searched in the following locations:
         https://repo.maven.apache.org/maven2/com/google/protobuf/protoc/3.8.0/protoc-3.8.0-osx-aarch_64.exe`

Expected to build without issue.

The issue seems like the same one mentioned here #3912 
It was fixed by adding the proto platform to local.properties using the setup script (mentioned here #3943)

In the build Gradle model file the variable is fetched from root project variables and not from the local.properties.
I solved the issue by correcting this. It still remains to be tested on other environments.

antonmagnus:
Fixes #4644 

In the build Gradle model file the variable is fetched from root project variables and not from the local.properties.
Now it instead fetches from the local.properties where it was saved with the build script.

Remains to be tested before on in other environments before merging.


model/build.gradle
```diff
@@ -4,11 +4,10 @@ apply plugin: 'com.google.protobuf'
 protobuf {
   protoc {
     // To build protoc in M1 mac. For context, see: #3912.
-    if (project.rootProject.hasProperty('protobuf_platform')) {
-      artifact = "com.google.protobuf:protoc:3.8.0:${project.rootProject.property("protobuf_platform")}"
-    } else {
-      artifact = "com.google.protobuf:protoc:3.8.0"
-    }
+    Properties properties = new Properties()
+    properties.load(project.rootProject.file("local.properties").newDataInputStream())
```
BenHenning, line 8: The project doesn't have a local.properties--is this something just on your local machine?

Also, we need to set this up such that it will work for other team members without them needing to set up their own local.properties (otherwise, it's really likely people will get stuck since it won't be clear that they need to do that).

Could we use a different approach?

BenHenning (changes requested):
Thanks @antonmagnus. It looks like CI is failing, and I think it's because of the local.properties reference (which I suspect won't work for a few reasons). PTAL at my in-file comment for more specifics.

PR #507, opened 2019-12-03
Rich text image centre fix

rt4914:
## Explanation

@BenHenning had observed one issue in which if we open exploration, again and again quickly, then in some cases (approximately 10-20% times) the image in content card would cut from the very start. Check below image:
Screenshot_1575389556

This has been fixed to keep image always in centre.

Also, to test this, you will need to open exploration again and again quickly as shown in gif. Make sure you test this atleast 10 times quickly.

image-issue-2

## Checklist

- [ ] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #4669, opened 2022-10-20
Fixes #3432: Use alt-with-value from <oppia-noninteractive-image> in RichText

### LINKED ISSUE #3432: [A11y] Use alt-with-value from <oppia-noninteractive-image> in RichText

vrajdesai78:
## Explanation
Fixes #3432: Created a `handleContentDescription` function in CustomHtmlContentHandler which is further override in ImageTagHandler. With current approach, I am setting up text (content description text from alt-with-value) to image. 

## Essential Checklist

### Demo Video

https://user-images.githubusercontent.com/43074241/199427651-4e3b571c-d8a9-4f02-b05f-943cd6f8a5a0.mp4


BenHenning:
As discussed, this requires further debugging by you, correct @vrajdesai78? Sending it back to you until the solution is fully working.

Also, should this be marked as a draft since it's not yet finished?

vrajdesai78:
> As discussed, this requires further debugging by you, correct @vrajdesai78? Sending it back to you until the solution is fully working.
> 
> Also, should this be marked as a draft since it's not yet finished?

Yes we can mark it as draft PR

rt4914:
Not sure why its working when the app is coming from background but maybe log all the important information which is responsible for displaying the item-selection-interaction and from that first try to understand what is different.

For example,
when we are changing the state, ie. going from one-question to another, activity/fragment lifecycle is not shifting to other methods like `onResume`, `onCreate`, `onPause`, `onStop`, onStart`, etc. but when app is in background and is again in foreground these methods are called and maybe there is some code in these methods which is fixing the issue.

vrajdesai78:
> Not sure why its working when the app is coming from background but maybe log all the important information which is responsible for displaying the item-selection-interaction and from that first try to understand what is different.
> 
> For example, when we are changing the state, ie. going from one-question to another, activity/fragment lifecycle is not shifting to other methods like `onResume`, `onCreate`, `onPause`, `onStop`, onStart`, etc. but when app is in background and is again in foreground these methods are called and maybe there is some code in these methods which is fixing the issue.

@rt4914 I have tried to investigate lifecycle methods like onResume and onStop which are getting called when we keep our app in background and again in foreground. Suprisingly, I haven't found any such method's overriding (onResume, onStop and onPause) in our current codebase. Can you help suggest me some next steps how can I investigate this issue.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -151,6 +152,8 @@ class UrlImageParser private constructor(
 
     override fun onResourceReady(resource: T, transition: Transition<in T>?) {
       val drawable = retrieveDrawable(resource)
+      // This statement is logged correctly
+      Log.d("TAGG", "onResourceReady: Top " + htmlContentTextView.text)
```
vrajdesai78: @rt4914 added a comment here, this log statement is logged correctly and printing, alt-with-value text

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -162,6 +165,9 @@ class UrlImageParser private constructor(
             )
           proxyDrawable.initialize(drawable, computeBounds(context, drawable, viewWidth, padding))
           htmlContentTextView.text = htmlContentTextView.text
+          // This is not getting logged, it get's logged only when we go in background and
```
vrajdesai78: This statment is not logged when we open activity for the first time (when images are not loaded)

vrajdesai78:
@BenHenning can you PTAL. I have solve the issue of loading images in SelectionInteractionView. Actually, there we were calling View.width { ... } which is creating the problem, so now I am calculating the width directly by calling View.width. Also, update the video in the description with the latest code changes.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -151,19 +151,21 @@ class UrlImageParser private constructor(
 
     override fun onResourceReady(resource: T, transition: Transition<in T>?) {
       val drawable = retrieveDrawable(resource)
+      // This statement is logged correctly
       htmlContentTextView.post {
-        htmlContentTextView.width { viewWidth ->
```
BenHenning: This was originally needed to fix an issue where the image may load faster than the TextView being laid out (which isn't completely surprising--post has inherent issues which is why we're considering outright banning it per #4689). See #507. You'll need to reevaluate to make sure that issue can't happen anymore even after this change.

rt4914 (commented):
@vrajdesai78 LGTM, Just check ben's comment, in that PR I have mentioned how to test that bug so please follow those steps and create a video too displaying that the current change does not re-surface that issue.

vrajdesai78:
@BenHenning @rt4914 I have checked whether the image is getting loaded or not if we open the screen frequently as mentioned in PR #507. You can watch below video. 


Uploading rich-text-flow.mp4…

vrajdesai78:
@BenHenning @rt4914 I have checked whether the image is getting loaded or not if we open the screen frequently as mentioned in PR https://github.com/oppia/oppia-android/pull/507. You can watch below video.

https://user-images.githubusercontent.com/43074241/199593014-23b56304-c97d-4dc4-b320-b90c8f5061ad.mp4

BenHenning:
Just to confirm @vrajdesai78 was that tested with the ``.width`` function removed?

Also, I'm wondering if checkpointing is "getting in the way of" hitting the bug. Perhaps temporarily update exploration routing logic to bypass checkpointing so that the exploration is directly opened and check again that the issue doesn't repro.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: I have checked by removing .width {} and it is again introducing the similar issue #507. This issue is happening after removing checkpointing.


https://user-images.githubusercontent.com/43074241/199662466-9ff924f2-53e2-418e-94bb-94c64855eec2.mp4

vrajdesai78:
@BenHenning @rt4914 what the next steps, removing .width {} is raising old issue #507.

vrajdesai78:
@BenHenning instead of using our .width{..} function I have created a new function afterLayout according to this [stackoverflow answer](https://stackoverflow.com/a/57495622/12314934). With this solution both problems are solved 

1. Loading images in selectionInteractionView
2. Issue #507


https://user-images.githubusercontent.com/43074241/199909026-86e20bd1-4230-4639-a6dd-245c1cdf1a37.mp4

vrajdesai78:
@BenHenning @rt4914 can you re-run CI checks, they are failing due to some flaky tests.

rt4914 (approved):
I have tested it locally and it does work correctly. Thanks @vrajdesai78

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,17 +463,17 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
...
+// Reference: https://stackoverflow.com/a/57495622/12314934
+private fun View.afterLayout(what: () -> Unit) {
+  if (isLaidOut) {
+    what.invoke()
+  } else {
     viewTreeObserver.addOnGlobalLayoutListener(object : ViewTreeObserver.OnGlobalLayoutListener {
       override fun onGlobalLayout() {
         viewTreeObserver.removeOnGlobalLayoutListener(this)
-        computeWidthOnGlobalLayout(width)
+        what.invoke()
```
BenHenning:
```suggestion
        what()
```

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,17 +463,17 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
   return margin.coerceAtLeast(0f)
 }
 
-// Reference: https://stackoverflow.com/a/51865494
-private fun TextView.width(computeWidthOnGlobalLayout: (Int) -> Unit) {
-  if (width == 0) {
+// Reference: https://stackoverflow.com/a/57495622/12314934
+private fun View.afterLayout(what: () -> Unit) {
```
BenHenning: Please name ``what`` to be something clearer (as it was before).

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,17 +463,17 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
...
+  if (isLaidOut) {
```
BenHenning: Perhaps refactor the existing one to match the new solution (they seem almost identical, so I don't know why it's being rearranged so much).

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -152,15 +152,18 @@ class UrlImageParser private constructor(
...
             Rect(
               htmlContentTextView.paddingLeft,
               htmlContentTextView.paddingTop,
               htmlContentTextView.paddingRight,
               htmlContentTextView.paddingBottom
             )
-          proxyDrawable.initialize(drawable, computeBounds(context, drawable, viewWidth, padding))
+          proxyDrawable.initialize(
+            drawable,
+            computeBounds(context, drawable, htmlContentTextView.width, padding)
```
BenHenning: Does this actually need to change or can we just pass in the width to the callback as we did before?

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
```diff
@@ -46,4 +48,23 @@ class ImageTagHandler(
...
+    if (contentDescription != null) {
+      val spannableBuilder = SpannableStringBuilder(contentDescription)
+      spannableBuilder.setSpan(
+        contentDescription,
+        /* start= */ 0,
+        /* end= */ contentDescription.length,
+        Spannable.SPAN_INCLUSIVE_EXCLUSIVE
+      )
+      output.replace(openIndex, closeIndex, spannableBuilder)
+    } else consoleLogger.e("ImageTagHandler", "Failed to parse $CUSTOM_IMG_ALT_TEXT_ATTRIBUTE")
```
BenHenning: Suggest making this a warning, instead, as the app still works but in a degraded state in this situation.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -185,6 +187,21 @@ class CustomHtmlContentHandler private constructor(
      * @param indentation The zero-based indentation level of this item.
      */
     fun handleClosingTag(output: Editable, indentation: Int, tag: String) {}
+
+    /**
+     * Called when a custom tag is encountered. This is always called after the closing tag.
```
BenHenning: This documentation doesn't actually seem correct per what this function is used for--please update.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -185,6 +187,21 @@ class CustomHtmlContentHandler private constructor(
...
+     *
+     * @param attributes the tag's attributes
+     * @param openIndex the index in the output [Editable] at which this tag begins
+     * @param closeIndex the index in the output [Editable] at which this tag ends
+     * @param output the destination [Editable] to which spans can be added
+     */
+    fun handleContentDescription(
```
BenHenning: In terms of names, 'handle content description' doesn't make a lot of sense to me.

But rather than considering renaming this, what's the intent for having a separate method vs. just including this logic in one of the other overrides? It seems that special content descriptions are only needed for images, so I don't see what value a separate method provides.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
BenHenning: > @BenHenning instead of using our .width{..} function I have created a new function afterLayout according to this [stackoverflow answer](https://stackoverflow.com/a/57495622/12314934). With this solution both problems are solved
> 
> 1. Loading images in selectionInteractionView
> 2. Issue [Rich text image centre fix #507](https://github.com/oppia/oppia-android/pull/507)
> 
>  fraction-frequently.mp4

Can you provide context on *why* this fixes both issues? I think that's important here (it feels a bit like we just copied the solution from SO, but we need to understand why the approach actually works).

BenHenning (commented):
Thanks @vrajdesai78. I've taken a full pass and have left some suggestions--PTAL.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
vrajdesai78: Removed

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
vrajdesai78: Removed handleContentDescription

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
vrajdesai78: done

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Refactored existing solution

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Renamed it to onGlobalLayoutListener thanks.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Updated

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Previously inside .width{..} method we are first calculating the width and checking if width is equals to zero or not and then we call onGlobalLayoutListener which actually won't work with the case when we have images inside selection interaction with alt-with-value (content description). So, instead of checking width here we are checking whether the view `isLaidOut` or not. `isLaidOut` will be used to check that the view has been through at atleast one layout.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Changed it back to width we get from callback

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,17 +463,17 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
   return margin.coerceAtLeast(0f)
 }
 
-// Reference: https://stackoverflow.com/a/51865494
+// Reference: https://stackoverflow.com/a/57495622/12314934
 private fun TextView.width(computeWidthOnGlobalLayout: (Int) -> Unit) {
-  if (width == 0) {
+  if (isLaidOut) {
```
BenHenning: Invert this condition (so again the difference is minimized between the two to better emphasize the change).

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,17 +463,17 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
   return margin.coerceAtLeast(0f)
 }
 
-// Reference: https://stackoverflow.com/a/51865494
+// Reference: https://stackoverflow.com/a/57495622/12314934
```
BenHenning:
```suggestion
// References: https://stackoverflow.com/a/51865494 and https://stackoverflow.com/a/57495622/12314934.
```

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -160,7 +160,10 @@ class UrlImageParser private constructor(
               htmlContentTextView.paddingRight,
               htmlContentTextView.paddingBottom
             )
-          proxyDrawable.initialize(drawable, computeBounds(context, drawable, viewWidth, padding))
+          proxyDrawable.initialize(
+            drawable,
+            computeBounds(context, drawable, viewWidth, padding)
+          )
```
BenHenning: Can the old code formatting be kept now?

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
BenHenning: Your explanation more or less just repeats _what_ the code is doing @vrajdesai78 but what I'm actually asking for is the _why_. Specifically, why does the width == 0 check fail when using content descriptions? And, why does isLaidOut work to address both that issue & the original issue fixed by this global layout codef?

Please be clear on the 'why'--it's really important as it's the core purpose of this code. Please also capture this in an in-line comment in the code so that future developers can better contextualize the method if future changes need to be made.

BenHenning (commented):
Thanks @vrajdesai78. Just had a few comments left--PTAL.

I also restarted the one failing CI check since it timed out.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: I have kept the original code of .width {..} method. I have just added requestLayout() before calling onGlobalLayoutListener. Without using requestLayout() onGlobalLayoutListener is not getting called when we have images inside an recyclerview with content description. The reason why it happened is because onGlobalLayout is invoked only when the global layout state changes and it might be possible that the view is created before the request to load images processed (may be the http request). [Stackoverflow](https://stackoverflow.com/a/35444014/12314934) answer to support my observation. Here requestLayout() will help to call onGlobalLayoutListener which solves the problem of loading images with content description in a recyclerview.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Now, we don't need this.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: done.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Done

vrajdesai78:
@BenHenning can you PTAL. I have replied to all of your comments and also mentioned why this solution works. Thanks

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,9 +460,13 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
   return margin.coerceAtLeast(0f)
 }
 
-// Reference: https://stackoverflow.com/a/51865494
+// Reference: https://stackoverflow.com/a/51865494 and
+// https://stackoverflow.com/a/35444014/12314934.
 private fun TextView.width(computeWidthOnGlobalLayout: (Int) -> Unit) {
   if (width == 0) {
+    // requestLayout() will request onGlobalLayoutListener when we have images inside recyclerview
+    // with a content description.
```
BenHenning: I don't think this quite fully explains the "why." What I think is missing: what about the situation of images in a recycler view leads to the global layout listener approach not working (whereas it works for other situations)? That, I think, is the missing context that we need to detail here.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
BenHenning: Thanks @vrajdesai78! That's really helpful--can you please also update the in-line comment to include this context? (I also left one new comment asking the same).

BenHenning (commented):
Thanks @vrajdesai78. Just a bit of work needed left on the UrlImageParser comment, but otherwise the PR LGTM.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Updated in inline comment, thanks.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
vrajdesai78: Updated the comment, I hope now it give context.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,9 +460,14 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
...
 }
 
-// Reference: https://stackoverflow.com/a/51865494
+// Reference: https://stackoverflow.com/a/51865494 and
+// https://stackoverflow.com/a/35444014/12314934.
 private fun TextView.width(computeWidthOnGlobalLayout: (Int) -> Unit) {
   if (width == 0) {
+    // In a recyclerview, where there are images with content descriptions, onGlobalLayoutListener
+    // will not be called because the view is created before the request to load images is
+    // processed, so calling requestLayout() will call onGlobalLayoutListener.
```
BenHenning:
```suggestion
    // processed, so calling requestLayout() will ensure that onGlobalLayoutListener is called.
```

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -460,9 +460,14 @@ private fun calculateInitialMargin(availableAreaWidth: Int, drawableWidth: Float
   return margin.coerceAtLeast(0f)
 }
 
-// Reference: https://stackoverflow.com/a/51865494
+// Reference: https://stackoverflow.com/a/51865494 and
```
BenHenning:
```suggestion
// References: https://stackoverflow.com/a/51865494 and
```

BenHenning (commented):
Thanks @vrajdesai78 just had a few nits--PTAL.

BenHenning (approved):
Thanks @vrajdesai78--the PR LGTM. Also, please make sure not to resolve revewiers' comments (since it makes it easier to then lose these comments).

vrajdesai78:
> Thanks @vrajdesai78 just had a few nits--PTAL.

@BenHenning fixed nit changes

vrajdesai78:
@BenHenning @rt4914 can you re-run CI checks, it failed due to connection timed out

Issue #4684, opened 2022-10-31
Promoted story cards fit images rather than constraining them


BenHenning:
**Describe the bug**
Promoted story thumbnail cards do not consistently size thumbnail image sizes.

**To Reproduce**
1. Open a tip-of-tree version of the app with production assets.
2. Navigate to the home screen
3. Ensure that "What is Addition?" shows up in the top recommended stories.
4. Observe that the "What is Addition?" thumbnail is much larger than others.

**Expected behavior**
Promoted story card thumbnails should be consistently sized and not dependent on their physical sizes.

**Demonstration**
Screenshot:

![image](https://user-images.githubusercontent.com/12983742/199122557-1b5ed681-ca17-4b79-a809-f47b09c88fcb.png)

**Environment**
- Device/emulator being used: Nexus 5X
- Android or SDK version (e.g. Android 5 or SDK 21): SDK 27
- App version (you can get this through system app settings or via the admin controls menu in-app): developer build using e4e4de63e0ade2a1f15e961be447dc3bf6d12eb0.

**Additional context**
Note this requires production assets--the local developer assets don't seem to trigger the issue.

BenHenning:
91014af586eb0f27f71b8af260fabf5f5f91bdac from #4587 seems suspicious since it changes promoted_story_card to wrap_content instead of match_parent.

BenHenning:
NB: This is a really significant visual regression, so we should fix this before the next beta release.

BenHenning:
I can confirm this is fixed after testing the latest beta version.

PR #4587, opened 2022-09-11
Fixes #4586: Improve text scaling according to accessibility scanner 

vrajdesai78:
## Explanation
Fixes #4586: Improved text scaling by setting layout_height and layout_width to wrap_content and setting min_height = 48dp and min_width = 48dp (may vary in some cases). 

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

##  A11YS result before changes
image
image
image

## A11Ys result after changes
image

image

PR #4695, opened 2022-11-04
Fixes #4586: Improve text scaling according to screen reader

### LINKED ISSUE #4586: Improve text scaling according to accessibility scanner
**Describe the bug**
In Android 12, accessibility scanner is showing suggestion to improve text scaling for cases in which we have set fixed width or height of view group which will create a problem in scaling test. 

**Expected behavior**
Accessibility scanner should not show any improvements for text scaling

**Demonstration**
image

image

**Environment**
- Device/emulator being used: Realme GT Master Edition
- Android version: Android 12
- App version : 1.0.0

vrajdesai78:
## Explanation
Fixes #4586: Changed layout_width to sp instead of dp as per this [comment](https://github.com/oppia/oppia-android/issues/4586#issuecomment-1304039347). This actually solves the problem of text scaling.  Improved text scaling by setting layout_height and layout_width to wrap_content and setting min_height = 48dp and min_width = 48dp (may vary in some cases). Introduced a tests which will make sure that width is set in SP not in DP. 

Note: In accessibility scanner it still shows to improve text scaling but visually or practically this solution looks accurate. 

## Essential Checklist

## Screenshots For Mobile (Promoted Story Card)

Default Fonts          |  Large Fonts
:-------------------------:|:-------------------------:
Screenshot from 2022-11-05 01-27-47  |  Screenshot from 2022-11-05 01-28-36
Screenshot from 2022-11-05 01-30-00  |  Screenshot from 2022-11-05 01-29-23

## Screenshots for Tablet (Promoted Story Card)
Default Fonts          |  Large Fonts
:-------------------------:|:-------------------------:
Screenshot from 2022-11-05 01-31-14  |  Screenshot from 2022-11-05 01-31-36
Screenshot from 2022-11-05 01-24-34  |  Screenshot from 2022-11-05 01-25-16

Failing robolectric tests when we change the width in dp. 
Screenshot from 2022-11-07 14-39-57


BenHenning:
Thanks @vrajdesai78. I think this seems sensible, but I have two questions:
1. Doesn't the rest of #4587 need to be rolled forward/re-added in order to fully fix #4586?
2. How do we ensure that this doesn't break again in the future? It'd be ideal to add some tests to ensure it doesn't break.

vrajdesai78:
> Thanks @vrajdesai78. I think this seems sensible, but I have two questions:
> 
> 1. Doesn't the rest of [Fixes #4586: Improve text scaling according to accessibility scanner  #4587](https://github.com/oppia/oppia-android/pull/4587) need to be rolled forward/re-added in order to fully fix [Improve text scaling according to accessibility scanner  #4586](https://github.com/oppia/oppia-android/issues/4586)?
> 2. How do we ensure that this doesn't break again in the future? It'd be ideal to add some tests to ensure it doesn't break.

I have added all changes from previous PR. For test I have added one test but not sure it was not failing when we have width set in dp.

app/src/main/res/layout/promoted_story_card.xml
```diff
@@ -17,6 +17,7 @@
     android:layout_marginStart="@dimen/promoted_story_card_layout_margin_start"
     android:layout_marginEnd="8dp"
     android:layout_marginBottom="8dp"
+    android:minWidth="@dimen/promoted_story_card_layout_width"
```
rt4914: This minWidth is of no use because above there is `layout_width` which is same as this.

app/src/main/res/layout/promoted_story_card.xml
```diff
@@ -36,6 +37,7 @@
         android:layout_width="match_parent"
         android:layout_height="0dp"
         android:importantForAccessibility="no"
+        android:minWidth="280dp"
```
rt4914: Are you sure this is needed?

We discussed using `sp` for width instead of `dp` which should have solved the problem.

app/src/main/res/layout/promoted_story_card.xml
```diff
@@ -17,6 +17,7 @@
     android:layout_marginStart="@dimen/promoted_story_card_layout_margin_start"
     android:layout_marginEnd="8dp"
     android:layout_marginBottom="8dp"
+    android:minWidth="@dimen/promoted_story_card_layout_width"
     app:cardCornerRadius="4dp"
     app:cardElevation="4dp">
```
rt4914: There are four constraints mentioned for `ConstraintLayout ` which is not needed as the parent of this layout is `MaterialCardView`

app/src/main/res/layout/recently_played_fragment.xml
```diff
@@ -27,4 +27,4 @@
       android:layout_height="6dp"
       android:background="@drawable/toolbar_drop_shadow" />
   </FrameLayout>
-</layout>
+</layout>
```
rt4914: EOF is missing, add one blank line at the end.

app/src/main/res/values/dimens.xml
```diff
@@ -540,7 +540,7 @@
 
   <dimen name="promoted_story_card_layout_margin_end">8dp</dimen>
   <dimen name="promoted_story_card_layout_margin_start">8dp</dimen>
-  <dimen name="promoted_story_card_layout_width">280dp</dimen>
+  <dimen name="promoted_story_card_layout_width">280sp</dimen>
```
rt4914, line 544: Please keep these two values `promoted_story_card_width ` and `promoted_story_card_layout_width ` close to each other.

Maybe just shift `promoted_story_card_width ` above this `promoted_story_card_layout_width `
Also add a comment mentioning why these values are in sp.

rt4914 (changes requested):
@vrajdesai78 PTAL

app/src/main/res/layout/promoted_story_card.xml
vrajdesai78: Removed

app/src/main/res/layout/promoted_story_card.xml
vrajdesai78: Ok removed those constraints

app/src/main/res/layout/promoted_story_card.xml
vrajdesai78: Removed it

app/src/main/res/layout/recently_played_fragment.xml
vrajdesai78: Done

app/src/main/res/values/dimens.xml
vrajdesai78: Done

app/src/main/res/layout/license_text_viewer_activity.xml
```diff
@@ -41,6 +43,7 @@
       </androidx.appcompat.widget.Toolbar>
     </com.google.android.material.appbar.AppBarLayout>
 
+
```
BenHenning: Nit: please remove the extra newline added here.

app/src/main/res/layout/mark_chapters_completed_fragment.xml
```diff
@@ -31,7 +31,7 @@
       <androidx.appcompat.widget.Toolbar
         android:id="@+id/mark_chapters_completed_toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
```
BenHenning, line 34: Doesn't a minHeight also need to be specified for this toolbar?

app/src/main/res/layout/mark_stories_completed_fragment.xml
```diff
@@ -31,7 +31,7 @@
       <androidx.appcompat.widget.Toolbar
         android:id="@+id/mark_stories_completed_toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
```
BenHenning, line 34: Ditto for this one.

app/src/main/res/layout/mark_topics_completed_fragment.xml
```diff
@@ -31,7 +31,7 @@
       <androidx.appcompat.widget.Toolbar
         android:id="@+id/mark_topics_completed_toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
```
BenHenning, line 34: And this one.

app/src/main/res/layout/my_downloads_fragment.xml
```diff
@@ -19,7 +19,7 @@
       <androidx.appcompat.widget.Toolbar
         android:id="@+id/my_downloads_toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
```
BenHenning, line 22: And this one.

app/src/main/res/layout/policies_activity.xml
```diff
@@ -19,7 +19,6 @@
         android:layout_height="wrap_content"
         android:layout_gravity="end"
         android:background="@color/oppia_primary"
-        android:minHeight="?attr/actionBarSize"
```
BenHenning: Why is this being removed? This seems inconsistent with other toolbars in the app.

app/src/main/res/layout/profile_and_device_id_activity.xml
```diff
@@ -19,7 +19,7 @@
     <androidx.appcompat.widget.Toolbar
       android:id="@+id/profile_and_device_id_toolbar"
       android:layout_width="match_parent"
-      android:layout_height="?attr/actionBarSize"
+      android:layout_height="wrap_content"
```
BenHenning, line 22: Ditto here (no minHeight).

app/src/main/res/layout/profile_edit_activity.xml
```diff
@@ -18,7 +18,7 @@
     <androidx.appcompat.widget.Toolbar
       android:id="@+id/profile_edit_toolbar"
       android:layout_width="match_parent"
-      android:layout_height="?attr/actionBarSize"
+      android:layout_height="wrap_content"
```
BenHenning, line 21: And here.

app/src/main/res/layout/profile_picture_activity.xml
```diff
@@ -27,7 +27,7 @@
       <androidx.appcompat.widget.Toolbar
         android:id="@+id/profile_picture_activity_toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
```
BenHenning, line 30: And here.

app/src/main/res/layout/profile_progress_header.xml
```diff
@@ -105,9 +105,10 @@
         <TextView
           android:id="@+id/ongoing_topics_description_text_view"
           style="@style/TextViewCenter"
-          android:layout_width="60dp"
+          android:layout_width="wrap_content"
```
BenHenning, line 109: Given that this change now lets this & the next changed text view grow beyond 60dp--is the result for longer text going to have the same effect of being cut off or will it lay out differently now?

app/src/main/res/layout/topic_practice_footer_view.xml
```diff
@@ -20,7 +20,8 @@
     <Button
       android:id="@+id/topic_practice_start_button"
       style="@style/StateButtonInactive"
-      android:layout_width="160dp"
+      android:layout_width="wrap_content"
```
BenHenning, line 23: Will this button actually look correct if the text extends past 160dp?

app/src/main/res/values/styles.xml
```diff
@@ -425,11 +429,14 @@
   </style>
   <!-- Continue Lesson Button -->
   <style name="ContinueLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
-    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_width">wrap_content</item>
```
BenHenning: Why is minWidth set to 48dp below when elsewhere we set the minWidth to the previously fixed width?

Also, similar question to an earlier one: will the button actually look correct if the text is larger?

Ditto on both for StartOverLessonButton.

app/src/main/res/values/dimens.xml
```diff
@@ -540,7 +539,10 @@
 
   <dimen name="promoted_story_card_layout_margin_end">8dp</dimen>
   <dimen name="promoted_story_card_layout_margin_start">8dp</dimen>
-  <dimen name="promoted_story_card_layout_width">280dp</dimen>
+  <!-- The width is set to sp so that it will scale according to the device screen -->
+  <dimen name="promoted_story_card_layout_width">280sp</dimen>
+  <!-- The width is set to sp so that it will scale according to the device screen -->
+  <dimen name="promoted_story_card_width">280sp</dimen>
```
BenHenning, line 546: Why aren't we using this solution elsewhere in the PR? How does a developer decide between using the minWidth/Height approach and using sp for such cases?

app/src/main/res/layout/promoted_story_card.xml
```diff
@@ -25,17 +25,14 @@
...
-      app:layout_constraintStart_toStartOf="parent"
-      app:layout_constraintTop_toBottomOf="@id/recently_played_stories_text_view"
       app:roundedRectDrawableWithColor="@{viewModel.promotedStory.lessonThumbnail.backgroundColorRgb}">
 
       <org.oppia.android.app.customview.LessonThumbnailImageView
         android:id="@+id/lesson_thumbnail"
         android:layout_width="match_parent"
         android:layout_height="0dp"
         android:importantForAccessibility="no"
+        android:minWidth="280sp"
```
BenHenning: Given this is 280sp, should it be tied to a specific dimension for parity with the broader promoted story card size?

app/src/main/res/layout/promoted_story_card.xml
```diff
@@ -45,7 +42,7 @@
         app:lessonThumbnail="@{viewModel.promotedStory.lessonThumbnail}" />
 
       <LinearLayout
-        android:layout_width="match_parent"
+        android:layout_width="0dp"
```
BenHenning: Why is this changed correct? It seems unrelated compared to other fixes in this PR.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -893,6 +895,39 @@ class HomeActivityTest {
...
+        ).check { _, _ ->
+          val promotedStoryCard =
+            activity.findViewById<LessonThumbnailImageView>(R.id.lesson_thumbnail)
+          val pixels = promotedStoryCard?.width?.toFloat()
+          val expectedSize = TypedValue.applyDimension(
+            TypedValue.COMPLEX_UNIT_SP,
+            280F,
+            context.resources.displayMetrics
+          )
+          assertThat(pixels).isEqualTo(expectedSize)
```
BenHenning: Float comparisons need to use a threshold to ensure that rounding errors don't cause inconsistent results. You can use a mid-sized epsilon for most cases, e.g.:

```kotlin
assertThat(pixels).isWithin(1e-5f).of(expectedSize)
```

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -893,6 +895,39 @@ class HomeActivityTest {
...
+        onView(
+          atPositionOnView(
+            recyclerViewId = R.id.home_recycler_view,
+            position = 0,
+            targetViewId = R.id.promoted_story_list_recycler_view
+          )
+        ).check { _, _ ->
+          val promotedStoryCard =
+            activity.findViewById<LessonThumbnailImageView>(R.id.lesson_thumbnail)
+          val pixels = promotedStoryCard?.width?.toFloat()
```
BenHenning: For this & the next variable, the names are a bit generic. Suggest renaming them to be more specific as to what they are actually representing ("pixels" is just a unit, and "size" could represent any number of things).

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -893,6 +895,39 @@ class HomeActivityTest {
     }
   }
 
+  @Test
+  fun testHomeActivity_promotedStoryHasScalableWidth() {
```
BenHenning, line 905: Please update the PR description with:
- A demonstration that this test fails with the old fix & passes with the new fix
- A demonstration that this entire test suite passes when run on Espresso

BenHenning (changes requested):
Thanks @vrajdesai78. Took a full pass. Many of the comments are duplicates, but I wanted to make sure each case I found for repeated issues were covered by comments.

From a top-level perspective:
- I think the PR title needs to be updated since this PR is now covering more than just the promoted story list fixes
- In the original PR there were changes made to app/src/main/res/layout/recently_played_fragment.xml -- why were these dropped here?
-  Are there any other files that require similar changes (e.g. for toolbars & nav menus)?
- How will developers make sure that they know how to address these sp vs. dp cases long-term? Having a test or check seems important here since it's clear that simply reusing an existing pattern may have different results depending on the view being modified.

BenHenning:
Also @vrajdesai78 it looks like you have conflicts to resolve from develop--please make sure to do that before sending this back.

app/src/main/res/layout/license_text_viewer_activity.xml
vrajdesai78: Removed

app/src/main/res/layout/mark_chapters_completed_fragment.xml
vrajdesai78: Added minHeight

app/src/main/res/layout/mark_stories_completed_fragment.xml
vrajdesai78: Done

app/src/main/res/layout/mark_topics_completed_fragment.xml
vrajdesai78: Done

app/src/main/res/layout/my_downloads_fragment.xml
vrajdesai78: Done

app/src/main/res/layout/policies_activity.xml
vrajdesai78: Added again.

app/src/main/res/layout/profile_and_device_id_activity.xml
vrajdesai78: Added

app/src/main/res/layout/profile_edit_activity.xml
vrajdesai78: Done

app/src/main/res/layout/profile_picture_activity.xml
vrajdesai78: Done

app/src/main/res/layout/profile_progress_header.xml
vrajdesai78: Here, changing it to wrap content is actually a correct solution, because with the fixed width the size of card increases which won't looks good. 

With Fixed Width     |  With wrap_content
:-------------------------:|:-------------------------:
image | image

app/src/main/res/layout/promoted_story_card.xml
vrajdesai78: Here, actually we don't need this minWidth as it's layout_width is match_parent.

app/src/main/res/layout/promoted_story_card.xml
vrajdesai78: changed back to match_parent

app/src/main/res/layout/topic_practice_footer_view.xml
vrajdesai78: This is how it looks with very large font size. I don't think it will create any issue if we have wrap_content as layout_width. 

image

app/src/main/res/values/dimens.xml
vrajdesai78: We are not using this solution elsewhere may be because we don't any special case like promoted_story_card where we want a consistent width/height to all items. If we use wrap_content with minWidth/minHeight items inside a recyclerview items width/height will not be consistent or same.

app/src/main/res/values/styles.xml
vrajdesai78: set minWidth to 144dp

With Fixed Width     |  With wrap_content
:-------------------------:|:-------------------------:
image | image

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Will add screenshots once test fails on setting width in dp. Currently, test is passing int both the case sp and dp.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Updated variable name.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Updated

vrajdesai78:
> Thanks @vrajdesai78. Took a full pass. Many of the comments are duplicates, but I wanted to make sure each case I found for repeated issues were covered by comments.
> 
> From a top-level perspective:
> 
> * I think the PR title needs to be updated since this PR is now covering more than just the promoted story list fixes
> * In the original PR there were changes made to app/src/main/res/layout/recently_played_fragment.xml -- why were these dropped here?
> * Are there any other files that require similar changes (e.g. for toolbars & nav menus)?
> * How will developers make sure that they know how to address these sp vs. dp cases long-term? Having a test or check seems important here since it's clear that simply reusing an existing pattern may have different results depending on the view being modified.

@BenHenning I have updated the title as per your suggestion. recently_played_fragment.xml is dropped from this PR because the toolbar from recently_played_fragment is shifted to recently_played_activity and in recently_played_activity.xml layout_height of toolbar is already wrap_content with minHeight. For tests, I am not able to retrieve the exact width of promoted_story_card inside a test. With my current implementation, everytime I run tests I get 840 width even after chainging it's width from dp to sp.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -904,26 +904,24 @@ class HomeActivityTest {
...
+        atPositionOnView(
+          recyclerViewId = R.id.home_recycler_view,
+          position = 1,
+          targetViewId = R.id.promoted_story_list_recycler_view
+        )
+      ).check { view, _ ->
+        val promotedStoryCard =
+          view.findViewById<LessonThumbnailImageView>(R.id.lesson_thumbnail)
+        val promotedStoryCardWidth = promotedStoryCard?.width?.toFloat()
+        val expectedWidthInSP = TypedValue.applyDimension(
```
BenHenning: Isn't this actually in pixels?

app/src/main/res/values/dimens.xml
```diff
@@ -541,9 +541,9 @@
   <dimen name="promoted_story_card_layout_margin_end">8dp</dimen>
   <dimen name="promoted_story_card_layout_margin_start">8dp</dimen>
   <!-- The width is set to sp so that it will scale according to the device screen -->
-  <dimen name="promoted_story_card_layout_width">280sp</dimen>
+  <dimen name="promoted_story_card_layout_width">280dp</dimen>
```
BenHenning: Reminder to switch this & promoted_story_card_width with their sp variants before the PR is merged.

app/src/main/res/layout/profile_progress_header.xml
BenHenning: Ah I see. This works because the card view is constrained using constraint layout so it'll eventually be made to fit.

app/src/main/res/layout/topic_practice_footer_view.xml
BenHenning: Hmm, 'Start' is rather short in English (as compared to what it might be in other languages), but it's also constrained by ConstraintLayout so I'm guessing the text will eventually be forced to wrap within the button.

app/src/main/res/values/dimens.xml
BenHenning: Got it, but I'm not sure that answers my second question--it seems really easy to miss this detail in future development. How do we make sure that doesn't happen?

app/src/main/res/values/styles.xml
BenHenning: Thanks--I think whether this approach works very much depends on the parent layout (the fact that a ConstraintLayout is being used ensures that the buttons always look correct, but other parent layouts may introduce issues with this approach I think).

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
BenHenning: > > Thanks @vrajdesai78. Took a full pass. Many of the comments are duplicates, but I wanted to make sure each case I found for repeated issues were covered by comments.
> > From a top-level perspective:
> > 
> > * I think the PR title needs to be updated since this PR is now covering more than just the promoted story list fixes
> > * In the original PR there were changes made to app/src/main/res/layout/recently_played_fragment.xml -- why were these dropped here?
> > * Are there any other files that require similar changes (e.g. for toolbars & nav menus)?
> > * How will developers make sure that they know how to address these sp vs. dp cases long-term? Having a test or check seems important here since it's clear that simply reusing an existing pattern may have different results depending on the view being modified.
> 
> @BenHenning I have updated the title as per your suggestion. recently_played_fragment.xml is dropped from this PR because the toolbar from recently_played_fragment is shifted to recently_played_activity and in recently_played_activity.xml layout_height of toolbar is already wrap_content with minHeight. For tests, I am not able to retrieve the exact width of promoted_story_card inside a test. With my current implementation, everytime I run tests I get 840 width even after chainging it's width from dp to sp.

I think it's possible to programmatically set text scaling, see:
https://github.com/oppia/oppia-android/blob/5cd31c36658a8fb840c0b1fc8209726d34d2529a/app/src/main/java/org/oppia/android/app/utility/FontScaleConfigurationUtil.kt#L16

You can also, in Robolectric, specify the smallest width size and the display density for the "device" being run, so you should be able to set up the environment to fit different sized versions of the app (in case you need a larger window for the sp-based elements to grow correctly). This may only work in Robolectric (not sure since the font scaling should, theoretically, work in Espresso), but that should be fine for this regression test.

BenHenning (commented):
Thanks @vrajdesai78. Took another pass--PTAL.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: No, we want to compare the width with it's equivalent size in SP only.

app/src/main/res/values/dimens.xml
vrajdesai78: Changed it back to sp

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Thanks @BenHenning for your suggestion, I have used adjustFontScale and updated the test accordingly. Now the test is failing when we set width in dp instead of sp. But as you said it works only on robolectric not in espresso.

app/src/main/res/values/dimens.xml
vrajdesai78: I think we can mention this in  [General tips to make accessible](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide#general-tips-to-make-app-accessible) guideline to use width in sp for recyclerview items instead of using wrap_content.

vrajdesai78:
@BenHenning I have updated test and also added screenshot of failing test case in the PR description. Can you PTAL. Thanks

app/src/main/res/values/dimens.xml
```diff
@@ -541,7 +540,10 @@
 
   <dimen name="promoted_story_card_layout_margin_end">8dp</dimen>
   <dimen name="promoted_story_card_layout_margin_start">8dp</dimen>
-  <dimen name="promoted_story_card_layout_width">280dp</dimen>
+  <!-- The width is set to sp so that it will scale according to the device screen -->
```
rt4914, line 543: `The width is set to sp so that it will scale according to the device screen and won't fail for Accessibility Scanner on latest android versions.`

Or something similar to add more context to it.

app/src/main/res/values/dimens.xml
rt4914: @BenHenning Ideally we should never use `sp` for any non-font based dimension. But promoted story card is special case because here we have constraints like keeping all items width equal + the width should be scalable too and its doable only with `sp` correctly. 

So basically developers should use `dp` only and that too normally with `minWidth` / `minHeight` i.e.e avoid using fixed withd/height.

rt4914 (approved):
@vrajdesai78 I have one suggested.

Also approving this PR not to block you and also relying for test on @BenHenning 's suggestions.

app/src/main/res/values/dimens.xml
vrajdesai78: Even if we use sp, accessibility scanner still shows suggestion to improve text scaling. If we set width to wrap_content then only accessiblity scanner doesn't show any improvement but here it's not possible.

rt4914: Ack.

app/src/main/res/values/dimens.xml
rt4914: Okay, till above comment I was assuming that it should pass on Accessibility Scanner but actually its not.
It fails with following error.



Honestly, I feel in this case Accessibility Scanner is incorrect because the width/height of the card is actually scalable and the UI does look correct.

The below screenshots shows how smallest and largest font size looks and affect the width/height of card 
 
Screenshot_20221107_231855


Any thoughts on this @BenHenning

rt4914 (changes requested):
Blocking this because of this comment thread https://github.com/oppia/oppia-android/pull/4695#discussion_r1015728054 @BenHenning

app/src/main/res/values/dimens.xml
rt4914: @BenHenning I have one suggested. I think we need design update here. Have a look at below design

479e97e9-cef1-45f6-9b3d-70079bad401b

In this design there are three parts:
1. The cards of full width, ie. match_parent (minus the margins) -- so no Accessibility issue
2. The cards slide automatically after some time (-- maybe not required for other case)
3. The dots at bottom represent the number + position of items.


I know the purpose of our cards and these cards are really different so another suggested solution is 

b0398533-7302-4832-a4f7-04a1119b402f

I have note tested this solution so cannot guarantee if it will definitely pass on Accessibility Scanner but the way this will get implemented, in that the item which is on display will be on `match_parent` and therefore should theoretically pass on Accessibility Scanner.

vrajdesai78: I get your point @rt4914 but still I am thinking how we can add all of the content (3 text views inside a linear layout) with your proposed layout. As for our application we have thumbnail images and we can't put text inside an image, like from UI perspective it won't looks good. Also, I still believe that with your approach of making a card of full width will re-create the issue #4684 as the thumbnail image sizes are not consistent.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -896,7 +901,9 @@ class HomeActivityTest {
   }
 
   @Test
+  @RunOn(TestPlatform.ROBOLECTRIC) // TODO(#3840): Make this test work on Espresso.
```
BenHenning: Is this the correct issue to be pointing to? The changes don't seem to have anything to do with languages.

app/src/main/res/values/dimens.xml
BenHenning: So I think we should at least make sure the accessibility guidelines call out this special case (i.e. the case of needing to use a fixed width container with text contained) along with an explanation for how to make it correct for accessibility.

Re: the banner approach, I think my main questions would be whether we can get the dragging functionality working there, and what the cycling experience is when using a screenreader. With regards to the contents, why wouldn't three text views & an image work? The first screenshot seems to do something quite similar.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
BenHenning: Sorry to clarify, the returned unit from applyDimension _has_ to be pixels, here, otherwise the comparison wouldn't work (since a view's width is always provided in pixels). I think we need to rename this value since it's unit isn't sp.

BenHenning (commented):
Thanks @vrajdesai78. Just to clarify one part of your PR description: should "Failing robolectric tests when we change the width in sp." be "Failing robolectric tests when we change the width **to dp**" (notice the emphasis)? The current wording doesn't seem correct since the entire intent is to be using sp here.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Filed a new issue #4700 and updated issue number in TODO.

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
vrajdesai78: Renamed it `expectedWidthInPixels`

app/src/main/res/values/dimens.xml
vrajdesai78: How can I update the accessibility guidelines in Wiki, or can you update it, if possible. Thanks

The exact guideline needs to be added as Follows:

- We can use width in SP if we have text inside a fixed width container. This will increase the size of container whenever we increase the font size, so the scaled text get enough size to fit inside a container.

vrajdesai78:
@BenHenning, I have replied to all of your comments, can you PTAL. Thanks

app/src/main/res/values/dimens.xml
rt4914: @BenHenning 
For banners the talkback will actually work with good experience. The main difference will be folloowing:

For sighted users
- The banners will be cyclic i.e. item-0,item1,item2,item0 and repeat

For talkback users
- the cyclic nature will stop
- the screenreader will start from item-0, item 1, item-2 and next it will go out of list.

rt4914:
My latest comment @vrajdesai78 @BenHenning https://github.com/oppia/oppia-android/pull/4695#discussion_r1016745130

app/src/main/res/values/dimens.xml
vrajdesai78: @BenHenning as you have suggested I have created a [google doc](https://docs.google.com/document/d/1FkZaFTT2hRo_3JShqhCwCZVVLGzUPiriW06Xme1EVaQ/edit?usp=sharing) to update the wiki. 

Also, I have filed a new issue #4703 to track the new approach suggested by Rajat

BenHenning: Thanks @vrajdesai78.

@rt4914 that does sound like a better experience. Do you think we should add a TODO anywhere for #4703 or keep it separate from the code? Trying to figure out how to best arrange things to minimize potential user error if we ever run into this situation again (the wiki changes are a good start). Also, could you take a look at Vraj's wiki changes? You have more context than I do here.

BenHenning (approved):
Thanks @vrajdesai78. PR LGTM, so I'm deferring to @rt4914 to resolve my last convo thread (but I'm happy to take another look if needed). I also took a pass on the doc--please let me know when it's ready for another pass.

rt4914 (approved):
Approving this PR.

Replying to @BenHenning [comment](https://github.com/oppia/oppia-android/pull/4695#discussion_r1017160358)

1. The implementation of promoted stories recyclerview will change as mentioned in https://github.com/oppia/oppia-android/issues/4703 . This change is needed to mainly fix one primary issue, i.e., currently we have a fixed-width of `280` for promoted-story-cards but it should be either `match_parent` so as to provide better experience to non-sighted users. Other details related to implementation are mentioned in the issue description
2. No TODO has been mentioned in the code as this is neither related to xml files directly nor to bindable adapter, mostly we will need to create a new adapter to implement this functionality. Currently not sure if we can tie this to one single file, therefore not mentioned in code.
3. I will apply those wiki changes as mentioned in https://github.com/oppia/oppia-android/pull/4695#discussion_r1016164073

app/src/main/res/values/dimens.xml
vrajdesai78: Resolving this conversation as @rt4914 already replied to your comment and it's blocking PR from getting merged

PR #4781, opened 2022-12-05
Add generates_api = True to compiler_annotation_processor.

tjgq:
So that Java header compilation works with HEAD Bazel.


BenHenning:
Thanks @tjgq. There's a bit of set up you'll need to do if you want to check this PR in, otherwise I can include it with one I was planning later in the week to update the app broadly for new Kotlin & Bazel support.

Do you have a preference?

tjgq:
@BenHenning please feel free to include this in your other changes - thanks!

PR #4814, opened 2022-12-30
Fix #4765: The Continue animation isn't interactive

### LINKED ISSUE #4765: The Continue animation isn't interactive
**Describe the bug**

The current continue button animation is too subtle and the user has to squint carefully to see what is going on.

JishnuGoyal:
## Explanation

Fixes https://github.com/oppia/oppia-android/issues/4765

Demo Video:

https://user-images.githubusercontent.com/64526117/210055420-016f2dd2-46d6-4166-ae2a-be4c98710299.mp4

This PR tries to handle the problem of the continue interaction button's animation not being eye-catching to the users. Currently, it does so by changing the animation to a shake animation. The button now does a little wobble, starting 45s after the screen is shown. It wobbles 2 times, and then repeats the animation after every 8 seconds. 

Only with further testing and trial, will we know how the animation appeals to the users and resolves the problem at hand.

**Technical detail:**
The anim file (xml) is set to play the animation 2 times. 
In the `ContinueButtonView.kt` file, the startAnimating() function has been made recursive, which calls itself after every 8 seconds in order to repeat the animation. The 8 second gap is achieved using the lifeCycleSafeTimerFactory.

## Essential Checklist


app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
```diff
@@ -116,9 +118,15 @@ class ContinueButtonView @JvmOverloads constructor(
...
-    val animation = AnimationUtils.loadAnimation(context, R.anim.scale_button_size)
+    val animation = AnimationUtils.loadAnimation(context, R.anim.wobble_button_animation)
     if (enableContinueButtonAnimation.value) {
       startAnimation(animation)
+      // Repeat the animation after a fixed interval.
+      lifecycleSafeTimerFactory.createTimer(INTERVAL_BETWEEN_CONTINUE_BUTTON_ANIM_MS).observe(
+        fragment
+      ) {
+        startAnimating()
+      }
```
BenHenning:
```suggestion
      lifecycleSafeTimerFactory.createTimer(
        INTERVAL_BETWEEN_CONTINUE_BUTTON_ANIM_MS
      ).observe(fragmen)) {
        startAnimating()
      }
```

Generally prefer line-breaking at the higher syntactical element.

That also being said, does just passing in ``::startAnimating`` instead of the lambda work?

BenHenning (commented):
Thanks @JishnuGoyal. Left one follow-up comment. Will follow up on the animation specifically in a follow-up comment.

BenHenning:
So I definitely think the wobble is better than the pulsing. It still has a disadvantage in that the touch target is moving, but it's much less likely to introduce clicking issues here than with the previous animation. I'm less certain about the 8 second delay between the animations--that seems a bit long.

@seanlip do you have any thoughts on the updated animation?

seanlip:
I think the updated animation looks fine -- let's go with it. I think the 8-second delay is OK since this isn't meant to be as "in your face".

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
JishnuGoyal: Added the suggestive commit. Re: ::startAnimating, this isn't the lambda but the recursive function itself; Am i missing something?

JishnuGoyal:
PTAL @BenHenning. Going with the current animation per the suggestion given by Sean.

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
BenHenning: The method reference is just a syntax sugar alternative to the lambda (it'll result in generating the observer implementation). Consider resolving this optional, but a nice-to-have.

BenHenning (approved):
Thanks @seanlip and @JishnuGoyal.

@JishnuGoyal the PR LGTM. There's one comment for you to resolve (I'm fine with you marking it as resolved once it's addressed one way or the other), and I think the PR needs to be updated with the latest develop before it can be merged. You'll also need @rt4914's approval for the XML changes.

rt4914 (changes requested):
Really nice and elegant solution. Thanks @JishnuGoyal

rt4914 (approved):
Meant to approve it.

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
JishnuGoyal: image
This would be the way to do it, and in my opinion, this looks a bit more complex. WDYT? I'll change as per the review if  you still suggest so!

JishnuGoyal:
Thank you so much @rt4914! Glad to hear this! <3
PTAL @BenHenning, left a reply to the review comment, will do as you suggest this time :)

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
BenHenning: Ah, no that's not what I was thinking, and is definitely less readable than the current code. I was instead thinking something like this:

```kotlin
lifecycleSafeTimerFactory.createTimer(
  INTERVAL_BETWEEN_CONTINUE_BUTTON_ANIM_MS
).observe(fragment, ::startAnimating)
```

This only works if startAnimating matches in signature to the Observer class SAM. I'm not sure it will in which case we should just keep the current code. If it does, the above is a bit cleaner.

BenHenning:
Thanks @JishnuGoyal. Sent a follow-up comment--happy to defer to your decision here.

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
JishnuGoyal: I'm getting a type mismatch error while doing this. Lets go with the current implementation.

JishnuGoyal:
Thanks Ben! I have left a reply on the last review thread. I suggest going with the current implementation itself since the function signature doesn't seem to match. PTAL @BenHenning.

app/src/main/java/org/oppia/android/app/customview/ContinueButtonView.kt
BenHenning: Sounds good--thanks for checking!

BenHenning (approved):
Thanks @JishnuGoyal. LGTM.

PR #4529, opened 2022-08-25
Fix #4519, #4522: Add concept card link support to hints and concept cards

BenHenning:
## Explanation
Fixes #4519
Fixes #4522

This PR expands concept card link support to hints/solutions, as well as other concept cards. Since both cases are dialogs, the behavior is that the newly opened concept card dialog "stacks" on top of the existing, open dialog (i.e. closing the concept card will show the original dialog from which it was opened).

Note that one necessary behavior change here was making only the hint & solution _headers_ clickable, rather than the whole card, when collapsing the hint/solution. This was done intentionally to avoid potential user confusion that could arise from the fact that the now-clickable description text stops that text view from being clickable (so part of the card, when tapped, wouldn't collapse).

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [ ] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
This PR introduces new UI flows for hints/solutions and concept cards, but the actual UI/UX hasn't changed for the app. To keep things simple, here's a video showing the new behavior:

TODO

The only directly affected Espresso test is ConceptCardFragmentTest. Here are its results:

TODO

Issue #4093, opened 2022-01-13
Implement image zooming functionality


BenHenning:
### Is your feature request related to a problem? Please describe.
As noted in #4027, many images in the app (concept cards, revision cards, questions, explorations, etc.) are too small on certain devices.  
Users with visual impairments or smaller screens cannot view details properly.  
Currently, there is no way to zoom into these images, which limits **accessibility** and **usability**.

#### Screenshots : 
 ![Screenshot_20230125-173248](https://user-images.githubusercontent.com/10575562/214553015-6313bb1d-d956-444f-9aa0-345b5cc36ff6.png)
 
 ![Screenshot_20230125-173403](https://user-images.githubusercontent.com/10575562/214553020-bc68a313-f0b8-436a-ab2e-932e6a745e10.png)
 
 ![image](https://user-images.githubusercontent.com/12983742/157579759-806dc27d-5b52-4712-9ac7-7d4f5d2d2a85.png)
 
 ![image](https://user-images.githubusercontent.com/12983742/157579913-b40371cb-781b-40f5-bb82-732a142986cf.png)
 
 ![image](https://user-images.githubusercontent.com/12983742/157575930-f58b67a7-6f5b-416d-9fd8-ab52d479ab8f.png)

### Desired Behavior
- Support **pinch-to-zoom** and **panning** on images across all relevant UI components.  
- Add **double-tap zoom** (zoom in/out).  
- Provide a **reset mechanism** (e.g., double-tap reset or a small "reset" button).  
- Ensure zooming does not interfere with other gestures (scrolling in RecyclerViews/ScrollViews).  
- Work smoothly across different screen sizes and orientations.  
- Preserve **accessibility support** (TalkBack, content descriptions).

### Alternatives Considered
- Simply resizing images to always fit the screen (#4027 (comment)) – not ideal for images with text or fine details.  
- Relying on OS/device zoom features – not always available or consistent.

### References / Possible Implementation Approaches
1. **Use a third-party library like PhotoView**  
   - Simplifies pinch-zoom, double-tap, and panning.  
   - [How to implement a pinch-to-zoom feature on an ImageView (StackOverflow)](https://stackoverflow.com/questions/47415681/how-to-implement-a-pinch-to-zoom-feature-on-an-imageview?utm_source=chatgpt.com)  
[...]

PR #4846, opened 2023-01-19
Fix #4833, #4834, #4835, #4838, #1050, #4519, #4522, #4837, #4836, #4855, #4856: Assorted alpha MR6 fixes

### LINKED ISSUE #1050: Solutions are not displayed correctly except for text and fraction input types

### LINKED ISSUE #4519: Concept card link is not showing up in hints.
**Describe the bug**
Content issue is seen for one of the hints in a question in Multi digit multiplication Part 1

**To Reproduce**
1. Go to 'Multiplication'
2. Click on 'Multi-Digit Multiplication, Part 1'
3. Click on Continue till you reach the question 'There are 3829 people in our small town'
4. Enter wrong answers to get 5hints

**Expected behavior**
If you need help adding numbers with carrying, take a look at this <link> for a refresher. should be seen

**Demonstration**

**Environment**
Device name: One plus Nord2 5G
Android version : Android 11
App version : 0.8-alpha-091b45a188

### LINKED ISSUE #4522: Concept card link is not showing up in concept cards (possibly revision cards, too).
**Describe the bug**
Concept card links do not show up in concept cards.

**To Reproduce**
1. Go to 'Percentages'- Open What is a Percentage
2. Click on 'add', subtract

**Expected behavior**
There should be a concept card after "(i.e., changing it to its ..."). This indicates a missing concept card.

**Note**:
- This is likely related to #4519.
- Additionally, for safety, please check revision cards as well -- they should support opening concept card links too. (Given that this was missed in hints and concept cards, it seems possible that it was missed there too.)

**Demonstration**

**Environment**
Device name: One plus Nord2 5G
Android version : Android 11
App version : 0.8-alpha-091b45a188

### LINKED ISSUE #4833: Update all hint/solution wordings to use 'show' instead of 'reveal'

### LINKED ISSUE #4834: Introduce new voiceover 'pause' learner analytics event
Should not be logged when autoplay stops.

### LINKED ISSUE #4835: Update voiceover learner analytics events to include language codes
Format for language codes is unimportant, but should be consistent across events. Probably prefer to use the content language code.

### LINKED ISSUE #4836: Nice-to-have: introduce easy in-lesson language switching between English and Swahili
This is for the upcoming Kenya user study. Should be gated behind the study's platform parameter, or a separate one.

### LINKED ISSUE #4837: Nice-to-have: remove extraneous spaces from Swahili strings

### LINKED ISSUE #4838: Introduce support for admins to mark lesson completions for profiles
This essentially exposes the part of the developer options menu that allows for easy lesson setting. This should be gated behind the learner study parameter and **only** be visible and accessible to admins.

Consider introducing a confirmation prompt since it's an irreversible action.

### LINKED ISSUE #4855: Bullet points show up weirdly in the app
**Describe the bug**
Bullet points seem to show up misaligned in the app.

**To Reproduce**

I only know how to reproduce this on the alpha-kenya build. Here are screenshots from a few places:

(a) Multiplication > Revision Card > Techniques of Multiplication
Screenshot_20230125-172138

(b) Multiplication > Lesson 1 > Final Card

Screenshot_20230125-172708

**Expected behavior**
The inner white bullet points should be closer to their text.

**Environment**
- App version (you can get this through system app settings or via the admin controls menu in-app): 0.10-alpha_kenya

### LINKED ISSUE #4856: The images in revision cards are bigger than expected.
**Describe the bug**
All the images in one of the Multiplication revision cards on the app are much bigger than expected. (Web page for reference: https://www.oppia.org/learn/math/multiplication/revision/multiplication-techniques).

It appears that the issue is that the Android app is using the size dimensions in the SVG file to calculate the correct size for the image, but on Web the sizing is actually determined by the filename. On Android we should use the filename sizes too (scaled appropriately to account for display density and expected distance from the user's eyes).

**To Reproduce**

In the production version of the app, go to Multiplication > Revision card > Techniques of multiplication.
Screenshot_20230126-120923

**Expected behavior**
The images in this card should be smaller.

**Environment**
- App version: 0.10-alpha_kenya

BenHenning:
## Explanation
Fixes #4833
Fixes #4834
Fixes #4835
Fixes #4838
Fixes #1050
Fixes #4519
Fixes #4522
Fixes #4837
Fixes #4836
Fixes #4855
Fixes #4856

This PR introduces a variety of fixes and user quality-of-life improvements in preparation for the start of the upcoming multi-week user research project in Kenya. This PR covers 9 different issues in one shot to keep the PR management simpler due to the time sensitive nature of these changes, so it's on the larger side. Please note that:
- These changes have been released to a user study specific release track since the research project is currently ongoing.
- Many of these changes will impact users outside of the study, as detailed below.

The specifics of the PR will be split across the different issues being addressed, in categories below.

### Improvements to solutions

#4833 needed to be addressed because the previous terminology ("reveal") has been found to be confusing to learners who might not yet have a strong grasp on English. This string update will, unfortunately, not be translated for the upcoming study (though the study predominantly uses English). This change affects all users of the app.

#1050 is a long-standing issue wherein solution answers would only show up correctly if they were either text or a fraction.  This PR fixes the issue by introducing generic answer support, expanding it to include: numeric input, numeric expression, algebraic expression, algebraic equation, and ratio expression. The only missing interaction that web supports is drag & drop, but this has been found to not be needed for the current shipped list of topics (as now enforced by an update to the content import pipeline) and introducing support for it would nontrivially increase the complexity of that side of the PR. This is something that, like other interactions, will be added in the future once needed. This fix will affect all users of the app.

### Improvements to voiceover telemetry

#4834 was addressed by introducing a new event to track explicit voiceover pauses (i.e. those that require clicking the 'pause' button and not the ones that are automatic, such as autoplay ending or the user navigating away from the lesson player). Because of the nature of the 'play voiceover' event (that it is logged both for clicking the 'play' button and when expanding the audio bar due to autoplay), it's possible for the play & pause event counts to not be equal. The intent behind these events is to track explicit user actions, not the occurrences of voiceovers playing/not playing. This event will be logged for all users, though sensitive identifiers will only be logged if the user study feature is enabled.

#4835 was addressed by updating both the existing 'play voiceover' and new 'pause vocieover' events to include the lesson player-reported language code (that is, the language code originating from Oppia web). We may change this to be something more strongly ensured (since Oppia Android is a bit more careful in its language code management due to Android complexities), but the code should be generally usable as-is for data purposes as it's implemented. This change will affect telemetry for all users of the app.

### Improvements to concept card availability

#4519 and #4522 were addressed by adding support for opening concept cards both in concept cards themselves, and from hints. As of this PR, just about all cases of lesson HTML rendering should now include support for linkifying and opening concept card references. Since both hints and concept cards are dialogs, the behavior is that the newly opened concept card dialog "stacks" on top of the existing, open dialog (i.e. closing the concept card will show the original dialog from which it was opened). This change affects all users of the app.

### Miscellaneous language improvements

#4837 was addressed to clean up a variety of extraneous spaces & newlines that were unintentionally added by translators for Swahili strings. These have a direct user benefit but are otherwise innocuous, non-semantic changes. This fix technically affects all users of the app, but only those who use the Swahili translations will actually interact with the changed strings.

### Study-specific improvements

#4838 was addressed by introducing a new link in the profile edit flow for administrators to access the previously developer-only flow for manually marking certain chapters as completed. This newly exposed mechanism allows study facilitators to set up profiles so that learners can begin at a specific point (which is particularly helpful if the data on the device ever needs to be reset, such as in the case where the administrator forgot their PIN). This change will only affect user study administrators as it's locked both behind the admin account and the learner study flag. The functionality *will* be available on production-optimized builds (that is, it won't be restricted to developer-only build flavors of the app).

#4836 was addressed by introducing a floating button within the lesson player for lesson cards whose content have available Swahili translations to quickly switch the content language between English and Swahili. This doesn't change app strings, only content strings, and only temporarily for the lifetime of that specific play session (that is, if they exit the lesson & return immediately they will still be in the switched language until the app itself restarts). This button only appears if the study feature is enabled and has been enabled for that user by the administrator profile (via a new setting on the "Edit Profile" screen). This button also has its own new event to track its usage.

A "share IDs" button was added to make it easier to share the device & learner IDs by collating the IDs into a single blob of text that can be shared to any app that accepts text intents. This is meant to help reduce potential error by study coordinators.

### Content display improvements

#4855 was addressed by changing a lot of how custom LI/OL tags are handled. Specifically, Android will automatically shift the leading span of cascading bullets/numbers, but the previous implementation was using the wrong value for each nested span's parent's leading margin (which is needed to adjust what leading margin to use for the child span). This was fixed by introducing more thorough tracking (which also included a bit of robustness in ensuring larger numbers can be aligned correctly). This doesn't quite match HTML rendering, but it's much closer. Bullet lists were also updated to use squares in addition to filled and open circles (for parity with web HTML), and now uses sizes & spacing that are a bit cleaner to see.

There's still more work to improve edge cases for bullet list rendering (see #4859).

#4856 was addressed by introducing a separate SVG "render size" that, unlike the intrinsic size, is based on the image filename. Since the image filename is given in pixels, there needs to be a translation to dp to ensure consistent viewing across devices. This computation has been implemented with three calculation passes:
1. Convert the image's target pixel size to "Oppia independent pixels" by using @seanlip's monitor display density as the basis.
2. Convert "Oppia independent pixels" into Android density-independent pixels.
3. Adjust the final value using a scalar (since the physical size of an image on @seanlip's display is too large for a small mobile screen, so all images are consistently further scaled down).

There's still more work to improve image rendering (see #4093).

### General technical details

Design choices & other details of note:
- The chapter completion flow was updated to support configuring a confirmation dialog (since the action is permanent). This confirmation only shows for the admin edit profile flow (the developer options menu version still does not show a confirmation).
- The hints & solution data flow was completely redone in the app layer since much of what it was doing before isn't actually necessary with ``HelpIndex`` as it greatly reduces the complexity of managing UI-side hint state. Since solutions were being updated, anyway, this unrelated complexity has been reduced (and other issues like directly constructing an injectable view model were addressed).
- Solution titles were previously set to their content ID (similar to hints before this was recently fixed), so this was addressed by using a new "Solution" string.
- Displayed solution answers didn't correctly differentiate between exclusive & example answers. This has been corrected in text (along with how the app generally concatenates the text as the previous approach wasn't very RTL friendly).
- Generalizing solutions was done by leveraging an InteractionObject for Solution's correct_answer field. No migration is being done here since assets are replaced across releases.
- There are a lot of changes to both json and textproto assets to accommodate the new correct_answer structure, among other various changes. A few of these were done manually, but most were generated (including new Swahili translations to allow local testing of the in-lesson language switch button).
- A bug was noticed & fixed wherein states with only a solution (and no hints) would never show the solution.
- Due to the concatenation strategy for solution correct answer text (see a few points above), the horizontal textual alignment was a bit off. The new approach addresses this (though LaTeX is still a bit off, but it's generally off everywhere in the app).
- To make the new in-lesson switch language button a bit more consistent and aesthetically pleasing, this PR repurposes the "Start over" button that currently shows on the checkpoint continue screen by making it a generic "secondary button" style (with corresponding colors) and using that both for the start over button and this new secondary button. Per the screenshots below, this seems to work well. The secondary button may also be used elsewhere in the app in the future.
- Note that some of the work of this PR originates from #4529, so it can be considered a replacement for that PR.
- Only the first test topic was updated to include Swahili translations to help decrease the size of the PR (plus, it's convenient to have non-Swahili lessons to verify cases when the switch button does **not** show up).
- The PR fixes an incidental issue with the hints & solution dialog whereby clicking any part of an expanded hint/solution card would collapse it (rather than just the header or collapse icon). This would make clicking links more difficult, and is just generally a likely unexpected behavior from users.
- The new event log & language code property have been manually verified as logging correctly using the developer Firebase project with debug view enabled.
- A new event was added to track when users leave revision cards.
- A bug was found & fixed that led to tab switches in the topic activity not having their corresponding events properly logged.
- All events have been updated to now include app string, content string, and voiceover languages per-profile for that event (which led to a fairly large amount of complexity around providing access to those languages at a low level like ``AnalyticsController``).
- A bug was found & fixed around session ID management: previously, the session ID was supposed to be reset upon a new lesson being started but it actually wasn't behaving this way. Per the new needs of analytics, this behavior was changed & fixed to reset upon profile login (so that behaviors for a single user session can be properly correlated).

Noteworthy test & exemption changes:
- ``StateRetrieverTest`` is largely not updated since all of the JSON changes in this PR only affect local development (all Bazel tests & production builds use textproto or binary proto versions of lessons). The tests locally passing plus some basic local ad hoc testing is considered sufficient for the JSON loading pipeline as it will, eventually, go away in favor of textproto (once Gradle is removed since textproto -> binary proto conversion is not easily implemented in the Gradle build environment).
- Some of the question test suites weren't updated since questions aren't currently enabled, and much of the question pipeline is already covered under state fragment tests. It's likely that the testing matrix for questions will need to be largely refined once they're ready to be enabled, so some of this work was skipped in favor of expediency.
- Some tests in StateFragmentTest & StateFragmentLocalTest are disabled and can't easily be verified because they would only pass on Espresso and Bazel (due to Robolectric having limitations in language resource handling & Gradle builds not supporting the app's language configurations). Since Espresso tests do not yet work in Bazel, these tests will be verified later (though they have been manually verified as part of development).
- A bunch of KDoc validity exemptions were removed since those classes are now, as of this PR, fully documented.
- The test_file_exemptions updates are due to HintsViewModel being split into two new view models, and the general convention on the team to not directly test view models.

## Essential Checklist

### Screenshots showing some of the new functionality
| **Scenario**                                 | LTR      | LTR       | LTR      | LTR       | RTL      | RTL       | RTL      | RTL       |
|----------------------------------------------|----------|-----------|----------|-----------|----------|-----------|----------|-----------|
|                                              | Handset  | Handset   | Tablet   | Tablet    | Handset  | Handset   | Tablet   | Tablet    |
|                                              | Portrait | Landscape | Portrait | Landscape | Portrait | Landscape | Portrait | Landscape |
| Fractions solution (exclusive)                          |      image    |      image     |   Skipped until requested     |      Skipped until requested    |   image     |   image           |     Skipped until requested     |    Skipped until requested       |
| Numeric solution (exclusive)                            |   image       |      image       |       Skipped until requested      | Skipped until requested         | image    |  image         |     Skipped until requested     |        Skipped until requested   |   
| Ratio solution (not exclusive)                              |   image       |       image    |      Skipped until requested       |      Skipped until requested    |  image   |       image    |       Skipped until requested         |      Skipped until requested     |
| Text solution (exclusive)                               |     image     |      image     |    Skipped until requested       |      Skipped until requested    | image      |     image      |    Skipped until requested      |         Skipped until requested  |         
| Numeric expression solution (not exclusive)                  |   image       |       image    |    Skipped until requested      |   Skipped until requested       |    image    |     image      |   Skipped until requested       |     Skipped until requested      |          
| Algebraic expression solution (not exclusive)                |   image       |       image    |        Skipped until requested  |    Skipped until requested      |   image     |   image        |        Skipped until requested  |        Skipped until requested   |         
| Math equation solution (not exclusive)                       |    image      |       image    |   Skipped until requested       |    Skipped until requested      |  image      |       image    |      Skipped until requested    |       Skipped until requested    |          
| Switch English to Swahili                    |     image     |     image      |  Skipped until requested        |     Skipped until requested     |  image      |      image     |   Skipped until requested       |  Skipped until requested         |          
| Switch Swahili to English                    |      image    |      image     |    Skipped until requested      |     Skipped until requested     |   image     |     image      |     Skipped until requested     |    Skipped until requested       |          
| Mark chapters as completed (in profile edit) |     image     |       image    |    image      |      image    |  image       |   image        |          Skipped until requested |    Skipped until requested       |          

A couple things to note:
- Switching from English to Swahili & back is a bit awkward in the RTL screenshots (partly because the RTL layout is maintained despite switching the content language). This isn't actually a realistic scenario to occur, but it's technically possible so demonstrating it above still makes sense. There are no plans to refine this flow.
- The landscape screenshots for the profile edit screens clearly show a broken "Profile Deletion" button placement. This is unfortunately an issue already on develop, and #4852 has been filed to track this.

### Videos demonstrating different changed/new flows
| Scenario                                           | Video demonstration |
|----------------------------------------------------|---------------------|
| Switch language in-lesson                          |      https://user-images.githubusercontent.com/12983742/213829720-075cbb64-cc5a-4a08-ab3b-f6d123650fac.mp4               |
| Mark chapters completed                            |         https://user-images.githubusercontent.com/12983742/213829850-e0e157c1-520b-4ff1-86f8-470eb8c37536.mp4            |
| Open concept card from hints & other concept cards |             https://user-images.githubusercontent.com/12983742/213829894-28228cf9-4c0a-4eca-a617-325676b8d538.mp4        |
| Accessibility flow for solutions (fractions, numeric, ratios, text)                   |             https://user-images.githubusercontent.com/12983742/213830378-a56235c2-bc78-4c1f-9248-9531ae0f024d.mp4        |
| Accessibility flow for solutions (numeric expressions, algebraic expressions, math equations)                   |             https://user-images.githubusercontent.com/12983742/213830403-b964485b-9b49-43b8-9dfd-35f904c0974e.mp4        |

### Espresso test results
The following tests have been modified as a result of this PR:
- MarkChaptersCompletedActivityTest
- MarkChaptersCompletedFragmentTest
- ExplorationActivityTest
- StateFragmentTest
- ProfileEditFragmentTest
- StoryFragmentTest
- NavigationDrawerActivityDebugTest
- ConceptCardFragmentTest

I have **not** run the Espresso tests for these suites. I ran into some issues getting the suites to kick off locally, and I won't have time to investigate this for a bit due to other competing priorities. Reviewers: please let me know if you would like these run and I will prioritize accordingly (I'm leaning toward not running them since many of the Espresso tests are already failing on develop, and we have no viable way to keep them passing until we can run them in CI).


app/src/main/AndroidManifest.xml
```diff
@@ -18,7 +18,7 @@
     <!-- NOTE TO DEVELOPER: You can enable debug logs for fast upload & better inspection in Firebase console using https://support.google.com/analytics/answer/7201382. -->
     <meta-data
       android:name="firebase_analytics_collection_deactivated"
-      android:value="true" />
+      android:value="false" />
```
BenHenning: Reminder to make sure this is disabled before review begins.

model/src/main/proto/oppia_logger.proto
```diff
@@ -207,8 +210,15 @@ message EventLog {
     // Defined attributes that are common among other exploration related event log contexts.
     ExplorationContext exploration_details = 1;
 
+    // Not sure if this was ever actually a value, but reserving it to ensure no incompatibilities
```
BenHenning: Maybe update this comment since it doesn't read well in-context.

BenHenning (commented):
Finished a first pass full self-review.

app/src/main/java/org/oppia/android/app/hintsandsolution/SolutionViewModel.kt
```diff
@@ -1,28 +1,266 @@
...
+      FRACTION -> correctAnswer.fraction.toAnswerString()
+      RATIO_EXPRESSION -> correctAnswer.ratioExpression.toAnswerString()
+      MATH_EXPRESSION -> when (interaction.id) {
+        "NumericExpressionInput" -> correctAnswer.mathExpression.toNumericExpressionHtml()
+        "AlgebraicExpressionInput" -> correctAnswer.mathExpression.toAlgebraicExpressionHtml()
+        "MathEquationInput" -> correctAnswer.mathExpression.toAlgebraicEquationHtml()
+        else -> error("Interaction ID not valid for math expressions: ${interaction.id}.")
+      }
+      LIST_OF_SETS_OF_TRANSLATABLE_HTML_CONTENT_IDS, SIGNED_INT, NON_NEGATIVE_INT,
+      TRANSLATABLE_SET_OF_NORMALIZED_STRING, TRANSLATABLE_HTML_CONTENT_ID,
```
seanlip, line 97: Does this work correctly for checkbox (item selection) answers? (answer would be displayed as a bullet list of html strings)

The other interaction I identified as potentially missed here was multiple choice (single-select, answer is represented as non-negative int but we should print the HTML string that it corresponds to in the answer ordering). Though actually I don't think we typically have solutions for those so it's fine to skip that one if difficult.

app/src/main/java/org/oppia/android/app/hintsandsolution/SolutionViewModel.kt
```diff
@@ -1,28 +1,266 @@
...
+      RATIO_EXPRESSION ->
+        correctAnswer.ratioExpression.toAccessibleAnswerString(appLanguageResourceHandler)
+      MATH_EXPRESSION -> when (interaction.id) {
+        "NumericExpressionInput" -> correctAnswer.mathExpression.toReadableNumericExpression()
+        "AlgebraicExpressionInput" -> correctAnswer.mathExpression.toReadableAlgebraicExpression()
+        "MathEquationInput" -> correctAnswer.mathExpression.toReadableAlgebraicEquation()
+        else -> error("Interaction ID not valid for math expressions: ${interaction.id}.")
+      }
+      LIST_OF_SETS_OF_TRANSLATABLE_HTML_CONTENT_IDS, SIGNED_INT, NON_NEGATIVE_INT,
+      TRANSLATABLE_SET_OF_NORMALIZED_STRING, TRANSLATABLE_HTML_CONTENT_ID,
```
seanlip, line 121: Ditto.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
```diff
@@ -26,9 +49,28 @@ class StateViewModel @Inject constructor() : ObservableViewModel() {
   val isHintBulbVisible = ObservableField(false)
   val isHintOpenedAndUnRevealed = ObservableField(false)
 
+  val hasSupportForSwitchingToSwahili: Boolean = enableLearnerStudy.value
+  val hasSwahiliTranslations: LiveData<Boolean> by lazy {
```
seanlip, line 56: Just FYI (may not affect this code) -- our current Swahili translation support for Multiplication is around 40% due to changes in the lessons. Wanted to make sure that what we're implementing here accounts for the case where some strings have not been translated and that Swahili would still show up as an option despite 40% being much less than 100%.

Note that the Swahili language switcher is an optional requirement overall, if handling the above correctly (in terms of not causing breakages/crashes, or the user not being able to get back to the "base" state) is difficult in any way.

model/src/main/proto/exploration.proto
```diff
@@ -172,12 +172,16 @@ message TrainingData {
...
   // Flag that is true if correct_answer is the only correct answer of the question.
   bool answer_is_exclusive = 2;
 
-  // correct_answer of the State.
-  CorrectAnswer correct_answer = 3;
+  // Represented a representation of correct_answer that was specific to strings and fractions.
+  reserved 3;
+
+  // The correct answer to the state's primary question prompt.
+  InteractionObject correct_answer = 6;
```
seanlip, line 184: Seems a weird name for a correct answer (sounds like it describes the entire interaction)...

model/src/main/proto/oppia_logger.proto
```diff
@@ -94,6 +94,9 @@ message EventLog {
       // The event being logged is related to playing a voiceover.
       PlayVoiceOverContext play_voice_over_context = 23;
 
+      // The event being logged is related to pausing a voiceover.
+      PlayVoiceOverContext pause_voice_over_context = 35;
```
seanlip: Should this be PauseVoiceOverContext?

seanlip (commented):
Thanks @BenHenning, took a pass and commented on what I could.

app/src/main/AndroidManifest.xml
BenHenning: Done in latest changes.

model/src/main/proto/oppia_logger.proto
BenHenning: Done in the latest changes.

app/src/main/java/org/oppia/android/app/hintsandsolution/SolutionViewModel.kt
BenHenning: Multiple choice & item selection aren't supported for solutions per web, same with image click. Drag & drop is, but I've intentionally left it out to help manage complexity (& I now audit imported lessons to make sure there are no such solutions; if that changes in the future, we'll have to finish adding support on Android in the same way as any of the other interactions).

app/src/main/java/org/oppia/android/app/hintsandsolution/SolutionViewModel.kt
BenHenning: Same in this case (see my reply to your earlier comment).

model/src/main/proto/oppia_logger.proto
BenHenning: No, PlayVoiceOverContext contains the necessary information for both play & pause events. I'm open to an alternative name (my concept is that it's a context for "audio playing" generally).

model/src/main/proto/exploration.proto
BenHenning: This name is per https://github.com/oppia/oppia/blob/develop/core/domain/state_domain.py#L517. I'm not inclined to deviate from web's terminology since that can increase confusion. Do you have any suggestions?

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
BenHenning: The logic is currently implemented as "if there are *any* Swahili translations, show the switch button." We could easily change it to a percentage or something else--do you have any suggestions?

model/src/main/proto/oppia_logger.proto
seanlip: VoiceoverActionContext or VoiceoverEventContext, perhaps?

model/src/main/proto/exploration.proto
seanlip: Wait, to be clear, I'm referring to "InteractionObject", not "correct_answer" (I'm fine with the latter). Previously this was typed as CorrectAnswer ... not sure why it changed?

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
seanlip: I think the best would be to show the switch button on a card-by-card basis if the *content* part of that card has a Swahili translation. If the user subsequently navigates to a card whose content is not translated then the switch button should not be shown and the display should be in English. Likewise if the user navigates to a card whose content has a Swahili translation then the switch button should be shown.

I know there's a gap here where the content may be translated but e.g. the hints/solutions aren't, but I think that that is OK to live with because the major issues with comprehension tend to stem from students not understanding the question they are being asked.

On the other hand, if the switch button is at the lesson level and cannot easily be made into a per-card thing, then I would suggest removing it entirely, because otherwise you'll get students reading the card in English while the button says that the card is in Swahili, and this would be confusing.

BenHenning (commented):
Finished a second self-review. Didn't notice anything worth fixing this time around.

BenHenning (commented):
Self-reviewed the latest changes. I've now reviewed everything & haven't found anything of concern. CI checks are also passing now.

BenHenning:
PTAL @seanlip for write-access (all files).

PTAL @rt4914 for codeowners (I think you can mainly focus on XML files here since there aren't many layout chnages, and to help minimize what you need to review).

PTAL @MohitGupta121 for layout changes, particularly around colors.

Edit: Note that the PR description is almost finished, but I'll be finishing it up over the next few hours (since it can be time consuming to collect the screenshots, videos, and Espresso test results). Please don't block on these unless you need to. I will re-request a check from you @seanlip on the screenshots/videos once they're available in case you have any product feedback for the updated UX.

MohitGupta121:
@BenHenning I have to review layout according to dark mode color patterns?

BenHenning:
> @BenHenning I have to review layout according to dark mode color patterns?

@MohitGupta121 I'm hoping you can take a pass on the non-string XML and make sure that they're in-line with our new color practices. There shouldn't be any dark mode-specific changes here since the new secondary button style is leveraging the same dark mode behaviors as the "start over" button that it previously was.

BenHenning:
The PR description is as complete as I'm willing to get it right now. There are still some outstanding bits I need to address from @seanlip before I deassign myself.

Otherwise, @seanlip PTAL at the videos & screenshots if you want to get an idea of the UI/UX changes introduced here.

seanlip (commented):
Thanks @BenHenning -- I reviewed the full description, images and videos. Thanks for the comprehensive writeup!

For images, they all look good, though I noticed two things:
- "Solution" in the header isn't translated in Arabic. However on translatewiki it seems like the [Arabic translations for the app are complete](https://translatewiki.net/w/i.php?title=Special:Translate&group=oppia-android-app&language=ar&filter=%21translated&action=translate), so just checking that something isn't off here (at the very least we might want to file a fix for the future if that string isn't translatable for some reason).
- Perhaps the "Switch Language to Swahili" button should be "Badilisha lugha hadi kiswahili" instead?

For the videos:
- One mild suggestion for the chapter completion confirmation dialog is to add a paragraph break before "Note that this is irreversible..." so that the warning stands out more.

app/src/main/AndroidManifest.xml
```diff
@@ -18,7 +18,7 @@
     <!-- NOTE TO DEVELOPER: You can enable debug logs for fast upload & better inspection in Firebase console using https://support.google.com/analytics/answer/7201382. -->
     <meta-data
       android:name="firebase_analytics_collection_deactivated"
-      android:value="false" />
+      android:value="true" />
```
seanlip: Just to check ... this is correct?

Also the expiration date a few lines below says "2020-09-01" ... is that correct? Seems old.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
```diff
@@ -98,16 +98,13 @@ class StateViewModel @Inject constructor(
...
-    hasEnabledSwahiliTranslations: Boolean
-  ): Boolean {
+  fun canQuicklyToggleBetweenSwahiliAndEnglish(hasSwahiliTranslations: Boolean): Boolean {
     // This logic has to be done in Kotlin since there seems to be a bug in the generated Java by
     // the databinding compiler that can result in a NPE being thrown in code that shouldn't
     // actually be throwing it (see https://issuetracker.google.com/issues/144246528 for context).
     // Essentially, the following example of generated code results in an NPE unexpectedly:
     //   Boolean value = boolean_value ? Boolean_value : false (Boolean_value can be null)
-    return hasSwahiliTranslations && hasEnabledSwahiliTranslations
+    return hasSwahiliTranslations && hasSupportForSwitchingToSwahili
```
seanlip: This doesn't seem to be passed in the method signature?

seanlip (commented):
(I just looked at the diff.)

app/src/main/res/layout/hint_summary.xml
```diff
@@ -12,7 +12,7 @@
```
MohitGupta121, line 12: I think instead of :-`app:cardBackgroundColor=component_color_hints_and_solution_summary_card_background_color` use `app:cardBackgroundColor=component_color_shared_screen_primary_background_color`


`component_color_hints_and_solutions_fragment_text_color` to `component_color_shared_primary_text_color`


`component_color_hints_and_solution_summary_button_text_color`   to  `component_color_shared_secondary_4_text_color`


`component_color_hints_and_solution_summary_divider_color`  to  `component_color_shared_divider_color`


`component_color_hints_and_solutions_fragment_description_color`  to  `component_color_shared_primary_text_color`

app/src/main/res/layout/hints_and_solution_fragment.xml
```diff
@@ -7,7 +7,7 @@
 
     <variable
       name="viewModel"
-      type="org.oppia.android.app.hintsandsolution.HintsViewModel" />
+      type="org.oppia.android.app.hintsandsolution.HintsAndSolutionViewModel" />
   </data>
 
   <androidx.constraintlayout.widget.ConstraintLayout
```
MohitGupta121, line 13: Instead of this we can use :- 

`android:background="@color/color_def_white"` to `android:background="@color/color_palette_primary_background_color"`


Also now no need of using `app:titleTextColor="@color/color_def_white"`  in `<androidx.appcompat.widget.Toolbar`  as it is already define in `app:titleTextAppearance="@style/ToolbarTextAppearance"`

app/src/main/res/layout/solution_summary.xml
```diff
@@ -130,32 +132,15 @@
           android:paddingBottom="16dp"
           android:visibility="@{isListExpanded? View.VISIBLE : View.GONE}">
 
-          <LinearLayout
-            android:layout_width="match_parent"
+          <TextView
```
MohitGupta121, line 135: I think instead of :-`app:cardBackgroundColor=component_color_hints_and_solution_summary_card_background_color` use `app:cardBackgroundColor=component_color_shared_screen_primary_background_color`


`component_color_hints_and_solutions_fragment_text_color` to `component_color_shared_primary_text_color`


`component_color_hints_and_solution_summary_button_text_color`   to  `component_color_shared_secondary_4_text_color`


`component_color_hints_and_solution_summary_divider_color`  to  `component_color_shared_divider_color`


`component_color_hints_and_solutions_fragment_description_color`  to  `component_color_shared_primary_text_color`

app/src/main/res/layout/state_fragment.xml
```diff
@@ -99,57 +99,87 @@
...
-          app:layout_constraintBottom_toBottomOf="parent"
-          app:layout_constraintStart_toStartOf="parent"
-          app:layout_constraintTop_toTopOf="parent"
-          app:srcCompat="@{viewModel.isHintOpenedAndUnRevealed() ? @drawable/ic_hint_bulb_filled_yellow_48dp : @drawable/ic_hint_bulb_white_48dp}" />
-
-        <TextView
-          android:id="@+id/hint_bar_title"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintBottom_toTopOf="@+id/animable_hints_and_solution_fragment_container"
+        android:background="@color/color_palette_transparent_background_color"
```
MohitGupta121: I think this is by mistake but here we can use `component_color_shared_transparent_background_color` in replace of `color_palette_transparent_background_color`



Also in this file please format code by code format shortcut in android studio.

app/src/main/res/values/component_colors.xml
```diff
@@ -22,6 +22,7 @@
   <color name="component_color_shared_previous_responses_header_item_container_background_color">@color/color_palette_highlighted_text_color</color>
   <color name="component_color_shared_transparent_background_color">@color/color_palette_transparent_background_color</color>
   <color name="component_color_shared_white_background_color">@color/color_palette_shared_white_background_color</color>
+  <color name="component_color_shared_secondary_button_background_color">@color/color_palette_secondary_button_background_color</color>
```
MohitGupta121: We already have `color_palette_primary_button_background_color` so we not need `component_color_shared_secondary_button_background_color` extra color. 
What you think @BenHenning ?

app/src/main/res/values/styles.xml
```diff
@@ -410,21 +410,23 @@
...
     <item name="android:paddingEnd">12dp</item>
     <item name="android:paddingStart">12dp</item>
     <item name="android:minWidth">144dp</item>
-    <item name="android:background">@drawable/start_over_button_background</item>
+    <item name="android:background">@drawable/secondary_button_background</item>
     <item name="android:fontFamily">sans-serif-medium</item>
-    <item name="drawableStartCompat">@drawable/ic_start_over_24dp</item>
     <item name="android:drawablePadding">4dp</item>
-    <item name="drawableTint">@color/component_color_shared_resume_lesson_start_over_button_color</item>
+    <item name="drawableTint">@color/component_color_shared_secondary_button_background_color</item>
```
MohitGupta121: As suggest in above comment we can use `color_palette_primary_button_background_color` instead of this `component_color_shared_secondary_button_background_color`

app/src/main/res/drawable/secondary_button_background.xml
```diff
@@ -5,5 +5,5 @@
   <solid android:color="@color/component_color_shared_white_background_color" />
   <stroke
     android:width="2dp"
-    android:color="@color/component_color_shared_resume_lesson_start_over_button_color" />
+    android:color="@color/component_color_shared_secondary_button_background_color" />
```
MohitGupta121: Same as per above comment use `color_palette_primary_button_background_color` here also.

MohitGupta121 (changes requested):
@BenHenning PTAL, These are some suggestions from my side.

rt4914 (approved):
Other than what is suggested by @MohitGupta121  and https://github.com/oppia/oppia-android/issues/4852
everything else looks good. 

Approving as Mohit has requested changes.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
BenHenning: The latest changes include the suggestion here to gate the button behind a per-card content translation check.

model/src/main/proto/exploration.proto
BenHenning: Ah I see. InteractionObject is an existing object that represents any rule input/answer types for an interaction.

We have better typing coming with the Oppia proto API.

model/src/main/proto/oppia_logger.proto
BenHenning: Went with ``VoiceoverActionContext`` in the latest changes--thanks for the suggestion!

app/src/main/AndroidManifest.xml
BenHenning: Yep, deactivated is set to true (so it's off).

App expiration is off, we set it temporarily when the build is created (as there's not much value in checking it into develop).

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
BenHenning: It's a class val in StateViewModel.

app/src/main/res/layout/hint_summary.xml
BenHenning: Since these seem largely unrelated to the PR, shouldn't we maybe do them separately?

app/src/main/res/layout/hints_and_solution_fragment.xml
BenHenning: These changes also seem unrelated to the PR.

app/src/main/res/layout/solution_summary.xml
BenHenning: It seems like the only part that's relevant here is component_color_hints_and_solutions_fragment_description_color. What's the thinking behind moving toward shared colors (vs. having more per-component colors)?

app/src/main/res/layout/state_fragment.xml
BenHenning: Done (to both).

app/src/main/res/values/component_colors.xml
BenHenning: I think these are conceptual different things (and your comment actually made me realize that this isn't just a 'background' color, it's actually a trim color that affects multiple styling components for the button), so I'm not sure that it makes sense to merge them.

app/src/main/res/values/styles.xml
BenHenning: See my earlier comment.

app/src/main/res/drawable/secondary_button_background.xml
BenHenning: See my earlier comment.

BenHenning (commented):
Thanks @seanlip @MohitGupta121 @rt4914 

I've addressed comments & wrapped up the remaining work code-wise. I still need to summarize the major changes from the last few commits & update the PR description, but I'll be doing that later this week (since the priority is getting this code shipped).

I've self-reviewed all changes up to this point, as well.

app/src/main/res/layout/solution_summary.xml
MohitGupta121: Actually, the main thinking behind moving this to shared colors is that while me and @rt4914 working on the dark-mode so we noticed there are so many commonly used colors. And to reduce Duplicacy and make the color file more short and easy to find the exact color. Before the shared colors, there is 345 lines and now it is reduced to 236. 

But it's okay actually we revisiting all the layouts and xml files and changing colors according to that. So for now this PR is not related to that.

MohitGupta121 (approved):

app/src/main/res/layout/hint_summary.xml
MohitGupta121: Yes, I working on this as I separately visiting all the xmls and layouts files and change according to that. So no need to change in this PR. Thanks.

app/src/main/res/layout/hints_and_solution_fragment.xml
MohitGupta121: Yes, same as the above comment. Sorry.

app/src/main/res/values/component_colors.xml
MohitGupta121: Yes looks good. We keep these two separate.

BenHenning:
PTAL @seanlip for the final follow-up changes (as explained in the latest changes to the PR description in 2 recent edits). I've self-reviewed this code, and recently worked through multiple areas (for analytics and the learner study screen) as part of #4896.

@MohitGupta121 PTAL as well since I think you have open conversation threads (please verify & mark them as resolved if you're happy with the latest changes).

MohitGupta121 (approved):
@BenHenning LGTM, 
I think I not see "mark resolved" in conversations due to permission issue. So you can mark them as resolved.
Thanks.

BenHenning:
No problem, thanks @MohitGupta121!

@seanlip since this is set to auto-merge, I'm going to wait to resolve @MohitGupta121's comments until after you approve (or, you're welcome to resolve them after approving if you'd like to merge the PR).

seanlip:
Whoops, I didn't realize resolving the comments would result in the PR getting merged.

That said, I'm fine with merging. I read through the comment updates but I didn't look at the code carefully, and the edited description looks fine to me. Thanks!

PR #4851, opened 2023-01-20
Fix #4850: Bump version codes for alpha 0.10 release, and fix version code ordering

### LINKED ISSUE #4850: Bump version codes for upcoming alpha releases, plus fix their order
This issue tracks two things:
1. The version codes need to be incremented for the upcoming release of 0.10 alpha
2. The version codes need to be ordered such that newer flavors take priority over older flavors (so that users get the alpha version of the app instead of beta for the same version name)

BenHenning:
## Explanation
Fixes #4850

This PR introduces the necessary version code updates for an upcoming 0.10 alpha release (2 different RCs). It also fixes the order of version codes such that less stable versions of the app (i.e. build flavors) take priority over more stable ones (so that users signed up for an earlier pre-release will get that over later versions, e.g. users would receive an alpha version of the app over beta if they're in both). This happens because Play Store uses version codes as the actual comparable versions for uploaded binaries, and it picks the binary with the highest version code to be installed based on all that a user may qualify for (based on which release tracks that they're participating in). The assumption here is that newer versions maintain backward compatibility with older versions, and this is something we will continue to aim for (though with minimal guarantees for pre-release versions).

Note that one unfortunate side effect of this change is that local installs of the app will require manually uninstalling when installing a "more stable" build flavor (though technically this was an issue in the other direction previously). This does make "upgrade" flows a little trickier to test locally unless using pre-assembled binaries.

This PR also updates app versioning such that the version now includes the local branch's commit rather than the common commit with develop (to ensure RC rebuilds have properly unique version names rather than sharing one version name for the release branch).

## Essential Checklist

## For UI-specific PRs only
N/A -- This is an infrastructure-only change.


BenHenning:
@seanlip PTAL for codeowners.

seanlip (approved):

seanlip:
@BenHenning Not sure if I should merge this yet, feel free to do so once you're ready. Thanks!

BenHenning (commented):
Took a self-review pass on the changes. I'll need to explain the changes in the main comment, as well, before this PR can be re-reviewed and merged.

BenHenning:
@seanlip PTAL for latest changes & PR description updates. I've also self-reviewed the latest changes, so I think this PR is ready to go if it looks good to you.

seanlip (approved):
No concerns on my end, thanks @BenHenning.

Issue #632, opened 2020-01-30
Replace current recyclerview implementation with BindableAdapter usage


rt4914:
Currently the entire project has two types of RecyclerView implementation:
1. Recyclerview using BindableAdapter (OngoingStoryListActivity, CompletedStoryListActivity, etc.) uses it)

2. Recyclerview with the RecyclerViewAdapter which inflates multiple layouts.

We need to change the implementation such that all RecyclerViewAdapter to  use BindableAdapter (i.e. approach similar to OngoingStoryListActivity, CompletedStoryListActivity).

## Checklist
- [x] `ContinuePlaying` (assigned to @veena14cs )
- [x] `HomeFragment` (assigned to @jcqli)
- [x] `StoryActivity` (Already done)
- [x] `TopicReview/Revision` (Assigned to @Sarthak2601 )
- [x] `TopicPlay/Lessons`
- [x] `TopicTrain/Practice` (assigned to @Luffy18346 )
- [x] `HelpActivity` (Assigned to @SayantanBanerjee16 )
- [x] `HintsAndSolutionAdapter`  (Assigned to @anandwana001 )
- [x] `OngoingListAdapter`  (open)
- [x] `TopicListAdapter` (assigned to @jcqli)
- [x] `LanguageSelectionAdapter` (assigned to @FareesHussain)
- [x] `StorySummaryAdapter`  (open)
- [ ] `PromotedStoryListAdapter` (assigned to @masclot )

veena14cs:
@rt4914 I will be taking up Continue Playing task in this list.

veena14cs:
@rt4914  StoryActivity already has implementation of Bindable adapter.

NullByte08:
@rt4914 Can I work on the `HomeFragment` task? I will replace both `TopicListAdapter` and `PromotedStoryListAdapter`.

rt4914:
> @rt4914 Can I work on the `HomeFragment` task? I will replace both `TopicListAdapter` and `PromotedStoryListAdapter`.

@NullByte08 I would suggest you to work on simpler one first as HomeFragment is the hardest one. Still its completely fine if you work on that.

Sarthak2601:
@rt4914 Can I work on `TopicReview/Revision` task ?

rt4914:
> @rt4914 Can I work on `TopicReview/Revision` task ?

Sure.

NullByte08:
@rt4914 Can I work on `FAQMain` task then?

[...]

Issue #3345, opened 2021-06-19
Images are not fully displayed


isalooo:
**Describe the bug**
Critical text in images is cut off. This makes it difficult for users to learn the math concepts. 

**To Reproduce**

In the place values lesson, there are multiple images are not fully displayed. 

**Expected behavior**
Images should be fully displayed. 

**Screenshots**
If applicable, add screenshots or videos to help explain your problem.
![Screenshot_20210526-055327_Oppia](https://user-images.githubusercontent.com/82181327/122643553-53c07f00-d108-11eb-95cd-00e8b12e7e14.jpg)
![Screenshot_20210526-055147_Oppia](https://user-images.githubusercontent.com/82181327/122643554-54f1ac00-d108-11eb-906f-6748d2facf2f.jpg)
![Screenshot_20210526-054559_Oppia](https://user-images.githubusercontent.com/82181327/122643556-558a4280-d108-11eb-8b46-dc0e60763689.jpg)

**Device** samsung Galaxy s9+
- Android version 8 

**Additional context**

rt4914:
@rishidyno Unassigning you from this issue because of inactivity, please re-assign yourself if you are currently working on this.

TheSwarnim:
Hi, @rt4914 I would like to work on this issue. Please assign it to me.

TheSwarnim:
Hi @rishidyno and @rt4914, could you tell me how can I produce the same behavior. I'm not able to find the place value session in the app

rt4914:
@TheSwarnim This has been explained in meeting but for reference adding it here too.

1. The issue is that images are cut-off and it can be because of two reasons (a.) the images are two big (b.) the images are the last thing in rich text. I think (b.) might be the reason because all the above screenshots show that the images are last part of rich-text.
2. The screenshots displayed above are from alpha app and therefore we cannot find them in current develop code. 
3. To reproduce them visit the "What is a Fraction?" exploration content in https://github.com/oppia/oppia-android/blob/develop/domain/src/main/assets/umPkwp0L1M0-.json and https://github.com/oppia/oppia-android/blob/develop/domain/src/main/assets/umPkwp0L1M0-.textproto and edit the `states` such that you can create a case where there are images at the bottom and there is not content after that image.

TheSwarnim:
Thanks, Rajat for your reference. It will help me a lot.

BenHenning:
For searching context, this issue was found in 0.5-alpha (MR2).

[...]

PR #4874, opened 2023-02-08
Fix part of #632: Move PromotedStoryListAdapter to BindableAdater

masclot:
## Explanation

Fixes part of #632: Reimplementation of PromotedStoryListAdapter as BindableAdater
Espresso tests are failing, but they were already failing. The errors with the new adapter are the same. The same tests pass when using gradle.

There was an error in MarginBindingAdapter which I'm also fixing in this cl: MarginBindingAdapter was replacing start/end margin changes for left/right margin changes. However, this replacement relied on getting the layout direction from the View being changed. When the direction is inherited and the View is not attached to a parent (recursively until top level View), the direction defaults to LTR, no matter the Locale/language. This situation happens for example in item Views within RecyclerView. The change does not replace start/end any more, but sets them through MarginLayoutParamsCompat so that it is compatible with api < 17.

## Essential Checklist

- [X] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [X] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [X] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [X] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".

| Current | new |
|-|-|
| bindable_adapter_current | bindable_adapter_new |

Unit tests:
recently_played_span_test

Espresso Tests. The run for the new version has an extra error, but it is due to timings. In other runs it passes.
recently_played_espresso_tests_current_vs_new

@rt4914 PTAL


masclot:
@rt4914 PTAL

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -282,6 +137,34 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
+        transformViewModel = { it as SectionTitleViewModel }
+      )
+      .registerViewDataBinder(
+        viewType = ViewType.VIEW_TYPE_PROMOTED_STORY,
+        inflateDataBinding = RecentlyPlayedStoryCardBinding::inflate,
+        setViewModel = RecentlyPlayedStoryCardBinding::setViewModel,
+        transformViewModel = { it as PromotedStoryViewModel }
+      )
+      .build()
+  }
```
BenHenning:
```suggestion
    }.registerViewDataBinder(
      viewType = ViewType.VIEW_TYPE_TITLE,
      inflateDataBinding = SectionTitleBinding::inflate,
      setViewModel = SectionTitleBinding::setViewModel,
      transformViewModel = { it as SectionTitleViewModel }
    ).registerViewDataBinder(
      viewType = ViewType.VIEW_TYPE_PROMOTED_STORY,
      inflateDataBinding = RecentlyPlayedStoryCardBinding::inflate,
      setViewModel = RecentlyPlayedStoryCardBinding::setViewModel,
      transformViewModel = { it as PromotedStoryViewModel }
    ).build()
  }
```

Nit: bit cleaner wrapping is possible.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -39,189 +35,48 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
       override fun getSpanSize(position: Int): Int {
-        return when (position) {
-          sectionTitle0Position, sectionTitle1Position, sectionTitle2Position -> {
-            /* number of spaces this item should occupy = */ spanCount
-          }
-          else -> {
-            /* number of spaces this item should occupy = */ 1
-          }
+        return if (adapter.getItemViewType(position) == ViewType.VIEW_TYPE_TITLE.ordinal) {
+          /* number of spaces this item should occupy = */ spanCount
```
BenHenning: This isn't actually an argument comment. While it's an existing issue, I think we should fix this. Given that it doesn't really add much context, I suggest just removing this & the next copy of the comment.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -39,189 +35,48 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
-      oldStoryCount > 0 && recentStoryCount > 0 -> {
-        recentStoryCount + oldStoryCount + 2
-      }
-      else -> {
-        recentStoryCount + oldStoryCount + 1
-      }
-    }
 
     val spanCount = activity.resources.getInteger(R.integer.recently_played_span_count)
-    promotedStoryListAdapter.setSpanCount(spanCount)
```
BenHenning, line 213: Just checking: was this not needed before with the SpanSizeLookup being provided?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+  private val topicListController: TopicListController,
+  @StoryHtmlParserEntityType private val entityType: String,
+  private val resourceHandler: AppLanguageResourceHandler,
+  private val translationController: TranslationController,
+) {
+
+  private var internalProfileId: Int = -1
+
+  /**
+   * Warning: it is required to call [setInternalProfileId] before accessing this property.
```
BenHenning: The KDoc should explain what this is first, before any such warnings.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+class RecentlyPlayedViewModel @Inject constructor(
+  private val activity: AppCompatActivity,
+  private val fragment: Fragment,
+  private val topicListController: TopicListController,
+  @StoryHtmlParserEntityType private val entityType: String,
+  private val resourceHandler: AppLanguageResourceHandler,
+  private val translationController: TranslationController,
+) {
+
+  private var internalProfileId: Int = -1
```
BenHenning: It may be cleaner to pass in a ProfileId, instead, and mark it as a lateinit var to ensure that it actually gets initialized.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Warning: it is required to call [setInternalProfileId] before accessing this property.
+   */
+  val recentlyPlayedLiveData: LiveData<List<RecentlyPlayedItemViewModel>> by lazy {
```
BenHenning: Perhaps instead: ``recentlyPlayedItems``?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -39,189 +35,48 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
     layoutManager.spanSizeLookup = object : GridLayoutManager.SpanSizeLookup() {
       override fun getSpanSize(position: Int): Int {
-        return when (position) {
-          sectionTitle0Position, sectionTitle1Position, sectionTitle2Position -> {
-            /* number of spaces this item should occupy = */ spanCount
-          }
-          else -> {
-            /* number of spaces this item should occupy = */ 1
-          }
+        return if (adapter.getItemViewType(position) == ViewType.VIEW_TYPE_TITLE.ordinal) {
```
BenHenning, line 78: Does this actually correctly replicate all of the cases covered in the code removed above? Are there tests to cover those cases, too?

This seems like a significant simplification (which is great), but I'm not sure that it actually handles the different cases above.

BenHenning (changes requested):
Thanks @masclot! Took a first pass--PTAL. Also, I think there are CI checks failing that should be fixed.

BenHenning:
Also, please make sure to include the 'Fixes part of #632' within the PR description.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Not needed: the method just sets a private field that is not used internally. https://github.com/oppia/oppia-android/blob/3059cc7870c521e6b52b5740e6b1982dd9fe0a2d/app/src/main/java/org/oppia/android/app/home/recentlyplayed/PromotedStoryListAdapter.kt#L23

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: I think both are equivalent in the end result. I think the new version is better (apart from less code) because it does not need to change the layout manager, which causes [very] short term discrepancy between adapter and layout manager.
There is an existing test class that covers the span count: https://github.com/oppia/oppia-android/blob/087f2fe85aa01655ad03bbebbc64ab5edd13089f/app/src/test/java/org/oppia/android/app/testing/RecentlyPlayedSpanTest.kt#L99

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Done

masclot:
I added unit tests
@BenHenning PTAL

app/src/main/java/org/oppia/android/app/home/recentlyplayed/PromotedStoryViewModel.kt
```diff
@@ -12,7 +12,7 @@ import org.oppia.android.domain.translation.TranslationController
 
 /** [ViewModel] for displaying a promoted story. */
 class PromotedStoryViewModel(
-  private val activity: AppCompatActivity,
```
BenHenning: Why is this change needed?

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -18,46 +16,62 @@ import org.oppia.android.util.parser.html.StoryHtmlParserEntityType
...
+  private val promotedStoryClickListener: PromotedStoryClickListener,
+  private val internalProfileId: ProfileId,
 ) {
 
-  private var internalProfileId: Int = -1
+  /**
+   * Factory of RecentlyPlayedViewModel.
+   *
+   * Dagger's AssistedInject would do all the wiring automatically, unfortunately it is not imported
+   * in oppia-android.
```
BenHenning: This seems more like a PR comment, not something that we'd want to keep in code long-term.

Also, I haven't used Dagger's assisted injection before. I'm not opposed to integrating it provided:
- We can make it work with both Gradle and Bazel.
- It will interoperate correctly with Dagger Hilt, since integrating with Hilt is a long-term goal.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
```diff
@@ -26,7 +26,7 @@ import javax.inject.Inject
  * instance).
  */
 class AppLanguageResourceHandler @Inject constructor(
-  private val activity: AppCompatActivity,
+  private val activity: Context,
```
BenHenning: Ditto  here. This seems incorrect--we cannot guarantee an activity is injected here without explicitly requesting that type (and this **must** be an activity for language switching to function correctly).

app/src/test/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModelTest.kt
```diff
@@ -0,0 +1,245 @@
...
+import kotlin.test.fail
+
+/** Tests for [RecentlyPlayedViewModel] data. */
+@RunWith(RobolectricTestRunner::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(
+  application = RecentlyPlayedViewModelTest.TestApplication::class,
+  manifest = Config.NONE
+)
+class RecentlyPlayedViewModelTest {
```
BenHenning: Thanks for adding this test suite!

We generally don't test view models, and I'm not sure we want to establish that pattern now, either. The main reason for this is that I think view models will largely go away (or become trivial data objects) in the future as we move toward more code generation to reduce the fairly significant amount of boilerplate needed today for writing UIs.

We can generally test view models thoroughly through their corresponding fragments/activities which is also generally more useful since we're verifying the integration of the UI component itself (and lifecycles play a pretty critical role in view models behaving correctly).

Might it be possible to repurpose these tests & move them to RecentlyPlayedFragmentTest?

BenHenning (changes requested):
Thanks @masclot! I really like the cleanup work being done here. I have some thoughts on the new test suite, and a few questions about various other changes. PTAL.

BenHenning:
Also, @masclot as FYI we generally rely on the original creator of a comment thread to resolve it (as part of verifying that the thing was addressed). I've unresolved past comment threads that you marked as resolved, so nothing to change here (but please keep this in mind for future comment threads :)).

seanlip:
HI @masclot -- just to check, are you still working on this? Thanks!

masclot:
Yes. I was out last week.

seanlip:
Ah, OK. Thanks!

app/src/test/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModelTest.kt
masclot: I added the test because one of the PR checks was complaining. It said something like "new file without test file" or something similar.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
masclot: The only reason was to be able to create robolectric tests without the burden of dealing with activities. If the test is not needed, then there is no point. I'm curious though: why does language switching does not work without activity? Can you point me to some code?

BenHenning: It has to do with how the Dagger graph is set up, actually. Injecting a context makes this handler exist 1-per-app rather than 1-per-activity, so we end up in a situation where we're using resources only bound to the application. This is an issue because language switching triggers an activity recreation via:

https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt#L90

And we change the activity's resources configuration dynamically based on the current language during activity creation:

https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/app/src/main/java/org/oppia/android/app/activity/InjectableAppCompatActivity.kt#L76

Theoretically, this can lead to using a Resources object that doesn't have the correct corresponding language. That's definitely wrong from an API perspective, but from a practical functionality perspective I'm less sure if it'd actually break in *this* specific case. Android may rely on other signals (like the default Locale object) for determining string retrieval, or perhaps it always uses the top-most activity of the process's task stack when retrieving strings. Either way, we want to keep things per-activity to increase predictability.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: I'll remove the PR part of the comment.
Re. your points, I have used Dagger Hilt with AssistedInject before with no problems. That said, I have no experience with Gradle and Bazel, so I cannot answer that part.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
masclot: Thanks for the explanation!

app/src/test/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModelTest.kt
masclot: In fact, RecentlyPlayedFragmentTest already does equivalent tests

masclot:
I fixed the MarginBindingAdapter class as explained in the last meeting. I'm attaching some screenshots of the home, select_profile and recently_played screens in RTL model.

| Current | New |
| - | - |
| Recently played ||
| recently_played_current | recently_played_new |
| Home ||
| home_current | home_new |
| Select profile ||
| select_profile_current | select_profile_new |

MarginBindingAdapters tests:
MarginBindingAdapters tests

masclot:
@BenHenning PTAL

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
```diff
@@ -14,12 +14,7 @@ public final class MarginBindingAdapters {
   public static void setLayoutMarginStart(@NonNull View view, float marginStart) {
     if (view.getLayoutParams() instanceof MarginLayoutParams) {
       MarginLayoutParams params = (MarginLayoutParams) view.getLayoutParams();
-      float marginEnd = params.getMarginEnd();
-      if (isRtlLayout(view)) {
-        setLayoutDirectionalMargins(view, (int) marginEnd, (int) marginStart);
-      } else {
-        setLayoutDirectionalMargins(view, (int) marginStart, (int) marginEnd);
-      }
+      MarginLayoutParamsCompat.setMarginStart(params, (int) marginStart);
```
adhiamboperes, line 17: Could you add an explanation in your PR description for why we made the change to `MarginLayoutParamsCompat`

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -39,199 +35,57 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
-  private val translationController: TranslationController
+  private val multiTypeBuilderFactory: BindableAdapter.MultiTypeBuilder.Factory,
+  private val recentlyPlayedViewModelFactory: RecentlyPlayedViewModel.Factory
 ) {
 
   private val routeToResumeLessonListener = activity as RouteToResumeLessonListener
   private val routeToExplorationListener = activity as RouteToExplorationListener
-  private var internalProfileId: Int = -1
+
+  private lateinit var internalProfileId: ProfileId
```
adhiamboperes: `internalProfileId: Int = -1` is usually the default value used when profileId is not initialized yet or not found, and I am concerned that we might be introducing a nullability issue by  lateinit-ing internalProfileId here.

adhiamboperes (commented):
You can format the images in the PR [current,new] next to each other by using tables. Here is [how to add table syntax in md](https://www.markdownguide.org/extended-syntax/). The outcome would be something like [this](https://github.com/oppia/oppia-android/issues/3345#issuecomment-1317551932).

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+import org.oppia.android.app.translation.AppLanguageResourceHandler
+import org.oppia.android.domain.topic.TopicListController
+import org.oppia.android.domain.translation.TranslationController
+import org.oppia.android.util.data.AsyncResult
+import org.oppia.android.util.data.DataProviders.Companion.toLiveData
+import org.oppia.android.util.parser.html.StoryHtmlParserEntityType
+import javax.inject.Inject
+
+/** View model for [RecentlyPlayedFragment]. */
+class RecentlyPlayedViewModel constructor(
```
BenHenning: We don't need to explicitly declare the constructor unless changing its visibility. That being said, if we can make this private that'd be preferred since the pattern here is to use the view model's factory to create instances of it.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+    return Transformations.map(promotedStoryListSummaryResultLiveData) {
+      when (it) {
+        // If there's an error loading the data, assume the default.
+        is AsyncResult.Failure, is AsyncResult.Pending -> PromotedActivityList.getDefaultInstance()
+        is AsyncResult.Success -> it.value
+      }
+    }
+  }
+
+  private fun processPromotedStoryList(promotedActivityList: PromotedActivityList):
```
BenHenning: You can break at the ``(`` instead to avoid the extra indentation being needed here (in the same way as ``addRecentlyPlayedStoryListSection`` below).

app/src/test/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModelTest.kt
BenHenning: Ah, we just exempt view models (longer term we really should remove the check for view models, but this is one of the things that's in a slightly temporary state for our static checks.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
BenHenning: Thanks for checking--this is excellent. Thanks @masclot!

app/src/main/java/org/oppia/android/app/home/recentlyplayed/PromotedStoryViewModel.kt
BenHenning: Looks like this was removed in the latest changes.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
BenHenning: Nice catch!

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
BenHenning: Thanks!

BenHenning (changes requested):
Thanks @masclot! Just had a few follow-up comments.

masclot:
> You can format the images in the PR [current,new] next to each other by using tables. Here is [how to add table syntax in md](https://www.markdownguide.org/extended-syntax/). The outcome would be something like [this](https://github.com/oppia/oppia-android/issues/3345#issuecomment-1317551932).

Thanks. I updated the comments as indicated.

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Changed back to -1

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Set to private

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
```diff
@@ -39,199 +35,57 @@ class RecentlyPlayedFragmentPresenter @Inject constructor(
...
-
-  private fun createOngoingStoryViewModel(
-    promotedStory: PromotedStory,
-    index: Int
-  ): RecentlyPlayedItemViewModel {
-    return PromotedStoryViewModel(
-      activity,
-      promotedStory,
-      entityType,
+    this.internalProfileId = ProfileId.newBuilder().setInternalId(profileId).build()
```
adhiamboperes: @masclot, It looks like you're building another instance of ProfileId here, but I think you should assign the variable created earlier on line 46.

adhiamboperes (changes requested):
I've left one code comment blocking approval.

Could you also PTAL @masclot at the failing script check and resolve it? Let me know if you need any help with that.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: I'm not sure I understand. I'm creating a ProfileId in order to assign to the variable created on line 46. The variable on line 46 holds a fully built proto, not a proto-builder.

Changed back to lateinit

masclot:
> I've left one code comment blocking approval.
> 
> Could you also PTAL @masclot at the failing script check and resolve it? Let me know if you need any help with that.

The failing check is due to a new class file without a matching test file. Since the new class is a ViewModel, @BenHenning  told me I didn't need to add a test file.

masclot:
PTAL @adhiamboperes

adhiamboperes:
> > I've left one code comment blocking approval.
> > Could you also PTAL @masclot at the failing script check and resolve it? Let me know if you need any help with that.
> 
> The failing check is due to a new class file without a matching test file. Since the new class is a ViewModel, @BenHenning told me I didn't need to add a test file.

For files we are exempting from tests, we add them to `test_file_exemptions.textproto`. Note that the same lexicographical ordering rule of imports applies here too.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
adhiamboperes: @masclot, tagging you here since this is the comment I was referring to on chat.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
adhiamboperes: Took another pass. This will actually work, and there's no need to change it.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
BenHenning: Nit: we also ought to call this ``profileId`` since it's not an "internal" profile ID anymore. Besides that, it should be fine for this variable to be a ``lateinit var``--no need to initialize it as @adhiamboperes suggests here (in case the most recent change was due to an old comment of mine--I'm not sure if that's the case here).

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
```diff
@@ -0,0 +1,172 @@
...
+
+/** View model for [RecentlyPlayedFragment]. */
+class RecentlyPlayedViewModel private constructor(
+  private val activity: AppCompatActivity,
+  private val topicListController: TopicListController,
+  @StoryHtmlParserEntityType private val entityType: String,
+  private val resourceHandler: AppLanguageResourceHandler,
+  private val translationController: TranslationController,
+  private val promotedStoryClickListener: PromotedStoryClickListener,
+  private val internalProfileId: ProfileId,
```
BenHenning: Nit: ``profileId``.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
BenHenning: Resolved in the latest changes.

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
BenHenning: Resolved in the latest changes.

BenHenning (changes requested):
Thanks @masclot! The PR mostly LGTM, just had a few follow-ups.

Also, please make sure to reply to all comments before sending the PR back to review (otherwise it can be difficult as a reviewer to know what you've intended to address).

BenHenning:
(Taking off Rajat since he doesn't need to review this PR).

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedViewModel.kt
masclot: Done both

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Done

app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedFragmentPresenter.kt
masclot: Changed back to lateinit

masclot:
> > > I've left one code comment blocking approval.
> > > Could you also PTAL @masclot at the failing script check and resolve it? Let me know if you need any help with that.
> > 
> > 
> > The failing check is due to a new class file without a matching test file. Since the new class is a ViewModel, @BenHenning told me I didn't need to add a test file.
> 
> For files we are exempting from tests, we add them to `test_file_exemptions.textproto`. Note that the same lexicographical ordering rule of imports applies here too.

Done

adhiamboperes:
> > > > I've left one code comment blocking approval.
> > > > Could you also PTAL @masclot at the failing script check and resolve it? Let me know if you need any help with that.
> > > 
> > > 
> > > The failing check is due to a new class file without a matching test file. Since the new class is a ViewModel, @BenHenning told me I didn't need to add a test file.
> > 
> > 
> > For files we are exempting from tests, we add them to `test_file_exemptions.textproto`. Note that the same lexicographical ordering rule of imports applies here too.
> 
> Done

Thanks @masclot. The PR LGTM. Can you respond to all the individual comments above that you have addressed, so we can mark them as resolved?

adhiamboperes (approved):
Aside from unresolved comments, this generally looks good.

masclot:
> Aside from unresolved comments, this generally looks good.

I cannot find any unresolved comment I haven't replied to. Is there a View in github that summarizes that lists the unresolved comments? Or maybe you can somehow highlight those comments for me/link to them?

masclot:
When I click on "Conversations" it shows three unresolved comments and I replied to all of them.

seanlip:
@masclot [This one](https://github.com/oppia/oppia-android/pull/4874#discussion_r1144089394) and [this one](https://github.com/oppia/oppia-android/pull/4874#discussion_r1144089796)? (I found these by scrolling up on this page.)

masclot:
Thanks Sean. The first one does not have a "Reply" box. The second one I already replied to with "Done both", trying to include the comment I cannot reply to. This comment has a "Pending" label; I don't know what that means.

seanlip:
Ah, for the first one, scroll up the page further and click "Load more". Here's the start of that thread: https://github.com/oppia/oppia-android/pull/4874#discussion_r1143479652 -- you can reply there.

If the comment box says Pending, that means that you have draft comments that haven't been submitted. Go to the "Files Changed" tab, and submit them using the green button in the top right.

masclot:
Thanks! I thought that big green button was for reviewers, not for replies to review-comments. I also don't understand why some of my comments went through without clicking that button. 
Anyway, there are no more pending labels and hopefully no more unreplied comments.

seanlip:
Ah ok, great, thanks @masclot! Not sure about why some of the comments went through, that seems odd.

Also just wondering, do you have any improvements to suggest to the docs [here](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#step-3-address-review-comments-until-all-reviewers-give-lgtm) that would have made the process clearer? If so, would you be up for making a PR to update them? (Now that the wiki docs are stored in the main repo, that should hopefully be straightforward.)

Thanks!

adhiamboperes:
> Ah ok, great, thanks @masclot! Not sure about why some of the comments went through, that seems odd.
> 
> Also just wondering, do you have any improvements to suggest to the docs [here](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#step-3-address-review-comments-until-all-reviewers-give-lgtm) that would have made the process clearer? If so, would you be up for making a PR to update them? (Now that the wiki docs are stored in the main repo, that should hopefully be straightforward.)
> 
> Thanks!

cc @seanlip. @masclot, I believe what happened here was: in some cases you replied from the conversation view, which adds a single comment, and in a later instance, you chose to 
*start a review*, from the *Files changed* tab.

**This**
Screenshot 2023-03-29 at 14 03 47

**vs**

Screenshot 2023-03-29 at 14 03 08

BenHenning (approved):
Thanks @masclot! This LGTM!

BenHenning:
Syncing with develop & enabling auto-merge.

masclot:
> cc @seanlip. @masclot, I believe what happened here was: in some cases you replied from the conversation view, which adds a single comment, and in a later instance, you chose to
> _start a review_, from the _Files changed_ tab.

Thant makes sense, thanks

masclot:
> Also just wondering, do you have any improvements to suggest to the docs [here](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#step-3-address-review-comments-until-all-reviewers-give-lgtm) that would have made the process clearer? If so, would you be up for making a PR to update them? (Now that the wiki docs are stored in the main repo, that should hopefully be straightforward.)

I'll try to compile a list of things I had trouble with. The documentation as a whole is big; I'm sure it has answers to at least some of my troubles, but I just could not remember the solution when I faced them.

seanlip:
@masclot Sounds good, thanks a lot -- appreciate it! 

/cc @MohitGupta121 too, who is doing some work to update the wiki to make it easier to follow.

MohitGupta121:
Thanks @seanlip, Yes @masclot glad to help and collaborate with you in wiki contribution.

PR #2730, opened 2021-02-16
Fix  #1497:Set up code coverage for Robolectric tests 

peculiaruc:
## Explanation
  Fixes #1497 Set up code coverage for Robolectric tests

### Output
domains

cli code coverage4

## Checklist

- [x] The PR title starts with "Fix #bugnum: ", followed by a short, clear summary of the changes. (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The PR explanation includes the words "Fixes #bugnum: ..." (or "Fixes part of #bugnum" if the PR only partially fixes an issue).
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary auto-generated code from Android Studio.
- [x] The PR is made from a branch that's **not** called "develop".
- [x] The PR is made from a branch that is up-to-date with "develop".
- [x] The PR's branch is based on "develop" and not on any other branch.
- [x] The PR is **assigned** to an appropriate reviewer in both the **Assignees** and the **Reviewers** sections.

PR #4886, opened 2023-02-28
Fix #5370, part of #59: Migrate to Bazel 6.5.0

### LINKED ISSUE #5370: [Feature Request]: Update Bazel Linux instructions to use JDK 11

JDK 8 is getting quite old, and it seems that some people run into issues with JDK 8.

We should update the wiki instructions to use JDK 11, instead, but this needs to be verified in a Linux environment. @kkmurerwa has already confirmed JDK 11 works on Mac, so we can update that wiki page at the same time.

None.

### Additional context

Came up during the CLaM meeting today.

BenHenning:
## Explanation

Fixes #5370
Fixes part of #59

This PR updates the project to use Bazel 6.5.0 instead of 4.0.0.

Note that most of the changes done so far in addressing bazelbuild/bazel#59 are centered around the concept of simplifying the Bazel maintenance as much as possible so that it's not too much more difficult than Gradle by the time we fully remove Gradle support from the project. While Bazel will always require more effort, there are many things that can be done to narrow the gap. This is a major step in that process since Bazel 4.x required using a custom Android toolchain (https://github.com/oppia/oppia-bazel-tools) which is not at all user friendly. Plus, there are many compatibility and performance improvements in later versions of Bazel that we want to be able to incorporate within the broader Oppia Android project.

Bazel 6.x was specifically chosen because:
- Bazel 4.x was missing support for the new D8 version which made it impossible to upgrade past 29.0.2 build tools https://github.com/bazelbuild/bazel/issues/13989.
- Bazel 5.x had some additional compatibility issues with the D8 change, so we weren't able to use it, either: https://github.com/bazelbuild/bazel/issues/15957.
- Bazel 7.x (which wasn't released when this work was originally done) introduces new bzlmod support that causes some additional build headaches that can be figured out later.
- Bazel 6.5.0 specifically was chosen since it's the latest 6.x version (as of this edit) and seems to work correctly with existing unit tests.

Some other important details to note:
- rules_kotlin 1.7.x is needed at a minimum for Bazel 5.x+ support. However, an additional fix was needed (https://github.com/bazelbuild/rules_kotlin/pull/940) in order to fix a deviation in functionality that occurred starting in Bazel 5.x's java_plugin support which led to some file duplication in rules_kotlin (that was fortunately easy to fix). Unfortunately, this change wasn't backported to 1.7.x so this PR makes use of a custom patch to rules_kotlin 1.7.1 (https://github.com/oppia/rules_kotlin) that includes the needed change. We'll get this change properly once we can upgrade to 1.8.x, though that will also require updating Kotlin itself to 1.8.x due to https://github.com/bazelbuild/rules_kotlin/issues/1019.
- Bazel 6.x (maybe 5.x) requires at least build tools 30.0.0 since it completely removed support for the old D8 compat dexer. 32.0.0 was chosen in this PR as it's simply a newer, more up-to-date build tools (and removes D8 completely). With this upgrade to Bazel 6.x we'll be able to update the build tools version more often (so long as it doesn't introduce AGP incompatibilities since we can't upgrade Gradle).
- As of Bazel 6.x, we're able to reenable Java header compilation and incremental dexing, both of which should have _significant_ performance improvements for incremental builds of the app (and in fact we will have build errors if we disable incremental dexing).
- In CI, we opted to **not** support build tools 29.0.2 or old builds of the app. Instead, we'll rely on build tools failing for certain PRs as an indicator that those PRs will require an update (once this PR is merged) in order to have CI run correctly. This is a lot easier than trying to figure out how to support before/after changes with some fairly complex environment differences.
- There are a bunch of version updates that were needed to support the minimum version of Kotlin for rules 1.7.x (1.6 I think) as well as JDK 11 (which I think was needed for Bazel 5.x), and these have largely been taken care of in previous PRs to this one (though the JDK 11 update in CI was done in this PR, along with wiki documentation updates to address bazelbuild/bazel#5370). One such case of a necessary version upgrade: https://github.com/google/dagger/issues/2511.
- There was a change needed for the databinding java_plugin declaration to specify that it generates an API (in order for it to be used correctly in builds).
- rules_java needed to be updated to support the newer version of Bazel.
- The desugaring hack needed for kotlinx-coroutines-core-jvm was removed since it's no longer needed with the build tools & Bazel upgrade introduced in this PR.
- This includes one small change in third-party to change all single-export wrappers that don't have additional plugins being enabled to aliases instead. This is more semantically correct as the wrappers may lose information (which caused problems when investigating adding Jetpack Compose support in bazelbuild/bazel#5401). While this isn't directly required for the Bazel upgrade, this is the last PR needed for Jetpack Compose support so it's being added here for simplicity.
- ``.bazelrc`` was updated to configure tools, tests, and builds to all use the remote JDK 11 available via Bazel rather than ever using the user's local JDK. This should improve build hermeticity and consistency across different user environments (see https://bazel.build/docs/bazel-and-java).
- Setup docs were updated to remove setting up JDK 11 (or Java at all for Linux & Mac) now that the user no longer needs to install Java (see previous point) except for Windows. The Python instructions were also removed since Bazel 6.x includes fixes for Android tools that previously depended on Python 2.x.
- CI was unchanged for Java setup since, as far as I can tell, it's still needed for sdkmanager.

There was also some small cleanup in unit_tests.yml that I noticed when updating CI versions.

## Essential Checklist

## For UI-specific PRs only
N/A -- This is a build infrastructure change. It shouldn't impact the end user experience.


gradle.properties
```diff
@@ -23,4 +23,5 @@ android.enableJetifier=false
 kotlin.code.style=official
 # Needed to enable Android data binding.
 android.databinding.enableV2=true
-
+# See: https://stackoverflow.com/q/70878643/3689782.
```
BenHenning: NB: This interestingly seems to only matter when upgrading to Moshi 1.14 (which ended up not happening in this PR), so I may revert this.

Rd4dev:
Hey @BenHenning, I'm reaching out to check if this PR is a working version of Bazel 6.2.0. Previously, I successfully ran Bazel builds and tests with version 4.0.0. However, I am encountering the following issue with this PR. 

**Screenshot:**
Bazel6 2 0error


Could it be related to something missing in my environment, or does Bazel 6.2.0 require additional dependencies?

theMr17:
Hi @Rd4dev, this might be due to something missing in your local environment because the build tests with bazel are passing on CI checks for this PR. Did you set your JDK version to 11 or higher as bazel 6.2.0 requires that?

Rd4dev:
Thank you for sharing the CI insights, @theMr17. 
Yes, as requested, I've been using JDK 11 with Bazel 6.2.0 configured. Interestingly, I've been able to successfully run the Bazel tests, and it appears that only the Bazel build is causing the issue. Upon examining the CI Bazel build, it seems that there may be something missing in my configuration. I'll take a closer look at the requirements and attempt to build again. Once again, thank you

Rd4dev:
@BenHenning, I analyzed the Android Bazel build environment setup CI workflow and then proceeded to create a new environment with JDK 11. I installed the necessary components, including 
- platform-tools
- platforms;android-31 
- and build-tools;32.0.0

One variation I made from the usual setup, which was used with Bazel 4.0.0, was to **omit the [Oppia Bazel Android Tools](https://github.com/oppia/oppia-bazel-tools) configuration**.

With this modified setup, I managed to advance through the build process without encountering any issues initially. However, I faced a setback when a memory regression issue triggered an "Out of memory" error in the JVM. This [issue has been known to occur](https://github.com/bazelbuild/rules_android/issues/302) with upgrades to Bazel 6.0.0 and above.

To address this issue, I attempted to adjust my .wslconfig file to maximize memory allocation. Though, this approach did not yield the desired result. Ultimately, I decided to **remove the .wslconfig** file altogether, allowing the system to fully utilize available resources, which led to a successful build.

I have attached a screenshot showcasing the successful build process: 

Screenshot (989)

I extend my gratitude to @theMr17 for providing the helpful hint regarding the CI setup, which contributed to the successful outcome.

.bazelrc
```diff
@@ -1,9 +1,6 @@
 # Configurations for arguments that should automatically be added to Bazel commands.
 build --android_databinding_use_v3_4_args \
     --experimental_android_databinding_v2 \
-    --java_header_compilation=false \
-    --noincremental_dexing \
-    --define=android_standalone_dexing_tool=d8_compat_dx \
```
BenHenning, line 6: Document these changes.

.github/actions/set-up-android-bazel-build-environment/action.yml
```diff
@@ -77,26 +77,18 @@ runs:
         $ANDROID_HOME/cmdline-tools/tools/bin/sdkmanager --install "platforms;android-33"
       shell: bash
 
-    - name: Install build tools 29.0.2
+    - name: Install build tools 29.0.2 (previous configured version)
```
BenHenning: Document the reason for this.

.github/workflows/build_tests.yml
```diff
@@ -25,15 +25,15 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up JDK 9
+      - name: Set up JDK 11
```
BenHenning, line 28: Document this.

.bazelversion
```diff
@@ -1 +1 @@
-4.0.0
+6.2.0
```
BenHenning: Try 6.5.0 & document reason for version.

.github/workflows/unit_tests.yml
```diff
@@ -86,16 +86,16 @@ jobs:
           echo "Affected tests (note that this might be all tests if configured to run all or on the develop branch): $TEST_BUCKET_LIST"
           echo "::set-output name=matrix::{\"affected-tests-bucket-base64-encoded-shard\":[$TEST_BUCKET_LIST]}"
           if [[ ! -z "$TEST_BUCKET_LIST" ]]; then
-            echo "::set-output name=have_tests_to_run::true"
+            echo "::set-output name=can_skip_tests::false"
```
BenHenning, line 89: Document this.

app/src/main/res/layout-land/story_chapter_view.xml
```diff
@@ -93,7 +93,7 @@
             android:layout_marginEnd="8dp"
             android:layout_marginBottom="4dp"
             android:contentDescription="@{@string/chapter_completed}"
-            android:visibility="@{viewModel.chapterSummary.chapterPlayState == ChapterPlayState.COMPLETED ? View.VISIBLE : View.GONE}"
+            android:visibility="@{viewModel.chapterSummary.chapterPlayState.equals(ChapterPlayState.COMPLETED) ? View.VISIBLE : View.GONE}"
```
BenHenning: Move these changes to earlier PR.

build_vars.bzl
```diff
@@ -1,2 +1,2 @@
 BUILD_SDK_VERSION = 33
-BUILD_TOOLS_VERSION = "29.0.2"
+BUILD_TOOLS_VERSION = "32.0.0"
```
BenHenning, line 2: Document this version upgrade.

gradle.properties
BenHenning: Revert this.

third_party/versions.bzl
```diff
@@ -153,16 +153,16 @@ HTTP_DEPENDENCY_VERSIONS = {
...
     "rules_jvm": {
         "sha": "c4cd0fd413b43785494b986fdfeec5bb47eddca196af5a2a98061faab83ed7b2",
         "version": "5.1",
     },
-    "rules_kotlin": {
-        "sha": "12d22a3d9cbcf00f2e2d8f0683ba87d3823cb8c7f6837568dd7e48846e023307",
-        "version": "v1.5.0",
+    "rules_kotlin_prerelease": {
+        "sha": "1d872b9c6546f0f737a356d873b164d70282760fe4c880349770abc9e494c9ce",
+        "version": "v1.7.2",
```
BenHenning: Document version upgrade & removal of patch.

This is a special fork (link to fork & explain why). Explain how this will change in future PRs.

third_party/versions.bzl
```diff
@@ -153,16 +153,16 @@ HTTP_DEPENDENCY_VERSIONS = {
         "version": "4.5",
     },
     "rules_java": {
-        "sha": "34b41ec683e67253043ab1a3d1e8b7c61e4e8edefbcad485381328c934d072fe",
-        "version": "4.0.0",
+        "sha": "c73336802d0b4882e40770666ad055212df4ea62cfa6edf9cb0f9d29828a0934",
+        "version": "5.3.5",
```
BenHenning, line 157: Document why this version upgrade is needed.

tools/android/BUILD.bazel
```diff
@@ -6,6 +6,7 @@ load("@rules_java//java:defs.bzl", "java_plugin")
 
 java_plugin(
     name = "compiler_annotation_processor",
+    generates_api = True,
```
BenHenning, line 9: Document why this is needed.

utility/build.gradle
```diff
@@ -83,7 +83,7 @@ dependencies {
       'androidx.lifecycle:lifecycle-livedata-ktx:2.2.0-alpha03',
       'androidx.work:work-runtime-ktx:2.4.0',
       'com.github.oppia:androidsvg:5bc9c7553e94c3476e8ea32baea3c77567228fcd',
-      'com.github.oppia:kotlitex:43139c140833c7120f351d63d74b42c253d2b213',
+      'com.github.oppia:kotlitex:ccdf4170817fa3b48b8e1e452772dd58ecb71cf2',
```
BenHenning: Move this to earlier PR.

BenHenning (commented):
Self-reviewed nearly all changes. There are a few big things to address & document before I take another pass.

.bazelrc
BenHenning: Done.

.github/actions/set-up-android-bazel-build-environment/action.yml
BenHenning: Removed instead.

.github/workflows/build_tests.yml
BenHenning: Done.

.bazelversion
BenHenning: Done.

.github/workflows/unit_tests.yml
BenHenning: Done.

build_vars.bzl
BenHenning: Done.

third_party/versions.bzl
BenHenning: Done.

tools/android/BUILD.bazel
BenHenning: Done.

third_party/versions.bzl
BenHenning: Done.

gradle.properties
BenHenning: Done.

app/src/main/res/layout-land/story_chapter_view.xml
BenHenning: Done.

BenHenning: Done.

utility/build.gradle
BenHenning: Done.

BenHenning (commented):
Self-reviewed all recent changes.

BenHenning:
@adhiamboperes PTAL.

Also assigning you @Rd4dev since you have familiarity with these changes, and this is the PR that you need submitted in order to start working on coverage.

BenHenning (commented):
Reviewed latest alias & Maven changes from merges.

FYI that I updated this PR to include some small third-party alias changes to better prepare for Jetpack Compose (per the findings in #5401).

theMr17:
I attempted to build this branch using Bazel in order to ultimately build bazelbuild/bazel#5401 and test Compose with Bazel. However, I am encountering some issues. Do you have any suggestions on what I might be doing wrong or how to fix it? I have already followed the instructions provided in https://github.com/oppia/oppia-android/pull/4886#issuecomment-2096772111.

image

Rd4dev:
> I attempted to build this branch using Bazel in order to ultimately build bazelbuild/bazel#5401 and test Compose with Bazel. However, I am encountering some issues. Do you have any suggestions on what I might be doing wrong or how to fix it? I have already followed the instructions provided in [#4886 (comment)](https://github.com/oppia/oppia-android/pull/4886#issuecomment-2096772111).
> 
> image

Hi @theMr17, while I haven't tested version 6.5.0 locally yet, during my initial testing with version 6.2.0, I encountered this error. Could you please verify if **JAVA_HOME** is correctly set and if **java** and **javac** are pointing to the same versions? 

Meanwhile, I will also conduct local testing to see if I encounter the same issue.

theMr17:
> Hi @theMr17, while I haven't tested version 6.5.0 locally yet, during my initial testing with version 6.2.0, I encountered this error. Could you please verify if **JAVA_HOME** is correctly set and if **java** and **javac** are pointing to the same versions?
> 
> Meanwhile, I will also conduct local testing to see if I encounter the same issue.

Thanks @Rd4dev! Earlier, `javac` was pointing to 1.8 and `java` to 11. I have changed both to point to 11. I have also updated `JAVA_HOME` to be set correctly. But, the issue still remains the same.

Rd4dev:
Thanks for confirming that @theMr17, 

I tried locally setting and testing this PR with **bazel-6.5.0** and **java version 11** and I was able to successfully build it.

image

In this case I mainly suspect 2 things with the setup that might be causing the failure in your case:

1. Any previous configurations that is messing up with the current setup.
2. The version of the command line tools used.
   - The [setup instructions](https://github.com/oppia/oppia-android/wiki/Bazel-Setup-Instructions-for-Windows#3-installing-the-android-sdk) points to the latest version of command tools: https://developer.android.com/studio#command-tools
   - But [digging into the CI setup](https://github.com/oppia/oppia-android/pull/4886/files#diff-feb59558444d6a0d22fd0121a77bfaab423a8e807ee9ffad1b648ff503162a32) we can find it uses a specific version of the command tools: 

image   

        wget https://dl.google.com/android/repository/commandlinetools-linux-7302050_latest.zip
        cp ./commandlinetools-linux-7302050_latest.zip $HOME/android-sdk
        cd $HOME/android-sdk
        unzip commandlinetools-linux-7302050_latest.zip
   - So would it be possible for you to check and see if using this version of command line tools works: https://dl.google.com/android/repository/commandlinetools-linux-7302050_latest.zip

If the issue still persists, I would recommend using a fresh wsl to confirm it is not related to any cache configurations that is causing this.

For your reference, I've documented the entire command sequence that I used to make it work. Please review it to ensure that each step has been followed correctly.

<details>
<summary>Bazel instructions</summary>
<br>

<h4> For Windows: </h4>

```
Installed ubuntu 22.04 [should work with Ubuntu-20.04 too (I can confirm as I tested that with bazel 6.2.0)
wsl --install Ubuntu-22.04

set username 
password

## Bazel instruction with wsl Ubuntu-22.04

$ sudo apt update && sudo apt upgrade

$ sudo apt install openjdk-11-jdk
$ java -version
openjdk version "11.0.22" 2024-01-16
OpenJDK Runtime Environment (build 11.0.22+7-post-Ubuntu-0ubuntu222.04.1)
OpenJDK 64-Bit Server VM (build 11.0.22+7-post-Ubuntu-0ubuntu222.04.1, mixed mode, sharing)

$ sudo apt install openjdk-17-jdk
$ java --version
openjdk 17.0.10 2024-01-16
OpenJDK Runtime Environment (build 17.0.10+7-Ubuntu-122.04.1)
OpenJDK 64-Bit Server VM (build 17.0.10+7-Ubuntu-122.04.1, mixed mode, sharing)

$ sudo apt install python2
$ sudo apt install gcc

$ mkdir -p $HOME/Android/Sdk

$ cp /mnt/c/Users/<UserName>/Downloads/commandlinetools-linux-7302050_latest.zip $HOME/Android/Sdk
cd $HOME/Android/Sdk

$ sudo apt install unzip
$ unzip commandlinetools-linux-7302050_latest.zip

$ cd cmdline-tools/
$ mkdir tools
$ mv -i * tools
$ cd ..

$ echo "export ANDROID_HOME=\$HOME/Android/Sdk" >> ~/.bashrc
$ echo "export PATH=\$ANDROID_HOME/cmdline-tools/tools/bin/:\$PATH" >> ~/.bashrc
$ source ~/.bashrc

$ sdkmanager
$ sdkmanager --install "platform-tools"
$ sdkmanager --install "platforms;android-33"
$ sdkmanager --install "build-tools;32.0.0"

$ sudo apt install apt-transport-https curl gnupg -y
$ curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor >bazel-archive-keyring.gpg
$ sudo mv bazel-archive-keyring.gpg /usr/share/keyrings
$ echo "deb [arch=amd64 signed-by=/usr/share/keyrings/bazel-archive-keyring.gpg] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list

$ sudo apt update && sudo apt install bazel
$ bazel --version
bazel 7.1.2

$ sudo apt install bazel-6.5.0
$ bazel --version
bazel 7.1.2

$ sudo rm /usr/bin/bazel
$ sudo ln -s /usr/bin/bazel-6.5.0 /usr/bin/bazel
$ bazel --version
bazel 6.5.0

$ mkdir $HOME/opensource
$ cd $HOME/opensource

$ git clone https://github.com/oppia/oppia-android.git

$ cd oppia-android/
$ scripts/setup.sh
-bash: scripts/setup.sh: Permission denied

$ chmod +x scripts/setup.sh
$ scripts/setup.sh
Using Ktlint 0.37.1
Ktlint file downloaded
Using Checkstyle version 8.37
Checkstyle file downloaded
Using Bufbuild version v0.37.1
Buf downloaded
Using Buildifier version 3.4.0
Buildifier file downloaded

$ sudo update-alternatives --config java
There are 2 choices for the alternative java (providing /usr/bin/java).

  Selection    Path                                         Priority   Status
------------------------------------------------------------
* 0            /usr/lib/jvm/java-17-openjdk-amd64/bin/java   1711      auto mode
  1            /usr/lib/jvm/java-11-openjdk-amd64/bin/java   1111      manual mode
  2            /usr/lib/jvm/java-17-openjdk-amd64/bin/java   1711      manual mode

Press <enter> to keep the current choice[*], or type selection number: 1
update-alternatives: using /usr/lib/jvm/java-11-openjdk-amd64/bin/java to provide /usr/bin/java (java) in manual mode

$ java --version
openjdk 11.0.22 2024-01-16
OpenJDK Runtime Environment (build 11.0.22+7-post-Ubuntu-0ubuntu222.04.1)
OpenJDK 64-Bit Server VM (build 11.0.22+7-post-Ubuntu-0ubuntu222.04.1, mixed mode, sharing)

$ git branch
$ git branch -r
origin/HEAD -> origin/develop
.....
origin/migrate-to-newer-bazel-and-kotlin
.....

$ git checkout -b migrate_to_bazel_6.5.0 origin/migrate-to-newer-bazel-and-kotlin
M       scripts/setup.sh
Branch 'migrate_to_bazel_6.5.0' set up to track remote branch 'migrate-to-newer-bazel-and-kotlin' from 'origin'.
Switched to a new branch 'migrate_to_bazel_6.5.0'

$ cat .bazelversion
6.5.0

$ bazel build //:oppia
Target //:oppia up-to-date:
  bazel-bin/oppia_deploy.jar
  bazel-bin/oppia_unsigned.apk
  bazel-bin/oppia.apk

```
image
<br>

</details>

If the issue still persists, feel free to share the entire log, and configurations and we may try to look into that.

**One additional note:**
Would recommend allocating atleast 10GB memory for the build.

BenHenning (commented):
Self-reviewed wiki formatting fix.

BenHenning:
Just to check @theMr17 did you ensure you aren't using the Android tools override as described in this wiki update: https://github.com/oppia/oppia-android/blob/1a51da121044d210cac3cef6b8376600a8bbf4c2/wiki/Troubleshooting-Installation.md#bazel-issues? The error you're seeing is different, but seems suspiciously similar in the fact that R8 seems missing from your build path (and it should be part of the bundled Android tools shipped with Bazel 6.x).

Rd4dev:
Hey @BenHenning,

###  Update:
The failing test cases seems to be due to the **`JAVA_HOME`** variable not set. 😐
After manually setting it via;
```
export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
export PATH=$JAVA_HOME/bin:$PATH
```
Now the test cases pass.
Apologies for any confusion, the issue was caused by my setup.

<details>
<summary>Old log of facing the error</summary>
<br>

I was wondering if tests run well with changes on jdk version to 11 or if I am missing something.
[I took HomeActivity.kt related tests for this case]

### With Bazel 6.5.0 and jdk 11 set

I was able to **successfully build** this PR with **`bazel 6.5.0`** and **`jdk 11`**.

Screenshot (1141)

But with no changes made as I try to test the target: 
`bazel test //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest`
It seems to throw errors saying: 
```
...
78) testHomeActivity_clickTopicSummary_opensTopicActivity(org.oppia.android.app.home.HomeActivityTest)
java.lang.NoSuchMethodError: 'java.lang.Class sun.misc.Unsafe.defineAnonymousClass(java.lang.Class, byte[], java.lang.Object[])'
    at org.robolectric.internal.bytecode.ProxyMaker.createProxyFactory(ProxyMaker.java:87)
    at org.robolectric.internal.bytecode.ProxyMaker$1.computeValue(ProxyMaker.java:39)
    at org.robolectric.internal.bytecode.ProxyMaker$1.computeValue(ProxyMaker.java:37)
...
```
Screenshot (1154)

Also with target: `//app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest`

Screenshot (1155)

### StackOverflow Thread

On research this seems like an issue with **`jdk 11`** but I might be wrong and could be missing something.
ref:

> sun.misc.Unsafe.defineClass has been replaced by java.lang.invoke.MethodHandles.Lookup.defineClass in jdk 9, sun.misc.Unsafe.defineClass is completely removed in jdk 11.

Link to the StackOverflow thread: https://stackoverflow.com/questions/55918972/unable-to-find-method-sun-misc-unsafe-defineclass

But the bazel test runs fine with test target: `//utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest` 
while having **`bazel 6.5.0`** and jdk set to 11

Screenshot (1156)

### With Bazel 4.0.0 and jdk 9 set

Also I tried running the targets: 
`//app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest`
`//app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest`
with **`bazel 4.0.0`** and **`jdk 9`** and they seem to run fine.

**HomeActivityTest.kt**
Screenshot (1149)

**HomeActivityLocalTest.kt**
Screenshot (1150)

Would appreciate to know if the tests pass in your case and make further progress figuring it out.
</details>

<details>
<summary>Updated pass test cases screenshots</summary>
<br>
<h3>HomeActivityTest.kt</h3>


<h3>HomeActivityLocalTest.kt</h3>


<br>
</details>

theMr17:
> For your reference, I've documented the entire command sequence that I used to make it work. Please review it to ensure that each step has been followed correctly.
> 
> Bazel instructions

Thanks @Rd4dev! Initially tried changing the command tools version. But, that had no effect. Later, I deleted everything and started with a fresh wsl. Following the compiled instructions you provided, everything worked fine and I was able to successfully build the project.

> Just to check @theMr17 did you ensure you aren't using the Android tools override as described in this wiki update: https://github.com/oppia/oppia-android/blob/1a51da121044d210cac3cef6b8376600a8bbf4c2/wiki/Troubleshooting-Installation.md#bazel-issues? The error you're seeing is different, but seems suspiciously similar in the fact that R8 seems missing from your build path (and it should be part of the bundled Android tools shipped with Bazel 6.x).

Thanks @BenHenning! I believe I had no issue with that. I am now able build this PR after reinstalling wsl. Probably, the caches that were generated when working with bazel 4.0.0 were causing some issue.

Rd4dev:
Hi @BenHenning ,

I was able to successfully execute the builds and tests!

> [still have concerns if coverage works well on all tests - will provide more details on that in the upcoming comment] 

In our wiki instructions regarding the [**`$JAVA_HOME`** configuration](https://github.com/oppia/oppia-android/pull/4886#issuecomment-2134167035), would it be necessary / possible to include the steps to set the variables persistently across sessions as the [existing link](https://www.java.com/en/download/help/path.html) directs to setting the `$PATH` variable alone?

```
echo "export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64" >> ~/.bashrc
echo "export PATH=$JAVA_HOME/bin:$PATH" >> ~/.bashrc
source ~/.bashrc
```

**Related issues:**
Linux Missing Java Home Instructions: https://github.com/oppia/oppia-android/issues/5293
My issue with executing bazel tests: https://github.com/oppia/oppia-android/pull/4886#issuecomment-2134167035 [solved]

Rd4dev:
Hi @BenHenning,

Though I was able to run tests with bazel 6.5.0 and jdk 11, I am facing issues with coverage that I would like to update.

I am using the HomeActivity.kt related test cases:
1. `//app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest`
2. `//app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest`

while both test target PASS (with bazel test execution)
### HomeActivityTest
Screenshot (1162)

### HomeActivityLocalTest
Screenshot (1164)

attempting to acquire coverage results seem to fail with message saying:

```
...
FAILURES!!!
Coverage runner: Not collecting coverage for failed test.
The following commands failed with status 1
...
Executed 1 out of 1 test: 1 fails locally.
```

Screenshot (1172)

**To note:** 
- [x] The tests pass with bazel tests command
- [ ] The tests fail when executed with bazel coverage command

To make sure it is not based on my local configurations I tried setting up [another repo](https://github.com/Rd4dev/cctest_oppia_android/pull/1) and tried to execute the [ci workflow](https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352) while adding the code to run bazel coverage in Run Bazel Tests step

```
while [ $i -ne 5 ]; do
    i=$(( $i+1 ))
    echo "Attempt $i/5 to run test targets"
    bazel test --keep_going -- $BAZEL_TEST_TARGETS
    bazel coverage $BAZEL_TEST_TARGETS
  done
```
image

And they too seem to replicate the same behaviour.

### HomeActivityLocalTest
**Test case execution:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161742#step:17:394
**Test case pass:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161742#step:17:981
**Coverage case fail:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161742#step:17:8634

### HomeActivityLocalTest
**Test case execution:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161495#step:17:219
**Test case pass:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161495#step:17:792
**Coverage case fail:** https://github.com/Rd4dev/cctest_oppia_android/actions/runs/9268684352/job/25498161495#step:17:8288

These mainly failed with warning related to: 
```
java.lang.IllegalAccessError: class org.oppia.android.app.home.HomeFragmentPresenter tried to access private method 'boolean[] 
```
Related StackOverflow Link: https://github.com/jmockit/jmockit1/issues/553#issuecomment-422519248

Also, I was able to run coverage commands for other tests
**eg:** `//utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest`

Screenshot (1171)

I am currently looking into the bazel configurations if anything with that is causing the issue.
I will update if I figured it out.

Would like to know if I am missing any necessary step?

Thanks.

BenHenning:
Thanks @Rd4dev. Regarding the JAVA_HOME/PATH updates, I was originally a bit hesitant to include those because it can vary based on each machine/installation. Are these steps specifically necessary for WSL? If so, I think it would make sense to include them.

Regarding code coverage, I'm not sure yet. I'll try to dig a bit and see if I can find anything useful.

BenHenning:
Interestingly, FractionParsingUiErrorTest seemed to run with coverage fine (I was wondering if it's just app module tests, but it's apparently more specific). HomeActivityTest also failed for me.

This definitely seems like something that's tied to Kotlin & instrumentation code generation being done by JaCoCo since it's happening within our code (for me, the issue is trying to access something in ``$jacocoInit()`` within ``BindableAdapter.MultiTypeBuilder.Factory``).

Edit: Ah, actually perhaps this is a Robolectric/JDK thing. I noticed that the stack trace has a part that says "HomeFragmentPresenter and BindableAdapter are in unnamed module of loader org.robolectric.internal.AndroidSandbox$SdkSandboxClassLoader".

Edit 2: I came across https://github.com/robolectric/robolectric/issues/8484 and tried to upgrade Robolectric to 4.12.2 along with AndroidX test libraries & Espresso to 3.5.1, but I still run into the same access issue. I am running on JDK 14, though, so I wonder if this is just tied to JVM modules.

Edit 3: also repros on JDK 11, though if I recall modules were added in JDK 9 (and I see some other folks having module issues on JaCoCo's issue tracker).

BenHenning:
Can't look into this anymore today, but https://github.com/jacoco/jacoco/issues/1487 seems interesting. I wonder if the version of JaCoCo we're using (which seems to probably be 0.8.7 or 0.8.8 per https://github.com/somethingvague/bazel/blob/master/third_party/java/jacoco/BUILD) might simply have some compatibility issues with newer JDK versions (and, specifically, isn't working correctly with whatever funkiness Robolectric is doing).

@Rd4dev do you have any better luck with using JaCoCo directly rather than via ``bazel coverage`` for one of the failing tests? We may need to wrap a later version of JaCoCo in our own Bazel command if so (I can help with that). I did see some differences in failures when upgrading to a newer version of Robolectric, but I'm thinking all the failures have the same root cause and it may go away with the latest JaCoCo.

Also, it may be worth using the add-opens JVM argument mentioned in that issue with the newer version of JaCoCo (I'd try it with and without the argument just to see if it's actually tied to JVM modules).

Rd4dev:
> Thanks @Rd4dev. Regarding the JAVA_HOME/PATH updates, I was originally a bit hesitant to include those because it can vary based on each machine/installation. Are these steps specifically necessary for WSL? If so, I think it would make sense to include them.
> 
> Regarding code coverage, I'm not sure yet. I'll try to dig a bit and see if I can find anything useful.

@BenHenning, while using **Bazel 4.0.0** and **JDK 9**, I didn't have the need to set the **`$JAVA_HOME`** to get the tests PASS.
But with **Bazel 6.5.0** and **JDK 11**, I had to do that. 
For confirmation I tried removing the variable `$JAVA_HOME` and doing that caused a FAIL case.

### Without **`$JAVA_HOME`** set
Screenshot (1177)

<details>
<summary>.bashrc file</summary>

</details>

### With **`$JAVA_HOME`** set
Screenshot (1182)

<details>
<summary>.bashrc file</summary>

</details>

### Entire Log
Screenshot (1183)

Also as we are juggling with multiple JDKs for handling sdkmanager and builds, I feel it might be safe to set $JAVA_HOME to ensure Bazel uses the specified JDK, instead of falling back to other installed versions preventing any discrepancies, while I didn't have the need to set the $PATH variable.
```
export PATH=$JAVA_HOME/bin:$PATH"
```
as my test passed without them.

BenHenning:
Interesting. I just checked and my local Ubuntu environment doesn't have ``JAVA_HOME`` set and Bazel 6.x tests pass just fine. I wonder what the difference is here. I know that ``PATH`` is the primary source for locating the ``java`` binary, but ``JAVA_HOME`` _can_ sometimes be used in certain situations. Some searching on the Bazel GitHub repo didn't yield anything specific to 6.x, either (though there are a lot of hits for ``JAVA_HOME`` and I admittedly didn't look through all of them).

Per https://bazel.build/docs/bazel-and-java we could update our .bazelrc to force all test execution to using the remote JDK like the tools do. This probably would be the best option to avoid system-specific differences like this (and honestly largely simplify which JDKs the user needs to install). I think even Android Studio uses a bundled JDK which means we could possibly remove all need for custom JDK installation, but it'll really depend on whether we can figure out a way to simplify the Android SDK installation part.

@Rd4dev could you try to add this CLI parameter:

```sh
bazel test --java_runtime_version=remotejdk_11 ...
```

With that, do your tests pass?

Rd4dev:
> Interesting. I just checked and my local Ubuntu environment doesn't have `JAVA_HOME` set and Bazel 6.x tests pass just fine. I wonder what the difference is here. I know that `PATH` is the primary source for locating the `java` binary, but `JAVA_HOME` _can_ sometimes be used in certain situations. Some searching on the Bazel GitHub repo didn't yield anything specific to 6.x, either (though there are a lot of hits for `JAVA_HOME` and I admittedly didn't look through all of them).
> 
> Per https://bazel.build/docs/bazel-and-java we could update our .bazelrc to force all test execution to using the remote JDK like the tools do. This probably would be the best option to avoid system-specific differences like this (and honestly largely simplify which JDKs the user needs to install). I think even Android Studio uses a bundled JDK which means we could possibly remove all need for custom JDK installation, but it'll really depend on whether we can figure out a way to simplify the Android SDK installation part.
> 
> @Rd4dev could you try to add this CLI parameter:
> 
> ```shell
> bazel test --java_runtime_version=remotejdk_11 ...
> ```
> 
> With that, do your tests pass?

Yes @BenHenning, the tests pass with the inclusion of the parameter;
```
--java_runtime_version=remotejdk_11
```
I also made sure to remove the set `$JAVA_HOME` varible
### With ----java_runtime_version=remotejdk_11 parameter
Screenshot (1200)

<details>
<summary>Log</summary>
rddev@DESKTOP-CLN30QH:~/opensource/oppia-android$ bazel test --java_runtime_version=remotejdk_11 //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest
Starting local Bazel server and connecting to it...
INFO: Analyzed target //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest (437 packages loaded, 6496 targets configured).
INFO: Found 1 test target...
....
....
Time: 49.896

OK (78 tests)


BazelTestRunner exiting with a return value of 0
JVM shutdown hooks (if any) will run now.
The JVM will exit once they complete.

-- JVM shutdown starting at 2024-05-29 19:50:13 --

================================================================================
Target //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest up-to-date:
  bazel-bin/app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.jar
  bazel-bin/app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest
INFO: Elapsed time: 455.668s, Critical Path: 285.48s
INFO: 276 processes: 4 internal, 148 linux-sandbox, 124 worker.
INFO: Build completed successfully, 276 total actions
//app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest    PASSED in 53.0s

rddev@DESKTOP-CLN30QH: ~/opensource/oppia-android$ echo $JAVA_HOME
rddev@DESKTOP-CLN30QH: ~/opensource/oppia-android$
rddev@DESKTOP-CLN30QH: ~/opensource/oppia-android$ ls
BUILD.bazel  app                  bazel-testlogs     build_vars.bzl         domain             gradlew.bat                    oppia_android_test.bzl  third_party  wiki
LICENSE      bazel-bin            buf.yaml           bundle_config.pb.json  gradle             instrumentation                scripts                 tools
NOTICE       bazel-oppia-android  build.gradle       config                 gradle.properties  model                          settings.gradle         utility
WORKSPACE    bazel-out            build_flavors.bzl  data                   gradlew            oppia_android_application.bzl  testing                 version.bzl
</details>

### Without the --java_runtime_version=remotejdk_11 parameter
Screenshot (1199)

<details>
<summary>Log</summary>
rddev@DESKTOP-CLN30QH:~/opensource/oppia-android$ bazel test //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest
INFO: Build option --java_runtime_version has changed, discarding analysis cache.
INFO: Analyzed target //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest (0 packages loaded, 6606 targets configured).
INFO: Found 1 test target...
.....
.....
78) testHomeActivity_clickTopicSummary_opensTopicActivity(org.oppia.android.app.home.HomeActivityTest)
java.lang.NoSuchMethodError: 'java.lang.Class sun.misc.Unsafe.defineAnonymousClass(java.lang.Class, byte[], java.lang.Object[])'
        at org.robolectric.internal.bytecode.ProxyMaker.createProxyFactory(ProxyMaker.java:87)
        at org.robolectric.internal.bytecode.ProxyMaker$1.computeValue(ProxyMaker.java:39)
        at org.robolectric.internal.bytecode.ProxyMaker$1.computeValue(ProxyMaker.java:37)
        at org.robolectric.internal.bytecode.ClassValueMap.get(ClassValueMap.java:21)
        at org.robolectric.internal.bytecode.ProxyMaker.createProxy(ProxyMaker.java:45)
        at org.robolectric.internal.bytecode.ShadowImpl.createProxy(ShadowImpl.java:31)
        at org.robolectric.internal.bytecode.ShadowImpl.directlyOn(ShadowImpl.java:26)
        at org.robolectric.shadow.api.Shadow.directlyOn(Shadow.java:46)
        at org.robolectric.shadows.ShadowResources.obtainAttributes(ShadowResources.java:90)
        at android.content.res.Resources.obtainAttributes(Resources.java)
        at android.content.pm.PackageParser.parseBaseApk(PackageParser.java:1821)
        at android.content.pm.PackageParser.parseBaseApk(PackageParser.java:1188)
        at android.content.pm.PackageParser.parseMonolithicPackage(PackageParser.java:1150)
        at android.content.pm.PackageParser.parsePackage(PackageParser.java:1054)
        at android.content.pm.PackageParser.parsePackage(PackageParser.java:1063)
        at org.robolectric.shadows.ShadowPackageParser.callParsePackage(ShadowPackageParser.java:47)
        at org.robolectric.android.internal.AndroidTestEnvironment.loadAppPackage_measured(AndroidTestEnvironment.java:326)
        at org.robolectric.android.internal.AndroidTestEnvironment.lambda$loadAppPackage$1(AndroidTestEnvironment.java:296)
        at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:53)
        at org.robolectric.android.internal.AndroidTestEnvironment.loadAppPackage(AndroidTestEnvironment.java:294)
        at org.robolectric.android.internal.AndroidTestEnvironment.installAndCreateApplication(AndroidTestEnvironment.java:195)
        at org.robolectric.android.internal.AndroidTestEnvironment.setUpApplicationState(AndroidTestEnvironment.java:171)
        at org.robolectric.RobolectricTestRunner.beforeTest(RobolectricTestRunner.java:319)
        at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:269)
        at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:840)

FAILURES!!!
Tests run: 78,  Failures: 78

WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by com.google.testing.junit.runner.util.GoogleTestSecurityManager (file:/home/rddev/.cache/bazel/_bazel_rddev/9cb02c856feab95ef2fdd476fda716be/external/remote_java_tools/java_tools/Runner_deploy.jar)
WARNING: Please consider reporting this to the maintainers of com.google.testing.junit.runner.util.GoogleTestSecurityManager
WARNING: System::setSecurityManager will be removed in a future release

BazelTestRunner exiting with a return value of 1
JVM shutdown hooks (if any) will run now.
The JVM will exit once they complete.

-- JVM shutdown starting at 2024-05-29 19:52:08 --

================================================================================
Target //app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest up-to-date:
  bazel-bin/app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.jar
  bazel-bin/app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest
INFO: Elapsed time: 26.181s, Critical Path: 15.57s
INFO: 6 processes: 4 internal, 2 linux-sandbox.
INFO: Build completed, 1 test FAILED, 6 total actions
//app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest    FAILED in 6.7s
  /home/rddev/.cache/bazel/_bazel_rddev/9cb02c856feab95ef2fdd476fda716be/execroot/__main__/bazel-out/k8-fastbuild/testlogs/app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest/test.log

Executed 1 out of 1 test: 1 fails locally.
rddev@DESKTOP-CLN30QH:~/opensource/oppia-android$ echo $JAVA_HOME

rddev@DESKTOP-CLN30QH:~/opensource/oppia-android$
</details>

BenHenning:
Thanks @Rd4dev! It sounds like perhaps that's the way to go, then. I'll actually update this PR to include the change in the shared .bazelrc and hopefully that at least partially simplifies the JDK setup work for everyone.

Separately, did you get a chance to test JaCoCo directly as I suggested in my earlier comment? I'm hoping this is as simple as just introducing a custom Bazel command for running coverage through our own JaCoCo version.

.bazelrc
```diff
@@ -1,9 +1,6 @@
 # Configurations for arguments that should automatically be added to Bazel commands.
 build --android_databinding_use_v3_4_args \
```
BenHenning, line 2: Per discussion with RD, update this file to include setting java_runtime_version to remotejdk_11 (or whatever the default is for builds; we should keep them the same).

Also, update docs since installing JDK should now be simpler.

BenHenning (commented):
Small merge conflict to verify from upstream sync; no issues found during self-review.

Rd4dev:
> Thanks @Rd4dev! It sounds like perhaps that's the way to go, then. I'll actually update this PR to include the change in the shared .bazelrc and hopefully that at least partially simplifies the JDK setup work for everyone.
> 
> Separately, did you get a chance to test JaCoCo directly as I suggested in my earlier comment? I'm hoping this is as simple as just introducing a custom Bazel command for running coverage through our own JaCoCo version.

Sorry for the delayed response, @BenHenning. I tried to come up with something useful, but testing with JaCoCo separately did not turn out as expected. I tried a few versions, but the results were unsatisfactory. I used the same setup as in PR bazelbuild/bazel#2730:

- Versions 0.8.8, 0.8.11, 0.8.12: Setting both `ext.jacocoVersion` and `jacocoToolVersion` generated a report but showed 0% coverage (except for 2 cases in the viewmodel). I also tested the utility and domain modules with 0.8.12, but they showed the same behavior, making me question if the setup was correct. I used the same setup as PR bazelbuild/bazel#2730 while testing various versions.

### 0% coverage
Screenshot (1213)

- While PR bazelbuild/bazel#2730 was built with JDK 8 and produced coverage for the domain module, it still fails for app module. The https://github.com/oppia/oppia-android/pull/2730#issuecomment-792213023 mentions to use 0.8.5 but that didn't work for app module.

### domain with 0.7.9 jdk 8
Screenshot (1227)

### app with 0.8.5 jdk 8
Screenshot (1229)

- Using the same versions as PR bazelbuild/bazel#2730, with `ext.jacocoVersion` set to 0.8.5 and `jacocoToolVersion` to 0.7.9, throws an error saying, "java.lang.RuntimeException: Class java/util/UUID could not be instrumented." This seems related to JDK 11. https://github.com/ASSERT-KTH/flacoco/issues/106. Updating to 0.8.7 was the proposed solution, but even with 0.8.7, 0% coverage was returned.

**Extra info:**
Majorly app modules coverage seems to fail. 81 out of 166 tests: 85 tests pass and 81 fail locally.

Apart from the app module, though I haven't run it for the entire testing module, the coverage for the target: `//testing/src/test/java/org/oppia/android/testing:FakeAnalyticsEventLoggerTest` too fails.

BenHenning:
Thanks @Rd4dev. It's interesting to see the differences with Gradle.

I spent some time today trying a bunch of different things to try and even get a newer version of JaCoCo running. It would be a lot to summarize, so here's the very high-level:
- We can build a deploy version of our Robolectric tests, but actually directly running them via JVM CLI is quite hard because there's a lot of environment setup stuff Bazel does behind the scenes.
- It _is_ possible to set up a custom Java toolchain to replace the built-in one that comes with Bazel (so that we can swap out the JaCoCo binary it points to).
- Swapping out JaCoCo is hard. Bazel builds its own wrapper around the JaCoCo library rather than using its CLI, so we'd have to build our own version of that which points to a newer JaCoCo library.
- The above isn't too much of an issue, except Bazel is in a weird mid-migration state where it's moving its remote_java_tools dependencies to a separate repo, but it's in between state is essentially:
  - Bazel itself instruments remote_java_tools when building (since rules_java still just points to the native Java rules as they haven't yet been fully migrated to Starlark).
  - Bazel does the above by pointing to remote_java_tools prebuilt deployments which are generated using a script that points back to Bazel in order to create the deployable Jars for its package.

That's as far as I've made it to updating JaCoCo even for a light test. Essentially next steps would be:
- Build a custom remote_java_tools deployment that contains a newer version of JaCoCo.
- Upload this somewhere (like a GitHub fork).
- Update my local checkout of Bazel to point to these new packages & rebuild.
- Use the locally built Bazel binary to try and run coverage.

It would probably be much easier to try and create a similar failure case outside of Bazel entirely (i.e. using direct ``javac`` and JaCoCo's CLIs), but without a better sense of what the root issue is that will presumably be hard to come up with. Hopefully I can make headway on the above to at least try to use a newer version of JaCoCo to see if that makes any difference on the failures.

wiki/Bazel-Setup-Instructions-for-Linux.md
```diff
@@ -1,31 +1,29 @@
...
 6. Verify the Android build
 
 ### 1. Install Bazel
 
-Install Bazel from [here](https://docs.bazel.build/versions/master/install.html). Make sure that you follow the instructions for installing a specific version (Oppia Android requires 4.0.0 and won't build on other versions).
+Install Bazel from [here](https://docs.bazel.build/versions/master/install.html). Make sure that you follow the instructions for installing a specific version (Oppia Android requires 6.5.0 and won't build on other versions).
    - Note: if you find any errors related to `cURL`, please set up cURL on your machine. For Linux, you can use `sudo apt install curl`.
 
-### 2. Install OpenJDK 8
+### 2. Install OpenJDK 11
```
adhiamboperes: Would it be possible to add $JAVA_HOME setup instructions in this PR? Re: https://github.com/oppia/oppia-android/issues/5293

adhiamboperes (changes requested):
Thanks @BenHenning!

I left a comment, but you also left an unresolved comment. PTAL.

.bazelrc
BenHenning: Addressed in latest changes.

wiki/Bazel-Setup-Instructions-for-Linux.md
BenHenning: I think we actually don't need this anymore with the other latest changes. Now that we're using the JDK bundled with Bazel, developers technically don't even need to install Java anymore (as far as I can tell). I've cleaned up these docs significantly in the latest changes.

BenHenning (commented):
Reviewed latest changes + upstream merge.

BenHenning (commented):
Verified latest change (undo action.yml change that removed Java verification).

BenHenning:
Thanks @adhiamboperes! I think the latest changes actually address your comment and fairly substantially simplify Bazel setup across all platforms. Could you PTAL for code changes, and also double check the Mac instructions to make sure they're correct?

Similarly, @Rd4dev could you also review this PR (and approve if it looks good to you?), and specifically verify that the Windows instructions are correct?

BenHenning (commented):
Self-reviewed upstream merge (was clean).

BenHenning:
PTAL @seanlip for owners.

adhiamboperes (approved):
LGTM, thanks @BenHenning!

seanlip (approved):
LGTM for owners, thanks!

BenHenning (commented):
Self-reviewed last few merges--nothing looks off.

BenHenning:
Currently waiting on @adhiamboperes and @Rd4dev (per a request in Chat) to verify the Mac & Windows compatibility (respectively) before merging this since this is a significant change to Bazel support.

I also will send out an email to the Android team letting them know of the changes they'll need to make in order to use Bazel moving forward.

Rd4dev:
> Currently waiting on @adhiamboperes and @Rd4dev (per a request in Chat) to verify the Mac & Windows compatibility (respectively) before merging this since this is a significant change to Bazel support.
> 
> I also will send out an email to the Android team letting them know of the changes they'll need to make in order to use Bazel moving forward.

Hey @BenHenning, I tried using a new WSL setup with Ubuntu 18.04.6 (Following [Bazel instructions for Windows](https://github.com/oppia/oppia-android/pull/4886/files#diff-17366a221a97bc7926ad10d58bccf8f32d1a05a83059d23f1fc7dfd048ca518f)) and followed the new command which is to use the JDK Bundled with Bazel. I was able to 
```
bazel build //:oppia 
```
but with 
```
bazel build //:oppia_dev
```
I am facing issues saying "Generating deployable AAB failed: (Exit 127): bash failed: error executing command (from target //:oppia_dev_raw_module_zip) /bin/bash -c ... (remaining 1 argument skipped)"

**Screenshot**
Screenshot (1292)

Am I missing any necessary steps?

Also I tried testing and coverages with working targets and they yielded appropriate results.

Screenshot (1293)

BenHenning:
So I spent quite a while trying to find an alternative to zip as that's what @Rd4dev reported wasn't working on Windows (@adhiamboperes confirmed via chat that the build works on Mac). However, ``//:oppia`` _is_ working for RD.

At this point, I realized that ``//:oppia_dev`` should _already_ be failing on Windows, so fixing this ought not block this PR. Going ahead and merging it after build stats runs.

Rd4dev:
> So I spent quite a while trying to find an alternative to zip as that's what @Rd4dev reported wasn't working on Windows (@adhiamboperes confirmed via chat that the build works on Mac). However, `//:oppia` _is_ working for RD.
> 
> At this point, I realized that `//:oppia_dev` should _already_ be failing on Windows, so fixing this ought not block this PR. Going ahead and merging it after build stats runs.

Got it, @BenHenning. Would that require making changes to [Bazel Instructions with Windows wiki](https://github.com/oppia/oppia-android/pull/4886/files#diff-17366a221a97bc7926ad10d58bccf8f32d1a05a83059d23f1fc7dfd048ca518fR177) to use `//:oppia` and not `//:oppia_dev`?

Rd4dev (approved):
Setting aside the existing `//:oppia_dev` issue on Windows, and the [coverage issues](https://github.com/oppia/oppia-android/pull/4886#issuecomment-2135684503) with a few test targets, everything else LGTM. Thanks @BenHenning!

BenHenning:
@Rd4dev that's probably a good idea (the suggested Windows doc change). Will make that and then enable auto-merge.

Separately, I realized that build stats won't run for this branch since it's using a newer version of Bazel. :)

wiki/Bazel-Setup-Instructions-for-Windows.md
```diff
@@ -227,40 +173,35 @@ scripts/setup.sh
...
-```sh
-sudo update-alternatives --config java
-```
-
-Select the number with JDK-8
-
-**Build**
 ```sh
-bazel build //:oppia
+bazel build //:oppia_dev
```
BenHenning:
```suggestion
bazel build //:oppia
```

wiki/Bazel-Setup-Instructions-for-Windows.md
```diff
@@ -227,40 +173,35 @@ scripts/setup.sh
...
 
 If everything is working, you should see output like the following:
 
 ```
-Target //:oppia up-to-date:
-  bazel-bin/oppia_deploy.jar
-  bazel-bin/oppia_unsigned/apk
-  bazel-bin/oppia/apk
+Target //:oppia_dev up-to-date:
+  bazel-bin/oppia_dev.aab
```
BenHenning:
```suggestion
Target //:oppia up-to-date:
  bazel-bin/oppia_deploy.jar
  bazel-bin/oppia_unsigned/apk
  bazel-bin/oppia/apk
```

wiki/Bazel-Setup-Instructions-for-Windows.md
```diff
@@ -227,40 +173,35 @@ scripts/setup.sh
...
+
+```sh
+bazel run //:install_oppia_dev
+```
+
+If this fails, you may need to use ADB from within a Windows command prompt and install ``bazel-bin/oppia.apk`` which can be produced via the following command to build an APK directly:
+
+```sh
+bazel build //:oppia
+```
```
BenHenning:
```suggestion
Note also that the ``oppia.apk`` under the ``bazel-bin`` directory of your local copy of Oppia Android should be a fully functioning development version of the app that can be installed using ``adb`` (though we recommend using ADB from within a Windows command prompt or shell since the Ubuntu subsystem may not have correct support to access devices or emulators connected to the native Windows machine).
```

BenHenning (commented):
Self-reviewed Windows doc changes (mostly reversions).

wiki/Bazel-Setup-Instructions-for-Windows.md
BenHenning: Addressed.

wiki/Bazel-Setup-Instructions-for-Windows.md
BenHenning: Addressed.

wiki/Bazel-Setup-Instructions-for-Windows.md
BenHenning: Addressed.

PR #4896, opened 2023-03-10
Fixes #4897: Follow up alpha MR6 fixes

### LINKED ISSUE #4897: Follow-up issues & fixes for Kenya user study research project
Some issues were found during the Kenya user study that led to #4896 being created, and some additional existing issues were found during development.

Specifically, this issue is tracking adding:
- Adding tracking information to the study-only learner & IDs screen that shows per-profile event tracking information (to help with server-side correlation).
- Sharing functionality for on-device stored events (so that we can work with these events directly in the event that they fail to upload).
- A force sync button to upload events immediately.

And, it's tracking fixing:
- The sync status indicator can be wrong sometimes: if there are pending events that are cached but haven't been uploaded yet, and then an event is later logged due to internet connectivity being restored, the indicator will specify that events have been uploaded (even though there are some still not uploaded).
- The possibility of cached events being lost if an event is logged with internet connectivity upon app open (with cached events) before the app's cached events are read.
- The possibility of cached events being duplicated in very specific situations (that are very unlikely to actually occur in practice, but are theoretically possible).

BenHenning:
## Explanation
Fixes #4897

### Overview of changes

This PR follows up from the changes introduced in #4846 by adding new functionality to help improve event reliability investigations, as well as telemetry diagnostics for research project facilitators that need to ensure events have been successfully uploaded for processing. To help with this, the following features have been introduced:
- A new "force upload" button that immediately uploads any pending cached events.
- Event counts to show both the number of events waiting to be uploaded & the events that are successfully uploaded.
- An improvement on the 'share' button to include a Base64 deflated binary representation of the entire event log store on the device.

Leading up to, and during, development, a number of theoretical issues were encountered that were also fixed in case they are playing a role in some of the ongoing concerns around Firebase event delivery reliability:
- The sync status reported on the learner analytics screen can be wrong in some cases, leading to the assumption that events have been uploaded when they may not have been.
- Events can, in some cases, be lost from previous app instances if they haven't been uploaded yet and, in other cases, duplicated. The former is much more likely to happen in practice, particularly with the connectivity patterns being used in the research project.

### Technical specifics

- ``FakeSyncStatusManager`` was renamed to ``TestSyncStatusManager`` since it's actually augmenting the production implementation of ``SyncStatusManager``, not replacing it.
- Events are now being cached indefinitely if the learner study feature is enabled. This allows for some robustness improvements (see below), and the ability to share the events themselves if the team suspects events are not arriving to Firebase.
- The issue of sync status being incorrect came from an unreliable means of reporting sync status during different event scenarios. Specifically, if the ``LogUploadWorker`` doesn't kick off and there's network connectivity, a new event being logged could trigger the sync status to become "data uploaded" even though there are still pending events. The new solution is to leverage the now-tracked uploaded events to determine whether there are still events yet waiting to upload. In this way, the sync status should always be the most up-to-date information with one caveat: network connectivity changes won't reflect until a new log is requested (this is a limitation in how the app tracks network connectivity, and it didn't seem very important to fix it).
- The issue of possible event duplication comes from ``PersistentCacheStore`` using ``mergeFrom`` with the *current* cached state and the on-disk proto being loaded. I originally thought that all pathways to loading the proto from disk could only execute once, but there's at least one pathway where it's possible to chain multiple calls into the cache store such that the load happens twice (and both results end up being merged together, hence the event duplication). This has been fixed by never reloading the cache store if it's already been loaded which should always be correct.
- The issue if possible events being lost comes from ``AnalyticsController`` previously not priming the event log cache. This means that if a new event is logged before the cache is read (e.g. early events like "open profile screen" can race against ``LogUploadWorker`` kicking off on app startup) then the cache will be completely reset and, once written, will overwrite any events currently cached on disk. This has been fixed by ensuring that the event log cache has been primed before any interactions with it (read or write) occur.

Note that all of the above has been thoroughly tested through automated testing to ensure that these issues have been properly fixed, and stay that way in the long-term.

### Test notes & exemptions

- In order to simplify testing ``SyncStatusManagerImpl`` and ``TestSyncStatusManager``, they ought to both pass a common test suite specific to ``SyncStatusManager`` itself. This has been done by introducing a ``SyncStatusManagerTestBase`` class that's inherited by both implementation test suites so that they inherit common API tests. This pattern is used in one other location in the codebase currently: https://github.com/oppia/oppia-android/blob/0290ef8037f798d8ba2721441612a80dec4a1f1a/testing/src/test/java/org/oppia/android/testing/threading/TestCoroutineDispatcherTestBase.kt#L33
- ``CircularProgressIndicatorAdaptersTestActivity`` has been exempted for an a11y label for the same reason as all other test activities: it technically doesn't need one.
- ``TestSyncStatusManagerTest`` has been allow-listed to use parameterized tests since it's aggressively verifying many different force/override sync status cases. Making sure that the test utility works correctly is extremely important to avoid false failures/passes of dependent tests (which are, in turn, guarding against regressions for the core issues covered by this PR, among other event system behaviors).
- ``SyncStatusManagerTestBase`` has been exempted from requiring KDocs since it's technically a test suite, so KDocs aren't required.
- A number of classes have been exempted from having new test suites as per existing conventions and requirements: ``ControlButtonsViewModel`` (renamed from ``ShareIdsViewModel``), ``CircularProgressIndicatorAdaptersTestActivity`` (simple test activities don't require their own tests), ``CircularProgressIndicatorAdaptersTestViewModel`` (view models aren't tested directly), ``SyncStatusManagerTestBase`` (the test base is more or less a test suite itself, so it doesn't make sense for it to have tests as well).
- ``TestSyncStatusManagerTest`` has been disabled in Gradle since it depends on ``SyncStatusManagerTestBase`` which is part of a different module. To make this work, the test base would have to be moved to src/main of the testing module which isn't a very clean approach. The current setup works fine with Bazel, and longer term this will get cleaner by moving testing utilities like ``TestSyncStatusManager`` to be near their corresponding production implementations (we just can't do this today because of Gradle limitations).
- Due to the asynchronous and data race nature of some of the issues discovered during test writing, changed & added tests were sometimes found to be flaky (and, in some cases, very rarely, e.g. 1/50). To ensure that all of the new & changed tests are stable, I made sure to run all 8 affected test suites 128 times to ensure stability. Here are the results of that run ([buildbuddy link](https://app.buildbuddy.io/invocation/5da310e1-3cc9-451b-a1d0-951a3322c8d7)):
128xruns_followup_mr6_fixes

## Essential Checklist

## For UI-specific PRs only

Following is a video demonstrating the new functionality introduced in this PR:

https://user-images.githubusercontent.com/12983742/224289397-73c5e3a7-4f7a-4116-8e71-3316e30fa9d7.mp4

Note that affected UI tests that could run on Espresso were **not** verified in this PR similar for the reasons listed in #4846 (specifically, the ongoing issue of Espresso tests being time consuming to investigate, often broken, and this PR being high priority).

Localization & accessibility changes aren't relevant for this change, and any tablet-specific issues aren't important since the affected screen is only expected to be accessed on handset devices.


BenHenning:
TODO: Only retain uploaded events if the study is on (& add corresponding tests).

app/src/main/java/org/oppia/android/app/application/alphakenya/BUILD.bazel
```diff
@@ -21,7 +21,7 @@ kt_android_library(
         "//app/src/main/java/org/oppia/android/app/application:common_application_modules",
         "//app/src/main/java/org/oppia/android/app/application/alpha:alpha_build_flavor_module",
         "//utility/src/main/java/org/oppia/android/util/logging:kenya_alpha_event_logging_configuration_module",
-        "//utility/src/main/java/org/oppia/android/util/logging/firebase:prod_module",
+        "//utility/src/main/java/org/oppia/android/util/logging/firebase:debug_module",
```
BenHenning: Revert, debug shouldn't be needed.

data/src/main/java/org/oppia/android/data/persistence/PersistentCacheStore.kt
```diff
@@ -272,6 +272,12 @@ class PersistentCacheStore<T : MessageLite> private constructor(
       return currentPayload.copy(state = CacheState.IN_MEMORY_ONLY)
     }
 
+    // It's possible for multiple load requests to happen simultaneously, but finish out of order.
+    // Attempting to load the cache a second time can result in a number of problems:
+    // - Losing in-memory cache state.
+    // - Duplicating loaded on-disk state (since "mergeFrom()" is used for loading).
+    if (currentPayload.state != CacheState.UNLOADED) return currentPayload
```
BenHenning, line 279: TODO: Include in PR description as callout.

domain/BUILD.bazel
```diff
@@ -216,8 +216,8 @@ TEST_DEPS = [
     "//testing/src/main/java/org/oppia/android/testing/data:async_result_subject",
     "//testing/src/main/java/org/oppia/android/testing/data:data_provider_test_monitor",
     "//testing/src/main/java/org/oppia/android/testing/logging:event_log_subject",
-    "//testing/src/main/java/org/oppia/android/testing/logging:fake_sync_status_manager",
     "//testing/src/main/java/org/oppia/android/testing/logging:sync_status_test_module",
+    "//testing/src/main/java/org/oppia/android/testing/logging:test_sync_status_manager",
```
BenHenning, line 220: TODO: Mention rename in PR description.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsController.kt
```diff
@@ -38,11 +58,29 @@ class AnalyticsController @Inject constructor(
...
+  @BackgroundDispatcher private val backgroundDispatcher: CoroutineDispatcher,
+  @EnableLearnerStudyAnalytics private val enableLearnerStudyParam: PlatformParameterValue<Boolean>
 ) {
+  // NOTE TO DEVELOPER: This log store should not be lazy since it needs to be primed as early as
+  // possible. When this is created shouldn't affect event record integrity, but it can affect how
+  // the sync status manager reports progress since it won't have a data source to properly monitor
+  // for changes.
   private val eventLogStore =
-    cacheStoreFactory.create("event_logs", OppiaEventLogs.getDefaultInstance())
+    cacheStoreFactory.create("event_logs", OppiaEventLogs.getDefaultInstance()).also { store ->
```
BenHenning, line 72: TODO: Mention this issue in PR description.

utility/src/test/java/org/oppia/android/util/logging/SyncStatusManagerTestBase.kt
```diff
@@ -0,0 +1,364 @@
...
+ * be used by implementation test suites of [SyncStatusManager].
+ *
+ * NOTE TO DEVELOPERS: This is essentially the only time inheritance is good practice for tests
+ * since JUnit doesn't provide a way to compose the tests themselves, only behaviors around the
+ * tests.
+ */
+// Same parameter value: helpers reduce test context, even if they are used by 1 test.
+// Function name: test names are conventionally named with underscores.
+@Suppress("SameParameterValue", "FunctionName")
+abstract class SyncStatusManagerTestBase {
```
BenHenning, line 36: TODO: Mention this pattern in the PR description.

BenHenning:
@seanlip PTAL at the PR as a whole since you have more context on the chain of changes leading up to this, and the nature of the problems being fixed.

@rt4914 OR @MohitGupta121 PTAL at the UI changes (I'll follow up via chat to see which one of you might have availability).

@adhiamboperes PTAL specifically at the analytics related changes, but feel free to look beyond that if you have time.

The only CI failure at the moment is our standard out-of-memory Gradle flake which I'll restart after I wake up next. I'll also address any open comments, but please err on the side of approving the PR if you only have small concerns (since I'd really like to merge this in order to simplify the release this is tied to since it needs to be cherry-picked).

Please let me know if you have any specific concerns.

Edit: Also, I have self-reviewed the entire PR.

MohitGupta121 (approved):
@BenHenning LGTM, I test on my emulator and from the video mention in PR looks perfect from UI side. Also pass review on the .xml files code. (But ignore the dark mode side, that I'll do in my revisit of all xml files for darkmode)
Thanks.

app/src/main/java/org/oppia/android/app/administratorcontrols/learneranalytics/ControlButtonsViewModel.kt
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * Note that a [totalEventsToUpload] value of 0 indicates there are 0 left in the **current**
+   * operation, but there may have been events to upload in the past.
+   *
+   * @property eventsUploaded the number of events that have uploaded so far
+   * @property totalEventsToUpload the total number of events to upload, or none
+   */
+  data class ForceSyncProgress(val eventsUploaded: Int = 0, val totalEventsToUpload: Int = -1) {
+    /** Returns whether there are events that can be uploaded. */
+    fun hasEventsToUpload(): Boolean = totalEventsToUpload != -1
```
seanlip: This feels a bit weird to me, not sure what a default of -1 means (and why it's not 0 instead). Maybe worth documenting.

app/src/main/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileLearnerIdItemViewModel.kt
```diff
@@ -51,6 +66,129 @@ class ProfileLearnerIdItemViewModel private constructor(
...
+
+  /**
+   * Stats structure to provide context on the synchronization & uploading information of a specific
+   * profile's events.
+   *
+   * @property learnerStats the stats corresponding to learner-specific events
+   * @property uncategorizedStats the stats corresponding to events not tied to an individual
+   *     profile, nor ``null`` if these aren't computed for this user
+   */
+  data class UploadEventsStats(
```
seanlip: Weird for a class name to start with a verb; I thought I was looking at functions when seeing this referenced elsewhere. Perhaps UploadedEventsStats or EventsUploadStats (if these include events that haven't been uploaded yet)?

Possibly also ProfileSpecificEventsUploadStats to reflect the profile-specific nature of this class (unlike categorized events below).

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsController.kt
```diff
@@ -38,11 +58,29 @@ class AnalyticsController @Inject constructor(
...
   private val translationController: TranslationController,
+  private val dataProviders: DataProviders,
   @EventLogStorageCacheSize private val eventLogStorageCacheSize: Int,
-  @BlockingDispatcher private val blockingDispatcher: CoroutineDispatcher
+  @BlockingDispatcher private val blockingDispatcher: CoroutineDispatcher,
+  @BackgroundDispatcher private val backgroundDispatcher: CoroutineDispatcher,
+  @EnableLearnerStudyAnalytics private val enableLearnerStudyParam: PlatformParameterValue<Boolean>
 ) {
+  // NOTE TO DEVELOPER: This log store should not be lazy since it needs to be primed as early as
+  // possible. When this is created shouldn't affect event record integrity, but it can affect how
```
seanlip: Minor suggestion: When --> The timestamp at which

(it's a bit of a "garden path sentence" otherwise)

model/src/main/proto/oppia_logger.proto
```diff
@@ -22,6 +23,11 @@ message EventLog {
   // The activity context of the event.
   Context context = 4;
 
+  // The profile to which this event corresponds, or empty if this event is not tied to a particular
```
seanlip, line 26: I'd suggest not going out of order (i.e. putting this after "7" instead of between "4" and "5"), it'd be harder for someone who adds more to the proto later to follow the sequential numbering and they might miss that "8" exists.

seanlip (approved):
@BenHenning This looks pretty good to me, thanks. Left some small notes but nothing blocking!

adhiamboperes (approved):
I have mainly focused on the analytics bit, it LGTM.

app/src/main/java/org/oppia/android/app/administratorcontrols/learneranalytics/ControlButtonsViewModel.kt
BenHenning: Added a constant to hide the sentinel value (and switched it to something more obviously wrong, INT_MIN), and added documentation to contextualize it.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsController.kt
BenHenning: Actually, I'm not sure that change makes sense here. The distinction isn't about timestamp but instead about when in the log store's lifecycle it's created relative to an outside observer (i.e. the sync status manager). I've rewritten this part of the documentation to avoid the potential ambiguity since it seems there were several ways to interpret the wording as it's currently written.

app/src/main/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileLearnerIdItemViewModel.kt
BenHenning: Good points. I like the suggestion of ``ProfileSpecificEventsUploadStats``--renamed accordingly.

model/src/main/proto/oppia_logger.proto
BenHenning: This actually gets a bit tricky since it's often more beneficial to logically organize proto fields vs. numerically (as numeric just provides an indication of _when_ something was added, but not its contextual importance relative to other fields). I'd rather not make the suggested change since it'd be inconsistent with other fields (such as the ``oneof`` in the nested ``Context.activity_context below`` in this proto).

I scanned through protobuf best practices and didn't see anything on managing the next ID (other than a suggestion to try and minimize the number of fields in a proto). I've seen other teams that manage especially large protos use a comment to track the next ID, but this would require CI enforcement to make sure it actually stays up-to-date.

Since there's not an obvious solution and this proto isn't very large (meaning it's not hard to find the next index), I'm opting to leave this as-is. Please let me know if you have any follow-up thoughts.

BenHenning:
Thanks @MohitGupta121 @adhiamboperes and @seanlip!

@seanlip I resolved your conversation threads since you approved, but feel free to go back over them. If you have any follow-up concerns or thoughts, I'll address them in a follow-up PR. Note that I didn't make changes per https://github.com/oppia/oppia-android/pull/4896#discussion_r1132700423 since it wasn't clear to me on what we should actually do here, so that one might be worth looking at in particular.

@rt4914 I think you might not have time to review this on short notice, so I'll mark this as a PR to follow-up review on just to make sure everything looks good for codeowners from your perspective.

BenHenning:
Force merging since @rt4914 won't be able to review this before we need the changes in as it's time-sensitive. Will ask for a post-merge approval if it's needed.

Issue #4362, opened 2022-05-16
When reading text size is extra large, resume lesson page and revision tabs content is seen in normal size


KolliAnitha:
**Describe the bug**
When reading text size is extra large, resume lesson page and revision tabs content is seen in normal size

**To Reproduce**
1. Go to 'Navigation menu'
2. Click on 'Options'- Change Reading text size to Extra large
3. Open any exploration which is in progress
4. The Resume page will be seen in normal text size

**Expected behavior**
Resume lesson page should also be in Extra large size
In the Revision tab, click on any tile the content is displayed in normal text size

![Screenshot_2022-05-16-16-35-00-35_943a62cb4c6fb83e010e1c2e82766a17.jpg](https://user-images.githubusercontent.com/101634267/168581188-f38f3c3b-6981-4f52-a3ea-47fd7ac0eebe.jpg)

Environment:
Device name: One plus Nord2 5G
Android version : Android 11
App version: 0.7-alpha-6c08d882d3

BenHenning:
This is a known issue, though I don't think there's anything tracking it currently. Adding to the CLAM team's board for triaging.

BenHenning:
For searching context, this issue was found in 0.7-alpha (MR4).

KevinGitonga:
Closed this since it is possibly the same issue as #4710 whereby resume lesson screen buttons are not scaling as expected when font size and display are set to max on device settings.

adhiamboperes:
Re-opening this since it is not actually tied to  https://github.com/oppia/oppia-android/issues/4710 even though the screenshots here have the same bug as  https://github.com/oppia/oppia-android/issues/4710.

adhiamboperes:
As per [issuecomment](https://github.com/oppia/oppia-android/issues/4710#issuecomment-1433999958), the app controlled reading text size is limited to in-lesson. Do we have plans to resolve this? (Adding this to the backlog).

Issue #3833, opened 2021-09-23
"Start over" button missing left border in the UI


srushtirk:
**Describe the bug**
The button "Start Over" after returning to an incomplete lesson in the app is missing left border around it or the left side of the button is possibly being cut off.

**To Reproduce**
1. Go to a lesson in the app
2. Play at least 2 cards and leave the lesson
3. Click on the incomplete lesson and see a screen that prompts to start over vs. continue from where you left off
4. See error in the 'Start Over' button

**Expected behavior**
'Start Over" button should have all borders consistent around it.

**Demonstration**
![image](https://user-images.githubusercontent.com/27875087/134443760-b1f61352-c91e-4db7-afc4-7864711e820f.png)

**Environment**
- Device/emulator being used: Honor-6X
- Android or SDK version (e.g. Android 5 or SDK 21): Android 7
- App version: 0.6-alpha-908d57b2b1

**Additional context**

BenHenning:
I believe this isn't a blocker for MR3 (& am thus removing it from our blocker list). If I got that wrong, please re-add it @srushtirk.

srushtirk:
Note that this issue is not seen on Pixel (3, 4XL, 5) devices (Android 11+).

BenHenning:
For searching context, this issue was found in 0.6-alpha (MR3).

KevinGitonga:
Closing this since it is possibly the same issue as #4710 whereby resume lesson screen buttons are not scaling as expected when font size and display are set to max on device settings.

PR #4899, opened 2023-03-13
Fix issue with resume buttons not scaling as expected

### LINKED ISSUE #4710: Improve support for scaling in resume lesson screen
**Describe the bug**
When we increase the font size to very large and open resume lesson screen the button's won't scaled properly. 

**To Reproduce**
1. Increase font size to very large (largest)
2. Open resume lesson's screen

**Expected behavior**
Buttons in resume lesson screen should scale properly when we have maximum font size. 

**Demonstration**
image

**Environment**
- App version (you can get this through system app settings or via the admin controls menu in-app): 1.0

KevinGitonga:
## Explanation
Fixes #4710 and #3833. The second issue was closed as it is possibly a duplicate of #4710

Updating width on both the buttons width 0dp actually gives them equal width enabling the button text and child drawables to scale as expected on max display and font config. Main issue is that the currently used width "144dp" doesn't work when the device's display and font config is set to max. This update works better and is per @adhiamboperes comments on previous PR submitted, The previous fix involved using a LinearLayout which had other caveats.

## Essential Checklist

## For UI-specific PRs only

## Before and after screenshots demo
Nexus 5x API 29
| Before Portrait            |  After Portrait |
:-------------------------:|:-------------------------:
 | 

Google Pixel 3xl
| Before Portrait            |  After Portrait |
:-------------------------:|:-------------------------:
 |

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -411,7 +411,7 @@
   </style>
   <!-- Start Over Lesson Button -->
   <style name="SecondaryButton" parent="TextAppearance.AppCompat.Widget.Button">
-    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_width">0dp</item>
     <item name="android:layout_height">wrap_content</item>
     <item name="android:paddingEnd">12dp</item>
     <item name="android:paddingStart">12dp</item>
@@ -429,7 +429,7 @@
   </style>
   <!-- Continue Lesson Button -->
   <style name="ContinueLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
-    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_width">0dp</item>
     <item name="android:layout_height">wrap_content</item>
     <item name="android:drawablePadding">4dp</item>
     <item name="android:paddingEnd">12dp</item>


BenHenning (changes requested):
Thanks @KevinGitonga.

Two things:
1. Does this actually fix #4362? It seems that issue is tied to also scaling the "resume lesson" screen and isn't particularly tied to the buttons.
2. Won't this '0dp' fix only work if the button is used in a constraint layout (since 0dp means 'match constraint')? I'm concerned with defaulting it at such a high level in styles since other uses of the button (both now and in the future) could lead to the button not displaying at all. Similarly, we need to make sure the min size of buttons is always at least 48dp per accessibility requirements, so you may need to add a ``minWidth`` to the styles as well.

KevinGitonga:
> Thanks @KevinGitonga.
> 
> Two things:
> 
> 1. Does this actually fix #4362? It seems that issue is tied to also scaling the "resume lesson" screen and isn't particularly tied to the buttons.
> 2. Won't this '0dp' fix only work if the button is used in a constraint layout (since 0dp means 'match constraint')? I'm concerned with defaulting it at such a high level in styles since other uses of the button (both now and in the future) could lead to the button not displaying at all. Similarly, we need to make sure the min size of buttons is always at least 48dp per accessibility requirements, so you may need to add a `minWidth` to the styles as well.

Thanks @BenHenning they seem related basing on the screenshots on the PR will double check on the revision tab which is mentioned on issue #4362 and provide update on this.

KevinGitonga:
> Thanks @KevinGitonga.
> 
> Two things:
> 
> 1. Does this actually fix #4362? It seems that issue is tied to also scaling the "resume lesson" screen and isn't particularly tied to the buttons.
> 2. Won't this '0dp' fix only work if the button is used in a constraint layout (since 0dp means 'match constraint')? I'm concerned with defaulting it at such a high level in styles since other uses of the button (both now and in the future) could lead to the button not displaying at all. Similarly, we need to make sure the min size of buttons is always at least 48dp per accessibility requirements, so you may need to add a `minWidth` to the styles as well.

Hey @BenHenning 
RE 1. they are not related so i have removed reference to #4362 from this fix as the other is related to local inapp changes in text size not working although the image demonstration convey's a different message that shows relation to #4710.
RE 2. The buttons already have a minimum width set per this file https://github.com/oppia/oppia-android/blob/develop/app/src/main/res/values/styles.xml. The buttons are also currently referenced only from constraintlayout.

BenHenning:
I'm not sure that fully addresses point (2) actually. Defaulting to using constraint layout is problematic since it won't be clear to users of this style in the future that the button must be in a constraint layout, leading to a potentially brittle style.

It also occurs to me that the current style very much wants the button to be the size it's set to (as the minimum is set to the same size). Perhaps we should be considering alternatives to resizing such as wrapping the buttons. Per https://m2.material.io/components/buttons#anatomy button text shouldn't ever wrap in Material design as it can hurt text legibility, so wrapping would probably be more in-line with Material as well.

PR #4914, opened 2023-03-22
Fix #4879: Add NPS survey models

### LINKED ISSUE #4879: Create Model Protos for Surveys
### Required Models for Android NPS Survey
All the descriptions are included in the technical design doc.

## Write new Protos for:
1. Survey
2. SurveyQuestion
3. SurveyQuestionAnswer
4. SurveyUserAnswer
5. ExplorationTimeDatabase
6. EphemeralSurveyQuestion
7. QuestionType
8. SurveyCompletionStatus
9. ExplorationTimeDatabase

## Make Modifications to existing protos
1. Profile
2. OppiaLogger
3. Screens

adhiamboperes:
## Explanation
Fix #4879.
This is Part 1 of 7 PRs that the add the new survey feature.

This PR adds the proto model and BUILD definitions required for the feature.

## Essential Checklist


model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+    // True if this is the last question in the survey.
+    bool terminal_question = 6;
+  }
+}
+
+// Corresponds to a question that the user has previously completed.
+message CompletedQuestion {
+  // The SurveyUserAnswer given to complete the question.
+  SurveyUserAnswer user_answer = 1;
+}
```
BenHenning: Nit: please add EOF newline.

model/src/main/proto/survey.proto
BenHenning, line 1: Are you also going to add the BUILD declarations for this proto in this PR, too?

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Corresponds to a single survey.
+message Survey{
```
BenHenning: Nit: space before the ``{`` for consistency. Ditto elsewhere.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  MULTIPLE_CHOICE_WITH_OTHER = 2;
+
+  // Corresponds to a question which takes a free-text response.
+  FREE_FORM = 3;
+}
+
+// Corresponds to the semantic name given to each question.
+enum QuestionName{
+  // The question name is unknown.
+   QUESTION_NAME_UNSPECIFIED = 0;
```
BenHenning: Two nits: extra space before the ``Q``, and missing a newline after the ``;``. Ditto elsewhere..

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  QuestionName question_name = 4;
+
+  // A list of multi-choice answer options for a given question.
+  repeated  SurveyQuestionAnswer answer_list = 5;
+}
+
+// The list of types a question can have that is used to determine how it is presented by the UI.
+enum QuestionType{
+  // The type of question is unknown.
+  QUESTION_TYPE_UNSPECIFIED=0;
```
BenHenning: Nit: one space on either side of the ``=``, and ditto elsewhere.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+// Represents a question that is part of the Survey.
+message SurveyQuestion{
+  // The ID of the survey question.
+  string question_id = 1;
+
+  // The type of question.
+  QuestionType question_type = 2;
+
+  // The actual text value of the question.
+  string  question = 3;
```
BenHenning: Nit: an extra space after repeated--please remove. Ditto elsewhere.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  VERY_DISAPPOINTED = 1;
+
+  // Corresponds to the answer for Disappointed.
+  DISAPPOINTED = 2;
+
+  // Corresponds to the answer for  Not Disappointed.
+  NOT_DISAPPONTED = 3;
+
+  // Corresponds to the answer for  N/A - Don't use Oppia Anymore.
+  NA_DONT_USE_OPPIA_ANYMORE = 4;
```
BenHenning: 'NA' is a bit ambiguous--perhaps 'NOT_APPLICABLE' would be clearer?

Also, wouldn't the context be "won't use Oppia anymore"? If they aren't using the app, they presumably wouldn't be filling out this survey.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  // Corresponds to a question which takes a free-text response.
+  FREE_FORM = 3;
+}
+
+// Corresponds to the semantic name given to each question.
+enum QuestionName{
+  // The question name is unknown.
+   QUESTION_NAME_UNSPECIFIED = 0;
+  // Corresponds to the name of the user type question.
+  QUESTION_USER_TYPE = 1;
```
BenHenning: For this & other proto enums, we don't need to prefix-qualify each one since they'll be part of a generated Java enum. I suggest removing the ``QUESTTION_`` prefix from these constants (except the first one since the unspecified constant should always repeat the enum name per proto convention).

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  QuestionType question_type = 2;
+
+  // The actual text value of the question.
+  string  question = 3;
+
+  // The semantic name of the question used to identify it in the analytics dashboard.
+  QuestionName question_name = 4;
+
+  // A list of multi-choice answer options for a given question.
+  repeated  SurveyQuestionAnswer answer_list = 5;
```
BenHenning: How is this populated for question types other than multiple choice?

I'm wondering if using a oneof might be a cleaner way to model these question answers.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  SURVEY_COMPLETED = 4;
+}
+
+// Corresponds to a survey question that exists ephemerally in the UI and will disappear
+// once the user finishes or exits the survey session.
+message EphemeralSurveyQuestion{
+  // The actual question the user is currently viewing.
+  SurveyQuestion  question = 1;
+
+  //Indicates whether this question has a previous question that has been answered,
```
BenHenning: Nit: Exactly one space should be between ``//`` and the first letter. Ditto elsewhere.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+    // A previous question completed by the user.
+    CompletedQuestion completed_question = 5;
+
+    // True if this is the last question in the survey.
+    bool terminal_question = 6;
+  }
+}
+
+// Corresponds to a question that the user has previously completed.
+message CompletedQuestion {
```
BenHenning: Will this ever contain anything other than ``SurveyUserAnswer``? If not, we could just reference that directly.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  //  Indicates possible states a question can take depending on whether a user needs to provide an
+  //  answer or has navigated to a previous question, or has reached the last question.
+  oneof question_type{
+    // A pending question that requires an answer to continue.
+    bool pending_question = 4;
+
+    // A previous question completed by the user.
+    CompletedQuestion completed_question = 5;
+
+    // True if this is the last question in the survey.
```
BenHenning: Probably more like: "Present if this is the last question in the survey (the actual value of the boolean does not matter)."

We may set it to true, but just its presence as being set is the key indicator (per how we handle this with other bool oneof cases).

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+
+  // Corresponds to the optional part of the survey being completed.
+  SURVEY_COMPLETED = 4;
+}
+
+// Corresponds to a survey question that exists ephemerally in the UI and will disappear
+// once the user finishes or exits the survey session.
+message EphemeralSurveyQuestion{
+  // The actual question the user is currently viewing.
+  SurveyQuestion  question = 1;
```
BenHenning: Nit: one space between the type & name.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  SURVEY_DECLINED = 1;
+
+  // Corresponds to the survey being exited before the mandatory questions are completed.
+  SURVEY_STARTED_NOT_COMPLETED = 2;
+
+  // Corresponds to the mandatory part of the survey being completed.
+  SURVEY_MANDATORY_PART_COMPLETED = 3;
+
+  // Corresponds to the optional part of the survey being completed.
+  SURVEY_COMPLETED = 4;
```
BenHenning: Perhaps 'FULLY_COMPLETED' might better correspond to 'MANDATORY_PART_COMPLETED'?

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
+syntax = "proto3";
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Corresponds to a single survey.
+message Survey{
+  // The ID of the survey.
```
BenHenning: For this & other IDs, we should try to explain what the ID format is or if it's an opaque type provided by the server (which I don't think it is in this case), and its uniqueness guarantees.

Technically, our audience for these doc comments are both callers to set and read the values, but I personally find it more helpful to focus on the proto's contract on the "read" side and assume the setter will comply with it.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+  QUESTION_DETRACTOR_FEEDBACK = 6;
+}
+
+// Represents the multiple choice answer available for a question.
+message SurveyQuestionAnswer{
+  // The ID of this answer option.
+  string question_answer_id = 1;
+
+  // The semantic name of the question this answer option is tied to.
+  QuestionName question_name = 2;
```
BenHenning: Why do these values need to be replicated for only a multiple choice option?

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+
+  // Corresponds to  the open-ended question for a NPS passive.
+  QUESTION_PASSIVE_FEEDBACK = 5;
+
+  // Corresponds to  the open-ended question for a NPS detractor.
+  QUESTION_DETRACTOR_FEEDBACK = 6;
+}
+
+// Represents the multiple choice answer available for a question.
+message SurveyQuestionAnswer{
```
BenHenning: I don't think this name well conveys that it represents a multiple choice option--I suggest renaming.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,182 @@
...
+// Represents an answer that was selected by a user for a survey question.
+message SurveyUserAnswer{
+  //  The ID of the survey to which this answer is tied.
+  string  survey_id = 1;
+
+  // Semantic name of the question the response is tied to.
+  QuestionName question_name = 2;
+
+  // The value of the answer the user gave.
+  string user_answer = 3;
```
BenHenning: What is this value for multiple choice--the actual label text? If so, is that what we're supposed to be submitting to the backend (vs. an index or something)? If so, how will this work with translated versions of the multiple choice options?

BenHenning (changes requested):
Thanks @adhiamboperes! Took a first pass--PTAL.

model/src/main/proto/survey.proto
adhiamboperes: Done

model/src/main/proto/survey.proto
adhiamboperes: Done

model/src/main/proto/survey.proto
adhiamboperes: Fixed

model/src/main/proto/survey.proto
adhiamboperes: Fixed.

model/src/main/proto/survey.proto
adhiamboperes: Fixed

model/src/main/proto/survey.proto
adhiamboperes: Corrected

model/src/main/proto/survey.proto
adhiamboperes: Fixed

model/src/main/proto/survey.proto
adhiamboperes: You're right about the context.
I was hoping to avoid extremely long names, but `NOT_APPLICABLE_WONT_USE_OPPIA_ANYMORE` is self documenting.

model/src/main/proto/survey.proto
adhiamboperes: This sounds good.

model/src/main/proto/survey.proto
adhiamboperes: Noted. I have fixed this.

model/src/main/proto/survey.proto
adhiamboperes: It is my understanding that the `answer_list` field can be left empty, as the respondent will provide their answer as free-form text.

It is not clear to me why a oneof _answer_ would be needed, since I have modelled the question around QuestionType.  Could you be suggesting that instead of modelling around QuestionType, I might consider using oneof _answer_ ?

model/src/main/proto/survey.proto
adhiamboperes: At the moment, that's all it contains, so I have  referenced it directly.

model/src/main/proto/survey.proto
adhiamboperes: Refactored the comment.

model/src/main/proto/survey.proto
adhiamboperes: I have updated the documentation for the IDs. Please let me know if this covers everything.

model/src/main/proto/survey.proto
adhiamboperes: Renamed to `SurveyAnswerOption`

model/src/main/proto/survey.proto
adhiamboperes: This message represent the answer in general, and according to the current implementation, is empty when the question is not a multiple choice question.

model/src/main/proto/survey.proto
adhiamboperes: Yes, I'll have this in the next update of the PR when I get back.

model/src/main/proto/survey.proto
adhiamboperes: Yes, it was meant to be the actual label text.
I see your point about translated text, and I think it might be advisable to use the associated enums instead.

adhiamboperes (commented):
Hey @BenHenning, aside from these two review comments, I have addressed all of your previous comments and I'd like you to take another look. Thanks.

model/src/main/proto/survey.proto
```diff
@@ -47,31 +51,33 @@ enum QuestionType{
...
   // Corresponds to  the open-ended question for a NPS detractor.
-  QUESTION_DETRACTOR_FEEDBACK = 6;
+  DETRACTOR_FEEDBACK = 6;
 }
 
 // Represents the multiple choice answer available for a question.
-message SurveyQuestionAnswer{
-  // The ID of this answer option.
+message SurveyAnswerOption {
+  // The ID of this answer option. It is a string value that uniquely identifies the answer option
```
BenHenning: Is it unique to the survey instance, profile, device installation, or global? Ditto elsewhere for new properties where we need to guarantee some level of uniqueness.

model/src/main/proto/survey.proto
```diff
@@ -163,20 +171,15 @@ message EphemeralSurveyQuestion{
...
   //  answer or has navigated to a previous question, or has reached the last question.
-  oneof question_type{
+  oneof question_type {
     // A pending question that requires an answer to continue.
     bool pending_question = 4;
 
-    // A previous question completed by the user.
-    CompletedQuestion completed_question = 5;
+    // The SurveyUserAnswer given to complete the previous question.
+    SurveyUserAnswer user_answer = 5;
```
BenHenning:
```suggestion
    SurveyUserAnswer completion_user_answer = 5;
```

Perhaps to keep the previous context that this represents a "completed" state?

model/src/main/proto/survey.proto
```diff
@@ -6,36 +6,40 @@ option java_package = "org.oppia.android.app.model";
 option java_multiple_files = true;
 
 // Corresponds to a single survey.
-message Survey{
-  // The ID of the survey.
+message Survey {
+  // The ID of the survey. This is a string value that uniquely identifies the survey.
+  // The ID is guaranteed to be unique across all surveys.
```
BenHenning: Unique across all surveys globally, for the device installation, or the profile?

model/src/main/proto/survey.proto
BenHenning: Enums sound good if we always have a fixed set of text choices to choose from.

model/src/main/proto/survey.proto
BenHenning: I think this relates to the oneof case above, so I'll respond there.

model/src/main/proto/survey.proto
BenHenning: One possible issue is that an empty proto value (whether it be text or answer_list) is ambiguous--does it mean "this isn't a multiple choice" or "there aren't any choices defined for this multiple choice question"? It makes it harder to reason about a model if it can take different forms but share properties that should be ignored for certain types (this can result in extra unnecessary error checking needing to be added, or potential for conceptual mix-ups later on).

In this case, you could replace QuestionType with a ``oneof`` and better separate the properties relevant to each type of question, e.g.:

```proto
message SurveyQuestion {
  string question_id = 1;
  QuestionName question_name = 2;
  oneof question_type {
    SurveyQuestionOptionList multiple_choice = 3;
    SurveyQuestionOptionList multiple_choice_with_other = 4;
    bool free_form_text = 5;
  }
}

message SurveyQuestionOptionList {
  repeated SurveyQuestionOption options = 1;
}
```

(Note the extra ``SurveyQuestionOptionList`` is needed because proto doesn't currently support having repeated values in ``oneof``s).

Note how in the case of free-form text we now no longer have to even question what to do with the multiple choice options as a field because the field can't exist in that case.

BenHenning (commented):
Thanks @adhiamboperes! Took an incremental pass. PTAL.

Once all of these are addressed, I'll take one more full pass just to make sure I didn't miss anything.

model/src/main/proto/survey.proto
adhiamboperes: SurveyIds are unique globally, since that is what we will use to group the responses given by a unique profile - if needed.

model/src/main/proto/survey.proto
adhiamboperes: At the moment, all the MCQs have a fixed set of answers, so I have opted to create a oneof of each of those answer representations that this field could take.

model/src/main/proto/survey.proto
adhiamboperes: I took this suggestion

model/src/main/proto/survey.proto
adhiamboperes: Your proposed `SurveyQuestion` message structure using `oneof` looks good. In this case, I can create a seperate UiModel for QuestionType which will be used for rendering.

model/src/main/proto/survey.proto
adhiamboperes: Re: translations, I have added a language code field to the question so that we can query open-ended responses based on this. It occurred to me that a generic query will just show all the answers without the language context.

model/src/main/proto/survey.proto
adhiamboperes: This is done

model/src/main/proto/oppia_logger.proto
```diff
@@ -2,7 +2,6 @@ syntax = "proto3";
 
 package model;
 
-import "exploration.proto";
```
BenHenning, line 5: Ah, I was removing this in one of my chained PRs since I also noticed the warning--thanks for removing it. :)

model/src/main/proto/survey.proto
```diff
@@ -61,43 +60,64 @@ enum QuestionName {
...
-  string user_answer = 3;
+  oneof answer {
+    // The enum representation of the type of user answer selected.
+    UserTypeAnswer user_type = 3;
+
+    // The enum representation of the market fit answer selected.
+    MarketFitAnswer market_fit = 4;
+
+    // The integer value representing the nps score selected by the user.
+    int32 nps = 5;
```
BenHenning: Perhaps ``nps_score``, instead, per the comment? Ditto above.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+
+package model;
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Corresponds to a single survey.
+message Survey {
+  // The ID of the survey. This is a string value that uniquely identifies the survey.
+  // The ID is guaranteed to be unique across all surveys globally for the purpose of analysis.
```
BenHenning:
```suggestion
  // The ID should ostensibly be unique across all surveys globally for the purpose of analysis.
```

We don't actually make guarantees since that would require a remote service like Firestore keeping track of used IDs. GUIDs are statistically unlikely to overlap, but that isn't the same as a guarantee.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+    MarketFitAnswer market_fit = 4;
+
+    // The integer value representing the nps score selected by the user.
+    int32 nps = 5;
+  }
+}
+
+// Represents the different types of users who could be taking the survey.
+// Consists of answer options for the user type question.
+enum  UserTypeAnswer {
```
BenHenning:
```suggestion
enum UserTypeAnswer {
```

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+  // Corresponds to a user who is a parent.
+  PARENT = 3;
+
+  // Corresponds to a user who is a learner.
+  LEARNER = 4;
+}
+
+// Represents general opinion of the user about the Oppia android app.
+// Consists of answer options for the market fit question.
+enum  MarketFitAnswer {
```
BenHenning:
```suggestion
enum MarketFitAnswer {
```

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+
+  // Indicates whether this question has a previous question that has been answered,
+  // or if this is the initial question.
+  bool has_previous_question = 2;
+
+  // Indicates whether this question has an upcoming question that can be navigated to.
+  bool has_next_question = 3;
+
+  //  Indicates possible states a question can take depending on whether a user needs to provide an
+  //  answer or has navigated to a previous question, or has reached the last question.
```
BenHenning:
```suggestion
  // Indicates possible states a question can take depending on whether a user needs to provide an
  // answer or has navigated to a previous question, or has reached the last question.
```

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+  }
+}
+
+// Represents the different types of users who could be taking the survey.
+// Consists of answer options for the user type question.
+enum  UserTypeAnswer {
+  // User type unknown.
+  USER_TYPE_UNSPECIFIED = 0;
+
+  // Corresponds to an unspecified or unknown type of user.
```
BenHenning, line 125: Out of curiosity, why not use ``USER_TYPE_UNSPECIFIED`` for this case? Why have a different OTHER enum?

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+  string language_code = 7;
+}
+
+// Represents a list of answer options for a multiple choice question.
+message SurveyQuestionOptionList {
+  repeated SurveyQuestionOption options = 1;
+}
+
+// Corresponds to the semantic name given to each question.
+enum QuestionName {
```
BenHenning: Should this be ``SurveyQuestionName`` for consistency elsewhere in this file? It seems like everything tends to start with 'survey'.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+    // The enum representation of the market fit answer option.
+    MarketFitAnswer market_fit = 5;
+
+    // The integer value representing the nps score option.
+    int32 nps = 6;
+  }
+}
+
+// Represents an answer that was selected by a user for a survey question.
+message SurveyUserAnswer {
```
BenHenning: Should this maybe be ``SurveyQuestionAnswer``? The 'User' part is throwing me off--I think this is meant to have parity with ``SurveyQuestionOption`` above, so a rename might help connect them better.

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,204 @@
...
+
+    // Corresponds to a question with multiple choices and a free-text 'other' option.
+    SurveyQuestionOptionList multiple_choice_with_other = 5;
+
+    // Indicates whether this question takes a free-text response.
+    bool free_form_text = 6;
+  }
+
+  // The language code to which this question is localized.
+  string language_code = 7;
```
BenHenning: We should use OppiaLanguage for this purpose, see:

https://github.com/oppia/oppia-android/blob/1bbadac452d87fa7748bcb5d5ad3a8b446473848/model/src/main/proto/languages.proto#L9

model/src/main/proto/survey.proto
BenHenning: Given the change, is this field still needed? I'm assuming the question type provides enough context to allow the UI to pick app strings to show for each question, so it isn't clear to me why this string will be needed.

model/src/main/proto/survey.proto
BenHenning: This was addressed in the latest changes.

BenHenning (commented):
Thanks @adhiamboperes! Took another pass--PTAL.

model/src/main/proto/survey.proto
adhiamboperes: I've updated both fields.

model/src/main/proto/survey.proto
adhiamboperes: Thanks, this makes sense. I can see why we really can't make the guarantee when thousands of devices are involved.

model/src/main/proto/survey.proto
adhiamboperes: Fixed

model/src/main/proto/survey.proto
adhiamboperes: Fixed.

model/src/main/proto/survey.proto
adhiamboperes: Fixed.

model/src/main/proto/survey.proto
adhiamboperes: > It seems like everything tends to start with 'survey'.

This is only because some of the message names have been used in other messages. 

I have updated this message to start with survey too, for consistency.

model/src/main/proto/survey.proto
adhiamboperes: I'm inclined to go with `SurveySelectedAnswer` so that it can make sense also in the context of `EphemeralSurveyQuestion`

model/src/main/proto/survey.proto
adhiamboperes: It is no-longer needed. Please counter-check, I removed it in a previous commit.

model/src/main/proto/survey.proto
adhiamboperes: Done

model/src/main/proto/survey.proto
adhiamboperes: Loosely speaking, it would be more difficult to distinguish between responses that were genuinely unspecified and those that were intended to be "other" choices. However the implementation should have checks to ensure each question has a selected answer before proceeding to the next question.

model/src/main/proto/survey.proto
BenHenning: That sounds good--thanks!

model/src/main/proto/survey.proto
```diff
@@ -0,0 +1,206 @@
...
+message SurveyQuestionOption {
+  // The ID of this question option. It is a string that uniquely identifies the question option
+  // within the context of a survey question. It is only unique within the scope of a survey.
+  string question_option_id = 1;
+
+  // The semantic name of the question this answer option is tied to.
+  SurveyQuestionName question_name = 2;
+
+  // The actual text value of the answer option.
+  string label = 3;
```
BenHenning: This part still seems wrong, unless I'm missing something. Since we localize text in the frontend, should this be an enum, instead?

Ditto for 'question' in SurveyQuestion above.

BenHenning (commented):
Thanks @adhiamboperes! The proto is looking really good, just had one comment about two fields left--PTAL.

model/src/main/proto/survey.proto
adhiamboperes: I have removed both string fields since we already have `SurveyQuestionName` in SurveyQuestion and `oneof answer`  in SurveyQuestionOption.

BenHenning (approved):
Thanks @adhiamboperes. This LGTM!

I'll leave it to you to merge since it'll affect your downstream chain PR.

PR #4916, opened 2023-03-22
Fix #4880: Add NPS Survey Gating Logic

### LINKED ISSUE #4880: Add Gating Logic for Android NPS Survey
### What
Selecting the target of the survey is depending on the below gating criteria (all must be true):

1. Spent at least 5 minutes [in aggregate] in a topic/lesson module
2. Completed at least one “Progress check within a Lesson/Chapter” successfully (i.e. they get the right response to the question - implying some kind of learning has happened). 
3. Completed tasks (1) & (2) in last 10 days [to ensure freshness]
4. Not exposed to survey in last 30 days [to prevent over-surveying + to ensure we get new opinions/inputs]

### How
The gating criteria is tracked separately for each unique profile on a device. We will use PersistentCacheStore’s createPerProfile() [method](https://github.com/oppia/oppia-android/blob/c77280545cdf4aad1918679603605a3e6b55b81d/data/src/main/java/org/oppia/android/data/persistence/PersistentCacheStore.kt#L405) to create a new cache file for the total time spent in an ongoing topic: exploration_time_database mapped to the [ExplorationTimeDatabase](https://docs.google.com/document/d/1_k3NKm8wHd90gY647idV5k3uPHfoWks1xlaQb7jCh60/edit#heading=h.6we3zk2ll3nq). Since there can be multiple ongoing topics for every profile, the database should also contain the topicId field that uniquely identifies the topic being played.

We will add a survey_last_shown_timestamp - the timestamp that the survey was last triggered for this profile, to the ProfileDatabase along with a [SurveyCompletionStatus](https://docs.google.com/document/d/1_k3NKm8wHd90gY647idV5k3uPHfoWks1xlaQb7jCh60/edit#heading=h.43jfyb190y7u) enum field which stores which action a user took on a survey which can be any of: declined to respond, completed the survey, responded but did not complete. The timestamp is saved based on the user’s locale in  milliseconds, as is the convention in [OppiaLocale](https://github.com/oppia/oppia-android/blob/292a53a700ba43b5fbe1184b38e04a1e5774428d/utility/src/main/java/org/oppia/android/util/locale/OppiaLocale.kt#L129-L172).

We will record the time when an exploration has been started or ended by getting the current timestamp when the DataProviders returned by [ExplorationDataController.startPlayingNewExploration()](https://github.com/oppia/oppia-android/blob/71e769bec2b853d51996432e31f8bec232f16928/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt#L74) and [ExplorationDataController.stopPlayingExploration()](https://github.com/oppia/oppia-android/blob/71e769bec2b853d51996432e31f8bec232f16928/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationDataController.kt#L240) have returned a success result.

To ensure that we are only logging the active time a learner is spending in the exploration, we can take a PerformanceMetricsController parameter in the ExplorationTimeController which we will use to get when the app is in the background by calling the [PerformanceMetricsController.getIsAppInForeground()](https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/PerformanceMetricsController.kt#L195) method. This value is updated by the [onAppInBackground()](https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt#L107) lifecycle event method in ApplicationLifecycleObserver. We can then use this value in a method that records that active time in an exploration has been paused.

When the [onAppInForeground()](https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt#L93) event is fired, the [isAppInForeground](https://github.com/oppia/oppia-android/blob/c7349c3c24f99b16c9f5115fd410fbe9c827e1d0/domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/PerformanceMetricsController.kt#L37) field is updated to true, which we can use to restart recording of the active exploration time.

adhiamboperes:
## Explanation
Fix #4880.

This is PR 2 of 6 planned PRs.

The changes introduced in this PR include:
### Tracking of Active Time in an Exploration
We are introducing a timing mechanism that kicks off when an exploration is started, and is stopped when an exploration is ended or the app goes into the background. This allows us to improve the precision of tracking when a learner is actively learning.

### Tracking of when a survey is shown on a profile
Record the last time when a profile was served the survey.

### Increased granularity in the time of day categorization.
We are breaking morning into `EARLY_MORNING` and `MID_MORNING`, and evening into `EVENING` and `LATE_NIGHT`

### Survey Gating
Using the 3 criteria tracked above, we decide who gets served with the survey.

### New Callback Pattern
We are introducing a set of daggerbound callbacks to improve sharing information more cleanly between domain controllers.

### Test file changes
Approximately 150 test files have changed to include `ExplorationprogressModule`, which is the application-scoped module used to provide dependencies needed to track active learning time.

### New Platform Parameters
We are itroducing two new platform parameters to be able to remotely tweak the minimimum aggregate time and the grace period required for survey gating.
 
### Proto Changes
- Added a `survey_last_shown_timestamp_ms` to the `profile.proto` which represents the epoch timestamp in milliseconds when the nps survey was previously shown on this profile.
- `topic.proto`
```
 // Top-level proto used to store the aggregate learning duration timestamps in a topic,
// on per-profile basis.
message TopicLearningTimeDatabase {
  // Map from topic ID to TopicLearningTime.
  map<string, TopicLearningTime> aggregate_topic_learning_time = 1;
}

// Represents the active learning time spent in a topic.
message TopicLearningTime {
  // The ID of the topic for which this learning time has been recorded.
  string topic_id = 1;

  // The total time spent by a learner in this topic, in milliseconds.
  int64 topic_learning_time_ms = 2;

  // Represents the last Unix epoch timestamp when the topic_aggregate_learning_time_ms was updated,
  // in milliseconds, coinciding with the last time the learner played an exploration in this topic.
  int64 last_updated_time_ms = 3;
}
```

### Domain BUILD.bazel
Added the `survey` package to the `MIGRATED_PROD_FILES` to be excluded automatically from the top-level utility library. 

### Test File Exemptions
Exempted `ExplorationProgressListener` and `ApplicationLifecycleListener` from this check.

## Essential Checklist


BenHenning (commented):
Thanks @adhiamboperes. A few things:
- If this is blocked on another PR to go in first, please update the PR description to include that information (e.g. "[Blocked: #4914]" at the end).
- If the PR description or PR code isn't complete yet, or if CI checks are failing prior to initial review, we should mark the PR as draft. (It's okay for check base branch check to fail, though, since this is a chained PR).
- Since this PR seems to be in draft (per the incomplete description & commented out code), please clarify what specific parts you'd like me to look at. I don't usually perform full reviews on draft PRs since they have a high chance of changing, so it's usually better to get the PR to code completion before a full review. However, I'm happy to look at & provide feedback on specific parts that you're unsure of (which can be especially useful if those are areas of code that are likely to change with feedback).

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
```diff
@@ -0,0 +1,219 @@
...
+
+  private fun createCacheStore(lastUpdatedTimeMs: Long) {
+    cacheFactory.createPerProfile(
+      CACHE_NAME,
+      createTopicLearningTimeDatabase(lastUpdatedTimeMs),
+      profileId
+    )
+  }
+
+  private fun createTopicLearningTimeDatabase(lastUpdatedTimeMs: Long): TopicLearningTimeDatabase {
```
adhiamboperes: I couldn't think of a way to enforce when the last record was updated, so I created this function. Could you suggest an alternative way that I can use to test this more robustly?

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
```diff
@@ -0,0 +1,219 @@
...
+  fun testRecordAggregateTopicLearningTime_returnsSuccess() {
+    val recordAggregateTimeProvider =
+      explorationActiveTimeController.recordAggregateTopicLearningTime(
+        profileId, TEST_TOPIC_ID_0, SESSION_LENGTH
+      )
+    monitorFactory.waitForNextSuccessfulResult(recordAggregateTimeProvider)
+  }
+
+  @Test
+  fun testSaveSessionLength_previousAggregateIsStale_overwritesOldAggregateWithCurrentLength() {
```
adhiamboperes: This test always fails and I am unable to figure out why.
I have tried to debug using breakpoints, and the the code always hits the wrong if/else branch -- i.e the one that evaluates to timestamp not stale, and I cannot tell whether it is an issue with test set up or API. The API looks correct to me.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt
```diff
@@ -0,0 +1,10 @@
+package org.oppia.android.domain.oppialogger.analytics
+
+/** Listener for when the app goes to the foreground or the background. */
+interface ApplicationLifecycleListener {
```
adhiamboperes, line 4: One implementation of these listeners that I attempted was to implement it in `ApplicationLifecycleObserver`, similar to [ApplicationStartupListener](https://github.com/oppia/oppia-android/blob/1ba98ddb66590e7879ec01928ac6e59cd9da4408/domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt#L80). The only reason it does not work is that an implementation to start and stop a timer here will be triggered even when the user is not currently in the exploration screen.

I am not certain how best to enforce a check for current screen when the app is in foreground, to ensure the timer is only started when in ExplorationActivity.

I think the alternative would be to use the activity's lifecycle onPause/onResume. What do you think about this.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
```diff
@@ -0,0 +1,219 @@
...
+
+  @Test
+  fun testSaveSessionLength_previousAggregateIsNotStale_incrementsOldAggregateByCurrentLength() {
+    oppiaClock.setFakeTimeMode(FakeOppiaClock.FakeTimeMode.MODE_FIXED_FAKE_TIME)
+    val currentTime = oppiaClock.getCurrentTimeMs()
+
+    // set aggregate learning time last updated timestamp to yesterday.
+    val lastUpdatedTimeMs = currentTime.minus(TimeUnit.DAYS.toMillis(1))
+
+    createCacheStore(lastUpdatedTimeMs)
```
BenHenning:
```suggestion
    explorationActiveTimeController.setExplorationSessionStarted()
    testCoroutineDispatchers.advanceTimeBy(TimeUnit.DAYS.toMillis(1))
    explorationActiveTimeController.setExplorationSessionPaused(profileId, TEST_TOPIC_ID_0)
```

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -111,7 +111,8 @@ class ExplorationProgressController @Inject constructor(
   private val loggingIdentifierController: LoggingIdentifierController,
   private val profileManagementController: ProfileManagementController,
   private val learnerAnalyticsLogger: LearnerAnalyticsLogger,
-  @BackgroundDispatcher private val backgroundCoroutineDispatcher: CoroutineDispatcher
+  @BackgroundDispatcher private val backgroundCoroutineDispatcher: CoroutineDispatcher,
+  private val explorationProgressListener: ExplorationProgressListener
```
```suggestion
  private val explorationProgressListeners: Set<explorationProgressListener>
```

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressModule.kt
```diff
@@ -0,0 +1,13 @@
...
+import dagger.Provides
+
+/** Module to provide dependencies corresponding to exploration progress. */
+@Module
+class ExplorationProgressModule {
+  @Provides
+  fun provideExplorationProgressListener(
+    explorationProgressListenerImpl: ExplorationProgressListenerImpl
+  ): ExplorationProgressListener = explorationProgressListenerImpl
+}
```
BenHenning:
```suggestion
@Module
interface ExplorationProgressModule {
  @Binds
  @IntoSet
  fun provideExplorationProgressListener(
    impl: ExplorationProgressListenerImpl
  ): ExplorationProgressListener
}
```

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt
BenHenning: As discussed in meeting today, I suggest refactoring the existing ApplicationLifecycleObserver such that we have a general handler for forwarding these lifecycle events, e.g.:

```kotlin
class ApplicationLifecycleObserver @Inject constructor(
  private val applicationLifecycleListeners: Set<ApplicationLifecycleListener>
): ApplicationStartupListener, LifecycleObserver {
  override fun onCreate() {
    ProcessLifecycleOwner.get().lifecycle.addObserver(this)
  }

  @OnLifecycleEvent(Lifecycle.Event.ON_START)
  fun onAppInForeground() {
    applicationLifecycleListeners.forEach(ApplicationLifecycleListener::onAppInForeground)
  }

  /** Occurs when application goes to background. */
  @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
  fun onAppInBackground() {
    applicationLifecycleListeners.forEach(ApplicationLifecycleListener::onAppInBackground)
  }
}
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
BenHenning: As discussed & found during the meeting, it seems the failure is due to the PersistentCacheStore not actually writing the expected contents ahead of expecting the value to be available. I also suggest following a pattern more like AppStartupStateControllerTest wherein we only use the ExplorationActiveTimeController's API for setting & verifying state (since we can use two instances of the test suite's TestApplicationComponent to simulate existing state from a "previous app run").

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
BenHenning: As discussed in an other comment, this probably isn't ideal since it leaks an implementation detail of the controller (using PersistentCacheStore), and tests are most robust when they focus on verifying a public API's contract.

BenHenning:
@adhiamboperes I followed up to the comments you left. Please let me know if there's anything else you'd like me to look at here.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt
adhiamboperes: I was able to refactor all the tests to use the public apis by simulating the previous run suggested. Thanks.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
adhiamboperes: Got it, thanks!
An interesting thing I encountered was the need to annotate the type argument here with @JvmSuppressWildcards as the [documentation](https://dagger.dev/dev-guide/multibindings.html) warns.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressModule.kt
adhiamboperes: I updated my module in this way

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt
adhiamboperes: I did this refactor and implemented the listeners in the `ExplorationProgressListenerImpl` as discussed, but I encontered these situations:
1. DRY: if the callbacks from both interfaces do the same thing, can we consolidate their implementations?
2. Not sure how the onAppInBackground() callback will receive the `profileId` and `topicId` needed when calling `setExplorationSessionPaused(profileId, topicId)` the callback itself takes no arguments, which is ideal since we fire it in the ApplicationLifecycleObserver like in your example above.

Here is what the impl looks like now:

```
@Singleton
class ExplorationProgressListenerImpl @Inject constructor(
  private val explorationActiveTimeController: ExplorationActiveTimeController
) : ExplorationProgressListener, ApplicationLifecycleListener {
  override fun onExplorationSessionStarted() {
    explorationActiveTimeController.setExplorationSessionStarted()
  }

  override fun onExplorationSessionPaused(profileId: ProfileId, topicId: String) {
    explorationActiveTimeController.setExplorationSessionPaused(profileId, topicId)
  }

  override fun onAppInForeground() {
    explorationActiveTimeController.setExplorationSessionStarted()
  }

  override fun onAppInBackground() {
    print("app is backgrounded")
  }
}
```
This way, with some logging, I can see that  `explorationActiveTimeController.setExplorationSessionStarted() `is always called, as soon as the app  comes into the foreground, regardless of the screen.

3. Where can I move this implementation so that the background/foreground behaviour is only reacted to within an exploration context?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
```diff
@@ -0,0 +1,27 @@
+package org.oppia.android.domain.exploration
+
+import org.oppia.android.app.model.ProfileId
+import org.oppia.android.domain.oppialogger.analytics.ApplicationLifecycleListener
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Implementation of [ExplorationProgressListener] */
```
BenHenning: Calling listener implementations generic "impl"s poses a problem of vagueness. Listeners are meant to exist more than once, so each implementation ought to be specific to whichever domain purpose it's fulfilling (versus utilities which generally only have one implementation). I suggest renaming this class to be more specific as to what it's responsible for rather than just naming it an implementation of the listener class (especially since it's an impl of multiple listeners, so there's clearly more specificity possible here).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+class ExplorationProgressListenerImpl @Inject constructor(
+  private val explorationActiveTimeController: ExplorationActiveTimeController
+) : ExplorationProgressListener, ApplicationLifecycleListener {
+  override fun onExplorationSessionStarted() {
+    explorationActiveTimeController.setExplorationSessionStarted()
+  }
+
+  override fun onExplorationSessionPaused(profileId: ProfileId, topicId: String) {
+    explorationActiveTimeController.setExplorationSessionPaused(profileId, topicId)
+  }
```
BenHenning: Is this 'paused' or 'ended'?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+
+  override fun onAppInForeground() {
+    explorationActiveTimeController.setExplorationSessionStarted()
```
BenHenning: Isn't this only true if there's an ongoing session? I'd also expect onAppInBackground to play a role in 'pausing' logic.

I think either the listener needs to determine session state based on its various signals & pass this along to the controller, or it needs to forward its received values to the controller and let it perform the determination.

Also a separate thought: you could alternatively do away with this class entirely and just have the controller implement the listeners & bind to the corresponding listener sets.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -570,6 +573,7 @@ class ExplorationProgressController @Inject constructor(
         recomputeCurrentStateAndNotifyAsync()
       }.launchIn(CoroutineScope(backgroundCoroutineDispatcher))
       explorationProgress.advancePlayStageTo(LOADING_EXPLORATION)
+      explorationProgressListeners.forEach(ExplorationProgressListener::onExplorationSessionStarted)
```
BenHenning: For this & the below callback: does exploration ID never need to be passed?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt
BenHenning: I think (2) and (3) relate to another comment I left in the listener impl--the correct way to do this is to probably use the signals of session activity & app activity independently to determine the current session state and use that for decision making. I think this also addresses (1) since I see the two listeners as providers of completely distinct information that we're using together to determine a particular domain state from which we can make decisions.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
```diff
@@ -273,21 +275,56 @@ class ExplorationActivityPresenter @Inject constructor(
...
           when (it) {
             is AsyncResult.Pending -> oppiaLogger.d("ExplorationActivity", "Stopping exploration")
             is AsyncResult.Failure ->
               oppiaLogger.e("ExplorationActivity", "Failed to stop exploration", it.error)
             is AsyncResult.Success -> {
               oppiaLogger.d("ExplorationActivity", "Successfully stopped exploration")
-              backPressActivitySelector()
-              (activity as ExplorationActivity).finish()
+              if (isCompletion) {
+                shouldShowSurveyDialog(profileId, topicId)
```
BenHenning: Perhaps instead: maybeShowSurveyDialog? The current name reads as a function that will return whether to show it as opposed to actually showing it.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
```diff
@@ -273,21 +275,56 @@ class ExplorationActivityPresenter @Inject constructor(
...
+        }
+      )
+  }
+
+  private fun createSurveyDialog(): AlertDialog {
+    val builder = AlertDialog.Builder(activity)
+    return builder.setTitle("NPS Survey")
+      .setMessage("Survey Intro message")
+      .create()
+  }
```
BenHenning: I assume this is temporary as it'll be filled in properly in a later PR? Reminder to remove this code in that case.

model/src/main/proto/profile.proto
```diff
@@ -80,6 +80,9 @@ message Profile {
 
   // Determines if the in-exploration continue button animation has been seen by the user or not.
   bool is_continue_button_animation_seen = 15;
+
+  // Represents the last time the nps survey was shown on this profile.
+  int64 survey_last_shown_timestamp = 17;
```
BenHenning: Please update comment & field name to indicate the time unit being used.

model/src/main/proto/topic.proto
```diff
@@ -671,3 +671,23 @@ message SubtopicRecord {
   // The old title corresponding to the subtopic (as a string).
   reserved 1;
 }
+
+// Top-level proto used to store the aggregate learning duration timestamps in a topic,
+// per-profile basis.
```
BenHenning:
```suggestion
// on per-profile basis.
```

model/src/main/proto/topic.proto
```diff
@@ -671,3 +671,23 @@ message SubtopicRecord {
...
+  // Map from topic ID to TopicLearningTime.
+  map<string, TopicLearningTime> aggregate_topic_learning_time = 1;
+}
+
+// Represents the active learning time spent in a topic.
+message TopicLearningTime {
+  // The ID of the topic for which this learning time has been recorded.
+  string topic_id = 1;
+
+  // The total time spent by a learner in this topic.
```
BenHenning:
```suggestion
  // The total time spent by a learner in this topic, in milliseconds.
```

model/src/main/proto/topic.proto
```diff
@@ -671,3 +671,23 @@ message SubtopicRecord {
...
+  int64 topic_learning_time_ms = 2;
+
+  // The last time the topic_aggregate_learning_time_ms was updated, coinciding with the last time
+  // the learner played an exploration in this topic.
```
BenHenning: I suggest indicating that this is a Unix epoch timestamp in milliseconds (as opposed to a duration or any other type of time value that could be stored here).

scripts/assets/test_file_exemptions.textproto
```diff
@@ -646,6 +646,8 @@ exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rule
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/ratioinput/RatioInputModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/textinput/TextInputRuleModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressModule.kt"
```
BenHenning: This actually could be tested--I suggest adding a basic test suite to demonstrate that the binding is set up correctly & as expected by the module (we have a few other module tests in the codebase that you could look at for reference).

utility/src/main/java/org/oppia/android/util/locale/OppiaLocale.kt
```diff
@@ -155,6 +155,9 @@ sealed class OppiaLocale {
 
     /** Represents different times of day. */
     enum class TimeOfDay {
+      /** Corresponds to the user's very early morning time. */
+      EARLY_MORNING,
+
       /** Corresponds to the user's morning time. */
       MORNING,
```
BenHenning: Perhaps this should be 'MID_MORNING' or 'LATE_MORNING' to contrast with 'EARLY_MORNING'?

utility/src/main/java/org/oppia/android/util/locale/MachineLocaleImpl.kt
```diff
@@ -50,9 +50,11 @@ class MachineLocaleImpl @Inject constructor(
 
   override fun getCurrentTimeOfDay(): TimeOfDay {
     return when (oppiaClock.getCurrentCalendar().get(Calendar.HOUR_OF_DAY)) {
-      in 4..11 -> TimeOfDay.MORNING
+      in 4..8 -> TimeOfDay.EARLY_MORNING
+      in 9..11 -> TimeOfDay.MORNING
       in 12..16 -> TimeOfDay.AFTERNOON
-      in 0..3, in 17..23 -> TimeOfDay.EVENING
+      in 17..22 -> TimeOfDay.EVENING
+      in 0..3, 23 -> TimeOfDay.NIGHT
```
BenHenning: Perhaps 'LATE_NIGHT'?

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
```diff
@@ -0,0 +1,89 @@
...
+ * Controller for retrieving survey gating criteria and deciding if a survey should be shown.
+ */
+class SurveyGatingController @Inject constructor(
+  private val profileManagementController: ProfileManagementController,
+  private val oppiaClock: OppiaClock,
+  private val explorationActiveTimeController: ExplorationActiveTimeController,
+  private val machineLocale: OppiaLocale.MachineLocale
+) {
+
+  private val surveyGracePeriodMillis = TimeUnit.DAYS.toMillis(30)
```
BenHenning: Is this something we should maybe implement as a platform parameter? It could be useful to be able to remotely tweak it in the future (and platform parameters make it easier to swap out this value while testing).

BenHenning (changes requested):
Thanks @adhiamboperes!

I actually realized a bit late that this PR is still in draft (so I was reviewing it as though it were fully in review). I stopped short of reviewing the controllers in depth, and the tests at all. I took a look at everything else, though, and left some thoughts. PTAL.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
adhiamboperes: Ended.

model/src/main/proto/profile.proto
adhiamboperes: I updated this.

model/src/main/proto/topic.proto
adhiamboperes: Noted

model/src/main/proto/topic.proto
adhiamboperes: Updated.

model/src/main/proto/topic.proto
adhiamboperes: I updated this comment

scripts/assets/test_file_exemptions.textproto
adhiamboperes: Noted, thanks.

utility/src/main/java/org/oppia/android/util/locale/OppiaLocale.kt
adhiamboperes: I went with MID_MORNING, given the range of hours under consideration.

utility/src/main/java/org/oppia/android/util/locale/MachineLocaleImpl.kt
adhiamboperes: Yes, this is definitely more appropriate.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
adhiamboperes: Since the purpose here is to start and stop the session timer, I have renamed it to `ExplorationSessionTimerController`

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
adhiamboperes: At the moment, I am not using the exploration ID for any logic, which is why I am not passing it.
The granularity currently required is at topic id level.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
adhiamboperes: I'll keep this in mind for when I implement the dialog, although this will be in theb next PR in the chain. I have already removed this placeholder code.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
adhiamboperes: Yes, I have removed the code.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListenerImpl.kt
adhiamboperes: I have added a check for an exploration being in progress.

I have opted to keep this class, the listeners determine the session state and passes it along to the controller.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
adhiamboperes: I did find out from testing that platform parameters would be preferrable here.

BenHenning:

Also @adhiamboperes it looks like there are CI failures--it might be good to address these first.

adhiamboperes:
> Also @adhiamboperes it looks like there are CI failures--it might be good to address these first.

Hi @BenHenning, I have resolved the CI failures.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -0,0 +1,182 @@
...
+    )
+  }
+
+  /**
+   * Records the tracked active exploration time.
+   *
+   * @param profileId the ID corresponding to the profile for which progress needs to be stored
+   * @param topicId the ID corresponding to the topic for which duration needs to be stored
+   * @param sessionDuration the tracked exploration duration between start and pause
+   * @return a [DataProvider] that indicates the success/failure of this record operation
```
adhiamboperes, line 466: Punctuation?

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -530,7 +531,9 @@ class ExplorationProgressController @Inject constructor(
         // processed (if there's a flow).
         else -> AsyncResult.Pending()
       }
-    } catch (e: Exception) { AsyncResult.Failure(e) }
+    } catch (e: Exception) {
+      AsyncResult.Failure(e)
```
adhiamboperes: Revert to one line

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
```diff
@@ -0,0 +1,12 @@
+package org.oppia.android.domain.exploration
+
+import org.oppia.android.app.model.ProfileId
+
+/** Listener for when an exploration is started or paused. */
```
adhiamboperes: `/** Listener for when an exploration is started or ended. */`

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt
adhiamboperes: I modified the implementation so that I receive signals from the two listeners independently then use them to decide the timer state.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleModule.kt
```diff
@@ -18,4 +19,10 @@ class ApplicationLifecycleModule {
   @Provides
   @LearnerAnalyticsInactivityLimitMillis
   fun provideLearnerAnalyticsInactivityLimitMillis(): Long = TimeUnit.MINUTES.toMillis(30)
+
+  @Provides
+  @IntoSet
+  fun provideApplicationLifecycleListeners(
+    impl: ExplorationSessionTimerController
```
adhiamboperes: Rename argument

instrumentation/src/java/org/oppia/android/instrumentation/application/TestApplicationComponent.kt
```diff
@@ -100,7 +101,7 @@ import javax.inject.Singleton
     MetricLogSchedulerModule::class, ActivityRouterModule::class,
     PerformanceMetricsAssessorModule::class, PerformanceMetricsConfigurationsModule::class,
     TestingBuildFlavorModule::class, EventLoggingConfigurationModule::class,
-    CpuPerformanceSnapshotterModule::class
+    ExplorationProgressModule::class, CpuPerformanceSnapshotterModule::class,
```
adhiamboperes: Remove comma

model/src/main/proto/profile.proto
```diff
@@ -80,6 +80,9 @@ message Profile {
 
   // Determines if the in-exploration continue button animation has been seen by the user or not.
   bool is_continue_button_animation_seen = 15;
+
+  // Represents the timestamp in milliseconds when the nps survey was shown on this profile.
```
adhiamboperes:
```suggestion
  // Represents the epoch timestamp in milliseconds when the nps survey was previously shown on this profile.
```

scripts/assets/test_file_exemptions.textproto
```diff
@@ -646,8 +646,10 @@ exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rule
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/ratioinput/RatioInputModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/classify/rules/textinput/TextInputRuleModule.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgress.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetriever.kt"
 exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt"
```
adhiamboperes: Remove excemption and actually add tests

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
```diff
@@ -231,3 +232,42 @@ annotation class EnableInteractionConfigChangeStateRetention
...
+ * provide informed feedback about their experience with the app.
+ */
+const val NPS_SURVEY_MINIMUM_AGGREGATE_LEARNING_TIME_IN_A_TOPIC_IN_MINUTES =
+  "nps_survey_minimum_aggregate_learning_time_in_a_topic_in_minutes"
+
+/**
+ * Default value of the minimum learning time in a topic, in minutes, that indicates a learner has
+ * had sufficient interaction with the app to be able to provide informed feedback about their
+ * experience with the app.
+ * */
```
adhiamboperes: Fix KDoc formatting

utility/src/test/java/org/oppia/android/util/locale/MachineLocaleImplTest.kt
```diff
@@ -299,12 +317,18 @@ class MachineLocaleImplTest {
 
   private companion object {
     // Date & time: Wed Apr 24 2019 08:22:03 GMT.
-    private const val MORNING_UTC_TIMESTAMP_MILLIS = 1556094123000
+    private const val EARLY_MORNING_UTC_TIMESTAMP_MILLIS = 1556094123000
```
adhiamboperes, line 320: Verify tests in this class for exhaustiveness per new classifications

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
```diff
@@ -0,0 +1,58 @@
...
+  private var topicId: String? = null
+
+  override fun onAppInForeground() {
+    this.isAppInForeground = true
+    startSessionTimer()
+  }
+
+  override fun onAppInBackground() {
+    this.isAppInForeground = false
+    if (profileId != null && topicId != null) {
```
adhiamboperes: Test these changes

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -570,6 +573,12 @@ class ExplorationProgressController @Inject constructor(
         recomputeCurrentStateAndNotifyAsync()
       }.launchIn(CoroutineScope(backgroundCoroutineDispatcher))
       explorationProgress.advancePlayStageTo(LOADING_EXPLORATION)
+      explorationProgressListeners.forEach {
+        it.onExplorationStarted(
```
adhiamboperes, line 575: Test that these listeners are called.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -0,0 +1,182 @@
...
+    this.startExplorationTimestampMs = oppiaClock.getCurrentTimeMs()
+  }
+
+  /**
+   * Stops tracking the active learning time in an exploration.
+   *
+   * This method is called when the [ExplorationProgressController.finishExplorationImpl] finishes
+   * executing, or the app goes to the background.
+   */
+  fun setExplorationSessionStopped(profileId: ProfileId, topicId: String) {
```
adhiamboperes: What happens if this is called before a session was started?

adhiamboperes (commented):
Adding some self-review comments.

adhiamboperes (commented):
Adding some self-review comments.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -135,11 +133,11 @@ class ExplorationActiveTimeController @Inject constructor(
...
     return retrieveCacheStore(profileId)
-      .transform(
-        RETRIEVE_AGGREGATE_LEARNING_TIME_PROVIDER_ID
-      ) { learningTimeDb ->
-        learningTimeDb.aggregateTopicLearningTimeMap[topicId]
-          ?: TopicLearningTime.getDefaultInstance()
+      .transformAsync(RETRIEVE_AGGREGATE_LEARNING_TIME_PROVIDER_ID) { learningTimeDb ->
+        return@transformAsync AsyncResult.Success(
+          learningTimeDb.aggregateTopicLearningTimeMap[topicId]
+            ?: TopicLearningTime.getDefaultInstance()
```
BenHenning: This can just be transform() if you're not calling a suspend function or returning a non-success AsyncResult.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -80,7 +78,7 @@ class ExplorationActiveTimeController @Inject constructor(
    * @param profileId the ID corresponding to the profile for which progress needs to be stored
    * @param topicId the ID corresponding to the topic for which duration needs to be stored
    * @param sessionDuration the tracked exploration duration between start and pause
-   * @return a [DataProvider] that indicates the success/failure of this record operation
+   * @return a [DataProvider] that indicates the success/failure of this record operation.
```
BenHenning: Nit: no punctuation at the end of an at-clause's fragment (we only add punctuation if multiple sentences are needed).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
```diff
@@ -2,11 +2,11 @@ package org.oppia.android.domain.exploration
 
 import org.oppia.android.app.model.ProfileId
 
-/** Listener for when an exploration is started or paused. */
+/** Listener for when an exploration is started or ended. */
```
BenHenning:
```suggestion
/** Listener for when an exploration has started or ended. */
```

Nit on verb use.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -1108,6 +1112,75 @@ class ProfileManagementControllerTest {
...
+      PROFILE_ID_1
+    )
+
+    val lastShownTimeMs = monitorFactory.waitForNextSuccessfulResult(
+      profileManagementController.retrieveSurveyLastShownTimestamp(
+        PROFILE_ID_1
+      )
+    )
+
+    assertThat(lastShownTimeMs).isEqualTo(oppiaClock.getCurrentTimeMs())
```
BenHenning: I suggest setting the time explicitly at the start of the test to make it clear that it's the expected current time at this point (since it's not obvious at a glance that getCurrentTimeMs() shouldn't return DEFAULT_SURVEY_LAST_SHOWN_TIMESTAMP_MILLIS which, if it did, could result in one or both of these tests falsely passing in cases when the production code regresses).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleModuleTest.kt
```diff
@@ -66,6 +68,10 @@ class ApplicationLifecycleModuleTest {
     assertThat(inactivityLimitMillis).isEqualTo(TimeUnit.MINUTES.toMillis(30))
   }
 
+  @Test
+  fun testInjectApplicationLifecycleListenerSet_includesExplorationSessionTimerController() {
+    assertThat(lifecycleListeners.any { it is ExplorationSessionTimerController }).isTrue()
+  }
```
BenHenning, line 77: Nit: newline after the brace for member spacing in the class.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationSessionTimerControllerTest.kt
```diff
@@ -0,0 +1,332 @@
...
+    oppiaClock.setFakeTimeMode(FakeOppiaClock.FakeTimeMode.MODE_UPTIME_MILLIS)
+
+    applicationLifecycleObserver.onAppInForeground()
+
+    explorationDataController.startPlayingNewExploration(
+      profileId.internalId, TEST_TOPIC_ID_0, TEST_STORY_ID_0, TEST_EXPLORATION_ID_2
+    )
+    testCoroutineDispatchers.runCurrent()
+
+    testCoroutineDispatchers.advanceTimeBy(SESSION_LENGTH)
```
BenHenning: Here & elsewhere: you can omit the runCurrent before thi scall since advanceTimeBy implies runCurrent (since it runs all tasks up to the scheduled delay).

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
```diff
@@ -0,0 +1,102 @@
...
+    val surveyLastShownCalendar = oppiaClock.getCurrentCalendar()
+    surveyLastShownCalendar.timeInMillis = lastShownTimestampMs
+
+    val currentCalendar = oppiaClock.getCurrentCalendar()
+    val showNextTimestamp = surveyLastShownCalendar.timeInMillis.plus(gracePeriodMillis)
+    val showNextCalendar = oppiaClock.getCurrentCalendar()
+    showNextCalendar.timeInMillis = showNextTimestamp
+
+    return currentCalendar.after(showNextCalendar) ||
+      showNextCalendar == currentCalendar
```
BenHenning: Is a ``Calendar`` really needed for this logic? I'm wondering if it's just checking/managing the Unix timestamp internally. This might be able to be simplified if so since then we can just avoid using ``Calendar`` and work with the timestamps directly.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private fun retrieveSurveyLastShownDate(profileId: ProfileId): DataProvider<Long> {
+    return profileManagementController.retrieveSurveyLastShownTimestamp(profileId)
+      .transformAsync(GET_SURVEY_LAST_SHOWN_TIMESTAMP_PROVIDER_ID) { lastShownTimestampMs ->
+        AsyncResult.Success(lastShownTimestampMs)
+      }
```
BenHenning: Can this not just be ``retrieveSurveyLastShownTimestamp()``? The transformation doesn't seem to be changing anything.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
```diff
@@ -0,0 +1,102 @@
...
+  private fun retrieveAggregateLearningTime(
+    profileId: ProfileId,
+    topicId: String
+  ): DataProvider<Long> {
+    return explorationActiveTimeController.retrieveAggregateTopicLearningTimeDataProvider(
+      profileId, topicId
+    ).transformAsync(GET_TOPIC_LEARNING_TIME_PROVIDER_ID) { topicLearningTime ->
+      val aggregateLearningTimeMs = topicLearningTime.topicLearningTimeMs
+      AsyncResult.Success(aggregateLearningTimeMs)
+    }
```
BenHenning:
```suggestion
    return explorationActiveTimeController.retrieveAggregateTopicLearningTimeDataProvider(
      profileId, topicId
    ).transform(
      GET_TOPIC_LEARNING_TIME_PROVIDER_ID, TopicLearningTime::getTopicLearningTimeMs
    )
```

Can be simplified a bit (and probably shortened if some of the variable names involved were shortened).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
```diff
@@ -0,0 +1,59 @@
...
+ * from [ExplorationProgressListener] and [ApplicationLifecycleListener].
+ */
+@Singleton
+class ExplorationSessionTimerController @Inject constructor(
+  private val explorationActiveTimeController: ExplorationActiveTimeController
+) : ExplorationProgressListener, ApplicationLifecycleListener {
+  private var isAppInForeground: Boolean = false
+  private var explorationStarted: Boolean = false
+  private var profileId: ProfileId? = null
+  private var topicId: String? = null
```
BenHenning: It seems like these variables will be accessed on different threads. These may need to be transferred over to an internal data class combined with either a command pattern (like we use in ``ExplorationProgressController`` via an actor) or to ``InMemoryBlockingCache`` to ensure synchronous access to the data).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
```diff
@@ -0,0 +1,59 @@
...
+import javax.inject.Singleton
+
+/**
+ * Controller to decide when the session timer is started and stopped based on signals received
+ * from [ExplorationProgressListener] and [ApplicationLifecycleListener].
+ */
+@Singleton
+class ExplorationSessionTimerController @Inject constructor(
+  private val explorationActiveTimeController: ExplorationActiveTimeController
+) : ExplorationProgressListener, ApplicationLifecycleListener {
```
BenHenning: Why not have the ``ExplorationActiveTimeController`` itself bind as an instance of these listeners? It seems like the extra wrapper doesn't add much benefit (unless I'm missing something).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
```diff
@@ -0,0 +1,12 @@
+package org.oppia.android.domain.exploration
+
+import org.oppia.android.app.model.ProfileId
+
+/** Listener for when an exploration is started or ended. */
+interface ExplorationProgressListener {
```
BenHenning, line 13: It's worthwhile to document whether this listener may be accessed on multiple threads (so that implementers can take precautions, if needed).

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -570,6 +571,12 @@ class ExplorationProgressController @Inject constructor(
         recomputeCurrentStateAndNotifyAsync()
       }.launchIn(CoroutineScope(backgroundCoroutineDispatcher))
       explorationProgress.advancePlayStageTo(LOADING_EXPLORATION)
+      explorationProgressListeners.forEach {
```
BenHenning, line 574: I suggest adding tests to ``ExplorationProgressControllerTest`` to verify both the session start & end cases to make sure that bound listeners are called when expected and passed the correct data.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -0,0 +1,182 @@
...
+    ).invokeOnCompletion {
+      if (it != null) {
+        oppiaLogger.e(
+          "ExplorationActiveTimeController",
+          "Failed to prime cache ahead of data retrieval.",
+          it
+        )
+      }
+    }
+    return cacheStore
```
BenHenning:
```suggestion
    return cacheStoreMap.getOrPut(profileId) {
      cacheStoreFactory.createPerProfile(
        CACHE_NAME,
        TopicLearningTimeDatabase.getDefaultInstance(),
        profileId
      ).also { cacheStore ->
        cacheStore.primeInMemoryAndDiskCacheAsync(
          updateMode = PersistentCacheStore.UpdateMode.UPDATE_IF_NEW_CACHE,
          publishMode = PersistentCacheStore.PublishMode.PUBLISH_TO_IN_MEMORY_CACHE
        ).invokeOnCompletion {
          if (it != null) {
            oppiaLogger.e(
              "ExplorationActiveTimeController",
              "Failed to prime cache ahead of data retrieval.",
              it
            )
          }
        }
      }
    }
```

Shorter, more idiomatic, and avoids the potential code smell of a forced null check.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
```diff
@@ -0,0 +1,182 @@
...
+      } else {
+        topicLearningTimeBuilder.topicLearningTimeMs += sessionDuration
+      }
+      topicLearningTimeBuilder.lastUpdatedTimeMs = oppiaClock.getCurrentTimeMs()
+
+      val topicLearningTime = topicLearningTimeBuilder.build()
+
+      val topicLearningTimeDatabaseBuilder = topicLearningTimeDatabase.toBuilder()
+        .putAggregateTopicLearningTime(topicId, topicLearningTime)
+      Pair(topicLearningTimeDatabaseBuilder.build(), TopicLearningTimeActionStatus.SUCCESS)
```
BenHenning:
```suggestion
      topicLearningTimeDatabase.toBuilder().apply {
        val topicLearningTime = aggregateTopicLearningTimeMap.getOrDefault(topicId) {
          TopicLearningTime.newBuilder().setTopicId(topicId).build()
        }.toBuilder().apply {
          topicLearningTimeMs = if (isLastUpdatedTimestampStale(lastUpdatedTimeMs)) {
            sessionDuration
          } else topicLearningTimeMs + sessionDuration
          lastUpdatedTimeMs = oppiaClock.getCurrentTimeMs()
        }.build()
        putAggregateTopicLearningTime(topicId, topicLearningTime)
      }.build() to TopicLearningTimeActionStatus.SUCCESS
```

Bunch of small idiomatic simplifications.

BenHenning (approved):
Thanks @adhiamboperes! I Took a full pass on the PR. While I am leaving a fair number of suggestions, many of these are actually superficial architectural suggestions or nits. I do suggest addressing all of them, and perhaps having another team member take a pass on the changes.

That being said, approving the PR since it largely looks good to me and I don't want to block this getting merged while I'm not performing code reviews over the coming weeks.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
adhiamboperes: Changed this to transform as it has no suspend context

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
adhiamboperes: Fixed

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
adhiamboperes: Corrected

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
adhiamboperes: Thanks for this suggestion, it does make a lot of sense.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleModuleTest.kt
adhiamboperes: Fixed

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
adhiamboperes: I was able to compare the timestamps directly without using calendar. I refactored the method.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
adhiamboperes: I did a refactor per the suggestion

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
adhiamboperes: I refactored this and removed ExplorationSessionTimerController and its test file

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
adhiamboperes: I updated this Kdoc

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
adhiamboperes: I accepted this suggestion as is, thanks!

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
adhiamboperes: I accepted this suggestion with some minor changes.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationSessionTimerControllerTest.kt
adhiamboperes: I ommitted all the occurance of runCurrent except for one instance where it is actually needed. In the test `testSessionTimer_explorationStarted_appInBackground_thenInForeground_resumesSessionTimer`, I need all previous operations to run before advancing time. I am using advanceTimeBy to get the clock to  a specific time, as opposed to changing the FakeTimeMode to MODE_FIXED_FAKE_TIME and manually setting time.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
adhiamboperes: I chose to use the command pattern approach since I could not find sufficient references to InMemoryBlockingCache to learn how to use it, due to limited use in the codebase.

This refactor, though it ended up being non-trivial, allowed me to have better control over the timer, since the timer now has its own session that I can start, pause, resume and stop.

adhiamboperes:
Hi @kkmurerwa, could you please take a look at the changes I have made on this PR(commits since last review).

I have attempted to address each comment on it's own commit, and the commit message  body should give you more information about what has changed.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
adhiamboperes: I made this change

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
adhiamboperes: I added one test for this

kkmurerwa:
Hi @adhiamboperes. I will review it tomorrow over the course of the day.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
kkmurerwa: This looks better now. It can be marked as resolved.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
kkmurerwa: This looks good to me.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
kkmurerwa: This is good now.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
kkmurerwa: This looks resolved.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
kkmurerwa: This is good now.

domain/src/main/java/org/oppia/android/domain/survey/SurveyGatingController.kt
kkmurerwa: The changes made look good.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressListener.kt
kkmurerwa: This has now been addressed.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
kkmurerwa: This can be marked as resolved.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
kkmurerwa: The changes LGTM.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationActiveTimeController.kt
kkmurerwa: This looks good now.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
kkmurerwa: I've reviewed the updated `testFetchSurveyLastShownTime_updateLastShownTimeFunctionCalled_returnsCurrentTime` test and it should work well now without the potential risk of false positives.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleModuleTest.kt
kkmurerwa: I am unable to confirm whether the suggested spacing was added.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationSessionTimerControllerTest.kt
kkmurerwa: All `runCurrent()` instances with an `advanceTimeBy()` method just below them seem to have been updated. This is good to go.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
kkmurerwa: This looks done.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationSessionTimerController.kt
kkmurerwa: The logic for this seems to have been moved to the `ExplorationActiveTimeController` class which is consistent with the other changes. The approach you chose also LGTM.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
kkmurerwa: I've reviewed the added test and it covers the required cases. This should be good to go.

kkmurerwa:
Hi @adhiamboperes. I've looked over the comments to make sure all suggestions have been implemented and everything seems good. Once the failing `testGetCurrentState_playInvalidExploration_returnsFailure_logsException` test on the `ExplorationProgressControllerTest` is fixed, this PR should be good to go.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleModuleTest.kt
adhiamboperes: There was previously no newline at ln78. I added it.

adhiamboperes:
> Hi @adhiamboperes. I've looked over the comments to make sure all suggestions have been implemented and everything seems good. Once the failing `testGetCurrentState_playInvalidExploration_returnsFailure_logsException` test on the `ExplorationProgressControllerTest` is fixed, this PR should be good to go.

Thanks @kkmurerwa! I have fixed the failing test and resolved the remaining threads.

adhiamboperes:
@seanlip, could you please code-owner approve and merge this PR for me? Thanks.

seanlip:
Hm, @BenHenning seems to have approved it already, so no further approval is needed from me AFAICT. I'll just merge it!

PR #4925, opened 2023-03-27
Fix part of #59: Migrate to rules_jvm_external 5.1

BenHenning:
## Explanation
Fix part of #59

At a high-level, this PR migrates the project to using rules_jvm_external 5.1 (from the current version 4.1). The main benefit of this is a **much** simpler ``maven_install.json`` file (>50% reduction in file size). This is especially appealing as downstream PRs for #59 make a _lot_ of changes to third-party dependencies, causing ``maven_install.json`` to be regenerated each time. A smaller and more compact format results in fewer actual deltas needing to be checked in between changes.

Introducing support for the new format has required a lot of changes elsewhere in the project, however, including:
- A new Maven install Moshi model needing to be defined.
- ``LicenseFetcher`` was extended to include support for verifying a URL's artifact validity (i.e. by using an HTTP HEAD request) for better robustness and performance when compiling licenses. This is because of a significant difference in the new ``maven_install.json`` format: we no longer know which Maven repositories correspond to which artifacts, so we need to test each artifact against all repositories when checking for licenses. This also has some repercussions on warnings from Bazel's own file downloader (see https://github.com/bazelbuild/rules_jvm_external/issues/349).
- Per the previous point, ``MavenDependenciesRetriever`` required substantial reworking to properly fetch pom information from dependencies. For robustness, it was also updated to use ``ScriptBackgroundCoroutineDispatcher`` to parallelize the fetching with some basic retry built into it (since I noticed on one of my workstations much more flakiness in the HEAD requests and license fetches).
- ``LicenseFetcher`` was also renamed to ``MavenArtifactPropertyFetcher`` since it's actually used for more than just license fetching (besides the new ``HEAD`` check, it's also used for fetching an artifact's POM file when processing the list of licenses for the artifact).

Separate details worthy of noting:
- The new version of rules_jvm_external introduces better support for duplicate dependency error handling and strict visibility.
  - Both of these are just nice-to-have robustness checks against the existing //third_party wrapping setup (so they won't impact developers directly unless changes are made to the //third_party wrapping Bazel code).
  - It also provides the ability to selectively override specific dependency targets (which is needed in #4931 to customize Guava in the build graph).
- Source Jars are no longer being fetched to cut down on the amount of files that need to be downloaded. This may be reverted in the future.
- Some minor dependency management was cleaned up in WORKSPACE.
- The new ``MavenArtifactPropertyFetcherImpl`` & model changes don't have test files for the same reasons the old ``LicenseFetcherImpl`` and models didn't.
- A lot of testing rework was needed for the license checks & update pathways since the means of representing licenses has fundamentally changed with the new ``maven_install.json`` format. The new retry functionality mentioned above also has resulted in a bit more output for these scripts (which results in changes to their tests).
- The new ``MavenCoordinate`` class added as part of ``MavenDependenciesRetriever`` also had a bunch of tests added since this is now a public class of the retriever (and was added to simplify a lot of the coord manipulation work).
- Some minor build time warnings were addressed during development.
- Some of the new licenses for Crashlytics transitive dependencies refer to broken links that I cannot be confident in replacements for, but it seems reasonable to assume all of Crashlytics falls under Apache & the Crashlytics ToS based on the existing Crashlytics GitHub repository.

Important: ``maven_dependencies.textproto`` was significantly changed due to two reasons:
1. The order of the output for ``GenerateMavenDependenciesList`` is different due to a combination of the reworking of ``MavenDependenciesRetriever`` and the new ``maven_install.json`` format.
2. Per my suspicion, I think the previous implementation wasn't including transitive dependencies correctly. The new ``maven_install.json`` format makes it much easier to track these dependencies, and so we'll now be including many more third party license information in released versions of the app.

It is **not** expected that there are any actual version differences in this PR (intentionally since it would be very hard to verify that, and this allows ``maven_install.json`` to be easier to review since it's generated file). This was verified by looking at the conflicts section of both the old and new ``maven_install.json`` files, and the Maven dependencies CI check.

## Essential Checklist

## For UI-specific PRs only
N/A -- This is a build system & script infrastructure change. It's expected to have no user-facing functional side effects.


BenHenning:
Still active.

app/BUILD.bazel
```diff
@@ -1006,7 +1006,6 @@ crashlytics_android_library(
 android_library(
     name = "crashlytics_deps",
     exports = [
-        "//third_party:com_crashlytics_sdk_android_crashlytics",
```
BenHenning: Revert this & the change in version?

BenHenning (commented):
Finished initial self-review pass.

app/BUILD.bazel
BenHenning: Done.

BenHenning (commented):
Self-reviewed most recent changes.

BenHenning:
This looks to be in a good place for review now. PTAL @adhiamboperes.

BenHenning:
@seanlip PTAL for owners group codeowners changes (I think this is mainly needed for ``maven_install.json`` which is auto-generated).

seanlip (approved):
LGTM for one codeowner file: scripts/assets/maven_dependencies.textproto

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
```diff
@@ -179,67 +179,82 @@ class MavenDependenciesRetriever(
...
+      }
+      while (pomCandidatesToUpdate.isNotEmpty()) {
+        println(
+          "Attempt ${++attemptCount} to download POM files for" +
+            " ${pomCandidatesToUpdate.size}/${pomCandidates.size} Maven artifacts..."
+        )
+        pomCandidatesToUpdate.map { pomCandidate ->
+          CoroutineScope(scriptBgDispatcher).async {
+            // Run blocking I/O operations on the I/O thread pool.
+            withContext(Dispatchers.IO) {
```
adhiamboperes, line 206: Is it worth simplifying this to run without wrapping the `withContext` in the outer CoroutineScope?

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
```diff
@@ -179,67 +179,82 @@ class MavenDependenciesRetriever(
...
+            }
+          }
+        }.awaitAll().forEachIndexed { index, pomFileText ->
+          val pomCandidate = pomCandidatesToUpdate[index]
+          pomCandidate.latestPomFileText = pomFileText
+          pomCandidate.downloadTryCount++
+        }
+        pomCandidatesToUpdate = pomCandidates.filter {
+          it.latestPomFileText == null && it.downloadTryCount < 10
+        }
```
adhiamboperes: Creating a  new list for every iteration looks expensive(using filter). Could this be implemented by using mutable lists that can be updated directly?

```kotlin
fun retrieveDependencyListFromPomAsync(
    finalDependenciesList: List<MavenListDependency>
): Deferred<MavenDependencyList> {
    return CoroutineScope(scriptBgDispatcher).async {
        val pomCandidates = finalDependenciesList.map { MavenListDependencyPomCandidate(it) }.toMutableList()
        var attemptCount = 0

        while (pomCandidates.any { it.latestPomFileText == null && it.downloadTryCount < 10 }) {
            val pomCandidatesToUpdate = pomCandidates.filter { it.latestPomFileText == null && it.downloadTryCount < 10 }
            println(
                "Attempt ${++attemptCount} to download POM files for" +
                " ${pomCandidatesToUpdate.size}/${pomCandidates.size} Maven artifacts..."
            )

            try {
                val results = pomCandidatesToUpdate.map { pomCandidate ->
                    withContext(Dispatchers.IO) {
                        mavenArtifactPropertyFetcher.scrapeText(pomCandidate.pomFileUrl)
                    }
                }

                pomCandidatesToUpdate.forEachIndexed { index, pomCandidate ->
                    pomCandidate.latestPomFileText = results[index]
                    pomCandidate.downloadTryCount++
                }
            } catch (e: Exception) {
                println("Error downloading POM file: ${e.message}")
            }
        }

        val failedToDownload = pomCandidates.filter { it.latestPomFileText == null }
        check(failedToDownload.isEmpty()) {
            "Failed to download ${failedToDownload.size}/${pomCandidates.size} POM files:" +
            " $failedToDownload."
        }

        return@async MavenDependencyList.newBuilder().apply {
            this.addAllMavenDependency(
                pomCandidates.map { pomCandidate ->
                    MavenDependency.newBuilder().apply {
                        this.artifactName = pomCandidate.dep.coord.reducedCoordinateString
                        this.artifactVersion = pomCandidate.dep.coord.version
                        // NB: the '!!' is fine here since previous checks should ensure the POM file text is
                        // impossible to be null at this point.
                        this.addAllLicense(extractLicenseLinksFromPom(pomCandidate.latestPomFileText!!))
                    }.build()
                }
            )
        }.build()
    }
}

```

adhiamboperes (changes requested):
Thanks @BenHenning. I have two questions, PTAL.

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
BenHenning: Unless I'm mistaken, I think this will be more efficient because of line 212. It's creating one async job for each candidate to update, then blocks on all finishing before proceeding with the outer operation. The I/O wrapping is just to ensure actual blocking I/O happens on Kotlin's large I/O executor (since blocking I/O is non-cooperative, unlike ``awaitAll()`` which is a coroutine operation).

Kotlin's documentation isn't specifically clear on this point, but I suspect we want to minimize use of the I/O dispatcher to only I/O operations where we can help it so I don't see a huge benefit to wrapping the outer context in the I/O dispatcher (plus, we'd need to re-wrap with our background executor and then I/O again for the individual ``map`` results in order to benefit from the ``awaitAll`` call).

Please let me know if I'm misunderstanding your suggestion.

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
BenHenning: FWIW the lists are fairly small (usually at most ~250 items with our current dependencies if I recall) so I'm not sure the optimization will matter that much.

That being said, I'm not opposed to making it mutable. Implemented a version based on your suggestion though I also noticed there are several lists being created each iteration, anyway (with the ``map`` call), so I'm not sure how much of a difference this makes.

BenHenning (commented):
Thanks @adhiamboperes!

Self-reviewed latest changes & followed-up to review comments. PTAL @adhiamboperes.

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
adhiamboperes: It appears that my suggestion was a bit vague, I was wondering why we needed to use both the background executor and withcontext while scraping the text. Your explanation makes sense, and I'm happy with the update.

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
adhiamboperes: This sounds good.

adhiamboperes (approved):

scripts/src/java/org/oppia/android/scripts/license/MavenDependenciesRetriever.kt
BenHenning: Seems like this is resolved now, unless I'm missing something (so going ahead and closing it).

BenHenning:
Thanks @adhiamboperes! Looks like this is ready to be merged, so syncing with the latest develop and enabling auto-merge.

PR #4937, opened 2023-04-01
Fix #4119, #4120, and part of #59: Upgrade to Kotlin 1.6.10

### LINKED ISSUE #4119: Migrate codebase to Kotlin 1.5

### LINKED ISSUE #4120: Migrate codebase to Kotlin 1.6

BenHenning:
## Explanation
Fixes #4119
Fixes #4120
Fixes part of #59

This PR finishes the migration of the codebase to Kotlin 1.6 (addressing both #4119 and #4120).

Kotlin 1.6 is needed as part of moving rules_kotlin to 1.7.x (which is, in turn, needed in conjunction with Bazel 6.x to enable strict dependency checking which significantly simplifies modularization which is planned for downstream PRs). This PR doesn't actually finish the movement to that version of rules_kotlin, but it does finish moving the codebase to a new enough (and no longer pre-release) version of rules_kotlin to allow using Kotlin 1.6 (over Kotlin 1.4 that the codebase currently uses): version 1.5.0.

Previous PRs (#5400 and #5402) prepared for the changes here by addressing large categories of build warnings that have either arisen from this migration, or from past work. Note that another large category of warnings have also been addressed in this PR: by moving to Kotlin 1.6, there's no longer a runtime incompatibility between the Kotlin SDK and the reflection APIs (which was causing a _lot_ of warning output previously). Between all three PRs, the output is now very clean and free of nearly all build warnings.

To try and keep the warnings clean long-term, this PR introduces warnings-as-errors for both Java and Kotlin code. However, please note some caveats:
- Dagger generated code doesn't follow the Java warnings-as-error flag, so those warnings were cleaned up manually (and will need to be generally watched for, unfortunately).
- The version of rules_kotlin used in this PR doesn't directly support turning on the functionality, but does internally (so a small patch file has been added to augment rules_kotlin). When the codebase is updated to rules_kotlin 1.7.x this patch will no longer be needed.
- To ease development, a build configuration flag was added to disable failure upon encountering build warnings (per https://bazel.build/run/bazelrc and https://bazel.build/docs/configurable-attributes#query-and-cquery as an example), though this needs to be opted into:
  ```sh
  bazel build --config=ignore_build_warnings <target>
  ```

Some other details to note:
- Version 1.6.10 is specifically picked in order to ensure Jetpack Compose compatibility (for preparation of the work being prototyped in #5401 to be compatible with the Oppia Android build environment).
- The vast majority of code in this PR is updating parameterized tests to use a cleaner repeatable annotation pattern that wasn't available in Kotlin 1.4.
- This upgrade absolutely does have runtime implications, but we're relying very heavily on existing automated tests to ensure correctness and no regressions.
- This PR doesn't make an effort to move toward newer Kotlin language features except where forced (API deprecations) or largely wanted (the repeatable annotation change).
- android-spotlight and kotlitex have been updated to support newer versions of Kotlin (as both are custom forks managed in the broader Oppia GitHub organization).
- Gradle files have been updated to match the same dependency versions as Bazel (where it was obvious to make changes; some might still be a bit off).
- The Gradle build configuration was also updated to use Kotlin 1.6.x (otherwise there would be build incompatibilities with Bazel). I think this is the last upgrade we can do for Gradle without upgrading AGP (which will cause us significant issues with the model module, so we're planning on instead dropping Gradle support).
- API changes that needed to be addressed in this PR due to deprecations include: ``String.captialize``, ``String.toLowerCase``, ``String.toUpperCase``, ``SendChannel.offer``, and ``Char.toInt``.
- New API changes that have been leveraged in this PR: ``Flow.lastOrNull`` and ``Deferred.asListenableFuture`` (to replace ``SettableFuture`` for safety; this also resulted in nice simplifications in ``CoroutineExecutorService``).
- The JVM coroutines dependency needed to be split out from Maven and manually imported with some empty internal Java class files since it otherwise has some issues being desugared: https://github.com/bazelbuild/bazel/issues/13553. This is a problem with the Desugarer used in Bazel 4.x (and maybe later versions, so this solution will probably need to kept for a while).
- Some Proguard rule updates were needed due to Kotlin SDK changes--see the Proguard file & comments for specifics.
- Due to dependency changes, the KitKat main dex file was also trimmed down. I'm fairly certain that it's already crashing on startup, so I don't care much about this change--it just needs to build. We plan to remove KitKat entirely eventually, anyway: #5012.
- Jetifier (that is, automatic conversion from support libraries to Jetpack/AndroidX) support was disabled in Gradle. We don't have it enabled in Bazel, and it could potentially encourage strange one-version violations if it was ever actually needed. This is a safer (and likely more performant) change to make.
- Moshi was updated to 1.13 to support the upgrade in Kotlin. This did result in a small configuration change due to its annotation processor being moved. Note that Moshi 1.14 couldn't be supported since it requires Kotlin 1.7+ which requires rules_kotlin 1.7+. This will be an option to upgrade in the future.
- Some improvements and fixes were made in ``FilterPerLanguageResources`` (I think it was outputting something incorrectly before and that's now been fixed as part of a broader logical reworking of the filtering logic).
- ``com.android.support:support-annotation`` was removed as a dependency since it was never used in Bazel, and shouldn't be used (since it's support library and not AndroidX).
- The updates to Moshi and Kotlin dependencies resulted in a bunch of other transitive dependency updates.
- Note that Gradle doesn't have ``allWarningsAsErrors`` enabled since it would require fixing more warnings than is exposed in Bazel, and we're using Bazel builds as the general source of truth for code quality.

## Essential Checklist

## For UI-specific PRs only
N/A -- This is an infrastructural change. While it could inadvertently affect user-facing code, it shouldn't based on the current passing state of automated tests.


.bazelrc
```diff
@@ -6,5 +6,7 @@ build --android_databinding_use_v3_4_args \
     --define=android_standalone_dexing_tool=d8_compat_dx \
     --android_databinding_use_androidx
 
+build --javacopt="-Werror"
```
BenHenning, line 11: Add comment explaining this.

tools/kotlin/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
+# This package defines codebase-wide third-party build configurations for Kotlin.
```
BenHenning: Convert to file docstring.

BenHenning:
Still active.

BenHenning (commented):
Self-reviewed all changes.

BenHenning:
PTAL @seanlip for CODEOWNERS & other OWNERS changes.

PTAL @adhiamboperes. Note that you may want to review the tests in isolation. They make up the very large majority of changes in this PR, and are largely mechanical (due to making some small formatting and syntax changes to move over the ``Iteration`` annotation code), but actually reviewing it may take some time as more than 2,000 iteration lines have been impacted.

seanlip (approved):
LGTM for 4 codeowner files

BenHenning (commented):
Self-review latest change (pulled in a Gradle file update from #4886).

BenHenning (commented):
Self-reviewed latest changes.

FYI I downgraded Kotlin to 1.6.10 for Jetpack Compose support (per the findings in #5401).

tools/kotlin/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
...
+load("@io_bazel_rules_kotlin//kotlin:core.bzl", "define_kt_toolchain", "kt_javac_options", "kt_kotlinc_options")
+
+# This exposes a patch file that retroactively adds opt-in support to kt_kotlinc_options (which was
+# added in a later version of rules_kotlin that's not available for the version of Bazel used at the
+# time of this introduction).
+exports_files(["add_kotlinc_optin_support.patch"])
+
+kt_kotlinc_options(
+    name = "oppia_kotlinc_options",
+    warn = "error",
```
BenHenning: Reminder to update this to be 'report' for local builds and 'error' for CI builds.

BenHenning (commented):
Pulled in upstream changes and self-reviewed. No concerns at the moment.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
```diff
@@ -73,10 +70,9 @@ import kotlin.reflect.KClass
  * ```
  *
  * Finally, regular tests can be added by simply using the JUnit ``Test`` annotation without also
- * annotating with [RunParameterized]. Such tests should not ever read from the
- * [Parameter]-annotated fields since there's no way to ensure what values those fields will
- * contain (thus they should be treated as undefined outside of tests that specific define their
- * value via [Iteration]).
+ * annotating with [Iteration]. Such tests should not ever read from the [Parameter]-annotated
+ * fields since there's no way to ensure what values those fields will contain (thus they should be
+ * treated as undefined outside of tests that specific define their value via [Iteration]).
```
adhiamboperes, line 75: Nit:
```suggestion
 * treated as undefined outside of tests that specifically define their value via [Iteration]).
```

tools/kotlin/BUILD.bazel
adhiamboperes: Is this to be done as part of this PR?

app/src/sharedTest/java/org/oppia/android/app/customview/interaction/MathExpressionInteractionsViewTest.kt
```diff
@@ -459,40 +450,38 @@ class MathExpressionInteractionsViewTest {
...
-    Iteration("math_eq_x-2", "type=MATH_EQUATION", "text=y = x − 2", "expLatex=y = x - 2"),
-    Iteration("math_eq_x*2", "type=MATH_EQUATION", "text=y = x × 2", "expLatex=y = x \\times 2"),
-    Iteration("math_eq_x/2", "type=MATH_EQUATION", "text=y = x ÷ 2", "expLatex=y = x \\div 2"),
-    Iteration(
-      "math_eq_long_exp",
-      "type=MATH_EQUATION",
-      "text=(x^2-1)/(x+1)=y/2",
-      "expLatex=(x ^ {2} - 1) \\div (x + 1) = y \\div 2"
-    )
+  @Iteration("numeric_-1", "type=NUMERIC_EXPRESSION", "text=  -  1           ", "expLatex=-1")
```
adhiamboperes, line 453: For tests with these large spaces in the parameters, are they intentional? Ditto [MathExpressionAccessibilityUtilTest.kt](https://github.com/oppia/oppia-android/pull/4937/files#diff-576b417a50621ba1b2b450deb862de133b43108cf491b587a874eae5df6f57e4) below.

domain/src/test/java/org/oppia/android/domain/classify/rules/numericexpressioninput/NumericExpressionInputMatchesUpToTrivialManipulationsRuleClassifierProviderTest.kt
```diff
@@ -76,19 +73,17 @@ class NumericExpressionInputMatchesUpToTrivialManipulationsRuleClassifierProvide
...
-    Iteration("2*3.14==2*3.14", "answer=2*3.14", "input=2*3.14"),
-    Iteration("2/3==2/3", "answer=2/3", "input=2/3"),
-    Iteration("2/3.14==2/3.14", "answer=2/3.14", "input=2/3.14"),
-    Iteration("2^3==2^3", "answer=2^3", "input=2^3"),
-    Iteration("2^3.14==2^3.14", "answer=2^3.14", "input=2^3.14"),
-    Iteration("sqrt(2)==sqrt(2)", "answer=sqrt(2)", "input=sqrt(2)")
-  )
+  @Iteration("-2==-2", "answer=-2", "input=-2")
+  @Iteration("1+3.14==1+3.14", "answer=1+3.14", "input=1+3.14")
+  @Iteration(" 1 +   3.14 ==1+3.14", "answer= 1 +   3.14 ", "input=1+3.14")
```
adhiamboperes, line 78: Same question about spaces.

adhiamboperes (commented):
Thanks @BenHenning!

Apart from a few clarifications and a nit, everything looks good.

I did also wonder about the comment regarding `FilterPerLanguageResources`. If the changes fixed an error output, is it testable?

tools/kotlin/BUILD.bazel
BenHenning: Yep, will be included in the latest batch of changes.

testing/src/main/java/org/oppia/android/testing/junit/OppiaParameterizedTestRunner.kt
BenHenning: Done in latest.

app/src/sharedTest/java/org/oppia/android/app/customview/interaction/MathExpressionInteractionsViewTest.kt
BenHenning: Yep, this is intentional. It's partly verifying that arbitrary spaces in inputs don't cause any issues with the math parsers used to analyze these raw expressions.

domain/src/test/java/org/oppia/android/domain/classify/rules/numericexpressioninput/NumericExpressionInputMatchesUpToTrivialManipulationsRuleClassifierProviderTest.kt
BenHenning: Yep it's intentional--see comment above.

BenHenning (commented):
Self-reviewed latest changes.

BenHenning:
> Thanks @BenHenning!
> 
> Apart from a few clarifications and a nit, everything looks good.
> 
> I did also wonder about the comment regarding `FilterPerLanguageResources`. If the changes fixed an error output, is it testable?

Thanks @adhiamboperes! I've addressed my earlier + your comments in the latest changes, and brought the PR up-to-date with develop. As part of this, I also updated the PR description to explain how to use the new opt-in build flag for disabling warnigns-as-errors. PTAL.

Regarding ``FilterPerLanguageResources``, I went back to look at the changes and tests. It seems to me that the existing test coverage seems pretty good for verifying that the changes still work as expected (and particularly ``testUtility_whenResourcesAreRemoved_printsDiagnosticInformation``). The exact failures that were originally fixed I have, unfortunately, forgotten as this change was made over a year ago and has kind of sat while the rest of the Bazel chain was being worked on. I think the important takeaway is that it's functionally still working. If you think additional testing is important here, could you please let me know where you think the functional gap is? I'm always happy to add more tests.

BenHenning (commented):
Self-reviewed lint fix.

adhiamboperes (approved):
Thanks @BenHenning! Everything looks good. I'm happy to leave the filter language resources as is.

adhiamboperes:
Merging since everything is up to date.

adhiamboperes:
1 test is flaky:

```
org.oppia.android.data.backends.gae.NetworkLoggingInterceptorTest > testLoggingInterceptor_makeNetworkCall_succeeds FAILED
    java.lang.IllegalStateException: This job has not completed yet
        at kotlinx.coroutines.JobSupport.getCompletedInternal$kotlinx_coroutines_core(JobSupport.kt:1199)
        at kotlinx.coroutines.DeferredCoroutine.getCompleted(Builders.common.kt:100)
        at org.oppia.android.data.backends.gae.NetworkLoggingInterceptorTest.testLoggingInterceptor_makeNetworkCall_succeeds(NetworkLoggingInterceptorTest.kt:99)
```

Issue #3384, opened 2021-06-28
Arrange string resources based on screens 


yashraj-01:
**Describe the bug**

Currently string resources are not organised correctly and are scattered in `strings.xml` file.

This should be fixed by following these points:
1. Make sure that each string is getting used in one specific view only. For example `Home` is part of multiple views like `Navigation Drawer`, `Home-Toolbar`. So for this we should have separate strings for both.
2. All strings related to one activity/fragment should be kept together by mentioning the activity name on top of the strings in a comment. Example: `<!-- ProfileResetPinActivity -->`
3. All activities/fragment sections should be in correct alphabetical order based on activity/fragment name. (The actual strings name are not required to be in alphabetical order but the comments should be in alphabetical order). Example:
```
<!-- HomeActivity -->
<string name="random_string_related_to_home">ABC</string>
<!-- ProfileResetPinActivity -->
<string name="random_string_related_to_profile_resent_pin">ABC</string>
<string name="random_string_related_to_profile_resent_pin_2">XYZ</string>
```

Arrange strings file as per below criteria:

**Screen/Toolbar Titles**
Used in manifest label, toolbar title and activity label test.
Pattern: `activity_name_activity_title`

Example: `administrator_controls_activty_title`

**App Content Strings**
Pattern: `activity_name_view_component_type`

`administrator_controls_activity_profile_management_label`
`administrator_controls_activity_account_actions_label`
`administrator_controls_activity_logout_item_text`
`administrator_controls_activity_app_version_item`

**Shared Strings**
Pattern: `view_component_description`

reading_text_size_item_text_small
reading_text_size_item_medium

**Test Strings**
Pattern: `test_file_name_view_component_type`

Dialogs
Pattern: `dialog_name_view_component_type`

Grouping Pattern: 

  `<!-- AdministratorControlActivity -->`
….
….
  `<!-- GeneralAvailabilityUpgradeNoticeDialogFragment -->`
…..
   `<!-- ProgressDatabaseFullDialog -->`

yashraj-01:
cc: @anandwana001 @rt4914

TheSwarnim:
Hello @rt4914 @anandwana001, please assign this issue to me. I'm working on it

TheSwarnim:
Hello @anandwana001, Sir I've added a PR, please review it.

anandwana001:
@rishidyno Could you please comment here so that I can get your GitHub handle in the assignee section to assign this issue to you?

[...]

PR #4940, opened 2023-04-08
Arranged string resources based on strings #3384

manan1979:
Arranged string resources based on strings #3384


BenHenning:
Hi @manan1979. Welcome! #3384 is currently assigned to @MohitGupta121. If you'd like to help with one of Oppia's issues, I suggest looking for one that hasn't been assigned yet (since then we can avoid people doing duplicate work). Thanks!

PR #4945, opened 2023-04-12
Fix#4881: Build UI for the Android NPS Survey

### LINKED ISSUE #4881: Build UI for the Android NPS Survey
### Screens

Link to [Mocks](https://www.figma.com/file/EuFsYZgT7je3xGuMNyNpMg/Oppia---%2335-Android-in-app-survey?node-id=39%3A1822&t=Bc25Go05SLpm5EKy-1)

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-7ad72452-7fff-66a8-bb8c-c9ba68ab1dd8"><div dir="ltr" style="margin-left:0pt;" align="left">

a | Required screens
-- | --
1 | Survey onboarding screen
2 | Survey Multiple Choice Questions with ‘other option’ screen
3 | Multiple Choice Question screen
4 | Survey NPS question screen
5 | Survey open-ended question screen
6 | Survey thank you screen
7 | Exit survey confirmation screen

</div></b>

This issue also tracks the databinding of these views

adhiamboperes:
## Explanation
Fixes #4881.

This is PR 3 of 6 that adds the UI for the NPS survey as well as completes the [key user flows](https://docs.google.com/document/d/1aIr7cu7V5-9uvhfDgekq-iOXF57hbxEmjiL3mElly0I/edit#heading=h.rkqdyq11md2s). 

#### Main Layouts
- Survey intro 
- Survey fragment for displaying the questions
- Exit confirmation dialog
- Thank you screen

#### Technical decisions
I used the `BindableAdapter.MultiTypeBuilder` from within the fragment to populate the views. This allowed me to reuse the survey question layout while creating individual layouts for each answer option group. 

I created a viewmodel hierachy where:

- The root fragment viewmodel is responsible for retrieving the correct question text, making the decision to allow navigation to the next question by enabling/disabling the next button based on selected answer availability, and updating the progressbar percentage and progress.
- For the answer option layouts, I created a  Super-class, `SurveyAnswerItemViewModel` for representing the different types of different layouts that can exist for the recyclerView responsible for displaying these views.
- Each question option layout has it's own viewmodel that is responsible for retrieving the answer options to display from the app strings file, notifying that an answer has been provided, and retrieving the selected answer for handling.

I used view binding to display all the strings and control navigation.

I also created custom views where using regular views proved to be complex to work with.

#### Custom Views
- SurveyOnboardingBackgroundView - for creating the custom background needed on the intro and outro layouts
- SurveyMultipleChoiceOptionView - for binding lists in multiple choice questions
- SurveyNpsItemOptionView - for binding the custom layout needed for the NPS score question

### Navigation and Progress
I found it difficult to decouple the UI development work from the progress work, since the survey and the questions are generated dynaically so I needed to develop both in tandem.

The following changes pertain to the core survey logic:
#### SurveyController
Creates a survey with a list of questions.

#### SurveyProgressController
Tracks the non-persisted progress of a survey and is responsible for handling the survey session, managing the commands for navigating through and answer submission.

I opted to use a command pattern similar to what is used in the [ExplorationProgressController](https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt).

#### Navigation
The survey can be viewed as a series of linear questions, which we will navigate through using methods defined in the SurveyQuestionDeck, modeled after the [StateDeck](https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt). To start with, we will create an EphemeralSurveyQuestion which represents the question the user is currently viewing. It contains a reference to the next and previous questions if applicable, and a reference to whether it is the last question in the deck. The SurveyQuestionDeck will define the conditions for navigating to the previous and next questions as well as the exceptions thrown in case a navigation action is invalid. 

A SurveyQuestionGraph class will be created to provide functionality for processing the behavior of a ‘next’ navigation action, which evaluates which question is shown next.

### Survey Proto Changes:
- Refactored the order of the enum items in the `UserTypeAnswer` enum so that they result into a properly indexed options list per the mocks.
- Modified the `EphemeralSurveyQuestion` to have a `current_question_index` and a `total_question_count` to make it easy to compute the survey progress.
- Added answer typecase `free_form_answer` to the `SurveySelectedAnswer` message to hold the value of a free text response.
-  Added ` SurveyQuestion optional_question = 3;` field to the survey so that we have a clear seperation between the mandatory and optional questions. This also supports the survey generation architecture.

### Bazel
Created `BUILD` definitions for all new classes and tests.

### Dagger
I changed [approximately 150 test files](https://pbs.twimg.com/media/EAeaeOnU4AcXRQw.jpg) to include `SurveyQuestionModule::class`, though I removed it in a subsequent refactor. When I removed the module and it's usages, it was not a very clean revert because some of the files were not properly formatted before, but they are now -- hence some test files here with only formatting changes.

### Usages, Future Work and Out of Scope
The domain layer of the survey is pretty much flexible to support any survey question, including should the current list be reordered or the number of questions changed. For example, if we wanted to show a subsequent survey on the same profile:

####  Show One Question Only
We would be able to show only one question if we only passed that one question name to the list in the `startSurveySession()` function. We would also need to set `showOptionalQuestion` to false, to disable adding the free-form question to the survey. This is otherwise created by default as `showOptionalQuestion` is always true otherwise.

Further work would need to be done in the UI to to update the navigation button behaviour, i.e currently the button will show next instead of submit, and to ensure the submit answer functionality works to exit survey once the one answer has been submitted.

#### Show Two or More Questions
We could for example want to show only the NPS score question and the feedback question. We would need to pass in the NPS question's name as an input to the `startSurveySession()` function. The feedback question name never needs to be explicitly passed since we create a default (promoter) question and typically update the actual question based on the NPS score.

The UI currently supports this scenario.

#### Add More Question Types
The domain layer would be able to handle any question types, but the UI will need to be updated to display this question by creating a new layout, and viewmodel for it, as well as updating the `SurveyAnswerItemViewModel`'s  `ViewType` enum with this new viewType(s).

A more longterm solution for a dynamic UI would be to create standardized survey questions and answer options, pretty much like state/exploration interactions.

## Essential Checklist

## For UI-specific PRs only
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide]

| LTR  | RTL |   
|---|---|
|https://github.com/oppia/oppia-android/assets/59600948/d2411422-aa99-482d-a0e7-c6fd163e85e1|  https://github.com/oppia/oppia-android/assets/59600948/329e0672-8d4c-4f43-8c6c-b9036a78980a|
|Updated LTR video with restored answers|
|https://github.com/oppia/oppia-android/assets/59600948/1a0decff-7dbf-4a6a-a225-5f38f4191fd6|


app/src/main/java/org/oppia/android/app/survey/SurveyFragmentPresenter.kt
```diff
@@ -0,0 +1,77 @@
...
+  private val oppiaLogger: OppiaLogger
+) {
+  private lateinit var profileId: ProfileId
+  private lateinit var topicId: String
+  private lateinit var binding: SurveyFragmentBinding
+  private lateinit var surveyToolbar: Toolbar
+  private lateinit var surveyProgressText: TextView
+  private lateinit var surveyProgressBar: ProgressBar
+  private lateinit var questionTextView: TextView
+  private lateinit var questionRecycleView: RecyclerView
```
adhiamboperes: Typo, Rename

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+    return surveyFragmentPresenter.handleCreateView(
+      inflater,
+      container,
+      internalProfileId,
+      topicId
+    )
+  }
+
+  fun handleKeyboardAction() = surveyFragmentPresenter.handleKeyboardAction()
+
```
adhiamboperes, line 69: TODO, incomplete code

app/src/main/java/org/oppia/android/app/survey/SurveyFragmentPresenter.kt
```diff
@@ -0,0 +1,77 @@
...
+    )
+    binding.lifecycleOwner = fragment
+    questionRecycleView = binding.root.findViewById(R.id.survey_question_recycler_view)
+    questionTextView = binding.root.findViewById(R.id.survey_question_text)
+    surveyToolbar = binding.surveyToolbar
+    surveyProgressText = binding.surveyProgressText
+    surveyProgressBar = binding.surveyProgressBar
+    activity.setSupportActionBar(surveyToolbar)
+
+    binding.surveyToolbar.setNavigationOnClickListener {
```
adhiamboperes: TODO, add exit confirmation dialog

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+          android:orientation="horizontal">
+
+          <ProgressBar
+            android:id="@+id/survey_progress_bar"
+            style="@android:style/Widget.ProgressBar.Horizontal"
+            android:layout_width="0dp"
+            android:layout_height="@dimen/survey_progress_bar_height"
+            android:layout_weight="1"
+            android:max="100"
+            android:progress="25"
```
adhiamboperes: Placeholder, populate via databinding.

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+      app:layout_constraintEnd_toEndOf="parent"
+      app:layout_constraintStart_toStartOf="parent"
+      app:layout_constraintBaseline_toBaselineOf="@id/survey_app_bar_layout">
+
+      <TextView
+        android:id="@+id/survey_question_text"
+        style="@style/TextViewStart"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="28dp"
```
adhiamboperes: Create style and add dimensions to file.

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+        android:textSize="18sp"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+      <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/survey_question_recycler_view"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="28dp"
+        android:layout_marginTop="52dp"
```
adhiamboperes: Add dimensions to dimens file

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+      <androidx.appcompat.widget.Toolbar
+        android:id="@+id/survey_toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="@color/component_color_shared_secondary_toolbar_color"
+        android:minHeight="?attr/actionBarSize"
+        android:theme="@style/Widget.AppCompat.ActionBar"
+        app:navigationIcon="@drawable/ic_close_white_24dp">
+
+        <LinearLayout
```
adhiamboperes, line 37: Maybe use constraint layout instead?

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+      android:layout_height="wrap_content"
+      app:layout_constraintBottom_toBottomOf="parent"
+      app:layout_constraintEnd_toEndOf="parent"
+      app:layout_constraintStart_toStartOf="parent">
+
+      <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/survey_navigation_buttons_container"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_margin="28dp">
```
adhiamboperes: Add dimensions to dimens file

app/src/main/res/layout/survey_fragment.xml
```diff
@@ -0,0 +1,130 @@
...
+
+        <Button
+          android:id="@+id/survey_previous_button"
+          style="@style/PreviousButtonActive"
```
adhiamboperes: Style causes CI failure, investigate.

app/src/main/res/layout/survey_nps_score_layout.xml
```diff
@@ -0,0 +1,164 @@
...
+  <androidx.constraintlayout.widget.ConstraintLayout
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <TextView
+      android:id="@+id/not_likely_text_view"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:fontFamily="sans-serif-light"
+      android:text="0 - Not at all likely"
```
adhiamboperes: Move to strings file.
Create style.
Should be italicized per design.
Ditto below.

app/src/main/res/layout/survey_nps_score_layout.xml
```diff
@@ -0,0 +1,164 @@
...
+        android:layout_height="wrap_content"
+        android:background="@drawable/survey_nps_radio_unselected_background"
+        android:button="@null"
+        android:fontFamily="sans-serif-medium"
+        android:gravity="center"
+        android:text="0"
+        android:textColor="@color/component_color_shared_active_text_color"
+        android:textSize="18sp" />
+
+      <RadioButton
```
adhiamboperes: Create style

app/src/main/res/layout/survey_nps_score_layout.xml
```diff
@@ -0,0 +1,164 @@
...
+        android:layout_height="wrap_content"
+        android:background="@drawable/survey_nps_radio_unselected_background"
+        android:button="@null"
+        android:fontFamily="sans-serif-medium"
+        android:gravity="center"
+        android:text="9"
+        android:textColor="@color/component_color_shared_active_text_color"
+        android:textSize="18sp" />
+
+      <RadioButton
```
adhiamboperes: Would it be better to generate the radiobuttons dynamically?

app/src/main/AndroidManifest.xml
```diff
@@ -311,6 +311,11 @@
     <activity
       android:name=".app.testing.activity.TestActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
+    <activity
+      android:name=".app.survey.SurveyActivity"
+      android:label="@string/survey_activity_title"
```
adhiamboperes, line 312: This activity doesn't really have a title, label for talkback.

adhiamboperes:
This PR has an existing test file that is failing on bazel but passing on gradle, and I am not sure why, yet. The changes in this PR do not relate to Checkpointing at all. I however do not think this should continue blocking review, and I am going to keep looking into it.

The only failure I expect that will be fixed by @MohitGupta121 in PR #5076 is the static check related to #5072.
|||
|---|---| 
|Screenshot 2023-06-27 at 18 08 49|Screenshot 2023-06-27 at 18 08 28|

adhiamboperes:
@MohitGupta121, could you please take a look at the files in the resource folder including layouts, strings, dimens and colors to verify that the naming conventions have been met, and also that they follow dark mode conventions.

@kkmurerwa, I'd like you to look at the PR as a whole, for android, and @seanlip could you please look at everything else?

app/src/main/java/org/oppia/android/app/survey/SelectedAnswerAvailabilityReceiver.kt
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   *  Called when the input answer availability changes.
+   */
+  fun onPendingAnswerAvailabilityCheck(inputAnswerAvailable: Boolean)
+}
+
+/** A callback that will be called when a user submits an answer. */
+interface SelectedAnswerHandler {
+  /** Return the current selected answer that is ready for submission. */
+  fun getPendingAnswer(selectedAnswer: SurveySelectedAnswer)
```
seanlip: "pending" and "freeform" aren't really in the same "category" of things. Are these two functions meant to be complementary to each other? If so, then maybe adjust the naming to reflect the difference (e.g. if the first function is for "selection" questions, then maybe getMultipleChoiceAnswer or similar)?

If they are indeed different things then is "free form" a subset of "pending"? If so then when would free-form be used (why wouldn't the caller always use "pending")?

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/MarketFitItemsViewModel.kt
```diff
@@ -0,0 +1,127 @@
...
+          )
+          getPendingAnswer()
+        }
+      }
+    isAnswerAvailable.addOnPropertyChangedCallback(callback)
+  }
+
+  private fun updateIsAnswerAvailable() {
+    val wasSelectedItemListEmpty = isAnswerAvailable.get()
+    if (selectedItems.isNotEmpty() != wasSelectedItemListEmpty) {
```
seanlip: If this is a boolean, I would suggest not starting it with a verb -- i.e. selectedItemListWasEmpty

app/src/main/res/color/component_color_shared_survey_option_selector.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+  <item android:color="@color/component_color_survey_radio_button_unchecked_color" android:state_checked="false" android:state_enabled="false" />
+  <item android:color="@color/component_color_survey_radio_button_unchecked_color" android:state_checked="false" android:state_enabled="true" />
+  <item android:color="@color/component_color_survey_radio_button_checked_color" android:state_checked="true" />
+</selector>
```
seanlip: Missing newline at end?

app/src/main/res/drawable/survey_next_button_background.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+  <!-- Enabled state -->
+  <item android:drawable="@drawable/rounded_primary_button_grey_shadow_color" android:state_enabled="true" />
+  <!-- Disabled state -->
+  <item android:drawable="@drawable/survey_rounded_disabled_button_background" android:state_enabled="false" />
+</selector>
```
seanlip: Ditto here and in a few other files.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterAlphaModule.kt
```diff
@@ -263,4 +269,28 @@ class PlatformParameterAlphaModule {
...
+
+  @Provides
+  @NpsSurveyMinimumAggregateLearningTimeInATopicInMinutes
+  fun provideNpsSurveyMinimumAggregateLearningTimeInATopicInMinutes(
+    platformParameterSingleton: PlatformParameterSingleton
+  ): PlatformParameterValue<Int> {
+    return platformParameterSingleton.getIntegerPlatformParameter(
+      NPS_SURVEY_MINIMUM_AGGREGATE_LEARNING_TIME_IN_A_TOPIC_IN_MINUTES
+    ) ?: PlatformParameterValue.createDefaultParameter(
+      NPS_SURVEY_MINIMUM_AGGREGATE_LEARNING_TIME_IN_A_TOPIC_IN_MINUTES_DEFAULT_VAL
```
seanlip: (Optional, feel free to disregard) If we are naming other things in a way that ends in DEFAULT_VALUE, we might want to do the same thing here. Perhaps "aggregate" can be replaced with "total", or "minutes" with "mins", if line length is an issue?

domain/src/main/java/org/oppia/android/domain/survey/SurveyConstantsProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+/** Qualifier corresponding to the total number of questions in a survey. */
+@Qualifier
+annotation class TotalQuestionCount
+
+/** Provider to return the constants required during the survey. */
+@Module
+class SurveyQuestionModule {
+  @Provides
+  @TotalQuestionCount
+  fun provideTotalQuestionCount(): Int = 4
```
seanlip: Just wondering -- isn't the list of questions stored somewhere else, and should this be computed from that list rather than be declared directly? It seems like a separate source of truth and I'm not sure what would happen if someone edited the list of questions but not this constant, or vice versa.

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgress.kt
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * Calling code should prevent this method from failing by checking state ahead of calling this method and providing
+   * more useful errors to UI calling code since errors thrown by this method will be more obscure. This method aims to
+   * ensure the internal state of the controller remains correct. This method is not meant to be covered in unit tests
+   * since none of the failures here should ever be exposed to controller callers.
+   */
+  fun advancePlayStageTo(nextStage: SurveyStage) {
+    when (nextStage) {
+      SurveyStage.NOT_IN_SURVEY_SESSION -> {
+        // All transitions to NOT_IN_SURVEY_SESSION are valid except itself.
```
seanlip: ...except those originating from itself.

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgress.kt
```diff
@@ -0,0 +1,109 @@
...
+        check(surveyStage != SurveyStage.NOT_IN_SURVEY_SESSION) {
+          "Cannot transition to NOT_IN_TRAINING_SESSION from NOT_IN_TRAINING_SESSION"
+        }
+        surveyStage = nextStage
+      }
+      SurveyStage.LOADING_SURVEY_SESSION -> {
+        // A session can only begun being loaded when not previously in a session.
```
seanlip: begun --> start

model/src/main/proto/survey.proto
```diff
@@ -122,17 +125,17 @@ enum UserTypeAnswer {
   // User type unknown.
   USER_TYPE_UNSPECIFIED = 0;
 
-  // Corresponds to an unspecified or unknown type of user.
-  OTHER = 1;
+  // Corresponds to a user who is a learner.
+  LEARNER = 1;
 
-  // Corresponds to a user who is a teacher.
+  // Corresponds to a user who is a parent.
```
seanlip: This comment and the next one are incorrectly switched.

seanlip (changes requested):
Thanks @adhiamboperes! I left a few inline comments, PTAL. Sorry that you had to edit so many files because of the Dagger thing :(

I also took a look at the videos. They look great! Some comments: 
- The multiple-choice option texts are a bit out-of-alignment vertically from the radio buttons. Could that be fixed?
- Are the user's previous answers saved, and if so, could they be shown on back-navigation (the user can, of course, still change them)? It seems odd to me that, when going back to a previous survey question, the original answer has disappeared. It feels like the user's answer wasn't saved at all.
- Could you show what happens when the user navigates back two questions and then forward one question?
- Could you show verification for when the survey completion events were registered and what data they contain (or is that going to be a different PR)?

Thanks!

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
```diff
@@ -500,4 +507,38 @@ class ExplorationActivityPresenter @Inject constructor(
       }
     }
   }
+
+  private fun maybeShowSurveyDialog(profileId: ProfileId, topicId: String) {
+    surveyGatingController.maybeShowSurvey(profileId, topicId).toLiveData()
```
kkmurerwa, line 513: I feel like the `surveyController.maybeShowSurvey()` method could be better named as `getSurveyGatingResult()` because that it's major purpose.

app/src/main/java/org/oppia/android/app/survey/ExitSurveyConfirmationDialogFragment.kt
```diff
@@ -0,0 +1,75 @@
...
+  lateinit var exitSurveyConfirmationDialogFragmentPresenter:
+    ExitSurveyConfirmationDialogFragmentPresenter
+
+  companion object {
+    internal const val PROFILE_ID_KEY = "ExitSurveyConfirmationDialogFragment.profile_id"
+
+    /**
+     * Creates a new instance of a DialogFragment to display an exit confirmation in a survey.
+     *
+     * @param profileId the ID of the profile viewing the survey
```
kkmurerwa, line 28: [Optional] The @param and @return below could end with a fullstop as on the [official Kdoc documentation](https://kotlinlang.org/docs/kotlin-doc.html#block-tags). I've looked at the rest of the project and a lot @param and @return annotations don't end with full stops (there is an inconsistency there) which is why I marked this as optional. Several other files have this issue. Check them out as well.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
```diff
@@ -0,0 +1,76 @@
...
+/** Fragment that represents the current state of a survey. */
+class SurveyFragment :
+  InjectableFragment(),
+  SelectedAnswerAvailabilityReceiver,
+  SelectedAnswerHandler {
+
+  companion object {
+    /**
+     * Creates a new instance of a SurveyFragment.
```
kkmurerwa, line 22: Add a new line at the end of the first comment line.
```suggestion
     * Creates a new instance of a SurveyFragment.
     *
```

app/src/main/res/drawable/rounded_primary_button_grey_shadow_color.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+ <item>
+    <shape android:shape="rectangle">
+      <solid android:color="#25000000" />
```
kkmurerwa, line 5: This could be made into a color resource.

kkmurerwa, line 5: This could be made into a color resource. Check the other resource files as well.

app/src/main/res/layout/survey_free_form_layout.xml
```diff
@@ -0,0 +1,42 @@
...
+
+  <androidx.constraintlayout.widget.ConstraintLayout
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <EditText
+      android:id="@+id/free_form_text"
+      style="@style/SurveyFreeFormAnswerEditText"
+      android:hint="@string/survey_free_form_text_hint"
+      android:inputType="text|textMultiLine"
```
kkmurerwa: Add `textCapSentences` to enable sentence case on the keyboard and input. 
```suggestion
      android:inputType="text|textMultiLine|textCapSentences"
```

model/src/main/proto/arguments.proto
```diff
@@ -303,3 +303,11 @@ message AppLanguageActivityStateBundle {
   // The default app language selected by the user.
   OppiaLanguage oppia_language = 1;
 }
+
+message SurveyActivityParams {
+  // The ID of the profile for which the survey is too be shown.
```
kkmurerwa:
```suggestion
  // The ID of the profile for which the survey is to be shown.
```

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -0,0 +1,429 @@
...
+/** Tests for [SurveyFragment]. */
+// FunctionName: test names are conventionally named with underscores.
+@Suppress("FunctionName")
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(
+  application = SurveyFragmentTest.TestApplication::class,
+  qualifiers = "port-xxhdpi"
+)
+class SurveyFragmentTest {
```
kkmurerwa, line 128: There seem to be missing tests for the Previous button scenario(s).

kkmurerwa (commented):
Hi Adhiambo. I have taken a look at the code and left a few comments. PTAL.

adhiamboperes:
> Thanks @adhiamboperes! I left a few inline comments, PTAL. Sorry that you had to edit so many files because of the Dagger thing :(
Thanks, I think this is one of the problems I want to solve in the coming weeks.
> 
> I also took a look at the videos. They look great! Some comments:
> 
> * The multiple-choice option texts are a bit out-of-alignment vertically from the radio buttons. Could that be fixed?

Thanks for catching this, I hadn't realized. Yes, I can definitely fix this.
> * Are the user's previous answers saved, and if so, could they be shown on back-navigation (the user can, of course, still change them)? It seems odd to me that, when going back to a previous survey question, the original answer has disappeared. It feels like the user's answer wasn't saved at all.

Yes, we do save them--I had put this logic in a different branch that handles saving and logging, but I can specifically cherry-pick it here if it makes better sense.

> * Could you show what happens when the user navigates back two questions and then forward one question?

I'll be adding a video and a UI test for that.

> * Could you show verification for when the survey completion events were registered and what data they contain (or is that going to be a different PR)?

Yes, this is in the next PR in the chain.

Thank you for taking a look!

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
adhiamboperes: Noted, I'm adding these.

app/src/main/java/org/oppia/android/app/survey/ExitSurveyConfirmationDialogFragment.kt
adhiamboperes: Per https://github.com/oppia/oppia-android/pull/4916/files/3dd5e8dac6277379ac5dadd924f552f8771d5953..dd0f8a811e33f0fcef6eee32c1498a3af1704807#r1213870876, it seems that the convention in oppia-android is to only add punctuation for an @-clause sentence if it is multiline or wraps.

domain/src/main/java/org/oppia/android/domain/survey/SurveyConstantsProvider.kt
adhiamboperes: We generate the list at the beginning of the survey, and do not store it. It only exists ephemerally during the survey.
We have an enum class of all the question names needed in the survey, and from this, I generate the actual question objects. The challenge I had is that this enum includes 6 questions, where the last 3 are conditional based on the answer selected in the 3rd question(nps score). So I opted to take the first four questions in the list, which is what this value  is used for, and then switch out the last question dynamically depending on the nps score selected.

If someone were to add a new question to the list and not edit the constant, then that could actually create a problem based on the position of the question, since the list is indexed.

I think I will need to fix this, but I am wondering whether what would be the best way to dictate the number of questions that go in the survey - a platform parameter or maybe just a variable in the controller. More drastically, the question list generation logic could change, but let me explore the former option first.

adhiamboperes: Following our conversation to avoid  using two sources of data that are dependent but represented independently, I came up with a solution that is more in line with your comment here:

It is now the responsibility of the caller of the  `startSurveySession()` to provide the list of questions at call site, and we compute the number of questions from this. I have overall modified how the survey is generated to support this, but the rest of the architecture remains intact.

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgress.kt
adhiamboperes: Fixed.

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgress.kt
adhiamboperes: Fixed

model/src/main/proto/survey.proto
adhiamboperes: Corrected

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterAlphaModule.kt
adhiamboperes: I updated this constant name.

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
adhiamboperes: The return here is boolean of whether the survey should be shown or not.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
adhiamboperes: Done.

app/src/main/res/drawable/rounded_primary_button_grey_shadow_color.xml
adhiamboperes: Did you mean the color value for `solid android:color="#25000000"`? I thought so as well, but when I tried it I ran into issue #5076

adhiamboperes: I am unable to do this https://github.com/oppia/oppia-android/pull/4945#discussion_r1247753990

app/src/main/res/layout/survey_free_form_layout.xml
adhiamboperes: Done.

model/src/main/proto/arguments.proto
adhiamboperes: Updated, thanks.

app/src/main/res/drawable/survey_next_button_background.xml
adhiamboperes: Fixed

app/src/main/res/color/component_color_shared_survey_option_selector.xml
adhiamboperes: I've added newlines to the end of all resource files.

app/src/main/java/org/oppia/android/app/survey/SelectedAnswerAvailabilityReceiver.kt
adhiamboperes: I renamed pending to getMultipleChoiceAnswer() because that is the distinction that I had hoped to make. One of the listeners is called when a multiplechoice question needs to be submitted, and the other for free form answers.

MohitGupta121 (approved):
@adhiamboperes all resources files look good to me. Only one concern Is you not added dark-mode colors for this? I think for that you need designs?

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/MarketFitItemsViewModel.kt
adhiamboperes: Thanks, I have updated this.

app/src/main/java/org/oppia/android/app/survey/ExitSurveyConfirmationDialogFragment.kt
kkmurerwa: We can leave it as is based on the linked discussion then.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
kkmurerwa: This looks good now.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
kkmurerwa: Is this done?

model/src/main/proto/arguments.proto
kkmurerwa: This looks good now.

app/src/main/java/org/oppia/android/app/application/BUILD.bazel
```diff
@@ -120,7 +120,6 @@ android_library(
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/exceptions:logger_module",
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler:metric_log_scheduler_module",
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/loguploader:worker_module",
-        "//domain/src/main/java/org/oppia/android/domain/survey:question_constants_module",
```
kkmurerwa: hhggas

app/src/main/res/drawable/rounded_primary_button_grey_shadow_color.xml
kkmurerwa: Does that mean this will have to suffice as an explicit color code?

kkmurerwa: If that is the case then you can ignore this and comment [#4945](https://github.com/oppia/oppia-android/pull/4945#discussion_r1247753990).

app/src/main/res/layout/survey_free_form_layout.xml
kkmurerwa: Confirmed.

kkmurerwa:
Hi @adhiamboperes. I have taken a look at the comments. With the exception of the test for the [SurveyFragmentTest.kt](https://github.com/oppia/oppia-android/pull/4945/files/6b6b4dbd31b4db4429fb65412924e33fae071717#diff-567442aed5d69f4d6c73d32813c3fbe95dd1088f503c324a8fd2eaaf5f17ec40) and the color issues, everything else looks good so far.

app/src/main/res/drawable/rounded_primary_button_grey_shadow_color.xml
adhiamboperes: Yes, for now

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
adhiamboperes: I have added these, PTAL.

adhiamboperes:
> > Thanks @adhiamboperes! I left a few inline comments, PTAL. Sorry that you had to edit so many files because of the Dagger thing :(
> > Thanks, I think this is one of the problems I want to solve in the coming weeks.
> > I also took a look at the videos. They look great! Some comments:
> > 
> > * The multiple-choice option texts are a bit out-of-alignment vertically from the radio buttons. Could that be fixed?
> 
> Thanks for catching this, I hadn't realized. Yes, I can definitely fix this.
> 
> > * Are the user's previous answers saved, and if so, could they be shown on back-navigation (the user can, of course, still change them)? It seems odd to me that, when going back to a previous survey question, the original answer has disappeared. It feels like the user's answer wasn't saved at all.
> 
> Yes, we do save them--I had put this logic in a different branch that handles saving and logging, but I can specifically cherry-pick it here if it makes better sense.
> 
> > * Could you show what happens when the user navigates back two questions and then forward one question?
> 
> I'll be adding a video and a UI test for that.
> 
> > * Could you show verification for when the survey completion events were registered and what data they contain (or is that going to be a different PR)?
> 
> Yes, this is in the next PR in the chain.
> 
> Thank you for taking a look!

Hi @seanlip,

I have addressed all your comments on the PR, PTAL!

I have also updated the PR description and the videos.

adhiamboperes:
> This PR has an existing test file that is failing on bazel but passing on gradle, and I am not sure why, yet. The changes in this PR do not relate to Checkpointing at all. I however do not think this should continue blocking review, and I am going to keep looking into it.
> 
> The only failure I expect that will be fixed by @MohitGupta121 in PR #5076 is the static check related to #5072.
> 
> Screenshot 2023-06-27 at 18 08 49	Screenshot 2023-06-27 at 18 08 28

This turned out to be a flaky test scenario, they are mostly passing.

app/src/main/java/org/oppia/android/app/application/BUILD.bazel
seanlip: I'm not sure I understand either the text of this comment, or why it got resolved (without any explanation...)

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/MarketFitItemsViewModel.kt
```diff
@@ -44,29 +51,42 @@ class MarketFitItemsViewModel @Inject constructor(
...
+    answerHandler.getMultipleChoiceAnswer(answer)
+  }
+
+  override fun getPreviousAnswer(): SurveySelectedAnswer {
+    return SurveySelectedAnswer.getDefaultInstance()
+  }
+
+  override fun restorePreviousAnswer(previousAnswer: SurveySelectedAnswer) {
+    val selectedAnswerOption = previousAnswer.marketFit.number
+    val optionIndex = if (selectedAnswerOption != 0) selectedAnswerOption - 1 else 0
```
seanlip: Does 0 indicate the case when no option was selected? Would null be a better sentinel value? (0 can easily be confused with an index.)

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/UserTypeItemsViewModel.kt
```diff
@@ -18,20 +19,26 @@ class UserTypeItemsViewModel @Inject constructor(
   private val resourceHandler: AppLanguageResourceHandler,
   private val selectedAnswerAvailabilityReceiver: SelectedAnswerAvailabilityReceiver,
   private val answerHandler: SelectedAnswerHandler
-) : SurveyAnswerItemViewModel(ViewType.USER_TYPE_OPTIONS) {
+) : SurveyAnswerItemViewModel(ViewType.USER_TYPE_OPTIONS), PreviousAnswerHandler {
```
seanlip, line 22: What does `, PreviousAnswerHandler` do here? I can't see where it is used.

seanlip (changes requested):
Thanks @adhiamboperes! Just a few more inline comments. Additionally, for the new video:

- At 0:07 and 0:10 it seems that an incomplete question displays and then later the options appear. This is quite jarring. Can all parts of the question be loaded at once?
- Around 0:30 there is a brief flash of the textarea and then it jumps back to the NPS question. This seems like a bug?

app/src/main/java/org/oppia/android/app/application/BUILD.bazel
adhiamboperes: This was an erroneous comment from @kkmurerwa. Maybe I should delete it instead.

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/UserTypeItemsViewModel.kt
adhiamboperes: It is an interface implemented by this class with the function   `override fun restorePreviousAnswer(previousAnswer: SurveySelectedAnswer)` for restoring the previous saved answer for a question on back/forward navigation.

app/src/main/java/org/oppia/android/app/survey/surveyitemviewmodel/MarketFitItemsViewModel.kt
adhiamboperes: Index 0 corresponds to `ANSWER_UNSPECIFIED` which is not a valid option so it needs to be filtered out. Valid enum type numbers start from 1 while the [option] list item indices start from 0, hence the minus(1) to get the correct index to update. Notice that for [getPendingAnswer] we increment the index first to get the correct typeCase to save. I have updated the code to be a bit more idiomatic.

adhiamboperes:
I have looked into these:
> * At 0:07 and 0:10 it seems that an incomplete question displays and then later the options appear. This is quite jarring. Can all parts of the question be loaded at once?

This is only visible because I had recorded the video at an extremely low bitrate(4Mbps). In reality, I am not able to observe serious visual jank.

> * Around 0:30 there is a brief flash of the textarea and then it jumps back to the NPS question. This seems like a bug?

I rechecked and this happened between actual next and prev button clicks in quick succession. I cannot repro this.

1. I have re-recorded the video with standard video settings. 
2. I have rechecked the code and could not find logic breaks to cause the text area question to load unless navigated to explicitly.

seanlip (approved):
Thanks @adhiamboperes! LGTM.

Just one thing I noticed when watching the demo again -- the fourth bullet point in the second question should be "N/A - I don't use Oppia anymore". I.e. there is a missing "I".

adhiamboperes:
> Thanks @adhiamboperes! LGTM.
> 
> Just one thing I noticed when watching the demo again -- the fourth bullet point in the second question should be "N/A - I don't use Oppia anymore". I.e. there is a missing "I".

I am curious about this: from the mocks there is no I, but other PMF samples I have seen seem to have it.

seanlip:
> I am curious about this: from the mocks there is no I, but other PMF samples I have seen seem to have it.

Yup, I think that was just a mistake, let's please add it.

(Also in the future, if you see anything that looks "off" please do bring it up directly with the design/product team. Thanks!)

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
kkmurerwa: Confirmed. These are good now.

kkmurerwa:
Hey @adhiamboperes. I have reviewed the remaining test issue and that is now resolved. Everything is good now.

adhiamboperes:
Thanks @seanlip, @kkmurerwa @MohitGupta121!

PR #4965, opened 2023-05-05
Fix part of #632: Simplify MarginBindingAdapters and fix some RTL cases. Fix #2116: UI misalignment due to not resetting the layout.

### LINKED ISSUE #2116: Fix UI for profile chooser screen in add-view
`profile_chooser_add_view.xml` needs to be updated so that the UI looks similar to mocks.

## Mocks
https://xd.adobe.com/view/ee9e607b-dbd6-4372-48dc-b687d32af3da-98af/grid/
https://xd.adobe.com/view/e8aa4198-3940-47f9-514a-f41cc54457f6-9e9b/grid
https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/33741ea7-89da-4e5d-b2a4-9872f6db48b4/
https://xd.adobe.com/view/d405de00-a871-4f0f-73a0-f8acef30349b-a234/screen/6fb19f15-699a-40aa-8392-cdfc6d7bf8c5/

https://github.com/oppia/oppia-android/wiki/Working-on-UI

## Current Screenshot
Screenshot_1605508088
Screenshot_1605508113
Screenshot_1605508138
Screenshot_1605508141

## Device:
Nexus-9 Emulator
API 29

masclot:
## Explanation

Fix #2116:
Fix part of #632:

Simplify MarginBindingAdapters.

The code is mostly the same as in PR #4874, which was reverted. This PR fixes the bug that caused the rollback.
The rest of the reverted changes in PR #4874 will be included in a separate PR.

The fix consists of resetting the layout params after changing the top and bottom margins, by calling View.setLayoutParams(), as explained in the [layout params documentation](https://developer.android.com/reference/android/view/ViewGroup.MarginLayoutParams#bottomMargin). The same call is not needed when calling a method instead of setting a property.

It also fixes an RTL issue: if the view is not attached yet, there is no layout direction information available, making the old implementation render in LTR always.

Some screenshots:

|new|current|
|-|-|
|home_rtl | home_rtl_current|
|profile_chooser_landscape|profile_chooser_landscape_current|
|profile_chooser_portrait|profile_chooser_portrait_current|
|recently_played_rtl|recently_played_rtl_current|

Test results:
Test results

## Essential Checklist

## For UI-specific PRs only


app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
```diff
@@ -29,36 +24,18 @@ public static void setLayoutMarginStart(@NonNull View view, float marginStart) {
...
     if (view.getLayoutParams() instanceof MarginLayoutParams) {
       MarginLayoutParams params = (MarginLayoutParams) view.getLayoutParams();
-      params.setMargins(
-          params.getMarginStart(),
-          (int) marginTop,
-          params.getMarginEnd(),
-          params.bottomMargin
-      );
+      params.topMargin = (int) marginTop;
+      view.setLayoutParams(params);
```
adhiamboperes, line 38: I notice that we are setting the layout parameters of the view only with MarginTop and Bottom below. Why is this not the case for marginStart/End, and `layoutMargin` as well?

adhiamboperes (changes requested):
Thanks @masclot! 
It seems that the main difference between this code and the one in the [previous PR](https://github.com/oppia/oppia-android/pull/4874) is the fact that we are calling `view.setLayoutParams(params)` before requesting the layout. It may be worth explaining in the PR description how or why this resolves the UI distortion issue.

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
masclot: I updated the PR description with an explanation

masclot:
@adhiamboperes PTAL
I updated the PR description with an explanation.

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
adhiamboperes: > The fix consists of resetting the layout params after changing one of its properties, as explained in the [layout params documentation](https://developer.android.com/reference/android/view/ViewGroup.MarginLayoutParams#bottomMargin). The same is not needed when calling a method instead of setting a property.

Maybe to clarify, my question is asking: why do we need to reset or change the value of marginTop and marginBottom, but not marginStart and marginEnd.

masclot: The documentation doesn't say, so I can only guess: marginTop/Bottom are java fields, while marginStart/End are java method calls. In java, when a field is changed externally, the object doesn't know, thus it doesn't know it needs to recompute the view layout. On the other hand, method calls are more flexible. If you look at Android's source code, you can see that calling setMarginStart does more than just setting a field: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;l=8619?q=Marginlayoutparams%20setmarginstart&sq=&ss=android

masclot:
@adhiamboperes PTAL

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
adhiamboperes: You're correct, according to the source code, setMarginStart/End returns the computed layout margins, therefore no need to call ViewGroup#setLayoutParams(LayoutParams) afterwards.

I think part of my confusion had to do with the fact that for start/end, we are relying oon MarginLayoutParamsCompat, but for top/bottom, we are using MarginLayoutParams. I have noticed that MarginLayoutParamsCompat does not provide means for setting the values of the top/bottom margins.

adhiamboperes: As for my other comment:

> It seems that the main difference between this code and the one in the https://github.com/oppia/oppia-android/pull/4874 is the fact that we are calling view.setLayoutParams(params) before requesting the layout. It may be worth explaining in the PR description how or why this resolves the UI distortion issue.

I think it is now clear that the distortion in was caused by the fact that in #4874, we were not calling `ViewGroup#setLayoutParams()` after updating the margin values for top and bottom. You could just update the description to reflect this.

Please assign Ben once this is done. Thanks!

adhiamboperes (commented):
Since this fix will remove the blocker to closing #2116, you can add  fix 2116 to your title and explanation.

adhiamboperes (dismissed):
These updates LGTM.

app/src/main/java/org/oppia/android/app/databinding/MarginBindingAdapters.java
```diff
@@ -29,36 +24,18 @@ public static void setLayoutMarginStart(@NonNull View view, float marginStart) {
   public static void setLayoutMarginEnd(@NonNull View view, float marginEnd) {
     if (view.getLayoutParams() instanceof MarginLayoutParams) {
       MarginLayoutParams params = (MarginLayoutParams) view.getLayoutParams();
-      float marginStart = params.getMarginStart();
-      if (isRtlLayout(view)) {
-        setLayoutDirectionalMargins(view, (int) marginEnd, (int) marginStart);
-      } else {
-        setLayoutDirectionalMargins(view, (int) marginStart, (int) marginEnd);
-      }
+      MarginLayoutParamsCompat.setMarginEnd(params, (int) marginEnd);
```
BenHenning, line 27: FWIW we might not actually need the compat version since it's a min SDK of 17 to directly set end, but I don't mind keeping it.

I actually didn't realize that MarginLayoutParams.setEnd() correctly adjusted the margin based on directionality (its documentation, unfortunately, doesn't indicate this).

BenHenning (dismissed):
Thanks @masclot! Sorry for the delayed review.

I think the solution looks excellent. Regarding the the specific part that caused the regression which led to a revert, what was the actual problem & how was it addressed? I didn't get that part from the PR description, but perhaps I misunderstood something.

Separately, is there any way to test both the issue being fixed (#2116) and the specific regression issue that led to the reversion?

BenHenning:
FYI @adhiamboperes I cleared your review since I have one leftover comment that needs to be addressed before this PR can get merged.

masclot:
> I think the solution looks excellent. Regarding the the specific part that caused the regression which led to a revert, what was the actual problem & how was it addressed? I didn't get that part from the PR description, but perhaps I misunderstood something.

As a sum-up:
Problem: 
Setting a top/bottom margin does not force a relayout of the view by default. However, setting the start/end margins _does_ force a relayout. Those views that were setting both top/bottom and start/end margins had no problem, as a relayout was requested. However, those views that were setting the top/bottom margin only, like the profile view, nothing was forcing a relayout and the top/bottom margins were not really applied.

Solution:
Forcing a relayout on top/bottom margin changes as well

> Separately, is there any way to test both the issue being fixed (#2116) and the specific regression issue that led to the reversion?

Do you mean, create a unit test for the profile view that checks that the top/bottom margins are applied?

masclot:
As discussed, I added tests for top/bottom margin changes
@adhiamboperes PTAL

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
```diff
@@ -134,6 +138,22 @@ class MarginBindingAdaptersTest {
...
   }
 
+  @Test
+  fun testMarginBindableAdapters_topAndBottomIsCorrect() {
+    activityRule.scenario.runWithActivity {
+      val textView: TextView = it.findViewById(R.id.test_margin_text_view)
+      setLayoutMarginBottom(textView, /* marginBottom= */ 24f)
+      setLayoutMarginTop(textView, /* marginTop= */ 40f)
+
+      ShadowLooper.idleMainLooper()
```
adhiamboperes: We typically don't use this directly to bring the test to an idle state. Instead, we have some convenience methods like runCurrent(), which you can use here, advanceTimeBy--used when you need to execute a task a specified amount of time in the future, and advanceUntilIdle. Please see the [interface](https://github.com/oppia/oppia-android/blob/develop/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchers.kt) and the [impl](https://github.com/oppia/oppia-android/blob/develop/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatcher.kt).

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
```diff
@@ -134,6 +138,22 @@ class MarginBindingAdaptersTest {
...
+
+      val computedTopMargin = textView.y
+      val computedBottomMargin = (textView.parent as View).height - textView.y - textView.height
```
adhiamboperes: Not sure if the computation here is correct. How about:

```
// Get the layout parameters of the TextView
val layoutParams = textView.layoutParams as ViewGroup.MarginLayoutParams

// Retrieve the top and bottom margins
val topMargin = layoutParams.topMargin
val bottomMargin = layoutParams.bottomMargin
```

adhiamboperes (changes requested):
Hey @masclot, I think we can expand the tests here a bit, since the distortion issue was observed in landscape in a tablet, we should also test for landscape, and landscape in large screen devices. (land, sw600dp-land).

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
masclot: The properties top/bottom margin in the layout were already correct before the fix. Using them does not prove that the fix works. The problem is that the new property values were not applied to the UI, i.e. the view was not layed out respecting the new top/bottom margins. That's way I'm computing the position of the View within its parent, to check that with the fix the View is layed out correctly.

app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
masclot: Done

masclot:
>

masclot:
> Hey @masclot, I think we can expand the tests here a bit, since the distortion issue was observed in landscape in a tablet, we should also test for landscape, and landscape in large screen devices. (land, sw600dp-land).

Done

masclot:
@adhiamboperes PTAL

adhiamboperes (approved):
Thanks @masclot, this LGTM.

PR #4891, opened 2023-03-07
Fix resume lesson fragment buttons scaling issue

KevinGitonga:
Fixes #4710, #4362 and also #3833. The second 2 consecutive issues were closed as they might be possibly duplicates of #4710
- Updating width on both the buttons width 0dp actually gives them equal width enabling the button text and child drawables to scale as expected on max display and font config. Main issue is that the currently used width "144dp" doesn't work when the device's display and font config is set to max.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

| Before Fix            |  After Fix |
:-------------------------:|:-------------------------:
Nexus 5X Api 29
 |  

Google Pixel 3XL
 |

PR #4987, opened 2023-05-17
"Fix #4710 the button scaling problem in largest text, and change button name"

### LINKED ISSUE #4710: Improve support for scaling in resume lesson screen

XichengSpencer:
Fix #4710 the button scaling problem in largest text setting in the resume fragment, and I will test on top of the comments.

The use of material button has better control for the icon inside the button so that it can group icon and text together as responsive design guidelines suggested. Particularly the icon gravity part that attached the icon to the text.

## Essential Checklist

## For UI-specific PRs only

Here are the testing on my own cell phone, even the second smallest text is too large for the screen so the buttons are wrapped in two rows. Only the smallest text can fit.
|Screenshot 1|Screenshot 2|Screeshot 3|
|--|--|--|
|Weixin Image_20230606114148|Weixin Image_20230606114203|Weixin Image_20230606114209|

- Add a screenshot demonstrating UI of tablet
|Screenshot 1| Screenshot 2|
|--|--|
|Screenshot 2023-06-05 120237|Screenshot 2023-06-05 120223|

LTR
|Screenshot 1|Screenshot 2|Screeshot 3|
|--|--|--|
|Image_20230606114738|Image_20230606114706|Image_20230606114719|


adhiamboperes:
Thanks for your PR @XichengSpencer.
This Pr shows only one video with the fix working if the texts are changed. But can you add one of it with the original text?

app/src/main/res/values/styles.xml
```diff
@@ -411,11 +411,8 @@
   </style>
   <!-- Start Over Lesson Button -->
   <style name="SecondaryButton" parent="TextAppearance.AppCompat.Widget.Button">
-    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_width">0dp</item>
```
adhiamboperes, line 414: If someone were to change the parent layout of these buttons to something other than ConstraintLayout, would they still render correctly?

app/src/main/res/values/styles.xml
```diff
@@ -429,12 +426,9 @@
   </style>
   <!-- Continue Lesson Button -->
   <style name="ContinueLessonButton" parent="TextAppearance.AppCompat.Widget.Button">
-    <item name="android:layout_width">144dp</item>
+    <item name="android:layout_width">0dp</item>
```
adhiamboperes: Would it be preferrable, in this case(refer to above comment on constraint layout), to have these attributes set at view level rather than at style level?

app/src/main/res/values/styles.xml
```diff
@@ -429,12 +426,9 @@
...
     <item name="android:layout_height">wrap_content</item>
     <item name="android:drawablePadding">4dp</item>
-    <item name="android:paddingEnd">12dp</item>
-    <item name="android:paddingStart">12dp</item>
-    <item name="android:minWidth">144dp</item>
```
adhiamboperes, line 437: I think the minWidth attribute here and above are important to keep as they prevent the buttons from being smaller than the specified size, unless you can provide a justification for removing them.

adhiamboperes (changes requested):
I have left 3 comments which are important.

General feedback:
- You do not need to close this PR if you make changes, instead, do push commits.
- Please check on your PR title and description formatting--adding the issue number in the description makes it easy to navigate to the linked issue. [See for example](https://github.com/oppia/oppia-android/pull/4891#issue-1613555564).

app/src/main/res/values/styles.xml
XichengSpencer: Unfortunately, I don't think it will work if we change to another type of layout. Since those buttons are pretty much defined by the constraints we set to dynamically fit different screen sizes. The constraint has higher precedence than those fixed numbers so those numbers basically been ignored.

app/src/main/res/values/styles.xml
XichengSpencer: I can certainly try to do that, but I would like to know what is the reason why the former code author put those attributes here instead of the view level. And to clarify, if it is just layout width or height or all the items in the style? I also see there are other style settings for edittext etc. are those also gonna be a problem?

adhiamboperes: The attributes were placed in syles for reusability. However, since the style naming is pretty specific, it seems this style is specific to this button, and in this case, it is okay to move the width and height attributes to the view level. This will make it easier to debug should the parent layout change. It is however useful to have a minWidth attribute in the syles to meet ccessibility requirements.

app/src/main/res/values/styles.xml
adhiamboperes: I do think having these as 0dp or match_constraint would effectively resolve this issue since the buttons will always scale to fit the screensize. However defaulting it at such a high level in styles could lead to the button not displaying at all with other uses of the button (both now and in the future), hence the suggestion to move the attributes to the view level.

adhiamboperes:
Another possible solution would be to wrap the buttons instead, so that the layout changes when the text size is large. Please refer to responsive button layouts in material [here](https://m3.material.io/components/buttons/guidelines#a9d9f5dc-4b15-41e8-b9f1-523690d7fcab). Material d[iscourages wrapping text in a button](url) as we are doing currently anyway.

Feel free to open a thread in the group chat if you'd like to discuss all 3 of these alternatives further.

app/src/main/res/values/styles.xml
XichengSpencer: I test and find out add minWidth doesn't impact my testing on both my emulator or my phone, I will leave it there. However, after testing, it doesn't see to me that it keeps 144dp minimum small device like wearable watch. 
Screenshot_20230524_190754
This is a screenshot of a wearable watch that has 360 width. But it seems in the image two of them combined only takes a bit more than half of the screen. Which is far less than 288/360 visually should be. I personally think it may be restricted by parent or maybe padding. So I am not sure if it is useful now.

app/src/main/res/values/styles.xml
XichengSpencer: Thank you so much, the explanation is really helpful. I will move attributes to the corresponding place in the fragment layout.

XichengSpencer:
> Thanks for your PR @XichengSpencer. This Pr shows only one video with the fix working if the texts are changed. But can you add one of it with the original text?

I did some tests on both my phone and the emulator and put screenshots in the description. For some reason, my phone behaves differently. I can definitely look into that but I still believe having shorter but not misleading words will be less likely to cause abnormal displays.

app/src/main/res/values/styles.xml
adhiamboperes: You don't need to worry about the wearables, since we are not designing for them. You can test wit different size mobile phones.

adhiamboperes:
@seanlip, could you PTAL at these screenshots for product? We decided to have the buttons wrap instead of being truncated.

seanlip:
Looks fine, that's definitely better than the status quo. Thanks for looping me in and thanks for fixing this @XichengSpencer!

app/src/main/res/drawable/secondary_button_background.xml
```diff
@@ -2,7 +2,7 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
   android:shape="rectangle">
   <corners android:radius="@dimen/state_previous_next_button_radius" />
-  <solid android:color="@color/component_color_shared_white_background_color" />
+  <solid android:tint="@color/component_color_shared_white_background_color" />
```
adhiamboperes, line 5: Did you verify that this drawable is not used elsewhere? And in case it is, have you verified that the display is still correct?
If you are unsure about how to check, or how affected views should look like, please let me know.

app/src/main/res/layout-land/resume_lesson_fragment.xml
```diff
@@ -54,36 +55,58 @@
...
-          app:layout_constraintHorizontal_chainStyle="spread"
-          app:layout_constraintStart_toStartOf="parent" />
+          app:backgroundTint="@null"
+          app:icon="@drawable/ic_start_over_24dp"
+          app:iconTint="@color/component_color_shared_secondary_button_background_trim_color"
+          app:iconGravity="textStart" />
+
 
-        <Button
+        <com.google.android.material.button.MaterialButton
```
adhiamboperes, line 93: Could you explain why this change of button class was necessary? It might be worth it to add the explanation to the PR description.

adhiamboperes (changes requested):
Hi @XichengSpencer, your PR looks good. I just had a couple of follow up comments. Please take a look and assign the PR back to me when done.

app/src/main/res/drawable/secondary_button_background.xml
XichengSpencer: image
I could locate the usage in file, but I am not sure if there is a way to know how to get to the specific fragment or activity in the app?

app/src/main/res/layout-land/resume_lesson_fragment.xml
XichengSpencer: The use of material button has better control for the icon inside the button so that it can group icon and text together as responsive design guidelines suggested. Particularly the icon gravity part that attached the icon to the text.

app/src/main/res/drawable/secondary_button_background.xml
adhiamboperes: We discussed this in the team meeting

adhiamboperes (approved):
Thanks @XichengSpencer, I don't think there's anything more to be done here.

PR #4999, opened 2023-05-29
Fix part of #5025: App and OS Deprecation Milestone 2 - Add protos and the DeprecationController

kkmurerwa:
## Explanation

Fix part of #5025: When this PR is merged, it will;
- Add a new `deprecation.proto` file that will allow the storage of deprecation responses as well as provide the various deprecation types.
- Add the `OPTIONAL_UPDATE_AVAILABLE` and the `OS_IS_DEPRECATED` startup modes on the `onboarding.proto` file for the two new startup modes being introduced.
- Create a `DeprecationController` class and add tests for the class in the `DeprecationControllerTest`.
- Modify BUILD.bazel files to provide the new proto files and the deprecation controller.
- Add `DeprecationControllerTest` to the `OppiaParameterizedTestRunner` file exemptions on the `file_content_validation_checks.textproto`.

## Essential Checklist


kkmurerwa:
Hey @BenHenning  and @adhiamboperes. Here is the PR for the second milestone of the App/OS Deprecation Feature. PTAL.

BenHenning:
Thanks @kkmurerwa! I'm going to be delegating this over to @adhiamboperes to approve since she'll be stepping in for me for codeowners starting this week through all of June.

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+enum DeprecationNoticeType {
+    // Unspecified notice type
+    DEPRECATION_NOTICE_TYPE_UNSPECIFIED = 0;
+
+    // App update notice
+    DEPRECATION_NOTICE_TYPE_APP_UPDATE = 1;
+
+    // OS update notice
+    DEPRECATION_NOTICE_TYPE_OS_UPDATE = 2;
+}
```
adhiamboperes: Please add a new line to eof

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
+
+package model;
+
+import "version.proto";
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Top-level proto used to store deprecation messages
```
adhiamboperes:
```suggestion
// Top-level proto used to store available deprecation messages.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+message DeprecationResponseDatabase {
```
adhiamboperes, line 12: Are these called responses because this model is used to decode server responses? Could we perhaps prefer `DeprecationMessage` or `DeprecationNotice`?

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+    // Stores the response to a request for an optional update.
+    DeprecationResponse app_deprecation_response = 1;
+
+    // Stores the response to a dialog notifying the user of an OS deprecation.
+    DeprecationResponse os_deprecation_response = 2;
```
adhiamboperes: Do we only ever expect to store one of each response at a time? This database would only contain 2 items at a time, is this by design?

I have added a comment in `DeprecationNoticeType` below that provides an alternative.

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Top-level proto used to store deprecation messages
+message DeprecationResponseDatabase {
+    // Stores the response to a request for an optional update.
+    DeprecationResponse app_deprecation_response = 1;
+
+    // Stores the response to a dialog notifying the user of an OS deprecation.
```
adhiamboperes:
```suggestion
    // Stores a response that indicates that the user should be notified that their OS is deprecated.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+package model;
+
+import "version.proto";
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+// Top-level proto used to store deprecation messages
+message DeprecationResponseDatabase {
+    // Stores the response to a request for an optional update.
```
adhiamboperes:
```suggestion
    // Stores a response that indicates that a user should be notified that an optional update is available.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+    DeprecationResponse app_deprecation_response = 1;
+
+    // Stores the response to a dialog notifying the user of an OS deprecation.
+    DeprecationResponse os_deprecation_response = 2;
+}
+
+// Stores the response to a request for an optional or OS update.
+message DeprecationResponse {
+    // The version code of the app in which the user has skipped.
```
adhiamboperes: I do not understand what this means. Did the user fail to install the last available version before this one that is available? What information is this field actually supposed to convey?

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+    int32 deprecated_version = 1;
+
+    // The number of milliseconds since epoch corresponding to the date that the user has dismissed
+    // the update notice dialog.
+    uint64 date_skipped_timestamp_millis = 2;
```
adhiamboperes: Suggestion: `notice_dismissed_timestamp_millis`

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+
+    // The type of deprecation notice to be displayed to the user.
+    DeprecationNoticeType deprecation_notice_type = 3;
+}
+
+// Distinguishes between the different deprecation notices to be displayed to the user.
```
adhiamboperes:
```suggestion
// Corresponds to/represents the different deprecation notices that can be shown to the user.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,42 @@
...
+enum DeprecationNoticeType {
```
adhiamboperes, line 38: 1. If you split `DEPRECATION_NOTICE_TYPE_APP_UPDATE` into  `FORCED_APP_UPDATE` and `OPTIONAL_APP_UPDATE`, you would not need to have two fields at DeprecationResponseDatabase level. You could instead have a `repeated DeprecationResponse` which stores a list of DeprecationResponses, and to know which type of response it is, a caller would need to do `deprecationResponsesList[index].deprecation_notice_type`
2. You could do away with the `DEPRECATION_NOTICE_TYPE` prefix, except for the `UNSPECIFIED` field.

model/src/main/proto/onboarding.proto
```diff
@@ -24,6 +24,15 @@ message AppStartupState {
     // continue using it. Instead, they should be shown a prompt suggesting that they update the app
     // via the Play Store.
     APP_IS_DEPRECATED = 3;
```
adhiamboperes, line 26: Perharps `FORCED_UPDATE_AVAILABLE` to correspond to `OPTIONAL_UPDATE_AVAILABLE` below?

adhiamboperes (changes requested):
Thanks @kkmurerwa. 
I have only reviewed the protos since there are changes that may cause you to update some parts of your implementation. I have suggested a way to structure your deprecation responses, and also requested changes to the field documentation since some of the comments are not clear to me. I think we should aim for clarity on how these models are used because this is impotant for maintenability.

model/src/main/proto/deprecation.proto
adhiamboperes: On second look, this suggestion may not be correct because it is not clear what this message is meant to represent.

model/src/main/proto/deprecation.proto
adhiamboperes: Same comment on clarity.

model/src/main/proto/deprecation.proto
kkmurerwa: I named theme responses because the user responds to the dialog prompt. I thought it would be a suitable name for the model.

model/src/main/proto/deprecation.proto
kkmurerwa: I had originally designed a database similar to the one on profile, one which can take a map of responses. However, we had a discussion with Ben where he suggested the new approach of using only two responses because that's all we are going to need. No need keeping responses from 10 versions ago, which might not be used. We only keep a response for the latest app version.

model/src/main/proto/deprecation.proto
kkmurerwa: The os_deprecation_response object will be used to store the response from the user after they have been shown a dialog that their OS is no longer supported. This will always be an acknowledgment since the user doesn't have a lot of options.

kkmurerwa: I can try and rename it based on the suggestion to make it easier to understand, since it clearly doesn't pass the right message if you didn't understand what it is supposed to do.

model/src/main/proto/deprecation.proto
kkmurerwa: Again, this object will store the response once a user is shown a dialog indicating that an optional update is available. The user only has two options, either update the app or dismiss the update. If they update the app, no response is stored. If they, however, choose to dismiss the update, this response will be stored on the DeprecationDatabase.

kkmurerwa: Again, let me work on rephrasing the comment to make it clearer.

model/src/main/proto/deprecation.proto
kkmurerwa: This is just a model for the response from the user. It is a structure for the response from the user once they have been shown a dialog notifying them that their app version or their os is now deprecated.

model/src/main/proto/deprecation.proto
kkmurerwa: I will update this.

model/src/main/proto/deprecation.proto
kkmurerwa: This will be included in the next commit.

model/src/main/proto/deprecation.proto
kkmurerwa: 1. We will not be logging responses to forced app updates. The options available for the forced app updates are closing the app and updating the app, and are not affected by previous responses, which is why we won't be storing the forced_update responses. I have explained why we shifted from having a list of responses above. 
2. I will update this.

model/src/main/proto/onboarding.proto
kkmurerwa: I don't understand this. Would you mind clarifying it?

model/src/main/proto/deprecation.proto
adhiamboperes: Okay, makes sense.

model/src/main/proto/onboarding.proto
adhiamboperes: I meant APP_IS_DEPRECATED means the same as should force update, hence could be renamed. However feel free to ignore this suggestion.

model/src/main/proto/deprecation.proto
adhiamboperes: Got it, thanks!

model/src/main/proto/deprecation.proto
kkmurerwa: I always forget to do this with new files. I have added one now.

model/src/main/proto/deprecation.proto
kkmurerwa: I have modified this and all comments on the protos to offer a better description since I realized that most of the comments were not clear.

model/src/main/proto/onboarding.proto
kkmurerwa: That's what I am using it for. APP_IS_DEPRECATED will show that the user is required to update the app.

kkmurerwa:
Hey @adhiamboperes. This is ready for another pass. PTAL.

domain/src/main/java/org/oppia/android/domain/onboarding/BUILD.bazel
```diff
@@ -9,14 +9,17 @@ kt_android_library(
     name = "state_controller",
     srcs = [
         "AppStartupStateController.kt",
+        "DeprecationController.kt",
```
adhiamboperes, line 12: Why did you prefer adding this as an extra src here over creating its own kt_android_library and including  that here as a dependency instead?

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+ */
+@Singleton
+class DeprecationController @Inject constructor(
+  cacheStoreFactory: PersistentCacheStore.Factory,
+  private val oppiaLogger: OppiaLogger,
+  private val dataProviders: DataProviders
+) {
+  /**
+   * Create an instance of [PersistentCacheStore] that contains a [DeprecationResponseDatabase].
+   */
```
adhiamboperes: Nit: Could this fit on a single line?

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+  private val deprecationStore by lazy {
+    cacheStoreFactory.create(
+      "deprecation_store",
+      DeprecationResponseDatabase.getDefaultInstance()
+    )
+  }
+
+  /**
+   * Enum states for the possible outcomes of a deprecation action.
+   */
```
adhiamboperes: Nit: This also looks short enough to fit on a single line.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+  private enum class DeprecationActionStatus {
+    /** Indicates that the deprecation operation succeeded. */
```
adhiamboperes:
```suggestion
    /** Indicates that the deprecation response write/store operation succeeded. */
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+    cacheStoreFactory.create(
+      "deprecation_store",
+      DeprecationResponseDatabase.getDefaultInstance()
+    )
+  }
+
+  /**
+   * Enum states for the possible outcomes of a deprecation action.
+   */
+  private enum class DeprecationActionStatus {
```
adhiamboperes:
```suggestion
  private enum class DeprecationResponseActionStatus {
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+    /** Indicates that the deprecation operation succeeded. */
+    SUCCESS,
+
+    /** Indicates that a deprecation write operation failed. */
```
adhiamboperes:
```suggestion
    /** Indicates that a deprecation response write operation failed. */
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+    FAILED_TO_STORE_DEPRECATION_RESPONSE,
+
+    /**
+     * Indicates that a deprecation read operation failed. This is usually the result when a
```
adhiamboperes:
```suggestion
     * Indicates that a deprecation response read operation failed. This is usually the result when a
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+     * requested [DeprecationResponse] was not found.
+     */
+    DEPRECATION_RESPONSE_NOT_FOUND
+  }
+
+  init {
+    /**
+     * Prime the cache ahead of time so that the deprecation response can be retrieved
+     * synchronously.
+     */
```
adhiamboperes: Nit: can be presented as comments instead of a KDoc

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+    }.invokeOnCompletion { primeFailure ->
+      primeFailure?.let {
+        oppiaLogger.e(
+          "DeprecationController",
+          "Failed to prime cache ahead of data retrieval for DeprecationController.",
+          primeFailure
+        )
+      }
+    }
+  }
```
adhiamboperes: Since we are not actually setting any values during priming, you can refactor this as:
```suggestion
    deprecationStore.primeInMemoryAndDiskCacheAsync(
      updateMode = PersistentCacheStore.UpdateMode.UPDATE_ALWAYS,
      publishMode = PersistentCacheStore.PublishMode.PUBLISH_TO_IN_MEMORY_CACHE
    ) .invokeOnCompletion { primeFailure ->
      primeFailure?.let {
        oppiaLogger.e(
          "DeprecationController",
          "Failed to prime cache ahead of data retrieval for DeprecationController.",
          primeFailure
        )
      }
    }
  }
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+
+  private val deprecationDataProvider by lazy { computeDeprecationProvider() }
+
+  private fun computeDeprecationProvider(): DataProvider<DeprecationResponseDatabase> {
+    return deprecationStore.transform(DEPRECATION_PROVIDER_ID) { deprecationResponsesDatabase ->
+      DeprecationResponseDatabase.newBuilder().apply {
+        appDeprecationResponse = deprecationResponsesDatabase.appDeprecationResponse
+        osDeprecationResponse = deprecationResponsesDatabase.osDeprecationResponse
+      }.build()
+    }
```
adhiamboperes: No computation is actually happening here, so the method name is misleading. This looks like you are retrieving an instance of the cachestore/db.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+   * affects what initial app flow the user is directed to.
+   */
+  fun getDeprecationDatabase(): DataProvider<DeprecationResponseDatabase> = deprecationDataProvider
+
+  /**
+   * Stores a new [DeprecationResponse] to the cache.
+   *
+   * @param deprecationResponse the deprecation response to be stored.
+   * @return [AsyncResult] of the deprecation action.
+   */
```
adhiamboperes: Nit: no punctuation at the end of an at-clause's fragment (we only add punctuation if multiple sentences are needed).

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+  fun setDeprecationResponse(deprecationResponse: DeprecationResponse): DataProvider<Any?> {
```
adhiamboperes: Perharps save or record? Set has idiomatic usages as in getters and setters.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+          .setAppDeprecationResponse(deprecationResponse)
+          .build()
+      } else {
+        it.toBuilder()
+          .setOsDeprecationResponse(deprecationResponse)
+          .build()
+      }
+
+      Pair(deprecationBuilder, DeprecationActionStatus.SUCCESS)
+    }
```
adhiamboperes:
```suggestion
      { deprecationResponseDb ->
        val deprecationBuilder = deprecationResponseDb.toBuilder().apply {
          if (deprecationResponse.deprecationNoticeType == DeprecationNoticeType.APP_DEPRECATION)
            setAppDeprecationResponse(deprecationResponse)
          else
            setOsDeprecationResponse(deprecationResponse)
        }
          .build()

        Pair(deprecationBuilder, DeprecationActionStatus.SUCCESS)
      }
```

This is a little cleaner as it avoids repetition of `  it.toBuilder()` and `build()`

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+    return dataProviders.createInMemoryDataProviderAsync(ADD_DEPRECATION_PROVIDER_ID) {
+      return@createInMemoryDataProviderAsync getDeferredResult(deferred)
+    }
+  }
+
+  /**
+   * Retrieves the [DeprecationResponse] from the cache.
+   *
+   * @param deferred a deferred instance of the [DeprecationActionStatus].
+   * @return [AsyncResult].
```
adhiamboperes: There  is typically no need to document private methods, unless extra context is required. You can keep it however.

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
```diff
@@ -0,0 +1,285 @@
...
+@Config(application = DeprecationControllerTest.TestApplication::class)
+class DeprecationControllerTest {
+  @Inject lateinit var context: Context
+  @Inject lateinit var deprecationController: DeprecationController
+  @Inject lateinit var testCoroutineDispatchers: TestCoroutineDispatchers
+  @Inject lateinit var monitorFactory: DataProviderTestMonitor.Factory
+
+  @Before
+  fun setUp() {
+    TestModule.buildFlavor = BuildFlavor.BUILD_FLAVOR_UNSPECIFIED
```
adhiamboperes: Why initialize build flavor if there are no tests relying on different build flavours?

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  @Test
+  fun testController_providesInitialState_indicatesNoUpdatesReceivedFromGatingConsole() {
+    val defaultDeprecationResponseDatabase = DeprecationResponseDatabase
```
adhiamboperes, line 63: For this and other tests it is enough to call here a public API of your controller that returns an instance of the database, and assign it to a variable. No need to initialize the database here. Your controller implementaion seems to return a default on null anyway.

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
```diff
@@ -0,0 +1,285 @@
...
+      .isEqualTo(defaultDeprecationResponseDatabase.osDeprecationResponse)
+
+    assertThat(deprecationResponseDatabase.appDeprecationResponse)
+      .isEqualTo(defaultDeprecationResponseDatabase.appDeprecationResponse)
+  }
+
+  @Test
+  fun testController_observedAfterSettingAppDeprecation_providesUpdatedDeprecationResponse() {
+    executeInPreviousAppInstance { testComponent ->
+      val appDeprecationResponse = DeprecationResponse.newBuilder().apply {
```
adhiamboperes, line 84: You could create these response objects in the companion object of this class

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.oppia.android.data.persistence.PersistentCacheStore
+import org.oppia.android.domain.oppialogger.OppiaLogger
+import org.oppia.android.util.data.AsyncResult
+import org.oppia.android.util.data.DataProvider
+import org.oppia.android.util.data.DataProviders
+import org.oppia.android.util.data.DataProviders.Companion.transform
+import javax.inject.Inject
+import javax.inject.Singleton
+
+private const val DEPRECATION_PROVIDER_ID = "deprecation_data_provider_id"
```
adhiamboperes:
```suggestion
private const val GET_DEPRECATION_RESPONSE_PROVIDER_ID = "get_deprecation_response_provider_id"
```

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,143 @@
...
+private const val ADD_DEPRECATION_PROVIDER_ID = "add_deprecation_data_provider_id"
```
adhiamboperes:
```suggestion
private const val ADD_DEPRECATION_RESPONSE_PROVIDER_ID = "add_deprecation_response_provider_id"
```

model/src/main/proto/deprecation.proto
adhiamboperes: Cool, I think we can go with updating the documentation instead.

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,49 @@
...
+
+import "version.proto";
+
+option java_package = "org.oppia.android.app.model";
+option java_multiple_files = true;
+
+
+// Top-level proto used to store deprecation responses. This proto is used to store the response to
+// a dialog notifying the user of an optional update, as well as the response to a dialog notifying
+// the user of an OS deprecation.
```
adhiamboperes:
```suggestion
// Top-level proto used to store deprecation responses, which correspond to a user's interaction with
// a dialog notifying them of either an optional update or of an OS deprecation.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,49 @@
...
+message DeprecationResponseDatabase {
+    // Stores the response to a dialog notifying the user that an optional update is available for
+    // download.
```
adhiamboperes:
```suggestion
    // Stores a user's response to a dialog notifying them that an optional update is available for
    // download.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,49 @@
...
+    DeprecationResponse app_deprecation_response = 1;
+
+    // Stores the response to a dialog notifying the user that their OS is deprecated.
```
adhiamboperes:
```suggestion
    // Stores a user's response to a dialog notifying them that their OS is deprecated.
```

model/src/main/proto/deprecation.proto
```diff
@@ -0,0 +1,49 @@
...
+    DeprecationResponse os_deprecation_response = 2;
+}
+
+// Holds the response to a dialog notifying the user that an optional update is available for
+// download, or that their OS is deprecated.
```
adhiamboperes:
```suggestion
// Represents a response to a dialog notifying the user that an optional update is available for
// download, or that their OS is deprecated.
```

Holds is synonymous with stores, which makes this sound like a data store.

adhiamboperes (changes requested):
Hi @kkmurerwa, I have requested some changes, most of which are suggestins for improvements. PTAL, thanks.

domain/src/main/java/org/oppia/android/domain/onboarding/BUILD.bazel
kkmurerwa: I usually try to make the most minimal changes I can. That's the only explanation. But I have moved the DeprecationController to a separate dependency/target and updated the test BUILD file to reflect this change.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: It fits snugly with no extra character to spare but this is done now.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Changed to a single-line Kdoc.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: This Kdoc has now been updated.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have renamed this enum to the more descriptive name.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have refactored this kdoc to include the suggestion.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: This suggestion has been implemented.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have changed this into a comment as suggested.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Nice spot. I have updated the method and tested it and it works as expected with more concise code.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have renamed this method from `computeDeprecationProvider()` to `fetchDeprecationProvider()`. This should be a bit better.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I noticed this inconsistency on your PR as well. Some parts of the existing code use fullstops at the end of @param and @return kdocs. I think we should come up with a common standard going forward. I have not made any action on this but I will once we have had a chat about this.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: This has been renamed to `saveDeprecationResponse`.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have implemented this change.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Noted. I have not made any changes to this section of the code.

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
kkmurerwa: I have noticed that this is unnecessary. I have removed the entire @Before setup method.

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
kkmurerwa: The test check if the initial condition is as expected which is why I am creating a default instance of the class.

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
kkmurerwa: The responses are only used once in the entire test class so I decided it was better if the variable remained in its related method. I have not made any changes on this yet as well because of that reason.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have renamed this variable.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have renamed this variable based on the suggestion.

model/src/main/proto/deprecation.proto
kkmurerwa: These comments have been updated based on your suggestions. They should be better now.

model/src/main/proto/deprecation.proto
kkmurerwa: I have updated this and other instances where the same suggestion could be applied.

model/src/main/proto/deprecation.proto
kkmurerwa: This has been updated.

model/src/main/proto/deprecation.proto
kkmurerwa: I have updated the comment here.

domain/src/main/java/org/oppia/android/domain/onboarding/BUILD.bazel
adhiamboperes: I think controllers should always have their own target so that they can be included independently if needed elsewhere.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
adhiamboperes: It seems that you updated the wrong kdocs. Please recheck.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
adhiamboperes: I mentioned this per Ben's previous comment:  https://github.com/oppia/oppia-android/pull/4916/files/3dd5e8dac6277379ac5dadd924f552f8771d5953..dd0f8a811e33f0fcef6eee32c1498a3af1704807#r1213870876, it seems that the convention in oppia-android is to only add punctuation for an @-clause sentence if it is multiline

domain/src/test/java/org/oppia/android/domain/onboarding/DeprecationControllerTest.kt
adhiamboperes: It looks like each app/os response is used twice, but I don't have strong feelings in this case.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
adhiamboperes: Also write/store are synomyms, so you could pick just one.

adhiamboperes (changes requested):
Hi @kkmurerwa, I have left a few further comments+some previoujs comments are not yet resolved. PTAL!

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Please check if this is resolved.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: I have updated this. I have also updated kdocs for the other enum items.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Ah. I see it now. I have made the changes to remove punctuation with that latest commit.

kkmurerwa:
Hi @adhiamboperes. I have pushed an updated commit. Let me know if something else is remaining.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
```diff
@@ -0,0 +1,132 @@
...
+    deferred: Deferred<DeprecationResponseActionStatus>
+  ): AsyncResult<Any?> {
+    return when (deferred.await()) {
+      DeprecationResponseActionStatus.SUCCESS -> AsyncResult.Success(null)
+      DeprecationResponseActionStatus.FAILED_TO_STORE_DEPRECATION_RESPONSE -> AsyncResult.Failure(
+        Exception("Failed to store deprecation response")
+      )
+      DeprecationResponseActionStatus.DEPRECATION_RESPONSE_NOT_FOUND -> AsyncResult.Failure(
+        Exception("Deprecation response not found")
+      )
```
adhiamboperes: I noticed that the convention is to define custom exceptions while returning a Deferred Result. PTAL at  [ProfileManagementController](https://github.com/oppia/oppia-android/blob/f7cf96a7b515e1639e97be0ac9311bb0fbc4ccca/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt#L99) for example

adhiamboperes (approved):
Hi @kkmurerwa I have left one additional comment, but otherwise this LGTM.

domain/src/main/java/org/oppia/android/domain/onboarding/DeprecationController.kt
kkmurerwa: Hi Adhiambo. I have been looking at the DeprecationController and I realized that only the `DeprecationResponseActionStatus.SUCCESS `is used. The other two statuses are never used. Can we remove them or do they stay?

adhiamboperes: You can remove them for now. They can be re-added on an as-needed basis in the future.

kkmurerwa: I have removed them. PTAL

Issue #4470, opened 2022-07-27
Handle configuration change using onSavedInstance


vrajdesai78:
**Problem**
For all interactions in exploration player and question player when we rotate our device, selected or entered answer doesn't persist. 

**Solution**
To solve this problem of configuration change we use onSavedInstance method to store the last entered user answer. 

**Alternatives**
- **Singleton Class Approach**: Complextiy of code will be increased if we use this approach as we have to write additional singleton class. Also, we need to handle issue of memory leak separately as it is not life-cycle aware. 
- **ViewModel Approach**: It is difficult to handle interactions when we have more than two interactions on one screen as all interactions are tightly coupled. 

Note: Closing issue #1737 as the approach for configuration change is changed to onSavedInstance from Singleton class approach.

masclot:
The ui/presenter is subscribed to get EphemeraState updates [here](https://github.com/oppia/oppia-android/blob/324023bb727007572fd2ab233c2e106a39bfe9eb/app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt#L279). The subscription is not removed after first data comes, so theoretically new data can come at any time that recreates the ui. In this case, the user input is also lost. Should I handle this case as well or in practice it won't happen?

If it can happen and I need to handle the case, what changes can I expect? could the interaction id have changed? Even if the interaction id is the same, can it belong to a different question, in which case the saved data must be discarded?

Potential scenario I'm imagining:
- The student opens Oppia both in smartphone and pc/web
- Student opens the same exploration in both
- Student types something in both, but submits only one

Another potential case:
- The student opens Oppia android
- Student navigates to an exploration and starts typing and answer
- There is a change in the exploration configuration received over the net (Is this actually possible?)
- The new exploration configuration changes the interaction id from a text input to multiple choice
- The Ui/presenter receives the new exploration configuration and recreates the view
- The student input is lost

@BenHenning 
@adhiamboperes

BenHenning:
[...]

PR #4668, opened 2022-10-19
Fixes #4470, #4471, #4472, #4473, #4474, #4708: Handle configuration change using on saved instance

vrajdesai78:
## Explanation
Fixes #4470: State of input interactions is retained on configuration change using onSavedInstance. 
Fixes #4471: Retain state for Text based interactions
Fixes #4472: Retain state for ImageRegion
Fixes #4473: Retain state for DragAndDrop
Fixes #4474: Retain State for SelectionInteraction
Fixes #4708: Fixed issue of returning null view in onSubmit time error

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

### Demo Video

https://user-images.githubusercontent.com/43074241/200193303-e1a69750-d0a2-4b4f-a8ac-84e419746e86.mp4

### Passing esspresso tests

Screenshot from 2022-11-07 00-50-29
Screenshot from 2022-11-07 00-50-00
Screenshot from 2022-11-07 00-34-11
Screenshot from 2022-11-07 00-33-47
Screenshot from 2022-11-07 00-31-04
Screenshot from 2022-11-07 00-30-08
Screenshot from 2022-11-07 00-29-21
Screenshot from 2022-11-07 00-20-01
Screenshot from 2022-11-13 08-30-37
Screenshot from 2022-11-13 08-29-58
Screenshot from 2022-11-13 08-27-40
Screenshot from 2022-11-13 07-26-20
Screenshot from 2022-11-13 08-05-20
Screenshot from 2022-11-13 08-04-47
Screenshot from 2022-11-13 08-03-54
Screenshot from 2022-11-13 08-01-50
[...]

PR #5008, opened 2023-05-31
Fix part of #4470: Save state of pending answer on configuration change in StateFragment

masclot:
## Explanation

Fix part of #4470: Save state of pending answer on configuration change in StateFragment. This PR only creates the infrastructure and applies to one single answer type: SelectionInteraction. If the approach is accepted, it can then be applied to other answer types in follow-up PRs.

The state is saved in a Bundle, using onSaveInstanceState. The ViewModels are not used as they are recreated on every configuration change.

[Demo onSaveInstanceState in StateFragment.webm](https://github.com/oppia/oppia-android/assets/103062089/94e589dd-13e4-4182-ad44-2a5fa58d752b)

## Essential Checklist

## For UI-specific PRs only


masclot:
@adhiamboperes PTAL
Please note that this is a draft. I'm looking for a LGTM for the strategy. I'll add tests once I have this draft LGTM.

adhiamboperes (changes requested):
Hey @masclot, thanks for putting this up.

While this is a good start, I am wondering if there is a way to do this without passing the entire Bundle first to the presenter then to the viewmodel?

An approach I saw in #4668 is to use a proto model to hold the answer, and onCreateView(), pass this proto to the presenter and subsequently to the view model.
This has the advantage that we can define keys for various information we want to restore, and we will retrieve them as needed, and only pass to the viewmodel information that is relevant.

I have previously experienced scenarios where a bundle is too large and that causes an app crash.

masclot:
> Hey @masclot, thanks for putting this up.
> 
> While this is a good start, I am wondering if there is a way to do this without passing the entire Bundle first to the presenter then to the viewmodel?
> 
> An approach I saw in #4668 is to use a proto model to hold the answer, and onCreateView(), pass this proto to the presenter and subsequently to the view model.
> This has the advantage that we can define keys for various information we want to restore, and we will retrieve them as needed, and only pass to the viewmodel information that is relevant.
> 
> I have previously experienced scenarios where a bundle is too large and that causes an app crash.

I'm not sure I understand. I see two options:

The entire thing needs to be stored in a Bundle (it is a requirement of onSaveInstanceState), so if the Bundle is too big, the app will crash. If we use a protocol buffer as an intermediate data format, that's fine to me, but please take into account that we are doubling the memory footprint.


Alternatively, if what you're saying is to not store the data in the Bundle, but to use onSaveInstanceState as a signal to store the data in another medium, be it memory or disk, this approach has the following problems from the top of my head:
- onSaveInstanceState must run on UI thread. You cannot do IO there. If the data is saved to disk, it must be scheduled to a background thread and then there is no guarantee the OS will allow it to run.
- if storing the data in memory, then we are basically either rewriting ViewModels (which I wouldn't do) or creating a new ViewModel to store the data of the other ViewModels (in the right way this time). ViewModels are faster than Bundles, but cover less cases than onSaveInstanceState as Ben explained.

Do you have any other option in mind?

adhiamboperes:
> > Hey @masclot, thanks for putting this up.
> > While this is a good start, I am wondering if there is a way to do this without passing the entire Bundle first to the presenter then to the viewmodel?
> > An approach I saw in #4668 is to use a proto model to hold the answer, and onCreateView(), pass this proto to the presenter and subsequently to the view model.
> > This has the advantage that we can define keys for various information we want to restore, and we will retrieve them as needed, and only pass to the viewmodel information that is relevant.
> > I have previously experienced scenarios where a bundle is too large and that causes an app crash.
> 
> I'm not sure I understand. I see two options:
> 
> The entire thing needs to be stored in a Bundle (it is a requirement of onSaveInstanceState), so if the Bundle is too big, the app will crash. If we use a protocol buffer as an intermediate data format, that's fine to me, but please take into account that we are doubling the memory footprint.
> 
> Alternatively, if what you're saying is to not store the data in the Bundle, but to use onSaveInstanceState as a signal to store the data in another medium, be it memory or disk, this approach has the following problems from the top of my head:
> 
> * onSaveInstanceState must run on UI thread. You cannot do IO there. If the data is saved to disk, it must be scheduled to a background thread and then there is no guarantee the OS will allow it to run.
> * if storing the data in memory, then we are basically either rewriting ViewModels (which I wouldn't do) or creating a new ViewModel to store the data of the other ViewModels (in the right way this time). ViewModels are faster than Bundles, but cover less cases than onSaveInstanceState as Ben explained.
> 
> Do you have any other option in mind?

To clarify, my suggestion was to use a proto model to represent the data that we are storing in the bundle.
For example, if we had 3 different objects we wanted to store in the saved bundle, we could represent them as named objects of a class, so that when we need a particular piece of data, 1 of the 3, we would retrieve it using the key associated with it, then send it to the viewmodel for processing. This is different from the current approach in this draft, where you are passing the entire bundle to the viewmodel.

masclot:
I changed the PR to use protos to save the state
@adhiamboperes PTAL

adhiamboperes:
Hi @masclot, I'll look this on Wednesday.

adhiamboperes:
Hey @masclot, I have looked into this and I am unable to approve this implementation approach for various reasons:
1.  We are saving an entire InteractionObject in the savedInstance bundle, and I don't hink this is advisable. We should instead prefer to save just the [simple information](https://proandroiddev.com/customizing-the-new-viewmodel-cf28b8a7c5fc) necesarry to restore the UI such as the selected item ids, etc. 
2. I have read up a bit on this and there seems to be a Saved State module for viewmodels that we can use, which sounds much simpler to me, and it is not clear to me why the issue description discourages using viewmodels. I am not sure if @vrajdesai78 considered the use of the [Savedstate](https://developer.android.com/jetpack/androidx/releases/savedstate) library. I would also like to know some examples of screens with multiple interactions.
3. Overall, I think this should be shelved until we can support the decision to use savedInstance as the best approach here.
I am going to unassign you from this issue for now.

Aside from the above, I think your overall approach was good. Thanks!

masclot:
> 1. We are saving an entire InteractionObject in the savedInstance bundle, and I don't hink this is advisable. We should instead prefer to save just the [simple information](https://proandroiddev.com/customizing-the-new-viewmodel-cf28b8a7c5fc) necesarry to restore the UI such as the selected item ids, etc.

We are only storing the selected item ids. InteractionObject is a one-of. This object is the same that is sent on submit so it contains only what's needed. I could create another object just for saving the state, but my intuition tells me that the new object will end up looking very similar to InteractionObject, if not equal, once we add support for the other interaction types. That said, if you prefer that, I am also fine with it.

masclot:
> 2\. I have read up a bit on this and there seems to be a Saved State module for viewmodels that we can use, which sounds much simpler to me, and it is not clear to me why the issue description discourages using viewmodels. I am not sure if @vrajdesai78 considered the use of the [Savedstate](https://developer.android.com/jetpack/androidx/releases/savedstate) library. I would also like to know some examples of screens with multiple interactions.

First time I hear about SavedState in ViewModels. Thanks for sharing!
Remember though that we are not using ViewModels properly in recently played UI: the ViewModels are recreated new, every time by Dagger, so the SavedState library won't help.

The only workaround I see, without fixing the current ViewModel usage, is to create yet another ViewModel, non-dagger, just for the saved state, and have the new ViewModel and the old ones talk to each other. I think this would be confusing and probably more complicated than onSavedInstanceState.

masclot:
> 3\. Overall, I think this should be shelved until we can support the decision to use savedInstance as the best approach here.
>     I am going to unassign you from this issue for now.

Ok. I'm interested in knowing the result and motivation of the final decision, even if the issue is assigned to someone else.

adhiamboperes:
> > 3. Overall, I think this should be shelved until we can support the decision to use savedInstance as the best approach here.
> > I am going to unassign you from this issue for now.
> 
> Ok. I'm interested in knowing the result and motivation of the final decision, even if the issue is assigned to someone else.

Let's discuss this a bit more at the meeting.

adhiamboperes:
> > 3. Overall, I think this should be shelved until we can support the decision to use savedInstance as the best approach here.
> > I am going to unassign you from this issue for now.
> 
> Ok. I'm interested in knowing the result and motivation of the final decision, even if the issue is assigned to someone else.



@Vishwajith-Shettigar, could you please summarize the solution you created for this issue?

Vishwajith-Shettigar:
@masclot, @adhiamboperes, to retain user input when device configurations change, I used savedInstanceState. Whenever the user enters input, we store it in savedInstanceState before the activity gets destroyed and use it after the activity is recreated. We only store necessary input data along with the answer error category. For text-based inputs(eg:- number, fraction, ratio, math expressions, text inputs), we store only the text, and for selection-based inputs, we store only the index of the selected items.

Here is the structure:
```
// Represents the state of item selection answers that the user has selected.
message ItemSelectionAnswerState {
  // List of selected item indices.
  repeated int32 selected_indexes = 1;
}

// Represents the user's answer state, retained across configuration changes.
message UserAnswerState {
  // Represents the last error state.
  AnswerErrorCategory answer_error_category = 1;

  // Type of answer input:
  oneof answer_input_type {
    // User's selection for selection input interactions.
    ItemSelectionAnswerState item_selection = 2;
    // Raw answer entered by the user in text-based interactions.
    string text_input_answer = 3;
  }
}

```

Saved answers will be removed when the user hits the submit button to ensure that no saved data will persist into upcoming interactions. This also ensures that no data pile-up occurs.

PR #5009, opened 2023-06-01
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates

PR #4762, opened 2022-11-24
Fix #4606: Language selector feature implementation

KevinGitonga:
## Explanation

Fixes part of #52
Fixes #4606

 Feature implementation for #4606
  - This PR introduces the ability to change App level language through the Options setting. It has involved below changes. 
      1.   Refactored the codebase to accomodate feature changes by changing from use of string variable to OppiaLanguage 
            proto enum object.
      2. Add Changes to allow display of languages loaded from proto in a human readable form similar to AudioLanguage 
           representation. 
  - When fully functional this PR should allow users to modify app language in App settings regardless of the language setting of their device. PersistentCache is used to locally persist language which has been selected by a user.

See also https://github.com/KevinGitonga/oppia-android/pull/5 for an additional explanation of more specific changes contained within this PR.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-(A11y)-Guide))
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

## Video demo of changes
https://user-images.githubusercontent.com/20886444/222683790-7b2d4fa8-7068-4774-8d1a-0494d4f9adf0.mp4

PR #5010, opened 2023-06-02
Fix #52: Finalize release 0.11

### LINKED ISSUE #52: Language selection screen [Blocked: #20, #44]

BenHenning:
## Explanation
Fixes #52

This PR finishes up a bunch of separate tasks relating to the 0.11 release of the Oppia Android app.

### Release plans
Note that a lot of the tasks addressed were completed out-of-band (hence the lack of tracking issues). These are needed to prepare for the following release changes:
- A new version of the alpha app will be launched to provide a basis of testing spotlights
- A new version of the Kenya alpha app will be launched (since those users are still in transitionary state to the beta version of the app until later this year)
- A new version of the beta version of the app to include dark mode, the new language selector, and more

#### Language support changes
- Arabic and Nigerian Pidgin (Naija) languages were added to production builds (Naija was added generally as it wasn't supported previously). Both languages were also added as possible audio languages.
- Some basic tests were added to cover Arabic and Naija now being available, though some tests ended up being omitted due to the inherent complexity of testing these particular languages in certain cases (Arabic because it's RTL and Naija because Android doesn't actually support it natively--see below explanation).
- Also added Nigeria as a language region. We don't yet have a strong grasp on the regions in which Arabic will be used, so that's being kept as a broad language for now. We may refine this in the future.
- The translations for Nigerian Pidgin are added here as a replacement to #5009 since the original translations had some formatting issues that were identified by @adhiamboperes and subsequently fixed by our translation volunteers. Since we can't wait for another push to #5009, this PR is just introducing the strings directly.
- ``AndroidLocaleFactory`` was largely rebuilt to make better use of code sharing, but its actual functionality needed to change due to the Naija support problems mentioned below.

#### New feature: spotlights (alpha-only)
- A new alpha module was added to allow the team to stage features specific to alpha builds, and spotlights were added as an initial use case for this new capability.
- As part of testing spotlights I noticed that the background overlay was too dark (it was fully occluding the background), so I changed it to something that's dark but still semi-translucent. See the UI part of the description below for a before-and-after.

#### New feature: in-app language selector (all users)
- This PR enables the new language selector introduced by #4762 by default as part of addressing #52.
- Some small fast-follow nits from #4762 were addressed (see https://github.com/oppia/oppia-android/pull/4762#discussion_r1213693593 and https://github.com/oppia/oppia-android/pull/4762#discussion_r1213698884).

#### Event system changes
- Introduced a new script for decoding the compressed base 64 event strings that can be generated during user studies (see explanation below for details). This has some basic fault tolerance built into it so that truncated event strings can still be partially decoded.
- A debug event property was added which essentially amounts to an event count since app startup that's attached to each event. While this won't help us identify events outright missing, it will help us identify events missing between received events. This is aimed to help an ongoing investigation that has found the high probability that events are being lost between user action and Firebase's storage layer.

#### Infrastructure changes
- The minor version of the app is being bumped (since this is a major beta release of the app).
- Version codes were bumped for 2 releases (since one re-release of the beta version of the app was needed a few weeks back due to it expiring).
- This PR also quiets the output when creating AABs since currently ~5k lines are outputted during the final bundle assembly steps and this output is never useful to the user unless there's a build pipeline failure.
- This PR fixes an error in the resource filtering output that is provided at the very end of the AAB creation process (previously it was outputting the number of configurations being filtered not the number of actual resources being removed).

#### Improved support for future user studies
- The EnableLearnerStudyAnalytics has been split into 3 flags (2 new ones): one that controls just whether to attach the sensitive profile IDs to events, one that controls the fast in-lesson language switcher (and corresponding admin-controlled profile setting for it), and one that controls access to the analytics screen + admins being able to mark lessons as completed for profiles (this is gated on the existing flag).
- The analytics screen flag has been default enabled for regular alpha builds of the app so that this channel can be better utilized for user studies. The user ID and in-lesson language switcher aspects of the old learner study analytics feature are being kept off in alpha builds, only the analytics screen & related behavior is being made available.

### Nigerian Pidgin support issues

The Android system does not formally support Nigerian Pidgin (Naija) despite the fact that it has a recognized ISO 639-3 language code.

Naija is a creole language which means it's mostly a derivative of English with vernacular adaptations (such as recreated words and grammar alterations). I think that because it's mostly a derivative language and is mainly used for casual speaking (vs. formal communications), there isn't as much of a desire on the Android side to formally support it as a UI language.

Fortunately, the Oppia Android app's localization system already supports forcing the system to rendering app strings that are not natively supported. However, this results in several considerations and has turned up one new issue that needed addressing:
- Custom language overriding only works by matching _both_ language and region, so the language strings needed to be moved to ``values-pcm-rNG`` (for Nigeria locking). This doesn't stop the language from being used outside Nigeria, it's just a limitation in Android's resource qualifier system when working with custom languages.
- It's not yet clear how the change in the values structure will affect Translatewiki--some additional work might be needed there (which will become clear once this PR is merged & Translatewiki attempts another branch sync).
- Custom languages don't have the benefit of guaranteeing rendering support, so the language's script largely comes to mind. Fortunately, as a derivative of English Naija uses Latin-based script so there are no concerns with language rendering compatibility.
- An issue was found in the localization system when handling fallbacks. Since Naija is English-based, this PR has configured its fallback language to be English. This led to the system actually prioritizing English app string translations rather than forcing the Naija strings. This required the changes that can be seen in ``AndroidLocaleFactory``.

### ``DecodeUserStudyEventString`` utility

A new utility was introduced to decode the compressed Base64 string of events logs that can be exported from devices being used for user studies (i.e. whose admins have access to the learner analytics screen). These logs represent the entirety of both pending and uploaded logs since the introduction of the feature (but only for devices that have the learner study enabled).

The utility reads an input file containing a single instance of the string (and automatically strips newline and horizontal space formatting since the output from the app includes line-wrapping) and outputs it to one of three indicated formats: textproto, JSON, or Yaml.

The script can be run in a local terminal from this branch (or ``develop`` once it's checked in) at the repository root by running:

```sh
bazel run //scripts:decode_user_study_event_string -- $(pwd)/input.log $(pwd)/output.json
```

(where 'input.log' exists in the local repository root, and 'output.json' will be written to the local repository root).

For use outside of Bazel, one can use a pre-built deployment Jar file by running:

```sh
java -jar decode_user_study_event_string_deploy.jar input.log output.json
```

(Note that the error output for this command will be assuming Bazel is being used, so some adjustment may be needed to interpret CLI errors when using the deploy Jar).

### Third-party dependency updates
Some new dependencies were needed for the new ``DecodeUserStudyEventString`` utility to make the implementation much simpler:
- Protobuf's Java util (https://cloud.google.com/java/docs/reference/protobuf/latest/com.google.protobuf.util) for conversion from proto to json.
- A converter from JSON to Yaml (since there's no direct proto-to-yaml conversion): https://github.com/xlate/yaml-json.
- The Jakarta JSON API (required by the converter): https://jakarta.ee/specifications/platform/9/apidocs/jakarta/json/json.
- The Jakarta JSON Parsson implementation (runtime dependency required by the converter): https://github.com/eclipse-ee4j/parsson.
- Snakeyaml implementation (runtime dependency required by the converter): https://github.com/snakeyaml/snakeyaml.

This required an update to maven_install.json, but no changes are needed in tracked Maven dependency licenses since these are not dependencies that are deployed with the app to end users.

The above do result in an incidental update to Gson 2.8.6 (from the current 2.8.5 used). I don't expect that this actually will affect the release much since Gson isn't used broadly, and this is a minor update to Gson.

### Exemptions & test notes
Two new files were exempted for tests:
- ``PlatformParameterAlphaModule``: while we can test modules, I'm not actually keen on testing the platform parameter modules at this time because staging is generally messy at the moment with a lot of duplication across multiple modules, so the test complexity isn't small. A lot of the platform parameter system needs to be streamlined, but that's largely dependent on #1720 being finished.
- ``DecodeUserStudyEventString`` while not particularly difficult to test, this is a special local analytics tool that will rarely be used and is very unlikely to break. Since I'm strapped for time on this PR, I'm forgoing adding tests for this utility.

Regarding broad testing coverage, everything has had corresponding testing additions or changes where applicable. This PR might not have quite as much testing thoroughness as other PRs, but it does largely cover the core scenarios.

## Essential Checklist

## For UI-specific PRs only

Spotlights changes before/after:
| Old spotlights background | New spotlights background |
|--------|--------|
| https://github.com/oppia/oppia-android/assets/12983742/a8df7807-1bb6-45be-b53a-6d057f4a5931 | https://github.com/oppia/oppia-android/assets/12983742/e499f4b6-4609-4e58-871f-a09ebd09a80d |

Video demonstrating the new functionality in the app, including alpha access to the learner analytics screen and support for both Nigerian Pidgin and Arabic (and via the in-app language selector):

https://github.com/oppia/oppia-android/assets/12983742/c0a92393-a8a3-4634-af5b-2e7fc35ecd97


BenHenning:
@seanlip @adhiamboperes PTAL for codeowners.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
```diff
@@ -26,7 +26,7 @@ import org.oppia.android.domain.translation.TranslationController
 import org.oppia.android.util.data.AsyncResult
 import org.oppia.android.util.data.DataProviders.Companion.toLiveData
 import org.oppia.android.util.locale.OppiaLocale
-import org.oppia.android.util.platformparameter.EnableLearnerStudyAnalytics
+import org.oppia.android.util.platformparameter.EnableFastInLessonLanguageSwitching
```
seanlip: I keep getting confused when reading this name because "in lesson" isn't hyphenated, so I wonder what "fast in lesson" means.

I wonder if EnableFastLessonLanguageSwitching might be sufficient (or EnableFastLanguageSwitchingInLesson)? In fact you use enableFastLanguageSwitching further down which is somewhat more readable (though I think the "in lesson" part is probably important to keep as part of the name).

model/src/main/proto/languages.proto
```diff
@@ -30,6 +30,11 @@ enum OppiaLanguage {
 
   // Corresponds to the Swahili (Kiswahili) macro language. IETF BCP 47 language tag: sw.
   SWAHILI = 7;
```
seanlip, line 32: Just to check: are we supporting Swahili in beta too, or alpha only?

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
```diff
@@ -111,6 +110,37 @@ const val LEARNER_STUDY_ANALYTICS = "learner_study_analytics"
...
+ * Qualifier for a feature flag that controls whether learners may be allowed (via an
+ * admin-controlled setting) to use a special in-lesson button for quickly switching between content
+ * languages.
+ *
+ * This is generally expected to only be used in tandem with [EnableLearnerStudyAnalytics].
+ */
+@Qualifier annotation class EnableFastInLessonLanguageSwitching
+
+/** The platform parameter name corresponding to [EnableFastInLessonLanguageSwitching]. */
+const val FAST_IN_LESSON_LANGUAGE_SWITCHING = "fast_in_lesson_language_switching"
```
seanlip: fast_language_switching_in_lesson or fast_language_switching_in_lessons, perhaps?

seanlip (approved):
Thanks @BenHenning -- this looks amazing from a product standpoint.

FYI I am unable to see the spotlight videos due to high download latency (for some reason they're only coming in at 3.5 kb/s), but I imagine your changes should be fine and am not worried about that.
- EDIT: Finally downloaded the videos! I think both are actually too dark, in terms of background -- I can barely see the background on either. I think it's worth learners still being able to see the background because it would help them see where the spotlit element is in relation to the other elements, so I would suggest having the background be a bit lighter still.


One other thought: I imagine that translatewiki will almost certainly push to values-pcm as it has done previously (both #5009 and #5019). What would you like to happen in that case -- should we accept it? /cc @adhiamboperes

app/src/main/java/org/oppia/android/app/drawer/NavigationDrawerFragmentPresenter.kt
```diff
@@ -427,7 +427,9 @@ class NavigationDrawerFragmentPresenter @Inject constructor(
 
         override fun onDrawerClosed(drawerView: View) {
           super.onDrawerClosed(drawerView)
-          fragment.activity!!.invalidateOptionsMenu()
+          // It's possible in some rare cases for the activity to be gone while the drawer is
```
adhiamboperes, line 430: Nice catch, I noticed this behaviour on my beta app and was able to reproduce it consistently today.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
```diff
@@ -157,6 +157,9 @@ class AppLanguageResourceHandler @Inject constructor(
       AudioLanguage.FRENCH_AUDIO_LANGUAGE -> getLocalizedDisplayName("fr")
       AudioLanguage.CHINESE_AUDIO_LANGUAGE -> getLocalizedDisplayName("zh")
       AudioLanguage.BRAZILIAN_PORTUGUESE_LANGUAGE -> getLocalizedDisplayName("pt", "BR")
+      AudioLanguage.ARABIC_LANGUAGE -> getLocalizedDisplayName("ar", "EG")
+      AudioLanguage.NIGERIAN_PIDGIN_LANGUAGE ->
+        resources.getString(R.string.nigerian_pidgin_localized_language_name)
```
adhiamboperes, line 162: You're adding the direct string here because of the lack of native support you mentioned in the description?

domain/src/test/java/org/oppia/android/domain/locale/LanguageConfigRetrieverProductionTest.kt
```diff
@@ -55,21 +55,29 @@ class LanguageConfigRetrieverProductionTest {
...
   @Test
-  fun testLoadSupportedLanguages_hasThreeSupportedLanguages() {
+  fun testLoadSupportedLanguages_hasSixSupportedLanguages() {
     val supportedLanguages = languageConfigRetriever.loadSupportedLanguages()
 
     // Change detector test to ensure changes to the configuration are reflected in tests since
     // changes to the configuration can have a major impact on the app (and may require additional
     // work to be done to support the changes).
-    assertThat(supportedLanguages.languageDefinitionsCount).isEqualTo(3)
+    assertThat(supportedLanguages.languageDefinitionsCount).isEqualTo(5)
```
adhiamboperes, line 64: Did you mean 6, per test name? also if 4 languages are supported+the 2 autogenerated fields. I don't understand why this test passed. Am I missing something?

domain/src/test/java/org/oppia/android/domain/locale/LanguageConfigRetrieverTest.kt
```diff
@@ -49,21 +49,21 @@ class LanguageConfigRetrieverTest {
...
   @Test
-  fun testOppiaLanguage_hasSupportForSevenLanguages() {
+  fun testOppiaLanguage_hasSupportForEightLanguages() {
     // While it's a bit strange to test a proto, and particularly in this file, this suite is
     // generally responsible for verifying language & region configuration sanity. Part of that
     // requires verifying that all languages are tested below. Note that '9' is because the base
     // 7 languages are supported + LANGUAGE_UNSPECIFIED and UNRECOGNIZED (auto-generated by
     // Protobuf). Finally, note that the values themselves are not checked since it doesn't provide
     // any benefit (being able to reference an enum constant without a compiler error is sufficient
     // proof that constant is available).
```
adhiamboperes: The comment here needs to be updated.

adhiamboperes (approved):
Thanks @BenHenning! I do have a couple of questions but nothing blocking.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
BenHenning: Seems reasonable. Updated everything to be consistent, and I went with ``EnableFastLanguageSwitchingInLesson`` since it has the same number of characters so I wouldn't have to worry about dealing with any line-wrapping changes.

model/src/main/proto/languages.proto
BenHenning: Alpha only. The languages support for beta & GA builds are specified here: https://github.com/oppia/oppia-android/blob/develop/config/src/java/org/oppia/android/config/productionlanguages/supported_languages.textproto. The languages the app supports at all are here: https://github.com/oppia/oppia-android/blob/develop/config/src/java/org/oppia/android/config/alllanguages/supported_languages.textproto.

utility/src/main/java/org/oppia/android/util/platformparameter/PlatformParameterConstants.kt
BenHenning: This has been adjusted per your earlier suggestion. It's now fast_language_switching_in_lesson.

app/src/main/java/org/oppia/android/app/drawer/NavigationDrawerFragmentPresenter.kt
BenHenning: Ah, if you have the context for it I think it's worth filing an issue explaining how to repro it. I found it completely accidentally and couldn't repro it again (I fixed it entirely based on the stack trace & reasoning). A consistent repro may allow us to possibly write a test for it (though Robolectric differs from Android most noticeably with activity lifecycle behaviors, so it may not be possible to test in a non-convoluted way).

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
BenHenning: Correct. Ideally we'll be moving audio languages to OppiaLanguage in general and relying only on our untranslated language names for displaying them. I needed to do it here because otherwise Android doesn't know what to call the language.

domain/src/test/java/org/oppia/android/domain/locale/LanguageConfigRetrieverProductionTest.kt
BenHenning: Good catch! Will fix the name in the latest changes. It was 6 before because I mistakenly enabled Swahili for prod builds. I forgot to update the name when fixing the configuration & test.

So this isn't a case where we're testing a generated proto enum (where there are extra cases). We're actually testing the loaded proto from here: https://github.com/oppia/oppia-android/blob/develop/config/src/java/org/oppia/android/config/productionlanguages/supported_languages.textproto. With this PR, there are now 5 language configurations supported in prod versions of the app hence the change to the test (as a quick parity check, we're adding Arabic and Nigerian Pidgin to the prod languages support, so it makes sense why this is going from 3 to 5).

domain/src/test/java/org/oppia/android/domain/locale/LanguageConfigRetrieverTest.kt
BenHenning: Great catch! I've updated the comment to be generic so that it doesn't need to be changed again in the future.

app/src/main/java/org/oppia/android/app/drawer/NavigationDrawerFragmentPresenter.kt
adhiamboperes: Okay, I will file the issue.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
adhiamboperes: Got it, thanks.

BenHenning:
Thanks @seanlip & @adhiamboperes. I addressed all follow-up comments.

@adhiamboperes could you PTAL at changes & address resolved comments from both you and Sean (since Sean is probably unavailable at the moment)?

> Thanks @BenHenning -- this looks amazing from a product standpoint.
> 
> FYI I am unable to see the spotlight videos due to high download latency (for some reason they're only coming in at 3.5 kb/s), but I imagine your changes should be fine and am not worried about that.
> 
> * EDIT: Finally downloaded the videos! I think both are actually too dark, in terms of background -- I can barely see the background on either. I think it's worth learners still being able to see the background because it would help them see where the spotlit element is in relation to the other elements, so I would suggest having the background be a bit lighter still.
> 
> One other thought: I imagine that translatewiki will almost certainly push to values-pcm as it has done previously (both #5009 and #5019). What would you like to happen in that case -- should we accept it? /cc @adhiamboperes

The background comes across lighter on a device than in the video, but I lowered the darkness by another 9%. See comparison:

| Previous new background | Updated new background |
|--------|--------|
| https://github.com/oppia/oppia-android/assets/12983742/8ce7e7f8-0f8e-4a3c-bd7e-f55fce376d66 | https://github.com/oppia/oppia-android/assets/12983742/e499f4b6-4609-4e58-871f-a09ebd09a80d |

Regarding Translatewiki, I think we might be stuck there temporarily (as in, merging those PRs will result in translation duplications that are ignored by the build). We should start with asking the Translatewiki team if it's possible for us to configure the export to ensure Nigerian Pidgin translations go to values-pcm-rNG instead of just values-pcm. If that's not an option, we're going to have to look into automation to fix the location of the files during build time (which isn't too bad, but it's a bit hacky so it's not preferred). I won't have time to look into this before I'm gone.

app/src/main/java/org/oppia/android/app/player/state/StateViewModel.kt
adhiamboperes: Verified all usages renamed.

domain/src/test/java/org/oppia/android/domain/locale/LanguageConfigRetrieverProductionTest.kt
adhiamboperes: Thanks for explaining.

adhiamboperes:
> Thanks @seanlip & @adhiamboperes. I addressed all follow-up comments.
> 
> @adhiamboperes could you PTAL at changes & address resolved comments from both you and Sean (since Sean is probably unavailable at the moment)?
> 
> > Thanks @BenHenning -- this looks amazing from a product standpoint.
> > FYI I am unable to see the spotlight videos due to high download latency (for some reason they're only coming in at 3.5 kb/s), but I imagine your changes should be fine and am not worried about that.
> > 
> > * EDIT: Finally downloaded the videos! I think both are actually too dark, in terms of background -- I can barely see the background on either. I think it's worth learners still being able to see the background because it would help them see where the spotlit element is in relation to the other elements, so I would suggest having the background be a bit lighter still.
> > 
> > One other thought: I imagine that translatewiki will almost certainly push to values-pcm as it has done previously (both #5009 and #5019). What would you like to happen in that case -- should we accept it? /cc @adhiamboperes
> 
> The background comes across lighter on a device than in the video, but I lowered the darkness by another 9%. See comparison:
> 
> Previous new background	Updated new background
> https://github.com/oppia/oppia-android/assets/12983742/8ce7e7f8-0f8e-4a3c-bd7e-f55fce376d66	https://github.com/oppia/oppia-android/assets/12983742/e499f4b6-4609-4e58-871f-a09ebd09a80d
> Regarding Translatewiki, I think we might be stuck there temporarily (as in, merging those PRs will result in translation duplications that are ignored by the build). We should start with asking the Translatewiki team if it's possible for us to configure the export to ensure Nigerian Pidgin translations go to values-pcm-rNG instead of just values-pcm. If that's not an option, we're going to have to look into automation to fix the location of the files during build time (which isn't too bad, but it's a bit hacky so it's not preferred). I won't have time to look into this before I'm gone.


Re: Spotlight background, I think the latest version has acceptable visibility.

model/src/main/proto/languages.proto
adhiamboperes: Not sure if you want to merge this right now so maybe you can resolve this comment unless @seanlip gets to it first.

adhiamboperes (commented):
All the latest changes look good, both mine and @seanlip's comments have been addressed.

seanlip:
Everything looks great. Thanks!

@BenHenning Noted on your comment re translatewiki. Will send them an email.

model/src/main/proto/languages.proto
seanlip: Looks good. Thanks!

BenHenning:
Thans @adhiamboperes & @seanlip!

PR #5019, opened 2023-06-05

translatewiki:
Translation updates


BenHenning:
This is being handled as part of #5010.

PR #5027, opened 2023-06-08
Fix #4790 Talkback will read toolbar title clickable

### LINKED ISSUE #4790: Talkback shouldn't read double tap to activate for the labels that are not clickable
**Describe the bug**
Talkback shouldn't read double tap to activate for the labels that are not clickable

**To Reproduce**
Enable Talkback
1. Login to Oppia as Admin
2. Click on any of the topic 'Place Values'
3. Focus will be on the Nav bar topic title

**Expected behavior**
Talkback should read topic title and not double tap to activate as it is not clickable

**Environment**
Device/emulator being used: Oneplus nord2 5g
Android or SDK version (e.g. Android 5 or SDK 21): Android version 12
App version (you can get this through system app settings or via the admin controls menu in-app): 0.10-beta-5e64fae55e

XichengSpencer:
## Explanation

Fix #4790 For toolbars that have marquee effects, add additional checks so that the effect will only be enabled when the accessibility service is not on. 
Also updated the existing test for the marquee effect. The test now checks if the effect is set and activated correctly depending on the accessibility service status.
Update the test setting to ensure that test behavior is correctly simulated.

Whenever users use vocal assisting tools such as talkback, the accessibility service we inject in the field will be enabled. We will choose either to use a marquee effect for visual display, or just pure vocal display when the user enables the service. 

For a toolbar that doesn't have an individual view for the toolbar title, setting 'onNavgation..listener' will make the whole toolbar clickable, and talkback will read "double click.."- 
While we want to keep the navigation function, we can use getChildAt(0) to set the listener to just the nav icon.
 

## Essential Checklist

## For UI-specific PRs only

Exploration Activity:

- Normal:

https://github.com/oppia/oppia-android/assets/74568012/74f6eaf8-8560-4d1a-bfb4-d378177e31b6

- Talkback On:

https://github.com/oppia/oppia-android/assets/74568012/11285aee-ed5e-4355-97b1-c18df646efd7

Topic Fragment:

- Talkback On:

https://github.com/oppia/oppia-android/assets/74568012/f8cec635-b1c3-4540-afcb-990a42a9a38b

## Issue with Espresso test:
when readerOff, toolbar.title.click() behaves different from manual test.

- Video record for manual test 

     Listener function is set and is called when click is performed. Regardless of title length, is selected will always be modified.

https://github.com/oppia/oppia-android/assets/74568012/7f9a04b6-e3e3-4803-9465-89543b091458

- Video record for espresso test 
     Listener function is set but not called when click is performed.
     The isSelected attribute does not been modified after click
    

https://github.com/oppia/oppia-android/assets/74568012/c2984b5f-c9fe-4a35-a36f-27d05a935210

## Solution to Solve Error in Espresso Tests.

- Setting the accessibility service before starting the activity solves accessibility service not correctly set except in TopicFragmentTest.
**Reason** The set listener code in the app executes before setting the accessibility service in the test.

- Add markSpotlightSeen before the start of the activity.
**Reason** The spotlight might cause the toolbar unable to reach to simulate a click.


app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -44,6 +46,7 @@ class TopicFragmentPresenter @Inject constructor(
   fun handleCreateView(
     inflater: LayoutInflater,
     container: ViewGroup?,
+    context: Context?,
```
adhiamboperes: I think you can get the context from either the activity or fragment that has been passed here already, e.g. `activity?.let{}`.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -61,12 +64,20 @@ class TopicFragmentPresenter @Inject constructor(
     this.internalProfileId = internalProfileId
     this.topicId = topicId
 
-    binding.topicToolbar.setNavigationOnClickListener {
+    binding.topicToolbar.getChildAt(0).setOnClickListener {
```
adhiamboperes: Could you tell me why you made this change?

adhiamboperes (changes requested):
Thanks @XichengSpencer, this approach looks good, and you can look into applying this fix for other toolbars with the marquee effect.

I have left a suggestion, and don't forget to add tests for your changes.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
XichengSpencer: Since there is no individual view for toolbar title, setting 'onNavgation..listener' will make the whole toolbar clickable which talkback will read "double click.."- While we want to keep the navigation function we can use getChildAt(0) to set the listener to just nav icon.

Also added this to explanation

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
XichengSpencer: That makes sense to me, I will change it back. But just let you know that, I got this solution referring to code in ExplorationActivityPresenter.

 fun handleOnCreate(
    context: Context,
    profileId: ProfileId,
    topicId: String,
    storyId: String,
    explorationId: String,
    parentScreen: ExplorationActivityParams.ParentScreen,
    isCheckpointingEnabled: Boolean
  )
  
You can see there is a context here-
  I am wondering if this context is also removable or if it is for another purpose, would you mind helping me explain a bit?

XichengSpencer: Also, I check the other activity presenter that deals with the toolbars, this seems to be the only one with context passed in as a parameter.

adhiamboperes: Yes, and if you check in Android Studio, this parameter is actually not used anywhere.

adhiamboperes: Yes, it could be replaced, though I suspect that this was done to ensure that the Activity context was associated with the entry point, since the ExplorationActivity has multiple entry points.

Generally getting the context from an activity or fragment helps to avoid potential memory leaks because the obtained Context is tied to the lifecycle of the enclosing Activity or Fragment.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
adhiamboperes: Mhh, and does the marquee effect still work when talkback is off?

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -61,12 +64,20 @@ class TopicFragmentPresenter @Inject constructor(
...
       (activity as TopicActivity).finish()
     }
 
-    binding.topicToolbar.setOnClickListener {
-      binding.topicToolbarTitle.isSelected = true
+    context?.let {
+      val accessibilityManager = context.getSystemService(Context.ACCESSIBILITY_SERVICE)
+        as AccessibilityManager?
+      val isAccessibilityManager = accessibilityManager!!.isEnabled
```
adhiamboperes:
```suggestion
      val isAccessibilityManagerEnabled = accessibilityManager!!.isEnabled
```

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
XichengSpencer: In the activity the 
intent.extractParams()
and context in side intent is passed to presenter. 

If that's the case, by assuming it is an Activity context, will this context associates with another activity other than the Exploration Activity?

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
XichengSpencer: It is a one-time action in the handleOnCreate function.

My thought on this is that, if someone needs to use accessibility manager, they will turn it on before even entering the app. I cannot figure out a reason for users to turn on the accessibility manager in half of the way like how we test the app.

My thought on this is if we want to switch between the talkback and the marquee effect within the activity, we may need to enable the listener in onResume and disable it in onStop. But I am not sure how to control this in a presenter.

adhiamboperes: My question is actually whether the marquee functionality works for user who are not using the talkback.

XichengSpencer: Yes, it works

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
XichengSpencer: While the context in this case has a relatively complex usage. I think I should keep it as it is.

adhiamboperes: That is fine. Either way, the contexts have the same meaning.

app/src/main/java/org/oppia/android/app/help/thirdparty/LicenseTextViewerActivityPresenter.kt
```diff
@@ -44,8 +46,12 @@ class LicenseTextViewerActivityPresenter @Inject constructor(
       (activity as LicenseTextViewerActivity).finish()
     }
 
-    binding.licenseTextViewerActivityToolbarTitle.setOnClickListener {
-      binding.licenseTextViewerActivityToolbarTitle.isSelected = true
+    activity.applicationContext?.let {
```
adhiamboperes: Here and everywhere else with the change, remove the application context reference since accessibilityService [handles its own context](https://github.com/oppia/oppia-android/blob/d049a5a8ba2423802391475f355f99880b479c48/utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityServiceImpl.kt#L10)

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -315,7 +319,7 @@ class TopicFragmentTest {
   }
 
   @Test
-  fun testTopicFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly() {
+  fun testTopicFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly_whenReaderOff() {
```
adhiamboperes:
```suggestion
  fun testTopicFragment_toolbarTitle_readerOff_marqueeInRtl_isDisplayedCorrectly() {
```

We define all the test conditions first then place expected outcome at the end of the test name.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -324,18 +328,58 @@ class TopicFragmentTest {
...
     testCoroutineDispatchers.runCurrent()
+    // setup accessibility manager
+    fakeAccessibilityManager.setScreenReaderEnabled(false)
+    val topicToolbarTitle: TextView =
+      activityTestRule.activity.findViewById(R.id.topic_toolbar_title)
+    ViewCompat.setLayoutDirection(topicToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
+    onView(withId(R.id.topic_toolbar_title)).perform(click())
+    assertThat(topicToolbarTitle.ellipsize).isEqualTo(TextUtils.TruncateAt.MARQUEE)
+    assertThat(topicToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
+  }
```
adhiamboperes, line 355: Please add a new line or use ctrl+alt+L to format your code.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -324,18 +328,58 @@ class TopicFragmentTest {
...
+  @Test
+  fun testTopicFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly_whenReaderOn() {
```
adhiamboperes:
```suggestion
testTopicFragment_toolbarTitle_readerOn_marqueeInRtl_isDisplayedCorrectly() {
```

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -324,18 +328,58 @@ class TopicFragmentTest {
...
+      activityTestRule.activity.findViewById(R.id.topic_toolbar_title)
+    ViewCompat.setLayoutDirection(topicToolbarTitle, ViewCompat.LAYOUT_DIRECTION_LTR)
     onView(withId(R.id.topic_toolbar_title)).perform(click())
     assertThat(topicToolbarTitle.ellipsize).isEqualTo(TextUtils.TruncateAt.MARQUEE)
     assertThat(topicToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
   }
 
   @Test
-  fun testTopicFragment_toolbarTitle_marqueeInLtr_isDisplayedCorrectly() {
+  fun testTopicFragment_toolbarTitle_marqueeInLtr_isDisplayedCorrectly_whenReaderOn() {
```
adhiamboperes:
```suggestion
testTopicFragment_toolbarTitle_readerOn_marqueeInLtr_isDisplayedCorrectly() {
```

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -324,18 +328,58 @@ class TopicFragmentTest {
...
       activityTestRule.activity.findViewById(R.id.topic_toolbar_title)
     ViewCompat.setLayoutDirection(topicToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
 
+    onView(withId(R.id.topic_toolbar_title)).perform(click())
+    assertThat(topicToolbarTitle.isFocused).isEqualTo(false)
+    assertThat(topicToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
+  }
+
+  @Test
+  fun testTopicFragment_toolbarTitle_marqueeInLtr_isDisplayedCorrectly_whenReaderOff() {
```
adhiamboperes:
```suggestion
testTopicFragment_toolbarTitle_readerOff_marqueeInLtr_isDisplayedCorrectly() {
```

adhiamboperes (changes requested):
Hi @XichengSpencer, really great job on this; your PR description is well detailed and good job on setting up the tests!

I have left some revision comments, and in addition:
- Do use the ktlint formatter comand to order your imports before committing.
- Please replicate these tests to the StoryFragmentTest and the corresponding test files of the modified activities.
- Reply (single comment) to the comments that you have addressed.
- Re-assign me  with my  @username PTAL.

XichengSpencer:
@adhiamboperes 
Hi Adhiambo, 
After I made changes to add the Accessibility Service to the constructor of different presenters, it seems to mess up many tests and gives errors like this
image
They all complain about somewhere in the tests have a different constructor setting than our modified constructors. Is there any suggestion for this error?

Thanks!

adhiamboperes:
> @adhiamboperes Hi Adhiambo, After I made changes to add the Accessibility Service to the constructor of different presenters, it seems to mess up many tests and gives errors like this image They all complain about somewhere in the tests have a different constructor setting than our modified constructors. Is there any suggestion for this error?
> 
> Thanks!

Hi @XichengSpencer,
What if you use field injection as opposed to constructor injection?

app/src/main/java/org/oppia/android/app/player/exploration/ExplorationActivityPresenter.kt
```diff
@@ -102,8 +105,12 @@ class ExplorationActivityPresenter @Inject constructor(
...
     activity.setSupportActionBar(explorationToolbar)
 
-    binding.explorationToolbarTitle.setOnClickListener {
-      binding.explorationToolbarTitle.isSelected = true
+    activity.applicationContext?.let {
+      if (!accessibilityService.isScreenReaderEnabled()) {
+        binding.explorationToolbarTitle.setOnClickListener {
+          binding.explorationToolbarTitle.isSelected = true
+        }
+      }
```
adhiamboperes: Please remove this usage of context

adhiamboperes (changes requested):
Thanks @XichengSpencer! I have looked at this and seems that there are still some outstanding changes to be made. Please refer to my last comment https://github.com/oppia/oppia-android/pull/5027#pullrequestreview-1511262094.

It is also the convention to allow the person that raised a comment to resolve it, so that they can verify that it is adequately fixed.

XichengSpencer:
> Thanks @XichengSpencer! I have looked at this and seems that there are still some outstanding changes to be made. Please refer to my last comment [#5027 (review)](https://github.com/oppia/oppia-android/pull/5027#pullrequestreview-1511262094).
> 
> It is also the convention to allow the person that raised a comment to resolve it, so that they can verify that it is adequately fixed.

The additional tests are added and succeed on my side, @adhiamboperes

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -357,6 +360,29 @@ class ExplorationActivityTest {
...
+      explorationActivityTestRule.activity.findViewById(R.id.exploration_toolbar_title)
+    ViewCompat.setLayoutDirection(explorationToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
+
+    onView(withId(R.id.exploration_toolbar_title)).perform(click())
+    assertThat(explorationToolbarTitle.isFocused).isEqualTo(false)
+    assertThat(explorationToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
+  }
+
+  @Test
+  fun testExploration_toolbarTitle_readerOff_marqueeInLtl_isDisplayedCorrectly() {
```
adhiamboperes: Nit: Ltr

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -366,6 +392,28 @@ class ExplorationActivityTest {
     assertThat(explorationToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
   }
 
+  @Test
+  fun testExploration_toolbarTitle_readerOn_marqueeInLtl_isDisplayedCorrectly() {
```
adhiamboperes: Nit: Ltr

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -227,11 +229,12 @@ class StoryFragmentTest {
     }
   }
 
-  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
   // its view
-  fun testStoryFragment_toolbarTitle_marqueeInRtl_isDisplayedCorrectly() {
+  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
```
adhiamboperes: Nit: some of the text in the comment seems misplaced

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -242,11 +245,44 @@ class StoryFragmentTest {
     assertThat(storyToolbarTitle.textAlignment).isEqualTo(TEXT_ALIGNMENT_VIEW_START)
   }
 
+  // its view
+  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
```
adhiamboperes: Nit: Please check the comment here as well.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -242,11 +245,44 @@ class StoryFragmentTest {
...
+    val storyToolbarTitle: TextView =
+      activityTestRule.activity.findViewById(R.id.story_toolbar_title)
+    ViewCompat.setLayoutDirection(storyToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
+
+    onView(withId(R.id.story_toolbar_title)).perform(click())
+    assertThat(storyToolbarTitle.isFocused).isEqualTo(false)
+    assertThat(storyToolbarTitle.textAlignment).isEqualTo(TEXT_ALIGNMENT_VIEW_START)
+  }
+
+  // its view
```
adhiamboperes: Nit: fix comment

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -242,11 +245,44 @@ class StoryFragmentTest {
...
+    activityTestRule.launchActivity(createFractionsStoryActivityIntent())
+    testCoroutineDispatchers.runCurrent()
+    accessibilityService.setScreenReaderEnabled(false)
+
+    val storyToolbarTitle: TextView =
+      activityTestRule.activity.findViewById(R.id.story_toolbar_title)
+    ViewCompat.setLayoutDirection(storyToolbarTitle, ViewCompat.LAYOUT_DIRECTION_LTR)
+
+    onView(withId(R.id.story_toolbar_title)).perform(click())
+    assertThat(storyToolbarTitle.isFocused).isEqualTo(false)
```
adhiamboperes: When the reader is off the toolbar should be clickable. Could you explain why this test passes while it is asserting the wrong behaviour, which is opposite what is expected?

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -242,11 +245,44 @@ class StoryFragmentTest {
...
+    assertThat(storyToolbarTitle.textAlignment).isEqualTo(TEXT_ALIGNMENT_VIEW_START)
+  }
+
   // its view
-  fun testStoryFragment_toolbarTitle_marqueeInLtr_isDisplayedCorrectly() {
+  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
```
adhiamboperes: Nit: fix comment

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
```diff
@@ -242,11 +245,44 @@ class StoryFragmentTest {
...
+  fun testStoryFragment_toolbarTitle_readerOn_marqueeInLtr_isDisplayedCorrectly() {
     activityTestRule.launchActivity(createFractionsStoryActivityIntent())
     testCoroutineDispatchers.runCurrent()
+    accessibilityService.setScreenReaderEnabled(false)
```
adhiamboperes: The test name implies that the reader should be on, not off.

adhiamboperes: It looks like you may have mistakenly switched this test and the one above, because they are both asserting behaviour opposite of their test names.

app/src/sharedTest/java/org/oppia/android/app/thirdparty/LicenseTextViewerActivityTest.kt
```diff
@@ -145,6 +156,86 @@ class LicenseTextViewerActivityTest {
...
+    ViewCompat.setLayoutDirection(activityToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
+
+    Espresso.onView(ViewMatchers.withId(R.id.license_text_viewer_activity_toolbar_title))
+      .perform(ViewActions.click())
+    assertThat(activityToolbarTitle.isFocused).isEqualTo(false)
+    assertThat(activityToolbarTitle.textAlignment).isEqualTo(View.TEXT_ALIGNMENT_VIEW_START)
+  }
+
+  @Test
+  fun testExploration_toolbarTitle_readerOff_marqueeInLtl_isDisplayedCorrectly() {
```
adhiamboperes: Nit: Ltr, here and the test below it.

adhiamboperes (approved):
Thanks for adding the tests, @XichengSpencer! 

I have left some comments for minor detail corrections, but there are two tests in StoryFragmentTest.kt that contain errors. Please fix these before merge.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
XichengSpencer: I am not sure what this comment means, but it is in the original test, so I thoughts it is for some future fixes and I carried it over. What should I do to this comment?

adhiamboperes: The original comment is

```
// TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
  // its view
```

While your version is

```
  // its view
  @Test // TODO(#4212): Error -> Only the original thread that created a view hierarchy can touch
```

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
XichengSpencer: There should be some improvement here, isfocused is inferring whether marquee effect is activated while 
assertThat(storyToolbarTitle.ellipsize).isEqualTo(TextUtils.TruncateAt.MARQUEE)
is determined if the marquee effect is set. (It is already set in xml, so always pass)
We should have both of them combined in all marquee tests

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
```diff
@@ -172,33 +191,56 @@ class RevisionCardActivityTest {
...
     ).use { scenario ->
       scenario.onActivity { activity ->
+        fakeAccessibilityService.setScreenReaderEnabled(false)
 
         val revisionCardToolbarTitle: TextView =
           activity.findViewById(R.id.revision_card_toolbar_title)
         ViewCompat.setLayoutDirection(revisionCardToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
 
-        onView(withId(R.id.revision_card_toolbar_title)).perform(ViewActions.click())
+        onView(withId(R.id.revision_card_toolbar_title)).perform(click())
```
XichengSpencer, line 207: I am not sure if this line clicks the right textview. During the test the onclick listener to activate the marquee effect is never called but in manual testing, it works as expected. Same question for all other tests when reader off.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
XichengSpencer: I figure out the reason, it is formatted by ktlint. Should I restore the change regardless of ktlint warning?

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
adhiamboperes: Okay, I understand. However, my point is majorly focussed on the behaviour being tested.
We established that `if accessibilityService.setScreenReaderEnabled(false)` then the correct behavior should be `assertThat(storyToolbarTitle.isFocused).isEqualTo(true)`. Is this correct? If this is the case, then the test behaviour described here in this test, is:
`if accessibilityService.setScreenReaderEnabled(false)`
`assertThat(storyToolbarTitle.isFocused).isEqualTo(false)`
Which is faulty. Do you agree?

XichengSpencer: Yes, and I post a comment regarding the problem when trying to fix this. Let me know if you can see it, I want to make sure the review is well published.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
XichengSpencer: @adhiamboperes

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
adhiamboperes: Yes, since it is broken. Alternativel, you can move it above the `@Test` line

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
adhiamboperes: I think that there is some confusion here.

The way I see it, most of the tests are correct. When we click the toolbar title, like here, marquee should be off. i.e, the behaviour should be `assertThat(revisionCardToolbarTitle.isFocused).isEqualTo(false)`. This should be the same behaviour as in manual tests.

If you find a test where the setup has not been done correctly, please feel free to update it.

app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt
adhiamboperes: I have seen it and replied.

app/src/sharedTest/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivityTest.kt
adhiamboperes: As an example, your tests in the [LicenseActivity](https://github.com/oppia/oppia-android/pull/5027/files#diff-b3b62a2ffea3ff795282ffd3b52ade34570f1ddc884b3912d71663049ba0165c) and the TopicFragment are correct imo, the rest should follow the same format.

XichengSpencer: Yes the test for `ReaderOn` cases is good, and error comes from the `ReaderOff`. I try to add a check for `title. is focused` for tests for `ReaderOff`, to make up the missing check for whether the marquee effect is activated. (This is the very original case) The tests failed with the title is not focused (Not activated). But we all know that the marquee effect works perfectly, that's why I am concerning if `perform.click()` has some problem.

adhiamboperes: As discussed in the call, let's update the checks from isFocused, to isSelected.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -64,11 +67,12 @@ class TopicFragmentPresenter @Inject constructor(
     binding.topicToolbar.setNavigationOnClickListener {
       (activity as TopicActivity).finish()
     }
-
-    binding.topicToolbar.setOnClickListener {
-      binding.topicToolbarTitle.isSelected = true
+    val screenReaderEnabled = accessibilityService.isScreenReaderEnabled()
```
adhiamboperes: I can't see any usages for this variable. Please let me know if I'm missing something.

app/src/sharedTest/java/org/oppia/android/app/thirdparty/LicenseTextViewerActivityTest.kt
```diff
@@ -145,6 +156,90 @@ class LicenseTextViewerActivityTest {
...
+    activityTestRule.launchActivity(
+      createLicenseTextViewerActivityIntent(
+        dependencyIndex = 0,
+        licenseIndex = 0
+      )
+    )
+
+    val activityToolbarTitle: TextView =
+      activityTestRule.activity.findViewById(R.id.license_text_viewer_activity_toolbar_title)
+    ViewCompat.setLayoutDirection(activityToolbarTitle, ViewCompat.LAYOUT_DIRECTION_RTL)
```
adhiamboperes: direction should be LTR per test name.

app/src/sharedTest/java/org/oppia/android/app/thirdparty/LicenseTextViewerActivityTest.kt
```diff
@@ -145,6 +156,90 @@ class LicenseTextViewerActivityTest {
...
```
adhiamboperes:
```suggestion
    ViewCompat.setLayoutDirection(activityToolbarTitle, ViewCompat.LAYOUT_DIRECTION_LTR)
```

adhiamboperes (changes requested):
Thanks @XichengSpencer! I only have a few minor corrections. See in-line.

I'd also like to point out that its good practice to avoid committing style only code changes in the future, as they can inflate the size of the PR. this happens when you format a file using Ctrl+Alt+L, but you can always be able to identify and revert unneeded formatting changes. Let me know if you need to learn how to do this in the next meeting.

XichengSpencer:
> Thanks @XichengSpencer! I only have a few minor corrections. See in-line.
> 
> I'd also like to point out that its good practice to avoid committing style only code changes in the future, as they can inflate the size of the PR. this happens when you format a file using Ctrl+Alt+L, but you can always be able to identify and revert unneeded formatting changes. Let me know if you need to learn how to do this in the next meeting.

Got it. I totally understand it will be great to address the code and style issues in the same commit. But I am not sure what needed formatting changes mean. Would you mind clarify a bit?

adhiamboperes (approved):

adhiamboperes:
This LGTM. Thanks @XichengSpencer

PR #5098, opened 2023-07-21
Fix #5001: Set Up Firestore and Upload Free Form Responses

### LINKED ISSUE #5001: Add Cloud Firestore to the App
## Add Firebase Firestore
This issue tracks milestone 7 of the NPS Survey.

- [x] Add bazel and gradle dependencies for Firestore 
- [x] #5130
- [x] Rig up the app to upload data to Firestore on our existing log upload pipeline.

### Writing to Firestore
Collections and Documents in Firestore are created implicitly on first write. For our use case, we are creating collections at the root level of our database to organize unique data sets. This means we are going to have one collection with the name nps_survey_open_feedback for storing answers to the open-ended survey responses, and should we want to store different data in the future, we will create a new collection for it.

We will add the following data to the document:

- survey_id: String. This is the id of the survey with which this free text answer is associated. 
- open_feedback_answer:  String. The value of the answer provided by the user.
- time_submitted: Long. We are storing a timestamp as a field in the documents to use for ordering in queries. We will also mark this field as a TTL field for automatic data deletion per the retention policy.

We are using the Firestore `add()` method as opposed to `set()` in order for Firestore to auto generate our document Ids, which need not be meaningful. This will also enforce uniqueness during write so that a user is unable to overwrite any other user’s document.

## Authenticate with Firebase anonymously
We are implementing anonymous auth in the background to allow users to be authenticated with Firebase Security Rules.

## Tasks
- [x] Configure Anonymous Auth
- [x] Configure Firestore upload
- [ ] Test security rules
- [x] Test offline auth, storage and upload

adhiamboperes:
## Explanation
Fixes #5001.
This is PR 5 of 6 Planned PRs.

We are adding a whole new infrastructure for uploading data to Firebase Firestore.

### Key Changes
- Introducing a `FirestoreDataController` that handles Anonymous Authentication and logging to Firestore, based on connectivity status.
- Adding the Firestore, Auth and AppCheck dependencies, and reppining these maven dependencies.
- New event logger for converting event logs to documents(hashmaps) and uploading them to firestore.
- Introduce AuthenticationController
This new controller provides functionality to retrieve the current signed in user and also sign in a user anonymously, and can be called from anywhere. It has a fake that is to be used in tests.
- Adding the newly created `utility` module files to `CODEOWNERS`
- Creating wrappers for FirebaseAuth, and Firestore in order to allow switching out with fakes in tests and development versions using dagger.

The implementation relies on an AuthenticationWrapper to switch out the real and fake authentication controllers for testing purposes. To do this, I created a dagger module to provide the correct listener where needed, hence the test file changes in the _app_ module tests.

### Technical Decisions
- I opted to create a new cache for events that will eventually be uploaded to Firestore. This means they will not be accidentally uploaded to Firebase.
- I opted to use EventLogs and convert them to Documents on the fly as opposed to creating an entirely new model and logging infrastruture that is Firestore-specific.
- I calculated the `FirestoreLogStorageCacheSize` based on the [average length of a Google Review ](https://www.meetsoci.com/resources/insights/brands-franchises/state-of-google-reviews/#:~:text=Average%20review%20length%20ranges%20from,per%20review%20for%20department%20stores.&text=In%20comparison%20to%20other%20local%20platforms%2C%20Google%20reviews%20skew%20positively.)and the [storage size computation specs](https://firebase.google.com/docs/firestore/storage-size) defined by Firebase. I computed it based on 10 profiles on a device over a 3 month period.
- I opted to use existing log upload infrastructure to upload to firebase, as opposed to creating a whole new upload workflow so I can reuse the existing test infrastructure as well.

### Dependency Updates
- I added dependencies for Firestore and Auth. The selected versions were both the earliest compatible with the rest of our existing dependencies and the latest that didn't cause proguarding issues. There were incidental updates to transitive dependencies, both major and minor.

### Proguard
Added new -dontwarns for the following:
`-dontwarn javax.naming.**`  [Reference](https://github.com/firebase/firebase-android-sdk/blob/00d4626/firebase-firestore/proguard.txt.)

## Essential Checklist

### Firestore Upload
Screenshot 2024-01-23 at 11 18 26

## Querying Synced Firestore Data on BigQuery
Screenshot 2024-01-23 at 11 26 26

Screenshot 2024-01-23 at 11 26 44


app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
```diff
@@ -47,6 +50,8 @@ abstract class AbstractOppiaApplication(
     // TODO(#4751): Re-enable WorkManager for S+.
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
       FirebaseApp.initializeApp(applicationContext)
+      Firebase.firestore
```
seanlip: I don't know enough Kotlin, but I'm a bit surprised that these lines don't seem to be calling anything (shouldn't they have a "()" after them if they are functions)?

If they aren't functions then what do these lines do?

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
```diff
@@ -14,11 +15,14 @@ import javax.inject.Inject
 @FragmentScope
 class ViewEventLogsViewModel @Inject constructor(
   debugAnalyticsEventLogger: DebugAnalyticsEventLogger,
+  debugFirestoreEventLogger: DebugFirestoreEventLogger,
   private val machineLocale: OppiaLocale.MachineLocale,
   private val resourceHandler: AppLanguageResourceHandler
 ) : ObservableViewModel() {
 
-  private val eventList = debugAnalyticsEventLogger.getEventList()
+  private val firebaseEvents = debugAnalyticsEventLogger.getEventList()
```
seanlip: What's the difference between a "firebase event" and a "firestore event"?

Is this explained anywhere, in the code/wiki?

seanlip (commented):
Hi @adhiamboperes, thanks for the PR. I don't really have much feedback! Just a couple of comments from looking at the files.

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
adhiamboperes: In Kotlin getters use direct access syntax like this, so this returns the FirebaseFirestore instance of the default FirebaseApp.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
adhiamboperes: Firestore events are eventlogs that are intended to be uploaded to Firestore, while Firebase events are event logs that are intended for upload to Firebase. 

> Is this explained anywhere, in the code/wiki?

More of implied via naming and KDocs? Since I have used this naming system in multiple files in this PR, I am not sure which would be a centralized place to add a general explanation.

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
adhiamboperes: I suppose this means that I can call them directly at usage site instead of globally in this init block.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
seanlip: But then should we rename "analytics event logger" here to "firebase event logger"?

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
seanlip: But where is the returned value being stored? I don't see any "val ... = ..." statement.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
adhiamboperes: The prod implementation of the AnalyticsEventLogger interface is named FirebaseAnalyticsEventLogger already. The debug version that is used here does not make that distinction, because the debug implementation is generic.
I have renamed this "firebaseEvents" variable to analyticsEvents to reflect that, and kept the "firestoreEvents" variable name because it is descriptive of its use. I have also added comments.

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
adhiamboperes: I have moved these getters to their call site where they are properly assigned to variables.

adhiamboperes:
I have 2 sets of failing tests in this PR, which on investigation seem to be caused by dependency version problems with Firestore. I think the best way to do this is to generate a PR that is exclusively for adding the new dependencies so @kkmurerwa and I can debug the proguard failures. I suspect that we mostly just need to lower the version to an older one, but I have no way of determining which one will be compatible with our dependencies so it will be more of a try-build agains different versions.

config/proguard/firebase-components-proguard-rules.pro
```diff
@@ -1,6 +1,8 @@
 # Reference: https://github.com/firebase/firebase-android-sdk/blob/82b02af331/firebase-components/proguard.txt.
+# Reference: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/proguard.txt.
```
seanlip: Just a note -- what is on "master" could change. You might want to pin a specific commit as was done in the previous line.

seanlip (approved):
Looks good. Just one comment, but no blockers from me. Thanks @adhiamboperes!

config/proguard/firebase-components-proguard-rules.pro
adhiamboperes: Thanks for catching this @seanlip, I have updated the url

adhiamboperes:
> I have 2 sets of failing tests in this PR, which on investigation seem to be caused by dependency version problems with Firestore. I think the best way to do this is to generate a PR that is exclusively for adding the new dependencies so @kkmurerwa and I can debug the proguard failures. I suspect that we mostly just need to lower the version to an older one, but I have no way of determining which one will be compatible with our dependencies so it will be more of a try-build agains different versions.

I was able to find and appropriately downgrade the Firebase Auth and Firestore dependencies. Proguard works fine now.

Gdakgdak22 (approved):

Gdakgdak22 (approved):

app/src/sharedTest/java/org/oppia/android/app/options/OptionsActivityTest.kt
```diff
@@ -86,7 +87,6 @@ import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
 import javax.inject.Inject
 import javax.inject.Singleton
-
```
BenHenning: Please re-add missing newline used for vertical spacing (consistency).

domain/build.gradle
```diff
@@ -93,9 +93,11 @@ dependencies {
       'com.google.dagger:dagger:2.24',
       'com.google.firebase:firebase-analytics-ktx:17.5.0',
       'com.google.firebase:firebase-crashlytics:17.0.0',
+      'com.google.firebase:firebase-auth-ktx:19.3.1',
       'com.google.guava:guava:28.1-android',
       'com.google.protobuf:protobuf-javalite:3.17.3',
-      "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+      "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version",
+      'org.mockito:mockito-core:2.19.0',
```
BenHenning: Production builds should never need to include Mockito (and in fact such cases often indicate a dependency that should go in testing/).

scripts/assets/maven_dependencies.textproto
```diff
@@ -830,6 +928,16 @@ maven_dependency {
     }
   }
 }
+maven_dependency {
+  artifact_name: "org.codehaus.mojo:animal-sniffer-annotations:1.18"
+  artifact_version: "1.18"
+  license {
+    license_name: "The Apache Software License, Version 2.0"
```
BenHenning: Do you know why this is saying Apache? Per https://mvnrepository.com/artifact/org.codehaus.mojo/animal-sniffer-annotations/1.18 the dependency seems to be using MIT. I didn't check any of the other dependencies, but can you please make sure that all of their licenses are the correct ones per their GitHub repositories (where available) & Maven artifact listing?

I wonder if a fix is needed in the license extractor.

domain/src/test/java/org/oppia/android/domain/auth/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
+"""
+Library for providing authentication with Firebase functionality in the app.
```
BenHenning:
```suggestion
Tests for libraries providing authentication with Firebase functionality.
```

utility/src/main/java/org/oppia/android/util/logging/firebase/BUILD.bazel
```diff
@@ -82,8 +83,49 @@ kt_android_library(
     deps = [
         ":dagger",
         ":debug_event_logger",
+        ":debug_firestore_logger",
         ":firebase_exception_logger",
     ],
 )
 
+kt_android_library(
+    name = "firestore_logger_impl",
```
BenHenning, line 93: Does this actually need to be accessed outside of this package? Normally implementations aren't directly accessed (and instead are indirectly depended on via modules).

app/src/main/java/org/oppia/android/app/survey/SurveyOutroDialogFragmentPresenter.kt
```diff
@@ -60,7 +60,7 @@ class SurveyOutroDialogFragmentPresenter @Inject constructor(
         when (it) {
           is AsyncResult.Pending -> oppiaLogger.d("SurveyActivity", "Stopping survey session")
           is AsyncResult.Failure -> {
-            oppiaLogger.d("SurveyActivity", "Failed to stop the survey session")
+            oppiaLogger.d("SurveyActivity", "Failed to stop the survey session, ${it.error}")
```
BenHenning: You could pass the throwable, instead. That will ensure the stack trace is printed, as well as any chained cause throwables (the stringified version of a throwable isn't always a reliable summary of the problem).

config/proguard/firebase-components-proguard-rules.pro
```diff
@@ -1,6 +1,9 @@
 # Reference: https://github.com/firebase/firebase-android-sdk/blob/82b02af331/firebase-components/proguard.txt.
+# Reference: https://github.com/firebase/firebase-android-sdk/blob/00d4626/firebase-firestore/proguard.txt.
 
 -dontwarn com.google.firebase.components.Component$Instantiation
 -dontwarn com.google.firebase.components.Component$ComponentType
+-dontwarn com.google.firebase.firestore.remote.FirestoreChannel
```
BenHenning: I don't see FirestoreChannel listed in the new proguard link added above. What's the actual error being run into? It may be the case of incompatible dependencies, not a warning that can safely be ignored.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -645,6 +645,9 @@ exempted_file_path: "data/src/main/java/org/oppia/android/data/backends/gae/mode
 exempted_file_path: "data/src/main/java/org/oppia/android/data/backends/gae/model/GaeVoiceover.kt"
 exempted_file_path: "data/src/main/java/org/oppia/android/data/backends/gae/model/GaeWrittenTranslation.kt"
 exempted_file_path: "data/src/main/java/org/oppia/android/data/backends/gae/model/GaeWrittenTranslations.kt"
+exempted_file_path: "domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt"
```
BenHenning: This class has logic in it--it seems useful to test to help catch issues earlier in the pipeline.

domain/src/main/java/org/oppia/android/domain/auth/FakeAuthenticationControllerFactory.kt
```diff
@@ -0,0 +1,13 @@
+package org.oppia.android.domain.auth
+
+import com.google.firebase.auth.FirebaseAuth
+import org.mockito.Mockito.mock
+import javax.inject.Inject
+
+/** A test specific fake for the AuthenticationController.Factory. */
+class FakeAuthenticationControllerFactory @Inject constructor() {
```
BenHenning: Should this be implementing AuthenticationController.Factory?

Also, why do we need this? If it's not implementing the ``Factory`` implementation, is the main benefit just binding the mock to be reusable? The class doesn't seem to be providing any specific fake functionality.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -751,6 +754,7 @@ exempted_file_path: "testing/src/main/java/org/oppia/android/testing/OppiaTestAn
 exempted_file_path: "testing/src/main/java/org/oppia/android/testing/OppiaTestRule.kt"
 exempted_file_path: "testing/src/main/java/org/oppia/android/testing/OppiaTestRunner.kt"
 exempted_file_path: "testing/src/main/java/org/oppia/android/testing/RichTextViewMatcher.kt"
+exempted_file_path: "testing/src/main/java/org/oppia/android/testing/TestAuthenticationModule.kt"
```
BenHenning: Modules are usually worth testing to ensure the bindings stay intentional as the modules change over time.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -842,11 +846,14 @@ exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/Loggin
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/SyncStatusManager.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/SyncStatusModule.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/DebugAnalyticsEventLogger.kt"
+exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/DebugFirestoreEventLogger.kt"
```
BenHenning: Test doubles are always worth testing as failures they introduce can make tests that depend on them more difficult to test/verify. Debug implementations that can also show up in developer builds should be assumed as 'production-ready', so they also ought to have tests.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -842,11 +846,14 @@ exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/Loggin
...
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/DebugLogReportingModule.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseAnalyticsEventLogger.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseExceptionLogger.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseLogUploader.kt"
 exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirebaseLogUploaderModule.kt"
+exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLogger.kt"
+exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLoggerProdImpl.kt"
```
BenHenning, line 866: Can tests be added for this implementation, too?

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationListener.kt
```diff
@@ -0,0 +1,14 @@
+package org.oppia.android.domain.auth
+
+import com.google.firebase.auth.FirebaseUser
+import kotlinx.coroutines.CompletableDeferred
+import org.oppia.android.util.data.AsyncResult
+
+/** Listener for getting the authentication state of the default FirebaseApp. */
+interface AuthenticationListener {
```
BenHenning: Is this actually a listener? It seems to be providing functionality that must be called by users of implementations of this listener rather than being a conduit for information from elsewhere in the app (e.g. onUserSignedIn(), onSignedInAnonymously(), etc.). The dataflow directionality here seems reversed from what I'd normally expect of a listener.

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLogger.kt
```diff
@@ -0,0 +1,16 @@
...
+interface FirestoreEventLogger {
+  /**
+   * Converts eventLogs to Firestore documents and uploads or save them on disk.
+   *
+   * @param eventLog which contains all the relevant data to be reported.
+   */
+  fun uploadEvent(eventLog: EventLog)
+
+  /** Returns the list of all [EventLog]s logged since the app opened. */
+  fun getEventList(): List<EventLog>
```
BenHenning: This doesn't seem like something we'd want to expose generally. Could we maybe have a separate interface, e.g. DebugFirestoreEventLogger that exposes this? We can bind the same implementation instances to both listener types in Dagger (such that the debug-only module provides an implementation for DebugFirestoreEventLogger that can be used by the debug-only view models).

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlinx.coroutines.CompletableDeferred
+import org.mockito.Mockito.mock
+import org.oppia.android.domain.auth.AuthenticationListener
+import org.oppia.android.util.data.AsyncResult
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** A test specific fake for the AuthenticationController. */
+@Singleton
+class FakeAuthenticationController @Inject constructor() : AuthenticationListener {
```
BenHenning: Is this a fake AuthenticationController or AuthenticationListener? Regardless, it's generally best practice to fake at the highest level below 1p code (e.g. the Firebase dependencies, in this case). It's also best practice to use existing fakes where available rather than reimplementing them. However, if there aren't fakes for an external library, we can wrap that entire dependency and provide fakes for the thin layer that sits on top of the dependency, keeping all downstream code unfaked.

BenHenning (changes requested):
Thanks @adhiamboperes! I took a high-level pass on architecture and boilerplate. I didn't dig into the details on implementation or tests yet.

I think overall the approach looks good, but I'm hoping for some clarification in comment, code, or both for a few different parts of the approach. PTAL.

domain/src/test/java/org/oppia/android/domain/auth/BUILD.bazel
adhiamboperes: Updated

app/src/sharedTest/java/org/oppia/android/app/options/OptionsActivityTest.kt
adhiamboperes: Done

scripts/assets/maven_dependencies.textproto
adhiamboperes: It could have been an error on my part. I deleted it and re-ran the script, found that it required manual update. 
I checked the other licenses and they are all correct, so I don't have a reason to think that there is a bug in the extractor.

config/proguard/firebase-components-proguard-rules.pro
adhiamboperes: The warning is:
```
Warning: com.google.firebase.firestore.remote.FirestoreChannel$1: can't find enclosing method 'void lambda$runBidiStreamingRpc$0$FirestoreChannel(com.google.firebase.firestore.remote.FirestoreChannel,io.grpc.ClientCall[],com.google.firebase.firestore.remote.IncomingStreamObserver,com.google.android.gms.tasks.Task)' in program class com.google.firebase.firestore.remote.FirestoreChannel
Warning: com.google.firebase.firestore.remote.FirestoreChannel$3: can't find enclosing method 'void lambda$runStreamingResponseRpc$1$FirestoreChannel(com.google.firebase.firestore.remote.FirestoreChannel,com.google.android.gms.tasks.TaskCompletionSource,java.lang.Object,com.google.android.gms.tasks.Task)' in program class com.google.firebase.firestore.remote.FirestoreChannel
Warning: com.google.firebase.firestore.remote.FirestoreChannel$4: can't find enclosing method 'void lambda$runRpc$2$FirestoreChannel(com.google.firebase.firestore.remote.FirestoreChannel,com.google.android.gms.tasks.TaskCompletionSource,java.lang.Object,com.google.android.gms.tasks.Task)' in program class com.google.firebase.firestore.remote.FirestoreChannel
```
I tried some fixes to resolve this but none of them work so far:
- Keeping firestore classes in proguard
- Declaring the `com.google.android.gms.tasks.TaskCompletionSource` dependency
- I couldn't find something related to `java.lang.Object`,

domain/build.gradle
adhiamboperes: I removed the dependency. It is already included in /testing

utility/src/main/java/org/oppia/android/util/logging/firebase/BUILD.bazel
adhiamboperes: It can be depended on via this package's `prod_module`, so I have updated the visibility.

app/src/main/java/org/oppia/android/app/survey/SurveyOutroDialogFragmentPresenter.kt
adhiamboperes: I'm not sure I understand - the [error](https://github.com/oppia/oppia-android/blob/08981272a8136b8da0be7c4197480790c36bf100/utility/src/main/java/org/oppia/android/util/data/AsyncResult.kt#L195) in AsyncResult.Failure is a throwable.

scripts/assets/test_file_exemptions.textproto
adhiamboperes: Added a test file and removed this exemption

domain/src/main/java/org/oppia/android/domain/auth/FakeAuthenticationControllerFactory.kt
adhiamboperes: I removed this class since I refactored AuthenticationController, now it does not use the factory pattern.

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
adhiamboperes: It is a fake AuthenticationController, since its the test fake implementation of AuthenticationListener/Wrapper. I removed the mock for FirebaseUser from this class. I hope that this accurately answers this comment.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationListener.kt
adhiamboperes: It is more of a wrapper, and I have renamed it as such.

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLogger.kt
adhiamboperes: Done. I created a `DebugFirestoreEventLogger` with these two functions, and changed `FirestoreEventLogger` to have only the `uploadEvent` function.

scripts/assets/test_file_exemptions.textproto
adhiamboperes: Since I created an interface DebugFirestoreEventLogger, I will keep this here. Its implementation, DebugFirestoreEventLoggerImpl, has tests added.

adhiamboperes:
@BenHenning, could you PTAL at this? I have addressed the comments.

scripts/assets/test_file_exemptions.textproto
adhiamboperes: Since `AuthenticationController` is just a thin wrapper, I would fail  to test it directly because I don't have any business logic to test here. I am just mapping from task to completable future. And so not testing this class is a valid conclusion. I mean in terms of code or behaviour coverage, testing this class impacts my confidence very lightly. What I lean towards, is testing it as part of an integration test, in FirestoreDataController, where there is usage of it.

config/proguard/firebase-components-proguard-rules.pro
adhiamboperes: This is now fixed

domain/build.gradle
BenHenning: It still seems present in the latest changes unless I'm missing something.

utility/build.gradle
```diff
@@ -89,6 +89,8 @@ dependencies {
       'com.google.firebase:firebase-analytics-ktx:17.5.0',
       'com.google.firebase:firebase-core:17.5.0',
       'com.google.firebase:firebase-crashlytics:17.0.0',
+      'com.google.firebase:firebase-firestore-ktx:21.5.0',
```
BenHenning: Is this intentionally a different version than the one used in Bazel? Prefer keeping all new dependencies aligned in versions to avoid any potential skew between builds (where possible).

third_party/maven_install.json
```diff
@@ -1,16 +1,20 @@
...
-        "__INPUT_ARTIFACTS_HASH": 521507787,
-        "__RESOLVED_ARTIFACTS_HASH": -1467267841,
+        "__INPUT_ARTIFACTS_HASH": -730560459,
+        "__RESOLVED_ARTIFACTS_HASH": 1784154617,
         "conflict_resolution": {
             "androidx.constraintlayout:constraintlayout:1.1.3": "androidx.constraintlayout:constraintlayout:2.0.1",
             "androidx.core:core:1.0.1": "androidx.core:core:1.3.1",
             "androidx.recyclerview:recyclerview:1.0.0": "androidx.recyclerview:recyclerview:1.1.0",
             "androidx.test:core:1.0.0": "androidx.test:core:1.4.0",
+            "com.google.auto.value:auto-value-annotations:1.7.5": "com.google.auto.value:auto-value-annotations:1.8.1",
```
BenHenning: For this and the checker-qual dependency, since these are just being resolved to higher dependencies anyway, can we go ahead and depend on those exact versions in versions.bzl?

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationWrapper.kt
```diff
@@ -0,0 +1,14 @@
+package org.oppia.android.domain.auth
+
+import com.google.firebase.auth.FirebaseUser
+import kotlinx.coroutines.CompletableDeferred
+import org.oppia.android.util.data.AsyncResult
+
+/** Wrapper for providing authentication functionality. */
+interface AuthenticationWrapper {
```
BenHenning: Suggest this being ``AuthenticationController`` and calling the other one ``AuthenticationControllerImpl`` or ``FirebaseAuthenticationControllerImpl``.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationWrapper.kt
```diff
@@ -0,0 +1,14 @@
...
+  /** Returns the current signed in user or null if there is no authenticated user. */
+  fun getCurrentSignedInUser(): FirebaseUser?
```
BenHenning: It would be ideal to abstract this so that we're not depending on Firebase types directly in the domain layer (in case we need to add onto or change the authentication system).

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.oppia.android.util.data.AsyncResult
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Controller for signing in and retrieving a Firebase user. */
+@Singleton
+class AuthenticationController @Inject constructor(
+  private val firebaseAuth: FirebaseAuth
+) : AuthenticationWrapper {
+  /** Returns the current signed in user or null if there is no authenticated user. */
```
BenHenning, line 18: Here & next method: no-need for KDocs on overrides.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseUser
+import kotlinx.coroutines.CompletableDeferred
+import org.oppia.android.util.data.AsyncResult
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Controller for signing in and retrieving a Firebase user. */
+@Singleton
```
BenHenning: This class doesn't have state, so it doesn't need to be scoped.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,33 @@
...
+  private val firebaseAuth: FirebaseAuth
+) : AuthenticationWrapper {
+  /** Returns the current signed in user or null if there is no authenticated user. */
+  override fun getCurrentSignedInUser(): FirebaseUser? {
+    return firebaseAuth.currentUser
+  }
+
+  /** Returns the result of an authentication task. */
+  override fun signInAnonymously(): CompletableDeferred<AsyncResult<Any?>> {
+    val deferredResult = CompletableDeferred<AsyncResult<Any?>>()
```
BenHenning, line 23: Is there a timeout functionality possible here? Concerned about cases where something breaks add neither onSuccess/onFailure callbacks are called.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,33 @@
...
+    firebaseAuth.signInAnonymously()
```
BenHenning: Should we have a check on whether the user is already signed in?

scripts/assets/test_file_exemptions.textproto
BenHenning: I think of it a bit differently. The parent class is a controller independent of Firebase that has its own behaviors that must be fulfilled. This implementation needs to fulfill that.

A wrapper is an implementation-specific thing that just forwards calls and can be swapped out for a test-only fake.

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
```diff
@@ -0,0 +1,42 @@
...
+
+  /** Sets whether sign in was successful. */
+  fun setSignInSuccessStatus(signInSuccessful: Boolean) {
+    signInIsSuccessful = signInSuccessful
+  }
+
+  /** Sets the current signed in user. */
+  fun setSignedInUser(firebaseUser: FirebaseUser) {
+    currentUser = firebaseUser
+  }
```
BenHenning: Could this maybe be consolidated to a sealed class state? Would ensure that the values always make sense together (e.g. "signInSuccessful = false && currentUser != null" probably doesn't make much sense).

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.google.firebase.auth.FirebaseUser
+import kotlinx.coroutines.CompletableDeferred
+import org.oppia.android.domain.auth.AuthenticationWrapper
+import org.oppia.android.util.data.AsyncResult
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** A test specific fake for the AuthenticationController. */
+@Singleton
+class FakeAuthenticationController @Inject constructor() : AuthenticationWrapper {
```
BenHenning: I suspect per my other comments that we can use the real implementation and separately wrap exactly the Firebase calls.

If an interface isn't needed for the controller, we can have a single AuthenticationController class that uses a FirebaseAuthWrapper interface (with corresponding FirebaseAuthWrapperImpl and FakeFirebaseAuthWrapper classes).

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
```diff
@@ -0,0 +1,169 @@
...
+   * error will be thrown if something went wrong during upload.
+   */
+  suspend fun uploadData() {
+    firestoreEventsStore.readDataAsync().await().eventLogsToUploadList.forEach { eventLog ->
+      authenticateAndUploadToFirestore(eventLog)
+    }
+  }
+
+  /**
+   * Logs a high priority event defined by [eventContext] corresponding to time [timestamp].
```
BenHenning: 'High priority' may not help here since this is the only choice of methods for logging an event.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
```diff
@@ -0,0 +1,169 @@
...
+   * priority.
+   */
+  private fun getLeastRecentEventIndex(oppiaEventLogs: OppiaEventLogs): Int? =
+    oppiaEventLogs.eventLogsToUploadList.withIndex().minByOrNull { it.value.timestamp }?.index
+
+  /** Returns a data provider for log reports that have been recorded for upload. */
+  fun getEventLogStore(): DataProvider<OppiaEventLogs> = firestoreEventsStore
+
+  /** Removes the first log report that had been recorded for upload. */
+  fun removeFirstEventLogFromStore() {
```
BenHenning: Why is this public? It doesn't seem that this is something we'd want random app module classes to call (which is fair game since this is being established as part of the controller's public API).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataControllerTest.kt
```diff
@@ -0,0 +1,465 @@
...
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testController_logEvent_withOptionalSurveyQuestionContext_checkLogsEvent() {
+    logOptionalSurveyResponseEvent()
+
+    val eventLog = fakeFirestoreEventLogger.getMostRecentEvent()
+
+    EventLogSubject.assertThat(eventLog).hasTimestampThat().isEqualTo(TEST_TIMESTAMP)
```
BenHenning: Prefer directly importing assertThat.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
```diff
@@ -14,11 +15,17 @@ import javax.inject.Inject
...
   private val resourceHandler: AppLanguageResourceHandler
 ) : ObservableViewModel() {
+  // Retrieves events from cache that are meant to be uploaded to Firebase Firestore.
+  private val firestoreEvents = debugFirestoreEventLogger.getEventList()
 
-  private val eventList = debugAnalyticsEventLogger.getEventList()
+  // Retrieves events from cache that are meant to be uploaded to Firebase Analytics.
+  private val analyticsEvents = debugAnalyticsEventLogger.getEventList()
+
+  private val eventList = analyticsEvents + firestoreEvents
```
BenHenning: Should these maybe be sorted by timestamp so that the two sets of events aren't concatenated?

domain/src/test/java/org/oppia/android/domain/oppialogger/loguploader/LogUploadWorkerTest.kt
```diff
@@ -402,6 +413,95 @@ class LogUploadWorkerTest {
...
+    val request: OneTimeWorkRequest = OneTimeWorkRequestBuilder<LogUploadWorker>()
+      .setInputData(inputData)
+      .build()
+
+    setUpFirestoreEventLoggerToFail()
+    workManager.enqueue(request)
+    testCoroutineDispatchers.runCurrent()
+    val workInfo = workManager.getWorkInfoById(request.id)
+
+    // assertThat(workInfo.get().state).isEqualTo(WorkInfo.State.FAILED)
```
BenHenning: Please remove any commented out code included in the PR.

adhiamboperes:
@BenHenning, PTAL.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
```diff
@@ -0,0 +1,171 @@
...
+    println("removing first event log from store")
```
BenHenning: Please make sure to remove debug lines the PR.

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
```diff
@@ -0,0 +1,3 @@
+package org.oppia.android.domain.auth
+
+class FirebaseAuthWrapperImplTest
```
BenHenning: What's this file for?

utility/src/main/java/org/oppia/android/util/logging/firebase/DebugFirestoreInstanceWrapperImpl.kt
```diff
@@ -0,0 +1,17 @@
+package org.oppia.android.util.logging.firebase
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * A test and debug fake for the [FirestoreInstanceWrapper]. This is also used in debug environments
+ * to make the debug implementation testable. [FirebaseFirestore] requires an instance of []FirebaseApp],
```
BenHenning:
```suggestion
 * to make the debug implementation testable. [FirebaseFirestore] requires an instance of [FirebaseApp],
```

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLogger.kt
```diff
@@ -0,0 +1,13 @@
+package org.oppia.android.util.logging.firebase
+
+import org.oppia.android.app.model.EventLog
+
+/** Logger for uploading data bundles to Firestore. */
+interface FirestoreEventLogger {
+  /**
+   * Converts eventLogs to Firestore documents and uploads or save them on disk.
+   *
+   * @param eventLog which contains all the relevant data to be reported.
```
BenHenning:
```suggestion
   * @param eventLog which contains all the relevant data to be reported
```

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLoggerProdImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+      "survey_id" to eventContext.surveyDetails.surveyId,
+      "open_feedback_answer" to eventContext.feedbackAnswer,
+      "time_submitted" to eventLog.timestamp
+    )
+
+    firestoreInstanceWrapper.firestoreInstance?.firebaseFirestore
+      ?.collection("nps_survey_open_feedback")
+      ?.add(document)
+      ?.addOnSuccessListener {
+        println("upload successful")
```
BenHenning: println here to remove as well.

utility/src/test/java/org/oppia/android/util/logging/firebase/DebugFirestoreEventLoggerImplTest.kt
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @Test
+  fun testDebugFirestoreEventLogger_logMultipleEvents_clearAllEvents_returnsEmptyList() {
+    eventLogger.uploadEvent(eventLog1)
+    eventLogger.uploadEvent(eventLog2)
+    debugFirestoreLoggerImpl.clearAllEvents()
+    val isListEmpty = debugFirestoreLoggerImpl.getEventList().isEmpty()
+
+    Truth.assertThat(isListEmpty).isTrue()
```
BenHenning: assertThat should be directly imported.

app/src/main/java/org/oppia/android/app/survey/SurveyOutroDialogFragmentPresenter.kt
BenHenning: ``oppiaLogger.d()`` has a parameter for passing the throwable so that its stack trace is logged, too.

BenHenning (changes requested):
Thanks @adhiamboperes! I originally was going to LGTM this right away, but given that there's 24 comments I think it'd nice to take one more pass (please let me know if you have any concerns with this).

utility/src/test/java/org/oppia/android/util/logging/firebase/DebugFirestoreEventLoggerImplTest.kt
adhiamboperes: Fixed this.

app/src/main/java/org/oppia/android/app/survey/SurveyOutroDialogFragmentPresenter.kt
adhiamboperes: Got it, I have modified this to use the overloaded function with the throwable.

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLoggerProdImpl.kt
adhiamboperes: Done

utility/src/main/java/org/oppia/android/util/logging/firebase/FirestoreEventLogger.kt
adhiamboperes: Fixed.

utility/src/main/java/org/oppia/android/util/logging/firebase/DebugFirestoreInstanceWrapperImpl.kt
adhiamboperes: Corrected

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
adhiamboperes: Removed🤦‍♀️

domain/src/test/java/org/oppia/android/domain/oppialogger/loguploader/LogUploadWorkerTest.kt
adhiamboperes: Uncommented this assertion. I previously missed this.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
adhiamboperes: I concatenated these because they are 2 seperate lists from different loggers. But I have made a change such that I add the lists to `eventList` then sort by timestamp.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataControllerTest.kt
adhiamboperes: Updated.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
adhiamboperes: Modified to ` Logs an event defined by [eventContext] corresponding to time [timestamp].`

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
adhiamboperes: I originally had this as the function was called in loguploadworker to remove logs after they have been uploaded. I have made a change so that it is called after each log has been uploaded within this controller. WDYT about this?

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
adhiamboperes: This is now implemented

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: I delegated this to the calling class, in this case the `FirestoreDataController`.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: Removed the Singleton annotation

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: This is outdated due to implementation change.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationWrapper.kt
adhiamboperes: Implemented a wrapper for FirebaseUser

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationWrapper.kt
adhiamboperes: Not sure if this is still valid per latest changes?

adhiamboperes: Not sure if this is still valid per latest chnages?

third_party/maven_install.json
adhiamboperes: Done

domain/build.gradle
adhiamboperes: Removed in latest changes

utility/build.gradle
adhiamboperes: Updated the gradle versions to match the bazel ones.

testing/src/main/java/org/oppia/android/testing/FakeAuthenticationController.kt
adhiamboperes: Done

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: One example of a scenario where on success/failure might not be called is a connection failure because firebase does not throw an exception in this case(it will keep retrying in the background). To mitigate against this, this method is never called unless a connection is established in `FirestoreDataController`

adhiamboperes:
@BenHenning, PTAL.

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
adhiamboperes: Tests for the prod impl of FirebaseAuthWrapper

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: This is now wrapped in a withTimeout of 3 seconds

scripts/assets/test_file_exemptions.textproto
adhiamboperes: As per the discussion, we will leave in this eemption for now.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
```diff
@@ -41,13 +41,19 @@ class FirestoreDataController @Inject constructor(
...
   suspend fun uploadData() {
-    firestoreEventsStore.readDataAsync().await().eventLogsToUploadList.forEach { eventLog ->
-      authenticateAndUploadToFirestore(eventLog)
+    val eventLogsToUpload = firestoreEventsStore.readDataAsync().await().eventLogsToUploadList
+
+    if (eventLogsToUpload.isNotEmpty()) {
+      eventLogsToUpload.forEach { eventLog ->
+        authenticateAndUploadToFirestore(eventLog)
+      }.also {
+        removeFirstEventLogFromStore()
```
BenHenning: Won't this only remove one event after all have been iterated?

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
```diff
@@ -1,3 +1,119 @@
...
+
+  @Test
+  fun testAuthWrapperImpl_getCurrentSignedInUser_userIsNotSignedIn_returnsNull() {
+    firebaseAuthWrapper.signInAnonymously(
+      onSuccess = {},
+      onFailure = {}
+    )
+    val user = firebaseAuthWrapperImpl.currentUser
+
+    Truth.assertThat(user).isNull()
```
BenHenning: assertThat should be directly imported.

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
```diff
@@ -1,3 +1,119 @@
...
+  fun setUp() {
+    setUpTestApplicationComponent()
+  }
+
+  @Test
+  fun testAuthWrapperImpl_getCurrentSignedInUser_userIsNotSignedIn_returnsNull() {
+    firebaseAuthWrapper.signInAnonymously(
+      onSuccess = {},
+      onFailure = {}
+    )
```
BenHenning: This is a bit confusing with respect to the test name--it seems like the test _is_ attempting to log in.

testing/src/main/java/org/oppia/android/testing/FakeFirebaseAuthWrapperImpl.kt
```diff
@@ -9,29 +9,41 @@ import javax.inject.Singleton
...
-    if (simulateSuccess) {
+    if (fakeAuthState == FakeAuthState.Success) {
       onSuccess.invoke()
     } else {
       onFailure.invoke(Exception("Sign-in failure"))
     }
   }
+
+  /** Represents a faked authentication state. */
+  sealed class FakeAuthState {
```
BenHenning: Probably could just be an enum since the classes aren't providing unique logic.

testing/src/test/java/org/oppia/android/testing/FakeFirebaseAuthWrapperImplTest.kt
```diff
@@ -64,6 +56,21 @@ class FakeFirebaseAuthWrapperImplTest {
...
+  @Test
+  fun testFakeAuthWrapper_simulateSignInSuccess_returnsFakeAuthStateSuccess() {
+    fakeFirebaseAuthWrapperImpl.simulateSignInSuccess()
+
+    val authState = fakeFirebaseAuthWrapperImpl.getAuthState()
+    assertThat(authState).isInstanceOf(FakeAuthState.Success::class.java)
+  }
+
+  @Test
+  fun testFakeAuthWrapper_simulateSignInSuccess_returnsFakeAuthStateFailure() {
```
BenHenning:
```suggestion
  fun testFakeAuthWrapper_simulateSignInFailure_returnsFakeAuthStateFailure() {
```

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
```diff
@@ -29,8 +37,14 @@ class ViewEventLogsViewModel @Inject constructor(
   }
 
   private fun processEventLogsList(): List<EventLogItemViewModel> {
-    return eventList.map {
-      EventLogItemViewModel(it, machineLocale, resourceHandler)
-    }.reversed()
+    return eventList
+      .apply {
+        addAll(analyticsEvents)
+        addAll(firestoreEvents)
```
BenHenning: This is going to change the eventList.

If we're not going to change eventList, you probably don't even need it anymore. Instead, you could just do something like:

```kotlin
return (analyticsEvents + firestoreEvents).map { ... }.sortedByDescending { ... }
```

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
BenHenning: I suggest moving the timeout to a constant.

That being said, is 3 seconds enough time? Normally network timeouts are on the order of ~30 seconds, so just wanted to check.

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,47 @@
...
+            },
+            onFailure = { exception ->
+              deferredResult.complete(AsyncResult.Failure(exception))
+            }
+          )
+        }
+      } catch (e: TimeoutCancellationException) {
+        // Handle timeout exception
+        deferredResult.complete(
+          AsyncResult.Failure(TimeoutException("Authentication attempt timed out"))
```
BenHenning: Suggest wrapping the original exception in something new, like an ``IllegalStateException``, to preserve the stack trace history.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
BenHenning: I think that's cleaner, but does that mean this can be private now?

BenHenning (approved):
Thanks @adhiamboperes! Just had a few follow-up comments. Overall the PR LGTM (so I've approved it as such), but feel free to assign me to take another pass on the final deltas if you'd like (I'm fine with you resolving my follow-up comments, as well).

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
adhiamboperes: Done

domain/src/test/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImplTest.kt
adhiamboperes: Fixed this

testing/src/main/java/org/oppia/android/testing/FakeFirebaseAuthWrapperImpl.kt
adhiamboperes: Update this to an enum

testing/src/test/java/org/oppia/android/testing/FakeFirebaseAuthWrapperImplTest.kt
adhiamboperes: Fixed, thanks!

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: I updated the timeout to 30s and also moved it to a constant.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
adhiamboperes: Yes, I changed access to private

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/FirestoreDataController.kt
adhiamboperes: Yes, I modified this logic to call `removeFirstEventLogFromStore()` on each iteration and added tests to verify it.

app/src/main/java/org/oppia/android/app/devoptions/vieweventlogs/ViewEventLogsViewModel.kt
adhiamboperes: I have made the change per your suggestion

domain/src/main/java/org/oppia/android/domain/auth/AuthenticationController.kt
adhiamboperes: Done

adhiamboperes:
@BenHenning, I have resolved all the conversations as I was able to verify the changes myself. This PR is ready to be merged, thanks!

PR #5103, opened 2023-07-24

translatewiki:
Translation updates


adhiamboperes:
We don't need this since we have updated this string and it needs to be re-translated.

PR #4929, opened 2023-03-29
Fix part of #59, part of #3926: Upgrade compute affected tests

BenHenning:
## Explanation
Fix part of #59
Fix part of #3926

As part of developing downstream PRs for #59, it was discovered that PRs which change a LOT of files (such as #4937) can run into problems where ComputeAffectedTests simply times out trying to compute the entire list of affected targets.

**Critical performance and compatibility fixes**

There have been past efforts to optimize the affected tests workflows (bucketing, breaking up some of the computations), but it was discovered that the most expensive part of the process is running the ``rbuildfiles`` query to figure out which BUILD files were affected by changed files. It was known this was an expensive step in the past, but it wasn't clear until this PR exactly how to address it. This PR changes the script to now:
- Filter ``rbuildfiles`` to only run under first-party targets (since it shouldn't be possible for third-party BUILD files to be affected by first-party changes). This reduces the search graph.
- Introduce Bazel command sharding for this step like the script already does for others. This breaks up the command to run on a subset of files multiple times, combining the output rather than running a single command with a large input. It seems that ``rbuildfiles`` does not scale linearly with its input size, so this drastically improves the script's performance. It's thought that this approach is also more logically correct due to more correct sibling matching semantics, but it's a bit hard to reason about ``bazel query`` behavior at times so I can't be 100% confident in this. Nevertheless, the existing tests pass and I haven't seen any issues from using these changes in downstream PRs.

Separately, another issue was discovered wherein some commands (including certain Bazel commands) can actually cause ``CommandExecutorImpl`` to soft-lock and always time out. This was due to an issue in the previous implementation wherein it would wait to read a command's stdout until after the timeout has been completed (i.e. it assumed the process would finish). This isn't correct, however: stdout is blocking I/O, and some commands are implemented to only continue execution after their standard output is read. The new implementation makes use of coroutine actors to consume stdout and stderr at the same time as waiting for execution to ensure all commands can continue execution and that they finish within the desired timeout. Note that the new ``ScriptBackgroundCoroutineDispatcher`` was actually introduced (in #5313) specifically to support this new ``CommandExecutorImpl`` implementation, though it has since been found to have lots of other nice benefits by providing scripts with a reliable mechanism for performing asynchronous operations without having to manage their own execution dispatcher.

[...]

PR #5138, opened 2023-08-22
[RunAllTests] Fix #5303, #5304, #5305, #5306, #5309, part of #5307, part of #5308: Fix a variety of dev platform-specific issues

### LINKED ISSUE #5303: [BUG]: AssertionHelpers.assertThrows() doesn't handle AssertionErrors correctly
### Describe the bug

AssertionHelpers.assertThrows() is a utility to verify that an exception is correctly thrown during test execution. However, during the development of #5138 it was discovered that test utilities trying to catch an AssertionError will not ever fail due to how assertThrows is implemented.

### Steps To Reproduce

Create a test that always throws an AssertionError and wrap that in assertThrows.

### Expected Behavior

assertThrows should allow catching _any_ exception type.

### Screenshots/Videos

_No response_

N/A -- test environment

N/A -- test environment

N/A -- test environment

### Additional Context

_No response_

### LINKED ISSUE #5304: [BUG]: LoggingIdentifierController has potential non-determinism in test environments
### Describe the bug

LoggingIdentifierController relies on an internal ``learnerIdRandom`` which is used for learner ID, installation ID, and session ID generation. This can create an order issue in tests since the order of tests will affect which of these IDs are generated first, ultimately changing other downstream IDs (even if the initial seed is fixed).

### Steps To Reproduce

Run individual tests in LoggingIdentifierControllerTest rather than the whole suite. Some of them will fail due to incorrectly generated IDs.

### Expected Behavior

LoggingIdentifierControllerTest & any other tests that depend on LoggingIdentifierController should be able to be run in any order.

### Screenshots/Videos

_No response_

N/A -- test environment

N/A -- test environment

N/A -- test environment

### Additional Context

_No response_

### LINKED ISSUE #5305: [BUG]: model/src/main/proto/BUILD.bazel has build issues
### Describe the bug

In model/src/main/proto/BUILD.bazel the proto feedback_reporting_proto requires languages.proto but doesn't actually depend on it. This results in a build error, but only when building all targets in the app (e.g. ``bazel build //...``) since, as of this issue, no library yet depends on feedback_reporting_proto.

### Steps To Reproduce

Try to build all proto libraries, e.g.: ``bazel build //model/...``.

### Expected Behavior

All model targets should build without any errors.

### Screenshots/Videos

_No response_

N/A

N/A

N/A

### Additional Context

_No response_

### LINKED ISSUE #5306: [BUG]: TestGitRepository can have environment-specific behavior
### Describe the bug

TestGitRepository is currently written to assume that a global user and email has been set in the local git environment. This is inherently fragile as configurations outside the repository shouldn't affect the test environment.

### Steps To Reproduce

Try to run TestGitRepositoryTest when no user.name or user.email is set for the machine's local global configuration.

### Expected Behavior

TestGitRepository should be robust against global configurations by manually maintaining and setting a repository-local user name and email.

### Screenshots/Videos

_No response_

N/A -- test environment

N/A -- test environment

N/A -- test environment

### Additional Context

_No response_

### LINKED ISSUE #5309: [BUG]: ProfileAndDeviceIdFragmentTest has order & machine-dependent behavior
### Describe the bug

When running certain tests in isolation within ProfileAndDeviceIdFragmentTest, a variety of order-independence issues arise causing difficulty in test behavior isolation. Additionally, there are proto comparisons happening that can have subtle differences depending on the local runtime environment.

### Steps To Reproduce

Run each of the tests in ProfileAndDeviceIdFragmentTest in isolation and notice different failures.

### Expected Behavior

All tests in ProfileAndDeviceIdFragmentTest should not have an order dependence, and should work in all runtime environments.

Note that some of these issues are caused by #5307 and #5304, but the proto behavior is specific to this test.

### Screenshots/Videos

_No response_

N/A -- test environment

N/A -- test environment

N/A -- test environment

### Additional Context

_No response_

BenHenning:
## Explanation
Fixes #5303
Fixes #5304
Fixes #5305
Fixes #5306
Fixes #5309
Fixes part of #5307
Fixes part of #5308

This PR fixes a variety of issues that were discovered while trying to begin development on #4929 with a different workstation than I usually build on.

Some of the issues were outright problems at the outset, but others were discovered when trying to fix those issues (e.g. during test utility upgrades).

Specific issues that were found and fixed:
- #5303: ``AssertionHelpers.assertThrows`` didn't handle cases when trying to catch an ``AssertionError`` itself (such in the case of ``TestGitRepositoryTest``). This could result in false positives, and it has since been fixed. Plus, the changes to ``AssertionHelpers`` also removes the dependency on Kotlin reflection.
- #5304: ``LoggingIdentifierController`` could have ID generation inconsistencies between environments due to how its ``Random`` instance was initialized. This could also result in a test order issue if a subset of tests were run, or run in a different order. The controller has been updated to have better robustness against the order in which IDs are generated which fixes the test and environment determinism issues. \
\
  The fix was essentially generating ID-specific seeds for ID-specific randoms that all originate from the application seed, and are always generated in the same order deterministically at the creation time of the controller. This fixes the IDs being generated in different orders leading to inconsistencies. Note that this is why ID changes are seen in a variety of tests (since there are now new seeds being used for generating these IDs as a by-product of the robustness improvements).
- #5305: ``model/src/main/proto/BUILD.bazel`` needed a fix due to a build warning (that only occurs when trying to build all targets).
- #5306: ``TestGitRepository`` was made more robust by better managing and setting the configured user & email used in its Git commands. The utility now manages a locally set configuration for each rather than relying on the global configuration (which may not be present on all systems, hence the original issue). Some other improvements in error detection were also added. Note that this likely wasn't discovered in CI because it seems that GitHub CI _does_ set up a Git user & email automatically, as do most people who work on the project.
- #5307 & #5308: ``ActivityTestRule`` is now deprecated in favor of ``ActivityScenarioRule`` since it can interact with activities outside their lifecycle. However, the scenario rule is also not good to use anymore since it can result in partial dependencies being initialized before test-only platform parameter states can be overridden. This PR introduces a new pattern to prevent using either of these rules, and the mentioned issues can be used to track the remaining tests that need to be cleaned up following this PR.
- #5309: ``ProfileAndDeviceIdFragmentTest`` is environment and order-dependent due to out-of-order platform parameter initialization. This issue ultimately arose from test parameters being initialized after they're needed in the launched UI. It is necessary when there are environment differences (e.g. local vs. CI) or when running certain tests individually. This was fixed by removing the usage of ``ActivityScenarioRule`` and instead managing the fragment's test activity lifecycle directly. \
\
  Separately, the test performs proto comparisons that can be machine-dependent. It seemed the test was suffering from some proto encoding inconsistencies that seem to occur between some development machines vs. on CI. The fix improves the test's robustness by extracting the raw encoded string, verifying that the other outputs in the intent message correctly correspond to that string, and that the string (as a parsed proto) contains the correct values. As a result, the test no longer depends on a hardcoded encoding value to be present for verification. This does result in a slightly lengthier test with more logic, but should be more stable in the long-term.

Some other miscellaneous notes:
- ``testCreateLearnerId_verifyCreatesCorrectRandomValue`` was removed from ``LoggingIdentifierControllerTest`` because it wasn't actually providing useful testing value. The application seed is not itself part of the class's public API. Instead, the ID generation methods should be ultimately verified.
- ``ComputeAffectedTestsTest`` had a shard increase to 24 and ``BazelClientTest`` has now been sharded to 4 (ditto for ``GenerateMavenDependenciesListTest`` and ``MavenDependenciesListCheckTest``). Both were done to help distribute the more expensive tests in each suite across multiple runners to try and reduce the long-tail of script test runs. More optimization will probably be useful in the future. Note also that some of these tests are *much* slower on certain systems (the one I've been working on, for instance), hence the need for these adjustments.
- ``testEnsureNextResultIsSuccess_failureThenSuccess_notified_throwsException`` was removed from ``DataProviderTestMonitorTest`` since it *seems* that it wasn't actually written correctly (and the issue wasn't discovered until ``assertThrows`` was fixed). Per the existing test coverage, it doesn't seem necessary to try and fix the test so it was instead removed.
- ``ProfileAndDeviceIdFragmentTest`` changes led to a new ``EventLogSubject.hasNoProfileId`` being added.
- There are a few miscellaneous script infrastructure changes that were pulled into this PR in preparation for downstream work. Specifically:
  - ``ComputeAffectedTests`` now allows for its command executor to be used for Bazel operations (which provides its tests with a bit more execution flexibility). Its tests were also updated to use a longer timeout for commands to try and improve test stability. ``BazelClientTest`` was similarly updated.
  - ``TestBazelWorkspace`` was updated to:
    - Have better workspace initialization robustness (since the workspace can actually be generated at several different points, not singularly with a call to ``initEmptyWorkspace``).
    - Add a ``.bazelrc`` file to disable bzlmod (since it causes network stability issues in some environments when using newer Bazel versions even if the project isn't upgraded, and we don't use bzlmod yet, anyway).
    - Add a ``.bazelversion`` file to ensure that tests are using the same Bazel version as the project build.
- ``AuthenticationControllerTest`` was largely updated since the changes to ``assertThrows`` behavior revealed that the failure test wasn't actually correctly verifying an exception was being passed (since it isn't actually being thrown; the test was a false positive possibly because it was using ``Throwable``, but I didn't dig into exactly why it was passing before). The changes ensure that the callbacks are actually verified rather than the user result (since there's a separate test to do that), as well as the exception's failure message.

## Essential Checklist

## For UI-specific PRs only
N/A since all of the changes in this PR are infrastructural or only affect tests.


BenHenning (commented):
Self-reviewed the whole PR and addressed issues in previous commits.

BenHenning:
This PR _should_ be ready for review, but I'll actually wait a bit to make sure no fast checks fail in CI before I send this out.

BenHenning:
Okay it just looks like the only failure pertains to a few new tests that were added but aren't being covered by Bazel. That's being addressed in e1802c8242eaa7877360c8301177a1e8980f722b from #5335, but there will also be follow-up fixes needed in the test itself in this PR.

BenHenning (commented):
Self-reviewed changes to AuthenticationControllerTest.

BenHenning:
This should be ready for review at this point, so sending it out.

BenHenning:
PTAL @adhiamboperes.

BenHenning:
NB: This was made to block on #5335 because, given the nature of this PR, I verify it locally by building & running all tests (which was failing without the fixes introduced in that PR). This also puts the upcoming Bazel chain in a good state for chained development.

scripts/src/javatests/org/oppia/android/scripts/common/GitClientTest.kt
```diff
@@ -37,14 +37,14 @@ class GitClientTest {
     // Print the status of the git repository to help with debugging in the cases of test failures
     // and to help manually verify the expect git state at the end of each test.
```
adhiamboperes:
```suggestion
    // and to help manually verify the expected git state at the end of each test.
```

adhiamboperes (approved):

Noticed a typo from an existing comment.

The refactor for assertThrows is a really nice touch, along with the regex checks. It would be bice to see some flaky tests fixed when these are removed.

scripts/src/javatests/org/oppia/android/scripts/common/GitClientTest.kt
BenHenning: Done!

BenHenning:
Thanks @adhiamboperes! Added the suggested fix and am now enabling auto-merge.

PR #5208, opened 2023-10-26
Fix #5112: Rendering Inline SVGs

### LINKED ISSUE #5112: [BUG]: The Meaning of “Equal Parts” under Fractions
### Describe the bug

I got stuck because of this mistake - "and ___ of it gets eaten". This happened while playing the lesson: The Meaning of “Equal Parts” under Fractions.

There's no image and the fraction is missing. I have attached a screenshot of this card.

### Steps To Reproduce

1. Go to the Topic page for Fractions 
2. Click on “The Meaning of Equal Parts”
3. This is the 13th card (if you get all answers correctly).
4. 

### Expected Behavior

I expect to see the fraction being referred to in the card to enable me solve the problem. 

### Screenshots/Videos

IMG_0705

Samsung A13

_No response_

_No response_

### Additional Context

_No response_

adhiamboperes:
## Explanation
Fixes #5112 

We discovered that `mathImg_` dimesions are in **ex**, and not **px**, and this unit is not natively rendered on android.  

This PR introduces a method that will convert ex to px and use the resulting dimesions to compute the rendering size of a math image to be rendered inline. The formula is based on the [androidSvg library](https://github.com/oppia/androidsvg/blob/5bc9c7553e94c3476e8ea32baea3c77567228fcd/androidsvg/src/main/java/com/caverock/androidsvg/androidrendering/SVGAndroidRenderer.java#L5018) that is also based on CSS3 specs.

To preserve the size computation of block SVGs, we have created a seperate method that handles the calculation of rendering dimensions for inline SVGs.

We have only tested these changes through visual observation, and confirming that the images scale correctly when reading text size is changed.

## Essential Checklist

## For UI-specific PRs only
|Before|After|
|--|--|
|IMG_0705|Screenshot_1698283405|
|Screenshot_1698284125|Screenshot_1698283411|
|Screenshot_1697714168|Screenshot_1698284519|
|Screenshot_1697745323|Screenshot_1698284666|
|Screenshot_1697714820|Screenshot_1698284933|

Images scaled to Largest text and display size, and dark mode
||||
|--|--|--|
|image|image|image|
|image|image|image|


WORKSPACE
```diff
@@ -125,7 +125,7 @@ git_repository(
 # to correctly size in-line SVGs (such as those needed for LaTeX-based math expressions).
 git_repository(
     name = "androidsvg",
-    commit = "1265eb1087056cf3fc2e10442e5545bc65c109ce",
+    commit = "5bc9c7553e94c3476e8ea32baea3c77567228fcd",
```
BenHenning, line 128: Ah, we didn't ever update this for the release?

WORKSPACE
```diff
@@ -125,7 +125,7 @@ git_repository(
...
     remote = "https://github.com/oppia/androidsvg",
     shallow_since = "1686302944 -0700",
```
BenHenning: Please update this, too (per the warning that should show up when building //:oppia_dev).

utility/src/main/java/org/oppia/android/util/parser/svg/SvgPictureDrawable.kt
```diff
@@ -100,11 +100,17 @@ abstract class SvgPictureDrawable(
...
    */
   protected fun reinitialize(textPaint: TextPaint?) {
-    picture = textPaint?.let {
-      scalableVectorGraphic.renderToTextPicture(it)
-    } ?: scalableVectorGraphic.renderToBlockPicture()
+    val newPicture = if (textPaint != null) {
+      intrinsicSize = scalableVectorGraphic.computeSizeSpecsForTextPicture(textPaint)
+      scalableVectorGraphic.renderToTextPicture(textPaint)
+    } else {
+      intrinsicSize = scalableVectorGraphic.computeSizeSpecs(null)
```
BenHenning: Does computeSizeSpecs still need to include this argument if it's not ever passed anymore?

Alternatively, might it be possible to consolidate the changes in the new function to be part of computeSizeSpecs? It would simplify this code (it wouldn't need to change) while also making computeSizeSpecs correct for the case when textPaint is provided.

BenHenning (commented):
Thanks @adhiamboperes! Left a couple thoughts--PTAL.

utility/src/main/java/org/oppia/android/util/parser/svg/SvgPictureDrawable.kt
adhiamboperes: The new function is necessary because the alternative looks like this: 
```
 return if (textPaint != null) SvgSizeSpecs(
      adjustedWidth,
      adjustedHeight,
      renderedWidth = adjustedWidth,
      renderedHeight = adjustedHeight,
      verticalAlignment
    ) else SvgSizeSpecs(
      intrinsicWidth,
      intrinsicHeight,
      renderedWidth = imageFileNameWidth ?: intrinsicWidth,
      renderedHeight = imageFileNameHeight ?: intrinsicHeight,
      verticalAlignment = 0f
    )
```

Besides, there are addtional computations in `computeSizeSpecsForTextPicture` that would increase the size of the `computeSizeSpecs` function.

WORKSPACE
adhiamboperes: No we didn't.

utility/src/main/java/org/oppia/android/util/parser/svg/SvgPictureDrawable.kt
adhiamboperes: I've removed the textPaint argument from computeSizeSpecs so it is not called where textpaint needs to be passed. I have also updated the KDocs.

WORKSPACE
adhiamboperes: Done

BenHenning (approved):
Thanks @adhiamboperes. The latest changes LGTM!

Issue #3363, opened 2021-06-23
Tests are failing on Espresso in AddProfileActivityTest[Blocked on #4991]


rt4914:
Tests are failing on Espresso in AddProfileActivityTest

<img width="1576" alt="Screenshot 2021-06-23 at 4 38 02 PM" src="https://user-images.githubusercontent.com/9396084/123086713-635e0180-d441-11eb-97e9-3b7957b8fb70.png">

`testAddProfileActivity_inputPin_configChange_downloadAccessSwitchIsOn()` was removed in #5229. The behaviour was deemed not important to the UX. 
The test failed when testing the checkbox should preserve its state after the screen rotated, which native checkbox doesn't have this ability and we agree that we don't need extra implementation to achieve this.

`testAddProfileActivity_configChange_inputShortPin_create_pinLengthError()` was improved in #5229, but it is still flaky. @XichengSpencer left some comments on this issue thread of his attempts to fix it, as well as a link to an external gist.

Developers should run the third test multiple times to deterrmine if it is still flaky.

FareesHussain:
@rt4914 It looks like a espresso scrolling issue,
I think we need to introduce [nestedScrollTo](https://stackoverflow.com/a/46037284/12088301) here.


On testing it manually, it was working fine
so I tried to use scrollTo() to the above view (to the targeted view) and it worked.


In case the nestedScrollTo() mentioned didn't work we can perform scrollTo() to the view above the targeted view.

rt4914:
Tested this again on `Pixel 3 XL` and only `testAddProfileActivity_configChange_inputShortPin_create_pinLengthError` is failing now.

XichengSpencer:
Try to repro the failing test result without emulator. They all ran 3 times and had identical results.
![Screenshot 2023-09-22 172511](https://github.com/oppia/oppia-android/assets/74568012/34bc5e7c-e63d-434f-8dcb-44710d64cfb8)
![Screenshot 2023-09-22 171358](https://github.com/oppia/oppia-android/assets/74568012/4a85ac51-380c-4e7c-938c-542c8235e1bc)
![Screenshot 2023-09-22 172356](https://github.com/oppia/oppia-android/assets/74568012/18ad7385-59ed-4604-aeee-c2beb0babf81)

Gist for failed test message: https://gist.github.com/XichengSpencer/c6f7b537ef783afc0ac56166f87ad1a4

XichengSpencer:
[...]

PR #5229, opened 2023-11-15
Fix part of #3363 AddProfileActivityTest some test didn't pass.

XichengSpencer:
## Explanation
Fix part of #3363 
Remove testAddProfileActivity_inputPin_configChange_downloadAccessSwitchIsOn() as discussed with Ben and Adhiambo.

-    The old test failed when testing the checkbox should persevere its state after the screen rotated, which native checkbox doesn't have this ability and we agree that we don't need extra implementation to achieve this. 

Negate download access logic in AddProfileActivityPresenter so it will correctly set the listener.

Improve testAddProfileActivity_configChange_inputShortPin_create_pinLengthError() by adding the swipeup() action to make the create button fully visible during the test but the test is still flaky

## Essential Checklist

## For UI-specific PRs only

Screenshot 2023-11-15 160038


app/src/main/java/org/oppia/android/app/profile/AddProfileActivityPresenter.kt
```diff
@@ -63,7 +63,7 @@ class AddProfileActivityPresenter @Inject constructor(
       lifecycleOwner = activity
       viewModel = profileViewModel
     }
-    if (!enableDownloadsSupport.value) {
+    if (enableDownloadsSupport.value) {
       binding.addProfileActivityAllowDownloadConstraintLayout.setOnClickListener {
         allowDownloadAccess = !allowDownloadAccess
```
adhiamboperes: I still think the code on line 68 is ambiguous--

`allowDownloadAccess` is false by default. On ln68, we are saying ` allowDownloadAccess = !allowDownloadAccess` which means we are are negating the current value(false) and transfmorming it to true(ie. !false == true).
And if this is the intention of this expression, then can we use the boolean values `true/false` to convey the intended meaning?

adhiamboperes (changes requested):
Thanks @XichengSpencer.

I have left a comment in-line.

Also, could you please update the PR description(the title too) as previously discussed?

app/src/main/java/org/oppia/android/app/profile/AddProfileActivityPresenter.kt
XichengSpencer: I am not very sure that I understand your question but since this listener is associated with a switch, and the user can toggle between on and off states many times it should not be trapped in either state forever.  That's why we use a negate logic other than assign it to a constant boolean value.

adhiamboperes (approved):
Thanks @XichengSpencer, This LGTM!

PR #5255, opened 2023-12-03
Remove quantity="zero" from English strings

amire80:
The "zero" quantity is not used in English. This has no effect on the app, but on translatewiki, where Oppia is localized, it is flagged as an error by the validator.


adhiamboperes (changes requested):
Thanks @amire80,

Could you please specify the translate wiki validation error? Possibly by sharing a screenshot.

We introduced `zero` since per [docs](https://developer.android.com/guide/topics/resources/string-resource#Plurals), it is handled differently in languages such as Arabic.

A more complete solution here would be to remove the plurals from english but ensure it is in the arabic file. It is my undersstanding that translations rely on what is in the english file as a base.

It is also part of our acceptance criteria to include screenshots that show user facing changes like this, and incase you don't have the project set up to do this verification, I suggest that you convert this to an issue instead, and one of our contributors will work on it.

amire80:
> Thanks @amire80,
> 
> Could you please specify the translate wiki validation error? Possibly by sharing a screenshot.

You can see, for example, here:
https://translatewiki.net/wiki/Special:LanguageStats?language=en&suppresscomplete=1&group=mediawiki&x=D

It's a list of projects with untranslated messages into English (note the "en" code at the top). Naturally, all the messages are "translated" into English, because English is the source language for everything, but "Oppia Android" is the only project in the list with three "untranslated" messages because if messages have any validation errors, they are considered "untransated".

> We introduced `zero` since per [docs](https://developer.android.com/guide/topics/resources/string-resource#Plurals), it is handled differently in languages such as Arabic.

"zero" is indeed necessary in Arabic. (I'm also quite happy to say that the specific case of Arabic plural forms is very well-documented in translatewiki: https://translatewiki.net/wiki/Localisation_guidelines/ar . There used to be issues with Arabic translations, but a year ago, I attended an international conference in Dubai about the Arabic language in our projects, and I worked with native Arabic speakers and experienced translators to address them all and write comprehensive documentation to avoid future issues.)

However, while "zero" is necessary in Arabic, it is not necessary in English. In fact, it has no effect in English at all (except producing validation errors in translatewiki). The plural support in Android is based on the CLDR, which defines only "one" and "other" for English, but zero, one, two, few, many, and other for Arabic. Other languages have different sets of forms. You can see the list for all the languages here: https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html .

> A more complete solution here would be to remove the plurals from english but ensure it is in the arabic file. It is my undersstanding that translations rely on what is in the english file as a base.

This is the correct thinking, and I'm glad to say that this is already ensured by the same validator in translatewiki. It sees unsupported "zero" forms in English, and marks the messages as "untranslated", and it applies the same logic to Arabic: if a translator tries to publish a translation without "zero" or "two", an error will be shown. The first two messages are already correctly translated into Arabic with all the forms:
* https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=oppia-android-app-completed_story_count&group=oppia-android-app&language=ar&filter=&optional=1&action=translate
* https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=oppia-android-app-ongoing_topic_count&group=oppia-android-app&language=ar&filter=&optional=1&action=translate

The third message (minutes_ago) is not yet translated into Arabic. A volunteer should do it, and if they omit any forms, they'll see a validation error. Here's a link to this particular message:
* https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=oppia-android-app-minutes_ago&group=oppia-android-app&language=ar&filter=&optional=1&action=translate

And here's a link to all the messages untranslated into Arabic, because why not:
* https://translatewiki.net/w/i.php?title=Special:Translate&group=oppia-android-app&language=ar&filter=%21translated&action=page

> It is also part of our acceptance criteria to include screenshots that show user facing changes like this, and incase you don't have the project set up to do this verification, I suggest that you convert this to an issue instead, and one of our contributors will work on it.

Sorry, I don't have a whole development environment for that. I made similar patches in a bunch of other projects (e.g. OpenStreetMap), and they were always accepted. If you insist, I'll make an issue, but I'm quite sure that the eventual patch by someone else will be the same :)

adhiamboperes (approved):
Thanks @amire80, this looks good.

PR #5277, opened 2023-12-19
Fix #4206 Create New HtmlParser Factory Method

### LINKED ISSUE #4206: Replace MathExpressionParserViewModel html parser usage with better factory function
#2173 introduces a new MathExpressionParserViewModel that uses html parser factory in a context where images shouldn't be loaded. This flow was simplified in #3852, so this issue tracks updating the model to use the new factory method.

Create a new factory method in HtmlParser, just like the [policies one](https://github.com/oppia/oppia-android/blob/6fec2e27244274e5e23a93f5051f5862218bc21c/utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt#L188C3-L188C5).  What is required here, is that the new create function is such that it does not request for the arguments that are not set currently here:

Screenshot 2023-12-08 at 01 30 21

XichengSpencer:
## Explanation

Fix #4206 Create New HtmlParser Factory Method to only take customOppiaTagActionListener and displayLocale as parameter and replace factory method usage in MathExpressionParserViewModel with this new implementation

## Essential Checklist

## For UI-specific PRs only

Screenshot 2024-01-07 142738

Screenshot 2024-01-07 145604


utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -240,6 +245,30 @@ class HtmlParser private constructor(
       )
     }
 
+    /**
+     * Returns a new [HtmlParser] with the empty entity type and ID for loading images,
+     * doesn't require GCS properties and imageCenterAlign set to false
+     * optionally specified [CustomOppiaTagActionListener] for handling custom Oppia tag events.
+     */
+    fun create(
```
adhiamboperes, line 253: I think as per the calling method, only two arguments are used, and these are `displayLocale` and `imageCenterAlign`, so this new factory method should only have those two.  `imageCenterAlign` can be set with a default of `false` here (this parert is already correct). I am not certain about the need for `customOppiaTagActionListener:`

adhiamboperes (changes requested):
Thanks @XichengSpencer! 

Do you need to update the MathExpressionParser tests as well to use this new method?

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
XichengSpencer: Yeah I will also take a look at the test. For customOppiaTagActionListener, I remove it in the input arg. And just to confirm I should set it as null in the function body?

adhiamboperes: Yes, unless you can find any usages where it is not null

XichengSpencer: I updated the tests related to the MathExpressionParserViewModel in the pr description. I search MathExpressionParserViewModel usages in test classes, and run the classes test for them. Let me know if there is any concern with this approach.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -240,6 +245,29 @@ class HtmlParser private constructor(
...
+        context,
+        urlImageParserFactory,
+        gcsResourceName = "",
+        entityType = "",
+        entityId = "",
+        imageCenterAlign = false,
+        consoleLogger,
+        cacheLatexRendering = enableCacheLatexRendering.value,
+        customOppiaTagActionListener = null,
+        null,
```
adhiamboperes: This should be a named variable like the rest. Do avoid mixing named and unnamed variables in the same function.

app/src/main/java/org/oppia/android/app/devoptions/mathexpressionparser/MathExpressionParserViewModel.kt
```diff
@@ -33,10 +33,6 @@ class MathExpressionParserViewModel @Inject constructor(
   private val htmlParser by lazy {
     // TODO(#4206): Replace this with the variant that doesn't require GCS properties.
     htmlParserFactory.create(
-      gcsResourceName = "",
```
adhiamboperes, line 36: Remove the TODO statement above because it will prevent the issue from being automatically closed.

adhiamboperes (changes requested):
Thanks @XichengSpencer!

Left a couple of minor comments.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
XichengSpencer: Hi Adhiambo, thanks for reviewing! Is this null variable the only unnamed variable? For others like consoleLogger, should we replace it as consoleLogger = consoleLogger?

adhiamboperes: It is adviasable to avoid mixing named an unnamed variables. So in this case, we should also replace the rest like you've metioned, e.g `consoleLogger = consoleLogger`

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -124,11 +123,17 @@ class HtmlParser private constructor(
     }
 
     val imageGetter = urlImageParserFactory?.create(
-      htmlContentTextView, gcsResourceName, entityType, entityId, imageCenterAlign
```
XichengSpencer, line 127: I found, that besides the two factory methods this PR is targeting, there are other unnamed variables in other functions. For example this one. Should I change them all?

adhiamboperes: It's best to just focus on this current PR. Thanks

adhiamboperes (approved):
Thanks @XichengSpencer! This LGTM!

PR #5211, opened 2023-10-28
Fix #5093 : Crash on api below 24

Shaik-Sirajuddin:
## Explanation

- Fixes #5093 , 
- Implemented Check for API Level 24 and above in all places where **computeIfAbsent of concurrenthashmap** was used and handled accordingly.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

Reference Videos : 
[api_23_crash_before.webm](https://github.com/oppia/oppia-android/assets/89742297/22ca6541-a7d2-4358-982c-d90323ab751e)
[api_23_crash_after_fix.webm](https://github.com/oppia/oppia-android/assets/89742297/424fabb6-915d-43d9-964f-a264850ede72)

Issue #5046, opened 2023-06-11
[BUG]: App can apparently crash due to an incompatible IETF BCP-47 language ID


BenHenning:
**Describe the bug**
Three devices in the latest release (beta 0.11 RC02) candidate's pre-launch report triggered the following stack trace:

```
Exception Process: org.oppia.android, PID: 8313
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.oppia.android/org.oppia.android.app.onboarding.OnboardingActivity}: java.lang.IllegalStateException: Invalid ID: # fg.fX@759b4323.
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3555)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3707)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:83)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2220)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  at android.os.Looper.loop (Looper.java:237)
  at android.app.ActivityThread.main (ActivityThread.java:8016)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:493)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1076)
Caused by java.lang.IllegalStateException: Invalid ID: # fg.fX@759b4323.
  at org.oppia.android.util.locale.AndroidLocaleFactory$LocaleSource.expectedProfile (AndroidLocaleFactory.java:212)
  at org.oppia.android.util.locale.AndroidLocaleFactory$LocaleSource.toForcedProposal (AndroidLocaleFactory.java:200)
  at org.oppia.android.util.locale.AndroidLocaleFactory$LocaleSource.computeForcedProposal (AndroidLocaleFactory.java:182)
  at org.oppia.android.util.locale.AndroidLocaleFactory$AndroidResourceCompatibilityPreferredChooser.findBestProposal (AndroidLocaleFactory.java:321)
  at org.oppia.android.util.locale.AndroidLocaleFactory$createAndroidLocale$1.apply (AndroidLocaleFactory.java:63)
[...]

PR #5291, opened 2023-12-28
Fix #5093: Fix start-up crash on SDK versions <24

### LINKED ISSUE #5093: [BUG]:Oppia-Android app not opening on the Android 6.0.1 Marshmello
### Describe the bug

I setup the oppia android project locally and deployed the app to my android device (6.0.1) but as i open the app it closes on its own.
But the app is working fine on higher specs devices.

### Steps To Reproduce

setup oppia-android locally 
deploy the app on a abdroid 6.0.1 device

### Expected Behavior

It should work fine for the android 6.0.1 devices.

### Screenshots/Videos

https://github.com/oppia/oppia-android/assets/109658545/eb8a8781-23c1-4a33-a842-c2ed1f43369c

### Stacktrace
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.oppia.android, PID: 1041
    java.lang.NoClassDefFoundError: org.oppia.android.util.locale.AndroidLocaleFactory$createAndroidLocale$1
        at org.oppia.android.util.locale.AndroidLocaleFactory.createAndroidLocale(AndroidLocaleFactory.kt:59)
        at org.oppia.android.util.locale.DisplayLocaleImpl$formattingLocale$2.invoke(DisplayLocaleImpl.kt:25)
        at org.oppia.android.util.locale.DisplayLocaleImpl$formattingLocale$2.invoke(DisplayLocaleImpl.kt:17)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at org.oppia.android.util.locale.DisplayLocaleImpl.getFormattingLocale(DisplayLocaleImpl.kt)
        at org.oppia.android.util.locale.DisplayLocaleImpl.setAsDefault(DisplayLocaleImpl.kt:46)
        at org.oppia.android.domain.locale.LocaleController.setAsDefault(LocaleController.kt:239)
        at org.oppia.android.app.translation.ActivityLanguageLocaleHandler.initializeLocaleForActivity(ActivityLanguageLocaleHandler.kt:41)
        at org.oppia.android.app.activity.InjectableAppCompatActivity.onInitializeLocalization(InjectableAppCompatActivity.kt:77)
        at org.oppia.android.app.activity.InjectableAppCompatActivity.attachBaseContext(InjectableAppCompatActivity.kt:33)
        at android.app.Activity.attach(Activity.java:6288)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2443)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2577)
        at android.app.ActivityThread.access$1000(ActivityThread.java:166)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1414)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:148)
        at android.app.ActivityThread.main(ActivityThread.java:5628)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:853)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:737)

Vivo y55s

Android 6.0.1

App version 1.0

### Additional Context

_No response_

BenHenning:
## Explanation
Fixes #5093

This replaces #5211.

The core crash has come from using Java 8's [Map.computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) on devices running SDK <24. This function isn't supported on these SDK versions because:
1. Java 8 isn't fully supported until SDK 24 (https://stackoverflow.com/a/35934691).
2. While the build toolchain "desugars" Java 8 syntax and some APIs to provide support for Java 8 before SDK 24, not all functionality is present. While https://developer.android.com/studio/write/java8-support-table suggests that ``computeIfAbsent`` _should_ be present, it's not clear why it isn't (it could be an out-of-date desugarer, or some other issue).

Only two uses for ``computeIfAbsent`` are present:
- ``AndroidLocaleFactory``: the main cause of the crash.
- ``FakeAssetRepository``: only used in tests, so ignored in this change (since it would require changing the production API for ``AssetRepository`` to fix).

A new regex pattern check was added to ensure this method can't inadvertently be used in the future since it won't work on lower API versions.

This PR mainly focuses on fixing ``AndroidLocaleFactory``. 

Need to review the code & finish it, and document the changes. #5211 explored different methods that we could take to keep synchronization when updating the factory to move away from ``computeIfAbsent``, but all of them required introducing locking mechanisms which could cause deadlocking. I also considered in this PR removing the memoization altogether, but this doesn't seem ideal either since the creation of profiles is non-trivial and locales are frequently fetched by nearly all core lesson data providers throughout the app. I landed on a solution that leveraged the app's blocking dispatcher and made profile creation asynchronous. This has some specific implications:
- Technically a "live lock" is still possible if the blocking dispatcher is starved. A better solution would be to use an actor-like pattern and funnel changes through the background dispatcher, but that's out of scope for this change and represents a problem with all uses of the blocking dispatcher.
- Creations of ``DisplayLocaleImpl`` get more complicated since creation of the ``Locale`` is now asynchronous. This was adjusted by passing in the display locale's ``Locale`` object rather than having it compute it, and callsites often already are operating within a coroutine context which makes the asynchronous aspect of ``Locale`` creation a bit simpler.
- Two cases where asynchronous creation cannot be used are the edge cases of initial app startup failing to fetching & create the locale in a timely manner, and initializing locale for all activities prior to the current locale being fetched. In both cases, the factory was updated to create a non-memoized ``Locale`` object specifically for these purposes. This method should only be used when necessary since it avoids the performance benefits of the memoized version.
- #5046 might be addressed since one plausible cause for that observed issue is a root locale being used, or a region-only locale (both of which should now be handled per this PR).

During development, there were two other changes that were made to ease development:
- ``AndroidLocaleProfile`` was refactored to use a sealed class to improve its typing, and to better facilitate the introduction of a root profile (which was added for better system compatibility in cases when the default app locale isn't supported on the system). Note that the new root profile only applies to app strings since it can actually have correct default behavior (whereas for content strings & audio voiceovers it can't actually be used effectively). The profile class was also updated to compute its own ``Locale`` (which is a simplification since before there were multiple ways to create a "correct" ``Locale``), and have an application-injectable factory.
- ``DataProviders.transformAsync`` was updated to handle caught exceptions in the same way as ``DataProviders.transform`` which helped while debugging the crash, and seems like it would have downstream benefits in the future. The method's tests were correspondingly updated.

The changes here led to some testing changes:
- ``testCreateDisplayLocaleImpl_defaultInstance_hasDefaultInstanceContext`` was removed since using the default context isn't valid in most cases (see below point).
- ``testReconstituteDisplayLocale_defaultContext_returnsDisplayLocaleForContext`` was renamed & a new test added to better represent that the default context is invalid except for app strings where it can represent root locale. For non-app cases, an exception should be thrown for default. This is also why ``testCreateLocale_appStrings_allIncompat_invalidLangType_throwsException`` was updated to instead verify that the root locale is returned.
- In ``TranslationControllerTest``, ``testGetSystemLanguageLocale_rootLocale_returnsLocaleWithBlankContext`` and ``testGetAppLanguageLocale_uninitialized_returnsLocaleWithSystemLanguage`` were updated to correctly indicate that there is no specified language for the root locale cases. To ensure coverage for valid IETF BCP-47 & Android resource language IDs, a new test was added: ``testGetAppLanguageLocale_ptBrDefLocale_returnsLocaleWithIetfAndAndroidResourcesLangIds``.

Finally, please note that this is fixing a release blocking issue for the upcoming 0.13 beta release of the app.

## Essential Checklist

## For UI-specific PRs only
This is mainly an infrastructure change. The only user behavior impacted is that the app no longer crashes on startup on SDK versions below 24.

Attempting to open the app on an SDK 23 emulator on develop (4a07d8db69e395c9e076c342c0d1dc3521ff3051):

open_app_without_fix_smaller

Attempting to open the app with the fixes from this branch:
open_app_with_fix_smaller


BenHenning:
NB: I made this blocking on #5138 since it makes it easier for me to develop the changes required for this branch.

scripts/assets/file_content_validation_checks.textproto
```diff
@@ -598,3 +598,10 @@ file_content_checks {
   exempted_file_name: "scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt"
   exempted_file_name: "testing/src/test/java/org/oppia/android/testing/espresso/TextInputActionTest.kt"
 }
+file_content_checks {
+  file_path_regex: ".+?\\.kt"
+  prohibited_content_regex: "computeIfAbsent\\("
+  failure_message: "computeIfAbsent won't desugar and requires Java 8 support (SDK 24+). Suggest using an atomic Kotlin-specific solution, instead."
+  exempted_file_name: "scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt"
+  exempted_file_name: "utility/src/main/java/org/oppia/android/util/caching/testing/FakeAssetRepository.kt"
```
BenHenning, line 606: Mention this in TODO description.

utility/src/main/java/org/oppia/android/util/locale/DisplayLocaleImpl.kt
```diff
@@ -13,16 +13,18 @@ import java.util.Locale
 import java.util.Objects
 
 // TODO(#3766): Restrict to be 'internal'.
-/** Implementation of [OppiaLocale.DisplayLocale]. */
+// TODO(#3766): Restrict formattingLocale to be 'internal'.
+/**
+ * Implementation of [OppiaLocale.DisplayLocale].
+ *
+ * @property formattingLocale the [Locale] used for user-facing string formatting
```
BenHenning, line 20: Add other properties.

Also, explain why this change was needed in PR description (creating a locale is no longer synchronous).

utility/src/main/java/org/oppia/android/util/data/DataProviders.kt
```diff
@@ -74,7 +75,12 @@ class DataProviders @Inject constructor(
         override fun getId(): Any = newId
 
         override suspend fun retrieveData(): AsyncResult<O> {
-          return this@transformAsync.retrieveData().transformAsync(function)
+          return try {
+            this@transformAsync.retrieveData().transformAsync(function)
+          } catch (e: Exception) {
```
BenHenning, line 80: TODO: Mention this in PR description & why.

utility/src/test/java/org/oppia/android/util/locale/DisplayLocaleImplTest.kt
```diff
@@ -56,13 +58,6 @@ class DisplayLocaleImplTest {
     setUpTestApplicationComponent()
   }
 
-  @Test
-  fun testCreateDisplayLocaleImpl_defaultInstance_hasDefaultInstanceContext() {
```
BenHenning, line 60: TODO: Mention why this was removed.

domain/src/main/java/org/oppia/android/domain/locale/LocaleController.kt
```diff
@@ -431,7 +436,7 @@ class LocaleController @Inject constructor(
     } ?: RegionSupportDefinition.newBuilder().apply {
       region = OppiaRegion.REGION_UNSPECIFIED
       regionId = RegionSupportDefinition.IetfBcp47RegionId.newBuilder().apply {
-        ietfRegionTag = countryCode
+        ietfRegionTag = machineLocale.run { countryCode.toMachineUpperCase() }
```
BenHenning, line 439: TODO: Verify there's test coverage for this.

domain/src/test/java/org/oppia/android/domain/locale/LocaleControllerTest.kt
```diff
@@ -114,12 +114,16 @@ class LocaleControllerTest {
   }
 
   @Test
-  fun testReconstituteDisplayLocale_defaultContext_returnsDisplayLocaleForContext() {
+  fun testReconstituteDisplayLocale_defaultContext_throwsException() {
```
BenHenning, line 117: TODO: Mention this in PR description.

domain/src/main/java/org/oppia/android/domain/locale/LocaleController.kt
```diff
@@ -473,9 +478,12 @@ class LocaleController @Inject constructor(
...
     appStringId = LanguageSupportDefinition.LanguageId.newBuilder().apply {
-      ietfBcp47Id = LanguageSupportDefinition.IetfBcp47LanguageId.newBuilder().apply {
-        ietfLanguageTag = systemLocaleProfile.computeIetfLanguageTag()
-      }.build()
+      // The root profile is assumed if there is no specific language ID to use.
+      if (systemLocaleProfile !is AndroidLocaleProfile.RootProfile) {
+        ietfBcp47Id = LanguageSupportDefinition.IetfBcp47LanguageId.newBuilder().apply {
+          ietfLanguageTag = systemLocaleProfile.ietfLanguageTag
+        }.build()
+      }
```
BenHenning, line 486: TODO: Verify test coverage.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
```diff
@@ -169,45 +180,41 @@ class AndroidLocaleFactory @Inject constructor(
...
-        // Empty region codes are allowed for Android resource IDs since they should always be used
-        // verbatim to ensure the correct Android resource string can be computed (such as for macro
-        // languages).
-        AndroidLocaleProfile(
-          it.languageCode,
-          regionCode = it.regionCode.ifEmpty { AndroidLocaleProfile.REGION_WILDCARD }
-        )
-      }
-    }
+    fun computeForcedProposal(allowRootProfile: Boolean): LocaleProfileProposal =
```
BenHenning: TODO: Rename all 'allowRootProfile' to 'fallBackToRootProfile' for clarity.

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleFactoryTest.kt
```diff
@@ -558,19 +612,17 @@ class AndroidLocaleFactoryTest {
   }
 
   @Test
-  fun testCreateLocale_appStrings_allIncompat_invalidLangType_throwsException() {
+  fun testCreateLocale_appStrings_allIncompat_invalidLangType_returnsRootLocale() {
```
BenHenning, line 615: TODO: Mention this behavior difference in the PR description.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
```diff
@@ -308,31 +328,43 @@ class AndroidLocaleFactory @Inject constructor(
...
-        ?: fallbackSource.computeSystemMatchingProposals().findFirstViable(machineLocale)
-        ?: fallbackSource.computeForcedAndroidProposal()?.takeOnlyIfViable(machineLocale)
-        ?: primarySource.computeForcedProposal()
+      // Note that defaulting to the root locale only makes sense for app strings (since app strings
+      // are picked based on the configured system locale).
+      return primarySource.computeSystemMatchingProposals().findFirstViable(localeProfileRepository)
+        ?: primarySource.computeForcedAndroidProposal()?.takeOnlyIfViable(localeProfileRepository)
+        ?: fallbackSource.computeSystemMatchingProposals().findFirstViable(localeProfileRepository)
+        ?: fallbackSource.computeForcedAndroidProposal()?.takeOnlyIfViable(localeProfileRepository)
+        ?: primarySource.computeForcedProposal(allowRootProfile = true)
```
BenHenning: TODO: Verify coverage on this boolean value.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
```diff
@@ -288,15 +308,15 @@ class AndroidLocaleFactory @Inject constructor(
...
     override fun findBestProposal(
       primarySource: LocaleSource,
       fallbackSource: LocaleSource
     ): LocaleProfileProposal {
-      return primarySource.computeSystemMatchingProposals().findFirstViable(machineLocale)
-        ?: fallbackSource.computeSystemMatchingProposals().findFirstViable(machineLocale)
-        ?: primarySource.computeForcedProposal()
+      return primarySource.computeSystemMatchingProposals().findFirstViable(localeProfileRepository)
+        ?: fallbackSource.computeSystemMatchingProposals().findFirstViable(localeProfileRepository)
+        ?: primarySource.computeForcedProposal(allowRootProfile = false)
```
BenHenning: TODO: Verify coverage on this boolean value.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
```diff
@@ -169,45 +180,41 @@ class AndroidLocaleFactory @Inject constructor(
...
+    private fun toForcedProposal(allowRootProfile: Boolean): LocaleProfileProposal {
+      return when (val languageTypeCase = languageId.languageTypeCase) {
         IETF_BCP47_ID -> createIetfProfile().expectedProfile()
         MACARONIC_ID -> createMacaronicProfile().expectedProfile()
-        LANGUAGETYPE_NOT_SET, null -> error("Invalid language case: $languageTypeCase.")
+        LANGUAGETYPE_NOT_SET, null -> {
+          if (allowRootProfile) {
+            AndroidLocaleProfile.RootProfile
+          } else error("Invalid language case: $languageTypeCase.")
+        }
```
BenHenning: TODO: Verify coverage on these cases.

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
```diff
@@ -112,9 +112,8 @@ class TranslationControllerTest {
     val appStringId = context.languageDefinition.appStringId
     assertThat(context.usageMode).isEqualTo(APP_STRINGS)
     assertThat(context.languageDefinition.language).isEqualTo(LANGUAGE_UNSPECIFIED)
-    assertThat(appStringId.languageTypeCase).isEqualTo(IETF_BCP47_ID)
-    assertThat(appStringId.ietfBcp47Id.ietfLanguageTag).isEmpty()
-    assertThat(appStringId.androidResourcesLanguageId.languageCode).isEmpty()
+    // TODO: Add tests for IETF_BCP47_ID type cases?
+    assertThat(appStringId.languageTypeCase).isEqualTo(LANGUAGETYPE_NOT_SET)
```
BenHenning, line 116: TODO: Verify why this & update on line 292 are correct, then document it.

Also, address TODO on 115 above.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
```diff
@@ -3,50 +3,156 @@ package org.oppia.android.util.locale
...
  * A profile to represent an Android [Locale] object which can be used to easily compare different
  * locales (based on the properties the app cares about), or reconstruct a [Locale] object.
  *
- * @property languageCode the IETF BCP 47 or ISO 639-2/3 language code
- * @property regionCode the IETF BCP 47 or ISO 3166 alpha-2 region code
+ * Subclasses of this sealed class operate on a language code and/or region code. The language code
+ * is an IETF BCP 47 or ISO 639-2/3 language code, or empty if unknown or not specified. The region
+ * code is an IETF BCP 47 or ISO 3166 alpha-2 region code, or empty if unknown or not specified.
+ *
+ * New instances should be created using [Factory].
```
BenHenning, line 16: TODO: Document this + why the profile changes were done in the first place.

Included moving Locale creation to the profile (in addition to derivation) to simplify Locale creation for the locale factory class (since there are several different ways to "correctly" create the Locale, plus there's now support for producing the root locale).

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleProfileTest.kt
```diff
@@ -298,262 +323,593 @@ class AndroidLocaleProfileTest {
...
+    val profile =
+      androidLocaleProfileFactory.createFromAndroidResourcesLanguageId(
+        languageId = LanguageId.newBuilder().apply {
+          ietfBcp47Id = IetfBcp47LanguageId.newBuilder().apply {
+            ietfLanguageTag = "pt-BR"
+          }.build()
+        }.build()
+      )
+
+    // This method can only be create a method via a valid Android resources language ID.
```
BenHenning: Fix grammar here & in other similar comments.

BenHenning (commented):
Self-reviewed to try and figure out what work is left & what needs docuemnting.

scripts/assets/file_content_validation_checks.textproto
BenHenning: Updated description.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
BenHenning: Updated PR description.

utility/src/main/java/org/oppia/android/util/data/DataProviders.kt
BenHenning: Updated PR description.

utility/src/main/java/org/oppia/android/util/locale/DisplayLocaleImpl.kt
BenHenning: Updated PR description.

utility/src/test/java/org/oppia/android/util/locale/DisplayLocaleImplTest.kt
BenHenning: Updated PR description.

domain/src/test/java/org/oppia/android/domain/locale/LocaleControllerTest.kt
BenHenning: Updated PR description.

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleFactoryTest.kt
BenHenning: Updated PR description.

utility/src/main/java/org/oppia/android/util/locale/DisplayLocaleImpl.kt
BenHenning: Updated docs in latest changes.

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleProfileTest.kt
BenHenning: Updated in latest changes.

domain/src/main/java/org/oppia/android/domain/locale/LocaleController.kt
BenHenning: Confirmed coverage.

domain/src/main/java/org/oppia/android/domain/locale/LocaleController.kt
BenHenning: Confirmed coverage.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
BenHenning: Refactored in latest changes.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
BenHenning: Confirmed coverage.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
BenHenning: Confirmed coverage.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
BenHenning: Confirmed coverage.

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
BenHenning: Added new test in latest changes & removed TODO.

BenHenning: Updated PR description.

BenHenning (commented):
Self-reviewed follow-up changes.

BenHenning:
PTAL @adhiamboperes.

utility/src/main/java/org/oppia/android/util/locale/DisplayLocaleImpl.kt
```diff
@@ -13,16 +13,21 @@ import java.util.Locale
 import java.util.Objects
 
 // TODO(#3766): Restrict to be 'internal'.
-/** Implementation of [OppiaLocale.DisplayLocale]. */
+// TODO(#3766): Restrict formattingLocale to be 'internal'.
```
adhiamboperes: I think you can remove the TODO on line 15

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
```diff
@@ -100,17 +206,48 @@ data class AndroidLocaleProfile(val languageCode: String, val regionCode: String
...
+     *
+     * This is meant to be used in cases when an [AndroidLocaleProfile] is needed to match a
+     * specific Android-compatible [Locale] (e.g. via [AndroidLocaleProfile.computeAndroidLocale])
+     * that can correctly match to specific Android app strings.
+     */
+    fun createFromAndroidResourcesLanguageId(languageId: LanguageId): AndroidLocaleProfile? {
+      val languageCode = languageId.androidResourcesLanguageId.languageCode
+      val regionCode = languageId.androidResourcesLanguageId.regionCode
+      return when {
+        !languageId.hasAndroidResourcesLanguageId() -> null
```
adhiamboperes, line 222: Shouldn't this check be made before attempting to retrieve the language and region codes in ln219 and 220?

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
```diff
@@ -100,17 +206,48 @@ data class AndroidLocaleProfile(val languageCode: String, val regionCode: String
...
       emptyRegionAsWildcard: Boolean = false
     ): AndroidLocaleProfile? {
-      return if (languageCode.isNotEmpty() && (regionCode.isNotEmpty() || emptyRegionAsWildcard)) {
-        val adjustedRegionCode = if (emptyRegionAsWildcard && regionCode.isEmpty()) {
-          REGION_WILDCARD
-        } else regionCode
-        AndroidLocaleProfile(languageCode, adjustedRegionCode)
-      } else null
+      return when {
+        languageCode.isEmpty() -> null
```
adhiamboperes, line 242: Does this qualify as an "else" case?

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
```diff
@@ -50,18 +73,17 @@ class AndroidLocaleFactory @Inject constructor(
    * - For other locale-based operations, the forced [Locale] will behave like the system's
    *   [Locale.ROOT].
    *
+   * Note that the returned [Locale] may be cached within the factory for performance reasons, so
+   * the returned value use a [Deferred] to ensure that this method can guarantee thread-safe
```
adhiamboperes:
```suggestion
   * the returned value uses a [Deferred] to ensure that this method can guarantee thread-safe
```

app/src/sharedTest/java/org/oppia/android/app/parser/HtmlParserTest.kt
```diff
@@ -852,8 +852,10 @@ class HtmlParserTest {
     }
   }
 
-  private fun createDisplayLocaleImpl(context: OppiaLocaleContext): DisplayLocaleImpl =
-    DisplayLocaleImpl(context, machineLocale, androidLocaleFactory, formatterFactory)
+  private fun createDisplayLocaleImpl(context: OppiaLocaleContext): DisplayLocaleImpl {
+    val formattingLocale = androidLocaleFactory.createOneOffAndroidLocale(context)
```
adhiamboperes, line 856: For this and other tests, are we only creating the locale synchronously because it is technically another situation where locale cannot be created in a timely manner?

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleProfileTest.kt
```diff
@@ -46,35 +51,46 @@ class AndroidLocaleProfileTest {
...
-    assertThat(profile.languageCode).isEqualTo("pt")
-    assertThat(profile.regionCode).isEqualTo("BR")
+    val languageAndRegionProfile = profile as? LanguageAndRegionProfile
+    assertThat(profile).isInstanceOf(LanguageAndRegionProfile::class.java)
+    assertThat(languageAndRegionProfile?.languageCode).isEqualTo("pt")
+    assertThat(languageAndRegionProfile?.regionCode).isEqualTo("br")
+  }
+
+  @Test
+  fun testCreateProfile_fromKenyaLocale_returnsRegionOnlyProfile() {
```
adhiamboperes, line 80: I suspect that changes in this PR also fixes the failure in https://github.com/oppia/oppia-android/issues/5046. We can now create a Profile even if we only have region code, and we nolonger use `computeIfAbsent` which caused the crash.

adhiamboperes (changes requested):
Thanks @BenHenning! I have left some thoughts.

app/src/sharedTest/java/org/oppia/android/app/parser/HtmlParserTest.kt
BenHenning: Actually, I did it this way to make things easier code-wise. More "correct" would be blocking on the async method like done for some tests, but it's logically close enough that it should be fine to use a mix.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleFactory.kt
BenHenning: Done.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
BenHenning: It doesn't need to be. Proto3 just defaults objects/strings if they aren't present, so ``languageCode`` and ``regionCode`` will be empty strings if there's no Android resource ID.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
BenHenning: I don't think so since we have to be sure ``languageCode`` is set before the subsequent cases. The alternative would be doing ``&&`` checks for the next two cases with ``languageCode.isNotEmpty``, unless I'm missing something?

utility/src/main/java/org/oppia/android/util/locale/DisplayLocaleImpl.kt
BenHenning: The first one is meant to correspond to the class, but per your comment this is confusing. I consolidated and updated the TODO.

utility/src/test/java/org/oppia/android/util/locale/AndroidLocaleProfileTest.kt
BenHenning: Ah, that's really insightful. I'll note it in the PR description and keep an eye out for the release to see if it goes away.

BenHenning (commented):
Self-reviewed the latest changes & addressed follow-up comments.

BenHenning:
Thanks @adhiamboperes! PTAL at latest.

utility/src/main/java/org/oppia/android/util/locale/AndroidLocaleProfile.kt
adhiamboperes: I see why the language code needs to be non-empty.

adhiamboperes (approved):
This LGTM, Thanks @BenHenning!

PR #5320, opened 2024-01-23
Fixes #5243: Technical Analytics Milestone 3 - New App Health Metrics

### LINKED ISSUE #5243: Implementation Request: Add new App Health Metrics

The following app health metrics should be implemented in the app;
- [x] App Onboarding Completed - When a user completes onboarding on the app for the first time.
- [x] App-reported errors - Log all errors logged via the ConsoleLogger to firebase.
- [x] In-app time spent - The time spent on the app.
- [x] Network Requests - Log all the network calls and the failed network calls.

The following fixes are still needed after a post-implementation review:
- [x] Verify why there are no Retrofit calls being logged (since they should be made from the worker that checks for platform parameters--it should actually be failing due to changes on Oppia web).
- [x] The ``complete_app_onboarding`` event is averaging 1.33 events per user per Firebase which should be impossible. It's overlogging (possibly due to screen rotations or other configuration changes).
- [x] ``feature_flag_list`` is being sent way too often. It averages 14 per session instead of 1--it only needs to be sent once per app instance since feature flag values cannot change after the app opens.
- [x] The ``feature_flag_names`` and ``feature_flag_sync_statuses`` parameters seem to be missing from the ``feature_flag_list`` event (only ``feature_flag_enabled_states`` is present).

An implementation of these logs.

N/A

### Additional context

N/A

kkmurerwa:
## Explanation

Fixes #5243: When merged, this PR will;
- Introduce the ability to log all network calls as well as failed network calls.
- Introduce the ability to log all console errors logged via the `ConsoleLogger`.
- Introduce the ability to log the time spent by a user with the app in the foreground.

## Essential Checklist

## For UI-specific PRs only


app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -96,9 +102,49 @@ class HomeFragmentPresenter @Inject constructor(
...
+            }
+            is AsyncResult.Success -> {
+              // It's possible for the observer to still be active & change due to the next activity
+              // causing a notification to be posted. That's always invalid to process here: the
+              // splash activity should never do anything after its initial state since it always
+              // finishes (or in the case of the deprecation dialog, blocks) the activity.
+              liveData.removeObserver(this)
+
+              if (startUpStateResult.value.startupMode ==
+                AppStartupState.StartupMode.USER_NOT_YET_ONBOARDED
```
adhiamboperes, line 125: It's unclear why we are logging the AppOnboardedEvent here, when the `StartupMode` is `USER_NOT_YET_ONBOARDED`.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -96,9 +102,49 @@ class HomeFragmentPresenter @Inject constructor(
...
+        override fun onChanged(startUpStateResult: AsyncResult<AppStartupState>?) {
+          when (startUpStateResult) {
+            is AsyncResult.Pending -> {
+              // Do nothing
+            }
+            is AsyncResult.Success -> {
+              // It's possible for the observer to still be active & change due to the next activity
+              // causing a notification to be posted. That's always invalid to process here: the
+              // splash activity should never do anything after its initial state since it always
+              // finishes (or in the case of the deprecation dialog, blocks) the activity.
```
adhiamboperes: This comment does not sound correct:
-  by the time we launch `HomeActivity`, the`SplashActivity` has already been destroyed.
- `HomeActivity` will never be launched if a deprecation state is encountered.
- Isn't startup state already checked in the splash activity?

data/src/main/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptor.kt
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * A flow that emits a [RetrofitCallFailedContext] when a network call fails.
+   */
+  val logFailedNetworkCallFlow: SharedFlow<RetrofitCallFailedContext> = _logFailedNetworkCallFlow
+
+  @Throws(IOException::class)
+  override fun intercept(chain: Interceptor.Chain): Response {
+    try {
+      val request = chain.request()
```
adhiamboperes: DRY: Can you lift assignment outside of try to avoild repetition below?

data/src/main/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptor.kt
```diff
@@ -0,0 +1,87 @@
...
+              .setUrlCalled(request.url.toString())
+              .setHeaders(request.headers.toString())
+              .setResponseStatusCode(response.code)
+              .setErrorMessage(responseBody ?: "")
+              .build()
+          )
+        }
+      }
+
+      return response
```
adhiamboperes: If you change above to return try, here will be `response`, instead of `return response`. i.e.
```kotlin
  return try {
  // handle response
  response
} catch(exception: Exception) {
// handle exception
  chain.proceed(request)
}
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+  private lateinit var topicService: TopicService
+
+  private val testVersionName = "1.0"
+  private val testVersionCode = 1
+
+  private val testUrl = "/"
+  private val testApiKey = "api_key"
+  private val testApiKeyValue = "api_key_value"
+  private val testResponseBody = "{\"test\": \"test\"}"
+  private val headerString = "$testApiKey: $testApiKeyValue"
```
adhiamboperes: You could make these constants in a companion object.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+
+      testCoroutineDispatchers.advanceUntilIdle()
+      failedNetworkJob.cancel()
+    }
+
+  private fun setUpTestApplicationComponent() {
+    ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
+  }
+
+  private fun setUpApplicationForContext() {
```
adhiamboperes: Could you add a comment explaining why this is needed?

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  @After
+  fun tearDown() {
+    mockWebServer.shutdown()
+  }
+
+  @Test
+  fun testLoggingInterceptor_makeCallToTopicService_logsNetworkCall() = runBlockingTest {
+    val mockWebServerUrl = mockWebServer.url(testUrl)
```
adhiamboperes: DRY: you can do this in the `@Before`

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+
+    val request = Request.Builder()
+      .url(mockWebServerUrl)
+      .addHeader(testApiKey, testApiKeyValue)
+      .build()
```
adhiamboperes: DRY: I think this could move to a function since it is the same in every test.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+    topicService = retrofit.create(TopicService::class.java)
+  }
+
+  // TODO(#89): Move this to a common test application component.
+  @Module
+  class TestNetworkModule {
+    @Provides
+    @Singleton
+    @NetworkApiKey
+    fun provideNetworkApiKey(): String = "test_api_key"
```
adhiamboperes: I haven't seen any usages of this in the test class.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+      mockWebServer.shutdown()
+
+      val failedNetworkJob = launch {
+        networkLoggingInterceptor.logFailedNetworkCallFlow.collect {
+          assertThat(it.urlCalled).isEqualTo(mockWebServerUrl.toString())
+          assertThat(it.responseStatusCode).isEqualTo(0)
+          assertThat(it.headers).contains(headerString)
+          assertThat(it.body).isEmpty()
+          assertThat(it.errorMessage).isNotEmpty()
+          assertThat(it.errorMessage).contains("Failed to connect to localhost/")
```
adhiamboperes:
```suggestion
          assertThat(it.errorMessage).contains("Failed to connect to localhost")
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+
+  private lateinit var topicService: TopicService
+
+  private val testVersionName = "1.0"
+  private val testVersionCode = 1
+
+  private val testUrl = "/"
+  private val testApiKey = "api_key"
+  private val testApiKeyValue = "api_key_value"
+  private val testResponseBody = "{\"test\": \"test\"}"
```
adhiamboperes: If we are using a random response body, then why are we mocking calls to TopicService? Should we not be checking that the call returned the value of what is expected from TopicService?

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
```diff
@@ -165,6 +172,30 @@ class ApplicationLifecycleObserver @Inject constructor(
...
+          appSessionId = sessionId,
+          foregroundTime = timeInForeground
+        ),
+        profileId = null
+      )
+    }.invokeOnCompletion { failure ->
+      if (failure != null) {
+        oppiaLogger.e(
+          "ActivityLifecycleObserver",
+          "Encountered error while trying to log app's performance metrics.",
```
adhiamboperes:
```suggestion
          "Encountered error while trying to log app's time in the foreground.",
```

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
```diff
@@ -165,6 +172,30 @@ class ApplicationLifecycleObserver @Inject constructor(
     }
   }
 
+  private fun logAppInForegroundTime() {
+    CoroutineScope(backgroundDispatcher).launch {
+      val sessionId = loggingIdentifierController.getSessionIdFlow().value
+      val installationId = loggingIdentifierController.fetchInstallationId()
+      val timeInForeground = oppiaClock.getCurrentTimeMs() - appStartTimeMillis
```
adhiamboperes, line 179: Technically speaking, since you are calling this at ` @OnLifecycleEvent(Lifecycle.Event.ON_START)` this time will be logged immediately the app comes into the foreground, meaning that it will always be 0 value. Or realistically speaking, just a few millis.

I believe the requirement is to log for how long the app has been in foreground, which would be best recorded when the app goes to the background. The time logged will be from appStartTimeMillis to current time when the app is leaving the foreground.

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
```diff
@@ -337,6 +352,78 @@ class OppiaLoggerTest {
...
+    assertThat(eventContext.retrofitCallFailedContext.headers).matches(TEST_HEADERS)
+    assertThat(eventContext.retrofitCallFailedContext.body).matches(TEST_BODY)
+    assertThat(eventContext.retrofitCallFailedContext.responseStatusCode)
+      .isEqualTo(TEST_RESPONSE_CODE)
+    assertThat(eventContext.retrofitCallFailedContext.errorMessage).matches(TEST_ERROR_LOG_MSG)
+  }
+
+  @Test
+  fun testController_createAppInForegroundTimeContext_returnsCorrectAppInForegroundTimeContext() {
+    val eventContext = oppiaLogger.createAppInForegroundTimeContext(
```
adhiamboperes, line 408: This test setup does not include any advancing time mechanisms, so there is no way to verify the actual app in foreground time. https://github.com/oppia/oppia-android/blob/develop/domain/src/test/java/org/oppia/android/domain/exploration/ExplorationActiveTimeControllerTest.kt will provide an example of how to set up advancing time in tests.

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
```diff
@@ -87,10 +92,20 @@ class OppiaLoggerTest {
...
     private const val TEST_ERROR_LOG_EXCEPTION = "test_error_log_exception"
 
+    private const val TEST_URL = "test_url"
+    private const val TEST_HEADERS = "test_headers"
+    private const val TEST_BODY = "test_body"
+    private const val TEST_RESPONSE_CODE = 200
+
+    private const val TEST_INSTALLATION_ID = "test_installation_id"
+    private const val TEST_APP_SESSION_ID = "test_app_session_id"
+    private const val TEST_FOREGROUND_TIME = 0L
```
adhiamboperes: This will always be true/pass in tests according to your current logic in lifecycleobserver. please use a real value like 3000L. TestCoroutineDispatchers has mechanisms for advancing time in tests.

model/src/main/proto/oppia_logger.proto
```diff
@@ -368,6 +384,69 @@ message EventLog {
...
+    string log_tag = 2;
+
+    // String containing the full error description.
+    string full_error_log = 3;
+  }
+
+  // Structure for the retrofit call context.
+  message RetrofitCallContext {
+    // String of the url being called.
+    string url_called = 1;
```
adhiamboperes:
```suggestion
    string request_url = 1;
```

model/src/main/proto/oppia_logger.proto
```diff
@@ -368,6 +384,69 @@ message EventLog {
...
+
+    // Headers passed as part of the retrofit call.
+    string headers = 2;
+
+    // Body of the retrofit call.
+    string body = 3;
+
+    // Response code from the call.
```
adhiamboperes:
```suggestion
    // Http response code from the call.
```
Ditto below.

model/src/main/proto/oppia_logger.proto
```diff
@@ -368,6 +384,69 @@ message EventLog {
...
+    int32 response_status_code = 4;
+  }
+
+  // Structure for the retrofit call failed context.
+  message RetrofitCallFailedContext {
+    // String of the url being called.
+    string url_called = 1;
```
adhiamboperes:
```suggestion
    string request_url = 1;
```

model/src/main/proto/oppia_logger.proto
```diff
@@ -368,6 +384,69 @@ message EventLog {
...
+
+    // Response code from the call.
+    int32 response_status_code = 4;
+
+    // The error body of the failed network call
+    string error_message = 5;
+  }
+
+  // Structure for the foreground time context.
+  message ForegroundAppTimeContext {
```
adhiamboperes: Maybe `AppInForegroundTimeContext`?

adhiamboperes (changes requested):
Thanks @kkmurerwa!

I have left some comments. PTAL.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
kkmurerwa: Once the user launches the app for the first time, the StartupMode will still be USER_NOT_YET_ONBOARDED. This will change (forever) on the next launch so it is safe enough to use it this way.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
kkmurerwa: I think this was there by mistake. I will remove it.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptor.kt
kkmurerwa: That makes sense. I will make the change.

data/src/main/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptor.kt
kkmurerwa: I have lifted the return statement out of the try catch block as suggested.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have tried removing it and the tests still pass so, as it turns out, it is not needed.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have moved the repetitive code to `@Before`.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have moved this initialization to `@Before` as well.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have removed the dependency.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: Done.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have removed references of the TopicService as it is not used in the final tests.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
kkmurerwa: Done

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
kkmurerwa: The `logAppInForegroundTime()` method is called in the `onAppInBackground()` method of the ApplicationLifecycleObserver so it should log the correct time.

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
kkmurerwa: I have modified this value to 5000L

domain/src/test/java/org/oppia/android/domain/oppialogger/OppiaLoggerTest.kt
kkmurerwa: I have added a test that does this in the `ApplicationLifecycleObserverTest` file.

model/src/main/proto/oppia_logger.proto
kkmurerwa: Done

model/src/main/proto/oppia_logger.proto
kkmurerwa: Done

model/src/main/proto/oppia_logger.proto
kkmurerwa: This was the original name. I have reverted back to it now.

model/src/main/proto/oppia_logger.proto
kkmurerwa: Done

kkmurerwa:
@adhiamboperes can you PTAL at this PR again.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+      .build()
+  }
+
+  @After
+  fun tearDown() {
+    mockWebServer.shutdown()
+  }
+
+  @Test
+  fun testLoggingInterceptor_makeCallToTopicService_logsNetworkCall() = runBlockingTest {
```
adhiamboperes: The assertions in this test are for the contents of the  network call, and not for  the event log as the name suggests.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+      }
+    }
+
+    client.newCall(request).execute()
+    testCoroutineDispatchers.advanceUntilIdle()
+    networkJob.cancel()
+  }
+
+  @Test
+  fun testLoggingInterceptor_makeFailingCallToTopicService_logsNetworkCallFailed() =
```
adhiamboperes: The test looks like it is testing something like:
```suggestion
  fun testLoggingInterceptor_makeNetworkCallWithInvalidUrl_failsAndCompletes() =
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
adhiamboperes:
```suggestion
  fun testLoggingInterceptor_makeNetworkCall_succeeds() = runBlockingTest {
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+      }
+
+      client.newCall(request).execute()
+      testCoroutineDispatchers.advanceUntilIdle()
+      networkJob.cancel()
+      failedNetworkJob.cancel()
+    }
+
+  @Test
+  fun testLoggingInterceptor_makeFailingCallToTopicService_logsNetworkCallFailed_withException() =
```
adhiamboperes:
```suggestion
  fun testLoggingInterceptor_makeFailingCall_faills withConnectionException() =
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+    client = OkHttpClient.Builder()
+      .addInterceptor(networkLoggingInterceptor)
+      .build()
+
+    // Use retrofit with the MockWebServer here instead of MockRetrofit so that we can verify that
+    // the full network request properly executes. MockRetrofit and MockWebServer perform the same
+    // request mocking in different ways and we want to verify the full request is executed here.
+    // See https://github.com/square/retrofit/issues/2340#issuecomment-302856504 for more context.
+    retrofit = Retrofit.Builder()
+      .baseUrl(mockWebServer.url(testUrl))
```
adhiamboperes, line 173: Why not:
```suggestion
      .baseUrl(testUrl)
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+    setUpTestApplicationComponent()
+    setUpRetrofit()
+
+    mockWebServerUrl = mockWebServer.url(testUrl)
+
+    request = Request.Builder()
+      .url(mockWebServerUrl)
+      .addHeader(testApiKey, testApiKeyValue)
+      .build()
+  }
```
adhiamboperes, line 79:
```suggestion
    request = Request.Builder()
      .url(testUrl)
      .addHeader(testApiKey, testApiKeyValue)
      .build()
  }
```

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -0,0 +1,240 @@
...
+      testCoroutineDispatchers.advanceUntilIdle()
+      failedNetworkJob.cancel()
+    }
+
+  private fun setUpTestApplicationComponent() {
+    ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
+  }
+
+  private fun setUpRetrofit() {
+    mockWebServer = MockWebServer()
```
adhiamboperes, line 163:
```suggestion
    mockWebServer = MockWebServer()
    mockWebServer.url(testUrl)
```

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
```diff
@@ -154,6 +161,30 @@ class ApplicationLifecycleObserver @Inject constructor(
...
+    }.invokeOnCompletion { failure ->
+      if (failure != null) {
+        oppiaLogger.e(
+          "ActivityLifecycleObserver",
+          "Encountered error while trying to log app's time in the foreground.",
+          failure
+        )
+      }
+    }
+  }
```
adhiamboperes: This seems off. There was no need to change this code at all.

adhiamboperes: Looks like this was meant for the `logAppInForegroundTime()` below.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+    val advanceTime = 10000L
+
+    val sessionIdProvider = loggingIdentifierController.getSessionId()
+    val sessionId = monitorFactory.waitForNextSuccessfulResult(sessionIdProvider)
+
+    val installationIdProvider = loggingIdentifierController.getInstallationId()
+    val installationId = monitorFactory.waitForNextSuccessfulResult(installationIdProvider)
+
+    applicationLifecycleObserver.onAppInForeground()
+    assertThat(performanceMetricsController.getIsAppInForeground()).isTrue()
```
adhiamboperes: We already have tests performing this assertion.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
     assertThat(event.currentScreen).isEqualTo(ScreenName.BACKGROUND_SCREEN)
   }
 
+  @Test
+  fun testObserver_onAppInForeground_thenInBackground_logsAppInForegroundTime() {
+    // Flaky test as the log is sometimes logged before and sometimes after the app in background
```
adhiamboperes: This is likely due to errors with the test setup.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+    // event.
+    setUpTestApplicationComponent()
+    fakeOppiaClock.setFakeTimeMode(FakeOppiaClock.FakeTimeMode.MODE_UPTIME_MILLIS)
```
adhiamboperes, line 410: To manually advance time, we need to use the `MODE_FIXED_FAKE_TIME`. This is already set in the `setUpTestApplicationComponent()` function.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+
+    applicationLifecycleObserver.onCreate()
```
adhiamboperes, line 412: You are missing a ` applicationLifecycleObserver.onAppInForeground()` call here.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+    testCoroutineDispatchers.runCurrent()
+
+    val advanceTime = 10000L
```
adhiamboperes: Make this a constant at the top of the file: `TEST_TIMESTAMP_APP_IN_FOREGROUND_MILLIS`

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+
+    val sessionIdProvider = loggingIdentifierController.getSessionId()
+    val sessionId = monitorFactory.waitForNextSuccessfulResult(sessionIdProvider)
+
+    val installationIdProvider = loggingIdentifierController.getInstallationId()
+    val installationId = monitorFactory.waitForNextSuccessfulResult(installationIdProvider)
+
+    applicationLifecycleObserver.onAppInForeground()
```
adhiamboperes, line 441: Move to before runCurrent in ln381

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +370,44 @@ class ApplicationLifecycleObserverTest {
...
+    assertThat(performanceMetricsController.getIsAppInForeground()).isTrue()
+
+    testCoroutineDispatchers.advanceTimeBy(advanceTime)
+
+    applicationLifecycleObserver.onAppInBackground()
+    testCoroutineDispatchers.runCurrent()
+
+    assertThat(performanceMetricsController.getIsAppInForeground()).isFalse()
```
adhiamboperes: Please remove as this is not the focus of this test. It is an integration test so it should fail if something is not working correctly.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
adhiamboperes: We probably need an integration with FakeEventLogger to verify that the logs were created.

utility/src/test/java/org/oppia/android/util/logging/ConsoleLoggerTest.kt
```diff
@@ -0,0 +1,163 @@
...
+
+  @Before
+  fun setUp() {
+    setUpTestApplicationComponent()
+    setUpApplicationForContext()
+  }
+
+  @Test
+  fun testConsoleLogger_logError_withMessage_logsMessage() = runBlockingTest {
+    val consoleJob = launch {
```
adhiamboperes, line 62: Where are you setting up the fake log? I would expect to see you call something like:
```kotlin
 consoleLogger.v(pass method params)
```
at the begining of the test, so that when the flow is subscribed to, we actually verify the passed params. 

Ditto the other public logging methods in the class.

utility/src/test/java/org/oppia/android/util/logging/ConsoleLoggerTest.kt
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+
+    consoleLogger.e(testTag, testMessage)
+
+    testCoroutineDispatchers.advanceUntilIdle()
+
+    consoleJob.cancel()
+  }
+
```
adhiamboperes, line 72: Can we verify that this event log has been added to the list of logs to be uploaded? (FakeEventLogger.)

adhiamboperes (changes requested):
Thanks @kkmurerwa -- this is taking shape but there are some changes needed in the new tests. PTAL.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have added them into a companion object

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: The logs can be tested in the ApplicationLifecycleObserverTest as part of the integration tests since it triggers the monitoring and logging. The NetworkLoggingInterceptor only emits the logs.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: Using the mockwebserver url allows it to intercept network calls and return the expected response.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: Done

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: Done

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: Done

utility/src/test/java/org/oppia/android/util/logging/ConsoleLoggerTest.kt
kkmurerwa: I don't get the question. But I have a `consoleLogger.e(testTag, testMessage)` method in the test that logs the error. I only log errors since the other logs are very many and would easily fill up the logs.

utility/src/test/java/org/oppia/android/util/logging/ConsoleLoggerTest.kt
kkmurerwa: I will include a seperate tests for these in the ApplicationLifecycleObserverTest to make sure they have been added to the logs.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
adhiamboperes: But the mock webserver url == test url.
mockWebserver.url("") just sets the value of the mock webserver URL,  AFAIK. So this isn't actually intercepting the network calls.

The network client does the intercepting

kkmurerwa: When I use `mockWebserver.url()`, it returns a url similar to `http://localhost:54086/` that contains a link to the local mock webserver client (and on the appropriate port). If I used the url directly, the fake network call would fail.

kkmurerwa:
@adhiamboperes I have added new integration tests for the console error logs and network calls to the ApplicationLifecycleObserverTest file in addition to fixing the other issues. PTAL

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have updated the test names. Check if they still align with Oppia naming standards.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have updated this test name.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: Done

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
adhiamboperes: Okay, sg

adhiamboperes (changes requested):
Thanks @kkmurerwa! I didn't leave any new comments, but a good number of my previous comments have not been responded to, so I could not resolve them.  PTAL.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: This won't work because the mockwebserver creates a server with a url that is not included if we use the suggested approach.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I don't think this is necessary because the `url` function only creates and returns the passed url with the server route and port number for the mockwebserver included.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
kkmurerwa: I have reverted this change.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: I have removed the unnecessary assertion.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: I have fixed the flaky test and removed the comment.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: The test fails if I do not set the time mode to MODE_UPTIME_MILLIS as the time does not advance.

kkmurerwa:
@adhiamboperes I have responded to the remaining comments. Some changes had been made and I had only forgotten to respond. PTAL

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
```diff
@@ -165,6 +172,30 @@ class ApplicationLifecycleObserver @Inject constructor(
...
+          installationId = installationId,
+          appSessionId = sessionId,
+          foregroundTime = timeInForeground
+        ),
+        profileId = null
+      )
+    }.invokeOnCompletion { failure ->
+      if (failure != null) {
+        oppiaLogger.e(
+          "ActivityLifecycleObserver",
```
adhiamboperes:
```suggestion
          "ApplicationLifecycleObserver",
```

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -369,6 +403,111 @@ class ApplicationLifecycleObserverTest {
...
+
+    testCoroutineDispatchers.advanceTimeBy(TEST_TIMESTAMP_APP_IN_FOREGROUND_MILLIS)
+    applicationLifecycleObserver.onAppInBackground()
+    testCoroutineDispatchers.runCurrent()
+
+    val eventLog = getOneOfLastTwoEventsLogged(ActivityContextCase.APP_IN_FOREGROUND_TIME)
+    val eventLogContext = eventLog.context
+
+    assertThat(eventLogContext.activityContextCase)
+      .isEqualTo(ActivityContextCase.APP_IN_FOREGROUND_TIME)
```
adhiamboperes: Please direct import ActivityContextCase.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
adhiamboperes: I have found the tests in this class to have this pattern:  `testLoggingInterceptor_makeFailingNetworkCall_emitsRetrofitCall_andCallFailedContext()` but this does not describe the expected test behaviour. I can see that all 3 tests assert the value of the network response, and none of them assert for the eventlog context.
 
 I would suggest replacing the "outcome" part of the test name with a description of what is expected, such as:
 
`testLoggingInterceptor_makeFailingNetworkCall_responseIsError`

adhiamboperes (changes requested):
Thanks @kkmurerwa! I left a couple of minor comments, but I would still like you to further change the network test names. Please see inline.

domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserver.kt
kkmurerwa: Done

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
kkmurerwa: Done

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: I have renamed the tests based on your suggestions. PTAL

kkmurerwa:
Hi @adhiamboperes. I have made the changes suggested. PTAL

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -102,7 +102,8 @@ class NetworkLoggingInterceptorTest {
   }
 
   @Test
-  fun testLoggingInterceptor_makeFailingNetworkCall_emitsRetrofitCall_andCallFailedContext() =
+//  fun testLoggingInterceptor_makeFailingNetworkCall_emitsRetrofitCall_andCallFailedContext() =
```
adhiamboperes: Please remove commented out line.

adhiamboperes (approved):
Thanks @kkmurerwa! LGTM. 

Please address the remaining comment before merge.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
kkmurerwa: Done

Issue #5070, opened 2023-06-20
[Feature Request]: Disable the submit button until an answer option is selected or an answer is typed


seanlip:
**Is your feature request related to a problem? Please describe.**

We've received some learner feedback that:
- The student did not understand that she is expected to tap on an answer option before she can proceed.
- The student did not understand that she is expected to input an answer in the input field. She was confused about not seeing the Continue button like in the previous cards.

**Describe the solution you'd like**

We plan to disable the Submit button until an answer/response is selected and provide "help text" to instruct the learner on what to do. Once the learner modifies the answer field, the Submit button should become active again.

Please see the [Mock up ](https://www.figma.com/proto/dHUz2vH8M0Tms4t0xwz3sj/Oppia-Anroid-App_HL?type=design&node-id=189-10117&scaling=min-zoom&page-id=1%3A17539&starting-point-node-id=189%3A10117&show-proto-sidebar=1) and the [Figma file](https://www.figma.com/file/dHUz2vH8M0Tms4t0xwz3sj/Oppia-Anroid-App_HL?type=design&node-id=1-17539&t=Y6WcnkfhuVIRBaCg-0) created by @Himali07, as well as [this design thread comment](https://github.com/oppia/design-team/issues/71#issuecomment-1581232000) for all the error messages.

This issue can be collaborated on by multiple contributors, by updating one interaction at a time. Here is the list of interactions:

- [ ] Number with units (e.g. 4 km)

- [x] Fraction input @masclot in https://github.com/oppia/oppia-android/pull/5224
- [x] Ratio input @theMr17 in https://github.com/oppia/oppia-android/pull/5263
- [x] Checkbox input @theMr17 https://github.com/oppia/oppia-android/pull/5319
- [x] Multiple choice input @theMr17 https://github.com/oppia/oppia-android/pull/5319
- [x] Math equation (e.g. 2x + 3y = 5) @Vishwajith-Shettigar https://github.com/oppia/oppia-android/pull/5317
- [x] Algebraic expression (e.g. 2x + 3y) @Vishwajith-Shettigar https://github.com/oppia/oppia-android/pull/5317
- [x] Number input @theMr17  https://github.com/oppia/oppia-android/pull/5310
- [x] Numeric expression (e.g. 5 + 2) @Vishwajith-Shettigar https://github.com/oppia/oppia-android/pull/5317
- [x] Text input @Vishwajith-Shettigar https://github.com/oppia/oppia-android/pull/5311
[...]

PR #5323, opened 2024-01-27
Fix part of #5070: Display empty answer message in drag & drop sort interaction

theMr17:
## Explanation

Fixes part of #5070

Enables the `submit_answer_button` when the pending answer is empty. Instead of disabling the button, an error message, stating "_**Arrange the boxes to continue.**_", is now displayed when the user attempts to submit without arranging the boxes.

https://github.com/oppia/oppia-android/assets/84731134/c0083266-8641-414a-b14d-de662123d4a2

## Essential Checklist

## For UI-specific PRs only


theMr17:
@adhiamboperes PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -23,6 +26,18 @@ import org.oppia.android.app.translation.AppLanguageResourceHandler
 import org.oppia.android.domain.translation.TranslationController
 import javax.inject.Inject
 
+/** Enum to the store the errors of drag and drop sort interaction. */
```
adhiamboperes:
```suggestion
/** Represents the type of errors that can be thrown by drag and drop sort interaction. */
```

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -97,6 +121,7 @@ class DragAndDropSortInteractionViewModel private constructor(
     // Update the data list if once drag is complete and merge icons are displayed.
     if (allowMultipleItemsInSamePosition) {
       (adapter as BindableAdapter<*>).setDataUnchecked(_choiceItems)
+      checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)
```
adhiamboperes: Is this needed? It looks as if we do not have any real time errors to report for this interaction type at the moment.

adhiamboperes: I think it should be submit time?

adhiamboperes (changes requested):
Thanks @theMr17!

PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
theMr17: > Is this needed? It looks as if we do not have an real time errors to report for this interaction type at the moment.

This is basically resetting the empty answer error message when user drags any of the boxes.

> I think it should be submit time?

This should be `REAL_TIME` because dragging of boxes is occurring in realtime, although at present real time error checking will just return null.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
theMr17: Done.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
theMr17: Making it `SUBMIT_TIME` will trigger the empty answer error message to display even without clicking on the submit button in the following situation.

https://github.com/oppia/oppia-android/assets/84731134/f5ce3f71-a28f-477a-97b7-0e6e8a950990

theMr17:
@adhiamboperes, I have addressed all the comments, PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
adhiamboperes: @theMr17, SUBMIT_TIME errors are errors that are triggered when the submit button is cliked, which is why the implementation is confusing. 

I think we need to take a step back here:
In the original implementation, the submit button is enabled by default. and the only error type available is submit time. We should keep this as it is. 

The change required in this PR is to notify the user that they need to rearrange the boxes to continue.

A few things come to mind:
1. `isAnswerAvailable` field is used to enable the submit button currently, by default. There is no need to update this.
2. we need to return  **DragAndDropSortInteractionError.EMPTY_INPUT** `if (_originalChoiceItems == _choiceItems)`, on clicking the submit button. We can handle this in the **override fun getPendingAnswer()** by calling `checkPendingAnswerError(AnswerErrorCategory.SUBMIT_TIME)` here. 
3. Your current `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is called in the onItemDragged, which is why the error is shown in realtime without clicking the submit button, because "onItemDragged" implements the actual real time dragging behaviour.

theMr17: @adhiamboperes, your solution still seems a bit unclear to me. Allow me to restate my approach for better clarity.

> 2\. We need to return **DragAndDropSortInteractionError.EMPTY_INPUT** if `_originalChoiceItems == _choiceItems` when the submit button is clicked. This can be handled in the `override fun getPendingAnswer()` by invoking `checkPendingAnswerError(AnswerErrorCategory.SUBMIT_TIME)` here.

`checkPendingAnswerError(AnswerErrorCategory.SUBMIT_TIME)` is already called in `getPendingAnswer()` in `QuestionPlayerViewModel` or `StateViewModel`, triggered when the submit button is clicked. Consequently, it returns **DragAndDropSortInteractionError.EMPTY_INPUT** or **DragAndDropSortInteractionError.VALID** as per the situation via the `getSubmitTimeError()` function.

> 3\. Your current `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is called in `onItemDragged`, which is why the error is shown in real-time without clicking the submit button. This is because "onItemDragged" implements the actual real-time dragging behavior.



(Note: The video linked in the [previous comment](https://github.com/oppia/oppia-android/pull/5323#discussion_r1489187304) is not the actual implementation. It was recorded solely to illustrate the problem that arises when `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is set to `SUBMIT_TIME` inside `onItemDragged`.)

To understand why `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is necessary in `onItemDragged`, let us consider the following scenario:

1. A question with a drag-and-drop sort interaction is given to a user.
2. The user submits the answer without altering the boxes, receiving an error message instructing them to arrange the boxes to submit the answer.
3. The user subsequently updates their answer by rearranging the boxes. However, the previous error message persists.

To address this scenario, `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is called in `onItemDragged`, returning null and removing the previous error after rearranging the boxes.

This is essentially analogous to [checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)](https://github.com/oppia/oppia-android/blob/993877939502fffee377a211142055ff38f97346/app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/RatioExpressionInputInteractionViewModel.kt#L108C9-L108C63) in `RatioExpressionInputInteractionViewModel`, which is inside `onTextChanged()`.

adhiamboperes: @theMr17, is the issue here to soley remove the previous error message once the boxes have been rearanged?

theMr17: Yes @adhiamboperes, thats why `checkPendingAnswerError(AnswerErrorCategory.REAL_TIME)` is called in `onItemDragged`, as it is triggered when the boxes are rearranged.

adhiamboperes: Sounds good.

adhiamboperes: This line should be moved out of the if condition. If you see my attached recording, it only works when `allowMultipleItemsInSamePosition` is true, as in with grouping items, but not when there is no grouping. Moving it outside the "if" means it will be called to reset the error everytime the boxes are sorted, which is what we want.
[device-2024-03-15-123308.webm](https://github.com/oppia/oppia-android/assets/59600948/a43d1035-fe1e-4bdb-acb8-979ccea073c9)

adhiamboperes (changes requested):
Thanks @theMr17. I tested this code manually and noticed a bug. PTAL.

adhiamboperes:
Please also add tests to verify that the submit button is re-enabled and the error message is hidden once a drag/sort interaction occurs.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
theMr17: Right, I have made the necessary changes.

theMr17:
@adhiamboperes PTAL. I have made the requested changes and have also added tests to verify that the submit button is re-enabled and the error message is hidden once a drag/sort interaction occurs for both grouping and non-grouping drag & drop interactions.

adhiamboperes (approved):
Thanks @theMr17! This LGTM.

Issue #4938, opened 2023-04-05
Android App - Onboarding Flow for learners and teachers/parents


tanzhirong:
**Is your feature request related to a problem?**
Profile onboarding experience is especially confusing for single-person device users. For example, users may be confused about when an administrator profile is required, and what differentiates a learner from an administrator. In addition, selection of key settings (e.g. language, audio) are currently missing from the onboarding flow.

**Describe the solution you'd like A clear and concise description of what you want to happen.**
The goal is to ensure that even without external assistance from an Oppia facilitator, users (either supervisors or learners)  (1) are well aware of the app’s value proposition, and (2) could (or in the case of supervisors, enable their students or children to) easily start his/her first lesson. These include features such as:
- Differentiating a learner from an administrator (e.g. teachers/parents) at the start of the onboarding flow
- Provide clear guidance on what the administrator does
- Allow users to set up the preferred settings (e.g. audio, language) in the onboarding process
- Ensure flow is easy to understand when a user (either the supervisor or the original learner) wants to add more learner profiles

Note that this project should be tackled in 3 parts in the following order (more details are in the "Implementation Process" section of the [PRD](https://docs.google.com/document/d/1th_V3ykAS-mUP5HjclMwO-CTxINarqgBaKBOqiGXNf4/edit#heading=h.liiqbmwmz8j8)):
- [x] Implement the onboarding flow for a learner profile.
- [x] Implement the onboarding "account type" selection.
- [ ] Implement the administrative account settings.
- [ ] Ensure compatibility with dark mode and accessibility tools

**Describe alternatives you've considered**
There are other features looking at audio/language change, but this request focuses on the onboarding flow.
Other alternatives include improved copywriting (which has been included in the proposed solution).

**Additional context Add any other context or screenshots about the feature request here.**

- PRD: https://docs.google.com/document/d/1th_V3ykAS-mUP5HjclMwO-CTxINarqgBaKBOqiGXNf4/edit#heading=h.liiqbmwmz8j8
- Mocks and Flow: https://www.figma.com/file/PnCtQKDR5VWbIlA3Ui7rTa/Oppia---Onboarding-Android-App?type=design&node-id=1446-11977&mode=design&t=6XYSCmBZqJCm6NA7-0
   - Old mocks here: https://docs.google.com/presentation/d/1cdvX9HUXuqJsV53odRM0TZlPA1EHOW3tpjK_tGpR73U/edit#slide=id.g1584efe35cd_0_270

[...]

Issue #4865, opened 2023-02-03
Refactor ProfileId provision process to multiple Activities


KevinGitonga:
### **Is your feature request related to a problem? Please describe.**
We want to provide ProfileId to all intents across the App in a pattern that will promote long term sustainability of the Oppia App by providing ProfileId in a decoupled manner. This is related to #4762.

### **Describe the solution you'd like**
Pack ProfileId into all Intents so that it can be easily accessed by underlying fragments and child classes.
https://github.com/oppia/oppia-android/pull/5248 introduced a `CurrentUserProfileIdIntentDecorator` utility to provide methods for adding the ProfileId to an intent, and for extracting it where needed. The PR also refactored a number of screens to use these utility methods.

### **Describe alternatives you've considered**
N/A

### **Additional context**
The pattern requested here was first introduced in [CurrentAppScreenNameIntentDecorator.kt](https://github.com/oppia/oppia-android/blob/4f3ca536f19da93a570f5ed9f7fbf576707ba11f/utility/src/main/java/org/oppia/android/util/logging/CurrentAppScreenNameIntentDecorator.kt) to allow the easy access by underlying child classes. 

The required decorator, `CurrentUserProfileIdIntentDecorator.kt` was added in https://github.com/oppia/oppia-android/pull/5248.

The remaining task is to refactor all screens(activities and fragments) so that the profileId is passed/retrieved entirely using this utility throughout the app.

**To pass the ProfileId to an intent or bundle, do**:
https://github.com/oppia/oppia-android/blob/8a7fe0ab737f65cc3d9a0773627d9058edadb347/app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt#L174-L180

**To retrieve ProfileId from an intent or Bundle, do**:
 https://github.com/oppia/oppia-android/blob/8a7fe0ab737f65cc3d9a0773627d9058edadb347/app/src/main/java/org/oppia/android/app/onboarding/IntroActivity.kt#L27

The best approach for this would be to make changes to a small to medium set of screens at a time, including their related tests to keep the PR sizes reasonably small. For every changed fragment, corresponding changes may be required in the test files.

### Completed packages:
[Fix part of #4865: Use profileId in classroom activity and presenter #5596](https://github.com/oppia/oppia-android/pull/5596)
[Fix part of #4865: Refactor topic to use profileId #5598](https://github.com/oppia/oppia-android/pull/5598)
[...]

PR #5387, opened 2024-04-15
Fix Part of #4938: Profile Configuration and Migration

adhiamboperes:
## Explanation
Fix Part of #4938:  Configure profile new creation and migration of existing profiles.

### Sole Learner Onboarding
Once the learner has created a profile, and have landed on the introduction screen, the profile is marked as `started_profile_oboarding`, and a corresponding event log is added.
This configuration facilitates the user resuming onboarding, if they didn't complete it the first time round.

After selecting the audio language, the profile is logged in, and routed to the home screen.
On the home screen, the profile is updated to indicate `completed_profile_oboarding` and 3 events are expected for the first time login flow, in order:
1.  Open home event
2.  Profile Onboarding Completed event
3.  App Onboarding Completed event

The  App Onboarding Completed event is only logged for the first profile on the device, while the  Profile Onboarding Completed event is logged for every profile on first login.

### Supervisor Profile Onboarding
From the Profile Type screen, the supervisor flow launches the profile chooser screen, showing the Admin profile, which on click launches the homescreen.

The same 3 events as above are expected for the first time login flow.

### Login Routing
Returning users should be routed to an appropriate landing page as follows:
A sole learner profile will be routed directly to their home screen 

Returning admin profiles and non-solo learner profiles will always be routed to the profile selection screen

A sole learner who started, but did not finish onboarding would be routed to the Introduction Screen.

Returning sole and none-sole learners created when the feature flag was disabled, would be directed to the onboarding screen. Profile migration happens in place when profiles are fetched by the controller.  

### Profile Creation and Migration
For migration purposes, the `getProfile()` function has been updated to compute the `profileType` field when the feature flag is enabled. Existing Admin profiles will be migrated to have the SUPERVISOR type, existing Admin profiles with no pins set will be migrated to have the SOLE_LEARNER type while the remaining accounts will be of the ADDITIONAL_LEARNER type. New profiles created will also contain the respective enum type based on the intended categorization. 
Flag off then on:
[device-2024-06-21-073957.webm](https://github.com/oppia/oppia-android/assets/59600948/7584ce32-305b-4743-878e-64c94dc52e66)

Flag on then off:
[device-2024-06-21-074114.webm](https://github.com/oppia/oppia-android/assets/59600948/87860084-c75c-4b6c-a37f-c2459a3e7794)

### Misc
- There are general Kdoc formatting fixes all over the place.
- `ProfileTestHelper` Has been modified to create a sole, pinless admin profile.
- There are other minor changes to fix tests that have been impacted by changes in this PR.
- The changes in `DeprecationController` are purely visual to make the code more readable.
- `ClassroomListFragmentTest` required refactor to support toggling multiple feature flags.

### Profile Onboarding Events
|Onborading Flow v1| Onboarding Flow v2|
|---|---|
|||

## Essential Checklist


app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -1892,6 +1967,77 @@ class HomeActivityTest {
...
+    setUpTestWithOnboardingV2Enabled()
+    profileTestHelper.addOnlyAdminProfileWithoutPin()
+    markSpotlightSeen(profileId)
+    launch<HomeActivity>(createHomeActivityIntent(internalProfileId)).use { scenario ->
+      pressBackUnconditionally()
+      // Pressing back should close the activity (and thus, the app) since the Sole learner has
+      // no profile chooser.
+      // Using Lifecycle.State.DESTROYED instead of activity.isFinishing because the app is already
+      // closed, therefore we cannot run `onActivity`
+      assertThat(scenario.state).isEqualTo(Lifecycle.State.DESTROYED)
```
adhiamboperes: Replace with isFinishing

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -192,6 +200,11 @@ class ProfileManagementController @Inject constructor(
     return profileDataStore.transformAsync(GET_PROFILE_PROVIDER_ID) {
       val profile = it.profilesMap[profileId.internalId]
       if (profile != null) {
+        if (enableOnboardingFlowV2.value) {
+          if (profile.profileType.equals(ProfileType.PROFILE_TYPE_UNSPECIFIED)) {
+            updateProfileType(profileId, computeProfileType(profile.isAdmin, profile.pin))
+          }
+        }
```
adhiamboperes, line 225: Rewrite more elegantly

app/src/main/java/org/oppia/android/app/home/HomeActivity.kt
```diff
@@ -106,4 +98,24 @@ class HomeActivity :
         .build()
     )
   }
+
+  override fun exitProfile(profileType: ProfileType) {
+    if (enableOnboardingFlowV2.value && profileType == ProfileType.SOLE_LEARNER) {
```
adhiamboperes: Condition should be checked before updating the isEnabled state of the OnBackPressedCallback: Ref https://stackoverflow.com/a/75462987/11396524

app/src/main/java/org/oppia/android/app/onboarding/AudioLanguageFragmentPresenter.kt
```diff
@@ -45,6 +46,14 @@ class AudioLanguageFragmentPresenter @Inject constructor(
...
       activity.finish()
     }
+
+    binding.onboardingNavigationContinue.setOnClickListener {
+      val intent =
+        HomeActivity.createHomeActivity(fragment.requireContext(), internalProfileId)
+      fragment.startActivity(intent)
+      fragment.activity?.finish()
+    }
+
```
adhiamboperes: Maybe login to profile here

app/src/main/java/org/oppia/android/app/onboarding/IntroActivityPresenter.kt
```diff
@@ -10,9 +10,12 @@ import javax.inject.Inject
 
 private const val TAG_LEARNER_INTRO_FRAGMENT = "TAG_INTRO_FRAGMENT"
 
-/** Argument key for bundling the profileId. */
+/** Argument key for bundling the profile's nickname. */
 const val PROFILE_NICKNAME_ARGUMENT_KEY = "profile_nickname"
 
+/** Argument key for bundling the profileId. */
+const val PROFILE_ID_ARGUMENT_KEY = "internal_profile_id"
+
```
adhiamboperes: Move to proto

app/src/main/java/org/oppia/android/app/onboarding/OnboardingProfileTypeFragmentPresenter.kt
```diff
@@ -34,6 +34,7 @@ class OnboardingProfileTypeFragmentPresenter @Inject constructor(
       profileTypeSupervisorNavigationCard.setOnClickListener {
         val intent = ProfileChooserActivity.createProfileChooserActivity(activity)
         fragment.startActivity(intent)
+        activity.finishAffinity()
```
adhiamboperes: Revert

adhiamboperes:
@BenHenning, PTAL. The CI checks seem stuck on this PR

BenHenning:
Hmm @adhiamboperes looking at the Bazel tests (which have better timeout cancellation than Gradle), I see at least ``SplashActivityTest`` is timing out at 5 minutes. I know that historically it takes much less time than that to run--have you tried running it locally? Changing anything with early app navigation is a bit tricky because there are a few things that need to be absolutely ordered perfectly for downstream activities to initialize properly. CI might be catching real failing tests; I'd focus on checking each of the failing Bazel tests locally to make sure they're actually passing as expected before delving too far into the Gradle tests.

I did re-run the Gradle workflows (and couldn't find anything about recent GitHub Actions issues), but if the timeout issues are real then they'll probably run into the same problem again.

model/src/main/proto/profile.proto
```diff
@@ -87,6 +87,12 @@ message Profile {
   // Represents the epoch timestamp in milliseconds when the nps survey was previously shown in
   // this profile.
   int64 survey_last_shown_timestamp_ms = 18;
+
+  // Represents the type of user which informs the configuration options available to them.
+  ProfileType profile_type = 19;
+
+  // Indicates whether this user has completed the onboarding flow.
+  bool already_onboarded_profile = 20;
```
adhiamboperes: Should be started and ended.

BenHenning:
Will need to take a pass on this a bit later since #5457 had a bunch of comments. Please make sure to address anything from #5457 that's relevant here, and then assign back.

adhiamboperes:
Resolved conflicts from upstream. Waiting for CI to pass.

adhiamboperes (commented):
Self review comments

adhiamboperes:
@BenHenning, assigning this back.

testing/src/main/java/org/oppia/android/testing/logging/EventLogSubject.kt
```diff
@@ -2400,6 +2455,36 @@ class EventLogSubject private constructor(
...
+    metadata: FailureMetadata,
+    private val actual: EventLog.ProfileOnboardingContext
+  ) : LiteProtoSubject(metadata, actual) {
+    /**
+     * Returns a [ComparableSubject] to test [EventLog.ProfileOnboardingContext.getProfileId].
+     *
+     * This method never fails since the underlying property defaults to empty string if it's not
+     * defined in the context.
+     */
+    fun hasProfileIdThat(): Subject = assertThat(actual.profileId)
```
BenHenning: Should this maybe return ``LiteProtoSubject`` instead?

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
```diff
@@ -138,6 +139,47 @@ class ProfileTestHelperTest {
...
+  @Test
+  fun testProfileOnboarding_markOnboardingCompleted_checkIsSuccessful() {
+    profileTestHelper.addOnlyAdminProfile()
+    val profileId = profileManagementController.getCurrentProfileId()
+    val onboardingProvider = profileTestHelper.markProfileOnboardingEnded(profileId!!)
+    monitorFactory.waitForNextSuccessfulResult(onboardingProvider)
+  }
+
+  @Test
+  fun testUpdateProfile_updateProfileType_checkIsSuccessful() {
```
BenHenning:
```suggestion
  fun testUpdateProfile_updateProfileType_profileTypeShouldBeUpdated() {
```

This should be more specific as to what is succeeding.

Please update all other new test names in this PR to match this level of specificity.

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
```diff
@@ -138,6 +139,47 @@ class ProfileTestHelperTest {
...
+  @Test
+  fun testProfileOnboarding_markOnboardingStarted_checkIsSuccessful() {
+    profileTestHelper.addOnlyAdminProfile()
+    val profileId = profileManagementController.getCurrentProfileId()
+    val onboardingProvider = profileTestHelper.markProfileOnboardingStarted(profileId!!)
+    monitorFactory.waitForNextSuccessfulResult(onboardingProvider)
+  }
+
+  @Test
+  fun testProfileOnboarding_markOnboardingCompleted_checkIsSuccessful() {
```
BenHenning, line 159: For this & the above: can they instead verify that the event was successfully logged rather than the provider succeeds (or better yet, have tests for both)? I think verifying both the provider result and outcome is important for consistency and correctness.

model/src/main/proto/profile.proto
```diff
@@ -93,6 +93,12 @@ message Profile {
 
   // Represents the type of user which informs the configuration options available to them.
   ProfileType profile_type = 20;
+
+  // Indicates whether this profile has started the onboarding flow.
+  bool started_profile_oboarding = 21;
+
+  // Indicates whether this profile has completed the onboarding flow.
```
BenHenning: What does started and completed mean? This is a bit difficult to grok for me because where users start and finish the onboarding flows depends on profile type, so the current framing is ambiguous and needs more specificity, I think.

model/src/main/proto/arguments.proto
```diff
@@ -257,6 +260,9 @@ message ReadingTextSizeFragmentStateBundle {
 message AudioLanguageActivityParams {
   // The default audio language previously selected by the user (upon opening the activity).
   AudioLanguage audio_language = 1;
+
+  // The internal Id associated with the profile.
+  int32 profile_id = 2;
```
BenHenning: Here & for the other new fields, prefer using ``ProfileId`` to avoid adding more work to #4865.

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
```diff
@@ -149,17 +160,23 @@ class ClassroomListFragmentPresenter @Inject constructor(
...
+        ) {
+        }
       }
     )
 
-    logAppOnboardedEvent()
+    if (enableOnboardingFlowV2.value) {
+      subscribeToProfileResult(profileId)
+    } else {
+      logAppOnboardedEvent(profileId)
```
BenHenning: We want to log this for all cases where the user successfully finishes the onboarding flow (there's a bug with it currently whereby it logs for every home screen, and we're looking into it).

It acts as a top-level app health metric, so please make sure to account for it with the new onboarding changes.

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
```diff
@@ -297,12 +314,71 @@ class ClassroomListFragmentPresenter @Inject constructor(
...
+    if (!profile.completedProfileOboarding) {
+      markProfileOnboardingEnded(profileId)
+      if (profile.profileType == ProfileType.SOLE_LEARNER ||
+        profile.profileType == ProfileType.SUPERVISOR
+      ) {
+        appStartupStateController.markOnboardingFlowCompleted()
+        logAppOnboardedEvent(profileId)
+      }
+    }
+  }
```
BenHenning: I'm quite uncomfortable, in general, with logging events in the app layer because its lifecycle is incredibly complex. It's very easily to accidentally duplicate logs (e.g. due to activity recreation, etc.). Can we log this in the domain layer, instead?

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
```diff
@@ -297,12 +314,71 @@ class ClassroomListFragmentPresenter @Inject constructor(
...
+
   private fun logHomeActivityEvent() {
     analyticsController.logImportantEvent(
       oppiaLogger.createOpenHomeContext(),
       profileId
     )
   }
+
+  private fun handleBackPress(profileType: ProfileType) {
+    activity.onBackPressedDispatcher.addCallback(
```
BenHenning: If the profile provider changes multiple times, it will lead to multiple callbacks being registered. Does this code correctly handle that scenario?

app/src/main/java/org/oppia/android/app/home/ExitProfileListener.kt
```diff
@@ -0,0 +1,13 @@
+package org.oppia.android.app.home
+
+import org.oppia.android.app.model.ProfileType
+/** Listener for when a user wishes to exit their profile. */
```
BenHenning, line 5: Nit: empty line above this please.

app/src/main/java/org/oppia/android/app/home/ExitProfileListener.kt
```diff
@@ -0,0 +1,13 @@
...
+interface ExitProfileListener {
+  /**
+   * Called when back press is clicked on the HomeScreen.
+   *
+   * A SOLE_LEARNER exits the app completely while other [ProfileType]s are routed to the
+   * [ProfileChooserActivity].
```
BenHenning: Suggest instead saying that routing behavior may change based on the profile type rather than specifying the actual implementation behavior (which seems a bit too specific at the interface level).

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -142,6 +155,52 @@ class HomeFragmentPresenter @Inject constructor(
...
+  private fun handleProfileOnboardingState(profile: Profile) {
+    // App onboarding is completed by the first profile on the app(SOLE_LEARNER or SUPERVISOR),
+    // while profile onboarding is completed by each profile.
+    if (!profile.completedProfileOboarding) {
+      markProfileOnboardingEnded(profileId)
+      if (profile.profileType == ProfileType.SOLE_LEARNER ||
+        profile.profileType == ProfileType.SUPERVISOR
+      ) {
+        appStartupStateController.markOnboardingFlowCompleted()
+        logAppOnboardedEvent(profileId)
```
BenHenning: Ditto here--we have multiple places where events need to be logged correctly in the UI layer. It seems *really* hard to verify this is actually correct--logging at the domain layer will likely be much easier to validate as correct and keep correct.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
```diff
@@ -209,4 +268,15 @@ class HomeFragmentPresenter @Inject constructor(
       ProfileId.newBuilder().apply { internalId = internalProfileId }.build()
     )
   }
+
+  private fun handleBackPress(profileType: ProfileType) {
+    activity.onBackPressedDispatcher.addCallback(
```
BenHenning, line 220: Ditto here--callback can be added multiple times.

app/src/main/java/org/oppia/android/app/onboarding/AudioLanguageFragmentPresenter.kt
```diff
@@ -158,6 +159,30 @@ class AudioLanguageFragmentPresenter @Inject constructor(
       }
   }
 
+  private fun loginToProfile(profileId: ProfileId) {
```
BenHenning:
```suggestion
  private fun logInToProfile(profileId: ProfileId) {
```

Nit: should be verb form.

app/src/main/java/org/oppia/android/app/onboarding/IntroFragmentPresenter.kt
```diff
@@ -67,4 +75,13 @@ class IntroFragmentPresenter @Inject constructor(
         R.string.onboarding_learner_intro_activity_text, profileName
       )
   }
+
+  private fun markProfileOnboardingStarted(profileId: ProfileId) {
+    profileManagementController.markProfileOnboardingStarted(profileId)
```
BenHenning: Ditto for this event: prefer logging in the domain layer, or via a UI call to ensure the user actually saw the correct flow prior to logging the event (though this call needs to be idempotent to be resilient against multiple calls which is possible in the UI layer).

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -252,8 +252,7 @@ class OnboardingFragmentPresenter @Inject constructor(
 
   private fun createDefaultProfile() {
     profileManagementController.addProfile(
-      name = "Admin", // TODO(#4938): Refactor to empty name once proper admin profile creation flow
```
BenHenning, line 254: Does  this need to be contingent based on whether the new onboarding flow flag is enabled?

app/src/main/java/org/oppia/android/app/onboarding/OnboardingProfileTypeFragmentPresenter.kt
```diff
@@ -66,4 +89,13 @@ class OnboardingProfileTypeFragmentPresenter @Inject constructor(
 
     return binding.root
   }
+
+  private fun markProfileOnboardingStarted(profileId: ProfileId) {
+    profileManagementController.markProfileOnboardingStarted(profileId)
+
+    analyticsController.logLowPriorityEvent(
```
BenHenning: Ditto for this event.

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivityPresenter.kt
```diff
@@ -3,27 +3,45 @@ package org.oppia.android.app.profile
...
+      profileManagementController.updateNewProfileDetails(
+        profileId = profileId,
+        profileType = profileType,
+        newName = "Admin",
+        avatarImagePath = null,
+        colorRgb = -10710042,
+        isAdmin = true
+      )
+    } else {
+      // TODO(#482): Ensures that an admin profile is present. Remove when there is proper admin account creation.
```
BenHenning: I think we can reframe this as: 'Remove when the new onboarding flow can be finalized as that properly creates an admin profile.' Your work *is* the solution desired by this comment.

app/src/main/java/org/oppia/android/app/profile/ProfileChooserFragmentPresenter.kt
```diff
@@ -267,4 +254,54 @@ class ProfileChooserFragmentPresenter @Inject constructor(
...
+    val intent = IntroActivity.createIntroActivity(activity)
+    intent.apply {
+      putProtoExtra(IntroActivity.PARAMS_KEY, introActivityParams)
+      decorateWithUserProfileId(profileId)
+    }
+
+    activity.startActivity(intent)
+  }
+
+  private fun loginToProfile(profile: Profile) {
```
BenHenning: Nit: ``logIn`` here, as well.

BenHenning (changes requested):
Thanks @adhiamboperes! I reviewed a bit over half of the PR. Will take a broader pass on the next round since there were a bunch of comments that would be nice to address, first.

Edit: Also, it seems that CI is failing--PTAL.

testing/src/main/java/org/oppia/android/testing/logging/EventLogSubject.kt
adhiamboperes: Yes, updated

model/src/main/proto/profile.proto
adhiamboperes: started - Indicates that the profile has viewed the relevant onboarding introduction screen.

completed - Indicates that the profile has reached the home screen for the first time.

I have updated the proto field comments.

model/src/main/proto/arguments.proto
adhiamboperes: Cleaned these up in favor of the decorator as per #4865.

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
adhiamboperes: `subscribeToProfileResult()` > `processProfileResult()` > `handleProfileOnboardingState(profile: Profile)`  ensures that all users who complete the new onboarding flow log a `started_profile_onboarding` and `completed_profile_onboarding` event.  At the same time, the first profile on the app(supervisor or sole learner) will additionally log the app onboarding complete event(existing event).

In this PR, I have retained the legacy onboarding flow to only log the app onboarding event.

I implemented this per TDD, although it is different from your comment(sounds like we want all profiles to log all 3 events?). Could you please clarify the intended behaviour?

adhiamboperes: Seperately, I experienced the bug you mention, and given some context I may be able to pick up debugging from my prior attempt at solving the https://github.com/oppia/oppia-android/issues/5434 (closed due to reverting adjacent changes that may have triggered the bug for me).

app/src/main/java/org/oppia/android/app/home/ExitProfileListener.kt
adhiamboperes: Fixed

app/src/main/java/org/oppia/android/app/home/ExitProfileListener.kt
adhiamboperes: Updated.

app/src/main/java/org/oppia/android/app/onboarding/AudioLanguageFragmentPresenter.kt
adhiamboperes: Refactored as suggested.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: The onboarding flow version has been gated at fragment level, such that we have presenter v1, and this, v2. Please see: https://github.com/oppia/oppia-android/blob/4730edb0263135e88dab08ab0f6dc724a1585a67/app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragment.kt#L36-L40

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivityPresenter.kt
adhiamboperes: Refactored

app/src/main/java/org/oppia/android/app/profile/ProfileChooserFragmentPresenter.kt
adhiamboperes: Updated.

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
adhiamboperes: As per the discussion in the meeting, I left the implementation of the logAppOnboardedEvent as is.

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
adhiamboperes: I have moved this to the controller.

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
adhiamboperes: Moved profile onboarding event to the controller and kept the implementation of the logAppOnboardedEvent.

app/src/main/java/org/oppia/android/app/onboarding/IntroFragmentPresenter.kt
adhiamboperes: Fixed

app/src/main/java/org/oppia/android/app/onboarding/OnboardingProfileTypeFragmentPresenter.kt
adhiamboperes: Refactored

app/src/main/java/org/oppia/android/app/home/HomeFragmentPresenter.kt
adhiamboperes: Refactored the code handling profile update and the callback.

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
adhiamboperes: Refactored function to remove this callback from any OnBackPressedDispatcher it is currently added to after every call.

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
adhiamboperes: Updated.

adhiamboperes:
@BenHenning, PTAL.

app/src/main/java/org/oppia/android/app/splash/SplashActivityPresenter.kt
```diff
@@ -265,36 +279,142 @@ class SplashActivityPresenter @Inject constructor(
...
+
+    val intent = IntroActivity.createIntroActivity(activity).apply {
+      putProtoExtra(PARAMS_KEY, introActivityParams)
+      decorateWithUserProfileId(profileId)
+    }
+
+    activity.startActivity(intent)
+  }
+
+  private fun loginToProfile(profileId: ProfileId) {
```
BenHenning:
```suggestion
  private fun logInToProfile(profileId: ProfileId) {
```

Nit for consistency & correct verb tense. Ditto elsewhere in the PR.

app/src/sharedTest/java/org/oppia/android/app/splash/SplashActivityTest.kt
```diff
@@ -1049,6 +1058,72 @@ class SplashActivityTest {
...
+
+    launchSplashActivityPartially {
+      intended(hasComponent(IntroActivity::class.java.name))
+      intended(hasProtoExtra(IntroActivity.PARAMS_KEY, params))
+      intended(hasProtoExtra(PROFILE_ID_INTENT_DECORATOR, profileId))
+    }
+  }
+
+  @Test
+  @RunOn(TestPlatform.ESPRESSO)
```
BenHenning: Why can't this run on Robolectric?

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -209,6 +218,11 @@ class ProfileManagementController @Inject constructor(
     return profileDataStore.transformAsync(GET_PROFILE_PROVIDER_ID) {
       val profile = it.profilesMap[profileId.internalId]
       if (profile != null) {
+        if (enableOnboardingFlowV2.value) {
```
BenHenning, line 221: Are there tests verifying this migration pathway and do they cover all potential scenarios? It seems they do, but I wanted to double check since it's essential that we get this part right (given it will affect all existing users of the app).

app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt
BenHenning: Ah, yes I think keeping the existing behavior is reasonable since fixing an existing event is outside the scope of this PR.

BenHenning (commented):
Thanks @adhiamboperes! Just had a few small comments left, otherwise the PR LGTM.

app/src/main/java/org/oppia/android/app/splash/SplashActivityPresenter.kt
adhiamboperes: Updated -- other occurances previously updated in profileChooser and AudioLanguageFragment.

app/src/sharedTest/java/org/oppia/android/app/splash/SplashActivityTest.kt
adhiamboperes: There seems to be a synchronizing issue with this test. I have updated the test code itself:
```kotlin
 @Test
  @RunOn(TestPlatform.ESPRESSO)
  fun testSplashActivity_onboardingV2Enabled_onboardedSoleLearnerProfile_routesToHomeActivity() {
    simulateAppAlreadyOnboarded()
    initializeTestApplication(onboardingV2Enabled = true)
    profileTestHelper.addOnlyAdminProfileWithoutPin()

    val profileId = ProfileId.newBuilder().setInternalId(0).build()
   monitorFactory.ensureDataProviderExecutes(
      profileTestHelper.updateProfileType(profileId, ProfileType.SOLE_LEARNER)
    )
    monitorFactory.ensureDataProviderExecutes(
      profileTestHelper.markProfileOnboardingStarted(profileId)
    )
    monitorFactory.ensureDataProviderExecutes(
      profileTestHelper.markProfileOnboardingEnded(profileId)
    )
    testCoroutineDispatchers.advanceUntilIdle()

    launchSplashActivityPartially{
      intended(hasComponent(HomeActivity::class.java.name))
    }
  }
```

`profileTestHelper.markProfileOnboardingEnded(profileId)` never executes, causing the test to run indefinitely due to never entering an idle state. 
On Espresso, this passes but ocassionally fails with:

```
Wanted but not invoked:
observer.onChanged(<Capturing argument>);
-> at org.oppia.android.testing.data.DataProviderTestMonitor.ensureNextResultIsPresent(DataProviderTestMonitor.kt:117)
Actually, there were zero interactions with this mock.

at org.oppia.android.testing.data.DataProviderTestMonitor.ensureNextResultIsPresent(DataProviderTestMonitor.kt:117)
at org.oppia.android.testing.data.DataProviderTestMonitor.waitForNextResult(DataProviderTestMonitor.kt:59)
```

For any random DataProvider in the test.

adhiamboperes: This turned out to be a dataprovider update loop, and I have fixed it.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
adhiamboperes: You're correct, these scenarios were implicitly tested, but I have added more explicit tests for them.

testing/src/main/java/org/oppia/android/testing/profile/ProfileTestHelper.kt
```diff
@@ -70,14 +70,13 @@ class ProfileTestHelper @Inject constructor(
    *
    * @returns the [AsyncResult] designating the result of attempting to log into the admin profile
```
BenHenning: It seems the KDoc needs to be updated with the recent changes.

app/src/sharedTest/java/org/oppia/android/app/splash/SplashActivityTest.kt
BenHenning: I am actually very interested in what the specific problems were, and if this is something we can try to document on the data providers wiki page. Let's discuss further.

BenHenning (approved):
Thanks @adhiamboperes! There just seems to be one small comment left. Feel free to address and resolve it on your own before submitting. I can take another pass if you need me to, but I don't expect that to be necessary.

BenHenning:
It looks like ``SplashActivityTest`` is timing out on Bazel--maybe on Gradle, too (hard to tell with the results since the Gradle tests are behaving oddly). I'm going ahead and updating the branch so that we can get fresh results.

testing/src/main/java/org/oppia/android/testing/profile/ProfileTestHelper.kt
adhiamboperes: This is done.

adhiamboperes:
@BenHenning, thanks for reviewing! I have adressed the remaining comment, and also added the comment on the unobserved dataprovider as discussed in the meeting. The tests pass, and I am proceeding with merging the PR.

Issue #5379, opened 2024-04-05
[BUG]: TestCoroutineDispatchers time out sometimes in large Espresso Tests[[Blocked on #4991]]


adhiamboperes:
### Describe the bug

`StateFragmentTest` and `ExplorationActivityTest`, do not run in espresso, with the failure:

```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: androidx.test.espresso.IdlingResourceTimeoutException: Wait for [TestCoroutineDispatcherIdlingResource] to become idle timed out
at androidx.test.espresso.Espresso.onIdle(Espresso.java:342)
at androidx.test.espresso.Espresso.onIdle(Espresso.java:361)
at org.oppia.android.testing.threading.TestCoroutineDispatchersEspressoImpl.advanceUntilIdle(TestCoroutineDispatchersEspressoImpl.kt:45)
at org.oppia.android.testing.threading.TestCoroutineDispatchersEspressoImpl.runCurrent(TestCoroutineDispatchersEspressoImpl.kt:34)
at org.oppia.android.app.player.exploration.ExplorationActivityTest.testExplorationActivity_loadingAudio_progressbarIsDisplayed(ExplorationActivityTest.kt:1321)
...
androidx.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:527)
at org.oppia.android.testing.OppiaTestRule$apply$1.evaluate(OppiaTestRule.kt:47)
at org.oppia.android.testing.junit.InitializeDefaultLocaleRule$apply$1.evaluate(InitializeDefaultLocaleRule.kt:53)
...
Caused by: java.util.concurrent.ExecutionException: androidx.test.espresso.IdlingResourceTimeoutException: Wait for [TestCoroutineDispatcherIdlingResource] to become idle timed out
at java.util.concurrent.FutureTask.report(FutureTask.java:123)
at java.util.concurrent.FutureTask.get(FutureTask.java:193)
at androidx.test.espresso.Espresso.onIdle(Espresso.java:334)
```

### Steps To Reproduce

- Checkout develop
- Select and run a test from the `StateFragmentTest` or `ExplorationActivityTest` in Espresso(Aside from the default `verifyScreenNameInIntent` test)

### Expected Behavior

Test builds, runs and passes.

### Screenshots/Videos

There are a variety of errors, but the most important one is:
[...]

PR #5388, opened 2024-04-17
Fix #4652 : Implemented Learner Progress for Exploration State

### LINKED ISSUE #4652: Show learner progress while playing through a lesson
### Discussed in https://github.com/oppia/oppia-android/discussions/4646

<div type='discussions-op-text'>

<sup>Originally posted by **Ronnie-Mercy** October 12, 2022</sup>
I'd like to share an idea I came up with while using the Oppia app recently. Simple pagination below the continue/submit button will help the user know how many more lessons they have left and also a clear indication of what page they are on currently. I added the screen I'm talking about and also
image
 a simple mockup below that can be improved. </div>

Rd4dev:
Fixes #4652

## Explanation

- This PR provides an implementation for displaying learning progress in the exploration lessons.
- As mentioned [in the issue](https://github.com/oppia/oppia-android/issues/4652#issuecomment-1314591992), Oppia lessons are dynamic in nature, and to define the progress, the flow of lessons is mandatory. 
  - One way we could fix this is by providing order numbers to the JSON files themselves, but that would make it difficult to scale and add other states in the future. 
  - So, I implemented a small parser that identifies the order of the lessons by going through the JSON.

Screenshot (450)

- The way the parser works is as follows: 
  - It starts off with **init_state_name**. 
  - With each state, it first identifies its **default_outcome** and assigns a position to them. 
  - Then, it checks the **answers_group** from bottom to top and assigns those states their positions one by one (note that only the first entries are noted). 
  - Thereby, the order of the exploration lessons is obtained. I have attached the pseudocode for reference and the obtained list of **statename** : **positions** below.

exploration_pseudo_code

- The way it handles the dynamic changes of questions is that the total number of questions remains the same as the total number of states. The current state is checked on every new exploration. Let's say we have 3 states; 
  - answering state 1 correctly will skip you to state 3, making the progress jump from 10% to 30%. 
  - If state 1 is answered incorrectly, the result_answers are checked, and state 2 is triggered, making the progress go from 10% to 20%. 
  - The same applies to the decrement of progress when recapping back to previous states.
- The progress bar uses the same design as the question player progress bar.

## TODO: 
- runBlocking should be properly checked and handled if needed, replaced with proper coroutine. 
- Tests will be updated soon. -[Hindered Implementation due to Bug #5379 
- Check landscape implementation
- To check visibility and proper positioning with audio block
## Note: 
- This is a proposed solution. I would love to hear feedback on whether this would be viable or not.

## Essential Checklist

## Demonstration
https://github.com/oppia/oppia-android/assets/122200035/1ac1b67a-6ab0-4b81-9511-23759a86b244

### DIFF
--- a/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt
+++ b/domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt
@@ -33,6 +33,11 @@ class ExplorationRetrieverImpl @Inject constructor(
     }
   }
 
+  override suspend fun loadExplorationPosition(explorationId: String): Pair<Int, MutableMap<String, Int>> {
+    val explorationObject = jsonAssetRetriever.loadJsonFromAsset("$explorationId.json")
+    return parseJsonObjectToCalculatePosition(explorationObject!!)
+  }
+
   // Returns an exploration given an assetName
   private fun loadExplorationFromAsset(explorationObject: JSONObject): Exploration {
     val innerExplorationObject = explorationObject.getJSONObject("exploration")
@@ -67,5 +72,42 @@ class ExplorationRetrieverImpl @Inject constructor(
       statesMap[key] = stateRetriever.createStateFromJson(key, statesJsonObject.getJSONObject(key))
     }
     return statesMap
+  }
+
+  private fun parseJsonObjectToCalculatePosition(explorationObject: JSONObject): Pair<Int, MutableMap<String, Int>> {
+    val states = explorationObject.getJSONObject("exploration").getJSONObject("states")
+    val initialState = explorationObject.getJSONObject("exploration").getString("init_state_name")
+
+    val statePosition = mutableMapOf<String, Int>()
+    var position = 0
+
+    // Function to recursively traverse the states and set their positions
+    fun getStatePosition(stateName: String) {
+      if (!statePosition.containsKey(stateName)) {
+        statePosition[stateName] = ++position
+
+        val state = states.getJSONObject(stateName)
+        val interaction = state.getJSONObject("interaction")
+
+        if (interaction.has("default_outcome") && !interaction.isNull("default_outcome")) {
+          val defaultOutcome = interaction.getJSONObject("default_outcome")
+          val dest = defaultOutcome.getString("dest")
+          getStatePosition(dest)
+        }
+
+        if (interaction.has("answer_groups") && interaction.getJSONArray("answer_groups").length() > 0) {
+          val answerGroups = interaction.getJSONArray("answer_groups")
+          for (i in answerGroups.length() - 1 downTo 0) {
+            val outcome = answerGroups.getJSONObject(i).getJSONObject("outcome")
+            val dest = outcome.getString("dest")
+            getStatePosition(dest)
+          }
+        }
+      }
+    }
+
+    getStatePosition(initialState)
+    return position to statePosition
+
   }
 }
--- a/app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
+++ b/app/src/main/java/org/oppia/android/app/player/state/StateFragmentPresenter.kt
@@ -42,6 +42,7 @@ import org.oppia.android.app.translation.AppLanguageResourceHandler
 import org.oppia.android.app.utility.SplitScreenManager
 import org.oppia.android.app.utility.lifecycle.LifecycleSafeTimerFactory
 import org.oppia.android.databinding.StateFragmentBinding
+import org.oppia.android.domain.exploration.ExplorationRetriever
 import org.oppia.android.domain.exploration.ExplorationProgressController
 import org.oppia.android.domain.oppialogger.OppiaLogger
 import org.oppia.android.domain.survey.SurveyGatingController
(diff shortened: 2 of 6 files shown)


Rd4dev:
@BenHenning, @adhiamboperes Would this approach be effective?

Rd4dev:
@adhiamboperes PTAL

adhiamboperes (commented):
Thanks @Rd4dev! This approach is really intersting. I'm handing over to @BenHenning to verify the approach, and @seanlip to look over the progressbar design for Product. 

As a workaround to the Espresso run issues, I usually modify the test setup to  unregister IdlingResource immediately after the profiles have been setup in the `setUpTest()` Function. This should allow you to write tests.

seanlip:
THis is quite interesting, thanks @Rd4dev!

From the product side, I don't have an objection, it would actually not be a bad idea to test this. The only request I have is to put this functionality behind a feature flag so that we can do some controlled testing before launching it more broadly.

The only other concern I have is that it's a bit weird how the content gets "sucked up" into the progress bar -- I'm not sure about the handling of the cutoff when scrolling. Would you mind filing an issue on the [design team repo](https://github.com/oppia/design-team/issues) to take a look at this, and I will talk to the Android design team to see if they can look at it? This doesn't need to block the merge of this PR though (assuming that the functionality remains gated behind a feature flag).

Thanks!

Rd4dev:
Thank you for the initial approval, @seanlip, I've filed a new issue in the [design team repository](https://github.com/oppia/design-team/issues/149). 
Will reserve the feature under a feature flag in the upcoming commits.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt
```diff
@@ -68,4 +73,41 @@ class ExplorationRetrieverImpl @Inject constructor(
     }
     return statesMap
   }
+
+  private fun parseJsonObjectToCalculatePosition(explorationObject: JSONObject): Pair<Int, MutableMap<String, Int>> {
```
BenHenning, line 77: @seanlip has anyone worked on the in-lesson progress bar on web? I'm wondering if anyone has already come up with an algorithm for computing the linear path through the lesson that we could copy here for parity.

@Rd4dev one potential issue I see with this approach is that it's sort of implicitly assuming that each destination is an equal outcome when making progress. So long as we never let the user go backwards in progress, that helps a bit with making sure the progress bar looks correct, but it may be a bit unpredictable since sometimes states with many outcomes will show "more progress" of completing than those with few.

I think the intended way to implement this is to use a pathfinding algorithm from the starting state to all ending states. There should be only one path (but we'll need some sort of tiebreaker if there is more than one). From there, we only count progress for states who have checkpoints marked (this is a property on Oppia web we don't currently import: https://github.com/oppia/oppia/blob/5ebbb3d369532a22ee1638a810d597ffb9719594/core/domain/state_domain.py#L3615).

@seanlip does the algorithm explained in the above paragraph sound correct to you (in the event there isn't an existing implementation to follow)? Also, is the checkpoint field being set in lessons now?

BenHenning (changes requested):
Thanks @Rd4dev! I moved this over to draft because I think it has a lot of iteration needed before it's actually ready to be finalized and go into full review. I left one main comment to analyze the algorithm first--once we finalize that, I'll have more comments on how to model the dataflow and API changes. After that's done, the final steps are focusing on the UI and testing side of things.

This PR will probably require a lot of back-and-forths to finalize given the nature of this problem: modeling progress is complex in Oppia lessons since they have non-linear paths that are potentially unique to an individual's learning experience. We've fortunately thought about this in some detail, but if you're unsure whether you'll have the time to finish this PR let me know and we can move the discussion for the algorithm over to the issue for whoever works on it next.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt
seanlip: @BenHenning Yes, we have done a bit of work here. We label some states as checkpoints, and use those checkpoints in the lessons as a measure of progress. The checkpoint states are always located at articulation points of the graph (i.e. if you remove the checkpoint state then it disconnects the graph). Lessons have 2-8 checkpoints and progress is measured by the achievement of checkpoints. More granular progress is not recorded.

The checkpoint field is being set in lessons, and the third paragraph you mention seems like a good approach to follow. It does match what we do on Web. You don't need a tiebreaker, given the "articulation points" property above.

Please let me know if you need more info. Also, I apologize for entirely missing this in my previous comment -- for some reason I didn't make the connection between the progress bar and the existing checkpoints functionality on Web.

BenHenning:
@Rd4dev are you planning on working on this? Otherwise, I think it might be best if we close it and unassign the issue so that others can take it up if they have availability.

It sounds based on Sean's latest feedback that we'd need to update the import asset script & domain protos for lessons to pull in the checkpoint properties so that we have a baseline for computing progress through an exploration.

Rd4dev:
@BenHenning I don't think I will be able to work on this right now. I was initially looking into pathfinding algorithm implementations, but after Sean mentioned the necessary asset updates, I think it's best to unassign myself for now. I'll let someone else take over, and if it's still available later, I'll pick it up when I can. For now, I'll unassign myself. Thanks for the ping.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationRetrieverImpl.kt
BenHenning: This approach sounds good and we should be able to utilize the new field.

PR #5400, opened 2024-05-21
Fix #1535, part of #4120: Upgrade to rules_kotlin v1.5.0 beta 3

BenHenning:
## Explanation
Fixes #1535
Fixes part of #4120

This PR preps for the codebase-wide migration to Kotlin 1.6 (done in #4937) by first upgrading rules_kotlin from 1.5.0 alpha 2 to 1.5.0 beta 3. This upgrade, while small, produces a few nice benefits:
- It removes the need for extra WORKSPACE dependency setup (thus addressing #1535).
- It moves a bunch of rules_kotlin macros to new locations which results in changing nearly every ``BUILD.bazel`` file in the codebase.

The changes are straightforward to review in isolation, so this PR acts as a means to help reduce the complexity of #4937 by pulling ahead these groups of ``BUILD.bazel`` changes while also moving rules_kotlin the smallest number of versions (ahead of the more major upgrades which will occur downstream).

Note that this change is needed because:
- We'll need to upgrade to a newer version of rules_kotlin in order to bring in Kotlin 1.6 support.
- rules_kotlin moved these macros and using the old ones results in a _lot_ of build warnings that spam the local console when trying to build.

Separately, this PR includes some minor trailing space cleanup in wiki/Oppia-Bazel-Setup-Instructions.md that was noticed when editing the file (since my local development environment auto-strips trailing spaces).

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
N/A -- This should only affect build infrastructure, and barely impact resulting binary builds.

PR #5222, opened 2023-11-03
Fix #5137: Upgrade builds to target SDK 33

BenHenning:
## Explanation
Fixes #5137
Partially mitigates #5233

This updates all build & target SDKs for Gradle & Bazel builds to now target SDK 33, per the new Play Store mandate (see https://support.google.com/googleplay/android-developer/answer/11926878?hl=en). The analysis of SDK 33 features, potential issues, and potential mitigations are described in #5137. It was determined that there are no obvious code changes needed beyond the minimum to target SDK 33. We'll be relying on some platform-level regression testing plus the Play Console's pre-submit app analysis report for finalizing the go/no-go decision for SDK 33 support.

Testing consisted of manually playing through the app and seeing if there were any issues using emulators both with SDK 33 and lower versions. #5137 documents the issues found and their ultimate causes. Since no functionality changes were needed for SDK 33, no tests needed updating. Separately, tests are still running on SDK 30 due to being stuck (see #4748).

One issue unrelated to SDK 33 was observed: when playing through the prototype lesson, I once again brought the app into a broken state wherein I couldn't leave the lesson via navigation. We've hit this issue a few times, but still don't know the root cause. #5233 was filed and a mitigation was introduced in this PR (+ a test for it): if the exploration player observes a failure when trying to stop the player session (for whatever reason), it still proceeds with its "end of activity" flow (e.g. finishing the activity). This provides at least an escape hatch for users who somehow hit this state. Note that this isn't a proper fix for #5233 given the lack of a known root cause, so this is only considered a mitigation.

The new test was verified as failing if the mitigation is omitted:

image

Note that ``ExplorationActivityTest``'s setup was tweaked to change the parent screen since this affects back navigation routing. I opted to using lessons tab as the parent screen since it's the most common route for users to hit, so it makes sense for our tests to default to that if they don't specifically need to use a different route (and none of the existing tests seemed to have an issue with this change).

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
[...]

Issue #3616, opened 2021-07-30
Migrate all non-test Android API deprecated references to the replacement versions with proper gating logic


yashraj-01:
As per: https://github.com/oppia/oppia-android/pull/3512#discussion_r679674825

We should update `NetworkConnectionUtilProdImpl` to use the correct methods to fetch connection status according to the SDK version.

1. Update `getCurrentConnectionStatus()` to switch logic based on Android SDK version:

- `if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) else {}`
- Add helper method for newer devices and older devices. Refer to [ConnectivityManager.getNetworkCapabilities()](https://developer.android.com/reference/android/net/ConnectivityManager#getNetworkCapabilities(android.net.Network)) for official documentation.

2. Update `NetworkConnectionUtilProdImplTest`

- Add tests for SDK 29+ using @Config(sdk = [Build.VERSION_CODES.Q]).
- Retain or tag existing tests for SDK <29 with @Config(sdk = [Build.VERSION_CODES.P]).

3. Test the app on emulators with newer and older sdk versions.
4. Push code and let CI run all tests to verify nothing is broken.

-----------
## Issue Breakdown 

Follow these steps to update [NetworkConnectionUtilProdImpl](https://github.com/oppia/oppia-android/blob/96af5e5f11dbf856eef5be9f0ccf264fde4491d6/utility/src/main/java/org/oppia/android/util/networking/NetworkConnectionUtilProdImpl.kt#L13) and its tests:

### 1. Add Dependency
Add the `"//third_party:org_mockito_mockito-core"` dependency in the corresponding build.bazel files to enable mocking in tests:

### 2. Update [NetworkConnectionUtilProdImpl](https://github.com/oppia/oppia-android/blob/96af5e5f11dbf856eef5be9f0ccf264fde4491d6/utility/src/main/java/org/oppia/android/util/networking/NetworkConnectionUtilProdImpl.kt#L13)
Modify the `getCurrentConnectionStatus()` method so that it switches logic based on the SDK version:

```kotlin
override fun getCurrentConnectionStatus(): ConnectionStatus {
  val connectivityManager = context.getSystemService(
    Context.CONNECTIVITY_SERVICE
  ) as ConnectivityManager

  return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) { // API 29+
    getCurrentConnectionStatusForNewerDevices(connectivityManager)
  } else {
[...]

PR #5402, opened 2024-05-22
Fixes part of #4120, part of #1051: Fix a lot of build-time warnings

BenHenning:
## Explanation
Fixes part of #4120
Fixes part of #1051

Similar to #5400, this brings forward changes that would otherwise go in #4937 to simplify the transition to Kotlin 1.6.

Part of #4937 is introducing warnings-as-errors for both Kotlin and Java in order to reduce developer error, simplify the codebase, and minimize warnings when building (which can result in developer habits of ignoring warnings that might have real consequences to end users of the app). In order to keep the main migration PR smaller, this PR fixes all existing warnings and any new ones detected with the Kotlin 1.6 compiler that are not tied to Kotlin 1.5/1.6 API changes (those are part of #4937, instead). Fortunately, most of the changes could be brought forward into this PR.

Specific things to note:
- A few new issues were filed for SDK 33 deprecations caused, but not noted by, #5222): #5404, #5405, and #5406 and corresponding TODOs added. This PR opts for TODOs over actual fixes to minimize the amount of manual verification needed, and to try and keep the PR more focused on non-functional refactor changes (to reduce the risk as reverting this PR may be difficult if an issue is introduced).
- A lot of the fixes were removing redundant casts or null checks.
- The old mechanism we used for view models is deprecated, and had a lot of problems  (partially documented in #1051). This PR moves the codebase over to directly injecting view models instead of using the view model provider (thus getting rid of true Jetpack view models entirely in the codebase).
  - We never used the Jetpack functionality, and we were leaking a lot of context objects that could theoretically result in memory leaks.
  - The migration of view models in this way has already been ongoing in the codebase; this PR just finishes moving the rest of them over to remove the deprecated JetPack view model reference.
  - Note that this doesn't actually change the scope of the view models, and in fact they should largely behave as they always have.
  - ``ObservableViewModel`` was subsequently updated, and may be something we could remove in the future now that it's no longer a Jetpack view model.
  - The old view model binding code was removed, along with its test file exemptions. It's no longer used now that the view models have been finished being migrated over to direct injection.
- Some of the binding adapters didn't correctly correspond to their namespaced properties. I _think_ that the databinding compiler was still hooking them up correctly, but they produced build warnings that have now been addressed (specifically, 'app' is implied). Some other properties were using unusual namespaces, so these were replaced with 'app' versions for consistency & correctness.
- Some cases where SAM interfaces could be converted to lambdas were also addressed (mainly for ``Observer`` callbacks in UI code).
- ``DrawerLayout.setDrawerListener`` was replaced with calls to ``DrawerLayout.addDrawerListener`` since the former [is deprecated](https://developer.android.com/reference/androidx/drawerlayout/widget/DrawerLayout#setDrawerListener(androidx.drawerlayout.widget.DrawerLayout.DrawerListener)). This isn't expected to have a functional difference.
- Some other minor control flow warnings were addressed (such as dead code paths).
- ``when`` cases were updated to be comprehensive (as this is enforced starting in newer versions of Kotlin even for non-result based ``when`` statements).
- Some unused variables were removed and/or replaced with ``_`` per Kotlin convention.
- Some parameter names needed to be updated to match their override signatures.
- One change in ``ExitSurveyConfirmationDialogFragment`` involved removing parsing a profile ID. Technically this is a semantic change since now a crash isn't going to happen if the profile ID is missing or incorrect, but that seems fine since the fragment doesn't even need a profile ID to be passed.
- Some of the test activities were updated to bind a ``Runnable`` callback rather than binding to a method (just to avoid passing the unused ``View`` parameter and to keep things a bit simple binding-wise).
- Some cases were fixed where variables were being shadowed due to reused names in deeper scopes.
- There were some typing issues going on in tests with custom test application components. This has been fixed by explicitly declaring the application component types rather than them being implicit within the generated Dagger code.
- ``getDrawable`` calls were updated to pass in a ``Theme`` since the non-theme version is deprecated.
- Some Java property references were updated, too (i.e. using property syntax instead of Java getters when referencing Java code in Kotlin).
- In some cases, deprecated APIs were suppressed since they're needed for testing purposes.
- Mockito's ``verifyZeroInteractions`` has been deprecated in favor of ``verifyNoMoreInteractions``, so updates were made in tests accordingly.
- ``ExperimentalCoroutinesApi`` and ``InternalCoroutinesApi`` have been deprecated in favor of a newer ``OptIn`` method (which can actually be done via kotlinc arguments, but not in this PR). Thus, they've been outright removed in cases where not needed, and otherwise migrated to the ``OptIn`` approach where they do need to be declared.
- In some cases, Kotlin recommends using a ``toSet()`` conversion for iterable operations when it's more performant, so some of those cases (where noticed) have been addressed.
- Some unused parameter cases needed to be suppressed for situations when Robolectric is using reflection to access them.
- In some cases Android Studio would recommend transformation chain simplifications; these were adopted where obvious.
- There are a few new TODOs added on #3616 as well, to clean up deprecated references that have been suppressed in this PR.
- ``BundleExtensions`` was updated to implement its own version of the type-based ``getSerializable`` until such time as ``BundleCompat`` can be used, instead (per #5405).
- A **lot** of nullability improvements needed to happen throughout the JSON asset loading process since there was a lot of loose typing happening there.
- Some Kotlin & OkHttp deprecated API references were also updated to use their non-deprecated replacements.
- ``NetworkLoggingInterceptorTest`` was majorly reworked to ensure that the assertions would actually run (``runBlockingTest`` was being used which is deprecated, and something I try to avoid since it's very difficult to write tests that use it correctly). My investigations showed that the assertions weren't being called, so these tests would never fail. The new versions will always run the assertions or fail before reaching them, and fortunately the code under test passes the assertions correctly. Ditto for ``ConsoleLoggerTest``.
- Some parts of ``SurveyProgressController`` were reworked to have better typing management and to reduce the need for nullability management.
- Some generic typing trickiness needed to be fixed ahead of the Kotlin version upgrade in ``UrlImageParser``. See file comments & links in those comments for more context.
- ``BundleExtensionsTest`` had to be changed since ``getSerializableExtra`` is now deprecated. We also can't update the test to run SDK 33 since that requires upgrading Robolectric, and Robolectric can't be upgraded without upgrading other dependencies that eventually lead to needing to upgrade both Kotlin and Bazel (so it's a non-starter; this is a workaround until we can actually move to a newer version of Robolectric).
- There was some minor code-deduplication & cleanup done in ``ClickableAreasImage``.
- Some incorrect comments were removed in tests (to the effect of "using not-allowed-listed variables should result in a failure."). These seemed to have been copied from an earlier test, but the later tests weren't actually verifying that behavior so the comment wasn't correct.
- An unused method was removed from ``ConceptCardRetriever`` (``createWrittenTranslationFromJson``) and some other small cleanup/consolidation work happened in that class.
- Some stylistic changes were done in ``TopicController`` for JSON loading to better manage nullable situations, and to try and make the JSON loading code slightly more Kotlin idiomatic.

Note that overall the PR has relied **heavily** on tooling to detect warnings to fix, and automated tests to verify that the changes have no side effects.

Note also that this PR does not actually enable warnings-as-errors; that will happen in a downstream PR.

## Essential Checklist

## For UI-specific PRs only
N/A -- While this changes UI code, it should change very few UI behaviors and only failure cases for those it does affect. It's largely infrastructural-only and falls mainly under refactoring/cleanup work.


BenHenning (commented):
Self-reviewed all changes.

BenHenning:
PTAL @adhiamboperes.

Note that this PR is pretty straightforward to review, but it's very large and caused my computer to have some issues when trying to view the actual changes. It probably is worth breaking this into chunks when reviewing it.

I think it'd also be completely fine to delegate review of most or all of it to other people. I've self-reviewed the changes and find them non-concerning, but I defer to whatever you want to do as the primary reviewer & approver of the PR.

BenHenning (commented):
Self-reviewed latest changes (pulled in some more fixes from #4886).

BenHenning (commented):
Acking latest changes were merged cleanly.

BenHenning (commented):
Latest merge commits went in without changes.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
```diff
@@ -464,6 +467,7 @@ class UrlImageParser private constructor(
       drawable?.colorFilter = colorFilter
     }
 
+    @Deprecated(message = "Deprecated in Java.")
```
adhiamboperes: Why do we need to annotate this usage of `getOpacity()` when it is already annotated as deprecated in the `Drawable` class?

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
```diff
@@ -52,9 +54,12 @@ class NetworkLoggingInterceptorTest {
...
 
-  @Inject lateinit var networkLoggingInterceptor: NetworkLoggingInterceptor
-  @Inject lateinit var context: Context
-  @Inject lateinit var testCoroutineDispatchers: TestCoroutineDispatchers
+  @Inject
+  lateinit var networkLoggingInterceptor: NetworkLoggingInterceptor
+  @Inject
+  lateinit var context: Context
+  @Inject
+  lateinit var testCoroutineDispatchers: TestCoroutineDispatchers
```
adhiamboperes, line 62: Why do we usually insist on formatting @inject fields on a single line when AS or ktlint formats them otherwise?

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgressController.kt
```diff
@@ -499,24 +504,37 @@ class SurveyProgressController @Inject constructor(
...
+    )
+  }
+
+  private suspend fun getStoredResponseToNpsScore(): Int {
+    return getStoredResponse(
+      SurveyQuestionName.NPS, AnswerCase.NPS_SCORE, SurveySelectedAnswer::getNpsScore
+    )
+  }
+
   private suspend inline fun <reified T : Any> getStoredResponse(
```
adhiamboperes, line 525: This is a nice improvement. I did have a hard time coming up with the suitable typing vs nullability management at the time of writing this code.

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
```diff
@@ -246,7 +246,7 @@ class TestBazelWorkspace(private val temporaryRootFolder: TemporaryFolder) {
     if (!isConfiguredForKotlin) {
       // Add support for Kotlin: https://github.com/bazelbuild/rules_kotlin.
       val rulesKotlinReleaseUrl =
-        "https://github.com/bazelbuild/rules_kotlin/releases/download/v1.5.0-beta-3" +
+        "https://github.com/bazelbuild/rules_kotlin/releases/download/v1.7.1" +
```
adhiamboperes, line 249: Is this a test-specific upgrade?

adhiamboperes (commented):
Thanks @BenHenning!
I noted the following:

1. You removed the following comment from a number of files, but it isn't clear why: `// Using not-allowed-listed variables should result in a failure.`
2. There are undocumented changes in [ConceptCardRetriever.kt](https://github.com/oppia/oppia-android/pull/5402/files#diff-f037b8ecc0269cc34361ae2838b18171794645256957f685631c3476164ca728) and [TopicController.kt](https://github.com/oppia/oppia-android/pull/5402/files#diff-0943fb0fc5b908e5ed8da64f53610e2c83c88815de2c2a253c89505da56485ea). I think it is worth including the reasons in the description.

I also left a couple of comments in-line.

utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt
BenHenning: Hmm, I'm actually not sure. I removed this line and it still builds, so it's possible it was needed earlier before other changes were in place, or it's a warning that came from Android Studio and not the Kotlin compiler.

data/src/test/java/org/oppia/android/data/backends/gae/NetworkLoggingInterceptorTest.kt
BenHenning: I prefer the format because it's less lines & easier to read. Eventually we'll be moving over to ktfmt which will hard enforce an exact style #2799 since ktlint is a bit too permissive in my opinion (I'd rather try and make all code look as though it was written by the same author).

There's not a lot we can do about Android Studio's formatting, unfortunately. I haven't found great ways of configuring it to actually match ktlint or ktfmt, but perhaps there are plugins that can help with his (I haven't looked in a long time).

FWIW from a reviewer side, I tend to only ask people to change it if the existing code is already single line (for consistency).

domain/src/main/java/org/oppia/android/domain/survey/SurveyProgressController.kt
BenHenning: Thanks--I'm not sure it's better from a readability perspective though. Lambdas and function references seem to often decrease the comprehensibility of code, so I was a bit on the fence with generalizing things as was done here.

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
BenHenning: Yeah. A lot of this code was originally written in #4937 and haphazardly copied to this PR, and that PR's code originated in #4886. The Bazel PR updates rules_kotlin to 1.7.1 and I figured we can just keep that bit here even though it's not in exact alignment for a few releases. It doesn't make much of a difference in the build graph between tests that use this utility and production.

BenHenning (commented):
Self-reviewed latest changes. No specific concerns.

I left replies to your comments and added one change to address one of them, plus pulled in the latest develop changes. PTAL @adhiamboperes.

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
adhiamboperes: Sounds good.

adhiamboperes (approved):
Everything looks good,  thanks @BenHenning!

BenHenning:
@seanlip PTAL for owners-specific changes.

seanlip:
@BenHenning Owners approval does not seem to be required for this PR, unless I'm missing something?

BenHenning:
Hmm not sure why I saw that originally @seanlip. Thanks for double checking!

Thanks @adhiamboperes for the review!

Enabling auto-merge now that this is up-to-date with the latest develop changes.

BenHenning (commented):
Reviewed latest merge. No conflicts so the merge went in cleanly.

PR #5361, opened 2024-03-14
Fix #5357: Remove CDATA from translatable strings.

BenHenning:
## Explanation
Fixes #5357

This PR removes all ``CDATA`` declarations in the translatable ``strings.xml`` file and instead escapes all necessary characters: < and > (& didn't need to be escaped since no strings use that character at the moment). This is needed because Translatewiki doesn't seem to extract the HTML within the CDATA declaration correctly, so it may not be translated (some existing strings were never translated, and per #5274 the latest FAQ changes aren't being processed correctly (leading to empty translation strings being submitted).

This PR also introduces a regex check + test to ensure that CDATA isn't used anymore in strings.

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
See video for a demonstration on all of the affected strings still rendering correctly (as far as I can tell) after this change:

https://github.com/oppia/oppia-android/assets/12983742/b1102d28-d3a5-4266-969c-d7d84f8d5d38

PR #5274, opened 2023-12-14
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates

PR #5403, opened 2024-05-22
Revert "Fix #5357: Remove CDATA from translatable strings."

### LINKED ISSUE #5357: [BUG]: CDATA causes issues with Translatewiki importing
### Describe the bug

As described in https://github.com/oppia/oppia-android/pull/5274/files#r1517940307 and observed in that PR, CDATA in strings causes one of three Translatewiki behaviors:
1. The string can be correctly translated.
2. The string is ignored (as is observed for privacy_policy_web_link, agree_to_terms, and terms_of_service_web_link.
3. The translation string is empty (#5274).

### Steps To Reproduce

See above.

### Expected Behavior

All strings should be translatable via Translatwiki.

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

I think this means we cannot use CDATA anymore and instead should prohibit it via a regex check. Existing CDATA strings will need to be escaped, instead (e.g. HTML tag markers will need to be replaced with '\&gt;' and '\&lt;').

adhiamboperes:
Reverts oppia/oppia-android#5361

There are some formatting issues that cause incorrectness in the HTML structure, so @translatewiki cannot parse them correctly. This is the cause of empty translatiion strings in #5274. Mainly, some elements are missing their closing tags, e.g.:

```xml 
<string name="faq_answer_bug_reporting">&lt;p&gt;&lt;ol&gt;&lt;li&gt;From your %s app home screen, tap the menu in the top left corner.&lt;/li&gt;&lt;li&gt;Tap &lt;strong&gt;Share feedback&lt;/strong&gt;.&lt;/li&gt;&lt;li&gt;Follow the instructions to report the bug or share feedback.&lt;/li&gt;&lt;/p&gt;</string>
```
is missing an \</ol\> tag before the closing \</p\> tag.

I have looked at several of the other "broken" strings, and they have similar issues. When @translatewiki encounters an unclosed tag, it continues reading until it encounters a corresponding closing tag, in a subsequent string, meaning any strings between the two points are assumed to be part of a single string. This is obviously a serious bug in translatewiki itself.

I think it is best to revert this PR and fix the issues first before merging it again.


BenHenning (approved):
Thanks @adhiamboperes! The plan SGTM and the revert seems clean.

BenHenning:
Also reopened #5357 since the PR is being reverted.

Issue #5344, opened 2024-02-17
[Feature Request]: Introduce support for multiple classrooms (GSoC'24 4.2)


BenHenning:
### Is your feature request related to a problem? Please describe.

The app is currently limited to only showing basic numeracy lessons, and the broader team plans to introduce lessons of other types (such as financial literacy, science, and more).

### Describe the solution you'd like

Mocks: see https://github.com/oppia/design-team/issues/125. The design contact is @ashley-ya. 

#### Suggested Milestones
- **Milestone 1**: Introduce new UIs for classroom selection.
  - Deliverable 1: Introduce new feature flag for the multiple classrooms feature.
  - Deliverable 2: Update the model & domain layer to support the definition of classrooms, and specify a classroom per-topic.
  - Deliverable 3: Introduce a ClassroomController.
  - Deliverable 4: Introduce a new activity & related fragments/views/tests for a new classroom selection landing page (recommendations section should be faked out for now).
  - Deliverable 5: Update 'recently played' topic cards to include classroom information.
- **Milestone 2**:
  - Deliverable 1: Update the asset download script ([#4885](https://github.com/oppia/oppia-android/pull/4885)) to support classrooms.
  - Deliverable 2: Implement new recommendations logic & UI support for the classroom selection 
  - Deliverable 3: Gated by the feature flag, hook up the new classroom selection page to replace the existing home activity upon profile login.
  - Deliverable 4: Ensure the existing event logs are captured in the new screen and related tests in preparation for removal.
  - Deliverable 5: Test, iterate, and work with the tech lead to finalize and launch the feature.
  - Deliverable 6: Audit home activity/fragment & recommendation tests to ensure the new utilities cover the same behaviors. Remove the old home activity/fragment.

<details>
<summary>Technical hints / guidance</summary>

#### Top-level components needed for supporting multiple classrooms in the app
- ``ClassroomListActivity.kt`` + presenter & layout: Top-level activity that has a row of available classrooms which is displayed as a carousel to facilitate addition of future classrooms. 
    - The list of topics will be filtered based on the selected classroom. 
    - The classrooms carousel shoud be sticky at the top when the topic list is scrolled. 
    - When the home screen is resumed, the last classroom selection will be retained.
- ``ClassroomListFragment.kt`` + presenter & layout: Top-level fragment for the available classrooms. The classrooms are displayed in a carousel(custom view as per existing patterns, see the "last played stories" carousel).
[...]

PR #5398, opened 2024-05-17
Fix #1547, parts of #169, #5344, #5365, #5411: Add data layer support for (multiple) classrooms & topic dependencies, and prepare for #4885

BenHenning:
## Explanation
Fixes #1547
Fixes part of #169
Fixes part of #5344
Fixes part of #5365
Fixes part of #5411

The main purpose of this PR is to introduce support for multiple classrooms in the data layer of the app (with minimal domain integration to avoid the change extending beyond the lesson structures). However, the PR is also doing a few more things including preparing the Android codebase for introducing an asset download script (#4885) and other peripheral cleanups of code (rather than updating it) that was noticed along the way.

### Preparing for multiple classrooms

This addresses part of #5365.

#5344 is introducing support for multiple classrooms in the app. To help prepare for those changes, this PR introduces the necessary data structure and domain loading changes to load a new proto structure:

```proto
message ClassroomList {
  repeated ClassroomRecord classrooms = 1;
}
message ClassroomRecord {
  string id = 1;
  map<string, TranslationMapping> written_translations = 2;
  SubtitledHtml translatable_title = 3;
  LessonThumbnail classroom_thumbnail = 4;
  map<string, TopicIdList> topic_prerequisites = 5;
  message TopicIdList {
    repeated string topic_ids = 1;
  }
}
```

Rather than just a flat topic list. Some important details to note:
- The recommended topics structure has been updated to use this new ``topic_prerequisites`` value being loaded from classrooms. This will also extend to production assets since the asset download script from #4885 is also being updated to include support for this multiple classrooms structure to address the remainder of #5365.
- To minimize domain changes, the new loading code assumes only **one** classroom is present. TODOs have been added on #5344 to extend this to support multiple classrooms.
- Current loading code (including for JSON) is ignoring all but: ``classrooms``, ``id``, and ``topic_prerequisites`` (including ``topic_ids``) from the protos above. These other fields are expected to have supported added as part of #5344.
- There were some color simplifications made in ``TopicListController``. These largely shouldn't have a major impact outside of developer assets. These changes were made to ensure non-specificity to the previous lesson classroom. In general, all of this should eventually be removed in favor of loading colors from lesson assets, but that will need to wait until the JSON pipeline is completely removed.

### Asset priming removal

This addresses part of #169.

``PrimeTopicAssetsController`` and its implementation were responsible for hackily pre-loading all lesson images and audio to be on-device to enable offline support. This was the first attempt at offline support early in the app's development, but it had a few significant drawbacks:
[...]

PR #5418, opened 2024-05-30
Fix part of #5344: Update models to support classrooms

theMr17:
## Explanation

Fixes part of #5344

 - Introduces models for `ClassroomIdList`, `ClassroomSummary`, & `EphemeralClassroomSummary` and updates `ClassroomList` to contain a list of `EphemeralClassroomSummary`s and a `written_translation_context`.
 - Topic and Story models are updated to include reference to `classroomId` & `classroomTitle`.
 - Test json & textproto data files are also added for 3 test classrooms, alongwith updating existing topic data files to include classroom information.

## Essential Checklist

## For UI-specific PRs only


theMr17:
@adhiamboperes PTAL. I have updated the classroom data files with the `topic_ids` list as discussed. This PR is now ready for a review.

domain/src/main/assets/test_classroom_id_0.json
```diff
@@ -0,0 +1,14 @@
...
+    "content_id": "classroom_title",
+    "html": "Science"
+  },
+  "thumbnail_bg_color": "#00FFFFFF",
+  "thumbnail_filename": "",
+  "topic_prerequisites": {
+    "test_topic_id_0": ["GJ2rLXRKD5hw"],
+    "test_topic_id_1": ["test_topic_id_0", "omzF4oqgeTXd"]
+  },
+  "topic_ids": ["test_topic_id_0", "test_topic_id_1"]
```
adhiamboperes: Per discussion on chat, this is not really necessary, we will get topic list from the `topic_prerequisites` object(unless you need it for something downstream). Same to the textproto.

model/src/main/proto/topic.proto
```diff
@@ -242,6 +248,12 @@ message CompletedStory {
   // The title of the topic this story is part of.
   SubtitledHtml topic_title = 9;
 
+  // The ID of the classroom this story is part of.
+  string classroom_id = 10;
+
+  // The title of the classroom this story is part of.
+  SubtitledHtml classroom_title = 11;
```
adhiamboperes: For this and `PromotedStory` below,  What's the purpose of adding classroom information to the story object?

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -794,51 +839,88 @@ class TopicListController @Inject constructor(
...
-    }
-    val topicPrereqs = topicPrereqsObj.keys().asSequence().associateWith { topicId ->
-      val topicIdArray = checkNotNull(topicPrereqsObj.optJSONArray(topicId)) {
-        "Expected topic $topicId to have a non-null string list."
+  private fun loadClassroomsFromJson(): List<ClassroomRecord> {
+    // Load the classrooms.json file.
+    val classroomIdsObj = jsonAssetRetriever.loadJsonFromAsset("classrooms.json")
+    checkNotNull(classroomIdsObj) { "Failed to load classrooms.json." }
+    val classroomIds = classroomIdsObj.optJSONArray("classroom_id_list")
+    checkNotNull(classroomIds) { "classrooms.json missing classroom IDs." }
```
adhiamboperes:
```suggestion
    checkNotNull(classroomIds) { "classrooms.json is missing classroom IDs." }
```

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -794,51 +839,88 @@ class TopicListController @Inject constructor(
...
+
+    // Initialize a list to store the classroom records.
+    val classroomRecords = mutableListOf<ClassroomRecord>()
+
+    // Iterate over all classroom IDs and load each classroom's JSON.
+    for (i in 0 until classroomIds.length()) {
+      val classroomId = checkNotNull(classroomIds.optString(i)) {
+        "Expected non-null classroom ID at index $i."
       }
```
adhiamboperes: Can potentially be more idiomatic.
```suggestion
 classroomIds.forEachIndexed { index, _ ->
    val classroomId = checkNotNull(classroomIds.optString(index)) {
        "Expected non-null classroom ID at index $index."
    }
```

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -794,51 +839,88 @@ class TopicListController @Inject constructor(
...
-          TopicIdList.newBuilder().apply {
-            addAllTopicIds(topicIds)
-          }.build()
+
+      // Load the topic ids of the current classroom.
+      val topicIdJsonArray = checkNotNull(classroomObj.getJSONArray("topic_ids")) {
+        "Expected classroom to have non-null topic_ids."
+      }
+      val topicIdListBuilder = TopicIdList.newBuilder()
+      for (i in 0 until topicIdJsonArray.length()) {
```
adhiamboperes: Use kotlin range operations as suggested above.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -794,51 +839,88 @@ class TopicListController @Inject constructor(
...
-      val topicIdArray = checkNotNull(topicPrereqsObj.optJSONArray(topicId)) {
-        "Expected topic $topicId to have a non-null string list."
+  private fun loadClassroomsFromJson(): List<ClassroomRecord> {
+    // Load the classrooms.json file.
+    val classroomIdsObj = jsonAssetRetriever.loadJsonFromAsset("classrooms.json")
+    checkNotNull(classroomIdsObj) { "Failed to load classrooms.json." }
+    val classroomIds = classroomIdsObj.optJSONArray("classroom_id_list")
+    checkNotNull(classroomIds) { "classrooms.json missing classroom IDs." }
+
+    // Initialize a list to store the classroom records.
```
adhiamboperes:
```suggestion
    // Initialize a list to store the [ClassroomRecord]s.
```

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
```diff
@@ -794,51 +839,88 @@ class TopicListController @Inject constructor(
...
+    val classroomRecords = mutableListOf<ClassroomRecord>()
+
+    // Iterate over all classroom IDs and load each classroom's JSON.
```
adhiamboperes:
```suggestion
    // Iterate over all classroomIds and load each classroom's JSON.
```

adhiamboperes (changes requested):
Thanks @theMr17!

I left a few comments inline, PTAL.

model/src/main/proto/topic.proto
theMr17: For `CompletedStory`, the classroom info shall not be required. I have removed it.

But for the `PromotedStory` object, the classroom info will be required for displaying the classroom name on the recommended stories carousel on the home screen.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
theMr17: Done.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
theMr17: This cannot be done, since `classroomIds` is a `JSONArray` here. `forEachIndexed` can be performed only on a Kotlin Array.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
theMr17: Unable to update it per suggestion for same reasons as above.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
theMr17: Done.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
theMr17: Done.

domain/src/main/assets/test_classroom_id_0.json
theMr17: Removed the topic IDs list.

theMr17:
@adhiamboperes, I have addressed all the comments, PTAL.

domain/src/main/java/org/oppia/android/domain/topic/TopicListController.kt
adhiamboperes: Okay, sounds good.

adhiamboperes (approved):
Thanks @theMr17! LGTM. 

@BenHenning, PTAL for codeowners.

model/src/main/proto/topic.proto
```diff
@@ -444,6 +463,12 @@ message UpcomingTopic {
   // The title of the topic.
   SubtitledHtml title = 7;
 
+  // The ID of the classroom this topic is part of.
+  string classroom_id = 8;
+
+  // The title of the classroom this topic is part of.
+  SubtitledHtml classroom_title = 9;
```
BenHenning, line 472: Here & elsewhere: I'm not sure this will work correctly. SubtitledHtml in activity containers is expected to be localized by the WrittenTranslationContext of that activity's ephemeral object. However, this title belongs to a different context (classroom), so it won't have translations in this activity's WrittenTranslationContext.

You could put the actual classroom title in the ephemeral version, or just rely on multiple fetches from the controller to get the title in cases where we actually need it.

model/src/main/proto/topic.proto
```diff
@@ -547,10 +572,10 @@ message EphemeralRevisionCard {
...
 }
 
-// Corresponds to a local file cataloging all available classrooms in the app.
-message ClassroomList {
-  // The list of classrooms available to the app.
-  repeated ClassroomRecord classrooms = 1;
+// Corresponds to a local file cataloging all classrooms available to load.
+message ClassroomIdList {
+  // The list of IDs corresponding to classrooms available on the local filesystem.
+  repeated string classroom_ids = 1;
```
BenHenning, line 580: Why split up classroom records like this? I'm not opposed, but the structures are pretty simple and the assumption is that, long term, the app will need all of this information to even show the option to download new classrooms.

model/src/main/proto/topic.proto
```diff
@@ -331,6 +343,13 @@ message TopicSummary {
   // The title of the topic.
   SubtitledHtml title = 8;
 
+  // The ID of the classroom this topic is part of.
+  string classroom_id = 10;
+
+  // The title of the classroom this topic is part of.
+  SubtitledHtml classroom_title = 11;
+
+
```
BenHenning: Nit: please remove extra newline.

BenHenning (commented):
Thanks @theMr17. I didn't look at the data changes closely yet because I had a few questions on the model changes; will take another pass after those are addressed.

model/src/main/proto/topic.proto
theMr17: As I understand your comment, you initially implemented a `classrooms.json` file containing a list of classroom records directly in #5398.

In my approach, I opted to maintain a separate list of `classroomId`s. The classroom IDs list is then used to fetch the corresponding classroom records. This decision is made to align with the existing structure used for topic records and topic IDs.

I understand your concern. Do you think this approach would cause any major issues, given that topics will also need to follow a similar method for downloading? If you prefer, I can revert to your original approach.

model/src/main/proto/topic.proto
theMr17: Done.

model/src/main/proto/topic.proto
theMr17: I have moved the classroom titles to the ephemeral versions and also added a `classroom_written_translation_context` wherever the `classroom_title` field is present.

theMr17:
Thanks for the review @BenHenning! I have addressed the comments, PTAL.

model/src/main/proto/topic.proto
BenHenning: Hmm I think it's fine to keep it this way.

domain/src/main/assets/omzF4oqgeTXd.json
```diff
@@ -21,6 +21,7 @@
   "topic_name": "Ratios and Proportional Reasoning",
   "topic_description": "Many everyday problems involve thinking about proportions. For example, if 30 apples cost 6 dollars, how much would 12 apples cost? In these lessons, you'll learn about how to use ratios and proportional reasoning to solve problems like this.",
   "topic_id": "omzF4oqgeTXd",
+  "classroom_id": "test_classroom_id_1",
```
BenHenning: Why is classroom ID being added to any of these assets? I'd only expect it to be in the summary and ephemeral objects passed to the UI layer.

I'm not very keen on duplicating IDs in the on-disk structures themselves since it puts a non-trivial burden on the script (and later, Oppia web) systems to ensure ID parity.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -473,6 +474,7 @@ class TopicController @Inject constructor(
           putAllWrittenTranslations(topicRecord.writtenTranslationsMap)
           title = topicRecord.translatableTitle
           description = topicRecord.translatableDescription
+          classroomId = topicRecord.classroomId
```
BenHenning: As mentioned in another comment, I'd instead expect that we'd load a topic in the context of a classroom (that is, pass the classroom ID down into ``TopicController``).

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -912,6 +922,56 @@ class TopicController @Inject constructor(
     }.build()
   }
 
+  // TODO(#5344): Move this to classroom controller.
+  fun loadClassroomById(classroomId: String): ClassroomRecord {
```
BenHenning: Do we want to expose loading a record like this? Usually we don't and instead opt for summary objects to be passed around.

Might it be possible to make this private and instead introduce a new method that, with additional parameters, returns the more direct object needed by callers?

Separately, this should also be tested in this PR. You can always move the tests to the new ClassroomControllerTest once it's added. It's good practice to always add tests with public API changes of a class.

BenHenning (changes requested):
Thanks @theMr17! Took a full pass and had just a few comments--PTAL.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
theMr17: These functions have now been removed from the `TopicController`, as they were previously required to load the classroom's ID and translation maps into the topic objects. The `Topic` and `EphemeralTopic` objects no longer contain references to their classrooms.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
theMr17: This is now removed.

domain/src/main/assets/omzF4oqgeTXd.json
theMr17: I have removed the `classroom_id` from the topic data files. Currently, the topic list returned from the `TopicListController` does not contain the classroom ID in the topics. However, the classroom ID will be present when the topic list is fetched from the `ClassroomController`.

Additionally, the `PromotedStory` object requires the classroom ID and title. Previously, this information was fetched directly from the `TopicRecord`. After this change, a new function `getClassroomIdByTopicId` has been introduced for this purpose.

The classroom ID cannot be passed down to this function because the `PromotedActivityList` is independent of the selected classroom and returns promoted stories belonging to various classrooms.

theMr17:
@BenHenning PTAL.

theMr17:
Merged the Kotlin and Bazel update changes.

BenHenning (approved):
Thanks @theMr17! The latest changes LGTM--no concerns here. I also find the latest code a lot cleaner, so I think the most recent changes have simplified things.

BenHenning:
@adhiamboperes / @theMr17 please go ahead and merge once you're ready (I didn't want to in case it broke any downstream chaining).

PR #5240, opened 2023-11-22
Technical Analytics: Milestone 2 - Add Ability To Log Feature Flags

kkmurerwa:
## Explanation

When merged, this PR will:
- Create a `FeatureFlagLogger.kt` file that aggregates and logs all feature flags.
- Add a `FeatureFlagLoggerTest.kt` to test the FeatureFlagLogger.
- Add logging logic to the `ApplicationLifecycleObserver.kt` file so that it is logged when the app is in foreground.

### Screenshot of the FeatureFlagContext log in the Event Logs page
<p align="center">
  Screenshot of the Event Logs page showing the FeatureFlagsEventContext being logged
</p>

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #5423, opened 2024-06-13
Fix part of #5343: Introduce Utilities to execute code coverage for a specific bazel test target

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 1.1 of Project 4.1]

### This PR introduces 2 utilities:
1. **RunCoverageForTestTarget** - A new script for running code coverage locally
2. **CoverageRunner** - A new scripting utility for running code coverage for a selected target

### Source:
- The RunCoverageForTestTarget script takes in a test target as an argument.
- The CoverageRunner executes the bazel coverage command asynchronously utilizing bazelClient.
- Implementation is added to BazelClient to execute code coverage commands.
- The getCoverage() from CoverageRunner returns a list of string containing the output result.
- The result is parsed via parseCoverageDataFile() to extract the coverage.dat file path.

The extracted coverage.dat file will then be used to acquire the actual coverage data.

### Tests:
- Tests are introduced for RunCoverageForTestTarget and CoverageRunner utilities.
- To test the execution, sample test files are generated and utilized for testing.
- Sample test files include - 
  - [sourceFile.kt, BUILD.bazel], 
  - [testFile.kt,BUILD.bazel], 
  - WORKSPACE 
  in their appropriate subpackages
- To add functionality to introduce source and test files with content TestBazelWorkspace utility is utilized.
- The addSourceAndTestFileWithContent() in TestBazelWorkspace utility achieves this.
- Tests are added for
  - BazelClientTest
  - TestBazelWorkspaceTest

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
@BenHenning can you PTAL, this PR is cherry-picked from #5420 as that was chained with `migration-to-newer-bazel-kotlin` branch, which recently got merged and the branch was deleted. This PR is worked with the develop branch with newer Bazel-6.5.0 version.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +130,19 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
     return correctedTargets
   }
 
+  /**
+   * Runs code coverage for the specified Bazel test target.
+   *
+   * @param bazelTestTarget Bazel test target for which code coverage will be run.
+   * @return generated coverageResult output
```
BenHenning:
```suggestion
   * @param bazelTestTarget Bazel test target for which code coverage will be run
   * @return generated coverage command result output lines
```

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +130,19 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+   */
+  fun runCoverageForTestTarget(bazelTestTarget: String): List<String> {
+    return executeBazelCommand(
```
BenHenning: Normally we try to process the Bazel command output and return something more useful for calling clients. If the only thing that we need here is the dat file, perhaps we could just extract that from the result, parse it, and read it here (even as a binary data blob)? That simplifies the contract of this method & thus calling code (though does add a bit of complexity to BazelClientTest).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.oppia.android.scripts.common.CommandExecutorImpl
+import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
+import java.io.File
+import java.util.concurrent.TimeUnit
+
+/**
+ * Class responsible for running coverage analysis asynchronously.
+ *
+ * @param repoRoot the absolute path to the working root directory
+ * @param targetFile Path to the target file to analyze coverage.
```
BenHenning: Not a parameter--please remove.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.oppia.android.scripts.common.CommandExecutor
+import org.oppia.android.scripts.common.CommandExecutorImpl
+import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
+import java.io.File
+import java.util.concurrent.TimeUnit
+
+/**
+ * Class responsible for running coverage analysis asynchronously.
+ *
+ * @param repoRoot the absolute path to the working root directory
```
```suggestion
 * @param repoRoot the absolute [File] to the working root directory
```

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+ * @param targetFile Path to the target file to analyze coverage.
+ */
+class CoverageRunner(
+  private val repoRoot: File,
+  private val scriptBgDispatcher: ScriptBackgroundCoroutineDispatcher
```
BenHenning: Add to docstring.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Runs coverage analysis asynchronously for the Bazel test target.
+   *
+   * @param repoRoot the absolute path to the working root directory
+   * @param scriptBgDispatcher the [ScriptBackgroundCoroutineDispatcher] to be used for running the coverage command
+   * @param bazelTestTarget Bazel test target to analyze coverage.
+   * @return a deferred value that contains the path of the coverage data file [will contain the proto for the coverage data].
+   */
+  suspend fun runWithCoverageAsync(
```
BenHenning: This is a bit contradictory. 'Async' usually means non-suspend (hence the ``Deferred`` return). We should either make this a suspend function that could block (and thus not 'async'), or non-suspend async.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param repoRoot the absolute path to the working root directory
+   * @param scriptBgDispatcher the [ScriptBackgroundCoroutineDispatcher] to be used for running the coverage command
+   * @param bazelTestTarget Bazel test target to analyze coverage.
+   * @return a lisf of string that contains the result of the coverage execution.
+   */
+  fun getCoverage(
+    bazelTestTarget: String
+  ): List<String> {
+    val commandExecutor: CommandExecutor = CommandExecutorImpl(
```
BenHenning: I suggest passing this in like we do for other utilities. You can default its construction, but being able to configure it can be really useful in the future if we need to tweak the runtime (in case some code coverage runs take longer than 5 minutes to complete).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+      scriptBgDispatcher, processTimeout = 5, processTimeoutUnit = TimeUnit.MINUTES
+    )
+    val bazelClient = BazelClient(repoRoot, commandExecutor)
```
BenHenning: Suggest initializing this via a lazy field (like we do in a bunch of other scripts and script utilities).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,79 @@
...
+    return coverageData
+  }
+
+  /**
+   * Parse the coverage command result to extract the path of the coverage data file.
+   *
+   * @param data the result from the execution of the coverage command
+   * @return the extracted path of the coverage data file.
+   */
+  fun parseCoverageDataFile(data: List<String>): String? {
```
BenHenning: This & the method above don't need to be public since they're never called outside this class--suggest making them private and/or inlining them.

They also won't need KDocs.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
```diff
@@ -0,0 +1,57 @@
...
+   * @return [Path of the coverage data file].
+   */
+  fun runWithCoverageAnalysis(): String? {
+    return ScriptBackgroundCoroutineDispatcher().use { scriptBgDispatcher ->
+      runBlocking {
+        val result =
+          CoverageRunner(repoRoot, scriptBgDispatcher).runWithCoverageAsync(targetPath).await()
+        result
+      }
+    }
```
BenHenning: This seems incomplete--shouldn't it be actually parsing the dat file and converting it to a proto for the test target?

BenHenning (changes requested):
Thanks @Rd4dev! Took a first pass (skipped tests for now since there are some high-level API clarifications that I'd like figured out, first, since it'll probably change the tests a bit). PTAL.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: I believe along with that doing this will also ensures we process the content immediately and acquire the necessary data from Bazel, preserving it using the byte array. Based on these understanding, I added implementations as suggested to [Bazel Client](https://github.com/oppia/oppia-android/pull/5423/commits/d410d17326b86178e954f155fa79530efbdaa452#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR180) to parse and get the coverage data file path, read it as and return a byte array. 

For the coverage runner, instead of returning the same byte array, I [convert it to a string](https://github.com/oppia/oppia-android/pull/5423/commits/ece147a249d6e955d4a8184a5a34c0ee11619f83#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR68) to simplify testing in PR 1.1 as we can confirm the data present with string. 
Future PRs will use this to parse it to proto and return it to the runCoverageForTestTarget. 

**Tests To Update:** 
- To also Update the tests for CoverageRunner and runCoverageForTestTarget
- Add tests for new implementations in BazelClientTest

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: **Done:** [Updated doc](https://github.com/oppia/oppia-android/pull/5423/commits/d410d17326b86178e954f155fa79530efbdaa452#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dL136) with implementation to return the coverage data.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: **Done:** Updated [BazelClientTest](https://github.com/oppia/oppia-android/pull/5423/commits/5db06508236e7089c30739537f5dbfd7251e3497)
**Done:** Updated [CoverageRunnerTest](https://github.com/oppia/oppia-android/pull/5423/commits/87be11408b172b752834f663c60a744a0abdfa3f)
**Done:** Updated [RunCoverageForTestTargetTest](https://github.com/oppia/oppia-android/pull/5423/commits/456e81b55b9fec782bbcc0f962577163c07b0fa4)

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Updated the documentation for `@param repoRoot` to align with standard definitions. Now specifies it as '_the root directory of the repository_'. Changes: [[link to the commit](https://github.com/oppia/oppia-android/pull/5423/commits/2c655249ae49fca8caee094a1430013036b382f5#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL17)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Removed the mistakenly mentioned `@param targetFile` - [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/2c655249ae49fca8caee094a1430013036b382f5#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL17)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Added `@param scriptBgDispatcher` to the docstring - [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/2c655249ae49fca8caee094a1430013036b382f5#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL17)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Thanks Ben, for the additional clarification. Initially, I marked the function as suspend to handle result retrieval, which remained unnecessary. I will remove the suspend keyword and maintain it as async to ensure concurrent execution. [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/62f9e1d9abf8ffb1f678d6eec93fa19dff2ddf2f)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: I agree. Implementing `processTimeout` via command-line arguments would be valuable, especially for managing longer processes like coverage runs, as I too have seen many failing due to this and have had to rerun them using cached results. 

**Custom Process Timeout:** [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/5e6d2ef8288011ac57c8d9ac1a1859f66474a305)]

I've updated it to accept `processTimeout` as an argument, allowing **dynamic timing** or **defaulting to 5 minutes**. I would like to know your thoughts on this approach. If it adds complexity at this stage, I can revert it back to the default of 5 minutes for all processes.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: **Done:** Initialized bazelClient using lazy field as done in MavenDependenciesRetriever.kt [[commit](https://github.com/oppia/oppia-android/pull/5423/files#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR23)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: **Done**. Earlier, I relocated these functions to the `BazelClient` where I implemented coverage execution and data reading functionalities, thus making them public there. Simultaneously, I made the `readDataFileAsBinary` function private. I assume (*_repo standards_*) that private functions typically do not require documentation, I have removed the documentation from both. [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/81d68550549d046aad154105185dae98b2d093bd)]

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Yes Ben, it should be and that is part of **PR 1.4** **[[Milestone Table](https://github.com/oppia/oppia-android/issues/5343#issuecomment-2146479220)]**.
Currently, it returns the acquired coverage data. Once PR 1.4 is implemented, this data will be converted to protos before being returned in the `CoverageRunner` utility.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: **Update:** Revoked those visibility changes [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/6cb1da7056f63c85f1327ba0366594b43428cf6d)] as they were called in `BazelClientTest` testing methods [[BazelClientTest.kt](https://github.com/oppia/oppia-android/pull/5423/files#diff-e50d7ab1968e374641f7298e27dca62ec182a42dc06dac4c47857cd7526de00fR406)].

Rd4dev:
I have addressed the first pass review comments @BenHenning, PTAL.

Rd4dev:
@BenHenning sorry for assigning, missed the Todo Check failing case. I'll figure that out and re-assign for review!

Rd4dev:
@BenHenning @adhiamboperes recently after the merge of the PR #5240, I am facing the Todo not corresponding to open issues - [Todo Check error](https://github.com/oppia/oppia-android/actions/runs/9548117968/job/26314577064?pr=5423#step:12:60), as the TODO was still mentioned in the [ApplicationLifecycleObserverTest.kt](https://github.com/oppia/oppia-android/blob/b9d9dffb4bf2c6cbd96636a933f42217a9d63651/domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt#L429) [line number: 429] should I update anything (remove todo) in this PR to solve this?

adhiamboperes:
> @BenHenning @adhiamboperes recently after the merge of the PR #5240, I am facing the Todo not corresponding to open issues - [Todo Check error](https://github.com/oppia/oppia-android/actions/runs/9548117968/job/26314577064?pr=5423#step:12:60), as the TODO was still mentioned in the [ApplicationLifecycleObserverTest.kt](https://github.com/oppia/oppia-android/blob/b9d9dffb4bf2c6cbd96636a933f42217a9d63651/domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt#L429) [line number: 429] should I update anything (remove todo) in this PR to solve this?

Hi @Rd4dev,  this should be fixed now...the issue number was incorrect.

Rd4dev:
Thanks @adhiamboperes, will update this with the latest changes!

Rd4dev:
@adhiamboperes @BenHenning, Update on the `testDebugUnitTest` I mentioned, the previous run had this issue with **Non-app Module Robolectric Tests** with [Execution failed for task ':domain:testDebugUnitTest'](https://github.com/oppia/oppia-android/actions/runs/9548117960/job/26314575296#step:13:6798) for the [`QuestionAssessmentProgressControllerTest`](https://github.com/oppia/oppia-android/actions/runs/9548117960/job/26314575296#step:13:6048) , but with merging latest changes (todo check update) all the checks with this PR **PASSED**.

Rd4dev:
@BenHenning, with the latest changes merged all checks pass and I have addressed all the review comment can you PTAL!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +132,52 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
     return correctedTargets
   }
 
+  /**
+   * Runs code coverage for the specified Bazel test target.
+   *
+   * @param bazelTestTarget Bazel test target for which code coverage will be run
+   * @return the coverage data file path generated from bazel coverage
```
BenHenning: It's returning the coverage file contents now, not the path--correct?

Also, should mention that null can be returned & why/when that happens.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @param repoRoot the root directory of the repository
+ * @param targetPath Bazel test target to analyze coverage
+ * @param commandExecutor Executes the specified command in the specified working directory
+ */
+class RunCoverageForTestTarget(
+  private val repoRoot: File,
+  private val targetPath: String,
+  private val commandExecutor: CommandExecutor,
+  private val scriptBgDispatcher: ScriptBackgroundCoroutineDispatcher
```
Rd4dev: Todo: Missed the doc string for scriptBgDispatcher.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +132,52 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+  /**
+   * Reads the content of the .dat file as a binary blob.
+   *
+   * @param filePath path to the .dat file
+   * @return content of the .dat file as binary data
+   */
+  fun readDatFileAsBinary(filePath: String?): ByteArray? {
+    val path = Paths.get(filePath!!)
+    return Files.readAllBytes(path!!)
+  }
```
BenHenning: Do these actually need to be public or can they instead be private helpers? Prefer not over-exposing internals where possible.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +132,52 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+   * @return the coverage data file path generated from bazel coverage
+   */
+  fun runCoverageForTestTarget(bazelTestTarget: String): ByteArray? {
+    val coverageData = executeBazelCommand(
+      "coverage",
+      bazelTestTarget
+    )
+    val coverageDataFilePath = parseCoverageDataFile(coverageData)
+    val coverageDataFileContent = readDatFileAsBinary(coverageDataFilePath!!)
+    return coverageDataFileContent
```
BenHenning: The nullability seems a bit off here since we can possibly return null. Perhaps something like this would be a bit more robust:

```suggestion
    return parseCoverageDataFile(coverageData)?.let { path ->
      readDatFileAsBinary(path)
    }
```

Note this means that ``readDatFileAsBinary`` can also be simplified since it never takes a null string value.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,64 @@
...
+  private val scriptBgDispatcher: ScriptBackgroundCoroutineDispatcher,
+  private val commandExecutor: CommandExecutor
+) {
+  private val bazelClient by lazy { BazelClient(repoRoot, commandExecutor) }
+
+  /**
+   * Runs coverage analysis asynchronously for the Bazel test target.
+   *
+   * @param bazelTestTarget Bazel test target to analyze coverage.
+   * @return a deferred value that contains the coverage data [will contain the proto for the coverage data].
```
```suggestion
   * @param bazelTestTarget Bazel test target to analyze coverage
   * @return a deferred value that contains the coverage data (will contain the proto for the coverage data)
```

Fixed periods (shouldn't end in punctuation). Also, I assume you meant to use a parenthetical here--square brackets mean 'linkify symbol' in KDocs markdown.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,64 @@
...
+    return coverageDataString
+  }
+
+  /**
+   * Converts a ByteArray to a String using UTF-8 encoding.
+   *
+   * @param coverageBinaryData byte array to convert
+   * @return string representation of the byte array
+   */
+  fun convertByteArrayToString(coverageBinaryData: ByteArray?): String? {
```
BenHenning: This very much feels like a helper and not something that should be part of the runner's contract.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  /**
+   * Runs coverage command for the Bazel test target.
+   *
+   * @param bazelTestTarget Bazel test target to analyze coverage.
+   * @return the generated coverage data as a string.
+   */
+  fun getCoverage(
```
BenHenning: Probably should be private--right?

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,64 @@
...
+    bazelTestTarget: String
+  ): String? {
+    val coverageDataBinary = bazelClient.runCoverageForTestTarget(bazelTestTarget)
+    val coverageDataString = convertByteArrayToString(coverageDataBinary!!)
```
BenHenning: Why convert the byte array to a string--it's not a file path anymore. Unless the data file is actually a text file, the charset conversion may lead to information being lost.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,64 @@
...
+   * @param bazelTestTarget Bazel test target to analyze coverage.
+   * @return a deferred value that contains the coverage data [will contain the proto for the coverage data].
+   */
+  fun runWithCoverageAsync(
+    bazelTestTarget: String
+  ): Deferred<String?> {
+    return CoroutineScope(scriptBgDispatcher).async {
+      val coverageDataFilePath = getCoverage(bazelTestTarget)
+      coverageDataFilePath
+    }
```
```suggestion
    return CoroutineScope(scriptBgDispatcher).async { getCoverage(bazelTestTarget) }
```

Can be simplified a bit.

Also, perhaps something like 'retrieveCoverageResult' would be a better name for ``getCoverage`` (since it's not a simple getter).

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: I think the latest approach is much cleaner. If the dat file is a text file of lines, though, you could also just parse it in BazelClient and return the list of lines as a ``List<String>``--that might be the cleanest option as it provides an almost structured return value for BazelClient.

Note that this approach can be done with Kotlin's ``readLines()`` function, as well: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/read-lines.html.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
BenHenning: Do they need to be? Generally we don't want to expose helper methods for the purpose of testing. Often we call helpers in tests for one of two reasons:
- We want to test that the helper is doing the right thing. The better practice is to figure out what situations of calling methods lead to that state of the helper method and verify those, instead. I call this process mapping certain program states back to public API behaviors that can be verified in tests.
- We need to use the helper to verify something, or to tweak conditions for a public API method. This is definitionally a "back door" into the program state for the sake of testing, and it's considered a bad practice because it increases the surface area of the public API of the class being tested past what we actually want to expose to non-test callers. Simply put, all public API methods/fields are "fair game" to be called in any place that has access to the Bazel target, so we should only expose things we want to be called outside of test cases.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
BenHenning: I think your approach works. I wasn't thinking about making the script itself configurable, just the script's logic class. However, if you think downstream code will need to customize the execution timeout then this approach makes sense.

BenHenning (commented):
Thanks @Rd4dev! Took a full pass on the PR. I think it's looking good, I mostly had nits plus a few suggestions to improve robustness in a few places. PTAL.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Added the docString : https://github.com/oppia/oppia-android/pull/5423/files#diff-04f7dc1ca7f9e9b2f3eac2fe9e9de7841146490f96e68f9059223c8c85588bfaR56

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Yeah Ben, it was! And now its returning the data as a List<String>, so updated the docString to align with the changes and also added points on cases of null : https://github.com/oppia/oppia-android/pull/5423/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR137-R144

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Got it, Ben! Changed their visibility to `private` and also updated the relevant test cases 
1. by removing them 
2. refactoring the tests to make sure those scenarios are still tested by the top-level public methods where possible.

Link to changes: https://github.com/oppia/oppia-android/pull/5423/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR161-R180

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Thanks Ben. 
I didn't realize the nullability issue there. I've now updated the code as suggested using safe calls. 

Revised versions:
**parseCoverageDataFile Changes:** https://github.com/oppia/oppia-android/pull/5423/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR150-R152

**readDatFile (previously readDatFileAsBinary) Changes:** https://github.com/oppia/oppia-android/pull/5423/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR181-R183

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Oh, okay! I just wanted to enclose them to suggest what is expected to return in the future. However, I think that might confuse things with this PR since we are only working with List<String> for now. So, I have removed that and kept it straightforward with just this PR. Maybe we can update that in PR 1.4 when we finally store the protos. [[updated docString](https://github.com/oppia/oppia-android/pull/5423/files#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR29)]

Regarding the ending periods, does it apply to every param KDoc? I think I have them in a bunch of other KDoc strings. Maybe I will update them then.

Rd4dev: **Done:** Removed the ending periods - https://github.com/oppia/oppia-android/pull/5423/commits/506916dec0e897b0cdb11cfbb29af1e83e76f2a4

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Yeah! Given the current changes with `bazelClient` returning data as **List<String>**, this is no longer needed, so I have removed it. [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/ee078e299a707ecdee4fbdc700ab2bc27360d900#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL61-L62)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Yeah! I should have used the method `runWithCoverageAsync`, to execute the coverage generation. I have updated the visibility of `getCoverage` to private and modified the test cases to use `runWithCoverageAsync` instead of `getCoverage`.

Also as per: https://github.com/oppia/oppia-android/pull/5423#discussion_r1646942599, the `getCoverage` is now named to `retrieveCoverageResult`

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Referring to the comment - https://github.com/oppia/oppia-android/pull/5423#discussion_r1641100458, the initial reason was to verify the accuracy of the content, as we previously used to return a byte array. To facilitate test case assertions, I converted them with the intention of removing the conversion methods once we switched to protos. 

However, with the new implementation, it's much more straightforward and easier to check the content directly as it is `List<String>` now, so I removed the conversion methods. [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/ee078e299a707ecdee4fbdc700ab2bc27360d900#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL61-L62)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Got it! Done [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/ee078e299a707ecdee4fbdc700ab2bc27360d900#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdL46)]

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Thanks for the revised approach and also linking the resources, Ben. I believe having it as a `List<String>` will make it easier to parse the information as protos later on as well. I've modified that and all related test cases accordingly - [[commit](https://github.com/oppia/oppia-android/pull/5423/commits/ee078e299a707ecdee4fbdc700ab2bc27360d900#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dL141-R153)]

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Makes sense with this information and what you explained in the meeting. Thanks, Ben. 
Also, aligning with this, I made all the helper methods private and updated the related tests.

Rd4dev:
Thanks for the review @BenHenning. I have addressed all the comments, can you PTAL!

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,51 @@
...
+   * Runs coverage analysis asynchronously for the Bazel test target
+   *
+   * @param bazelTestTarget Bazel test target to analyze coverage
+   * @return a deferred value that contains the coverage data
+   */
+  fun runWithCoverageAsync(
+    bazelTestTarget: String
+  ): Deferred<List<String>?> {
+    return CoroutineScope(scriptBgDispatcher).async {
+      val coverageDataFilePath = retrieveCoverageResult(bazelTestTarget)
```
Rd4dev: TODO: This is no longer the coverageDataFilePath.
           Need to change it to more appropriate variable name (may be coverageResult or coverageData)

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.async
+import org.oppia.android.scripts.common.BazelClient
+import org.oppia.android.scripts.common.CommandExecutor
+import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
+import java.io.File
+
+/**
+ * Class responsible for running coverage analysis asynchronously
```
```suggestion
 * Class responsible for running coverage analysis asynchronously.
```

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+      readDatFile(path)
+    }
+  }
+
+  /**
+   * Parse the coverage command result to extract the path of the coverage data file.
+   *
+   * @param data the result from the execution of the coverage command
+   * @return the extracted path of the coverage data file.
+   */
```
BenHenning: Here & elsewhere in the PR: please remove KDocs for private methods.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param bazelTestTarget Bazel test target to analyze coverage
+   * @return a deferred value that contains the coverage data
+   */
+  fun runWithCoverageAsync(
+    bazelTestTarget: String
+  ): Deferred<List<String>?> {
+    return CoroutineScope(scriptBgDispatcher).async {
+      val coverageDataFilePath = retrieveCoverageResult(bazelTestTarget)
+      coverageDataFilePath
```
```suggestion
      retrieveCoverageResult(bazelTestTarget)
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
```diff
@@ -0,0 +1,86 @@
...
+    ).runCoverage()
+  }
+}
+
+/**
+ * Class responsible for analyzing target files for coverage and generating reports.
+ *
+ * @param repoRoot the root directory of the repository
+ * @param targetPath Bazel test target to analyze coverage
+ * @param commandExecutor Executes the specified command in the specified working directory
```
BenHenning:
```suggestion
 * @param commandExecutor executes the specified command in the specified working directory
```

scripts/src/javatests/org/oppia/android/scripts/common/BazelClientTest.kt
```diff
@@ -20,10 +20,10 @@ import java.util.concurrent.TimeUnit
  * Tests for [BazelClient].
  *
  * Note that this test executes real commands on the local filesystem & requires Bazel in the local
- * environment.
+ * environment
```
BenHenning:
```suggestion
 * environment.
```

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+  private fun parseCoverageDataFile(data: List<String>): String? {
```
BenHenning: Suggest:
- Renaming this to 'parseCoverageDataFilePath' since that's what it's actually doing.
- Renaming 'data' to be 'commandOutputLines' to be more specific. 'data' is generally one of the more vague or ambiguous names to use, so it's often better to avoid it in favor of something more specific.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+  /**
+   * Reads the content of the .dat file as a list of lines.
+   *
+   * @param filePath path to the .dat file
+   * @return content of the .dat file as a list of strings
+   */
+  private fun readDatFile(filePath: String?): List<String>? {
+    return filePath?.let { path ->
+      val pathObj = Paths.get(path)
+      File(pathObj.toUri()).readLines()
```
BenHenning: Why doesn't using ``File`` directly work?

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+    return null
+  }
+
+  /**
+   * Reads the content of the .dat file as a list of lines.
+   *
+   * @param filePath path to the .dat file
+   * @return content of the .dat file as a list of strings
+   */
+  private fun readDatFile(filePath: String?): List<String>? {
```
BenHenning: ``filePath`` is never passed in as null, so this can be simplified.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+  /**
+   * Runs code coverage for the specified Bazel test target
+   *
+   * @param bazelTestTarget Bazel test target for which code coverage will be run
+   * @return the generated coverage data as a list of strings
+   * or null if the coverage data file could not be parsed
+   *
+   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
+   * This can happen due to: Test failures or misconfigurations that prevent the coverage data
+   * from being generated properly
```
BenHenning: ``@return`` should be the last line here--move this to be above the at-clauses.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +131,59 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
+   */
+  fun runCoverageForTestTarget(bazelTestTarget: String): List<String>? {
+    val coverageData = executeBazelCommand(
```
BenHenning: 'coverageCommandOutputLines' would be a much clearer name. As-is, it's currently confusing whether this contains the command's output lines or the actual coverage dat file contents.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
BenHenning: Yeah, we never capitalize or end-punctuate at-clauses, except for when there's more than one sentence in the at-clause (in which case we punctuate, but still don't capitalize). This is conventional, and it also affects the generated KDoc HTML (though we don't currently generate doc pages).

Separately, the summary (i.e. the text above at-clauses) is always capitalized and punctuated.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: Ah, I think part of my earlier comment might not have been as clear so I left a newer comment. The other aspect of this has been addressed, though, so resolving this thread.

BenHenning (changes requested):
Thanks @Rd4dev! Had a few more thoughts--PTAL.

Also, if it saves time feel free to avoid linking the commits in your follow-up comments (it's actually not needed at all since GitHub provides pretty easy ways to see the changes between review passes). If you prefer continuing to include them, feel free but it isn't needed for reviewing. :)

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Oops, I think I removed bunch of punctuations from the summary, will probably re-add them. Thanks for the explaining the KDoc standards.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Removed coverageDataFilePath

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Done: Added periods on all summary KDoc strings

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Done: Removed it for parseCoverageDataFilePath and readDatFile methods

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Done

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Done

scripts/src/javatests/org/oppia/android/scripts/common/BazelClientTest.kt
Rd4dev: Done

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Got it, Ben. Renamed to `parseCoverageDataFilePath` and for `data` as per https://github.com/oppia/oppia-android/pull/5423#discussion_r1649508527, I renamed that to `coverageCommandOutputLines`.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: My bad! I was messing around with Paths when trying to work with byte arrays - [[old commit](https://github.com/oppia/oppia-android/pull/5423/commits/d410d17326b86178e954f155fa79530efbdaa452#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR180-R182)] and tried replicating the same behaviour without considering a simpler approach. Simplified it now to just be: `File(filePath).readLines()`

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Done, Thanks!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Got it! Done.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Yeah! Changed it to `coverageCommandOutputLines` throughout the implementation. Thanks!

Rd4dev:
> Thanks @Rd4dev! Had a few more thoughts--PTAL.
> 
> Also, if it saves time feel free to avoid linking the commits in your follow-up comments (it's actually not needed at all since GitHub provides pretty easy ways to see the changes between review passes). If you prefer continuing to include them, feel free but it isn't needed for reviewing. :)

Thanks for letting me know! That will make it a bit faster for me to respond. 
Also addressed all the recent review comments too @BenHenning, can you PTAL!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -132,35 +131,28 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
   }
 
   /**
-   * Runs code coverage for the specified Bazel test target
+   * Runs code coverage for the specified Bazel test target.
+   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
```
BenHenning:
```suggestion
   * Runs code coverage for the specified Bazel test target.
   *
   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
```

For readability.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -171,17 +163,8 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
-   *
-   * @param filePath path to the .dat file
-   * @return content of the .dat file as a list of strings
-   */
-  private fun readDatFile(filePath: String?): List<String>? {
-    return filePath?.let { path ->
-      val pathObj = Paths.get(path)
-      File(pathObj.toUri()).readLines()
-    }
+  private fun readDatFile(filePath: String): List<String> {
```
BenHenning: Given how simple this is, I suggest just inlining it and removing the helper at this point.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: Ah makes sense--thanks for cleaning it up!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -130,6 +130,43 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
 
+  /**
+   * Runs code coverage for the specified Bazel test target.
+   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
+   * This can happen due to: Test failures or misconfigurations that prevent the coverage data
+   * from being generated properly.
+   *
+   * @param bazelTestTarget Bazel test target for which code coverage will be run
+   * @return the generated coverage data as a list of strings
+   * or null if the coverage data file could not be parsed
```
BenHenning: Indentation on continued lines for at-clauses should be +4 additional spaces.

BenHenning (commented):
Thanks @Rd4dev! Just a few follow-ups, all nits--PTAL.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Done!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Sure, Done!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Okay!

Rd4dev:
@BenHenning, fixed the nits, can you PTAL!

BenHenning (approved):
Thanks @Rd4dev! This LGTM. Feel free to merge when you're ready to update your chained PR (let me know if you need the commands for that).

Rd4dev:
Hey @BenHenning, as per at https://github.com/oppia/oppia-android/pull/5433#discussion_r1649961400, I have removed the **RunCoverageForTestTarget** in this PR. With this, this PR will now only handle the execution of Bazel commands for coverage and include updates to the TestBazelWorkspace utility to test coverage executions. Can you PTAL!

Additionally, this PR is not from upstream (as this was created before getting permissions). However, PR 1.3 is properly chained and has been updated accordingly with the removal of **RunCoverageForTestTarget** in sync with the upstream **code_coverage_bazel_command_execution** branch. 

> Feel free to merge when you're ready to update your chained PR (let me know if you need the commands for that).

That should be really helpful if you can share them / confirm the below previously provided commands.

You had previously provided some commands addressing the same, if that's the same can I just confirm the process for the chaining with **code_coverage_bazel_command_execution** and **code_coverage_using_filename**, (as per Meeting Notes, June 6):

**base-branch:** code_coverage_bazel_command_execution
**feature-branch:** code_coverage_using_filename

```
git checkout base-branch
git checkout develop
git pull origin develop
git checkout base-branch
git merge develop
git checkout feature-branch
git pull origin feature-branch
git merge base-branch
git push origin feature-branch
```

BenHenning (approved):
Latest changes (removing the per-target script) LGTM.

BenHenning:
Your commands are correct @Rd4dev but I just noticed this PR is used your fork branch, not the one on Oppia Android. You may need to do other steps to get it to sync properly, and it might be easier doing that together over chat.

Please send this back to me once you sync it with the latest develop & when you're ready to merge it; I'll merge it and we can work together to get #5433 properly synced.

Edit: I actually went ahead and synced, so please assign back when CI is passing and you're ready for this to be merged.

Rd4dev:
Thanks @BenHenning, all the CI Checks have passed.

BenHenning:
Looks like another PR has been merged. :) Syncing again.

BenHenning:
Enabling auto-merge.

PR #5426, opened 2024-06-18
Fix part of #5343: Update the test exemption check script & its exemption format

Rd4dev:
## Explanation

Fixes part of #5343

**Project**
[PR 1.2 of Project 4.1]
This PR is based on develop as it had no other dependency with other PRs.

### Changes Implemented
This PR implements updates to enhance the handling of file exemptions and code coverage settings:

1. Updates `script_exemptions.proto` to support two possible states:
   - Exemption from requiring a test file (current behavior).
   - Override of the minimum code coverage to meet a specific percentage (including 0%).

2. Aligns `test_file_exemptions.textproto` with the new proto format.

3. Modifies `TestFileCheck.kt` to integrate the updated proto format.

## Project Context
In Project 4.1, we are introducing code coverage capabilities where certain files may require exemptions from the default minimum code coverage percentage. These exemptions can now be managed in `test_file_exemptions.textproto` by setting the `override_min_coverage_percentage`.

## Implementation Details
Initially, I explored using the `TodoOpenCheck` regeneration structure to implement these changes. However, due to significant differences in regeneration requirements (creating a new proto format to regenerate the exemptions list), conflicts arose with `generate_test_file_assets_list_from_text_protos`. 

Given that the regeneration process is solely aimed at rebuilding the textproto file and no longer serves any other purpose in this project, I opted to manually compile a list of exempted files and used it to generate the new proto format. I then transferred this newly generated format to the textproto file. The regeneration implementation can be found in my [local test_exemption branch](https://github.com/Rd4dev/oppia-android/commit/189a1da34d680e2a9e14b1a3dbeed01bfe751786#diff-fd100a053f2b965df76d989b0d97b5400c051c7fff9507afc0d891cfaf36e238R925)

Post-generation, the regeneration functionality was removed, as it no longer aligns with the project's needs. Further validation and adjustments, if necessary, will be discussed with Ben and implemented.

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
@BenHenning can you PTAL!

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
```diff
@@ -26,7 +26,9 @@ fun main(vararg args: String) {
   // A list of all the files to be exempted for this check.
   // TODO(#3436): Develop a mechanism for permanently exempting files which do not ever need tests.
   val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptiontextProto)
-    .getExemptedFilePathList()
+    .testFileExemptionList
+    .filter { it.testFileNotRequired }
+    .map { it.exemptedFilePath }
```
BenHenning: Can you also add a test in ``TestFileCheckTest`` to verify that a min coverage exemption results in the file _not_ being exempted here, as expected? We aren't going to verify beyond that in ``TestFileCheck`` so it seems useful to verify here.

BenHenning (changes requested):
Thanks @Rd4dev! The approach LGTM--just had one comment for adding a new test to verify the "do not exempt" behavior that ``TestFileCheck`` now has--PTAL.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
Rd4dev: @BenHenning, I would like to confirm the approach that I can take to write the test for this case.
So the existing cases use the `main` call to execute the tests and that internally uses the actual test_file_exemption.textproto values to check the test exemptions

1. In our case to check that the `override_min_coverage_percentage` exempted tests don't interfere with the actual test file checks we might need to have the actual data inside, so would we need to add atleast one `override_min_coverage_percentage` to the actual "test_file_exemptions.textproto" to facilitate this check as done with other test cases or

2. As the entire logic is in `main`, should it be separate to appropriate classes and functions (**eg:** Class `TestFileCheck`, `fun getTestFileExemptions`) so we can call it passing our own "testing_test_file_exemptions.textproto" that includes a `override_min_coverage_percentage` value and use that for testing.

or is there any other much straight forward way to do that?

Rd4dev: So Ben, I refactored the `TestFileCheck` to separate the logic from `main` primarily to have the ability to pass in a custom protobuf (as path) to check the `override_min_coverage_percentage` exempted file is not being exempted from the actual test file check. 

Now it has 2 functions:
1. **`getTestFileExemptionList`** 
   that gets the path of the test file exemption textproto file and returns a list of file paths that are exempted from requiring a test file.
3. **`execute`** 
  that actually executes the test file check.

Utilizing this, I created the test `testTestFileCheck_minCoverageExemptedFile_notExemptedFromTestFileCheck` that sends a custom `proto` we built 
```
test_file_exemption {
  exempted_file_path: "test/TestExemptedFile.kt"
  test_file_not_required: true
}
test_file_exemption {
  exempted_file_path: "coverage/CoverageTestFile.kt"
  override_min_coverage_percent_required: 5
}
```
to send to `TestFileCheck(_, _).getTestFileExemptionList(testFileExemptiontextProto)`, which filters and just returns the list of files that need to be exempted.

### Things I tried:
To not split up the `getTestExemptionList` function as it may result in making it public, but that is necessary or else it will still use the old `protobuf`. So, this, for now, breaks the part of over-exposing, but I guess that is needed here.

Rd4dev:
@BenHenning added the test case to verify that a min coverage exemption results in the file not being exempted, PTAL!

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
```diff
@@ -23,44 +23,79 @@ fun main(vararg args: String) {
...
-    }
+  val testFileCheck = TestFileCheck(repoPath, testFileExemptiontextProto)
+  testFileCheck.execute()
+}
+
+/**
+ * Class for checking the presence of test files in a repository.
+ *
+ * @param repoPath The path of the repo
+ * @param testFileExemptiontextProto the location of the test file exemption textproto file.
```
BenHenning:
```suggestion
 * @param testFileExemptiontextProto the location of the test file exemption textproto file
```

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
```diff
@@ -23,44 +23,79 @@ fun main(vararg args: String) {
...
+   * Retrieves a list of files exempted from requiring a test file based on the specified protocol buffer.
+   *
+   * This function loads the test file exemptions from the provided protocol buffer file path,
+   * filters and returns list of paths for files that are not required for having a test file.
+   *
+   * The files specifically marked as exempt from needing a test file are included,
+   * rather than those exempt from meeting a minimum coverage percentage.
+   *
+   * @param testFileExemptiontextProto the location of the test file exemption textproto file.
+   * @return A list of file paths that are exempted from requiring a test file.
```
BenHenning:
```suggestion
   * @param testFileExemptiontextProto the location of the test file exemption textproto file
   * @return a list of file paths that are exempted from requiring a test file
```

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
```diff
@@ -23,44 +23,79 @@ fun main(vararg args: String) {
...
+   */
+  fun getTestFileExemptionList(testFileExemptiontextProto: String): List<String> {
```
BenHenning: Why is this exposed? Generally, prefer never exposing methods outside the scope of the class.

If it's because it's going to be needed in the coverage scripts, I suggest just duplicating it there since the load logic is quite straightforward (plus we use utilities when sharing functionality between scripts, we usually don't have the scripts be cross-dependent).

scripts/src/javatests/org/oppia/android/scripts/testfile/TestFileCheckTest.kt
```diff
@@ -125,6 +128,32 @@ class TestFileCheckTest {
...
+
+    val coverageTestExemptiontextProto = tempFolder.newFile("test_exemption.pb")
+    coverageTestExemptiontextProto.outputStream().use { (testFileExemptions.writeTo(it)) }
+
+    val result = TestFileCheck(
+      "${tempFolder.root}",
+      "${tempFolder.root}/test_exemption"
+    ).getTestFileExemptionList("${tempFolder.root}/test_exemption")
+
+    assertEquals(listOf("test/TestExemptedFile.kt"), result)
```
BenHenning: Use Truth.assertThat(); we don't use assertEquals().

scripts/src/javatests/org/oppia/android/scripts/testfile/TestFileCheckTest.kt
```diff
@@ -125,6 +128,32 @@ class TestFileCheckTest {
...
+      )
+      .build()
+
+    val coverageTestExemptiontextProto = tempFolder.newFile("test_exemption.pb")
+    coverageTestExemptiontextProto.outputStream().use { (testFileExemptions.writeTo(it)) }
+
+    val result = TestFileCheck(
+      "${tempFolder.root}",
+      "${tempFolder.root}/test_exemption"
+    ).getTestFileExemptionList("${tempFolder.root}/test_exemption")
```
BenHenning: Instead of validating via getTestFileExemptionList, you should validate based on the changed behavior of the script. As-is, this is verifying the internal state of TestFileCheck, not its end user behavior.

I'd imagine the difference is the check shouldn't fail when execute() is called. A quick litmus test would be to remove the exemption for CoverageTestFile and verify that the test correctly fails. This is a good idea to be confident the test is actually verifying what you want it to.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
BenHenning: It shouldn't actually be needed to be exposed--I left a few other comments for considerations that, when taken together, should mean that we don't need to approach the test part quite this way.

Resolving this comment, though, since a test was added.

BenHenning (changes requested):
Thanks @Rd4dev! Took another pass--PTAL.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
Rd4dev: Hhmm... Okay! Again my bad🥲! So for all the time I thought, the **searchFiles = RepositoryFile.collectSearchFiles()** is actually retrieving the original app's repository files and that is why I was trying to create the seperation,
**ie**. to not touch the searchFiles, as the original sources have a seperate script_exemption.textproto list to handle the test file check situation which is different from the custom **test_exemption.pb** list that I provided to it.

So, I didn't have a proper realization that it was actually only searching within the given **repoPath**,
```
val searchFiles = RepositoryFile.collectSearchFiles(
      repoPath = repoPath,
      expectedExtension = ".kt",
      exemptionsList = testFileExemptionList
    )
```
that single thing just clears up everything for me and just realized it is just a simple call to the execute(), and me just complicated the tests.

Sorry for that, I re-wrote the tests with a success and an error case. Hope this provides proper testing to the actual end user behaviour.

scripts/src/javatests/org/oppia/android/scripts/testfile/TestFileCheckTest.kt
Rd4dev: Thank you for providing these insights, Ben. I believe the current approach implemented in tests should adequately cover the required validation. I'll await your review to see if any additional tests are necessary.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
Rd4dev: So one another thing, is the original source has KDocs for the private methods - just wondering if they have separate standards / practice or is that something to remove or can I just ignore them as it is not related : https://github.com/oppia/oppia-android/blob/dcde7b359e3f3ba63f50f7585b45ff6ef6b12440/scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt#L67-L97

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
Rd4dev: Done!

Rd4dev:
Thanks for the review and suggestions @BenHenning, added the required test cases for validating coverage exemption checks too, can you PTAL!

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
Rd4dev: Recent changes removed the function itself, so these are no longer available.

scripts/src/javatests/org/oppia/android/scripts/testfile/TestFileCheckTest.kt
Rd4dev: Okay Ben. I instead replaced it with `assertThat` checks.
Would this practice also apply to sample test cases that I created in `TestBazelWorkspaceTest`, `BazelClientTest`, and other utility Tests that I introduced recently. I kinda utilized assertEquals extensively over there! [[commit](https://github.com/oppia/oppia-android/pull/5423/files#diff-e50d7ab1968e374641f7298e27dca62ec182a42dc06dac4c47857cd7526de00fR407-R419)]

BenHenning: I think it's fine there since it avoids extra dependency complexity in setting up the tests (which is why I didn't comment on it); feel free to keep those as-is.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
BenHenning: Good question--it's fine to remove it (we shouldn't have added it originally, but it was missed during review). However, it's often easier just to maintain existing code and only clean it up if we're already making changes to it. Since this PR is moving code, either outcome seems reasonable.

BenHenning (approved):
Thanks @Rd4dev! This LGTM.

BenHenning:
Feel free to merge this @Rd4dev if you don't intend to make any follow-up changes. If you do make follow-up changes, please assign back so I can look at them.

Rd4dev:
Hey @BenHenning, I removed the kDocs for the private methods in `TestFileCheck.kt`. Can you PTAL? Also, I'm wondering if I can merge this if it's not from upstream?

BenHenning (approved):
Thanks @Rd4dev! Seems fine to merge, so going ahead and merging it. Latest changes LGTM.

PR #5248, opened 2023-11-28
 Fix part of #4865 and Fix #4986 : Move all fragment arguments, activity intent extras, and saved instance state over to protos

Vishwajith-Shettigar:
## Explanation
Fixes #4986 
Move all fragment arguments, activity extras, and saved instance state bundle usage over to using protos entirely.

Fix part of #4865
Introduce a ProfileID decorator to pack ProfileId into all Intents so that it can be easily accessed by underlying fragments and child classes. Some classes already use this utility where it was not easy to add profileId to the intent proto bundle.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #5433, opened 2024-06-20
Fix part of #5343: Implement script to run code coverage for a specific file

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 1.3 of Project 4.1]

### Changes Made

- Added a new script utility `RunCoverage.kt` that takes a filename as an argument.
- Implements handling for exempted files and test target mapping.

---

### Implementation Details

- **Exemption Check**: The script first checks if the provided file is exempted from having test file. If exempted, it exits early with the message "The file is exempted hence no coverage!".
  
- **Test Target Mapping**:
  - **Script, App, and Default Subpackages**: Constructs mappings for files within these subpackages to their corresponding test targets.
  - **App Subpackage**: Handles mapping for both `test` and `Localtests`.
  
- **Code Coverage Execution**:
  - Utilizes the `RunCoverageForTestTarget` script to execute code coverage analysis.
  - Returns a list of coverage data strings for each run, which will be subsequently saved as protos in PR 1.4.

### Example Usage

```bash
bazel run //scripts:run_coverage -- $(pwd) utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
```

## Essential Checklist

## For UI-specific PRs only


scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * Example:
+ *     bazel run //scripts:run_coverage -- $(pwd)
+ *     utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
+ * Example with custom process timeout:
+ *     bazel run //scripts:run_coverage -- $(pwd)
+ *     utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt processTimeout=10
+ *
+ */
+fun main(vararg args: String) {
```
Rd4dev, line 30: With the introduction of the entry point in RunCoverage, may be I can remove that in RunCoverageForTestTarget https://github.com/oppia/oppia-android/pull/5423/files#diff-04f7dc1ca7f9e9b2f3eac2fe9e9de7841146490f96e68f9059223c8c85588bfaR28 as that will no longer be used.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * Finds potential test file paths corresponding to a given source file path within a repository.
+   *
+   * @param repoRoot the root directory of the repository
+   * @param filePath The file path of the source file for which the test files are to be found.
+   * @return A list of potential test file paths that exist in the repository.
+   */
+  fun findTestFile(repoRoot: String, filePath: String): List<String> {
```
Rd4dev: TODO: This needs to be made private and update the tests to handle cases with different subpackage paths to handle this behaviour.

Rd4dev:
Hey @BenHenning,
With PR 1.3, I've completed the logic implementation (updated with recent changes to PR 1.1 bazelClient returning the coverage data as `List<String>`. The tasks remaining are to update the tests (handling multi module test) (and to make `findTestFile` private and its related test cases). Could you please provide an initial review to see if any logical changes are needed?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Why not update ``RunCoverageForTestTarget`` instead of creating a new file? I'm a bit confused about the implementation order, at the moment. Why introduce ``RunCoverageForTestTarget`` at all?

Also, perhaps 'ComputeCoverageForSourceFile' might be a clearer name for this script--WDYT?

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageRunnerTest.kt
```diff
@@ -37,7 +37,7 @@ class CoverageRunnerTest {
   }
 
   @Test
-  fun testRunCoverage_emptyDirectory_throwsException() {
+  fun testCoverageRunner_emptyDirectory_throwsException() {
```
BenHenning: The test name format that I encourage is:

``testAction_withCondition_andCondition_hasOutcome``

In this case, the action is 'runWithCoverageAsync'. I suggest renaming based on that for better clarity. It should be contextually evident that ``CoverageRunner`` is being tested here since these tests are in ``CoverageRunnerTest``.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+    RunCoverage(repoRoot, filePath, commandExecutor, scriptBgDispatcher).execute()
+  }
+}
+
+/**
+ * Class responsible for executing coverage on a given file.
+ *
+ * @param repoRoot the root directory of the repository
+ * @param filePath the relative path to the file to analyse coverage
+ * @param commandExecutor Executes the specified command in the specified working directory
```
BenHenning:
```suggestion
 * @param commandExecutor executes the specified command in the specified working directory
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+   * a Bazel client, finds potential test file paths, retrieves Bazel targets, and initiates
+   * coverage analysis for each test target found.
+   */
+  fun execute(): MutableList<List<String>> {
+    var coverageDataList = mutableListOf<List<String>>()
+    val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptionTextProto)
+      .getExemptedFilePathList()
+
+    val isExempted = testFileExemptionList.contains(filePath)
+    if (isExempted) {
```
BenHenning:
```suggestion
    if (filePath in testFileExemptionList) {
```

Can be simplified a bit.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * Executes coverage analysis for the specified file.
+   *
+   * Loads test file exemptions and checks if the specified file is exempted. If exempted,
+   * prints a message indicating no coverage analysis is performed. Otherwise, initializes
+   * a Bazel client, finds potential test file paths, retrieves Bazel targets, and initiates
+   * coverage analysis for each test target found.
+   */
+  fun execute(): MutableList<List<String>> {
```
BenHenning: Why is a mutable list being returned? We should generally never return mutable structures unless it makes sense for the caller to mutate the result (which would be a very specific and pretty complicated requirement to do correctly, so let's avoid that if we can).

Also, the method output should be documented.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+      return mutableListOf()
+    }
+
+    val testFilePaths = findTestFile(repoRoot, filePath)
+    val testTargets = bazelClient.retrieveBazelTargets(testFilePaths)
+
+    for (testTarget in testTargets) {
+      val coverageData = RunCoverageForTestTarget(
+        rootDirectory,
+        testTarget.substringBeforeLast(".kt"),
```
BenHenning:
```suggestion
        testTarget.removeSuffix(".kt"),
```

This is probably a bit clearer.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+        commandExecutor,
+        scriptBgDispatcher
+      ).runCoverage()!!
+      coverageDataList.add(coverageData)
+    }
+    println("Coverage Data List: $coverageDataList")
+    return coverageDataList
```
BenHenning: I'm confused about the output here. Per the implementation plan, shouldn't this script be producing a result proto for the source file that combines the results from the parsed lcov dat files?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Finds potential test file paths corresponding to a given source file path within a repository.
+   *
+   * @param repoRoot the root directory of the repository
+   * @param filePath The file path of the source file for which the test files are to be found.
+   * @return A list of potential test file paths that exist in the repository.
+   */
+  fun findTestFile(repoRoot: String, filePath: String): List<String> {
+    val file = File(filePath)
+    val parts = file.parent.split(File.separator)
```
BenHenning: It should be fine just using '/', I think (which you're already doing below, so it's cleaner to avoid the inconsistency and instead commit to a single standard).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+        testFiles.add(localTestFilePath)
+      }
+    } else {
+      val defaultTestFilePath = filePath.replace("/main/", "/test/").replace(".kt", "Test.kt")
+      if (File(repoRoot, defaultTestFilePath).exists()) {
+        testFiles.add(defaultTestFilePath)
+      }
+    }
+    return testFiles
+  }
```
BenHenning:
```suggestion
    val possibleTestFilePaths = when {
      filePath.startsWith("scripts/") ->
        listOf(filePath.replace("/java/", "/javatests/").replace(".kt", "Test.kt"))
      filePath.startsWith("app/") -> {
        listOf(
          filePath.replace("/main/", "/sharedTest/").replace(".kt", "Test.kt"),
          filePath.replace("/main/", "/test/").replace(".kt", "Test.kt"),
          filePath.replace("/main/", "/test/").replace(".kt", "LocalTest.kt")
        )
      }
      else -> listOf(filePath.replace("/main/", "/test/").replace(".kt", "Test.kt"))
    }
    return possibleTestFilePaths.map { File(repoRoot, it) }.filter(File::exists)
  }
```

Some Kotlin patterns can simplify this somewhat by removing duplication and avoiding using a mutable structure, and possibly make it a bit easier to follow as a reader.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,154 @@
...
+    val builder = TestFileExemptions.getDefaultInstance().newBuilderForType()
+
+    // This cast is type-safe since proto guarantees type consistency from mergeFrom(),
+    // and this method is bounded by the generic type T.
+    @Suppress("UNCHECKED_CAST")
+    val protoObj: TestFileExemptions =
+      FileInputStream(protoBinaryFile).use {
+        builder.mergeFrom(it)
+      }.build() as TestFileExemptions
+    return protoObj
```
BenHenning:
```suggestion
    return File("$testFileExemptiontextProto.pb").inputStream().use { stream ->
      TestFileExemptions.newBuilder().also { builder ->
        builder.mergeFrom(stream)
      }.build()
    }
```

Can be simplified a bunch, and no unsafe cast is needed.

BenHenning (changes requested):
Thanks @Rd4dev! I skipped ``RunCoverageTest`` for my first pass review, but I looked at everything else. Most of my suggestions are nits or suggestions for clarity, however I am a bit confused about the data flow for the new script (and, in particular, what it's actually outputting and how that affects later PRs). PTAL.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Yeah! actually the `RunCoverageForTestTarget` script now just acts as an intermediary for passing data and nothing useful. Initially, I followed it to stick with what we decided, ie. to have '3 scripts' when we altered our initial plan. Originally, we had it planned to provide test target inputs and obtain coverage reports. However, we later decided it would be more user-friendly to use filename inputs instead.

With this change, we specified having three separate scripts:
1. Running coverage analysis.
2. Running a single target.
3. Running coverage for a specific file.

I viewed this as a separation of concerns and proceeded accordingly. However, the `bazelClient` already handles running coverage for a single test target, making this script redundant—it's now just an intermediary and nothing more.

Would you prefer I remove this script and consolidate it into two utilities instead: **RunCoverage** and **CoverageRunner**? We could possibly introduce a **CoverageReporter** later, perhaps with **PR 1.5** for generating reports.

Rd4dev: > Also, perhaps 'ComputeCoverageForSourceFile' might be a clearer name for this script--WDYT?

Uh... just few thoughts on why I initially preferred keeping it **runCoverage**, for the actual script run is to keep it simple and easy for developers to remember. So, I thought we could keep it abstract, as developers would follow the wiki instructions and not everyone (especially those coming from gradle using bazel just for coverages) would delve into whether it's running for source files or test targets.

However, naming it runCoverage might cause confusion for those who are working with bazel, as they might expect it to run test targets so having a proper naming would work. But this wouldn't mean much as over time, developers would get used to the command, so a descriptive command as you mentioned should be the way to go.

I've been used to calling it runCoverage for a while now (feels like trying to rename my childhood teddy bear 😅), but having a descriptive name would make things much clearer too, so if confirmed, I can refactor it while simplifying the scripts into one.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Made as private! with altering related tests

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageRunnerTest.kt
Rd4dev: Ah Okay! Renamed them to: 

- testRunWithCoverageAsync_emptyDirectory_throwsException
- testRunWithCoverageAsync_invalidTestTarget_throwsException
- testRunWithCoverageAsync_validSampleTestTarget_returnsCoverageData

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Done!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Done! Simplified as mentioned.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: So, I had it as a `MutableList` to hold `coverageDataList` to which we add coverage data results. I understand the practice to avoid mutable structure. So, in the revised approach the coverageDataList is still mutable to allow adding the coverage data, and it is then converted to an immutable list using `toList()` and passed as a result.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Got it! Done.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: With the new refactored structure utilizing Kotlin patterns this implementation is removed now!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks @BenHenning, that was super helpful! Updated with the provided pattern with the addition to just convert and return them as relative paths.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Okay, I just duplicated them from the existing implementations. Now they are updated as suggested. Thanks!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Ah yes, Ben. Since Proto generation is planned for PR 1.4, I kept the coverage data results as a list of strings for now. This approach allows to check and test for accurate content more easily. Additionally, it closely resembles the flow of what we aim to achieve with Proto generation in PR 1.4. With PR 1.4, I just had to update these results to Protos and verify them accordingly. [[PR 1.4 proto generation](https://github.com/oppia/oppia-android/pull/5439)]. So is that how we planned? and also want to make sure if these PRs are in the right track?

Rd4dev:
Hi @BenHenning,

I have addressed the comments with one thing to maybe just have 2 scripts, removing the redundant coverage script (`runCoverage`, `runCoverageForTestTarget`). Still left to confirm, and I will just reformat that. Other than that, everything else is done. Here are a few changes:

- Changed the visibility of `findTestFile` and moved its functionality to be handled through public calls for separate app, scripts, and default subpackages.
- Converted the mutable list return to immutable lists.
- Other code simplifications.

However, I still have one issue with this PR, which I mentioned earlier: **having tests run on multiple packages**. 
____

### Here is what works:
When I have tests run on:

**sourceSubpackage = "coverage/main/java/com/example"
testSubpackage = "coverage/test/java/com/example"**

They successfully provide the results. The flow happens like this:

1. We give the source file as **coverage/main/java/com/example/TwoSum.kt**
2. It is mapped to its test files through findTestFile and returns **coverage/test/java/com/example/TwoSumTest.kt**
3. Then it retrieves its test target as **//coverage/test/java/com/example:TwoSumTest.kt** (where we clip off the .kt)
4. It is finally sent to bazelClient's `runCoverageForTestTarget`, which eventually returns the results

This flow works smoothly with having subpackages as:

1. default package name:
**sourceSubpackage = "coverage/main/java/com/example"
testSubpackage = "coverage/test/java/com/example"**

In the test case `testRunCoverage_validSampleTestFile_returnsCoverageData`, as mentioned, I also tried to test the same with the script subpackage, having:

2. scripts package name:
**sourceSubpackage = "scripts/java/com/example"
testSubpackage = "scripts/javatests/com/example"**

Providing filename as **scripts/java/com/example/TwoSum.kt** in the test case `testRunCoverage_validScriptPathSampleTestFile_returnsCoverageData`, the same also works with the app subpackage (but only with /test/ subpackages) having it as:

3. app package name:
**sourceSubpackage = "app/main/java/com/example"
testSubpackage = "app/test/java/com/example"**

And providing the filename as **app/main/java/com/example/TwoSum.kt** in the test case `testRunCoverage_validAppPathSampleTestFile_returnsCoverageData`.

___
### Here is what didn't work:
But the place where it fails is when I try to introduce the subpackage **sharedTest** within the app. When I try to use:

**sourceSubpackage = "app/main/java/com/example"
testSubpackage = "app/sharedTest/java/com/example"**
(this test is necessary to test the functionality of coverage execution with existing app module tests that are having multiple tests including a sharedTest Test.kt and / or test LocalTest.kt)

As done in the reference fail commit [commit](https://github.com/oppia/oppia-android/pull/5433/commits/6720de78593da40404368d27603918e9d75e4c8c#diff-5a779ea4a8aa2caf5249ced27465f76b8ac19bed5cf250475f6033dfad464908R384), the coverage returns null. 
- It successfully retrieves the file, 
- queries the Bazel test targets, and 
- gives **//app/sharedTest/java/com/example:TwoSumTest.kt** but 
- fails to retrieve the coverage. 

___
### Things I tried:
I doubt if I am missing any configuration that needs to be done to make coverage work on packages that don't follow the regular pattern for **src/test/java/com**.

One thing I tried doing is adding the **visibility** to sharedTest - [commit](https://github.com/oppia/oppia-android/pull/5433/commits/e4ede829edd60af932a6c36b06d74f38d19a61c7#diff-af480691578c810595659968b4cc9141b6cb4ca63c81cbf6fb3d05460f8f471bR124) but that went in vain.

Can I understand what I might be missing? 
**Note:** another thing I tried was adding a new functionality to TestBazelWorkspace utiility `addMultiLevelSourceAndTestFileWithContent` that adds a local, shared, and source file specific to this case. The local tests LocalTest.kt returns the expected result; it's only with the sharedTest - [commit](https://github.com/oppia/oppia-android/pull/5433/files#diff-af480691578c810595659968b4cc9141b6cb4ca63c81cbf6fb3d05460f8f471bR96) that fails.

I thought maybe the files need to have the test subpackage mandatory for Bazel coverage to find it, so I tried providing the subpackage as app/test - [commit](https://github.com/oppia/oppia-android/pull/5433/commits/985e861e342d162231e695a999a0728e982d2f04), but it's the same when I try to make it sharedTest.

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
```diff
@@ -83,6 +83,51 @@ class TestBazelWorkspace(private val temporaryRootFolder: TemporaryFolder) {
...
+   * Adds a source file and 2 test files with the specified name and content,
+   * and updates the corresponding build configuration.
+   *
+   * @param filename the name of the source file (without the .kt extension)
+   * @param sourceContent the content of the source file
+   * @param testContentShared the content of the test file for SharedTest Package
+   * @param testContentLocal the content of the test file for Test Package
+   * @param subpackage the subpackage under which the source and test files should be added
+   */
+  fun addMultiLevelSourceAndTestFileWithContent(
```
BenHenning, line 96: Reminder to add tests for this new method.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Thanks for this context--it's really helpful! I think keeping the current name is fine--that suggestion was more to differentiate from having multiple scripts.

I think the simplest user experience is probably to have no more than two scripts at the end of the project:
- One for generating a coverage report for a selected target or targets (based on what we need).
- One for taking one or more coverage reports and generating a combined summary in MD or HTML.

I could see the need for a third script to perform the report combining separately from the MD/HTML generation, but I actually don't see a need for a different script for actually running coverage since we can always extend it to support N targets.

Would the plan then be to keep this PR as-is and remove RunCoverageForTestTarget upstream?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Avoiding mutability seems like the best approach, then, given the context provided.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: I think that's fine, then. I personally find it a bit easier when splitting up PRs to never split up the dataflow at the end of its pipeline since changing it throughout multiple PRs can be awkward, but it does make sense to continue as currently planned.

BenHenning (commented):
Thanks @Rd4dev! Will take a full pass on RunCoverage and RunCoverageTest after the design decision question about RunCoverageForTestTarget is addressed.

BenHenning:
@Rd4dev re: your comment above, what do you mean by 'coverage is failed to be retrieved'? I take this to mean that null is returned, but have you looked at the full Bazel command output (e.g. by printing it out in BazelClient when running the test)?

You could also try recreating the exact workspace that the test since (since it's pretty simply; you can just print out the exact files being written or write a bit of code in the test to output the directory structure & file contents in the test workspace) and then running the exact Bazel command that is being run in the test. This would provide you a much faster iteration loop to try things and try to figure out what's going wrong, and what needs to change (which is then hopefully adaptable to the test environment).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Yeah Ben, I think we can remove **RunCoverageForTestTarget** and I proceeded to do that and updated this PR too.
[[PR 1.1 commit](https://github.com/oppia/oppia-android/pull/5423/commits/c834aa77f3e6e71b3f15dcf7d606a2f889b4271b)]

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
Rd4dev: Thanks, added tests for checking the source and build files with `addMultiLevelSourceAndTestFileWithContent`.

Rd4dev:
> @Rd4dev re: your comment above, what do you mean by 'coverage is failed to be retrieved'? I take this to mean that null is returned, but have you looked at the full Bazel command output (e.g. by printing it out in BazelClient when running the test)?
> 
> You could also try recreating the exact workspace that the test since (since it's pretty simply; you can just print out the exact files being written or write a bit of code in the test to output the directory structure & file contents in the test workspace) and then running the exact Bazel command that is being run in the test. This would provide you a much faster iteration loop to try things and try to figure out what's going wrong, and what needs to change (which is then hopefully adaptable to the test environment).

Thanks a lot, @BenHenning. Recreating the workspace and doing testing with that probably saved me a ton of time. Here's what happened:

When running coverage with **sharedTest**, the coverage command ran but failed to generate coverage, returning "**There was no coverage.**" Since **no coverage.dat** file was produced, it returned **null**.

After investigating, I found that Bazel couldn't figure out which package it was looking for. By setting the instrumentation filter, Bazel was able to determine where exactly to inspect. So, I added **--instrumentation_filter=//** to the BazelClient's command to inspect everywhere, which achieved the desired results.

One thing to note is that this also includes coverage related to tests (_But I wonder if tests have their own coverage_). For example, when I tried to get coverage with the instrumentation filter for the target //utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest, it generated coverage results for both SF:utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt and SF:utility/src/test/java/org/oppia/android/util/parser/math/MathModelTest.kt. 

<details>
<summary>Log</summary>
<br>
cat .../testlogs/utility/src/test/java/org/oppia/android/util/parser/math/MathModelTest/coverage.dat
SF:utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
FN:58,org/oppia/android/util/parser/math/MathModel$MathModelSignature$Companion::createSignature$utility_src_main_java_org_oppia_android_util_parser_math_math_latex_model_kt (Ljava/lang/String;FZ)Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;
FN:33,org/oppia/android/util/parser/math/MathModel$MathModelSignature::<init> (Ljava/lang/String;IZ)V
FN:35,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getLineHeightHundredX ()I
FN:34,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getRawLatex ()Ljava/lang/String;
FN:36,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getUseInlineRendering ()Z
FN:41,org/oppia/android/util/parser/math/MathModel$MathModelSignature::updateDiskCacheKey (Ljava/security/MessageDigest;)V
FN:14,org/oppia/android/util/parser/math/MathModel::<init> (Ljava/lang/String;FZ)V
FN:16,org/oppia/android/util/parser/math/MathModel::getLineHeight ()F
FN:15,org/oppia/android/util/parser/math/MathModel::getRawLatex ()Ljava/lang/String;
FN:17,org/oppia/android/util/parser/math/MathModel::getUseInlineRendering ()Z
FN:21,org/oppia/android/util/parser/math/MathModel::toKeySignature ()Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;
FNDA:1,org/oppia/android/util/parser/math/MathModel$MathModelSignature$Companion::createSignature$utility_src_main_java_org_oppia_android_util_parser_math_math_latex_model_kt (Ljava/lang/String;FZ)Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;
FNDA:1,org/oppia/android/util/parser/math/MathModel$MathModelSignature::<init> (Ljava/lang/String;IZ)V
FNDA:1,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getLineHeightHundredX ()I
FNDA:0,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getRawLatex ()Ljava/lang/String;
FNDA:1,org/oppia/android/util/parser/math/MathModel$MathModelSignature::getUseInlineRendering ()Z
FNDA:1,org/oppia/android/util/parser/math/MathModel$MathModelSignature::updateDiskCacheKey (Ljava/security/MessageDigest;)V
FNDA:1,org/oppia/android/util/parser/math/MathModel::<init> (Ljava/lang/String;FZ)V
FNDA:0,org/oppia/android/util/parser/math/MathModel::getLineHeight ()F
FNDA:0,org/oppia/android/util/parser/math/MathModel::getRawLatex ()Ljava/lang/String;
FNDA:0,org/oppia/android/util/parser/math/MathModel::getUseInlineRendering ()Z
FNDA:1,org/oppia/android/util/parser/math/MathModel::toKeySignature ()Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;
FNF:11
FNH:7
BRDA:46,0,0,1
BRDA:46,0,1,1
BRF:2
BRH:2
DA:14,1
DA:15,1
DA:16,1
DA:17,1
DA:21,1
DA:33,1
DA:34,1
DA:35,1
DA:36,1
DA:41,1
DA:42,1
DA:43,1
DA:44,1
DA:45,1
DA:46,1
DA:47,1
DA:49,1
DA:58,1
DA:59,1
LH:19
LF:19
end_of_record
SF:utility/src/test/java/org/oppia/android/util/parser/math/MathModelTest.kt
FN:9,org/oppia/android/util/parser/math/MathModelTest::<init> ()V
FN:90,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_diffModelByLineHeight_outsideTwoDecimals_returnsDiffKeyWithDiffDigest ()V
FN:69,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_diffModelByLineHeight_withinTwoDecimals_returnsSameKeyWithSameDigest ()V
FN:109,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByInlineRendering_returnsDifferentKeyWithDifferentDigest ()V
FN:31,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByLatex_returnsDifferentKeyWithDifferentDigest ()V
FN:50,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByLineHeight_returnsDifferentKeyWithDifferentDigest ()V
FN:13,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_sameModelByValues_returnsSameKeyWithSameDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::<init> ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_diffModelByLineHeight_outsideTwoDecimals_returnsDiffKeyWithDiffDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_diffModelByLineHeight_withinTwoDecimals_returnsSameKeyWithSameDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByInlineRendering_returnsDifferentKeyWithDifferentDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByLatex_returnsDifferentKeyWithDifferentDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_differentModelByLineHeight_returnsDifferentKeyWithDifferentDigest ()V
FNDA:1,org/oppia/android/util/parser/math/MathModelTest::testToKeySignature_sameModelByValues_returnsSameKeyWithSameDigest ()V
FNF:7
FNH:7
DA:9,1
DA:10,1
DA:13,1
DA:14,1
DA:15,1
DA:16,1
DA:18,1
DA:19,1
DA:20,1
DA:21,1
DA:23,1
DA:24,1
DA:25,1
DA:26,1
DA:27,1
DA:31,1
DA:32,1
DA:33,1
DA:34,1
DA:36,1
DA:37,1
DA:38,1
DA:39,1
DA:42,1
DA:43,1
DA:44,1
DA:45,1
DA:46,1
DA:50,1
DA:51,1
DA:52,1
DA:53,1
DA:55,1
DA:56,1
DA:57,1
DA:58,1
DA:61,1
DA:62,1
DA:63,1
DA:64,1
DA:65,1
DA:69,1
DA:70,1
DA:71,1
DA:72,1
DA:73,1
DA:75,1
DA:76,1
DA:77,1
DA:78,1
DA:82,1
DA:83,1
DA:84,1
DA:85,1
DA:86,1
DA:90,1
DA:91,1
DA:92,1
DA:93,1
DA:95,1
DA:96,1
DA:97,1
DA:98,1
DA:101,1
DA:102,1
DA:103,1
DA:104,1
DA:105,1
DA:109,1
DA:110,1
DA:111,1
DA:112,1
DA:113,1
DA:115,1
DA:116,1
DA:117,1
DA:118,1
DA:121,1
DA:122,1
DA:123,1
DA:124,1
DA:125,1
LH:82
LF:82
end_of_record
</details>

However, this shouldn't hinder our execution since we will only keep the single related source file coverage result for the final report, can you PTAL!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -144,7 +144,8 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
   fun runCoverageForTestTarget(bazelTestTarget: String): List<String>? {
     val coverageCommandOutputLines = executeBazelCommand(
       "coverage",
-      bazelTestTarget
+      bazelTestTarget,
+      "--instrumentation_filter=//"
```
BenHenning: Is this actually needed? I worry that this will substantially slow down performance for coverage, and we technically can normally defer to Bazel to select the correct instrumentation filter for the target, can't we?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * @return a list of lists containing coverage data for each requested test target, if
+   *     the file is exempted from having a test file, an empty list is returned
+   */
+  fun execute(): List<List<String>> {
+    val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptionTextProto)
+      .getExemptedFilePathList()
+
+    if (filePath in testFileExemptionList) {
+      println("This file is exempted from having a test file. Hence No coverage!")
```
BenHenning:
```suggestion
      println("This file is exempted from having a test file; skipping coverage check.")
```

Possible clarification in warning output.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -0,0 +1,435 @@
...
+        "LF:4",
+        "end_of_record"
+      )
+    )
+
+    assertThat(result).isEqualTo(expectedResultList)
+  }
+
+  @Test
+  fun testRunCoverage_validScriptPathSampleTestFile_returnsCoverageData() {
```
BenHenning: I'm not sure that I fully understand what is meant by this condition (valid script path sample test file) vs. the one above or tests below. Could you maybe clarify in the test names exactly the differences between the tests?

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -0,0 +1,435 @@
...
+import org.junit.rules.TemporaryFolder
+import org.oppia.android.scripts.common.CommandExecutorImpl
+import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
+import org.oppia.android.scripts.testing.TestBazelWorkspace
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+import java.util.concurrent.TimeUnit
+
+/** Tests for [RunCoverage]. */
+class RunCoverageTest {
```
BenHenning, line 19: Other test suggestions:
- Invalid source file
- Missing test file (without exemption)
- Tests to verify test/, local test, and sharedTest/ logic all works correctly (check by commenting out the bits in your script to verify the correct tests fail)

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: SGTM--this comment is addressed from my perspective but technically you're the one that opened it, so please resolve it once you're happy with things.

BenHenning (changes requested):
Thanks @Rd4dev! Took a look at the rest of the PR (it's now fully reviewed) and had a few follow-up suggestions--PTAL.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Okay Done!

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Included source code to handle cases when 
- a invalid / no available source file path is provided (added test)
- cases there are no test files for the provided path (added test)
- Not sure what this means the existing `testRunCoverage_sharedAndLocalTests_returnsCoverageData` (this was previously named as `testRunCoverage_validMultiPathSampleTestFile_retrunsCoverageData`) handles the scenario of testing the app module when they have multiple test files (shared/localtest)

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Now the instrumentation is derived from the filepath to give `//$module_name/...`, this can be a bit more efficient, may be?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Made it to **"This file is exempted from having a test file; skipping coverage check."**

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: The test cases are for scenarios to test for various modules
1. script module needs to have the `javatests/...`
2. app module have `test/.../___Test.kt `or `test/.../___LocalTest.kt` or `sharedTest/...`
3. default behaviour should correspond to `test/...`

And this particular `testRunCoverage_validScriptPathSampleTestFile_returnsCoverageData` test case checks the scenario if a file provided from the scripts module correctly maps its test file and Bazel targets and executes them. Same with others for respective app, default behaviours

And renamed them to just:
`testRunCoverage_validScriptPathSampleTestFile_returnsCoverageData` -> `testRunCoverage_scriptTests_returnsCoverageData` (similarly others)

Rd4dev:
@BenHenning, modified the instrumentation and added handling of edge cases (included tests for them), can you PTAL!

Edit: Also changed the base branch to develop, hope that's done right, if yes then I will proceed to delete the upstream from Oppia Android.

Rd4dev:
@BenHenning @adhiamboperes, so there was a TodoCheck failure happening in the file: TopicPracticeFragment.kt [[stack trace](https://github.com/oppia/oppia-android/actions/runs/9672279680/job/26684317915?pr=5433)], since the merge of https://github.com/oppia/oppia-android/pull/5248, can you PTAL or maybe I can just create a separate PR to just remove that Todo if that's done.

Edit: After merging with develop all the Checks have passed. So, I think this irrelevant now.
Will revert back the Assignments, thanks!

scripts/src/java/org/oppia/android/scripts/coverage/BUILD.bazel
```diff
@@ -4,6 +4,20 @@ Libraries corresponding to developer scripts that obtain coverage data for test
...
+kt_jvm_library(
+    name = "run_coverage_lib",
+    testonly = True,
+    srcs = [
+        "RunCoverage.kt",
+    ],
+    visibility = ["//scripts:oppia_script_binary_visibility"],
+    deps = [
+        "//scripts/src/java/org/oppia/android/scripts/common:bazel_client",
+        "//scripts/src/java/org/oppia/android/scripts/coverage:coverage_runner",
```
Rd4dev: TODO: "//:coverage_runner",

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -0,0 +1,148 @@
...
+  ScriptBackgroundCoroutineDispatcher().use { scriptBgDispatcher ->
+    val processTimeout: Long = args.find { it.startsWith("processTimeout=") }
+      ?.substringAfter("=")
+      ?.toLongOrNull() ?: 5
+
+    val commandExecutor: CommandExecutor = CommandExecutorImpl(
+      scriptBgDispatcher, processTimeout = processTimeout, processTimeoutUnit = TimeUnit.MINUTES
+    )
+
+    println(RunCoverage(repoRoot, filePath, commandExecutor, scriptBgDispatcher).execute())
```
Rd4dev: TODO: remove print statement

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: So, is the suggestion to add test for local, sharedTests separately alongside with multi tests case, I think then will probably update the TestBazelWorkspace and implement this.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -40,6 +42,28 @@ class RunCoverageTest {
...
+    testBazelWorkspace.initEmptyWorkspace()
+    val exception = assertThrows<IllegalStateException>() {
+      main(tempFolder.root.absolutePath, "file.kt")
+    }
+
+    assertThat(exception).hasMessageThat().contains("File doesn't exist")
+  }
+
+  @Test
+  fun testRunCoverage_missingTestFileNOtExempted_throwsException() {
```
BenHenning:
```suggestion
  fun testRunCoverage_missingTestFileNotExempted_throwsException() {
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -31,6 +31,10 @@ fun main(vararg args: String) {
   val repoRoot = args[0]
   val filePath = args[1]
 
+  if (!File(repoRoot, filePath).exists()) {
+    error("File doesn't exist.")
```
BenHenning:
```suggestion
    error("File doesn't exist: $filePath.")
```

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
BenHenning: The renamed methods are much clearer--thanks!

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: Hmm not sure that I have a better suggestion, so I think it's fine to go with this for now. There are probably things we could do, but per our discussion it sounds like the best option is just to remove this filter entirely once we no longer can have multiple tests for a single source file.

BenHenning (approved):
Thanks @Rd4dev! Just had a couple of comments.

Note that I'm approving the PR, but please have @adhiamboperes verify the last couple comments before merging (just to make sure it's reviewed by someone).

BenHenning:
Also well done on getting this branched onto develop correctly @Rd4dev. :)

Rd4dev:
Thank you so much @BenHenning, will make sure to request @adhiamboperes on any further progress.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Done!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Done!

adhiamboperes (approved):
Thanks @Rd4dev! I have take a full pass, and resolved the remaining comments. LGTM!

adhiamboperes:
Synced with develop and enabling auto-merge.

PR #5439, opened 2024-06-21
Fix part of #5343: Building proto with coverage results for data processing

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 1.4 of Project 4.1]

### Changes Made
- Introduced `coverage.proto` to store the coverage data generated during coverage analysis.
- Parsed coverage data according to the LCOV (geninfo) file format specifications: [LCOV geninfo specifications](https://manpages.debian.org/stretch/lcov/geninfo.1.en.html#FILES)
- This maps the Bazel test target to the corresponding SF: source file path and filters the data to generate the Coverage Report proto that will only the coverage of the tests related to the single source file.
- With the previous implementation in PR 1.3 to get the related LocalTest and Test file mappings, we utilize that to run coverage for both the test and return them as a list of Coverage Report protos.

**To Note:**
- Only the app module has the multi test file case.
- But as of date, the many app module coverage executions fail.
- So to test acquiring multi test coverage execution as list of Coverage Report Proto, I hardcoded and tested the implementation [[hardcoded comment](https://github.com/oppia/oppia-android/pull/5439/files#diff-62c81482e7c4b720da2d4efc73a48d10369eee097e0bf3439def6648ff150c23R89-R93)] (just for testing the actual code wouldn't have that)

**Result:**
A representation of the list of Coverage Report proto is presented below:

This is the output on running 2 hardcoded test cases:
`math:MathModelTest` and `math:FloatExtensionsTest`
<details>
<summary>List of Coverage Report</summary>
<br>

<pre>
Parsed Coverage Data File: /home/rddev/.../bazel-out/.../util/parser/math/MathModelTest/coverage.dat
Parsed Coverage Data File: /home/rddev/.../bazel-out/.../util/math/FloatExtensionsTest/coverage.dat
Coverage Data List: [bazel_test_target: "//utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest"
covered_file {
  file_path: "utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt"
  file_sha1_hash: "37eb0c6a3a16c300da100e245e7d320c61cc008f"
  covered_line {
    line_number: 14
    coverage: FULL
  }
  covered_line {
    line_number: 15
    coverage: FULL
  }
  covered_line {
    line_number: 16
    coverage: FULL
  }
  covered_line {
    line_number: 17
    coverage: FULL
  }
  covered_line {
    line_number: 21
    coverage: FULL
  }
  covered_line {
    line_number: 33
    coverage: FULL
  }
  covered_line {
    line_number: 34
    coverage: FULL
  }
  covered_line {
    line_number: 35
    coverage: FULL
  }
  covered_line {
    line_number: 36
    coverage: FULL
  }
  covered_line {
    line_number: 41
    coverage: FULL
  }
  covered_line {
    line_number: 42
    coverage: FULL
  }
  covered_line {
    line_number: 43
    coverage: FULL
  }
  covered_line {
    line_number: 44
    coverage: FULL
  }
  covered_line {
    line_number: 45
    coverage: FULL
  }
  covered_line {
    line_number: 46
    coverage: FULL
  }
  covered_line {
    line_number: 47
    coverage: FULL
  }
  covered_line {
    line_number: 49
    coverage: FULL
  }
  covered_line {
    line_number: 58
    coverage: FULL
  }
  covered_line {
    line_number: 59
    coverage: FULL
  }
  lines_found: 19
  lines_hit: 19
  function_coverage {
    line_number: 58
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature$Companion::createSignature$utility_src_main_java_org_oppia_android_util_parser_math_math_latex_model_kt (Ljava/lang/String;FZ)Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 33
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature::<init> (Ljava/lang/String;IZ)V"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 35
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature::getLineHeightHundredX ()I"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 34
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature::getRawLatex ()Ljava/lang/String;"
    execution_count: 0
    coverage: NONE
  }
  function_coverage {
    line_number: 36
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature::getUseInlineRendering ()Z"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 41
    function_name: "org/oppia/android/util/parser/math/MathModel$MathModelSignature::updateDiskCacheKey (Ljava/security/MessageDigest;)V"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 14
    function_name: "org/oppia/android/util/parser/math/MathModel::<init> (Ljava/lang/String;FZ)V"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 16
    function_name: "org/oppia/android/util/parser/math/MathModel::getLineHeight ()F"
    execution_count: 0
    coverage: NONE
  }
  function_coverage {
    line_number: 15
    function_name: "org/oppia/android/util/parser/math/MathModel::getRawLatex ()Ljava/lang/String;"
    execution_count: 0
    coverage: NONE
  }
  function_coverage {
    line_number: 17
    function_name: "org/oppia/android/util/parser/math/MathModel::getUseInlineRendering ()Z"
    execution_count: 0
    coverage: NONE
  }
  function_coverage {
    line_number: 21
    function_name: "org/oppia/android/util/parser/math/MathModel::toKeySignature ()Lorg/oppia/android/util/parser/math/MathModel$MathModelSignature;"
    execution_count: 1
    coverage: FULL
  }
  functions_found: 11
  functions_hit: 7
  branch_coverage {
    line_number: 46
    block_number: 0
    branch_number: 0
    hit_count: 1
    coverage: FULL
  }
  branch_coverage {
    line_number: 46
    block_number: 0
    branch_number: 1
    hit_count: 1
    coverage: FULL
  }
  branches_found: 2
  branches_hit: 2
}
, bazel_test_target: "//utility/src/test/java/org/oppia/android/util/math:FloatExtensionsTest"
covered_file {
  file_path: "utility/src/main/java/org/oppia/android/util/math/FloatExtensions.kt"
  file_sha1_hash: "178a2d84e9182a11b41adf34abc0a92a7d365f50"
  covered_line {
    line_number: 28
    coverage: FULL
  }
  covered_line {
    line_number: 36
    coverage: FULL
  }
  covered_line {
    line_number: 43
    coverage: FULL
  }
  lines_found: 3
  lines_hit: 3
  function_coverage {
    line_number: 36
    function_name: "org/oppia/android/util/math/FloatExtensionsKt::isApproximatelyEqualTo (DD)Z"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 28
    function_name: "org/oppia/android/util/math/FloatExtensionsKt::isApproximatelyEqualTo (FF)Z"
    execution_count: 1
    coverage: FULL
  }
  function_coverage {
    line_number: 43
    function_name: "org/oppia/android/util/math/FloatExtensionsKt::toPlainString (D)Ljava/lang/String;"
    execution_count: 1
    coverage: FULL
  }
  functions_found: 3
  functions_hit: 3
  branch_coverage {
    line_number: 28
    block_number: 0
    branch_number: 0
    hit_count: 1
    coverage: FULL
  }
  branch_coverage {
    line_number: 28
    block_number: 0
    branch_number: 1
    hit_count: 1
    coverage: FULL
  }
  branch_coverage {
    line_number: 36
    block_number: 0
    branch_number: 0
    hit_count: 1
    coverage: FULL
  }
  branch_coverage {
    line_number: 36
    block_number: 0
    branch_number: 1
    hit_count: 1
    coverage: FULL
  }
  branches_found: 4
  branches_hit: 4
}
]
</pre>
</details>

**Representation in Protobuf**
- Function Coverage: Interpreted execution counts to determine coverage status. Any execution count greater than 0 signifies full coverage, otherwise considered null.

- Branch Coverage: Determined coverage based on the recorded hit count for each branch. A hit count greater than 0 denotes full coverage, otherwise marked as null.

**Usage:**
The stored proto will be used to generate coverage reports in **md** and **HTML** formats [to be handled in PR 1.5]

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
@BenHenning, This PR actually covers **PR 1.4** Stage **A**, **B** and **partially C** (as modules possible of having multiple tests is yet to be added. Will add that for PR 1.3 and pass it to over here). But the logic is done and the existing related test cases are updated as well to test the generated protos. Would work on the logical changes as per review and address the remaining test cases then too. PTAL

BenHenning:
Got it, thanks @Rd4dev. Marking this as draft until it's full completed (though I'll still take a review pass on it).

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,63 @@
+syntax = "proto3";
+
+package proto;
+
+option java_package = "org.oppia.android.scripts.proto";
+option java_multiple_files = true;
+
+message CoverageReport {
```
BenHenning, line 10: Reminder to add documentation comments for all protos and all fields in this file.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,63 @@
...
+  optional int32 functions_found = 7;
+  optional int32 functions_hit = 8;
+  repeated BranchCoverage branch_coverage = 9;
+  optional int32 branches_found = 10;
+  optional int32 branches_hit = 11;
+}
+
+message CoveredLine {
+  int32 line_number = 1;
+  CoveredLine.Coverage coverage = 2;
```
BenHenning: Can just be ``Coverage``. Ditto below.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,63 @@
...
+
+  enum Coverage {
```
BenHenning: Suggest moving this up & sharing it across the protos as needed.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,63 @@
...
+    UNSPECIFIED = 0;
+    FULL = 1;
+    NONE = 2;
+  }
+}
+
+message BranchCoverage {
+  int32 line_number = 1;
+  optional int32 block_number = 2;
```
BenHenning: Drop all optionals in this file--optional is the default (and only) property for proto3.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,63 @@
+syntax = "proto3";
+
+package proto;
+
+option java_package = "org.oppia.android.scripts.proto";
+option java_multiple_files = true;
+
+message CoverageReport {
+  string bazel_test_target = 1;
+  repeated CoveredFile covered_file = 2;
```
BenHenning: I'm curious: how valuable is it to store all of the covered files if we (generally) only care about a single source file per test target?

Separately, I wonder if we should model this report in the reverse order: have a single report for a source file and a list of test targets that were run for that _combined_ report. That better aligns with how we'll be using the protos, after all.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
```diff
@@ -65,7 +66,7 @@ class RunCoverageForTestTarget(
   /**
    * Analyzes target file for coverage, generates chosen reports accordingly.
    */
-  fun runCoverage(): List<String>? {
+  fun runCoverage(): CoverageReport? {
```
BenHenning: Here & elsewhere: we probably should throw an exception rather than return null if we can produce a report (especially since that should only happen in exceptional cases).

BenHenning (changes requested):
Thanks @Rd4dev! Took a first pass only on the data structures as that's generally the part which will affect the broader PR the most. Once that's done, I'll focus on the logic and algorithmic sides since there aren't many design decisions to consider in this PR. I'll review tests last.

Some of the comments in this & future rounds may be easier to address in upstream PRs, instead. Feel free to do that where you think it makes sense.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: So @BenHenning, Can I just understand if this would affect storing and accessing values in our proto? Initially, I used int32 for execution_count, hit_count, block_number, branch_number, etc., but since 0 is considered a default value, these fields are not actually represented in the proto. This resulted in:

```
functions_found: 11
functions_hit: 7
branch_coverage {
  line_number: 46
  block_number: 1
  hit_count: 1
}
branch_coverage {
  line_number: 46
  branch_number: 1 // hit_count not shown as 0 (default)
}
```
Would this cause issues when setting Coverage as None or when trying to parse the data into HTML? Sorry if I missed any alternative ways to retrieve the data while still ignoring optional fields. I'll look into the available docs further in meantime, but would having no value representation be considered as 0 and parsed accordingly?

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Hey @BenHenning, I'm not entirely sure if I understand your question correctly. Are you asking why I'm storing all the covered files? 

Let's consider the scenario in **util/math** where we have several files like **RealExtensions.kt**, **FloatExtensions.kt**, **ExpressionToPolynomialConverter.kt**, **ExpressionToComparableOperationConverter.kt**, **ComparableOperationExtensions.kt**, and **others**. 

Typically, (_Bazel way of getting coverage_) when we perform coverage analysis, bazel generates a combined coverage report that includes data for every file in the package—that's the default behavior. 

However, in our case, we specifically want to gather coverage data for just a single file. 
**For example**, if we're interested in the coverage of **RealExtensions.kt**, the report should **only** **contain** RealExtensions.kt's coverage data, **disregarding others in the package**. 

This required approach was kept in mind and was implemented in PR 1.4 [[implementation](https://github.com/oppia/oppia-android/pull/5439/files#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR81-R89)], where it checks for the filename and parses the coverage data for the given filename while skipping the rest. This aligns with our original design intent.

Can I know if there are any conflicts with this approach or any concerns or if I've missed anything that needs to be addressed? Also, I'm not quite clear on what you mean by asking how valuable it is to store all covered files. 

In the later part where you mentioned about 'modelling this report in the reverse order' Could you please provide more details? Regarding your suggestion for a combined report, do you think it's necessary to have a combined report for all files in the packages? Should we proceed with that approach? Just wanted to know why?... as that could be helpful while generating custom reports!

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Added Docs @BenHenning. But I have trouble understanding the difference between the 
  
Block number and branch number in the coverage data.
  - int32 **block_number** = 2;
  - int32 **branch_number** = 3;
  
These are generally represented as: 
```
BRDA:46,0,0,1
BRDA:46,0,1,1
```
what does a block and a branch mean?
So I assumed branch would be like 
```
if(condition) { // branch 0
// code block
} else { // branch 1
// code block
}
```
if that's right then what is a block in it?
```
if(condition) { // is the condition called a block - condition 1 - block 0, condition 2 - block 1
// code block // or the actual code block a block?
} else { // branch 1
// code block
}
```
I am also looking into few results to figure this out, but if it would be possible to share some insights on this that would be helpful.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Oops, yeah! Updated the proto structure.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Done

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Removed the optional specifications from all int32 as they just work as intended.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -41,4 +53,176 @@ class CoverageRunner(
...
   }
+
+  private fun parseCoverageData(
+    coverageData: List<String>,
+    bazelTestTarget: String
+  ): CoverageReport {
+    var filePath = ""
+    var linesFound = 0
+    var linesHit = 0
+    val coveredLines = mutableListOf<CoveredLine>()
```
Rd4dev: TODO: refactor to avoid mutability

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Refactored it now as to throw RuntimeException
```
fun runWithCoverageAsync(
    bazelTestTarget: String
  ): Deferred<CoverageReport> {
    return CoroutineScope(scriptBgDispatcher).async {
      val coverageResult = retrieveCoverageResult(bazelTestTarget)
        ?: throw RuntimeException("Failed to retrieve coverage result for $bazelTestTarget")
}
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -95,7 +97,7 @@ class RunCoverage(
     }
   }
 
-  private fun runCoverageForTarget(testTarget: String): List<String>? {
+  private fun runCoverageForTarget(testTarget: String): CoverageReport? {
```
Rd4dev: TODO: Refactor to throw an exception (possibly)

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Would this also apply to the CoverageRunners's retrieveCoverageResult that retrieves the result from bazelClient?
```
  private fun retrieveCoverageResult(
    bazelTestTarget: String
  ): List<String>? {
    return bazelClient.runCoverageForTestTarget(bazelTestTarget)
  }
```

Rd4dev:
@BenHenning, I addressed the existing review comments, but still need to do some refactorings, but before proceeding can I just confirm if everything is as plan, as per concerns with https://github.com/oppia/oppia-android/pull/5439#discussion_r1649554653, and have another review if possible, PTAL!

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -41,4 +53,176 @@ class CoverageRunner(
...
+                if (hitCount > 0)
+                  Coverage.FULL
+                else
+                  Coverage.NONE
+              coveredLines.add(
+                CoveredLine.newBuilder()
+                  .setLineNumber(lineNumber)
+                  .setCoverage(coverage)
+                  .build()
+              )
```
```suggestion
              val (lineNumStr, hitCountStr) = line.substringAfter("DA:").split(",")
              coveredLines += CoveredLine.newBuilder().apply {
                this.lineNumber= lineNumStr.toInt()
                this.coverage = if (hitCountStr.toInt() > 0) Coverage.FULL else Coverage.NONE
              }.build()
```

Destructuring and ``apply`` let you streamline this type of code a lot more, as an option.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
BenHenning: Mutability here isn't strictly bad, but it could be avoided by taking multiple passes on the data file lines (and specifically extracting out lines with the prefixes that are needed for each thing). It's definitely not as performant, and that could matter for large files.

For values where we need to guarantee the value is set, I highly suggest using lateinit var and not specifying a default, or specifying a null default and checking for nullability at the end (to ensure all required values are set).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -41,4 +53,176 @@ class CoverageRunner(
...
+    val coveredLines = mutableListOf<CoveredLine>()
+    val branchCoverage = mutableListOf<BranchCoverage>()
+    val functionCoverage = mutableListOf<FunctionCoverage>()
+
+    var functionsFound = 0
+    var functionsHit = 0
+    var branchesFound = 0
+    var branchesHit = 0
+
+    var parseFile = false
```
BenHenning: A cleaner option might be to do something like this:

```kotlin
val sfStartIdx = coverageData.indexOfFirst { it.startsWith("SF:") }
val eofIdx = coverageData.indexOfFirst { it.startsWith("end_of_record") }
val fileSpecificCovDatLines = coverageData.subList(sfStartIdx, eofIdx)
```

Now you can avoid the slightly awkward "am I parsing right now?" state of ``parseFile`` and eliminate a whole ``when`` block nesting below since we know exactly what lines to parse from ``fileSpecificCovDatLines``.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -41,4 +53,176 @@ class CoverageRunner(
   ): List<String>? {
     return bazelClient.runCoverageForTestTarget(bazelTestTarget)
   }
+
+  private fun parseCoverageData(
+    coverageData: List<String>,
```
BenHenning, line 55: Perhaps instead: ``coverageDataFileLines`` might be a bit clearer/more specific.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -30,9 +39,12 @@ class CoverageRunner(
    */
   fun runWithCoverageAsync(
     bazelTestTarget: String
-  ): Deferred<List<String>?> {
+  ): Deferred<CoverageReport> {
     return CoroutineScope(scriptBgDispatcher).async {
-      retrieveCoverageResult(bazelTestTarget)
+      val coverageResult = retrieveCoverageResult(bazelTestTarget)
+        ?: throw RuntimeException("Failed to retrieve coverage result for $bazelTestTarget")
```
BenHenning: Use ``error`` rather than a generic ``RuntimeException``--it throws an ``IllegalStateException`` which is contractually more appropriate for the failure being caught here.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
BenHenning: Another option could be to generically parse the structure since it seems well-defined as properties, and then construct the sub-elements of the proto by pulling from the generic parsed structure. In this case, I think you could parse all the lines into a ``Map<String, List<List<String>>>`` data structure using something like:

```kotlin
val coverageDataProps = coverageData.groupBy { line ->
  val (key, value) = line.split(":", limit = 2)
  return@groupBy key
}.mapValues { (_, line) ->
  line.substringAfter(":").split(",")
}
```

Then it can become pretty straightforward:

```kotlin
val functionsFound = coverageDataProps.getValue("FNF").single().single().toInt()
val coveredLines = coverageDataProps.getValue("DA").map { (lineNumStr, hitCountStr) ->
  CoveredLine.newBuilder().apply {
    this.lineNumber = lineNumStr.toInt()
    this.coverage = if (hitCountStr.toInt() > 0) Coverage.FULL else Coverage.NONE
  }.build()
}
```

As a couple of examples. This shows cases where:
- Lines must be present in the data file (FNF, DA).
- A case where exactly one entry should be present (FNF).
- A case where an entry has exactly one property (FNF).
- A case where a line can occur multiple times in the file (DA).
- Using destructuring to extract properties from a multi-property line (DA).

This could be another approach to eliminate the mutability _and_ nesting, introduce stronger assertion checks (e.g. with ``getValue()`` and ``single()``), and do this without a performance hit (since the map structure should be well optimized for this type of repeated retrieval and iteration pattern unlike the raw lines).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -74,7 +75,8 @@ class RunCoverage(
    * @return a list of lists containing coverage data for each requested test target, if
    *     the file is exempted from having a test file, an empty list is returned
    */
-  fun execute(): List<List<String>> {
+
```
BenHenning: Nit: remove empty line (KDocs should be immediately adjacent the method being documented).

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -0,0 +1,84 @@
...
+// Bazel coverage execution.
+message CoverageReport {
+  // The test target for which the coverage report is generated.
+  string bazel_test_target = 1;
+  // A list of files covered in this coverage report.
+  repeated CoveredFile covered_file = 2;
+}
+
+// Information about the single file that was covered during the tests.
+message CoveredFile {
```
BenHenning: After looking at the parsing logic, how many of these properties do we actually need? I highly suggest dropping everything that doesn't actually have use for enforcement or final HTML/MD report generation.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
BenHenning: I'm not 100% certain that I understand the distinction, either. I'm having some difficulty in finding a well-define page documenting the file format to explain the distinction, as well.

I suspect we can probably ignore all but lines for coverage since that's what we're presumably using for computing line-of-code percentage. Branch & block analysis could be interesting to add in the future to try and enforce more thorough tests (or for the final HTML/MD report generation).

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
BenHenning: Let me try explaining my thoughts a different way, and it may be a bit repetitive to what you already said (apologies if that's the case).

The current modality of coverage in Bazel is to run coverage on one or more test targets. For parity with how we're using the script, let's assume one target. As you mentioned, Bazel will instrument all of the files under the instrumentation filter and should generate a single coverage.dat file that covers all of those files. To that end, it would make sense to design the proto to contain the tested target and all of the covered source files as part of the instrumentation filter.

However, how we want our coverage tool to work is to provide coverage analysis for only a single source file: the one that corresponds to the exact test target being checked. This begs two questions per the current design in this PR:
- There's really only one covered file we care about for each test target, so why have a list here?
- There's actually a many:1 mapping from target to source file (since a single source file may have a regular test, a local test, and a shared test). How do we consolidate these into a proper, single report for the original source file?

Separately, we also need to be able to collate the coverage results for each target/source file pair into a high-level report for eventual HTML/Markdown generation (since that step should be summarizing the code coverage for _all_ targets, not just a single one). How are we representing this part of the plan in the data structures being introduced here?

Does this help clarify?

BenHenning (dismissed):
Thanks @Rd4dev--left some more thoughts PTAL.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Got it, Ben. As discussed, now the current implementations will only represent the line coverage for both HTML and MD reports.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: I think with this comment https://github.com/oppia/oppia-android/pull/5443#issuecomment-2189239096 and discussions things are clear, and we will need to possibly have the ability to run multiple targets/paths at once to replicate same wasy as to be done with CI for M2.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: :astonished: thanks @BenHenning, for just giving the entire parsing logic. Updated the source with the same!

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Just removed everything else and kept only those related to lines:
```
message CoverageReport {
  string bazel_test_target = 1;
  string file_path = 2;
  string file_sha1_hash = 3;
  repeated CoveredLine covered_line = 4;
  int32 lines_found = 5;
  int32 lines_hit = 6;
}

message CoveredLine {
  int32 line_number = 1;
  Coverage coverage = 2;
}
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Now it uses - `error("Failed to retrieve coverage result for $bazelTestTarget")`

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Thanks Ben! Updated the parsing logic.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Thanks Ben! Updated with the exact logic.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Okay Done!

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Done used `error` accordingly!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Got it! Done.

Rd4dev:
@BenHenning, also would it be possible to have a Stage A and B review on this, the tests need to be fixed yet. But can I know if the logic and proto structures are right? Can you PTAL!

BenHenning:
At a very quick glance it seems really solid @Rd4dev. @adhiamboperes will probably need to take the more thorough next passes.

Just to check: are you planning to change the script to take a source file and derive the test targets from that?

Rd4dev:
> Just to check: are you planning to change the script to take a source file and derive the test targets from that?

Yeah @BenHenning, that is what is implemented in the https://github.com/oppia/oppia-android/pull/5433, developer inputs a sourceFile.kt and we map them to test targets. (this was decided based on user experience)

Edit: If that question is something related to M2 (I am right now uncertain which would be the best strategy and that would be based on the ci runs. If we could get the list of **changed files** (not affected), then that that should be straightforward with the existing script.

Rd4dev:
The suggestion to add test cases as in https://github.com/oppia/oppia-android/pull/5433#discussion_r1650352941 are implemented in this PR with [[commit](https://github.com/oppia/oppia-android/pull/5439/commits/cf405c617f04fd2b1ce539fcdeb04d6e166d4e33)]

Rd4dev:
Hi @adhiamboperes, this PR is now ready for review, addressed all the review comments Ben left. Can you PTAL!

adhiamboperes (approved):
Thanks @Rd4dev. I took a full pass and the implementation looks clean, and I think the structure of the `coverage.proto` looks great.

Rd4dev:
Thank you so much @adhiamboperes :smiley:

Rd4dev:
@adhiamboperes, also I am unsure how to get this pass the unresolved conversations, to get it merged?

DubeySandeep:
@seanlip Can you help get this PR merged? @BenHenning is on leave and was [fine with the PR changes](https://github.com/oppia/oppia-android/pull/5439#issuecomment-2191082065).

seanlip:
Yup, I think that's fine. I'll see if I can merge this.

seanlip:
@adhiamboperes Could you please check the previous conversations and resolve them to confirm they are resolved, since that's a requirement for merge? Thanks.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes: @Rd4dev  please see comment you created.

Rd4dev: Oops! implemented that, will resolve this comment now.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
adhiamboperes: This file is nolonger in this PR. Did you move it to a different one?

adhiamboperes:
@Rd4dev, there are two comments to respond to, but I have verified everything else as addressed and resolved them.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverageForTestTarget.kt
Rd4dev: Actually, the **file was removed** to keep the scripts clean and non-redundant.

Original implementation involved introducing 3 scripts, but then felt the 3rd one redundant, so it was then decided to be removed as per https://github.com/oppia/oppia-android/pull/5433#discussion_r1648659133.

The above mentioned `error` change was updated in other relevant references in the script: https://github.com/oppia/oppia-android/pull/5439/files#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR42

Rd4dev:
> @Rd4dev, there are two comments to respond to, but I have verified everything else as addressed and resolved them.

Thanks @adhiamboperes, 

- Replied to the conversation on non-existing file: https://github.com/oppia/oppia-android/pull/5439/files/25a064e555894a21b18a0c5aa759e9f8e28b3a19#r1656625309
- And resolved the conversation I opened: https://github.com/oppia/oppia-android/pull/5439/files/bc9a5c084cbcebb97ed449c624489215a14373e9#r1656590468 (as it was implemented)

adhiamboperes:
Thanks @Rd4dev. Everything looks good.

PR #5443, opened 2024-06-24
Fix part of #5343: Generate Code Coverage Report in HTML and MARKDOWN formats

Rd4dev:
## Explanation

Fixes part of #5343 

### Project 
[PR 1.5 of Project 4.1]

### Changes Made
This PR introduces the `CoverageReporter` utility to generate the code coverage report. The main features and changes include:

- **New Utility:**
  - `CoverageReporter` to generate code coverage reports.

- **Command Line Arguments:**
  - Now support output formats: `HTML` or `MARKDOWN`.

- **Report Generation:**
  - Generates coverage reports from the generated proto for the requested format.
  - Outputs are generated and stored in the `coverage_reports` folder relative to the `repoRoot`.

- **Gitignore Update:**
  - The `coverage_reports` folder is added to `.gitignore` to ensure it is not added to the repository.

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
Hey @BenHenning, just wanted to clear 2 main things up at the start. 
### 1. MARKDOWN report format:
The .md report as I assume it to be most useful for CI runs, containing essential result data about the percentage of covered code, and can be added as a comment. This base template was discussed in our meeting notes.
```
Total coverage:
Files covered: (# changed / # run with coverage)
Coverage percentage: ##% covered / ##% expected
LOC: # covered / # instrumented

(indent left) Specific coverage:
app/src/.../app
home
HomeActivity.kt - 87% (110/115)
...
...
```
The current Markdown report follows this foundational format and now appears as:

# Coverage Report

**Total coverage:**
- **Files covered:** 1
- **Covered File:** utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
- **Coverage percentage:** 87.50% covered
- **Line coverage:** 19 covered / 19 found
- **Function coverage:** 7 covered / 11 found
- **Branch coverage:** 2 covered / 2 found

Probably simplifying it to 
Line coverage: 19 / 19 covered, 
Function coverage: 7 / 11 covered, 
Branch 2 / 2 covered

I would like to know if any additional details that should be included in the report for local generation.

___

### 2. Path to Store Reports

**Variable Path Definitions Across Different OS**

Windows users typically use absolute paths like `C:/Users/...`, whereas in WSL, paths are like `/mnt/c/Users`. 
So if we direct developers to use an absolute path, would this cause any confusion, and probably would it need to be handled in our file creation processes.

**Providing and Handling Different Kinds of Paths**

- **Path Variants:**
  - When developers are directed to provide relative paths, some paths might be specified as `coverage/coverage.html`, where the file extension could be incorrect (e.g., it might be a Markdown generation) - with command 
```
bazel run //scripts:run_coverage -- $(pwd) utility/...parser/math/MathModel.kt MARKDOWN
```
  - Others might simply specify a directory like `coverage/`, requiring us to check if a default filename should be provided.

**Subpaths and Endless/Long Chains**

- **Managing Subdirectories:**
  - For structured coverage reports, like for `utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt`, I propose creating subdirectories under the same path to maintain organized coverage results.
  - Example structure: `(absolute/relative to $repoRoot)/utility/src/main/java/org/oppia/android/util/parser/math/MathModel/coverage.html`.
  - I took this approach just to ensure the coverage reports are structured and possible re-runs of **other files** don't overwrite the same file, which would otherwise complicate referencing other file coverages (when needed). 

- **Handling Long Input Paths:**
  - But with this above approach if we consider scenarios where users provide long paths themselves (`bazel run ... /coverageData/report/path/place/here/`) our mapping would result in $repoRoot/(user_defined_path)coverageData/report/path/place/here/(our_default_path)/utility/src/main/java/org/oppia/android/util/parser/math/MathModel/coverage.html. 

I just don't know, if this is a simple process and its just something that I complicate thinking of weird possibilities and approaches

#### Just stuck to something simple on both ends for now

I've implemented a solution based on these considerations, but it deviates from our previous discussions. I'm seeking your feedback to refine and adjust as necessary. 

So basically, I just made it simple to not get developers input itself and make use of the filePath argument and create a dir coverage_reports in the repoRoot, based on the HTML or MARKDOWN arg

so for a command -
``` 
bazel run //scripts:run_coverage -- $(pwd) utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt HTML
// it would be
reportOutputPath: oppia-android/coverage_reports/utility/src/main/java/org/oppia/android/util/parser/math/MathModel/coverage.html
```
and for 
```
bazel run //scripts:run_coverage -- $(pwd) app/src/main/java/org/oppia/android/app/mydownloads/MyDownloadsActivity.kt MARKDOWN
// it would be
reportOutputPath: oppia-android/coverage_reports/app/src/main/java/org/oppia/android/app/mydownloads/MyDownloadsActivity/coverage.md
```
_(Even this feels like a train)_

- **Version Control Consideration:**
  - To prevent unintentional inclusion in the GitHub repository, the coverage_reports dir is added to `.gitignore`.

Would it be possible to clear this up, if we can just proceed this way or should taking inputs from developers/users be mandatory?

Rd4dev:
@BenHenning,

### Update on progress: 

Implemented a base html layout for code coverage (TODO: still doesn't have computations done to have a cumulative coverage representation)

### The Approach:
The lcov genhtml generation produces coverage report with 2 views
1. For Line Coverage
2. For Functions

**Line Coverage**
Screenshot (1396)

**Function Coverage**
Screenshot (1397)


And as far as seen no branch coverage (yet to be tested with multiple targets)

So I wanted every detail to be in one single place so wanted to combine them like this base structure (derived from other coverage reports)
Screenshot (1402)

This has 4 columns:
1. For Line number
2. For Branch coverage - provides hit count
3. For Function execution coverage - provides execution count
4. Actual source code

Am also thinking if we can include the actual function hit with the function names like the lcov genhtml function view did, either separately or right below the available html report.

Also, the coverage (representative color) is decided based on the cumulative coverage result of the line with line, branch and functions.
So like
|Line|Branch|Function|Coverage color|
|---|---|---|---|
|FULL|FULL|FULL|GREEN|
|FULL|FULL|NONE|YELLOW|
|FULL|NONE|NONE|YELLOW|
|NONE|FULL|NONE|YELLOW|
|NONE|NONE|NONE|RED|

This still needs a lot of formatting to be done.
And for now the coverage representations are **hardcoded.**

But would like to get an approval on the html layout before proceeding.

.gitignore
```diff
@@ -25,3 +25,4 @@ bazel-*
 .bazelproject
 .aswb
 *.pb
+coverage_reports
```
BenHenning, line 28: Is this generated by Bazel? Otherwise, where does it come from & why should it be ignored?

scripts/src/java/org/oppia/android/scripts/coverage/BUILD.bazel
```diff
@@ -14,6 +14,7 @@ kt_jvm_library(
     deps = [
         "//scripts/src/java/org/oppia/android/scripts/common:bazel_client",
         "//scripts/src/java/org/oppia/android/scripts/coverage:coverage_runner",
+        "//scripts/src/java/org/oppia/android/scripts/coverage:coverage_reporter",
```
BenHenning:
```suggestion
        ":coverage_reporter",
```

Doesn't need to be fully qualified since it's in the same package (ditto elsewhere in your other PRs).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -64,6 +75,12 @@ class RunCoverage(
   private val rootDirectory = File(repoRoot).absoluteFile
   private val testFileExemptionTextProto = "scripts/assets/test_file_exemptions"
 
+  companion object {
+    // The minimum coverage percentage threshold for coverage analysis,
+    // The script will fail if the file has less than the minimum specified coverage.
+    const val MIN_COVERAGE_PERCENTAGE = 20
```
BenHenning: Where did this number (20) come from? We haven't discussed yet what the default value should be for new code, so I'm curious how you came to this specific value.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -31,6 +33,13 @@ import java.util.concurrent.TimeUnit
 fun main(vararg args: String) {
   val repoRoot = args[0]
   val filePath = args[1]
+  val reportFormat = when (args.getOrNull(2)) {
```
BenHenning: I suggest instead taking as an argument the path to the report file to generate and then verify that it ends in either ".md" or ".html" (and error otherwise). The extension then acts as your mode, and you don't need to default any assumptions on file location or name.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -0,0 +1,310 @@
+package org.oppia.android.scripts.coverage
+
+import org.oppia.android.scripts.proto.CoverageReport
+import org.oppia.android.scripts.proto.Coverage
+import java.io.File
+
+class CoverageReporter(
```
BenHenning, line 14: Some thoughts (without reviewing the code in significant detail on this pass):
- As I was mentioning in the 1.4 PR, the expectation is that MD/HTML report generation is for the total summary of results from an entire run of coverage, not for a single target.
- This might differ from a comment I left below about the extension ending in HTML, but I was actually expecting that the HTML file generated would be a rich annotation of the source file being covered (e.g. with red/green/non highlighted lines). This conflicts a bit with a comment I left below about detecting the report type based on extension, but it's adaptable: HTML would require the user to pass a directory path wherein we would probably replicate the existing source file structure and generate an HTML file for each source file covered by tested targets.

I think that you need to think about code coverage, as a whole, like a pipeline:
- We can run coverage for a single target and get a single report for that target.
- We can run coverage for a bunch of targets by running the coverage script for each one in isolation (leading to a bunch of code coverage reports).
- We can combine all of these reports into a single, combined/aggregated report.
- We can generate either a single summarization of all the results (MD), or a rich view of exact coverage on a per-source file basis (HTML) using this combined/aggregated report.

I actually thought the original plan in the design document was to introduce multiple scripts, but perhaps I'm misremembering whether part of that is planned for M2. I can see how the existing technical requirements written in the issue don't go into significant detail on the multi vs single target situation (I intended for M1 to be multiple targets only hence why M2 had a deliverable for single target support).

I do recall we discussed this in some detail in the design document, and I had thought the document moved ultimately toward this approach. It'd be really helpful to understand what you're thinking here broadly.

BenHenning (dismissed):
Thanks @Rd4dev. I get the sense that we actually have differing expectations for what we ultimately need these utilities to be able to do--PTAL closely at my top CoverageReporter comment, plus my comments in the 1.4 PR. We definitely need to get on the same page ASAP because I think, as they currently stand, you're going to have difficulty getting these scripts working the way we need them to come integration time in M2 since they're not being designed with multiple targets in mind (unless I'm missing something significant).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
BenHenning: Follow-up: ah I think part of my comment above is incorrect since I just realized you are, indeed, coloring the lines of the source file in the generated HTML (it looks really nice!), but the single vs. multi targets part is still the key difference, I think, that needs to be figured out.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Sorry for introducing that. I was testing a default minimum threshold for the package files to exit on getting lower coverage results and chose a slightly relevant value based on the issue tracker specifications, which stated: 
>Note that the minimum enforced coverage for now should be something very modest, like 10%. This number will be increased in future changes outside the scope of this project (as the number may not be believable until we understand the limitations of what can be covered by JaCoCo).

However, that was **just an experiment** and will be adjusted according to the final decision. Or may be I can just remove that for now?

.gitignore
Rd4dev: It seems my previous comment, placed way above, was missed. Could you please take a look at it when you get a chance? It provides context for the changes I made. https://github.com/oppia/oppia-android/pull/5443#issuecomment-2186663348

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks for the suggestion @BenHenning, can you also just confirm if getting the path from user would be more useful / simpler or doing it ourself and keeping it simple with just getting HTML or MARKDOWN as arg would be a option as per https://github.com/oppia/oppia-android/pull/5443#issuecomment-2186663348

The following is the same as my concerns in the comment:
So getting the user type the path and checking if it ends with ".md" or ".html" should be straightforward but 
- would users/developers input an absolute or relative path?
- would a windows user understand the wsl way of providing the path
- if relative path, would that mean having it in the repoRoot?
- if added to repoRoot would that make sure they are not added accidentally to the original repo.

Rd4dev:
Edit: Sorry this is long, to just address your comments here is a shorter comment: https://github.com/oppia/oppia-android/pull/5443#issuecomment-2189264044

@BenHenning, to some extent I think we might be on the same page, but I just wanted to confirm with actually showing you what and how I intend to go with the overall flow of the project

### M1 phase:
**Scripts:**
1. RunCoverage.kt
2. CoverageRunner.kt
3. CoverageReporter.kt

**Start:**

1. Developer runs the command to get coverage analysis for a certain file
Developer inputs:

```
bazel run //scripts:run_coverage -- $(pwd) utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt HTML
```

Now the run_coverage scripts ie. RunCoverage.kt is triggered

2. RunCoverage.kt 
Receives the file and checks if the file is exempted 
```
if yes
then it skips coverage
if no
then 
- it tries to map the file path to its respective test targets and returns a list of test targets
```

so for **1:1 mapping**
```
utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt -> 
[//utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest]
```

for **many:1 mapping**
```
app/src/main/java/org/oppia/android/app/home/HomeActivity.kt -> 
[//app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest,
//app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest]
```

3. Run Coverage analysis asynchronously
the list of test targets are passed to the CoverageRunner to runCoverageAnalysis asynchronously on the provided test targets and returns a CoverageReport for each test target separately
The **CoverageReport** is the coverage data stored in proto **for a single test target**

we then collect it and make it as a list with the returned CoverageReport
so now we would have a **list of CoverageReports**

the coverage reports would be like
for **1:1 mapping** (just for representation not acutal data)
```
[{
  bazel_test_target://utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest,
  CoveredLine{line_number:2}
}]
```

for **many:1 mapping** (just for representation not acutal data)
```
[{
  bazel_test_target://app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest,
  CoveredLine{line_number:2}
 },
 {
  bazel_test_target://app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest,
  CoveredLine{line_number:3}
}]
```

I think until this there needs no huge change in the way it is implemented (except for proto data structure)
___

Now we have the list of coverage reports and the thing left to do is actually make it something useful by creating a **.md** or **.html** file
So to do this we pass this list of coverage reports to the CoverageReporter.kt

4. CoverageReporter.kt
_(Heads up: So for now the PR1.5 just provides implementation to generate reports for a single target 1:1 mapping targets
But work needs to be done to figure out how we can get the combined reports for many:1 mapping.)_

### a. MD reports
I consider **.md reports** are mostly helpful for **ci runs**
and **.html reports** for **local development** purposes

so in here the process flows like
with the list of coverage reports
we either send these list of coverage reports to generating html or generating md reports based on the input

for **1:1 mapping**
The md text is generated just with the one - coverage report and[0] and saved as md file to the output path.

for **many:1 mapping** (Haven't been implemented but the plan is)
say for the coverage reports
```
[{
  bazel_test_target://app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest,
  CoveredLine{line_number:2}
 },
 {
  bazel_test_target://app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest,
  CoveredLine{line_number:3}
}]
```

Both are essentially going to have the same number of founds (as both are for same file)
```
coverage reports[0] had line 1: FULL, line 2: FULL, line 3: NONE, line 4: NONE
coverage reports[1] had line 1: NONE, line 2: FULL, line 3: FULL, line 4: NONE
(to be computed) so both should be computed to result in 
line 1: covered
line 2: covered
line 3: covered
line 4: not covered
```

and the combined result will be made as a **md file** and saved to the output path.
so thereby we achieve the **consolidated output**

### [M2 phase future]
_This is something I still haven't tried but I am planning to do this to achieve the report generation of all the affected files in the ci run_
It would probably be derived from the unit test, compute affected test runs

So the plan is to return the actual md data
like for file 1:
```
**Lines:** 1/1 covered
```

So when a CI run is started,
1. We should 1st get the **list of files that are changed**
2. We will pass the file names one by one to perform the regular code coverage analysis as done earlier in M1
3. So it will return the **actual md string** as an output for that file

say file1: 
```
**file_path:** /file1.kt
```

and this md will be saved in the ci run in **a variable** or in a **log file**

then again, the process continues for file 2
1. We pass the file 2 to do coverage analysis
2. Get its md string output

say file2:
```
**file_path:** /file2.kt
```

and this md will be added to the previous saved md string or log

so now the log / stored variable will look like
```
**file_path:** /file1.kt
**file_path:** /file2.kt
```

and probably by the end it will have all the md reports in one place and by the end of all ci runs on affected / changed files
we will have a single md report in string variable or log, that can be **uploaded as a comment**

This is what I had in my mind for the entire process
Hmm.. are there any conflicts?

### b. HTML reports
As with .html, I assume this to be most useful with local developments
and in the first Android CLAM meet we had discussions on this and decided to just have .html for just local development for now since I mentioned there might be few things to note on how to collate the entire html report
and we decided to stick with just md for now as mentioned in M2 milestone PR 2.1

with HTML the flow will be the same as done with md reports

for **1:1 mapping**
The html file content is generated just with the one - coverage report and[0] and saved as .html file to the output path.
And I am right now sticking with generating it to the path of **$repoRoot/coverage_reports/..../coverage.html** as for few reasons that was described in https://github.com/oppia/oppia-android/pull/5443#issuecomment-2186663348
And that might be changed later as required

For now I am trying to accumulate all line, branch and function coverage details
But since you mentioned we can skip branch coverage, I am thinking of sticking with line and function coverages
And for 1:1 mapping this should be straightforward of just mapping the coverage report data to appropriate representations and the result will look something like below (this is just line coverage, functions need to be added)

Screenshot (1407)

(this is yet to be implemented in PR 1.5)
for **many:1 mapping**
This will also be the same for md generation
**ie.** There will be just one single file say **$repoRoot/coverage_reports/HomeActivity/coverage.html** (as both targets represent coverages for same file)
and first a collective result of the lines covered and not need to be found, similar to md as, 

```
coverage reports[0] had line 1: FULL, line 2: FULL, line 3: NONE, line 4: NONE
coverage reports[1] had line 1: NONE, line 2: FULL, line 3: FULL, line 4: NONE
(to be computed) so both should be computed to result in 
line 1: covered
line 2: covered
line 3: covered
line 4: not covered
```

and this will be same for functions coverage too, and the color of the coverage representation will look like - 

|Line|Branch|Function|Coverage color|
|---|---|---|---|
|FULL|FULL|FULL|GREEN|
|FULL|FULL|NONE|YELLOW|
|FULL|NONE|NONE|YELLOW|
|NONE|FULL|NONE|YELLOW|
|NONE|NONE|NONE|RED|

and the combined result will be made as a html file and saved to the output path.
so thereby we achieve the **consolidated output**

### [No M2 phase for HTML]
But I can see in your comments you mentioning about M2 HTML genertions, if that is something we absolutely need... (I still need to figure that out then) would confirm that and may be look into that for M2

So this is just the entire high level flow of the entire project majorly on M1 details and a outline on M2, from the comments I sense the concerns on 2 main things
1. many:1 report generation
2. the final cumulative report generation for all affected files

Another concern on proto structure to have the covered file as list, that is actually something I too wanted to discuss on I just followed the pattern provided in the issue tracker
```
message CoverageReport {
  string bazel_test_target = 1; // The Bazel test target that was run.
  repeated CoveredFile covered_file = 2; // Files with some coverage in this report.
}
```
but felt irrelevant to how we are proceeding so I think we will need to remove that and have everything under CoverageReport itself.

I think these might address few of the concerns but would definitely need to know if the flow is the same as you think it should be and was the flow right and would like to hear and work on feedbacks.

Rd4dev:
@BenHenning, Hmm... Okay the previous comment just looks like a novel sorry, may be in short to just address the comments,

### M1 phase
We will basically have the list of CoverageReports in RunCoverage.kt that will be sent to generate reports to CoverageReporter.kt
with the list,
eg: 
```
List of CoverageReport ->
[{
  bazel_test_target://app:src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest,
  CoveredLine{line_number:2}
 },
 {
  bazel_test_target://app:src/test/java/org/oppia/android/app/home/HomeActivityLocalTest,
  CoveredLine{line_number:3}
}]
```

we will need to generate a cummulative coverage data in the way like
```
coverage reports[0] had line 1: FULL, line 2: FULL, line 3: NONE, line 4: NONE
coverage reports[1] had line 1: NONE, line 2: FULL, line 3: FULL, line 4: NONE
(to be computed) so both should be computed to result in 
line 1: covered
line 2: covered
line 3: covered
line 4: not covered
```

and use this to data generate md or html reports and save them
so thereby we achieve the **consolidated output**

### M2 phase
The ci run will run this command for each affected file and the plan is to store each run's output md string to a variable or log and at the end upload the collective generated md string as a comment.

BenHenning:
I think we discussed what was left to do here in our meeting today @Rd4dev (and I conveyed that to @adhiamboperes separately). Please let me know if anything else is needed from here, otherwise I'm deferring to @adhiamboperes for reviews over the next ~1.5 days.

Rd4dev:
Thanks @BenHenning, just one thing to confirm. Can I proceed with having the script itself set the default output storage path for the md and html files (as shown in the demo M1)?

The reasons are discussed here https://github.com/oppia/oppia-android/pull/5443#issuecomment-2186663348
But in-short, just for 2 things
1. Handle different OS paths and ensure consistent file structure.
2. Use a default directory for coverage reports to avoid overwriting.

Rd4dev:
### Update on progress: 
For now, the entire logic is done for 
### 1. Generating md reports
**md report:**
## Coverage Report

- **Covered File:** utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
- **Coverage percentage:** 100.00% covered
- **Line coverage:** 19 / 19 lines covered

### 2. Generating html reports
**html report:**
Screenshot (1417)
___
**html report 2: to represent fail cases**
Screenshot (1414)

### TODO:
And things left to do are to add tests

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: As discussed, we will need to do / add 2 things but for M2,
1. Make sure multi targets work in a consolidated way for a single source file. ie. if app/main/java/.../HomeActivity.kt is provided as developer/user input -> this should fetch coverage report from both HomeActivityLocalTest and HomeActivityTest targets for the single source file and present them as a combined report.
2. We need to have the same way of executing ci checks for local checking too. ie. as ci runs will need be based on multiple files/targets provided, those need to be doable in local development too. So something like
```
bazel run //scripts:run_coverage -- $(pwd) [utility/.../MathModel.kt, app/.../HomeActivity.kt, app/..../HelpActivity.kt]
``` 
should be doable

adhiamboperes:
> Thanks @BenHenning, just one thing to confirm. Can I proceed with having the script itself set the default output storage path for the md and html files (as shown in the demo M1)?
> 
> The reasons are discussed here [#5443 (comment)](https://github.com/oppia/oppia-android/pull/5443#issuecomment-2186663348) But in-short, just for 2 things
> 
> 1. Handle different OS paths and ensure consistent file structure.
> 2. Use a default directory for coverage reports to avoid overwriting.


@Rd4dev, I think having the script set the path may be the more logical path here, so the developer doesn't have to figure this out.

Please let me know if you still have other open questions.

Rd4dev:
> @Rd4dev, I think having the script set the path may be the more logical path here, so the developer doesn't have to figure this out.
> 
> Please let me know if you still have other open questions.

Thanks @adhiamboperes, that clears up everything!

Also, as update: The logic and the report generation for this PR are done and tests for `CoverageReporter` are added only thing left is to update the test cases for `RunCoverage`.
(each test runs are taking a bit of time to complete)
Log:
```
INFO: From Testing //scripts/src/javatests/org/oppia/android/scripts/coverage:RunCoverageTest:
==================== Test output for //scripts/src/javatests/org/oppia/android/scripts/coverage:RunCoverageTest:
JUnit4 Test Runner
.
Time: 208.708

OK (1 test)
```
I will push them too once everything is tested and open this PR for review.

scripts/src/java/org/oppia/android/scripts/coverage/BUILD.bazel
Rd4dev: Done! Updated dependency references across all occurrences.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Removed the const declaration, that will be re-introduced in M2 to check for coverage fail / pass cases.

Rd4dev:
@adhiamboperes, the PR 1.5 tasks are done.

- This now introduces the `CoverageReporter` that outputs the coverage data in `.md` and `.html` formats as represented in the https://github.com/oppia/oppia-android/pull/5443#issuecomment-2191913964
- Also updated the tests to check for (app, scripts and default modules) in both (html and md formats)

As of now the reports are returning a pair of **computed coverage ratio** and the actual **report text**.
And `RunCoverage` takes the report text and saves it to the output file (in location _coverage_reports/.../coverage.md or .html_)

**M2**
These report texts and computed coverage ratios are returned with the plan to make them useful in M2 
1. **report text** - to gather all the reports (failed) and upload them as a single _comment_
2. **computed coverage ratio** - to check for minimum threshold to decide _pass / fail_ case

with these plans, I have addressed the existing review comments, would it be possible to PTAL!

.gitignore
Rd4dev: This conversation is addressed with confirmation to https://github.com/oppia/oppia-android/pull/5443#issuecomment-2192915627 by sticking to the approach of letting the script decide the output paths.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
```diff
@@ -0,0 +1,228 @@
...
+      tempFolder.root.absolutePath,
+      listOf(report),
+      ReportFormat.MARKDOWN
+    )
+    val (coverageRatio, _) = reporter.generateRichTextReport()
+    assertThat(expectedZeroCoverageRatio).isEqualTo(coverageRatio)
+  }
+
+  @Test
+  fun testCoverageReporter_markdownReportText_check() {
```
adhiamboperes:
```suggestion
  fun testCoverageReporter_generateMarkdownReport_hasCorrectContentAndFormatting() {
```

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -131,52 +135,101 @@ class RunCoverageTest {
...
+
+    val sourceContent =
+      """
+      package com.example
+      
+      class TwoSum {
+      
+          companion object {
+              fun sumNumbers(a: Int, b: Int): Any {
+                  return if (a ==0 && b == 0) {
```
adhiamboperes: Nit:
```suggestion
                  return if (a == 0 && b == 0) {
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -136,6 +170,19 @@ private fun findTestFile(repoRoot: String, filePath: String): List<String> {
...
+  repoRoot: String,
+  filePath: String,
+  reportFormat: ReportFormat
+): String {
+  val fileWithoutExtension = filePath.substringBeforeLast(".")
+  val defaultFilename = when (reportFormat) {
+    ReportFormat.HTML -> "coverage.html"
+    ReportFormat.MARKDOWN -> "coverage.md"
+  }
+  return "$repoRoot/coverage_reports/$fileWithoutExtension/$defaultFilename"
```
adhiamboperes, line 184: Would it be better to have the name as maybe MathModelCoverage.md compared to /MathModel/coverage.md? I am asking in case I am missing something about how the coverage reports are to be used for the combined report in M2.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -0,0 +1,198 @@
...
+   */
+  fun generateRichTextReport(): Pair<Float, String> {
+    println("report format: $reportFormat")
+    return when (reportFormat) {
+      ReportFormat.MARKDOWN -> generateMarkdownReport()
+      ReportFormat.HTML -> generateHtmlReport()
+    }
+  }
+
+  private fun generateMarkdownReport(): Pair<Float, String> {
```
adhiamboperes, line 42: My concern with the MD report is that unlike the html report, it does not show any information about which lines were not covered, so that it does not tell the developer what to do to next. Is that by design?

adhiamboperes (changes requested):
@Rd4dev, PTAL at some comments that I have.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes: I think I have answered my own question, I generated reports with both formats and the resulting directory structure looks good.
Screenshot 2024-06-27 at 21 39 14

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -31,6 +33,13 @@ import java.util.concurrent.TimeUnit
 fun main(vararg args: String) {
   val repoRoot = args[0]
   val filePath = args[1]
+  val reportFormat = when (args.getOrNull(2)) {
+    "HTML" -> ReportFormat.HTML
+    "MARKDOWN", null -> ReportFormat.MARKDOWN
```
adhiamboperes: Can these be specified as ignorecase? using the format with lowercase fails

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -0,0 +1,198 @@
...
+            display: inline-block;
+        }
+        .covered-line {
+            background-color: #c8e6c9; /* Light green */
+        }
+        .not-covered-line {
+            background-color: #ffcdd2; /* Light red */
+        }
+        .uncovered-line {
+            background-color: #fafafa; /* Half white */
```
adhiamboperes: Given that this file renders in light mode, the white is barely visible and it took me a really long time to notice. Maybe change to something like light gray or light yellow.

adhiamboperes (commented):
A couple more followups

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Yeah @adhiamboperes, it makes sense that it looks like lacking details, but the plan (as far as I understood) was MD reports are majorly for ci runs to provide a short summary of the coverage that will be uploaded as GitHub comments (in ci which would be multiple files) and HTML reports for detailed coverage.

And during the meetings Ben suggested the templates for how these should look like:

> Any template for report generation?
> HTML: directory structure that matches the files under test, and then copies code over with red/green/yellow highlighting based on coverage type (probably can’t do, so just red/green for miss/hit).
> MD: should focus on a summarization of the test results, probably something like a cascading bullet list representing the covered tests and their percentages and covered/total lines of code, so something like:
> 
> Total coverage:
> Files covered: (# changed / # run with coverage)
> Coverage percentage: ##% covered / ##% expected
> LOC: # covered / # instrumented
> 
> (indent left) Specific coverage:
> app/src/.../app
> home
> HomeActivity.kt - 87% (110/115)
> ...
> ...
> Note: the path abbreviation is a good nice-to-have, but not required. Want to reduce cascading. Nice-to-have also would be adding a Markdown link to the file, e.g. for HomeActivity.kt above.

_(also the provided template it for M2 (when we have multiple files to check coverage for))_

and this was followed along, but if that's absolutely needed and could be useful maybe we can add it later as per feedback?

adhiamboperes: Ok, it makes sense that we plan on expanding the details when we have more files/data

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks, Adhiambo, certainly missed considering the case sensitivity scenarios. Updated them with `ignoreCase`.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Thanks Adhiambo, so the uncovered lines are the ones represented in the kind of half white `#fafafa`. They basically didn't require any highlighting (lcov genhtml reports leave them as plain lines), I thought maybe it would look nice to separate lines of code from the background so added that. But as said it's barely visible because I didn't want to really make it any differentiable but still give a slight elevation. 

I tried adding a light gray `#F1F1F1`, I think that looks like a better way to represent as you said too. (Yellow `#FCEE9E` couldn't be possibly added as it is something I am thinking of reserving for partially covered data when working with many:1 target files). Also, I think the top summary possibly needed some design changes too so just retouched on few things. Can you check if this looks decent? 
_(Definitely needs a lot of re styling to look good but just wanted to keep it simple and usable for now)_

Screenshot (1426)

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks @adhiamboperes, hope no changes need to be made then? or was the suggestion to rename it to coverage_reports/utility/..../parser/math/MathModelCoverage.md

Not sure, but to also answer that (just incase), just having the structure to have `coverage_reports/utility/..../parser/math/MathModelCoverage.md` could cause a collapse of multiple other scripts/files reports to be in the same place, causing something like
```
(mock)
coverage_reports/
  utility/
    ..../
      parser/
        math/
          DivModelCoverage.md 
          FractionModelCoverage.md 
          FractionModelCoverage.html
          MathModelCoverage.md 
```

Guess the current approach would make it clean.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Thanks! Fixed it in all occurrences.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
Rd4dev: Done, also updated the `testCoverageReporter_htmlReportText_check` in the same way to `testCoverageReporter_generateHTMLReport_hasCorrectContentAndFormatting`

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: This is also addressed as per https://github.com/oppia/oppia-android/pull/5443#issuecomment-2192915627

Rd4dev:
Thanks for the review @adhiamboperes, addressed all the review comments and updated tests accordingly. Can you PTAL!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes: Yes, I'm happy with the current solution.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
adhiamboperes: Verifying that for this PR, the script runs for a single source file.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
adhiamboperes: Light gray loooks ok. Half white was not visible, so it looked a bit like nothing had happened to those lines. Gray gives the impression that they were checked, just uncovered.

adhiamboperes (approved):
Thanks @Rd4dev, LGTM!

adhiamboperes:
@seanlip, could you PTAL? I have verified that @BenHenning's changes have been addressed.

seanlip (approved):
Approving for codeowners since Ben is away

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Ah got it! Thanks for the suggestion.

PR #5453, opened 2024-06-28
Fix part of #5343: Code Coverage script edge cases

Rd4dev:
## Explanation

Fixes part of #5343 

This is an extended part of Milestone 1 to address any edge cases.

### Project
[Project 4.1]

### Changes Made

- Added arg names to the run_coverage scripts to cover edge cases on handling formats and reordering the arguments. 

Updated script run command:
```
bazel run //scripts:run_coverage -- $(pwd) utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt format=HTML
```

---

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
Hi @adhiamboperes, addressed the argument name / ordering edge case as suggested. Can you PTAL!
Also changed exception statement to be more descriptive.

Rd4dev:
@adhiamboperes added tests for case sensitivity and reordered arguments, would it require any other tests?

adhiamboperes (approved):
Thanks @Rd4dev. This LGTM.

Rd4dev:
@adhiamboperes the tests fail due to **TIMEOUT**, I tried changing them in the test `longCommandExecutor` but that didn't seem to work: https://github.com/oppia/oppia-android/actions/runs/9717824867/job/26824483404?pr=5453#step:17:932

adhiamboperes:
@seanlip, we discovered some edge cases from RD's last PR, and made a follow up PR instead of reverting the last one. cc @DubeySandeep

adhiamboperes:
@Rd4dev PTAL for CI failures

Rd4dev:
Hey @adhiamboperes, I just tried reverting all the changes to see if those are causing any issues and the tests though timed out at Attempt 1 seems to pass from Attempt 2: [stack trace from Attempt 2](https://github.com/oppia/oppia-android/actions/runs/9721159638/job/26833526588?pr=5453#step:17:335)
_(All checks passed)_

So, there is something from the newly introduced changes (which is only the arg parsing) that was messing up the tests.
I will try to re-introduce them and see what's actually happening.

Edit: There is nothing failing with the logic (all checks passed reverting the test cases), its just the addition of tests making it wait too long causing **TIMEOUT**!

### Passing locally:

image

Edit:
Tried changing the processTimeout in computeAffectedTestsTest and they properly reflect the timeout setting: https://github.com/oppia/oppia-android/pull/5453/commits/100d8f11de220618227648e9f2344d4fd2ab1a9e

Edit:
The processTimeout works if I set it to a lower value but doesn't reflect if I choose a higher value, guess that's a limitation to timeout at max 300s.

This is the stack trace of : https://github.com/oppia/oppia-android/pull/5453/commits/e82f2fdb60d111f380217a13f05c4c5c4103d626 where I tried to set `processTimeout` to 1L MILLISECOND and that is reflected.
image

I am not sure where the actual max limitation is set (if that's something that is restricting the processTimeout to go beyond 300s which is actually needed to run coverages).

Edit: I probably doubt if this could be due to varied processTimeout references at various places.
If that's TODO: Test with a common `commandExecutor` value for RunCoverage

Edit: Tried increasing the processTimeout in `computeAffectedTestsTest` to 10 minutes while adding a sleep of 350 seconds to the test and that works [[stack trace](https://github.com/oppia/oppia-android/actions/runs/9728015513/job/26848316526?pr=5453#step:17:1391)], so there is no max limit its just the processTimeout somewhere messing in the `RunCoverageTest`.

Edit: I doubt the timeout is not actually related to RunCoverage script settings solely but was failing because of the kotlin coroutine timeouts 

Edit: Tried testing with a more time intensive test case locally - assertEquals(computeFibonacci(45), 1134903170L) and that failed with kotlin coroutine timed out error and processTimeout set via RunCoverage was ignored. Its ultimately the runBlocking timing out (earlier the `computeAffectedTestsTest ` processed with a higher time out value since it didn't use any coroutines)

Edit: Tried adding a delay to the MavenDependeciesCheckList, in ci it throws the same TIMED OUT error on Coroutines but here the timeout occured after 800+ seconds which was 300s for RunCoverage [[stack trace](https://github.com/oppia/oppia-android/actions/runs/9730804785/job/26854463693?pr=5453#step:17:3868)]

### Fix
Edit: Ok, I think I understood where the actual problem is, so it was basically due to the test exceeding bazel's --test_timeout value, it was set to [bazel runner doc](https://bazel.build/reference/test-encyclopedia#role-test-runner) bazel test's moderate value causing a timeout at 300s. And increasing the timeout value using `--test_timeout` flag fixes that. [[checks passing in ci](https://github.com/oppia/oppia-android/actions/runs/9736939000/job/26868354715?pr=5453#step:17:3999)]

Rd4dev:
Hey @adhiamboperes, following up on [this workaround](https://github.com/oppia/oppia-android/pull/5453#issuecomment-2197886284), I was able to figure out what the issue is.

The tests exceeded bazel's timeout causing SIGTERM and as per [bazel's timeout doc](https://bazel.build/reference/be/common-definitions#test.timeout),

> Bazel uses the size to determine a default timeout, which can be overridden using the timeout attribute.

| Size     | RAM (in MB) | CPU (in CPU cores) | Default timeout       |
|----------|--------------|--------------------|-----------------------|
| small    | 20           | 1                  | short (1 minute)      |
| medium   | 100          | 1                  | moderate (5 minutes)  |
| large    | 300          | 1                  | long (15 minutes)     |
| enormous | 800          | 1                  | eternal (60 minutes)  |


so I guess, it was allocating a default timeout of 300 seconds for this test, which was causing the max limit.

And I tested them both locally and in ci with a `--test_timeout` flag and they both **PASSED** successfully.

In CI it takes ~350 to 400 seconds: [[stack trace](https://github.com/oppia/oppia-android/actions/runs/9739581796/job/26875117040?pr=5453#step:17:4214)]
**Screenshot**
image

And locally it took me ~830 seconds
**Screenshot**
Screenshot (1440)

I tried setting this in unit test workflow run to `--test_timeout=600` https://github.com/oppia/oppia-android/pull/5453/commits/fe642a6b060e6f5579879e7bac907cfcfbacd0c8 and the tests ran successfully.

But the timeout set was just an approximation to be over what it took in the past ci run and doing this would also set it for the other test runs too and that is something to look into.

Can you PTAL whenever possible!

_(Also did few clean ups in the tests in this PR as felt I added a lot of redundant code earlier for source and test content)_

Rd4dev:
Additionally, I am concerned that the extended test duration may be due to the broader instrumentation filter? [[PR 1.3 modified instrumentation](https://github.com/oppia/oppia-android/pull/5433/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63d)]

.github/workflows/unit_tests.yml
```diff
@@ -286,7 +286,7 @@ jobs:
           while [ $i -ne 5 ]; do
             i=$(( $i+1 ))
             echo "Attempt $i/5 to run test targets"
-            bazel test --keep_going --remote_http_cache=$BAZEL_REMOTE_CACHE_URL --google_credentials=./config/oppia-dev-workflow-remote-cache-credentials.json -- $BAZEL_TEST_TARGETS
+            bazel test --keep_going --test_timeout=600 --remote_http_cache=$BAZEL_REMOTE_CACHE_URL --google_credentials=./config/oppia-dev-workflow-remote-cache-credentials.json -- $BAZEL_TEST_TARGETS
```
Rd4dev: **Try / TODO:**
If the broader instrumentation filter range is causing the issue, try narrowing it down more!

**eg:**

- for target - //utility/src/test/java/org/oppia/android/util/parser/math:MathModelTest -> //utility/src/test/java/org/oppia/android/util/parser/math/... (not sure if that's to be done with test or actual source file (if source) -> (same logic to map tests but in reverse) //utility/src/main/java/org/oppia/android/util/parser/math/...

- for target - //app:src/sharedTest/java/org/oppia/android/app/home/WelcomeViewModelTest -> //app/... (basically separating the package structure by ':')

Rd4dev: Ok, I think that's not going to be helpful, at least for the test cases.
Reason being these test cases don't actually have any other coverages to analyse so having a broader instrumentation shouldn't be making much of a difference here?

Narrowing it down for the actual script runs could be something to think, but 
1. The `--instrumentation_filter` was predominantly introduced to make sure sharedTest tests get covered, but if the sharedTest's get modularized then that will also be narrowed to their sub packages, which will again make the coverage attempt fail.
2. If that's necessary, then we will need to have separate ways of handling the instrumentation for app modules (broader `//app/...`) from the default and script modules (`//default/src/main/java/com/example/package/...`).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -60,7 +60,9 @@ class CoverageRunner(
     val sfStartIdx = coverageData.indexOfFirst {
       it.startsWith("SF:") && it.substringAfter("SF:").substringAfterLast("/") == extractedFileName
     }
-    if (sfStartIdx == -1) throw IllegalArgumentException("File not found")
+    if (sfStartIdx == -1) throw IllegalArgumentException(
+      "Coverage data not found for the file: $extractedFileName"
+    )
```
```suggestion
    if (sfStartIdx == -1) throw IllegalArgumentException(
      "Coverage data not found for the file: $extractedFileName"
    )
```
```suggestion
    require(sfStartIdx != -1) {
      "Coverage data not found for the file: $extractedFileName"
    }
```

More Kotlin-idiomatic way to handle this.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -68,64 +105,120 @@ class RunCoverageTest {
...
+    )
+
+    val outputFilePath = "${tempFolder.root}" +
+      "$coverageDir/${filePath.removeSuffix(".kt")}/coverage.md"
+
+    assertThat(File(outputFilePath).exists()).isTrue()
+  }
+
+  @Test
+  fun testRunCoverage_ignoreCaseHTMLArgument_returnsCoverageData() {
```
BenHenning:
```suggestion
  fun testRunCoverage_ignoreCaseHtmlArgument_returnsCoverageData() {
```

Per https://developer.android.com/kotlin/style-guide#camel_case. Ditto elsewhere in this PR for identifiers.

.github/workflows/unit_tests.yml
BenHenning: Should this be reverted? It doesn't seem relevant to this PR (unless I'm missing something).

BenHenning (commented):
Thanks @Rd4dev! Looks like a nice simplification.

.github/workflows/unit_tests.yml
Rd4dev: Sorry @BenHenning if this https://github.com/oppia/oppia-android/pull/5453#issuecomment-2199847753 ([Issue stack trace comment](https://github.com/oppia/oppia-android/pull/5453#issuecomment-2197886284)) was missed earlier or if I'm just sending a duplicate. Because of timeout issues, I had to introduce `--test_timeout=600`.

The current comment thread was something I thought could be a workaround, but I don't think that could help with the tests for the reasons mentioned above so I opted to leave it unresolved to explore other potential possibilities.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -49,7 +53,7 @@ fun main(vararg args: String) {
   ScriptBackgroundCoroutineDispatcher().use { scriptBgDispatcher ->
     val processTimeout: Long = args.find { it.startsWith("processTimeout=") }
       ?.substringAfter("=")
-      ?.toLongOrNull() ?: 5
+      ?.toLongOrNull() ?: 10
```
adhiamboperes: This can still be 5 because thet imeout issue is test specific and this change does not impact it.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Thanks for the resource. Updated all occurrences as guided.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Got it! Updated accordingly. Thanks.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -23,20 +23,57 @@ class RunCoverageTest {
...
+        @Test
+        fun testSumNumbers() {
+          assertEquals(TwoSum.sumNumbers(0, 1), 1)
+          assertEquals(TwoSum.sumNumbers(3, 4), 7)         
+          assertEquals(TwoSum.sumNumbers(0, 0), "Both numbers are zero")
+        }
+      }
+      """.trimIndent()
+
     System.setOut(PrintStream(outContent))
```
adhiamboperes: This is not needed as we are not using the value anyway in the test.

.github/workflows/unit_tests.yml
adhiamboperes: Inline with the discussion at the meeting -- the timeouts are more likely  to be an issue with the test setup than the CI configs, so  you will need to revert this and below. 

For instance, I pulled your changes and ran the RunCoverageTest on my Mac, and had similar timeouts. These yml configs have no impact on running tests locally, so adjusting the timeout on CI is more likely burying the real problem.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes, line 1: The returned value of the execute() function is actually unused, so it can be refactored to be a unit function instead.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -2028,49 +816,46 @@ class RunCoverageTest {
...
       """.trimIndent()
 
-    assertThat(outputReportText).isEqualTo(expectedResult)
+    return htmlText
   }
 
   private fun initializeCommandExecutorWithLongProcessWaitTime(): CommandExecutorImpl {
     return CommandExecutorImpl(
-      scriptBgDispatcher, processTimeout = 5, processTimeoutUnit = TimeUnit.MINUTES
+      scriptBgDispatcher, processTimeout = 10, processTimeoutUnit = TimeUnit.MINUTES
```
adhiamboperes: Same comment on reverting the timeout

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes: Are there long running operations in this function that could be run on the background thread? This could be the cause of  some of the tests running long.

adhiamboperes (changes requested):
@Rd4dev, PTAL.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Before reverting the change, I wanted to explain my reasoning behind the `processTimeout` update in **RunCoverage.kt**.

During local coverage analysis, I often encountered files in large packages taking over **300 seconds** to process, leading to **"process didn't complete within the intended time"** errors. 

image

To address this, I proposed increasing the timeout to 10 minutes _(by default to give ample time)_. This duration can vary depending on the files and the device and having the `processTimeout` configurable via command-line arguments should be useful for this.

Would you prefer keeping it at 5 minutes and allowing developers to adjust the processTimeout through command-line arguments if needed, or can we set it to 10 minutes as a middle ground? Would like to confirm how it could be done, and I will revert the changes as guided.

Thanks.

.github/workflows/unit_tests.yml
Rd4dev: Hey @adhiamboperes, thanks for testing it. 
So currently there are 2 issues _(provided in the Android CLaM meeting chat thread)_.

---

### Error with fetching in ci (not related to this current comment thread)
_(not related to `--test_timeout` issue)_

The issue we discussed in the meeting was related to the [ERROR: An error occurred during the fetch of repository 'remotejdk11_linux':](https://github.com/oppia/oppia-android/actions/runs/9860889216/job/27228421626?pr=5453#step:17:348) and that might be due to some mismatch of configurations with the `TestBazelWorkspace` setups (as you mentioned) which I am still trying to figure out. 

To give a bit of context, I tried running the **RunCoverageTest.kt** locally a couple of times and they do pass locally. So, its somewhere in the setup that has to have some mismatch configurations with ci, and causing the underlying issue.

image

image


---

### --test_timeout
_(related to the current comment thread)_

Regarding the introduction of **--test_timeout**, I was experimenting with a few solutions, and this was one potential fix I found. However, it was **not intended to be merged** without further discussion with Ben. To keep this issue on our radar, I left the comment unresolved. I plan to discuss the underlying issues further to determine the best course of action.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Hey @adhiamboperes, thanks for the suggestion on background threads.

Actually, `retrieveCoverageResult` is the operation that should supposedly take too long, but it's already managed within **CoroutineScope(scriptBgDispatcher).async { ... }** using **ScriptBackgroundCoroutineDispatcher**. I think this setup should delegate the operation to the background threads. 

**Log of thread name:**
```
//utility/src/test/java/org/oppia/android/util/math:NumericExpressionEvaluatorTest start: 2024-07-08T22:32:08.714519 on thread: pool-1-thread-1
```

While these operations being asynchronous and running in the background, we utilize **runBlocking** to ensure all tasks are completed and final output is obtained as a curated list of outputs.

Would this require any changes?

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: @adhiamboperes, just to clarify, was the suggestion to remove `outContent`? It's used in the `testRunCoverage_testFileExempted_noCoverage` test case to capture output. Or was it a suggestion to remove something else?

Rd4dev:
Hey @adhiamboperes,

Regarding the error on fetching the repo issue we discussed, I checked the setup, and although it passes locally, I suspected misconfigurations in the **TestBazelWorkspace.kt** file. It appears that the existing implementations didn't specify a **Java runtime version** in their **.bazelrc** file. But that was introduced in the [Migration to Bazel 6.5.0](https://github.com/oppia/oppia-android/pull/4886/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832fR8) PR to have `build --java_runtime_version=remotejdk_11 --tool_java_runtime_version=remotejdk_11` which might be causing this misconfiguration. 

After setting up the Java version in the **TestBazelWorkspace's** **.bazelrc** (bazelRcFile) file - https://github.com/oppia/oppia-android/pull/5453/commits/4c8757306e71ba5476fe63ae8c64d8905eea02b6, the recent CI runs seem to pass.

**To note:**
- Although remotejdk tools were introduced in the Migration to Bazel 6.5.0, the following PRs didn't cause this issue, which makes me wonder (if any of the recent changes to any remote JDK (I don't even know if that's a possible scenario) might be the cause).
- These needs confirmation before proceeding, as changes were made to existing implementations.
_(Will also make sure to run it a couple more times in ci to confirm this)_

And addressed the review comments and would like to clarify changes requested on few places and then proceed accordingly, Can you PTAL.

---

Also @BenHenning, I addressed the other review comments, except the issue related to `--test_target` introduction.
Also, for reference, pointing to the [latest stack trace ](https://github.com/oppia/oppia-android/actions/runs/9883098823/job/27297265069?pr=5453#step:17:4100) which took **RunCoverageTest** **~350 seconds** to execute.

In the meantime, I will look into what's causing the test to take this long and see if I can identify any issues. And will discuss any findings whenever possible.

.github/workflows/unit_tests.yml
Rd4dev: Oh, also missed mentioning this in the previous comment.

> For instance, I pulled your changes and ran the RunCoverageTest on my Mac, and had similar timeouts. These yml configs have no impact on running tests locally

Yeah! that's the case, so even with the local test it fails with the **SIGTERM** error. I had to add `--test_timeout` to the bazel test to give it the time to actually complete the test, and I replicated the same in the ci.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Got it! Refactored the execute() to be a unit function.

Rd4dev:
@BenHenning, @adhiamboperes, I think I finally figured it out _(shouldn't have taken this long)._ 

The test runtimes consistently exceed 300 seconds, so the previous implementation introduced `--test_timeout` to the Bazel test runs. However, there were concerns that this would affect the entire run rather than specific test targets. 

It turned out to be a straightforward fix https://github.com/oppia/oppia-android/pull/5453/commits/75e28e2d421f65290e515df66108084c88099482 to add `size = "large"` to the test rule [[bazel implied timeout labels](https://bazel.build/reference/test-encyclopedia#role-test-runner)], similar to what was done for other test targets ([MavenDependenciesListCheckTest](https://github.com/oppia/oppia-android/blob/1c4b17a10fc42dd8efb14d7e7f4841ddff598607/scripts/src/javatests/org/oppia/android/scripts/license/BUILD.bazel#L20), [ComputeAffectedTestsTest](https://github.com/oppia/oppia-android/blob/1c4b17a10fc42dd8efb14d7e7f4841ddff598607/scripts/src/javatests/org/oppia/android/scripts/ci/BUILD.bazel#L9)). This change should set the timeout for the test target to (long) 900 seconds, allowing it to complete its execution.

> Most recent ci run passing in ~480 seconds [[stack trace](https://github.com/oppia/oppia-android/actions/runs/9893021303/job/27327520924?pr=5453#step:17:3860)]

Uhm.. I'd like to know if this is the right approach and hope this resolves everything related to `--test_timeout` issues!

Edit: Also, I wonder if this should be subdivided to help parallelize test execution using `shard_count`, as done with other large test targets. If yes, the count needs to be determined. Looking to discuss this and proceed.

.github/workflows/unit_tests.yml
Rd4dev: This should now be addressed with replacing it with `size = "large"` for specific test targets as mentioned in https://github.com/oppia/oppia-android/pull/5453#issuecomment-2223125478

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
adhiamboperes: Frankly, I only mention 5 minutes becase it is consistent with the rest of the test configuration, and since we allow the developer to configure a timeout. Configuring a timeout is agreat workaround and the duration wil depend on a developer's machine performance. (Apologies, I missed this comment.)

scripts/src/java/org/oppia/android/scripts/testing/TestBazelWorkspace.kt
```diff
@@ -30,7 +30,14 @@ class TestBazelWorkspace(private val temporaryRootFolder: TemporaryFolder) {
...
   private val bazelRcFile by lazy {
-    temporaryRootFolder.newFile(".bazelrc").also { it.writeText("--noenable_bzlmod") }
+    temporaryRootFolder.newFile(".bazelrc").also {
+      it.writeText(
+        """
+          --noenable_bzlmod
+          build --java_runtime_version=remotejdk_11 --tool_java_runtime_version=remotejdk_11
+        """.trimIndent()
+      )
+    }
```
adhiamboperes, line 40: Looks like a reasonable workaround for the jdk issue. I  am highlighting this for @BenHenning to look at post-merge. Same to the corresponding test file changes.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
adhiamboperes: If it affects one test I think it is safe to do that at the start of the test as opposed to here(because it does it now for every test).

adhiamboperes:
@Rd4dev, good job on identifying the problem! 

Re: 

>  I wonder if this should be subdivided to help parallelize test execution using shard_count, as done with other large test targets.

That could be discussed later, I will not block merging this PR on that. Can you add it to your list of tasks for CI validations later in M2?

.github/workflows/unit_tests.yml
Rd4dev: Resolving this comment since the `--test_timeout` has been reverted and replaced by setting `size = "large"` for the specific test target.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Got it! I've reverted it to **5 minutes** for consistent timing _(considering we do have the configuration available)_. Thanks!

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Okay! I've moved it down to the specific test case.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Done!

Rd4dev:
> @Rd4dev, good job on identifying the problem!
> 
> Re:
> 
> > I wonder if this should be subdivided to help parallelize test execution using shard_count, as done with other large test targets.
> 
> That could be discussed later, I will not block merging this PR on that. Can you add it to your list of tasks for CI validations later in M2?

Thanks!

Sure, I'll include it as a sub-task for the M2 list.

Rd4dev:
Hey @adhiamboperes, 

The tests failed with the following error:

```
1) testRunCoverage_scriptTestsMarkdownFormat_returnsCoverageData(org.oppia.android.scripts.coverage.RunCoverageTest)
java.lang.IllegalStateException: Expected non-zero exit code (not 1) for command: [bazel, query, --noshow_progress, --keep_going, set(scripts/javatests/com/example/TwoSumTest.kt)].
Standard output:
Error output:
$TEST_TMPDIR defined: output root default is '/home/runner/.cache/bazel/_bazel_runner/ef69b31fa14ea8341d6589b0a8832bdd/sandbox/linux-sandbox/652/execroot/__main__/_tmp/fdac3d66717137a77a91b3f32165a80f' and max_idle_secs def
================================================================================
INFO: Elapsed time: 333.443s, Critical Path: 333.28s
```

They passed locally, so I tried adding **shards_count** and set it to 4, which is the minimum for other large tests. This significantly reduced the run time _(from ~350 seconds to ~240 seconds)_ as they now run in parallel. It also seems to help with timeouts and the **max_idle_secs** issue, as introducing it solved the problem. I verified it by removing and re-adding it; the issue reappears when it's removed [[failure stack trace](https://github.com/oppia/oppia-android/actions/runs/9911430499/job/27384328646?pr=5453#step:17:4004)] and is resolved when re-added [[success stack trace](https://github.com/oppia/oppia-android/actions/runs/9911973922/job/27385990088?pr=5453#step:17:4844)]. I guess the issue should be with tests taking too long to complete and hitting idle times and causing this issue.

And this was something mentioned to be considered for M2 but given the priority to merge this PR ASAP and the small fix it represents; I'm including it here. The exact number of shards can be updated in M2 if needed. Also, I believe this will be very beneficial for upcoming PRs, as they are likely to introduce more test cases with multiple test suites. Can you PTAL.

Thanks!

adhiamboperes (approved):
Thanks @Rd4dev! Latest changes LGTM.

PR #5457, opened 2024-07-05
Fix Part of #4938: Language Selection Config and New Profile Creation Flow

adhiamboperes:
## Explanation
Fixes Part of #4938: Modifies the profile creation flow and sets the app/audio language selection during onboarding.

### Default Profile Creation
A default empty profile is created when the Onboarding screen is opened for the first time. This is necessary to provide a ProfileId to use when saving the selected app language.

Because this will be the first profile on the app in onboarding v2, it is an admin.

Provisions have been made so that, should the user exit the app before completing the onboarding flow, this profile will be fetched, to prevent multiple profile creation.

### App Language Selection
The language selector will be shown to the user on initial app launch, or if profile onboarding is not complete. 

- There will be a pre-filled language option based on the locale of the device when the app is installed. If the locale is unsupported, English will be the default selection.
- A user can select any preferred supported app language from the dropdown list.
- The existing language functionality/behavior will be retained.

Tests have been added to verify these requirements, and efforts have been made to ensure language selection persists on configuration change. I noticed during testing that failure to do this resulted in an unpleasant UX.

### Profile Nickname and Picture
The "Create profile Screen" has been repurposed to update the default profile instead, providing the remaining profile properties(ProfileType, name, avatar)

Checks have been added to check for profile creation errors, consistent with the legacy flow. 

A new function has been added to the ProfileManagementController to allow for batch update of these fields, and corresponding tests have been added.

### Audio Language Selection
The final step of onboarding is the audio language selection, and there will be a pre-filled language selectionas follows:

- Selected app language(from first onboarding screen) if available as audio language. (Audio language need not be completed)
- Otherwise (if available  as audio language), audio language of the administrator account -- this will be added downstream in M3, as it only impacts additional learners.
- Otherwise (if available as audio language), device language.
- Else, English.

There have been some incidental changes in AudioLanguageFragment and OptionsFragment, and their related tests due to sharing of the classes between the existing screens and the new screens.

### ProfileTestHelper.kt
A new function, and related tests, have been added to create a default profile to be used in tests.

## Essential Checklist

## For UI-specific PRs only
The screen recordings are in [this drive location](https://drive.google.com/drive/folders/1CXTAALPgpCKfekOQKjRFTb5G2fPgWkRt?usp=sharing), since github does not support webm.


model/src/main/proto/arguments.proto
```diff
@@ -841,3 +844,27 @@ message IntroActivityParams {
...
+message CreateProfileFragmentArguments {
+  // The ProfileType of the new profile as implied by the user's selection.
+  ProfileType profile_type = 1;
+}
+
+// The bundle of properties that are saved on configuration change in OnboardingFragment.
+message OnboardingFragmentStateBundle {
+  // The selected language display name.
+  string selected_language = 1;
+}
```
adhiamboperes: newline at eof.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
```diff
@@ -188,11 +188,45 @@ class AppLanguageResourceHandler @Inject constructor(
     }
   }
 
+  /**
+   * Returns an [OppiaLanguage] from its human-readable, localized representation.
```
adhiamboperes: Update kdocs here and below.

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
```diff
@@ -305,22 +316,13 @@ class AudioLanguageFragmentTest {
...
-      onView(withId(R.id.audio_language_subtitle)).check(
-        matches(withText(context.getString(R.string.audio_language_fragment_subtitle)))
-      )
-      onView(withId(R.id.onboarding_navigation_back)).check(
-        matches(withEffectiveVisibility(Visibility.VISIBLE))
-      )
-      onView(withId(R.id.onboarding_navigation_continue)).check(
-        matches(withEffectiveVisibility(Visibility.VISIBLE))
-      )
+      // Verifies that aceepting the default language selection works correctly.
```
adhiamboperes: Typo

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -658,6 +660,69 @@ class ProfileManagementController @Inject constructor(
     }
   }
 
+  /**
+   * Updates the provided details of an newly created profile to migrate onboarding flow v2 support.
+   * @param profileId The ID of the profile to update.
+   * @param avatarImagePath The path to the selected image.
+   * @param colorRgb The randomly selected unique color to be used in place of a picture.
+   * @param newName The nickname to identify the profile.
+   * @param isAdmin Boolean representing whether the profile has administrator privileges.
```
adhiamboperes: Newline between last param and return.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileViewModel.kt
```diff
@@ -9,6 +9,9 @@ import javax.inject.Inject
 @FragmentScope
 class CreateProfileViewModel @Inject constructor() : ObservableViewModel() {
 
-  /** ObservableField that tracks whether creating a nickname has triggered an error condition. */
+  /** [ObservableField] that tracks whether creating a nickname has triggered an error condition. */
```
adhiamboperes:
```suggestion
  /** [ObservableField] that tracks whether creating a profile has triggered an error condition. */
```

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingFragmentTest.kt
```diff
@@ -792,6 +806,301 @@ class OnboardingFragmentTest {
...
+      val displayLocale = appLanguageLocaleHandler.getDisplayLocale()
+      val layoutDirection = displayLocale.getLayoutDirection()
+      assertThat(layoutDirection).isEqualTo(ViewCompat.LAYOUT_DIRECTION_LTR)
+    }
+  }
+
+  @Test
+  @DefineAppLanguageLocaleContext(
+    oppiaLanguageEnumId = OppiaLanguage.LANGUAGE_UNSPECIFIED_VALUE,
+    appStringIetfTag = "fr",
```
adhiamboperes, line 971:
```suggestion
    appStringIetfTag = "fr-CA",
```

adhiamboperes (commented):
Self-review pass

scripts/assets/file_content_validation_checks.textproto
```diff
@@ -271,6 +271,7 @@ file_content_checks {
   exempted_file_name: "app/src/main/java/org/oppia/android/app/player/audio/LanguageDialogFragment.kt"
   exempted_file_name: "app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/AppVersionActivityTest.kt"
   exempted_file_name: "app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt"
+  exempted_file_name: "app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingFragmentTest.kt"
```
adhiamboperes, line 279: @BenHenning, not sure there is a way to use OppiaLocale in forcing default locale in tests, so I exempted the test file from this check.

app/src/main/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdapters.java
```diff
@@ -15,4 +16,14 @@ public static void setErrorMessage(
   ) {
     textInputLayout.setError(errorMessage);
   }
+
+
+  /** Binding adapter for setting the text of an [AutoCompleteTextView]. */
+  @BindingAdapter({"selection", "filter"})
+  public static void setSelection(
```
BenHenning: Any tests for this?

app/src/main/java/org/oppia/android/app/onboarding/AudioLanguageFragmentPresenter.kt
```diff
@@ -52,22 +87,78 @@ class AudioLanguageFragmentPresenter @Inject constructor(
...
+  private fun updateSelectedAudioLanguage(selectedLanguage: String, profileId: ProfileId) {
+    val audioLanguage =
+      appLanguageResourceHandler.getAudioLanguageFromLocalizedName(selectedLanguage)
+    profileManagementController.updateAudioLanguage(profileId, audioLanguage).toLiveData()
+      .observe(fragment) {
+        when (it) {
+          is AsyncResult.Success -> {
+            val intent = HomeActivity.createHomeActivity(fragment.requireContext(), profileId)
+            fragment.startActivity(intent)
+            fragment.activity?.finishAffinity()
```
BenHenning: Why use finishAffinity here? Seems worthwhile to add a comment, and maybe a test (to ensure its being used over just finish()).

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivityPresenter.kt
```diff
@@ -1,12 +1,20 @@
 package org.oppia.android.app.onboarding
 
+import android.os.Bundle
+import android.view.Gravity.apply
```
BenHenning: Unintentional import?

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
```diff
@@ -107,19 +142,102 @@ class CreateProfileFragmentPresenter @Inject constructor(
...
+        }
+      }
+    )
+  }
+
+  /** Randomly selects a color for the new profile that is not already in use. */
+  private fun selectUniqueRandomColor(): Int {
+    return COLORS_LIST.map {
+      ContextCompat.getColor(fragment.requireContext(), it)
+    }.random()
```
BenHenning:
```suggestion
    return ContextCompat.getColor(fragment.requireContext(), COLORS_LIST.random())
```

Bit more compact and efficient.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -36,13 +84,197 @@ class OnboardingFragmentPresenter @Inject constructor(
...
+              val intent =
+                OnboardingProfileTypeActivity.createOnboardingProfileTypeActivityIntent(activity)
+              intent.decorateWithUserProfileId(profileId)
+              fragment.startActivity(intent)
+            }
+            is AsyncResult.Failure -> oppiaLogger.e(
+              "OnboardingFragment",
+              "Failed to set AppLanguageSelection",
+              result.error
+            )
```
BenHenning, line 166: Should we have some sort of fail-safe here? Otherwise the user will potentially get stuck in this flow and be unable to proceed.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -36,13 +84,197 @@ class OnboardingFragmentPresenter @Inject constructor(
...
+            is AsyncResult.Pending -> {}
+          }
+        }
+      )
+  }
+
+  private fun getSystemLanguage() {
```
BenHenning: Alternative name suggestion: initializeSelectedLanguageToSystemLanguage.

Ditto for ``getSupportedLanguages`` (since these aren't traditional getters).

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -36,13 +84,197 @@ class OnboardingFragmentPresenter @Inject constructor(
...
+              "OnboardingFragment", "Error creating the default profile", result.error
+            )
+            is AsyncResult.Pending -> {}
+          }
+        }
+      )
+  }
+
+  private fun retrieveProfileId(profileList: List<Profile>) {
+    profileId = profileList.firstOrNull()?.id ?: ProfileId.getDefaultInstance()
```
BenHenning: Suggest ininlining this method above, and reworking the logic to avoid the double "if empty" checks happening here (for robustness and simplicity).

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -36,13 +84,197 @@ class OnboardingFragmentPresenter @Inject constructor(
...
+    return profileList
+  }
+
+  private fun createDefaultProfile() {
+    profileManagementController.addProfile(
+      name = "Admin", // TODO(#4938): Refactor to empty name once proper admin profile creation flow
+      // is implemented.
+      pin = "",
+      avatarImagePath = null,
+      allowDownloadAccess = allowDownloadAccess,
```
BenHenning: Can be true always, I think, since the feature is off (and when we enable it in the future there'd be no way to retroactively turn this back on, I think, without a required profile migration).

app/src/main/java/org/oppia/android/app/options/AudioLanguageSelectionViewModel.kt
```diff
@@ -1,19 +1,110 @@
...
+      systemLanguageProvider,
+      PRE_SELECTED_LANGUAGE_PROVIDER_ID
+    ) { appLanguageSelection: AppLanguageSelection, displayLocale: OppiaLocale.DisplayLocale ->
+      val appLanguage = appLanguageSelection.selectedLanguage
+      val systemLanguage = displayLocale.getCurrentLanguage()
+      getPreselection(appLanguage, systemLanguage)
+    }
+  }
+
+  private fun getPreselection(
```
BenHenning: Suggest: computePreselection, though this conflicts with the version below. Maybe rethink the helper names a bit.

Also here & elsewhere in the PR: please make note of member ordering. We generally keep fields together and methods together rather than interlacing them (since that can make it easier to find stuff). Separately, we usually keep public methods above private methods, though there's less consistency for public vs. private fields.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
```diff
@@ -188,11 +188,45 @@ class AppLanguageResourceHandler @Inject constructor(
...
+   * It is expected that each input string is localized to the user's current locale, as per
+   * [computeLocalizedDisplayName].
+   */
+  fun getOppiaLanguageFromDisplayName(displayName: String): OppiaLanguage {
```
BenHenning: I'm not opposed to these methods (this & the next), but I wonder: is it possible to use ``OppiaLanguage`` and ``AudioLanguage`` through the entire data pipeline all the way to the UI, and convert to display name only at the very end so that we never need to convert back to the proto version? I.e. bind the actual proto values to the UI element and just display the human-readable strings.

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -451,6 +466,104 @@ class CreateProfileFragmentTest {
...
+        )
+      testCoroutineDispatchers.runCurrent()
+
+      onView(withId(R.id.create_profile_nickname_error))
+        .check(matches(withEffectiveVisibility(Visibility.GONE)))
+    }
+  }
+
+  @Test
+  fun testFragment_configChange_inputNameWithNumbers_thenInputNameWithLetters_errorIsCleared() {
```
BenHenning: For here & the above: these are technically testing landscape, not configuration changes. I suggest renaming them, and adding separate tests for verifying that a configuration change after the error is initiated retains the error (for both portrait & landscape). That's the more "interesting" behavior to test for config changes.

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
```diff
@@ -335,19 +337,78 @@ class AudioLanguageFragmentTest {
...
-      onView(withId(R.id.onboarding_navigation_continue)).check(
-        matches(withEffectiveVisibility(Visibility.VISIBLE))
-      )
+      // Verifies that accepting the default language selection works correctly.
+      intended(hasComponent(HomeActivity::class.java.name))
+    }
+  }
+
+  @Test
+  fun testFragment_languageSelectionChanged_languageIsUpdated() {
```
BenHenning: For this & the below test: are these actually verifying the language has been set at the domain layer? That seems to be the thing we'd want to be sure of to verify that it's actually being updated.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -392,6 +395,36 @@ class ProfileManagementController @Inject constructor(
     }
   }
 
+  /**
+   * Updates the profile type field of an existing profile.
+   *
+   * @param profileId The ID of the profile to update.
+   * @return A [DataProvider] that represents the result of the update operation.
```
BenHenning:
```suggestion
   * @param profileId The ID of the profile to update
   * @return [DataProvider] that represents the result of the update operation
```

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -658,6 +691,70 @@ class ProfileManagementController @Inject constructor(
...
 
+  /**
+   * Updates the provided details of an newly created profile to migrate onboarding flow v2 support.
+   *
+   * @param profileId The ID of the profile to update
+   * @param avatarImagePath The path to the selected image
+   * @param colorRgb The randomly selected unique color to be used in place of a picture
+   * @param newName The nickname to identify the profile
+   * @param isAdmin Boolean representing whether the profile has administrator privileges
+   * @return A [DataProvider] that represents the result of the update operation
```
BenHenning:
```suggestion
   * @param profileId the ID of the profile to update
   * @param avatarImagePath the path to the profile's avatar image, or null if unset
   * @param colorRgb the randomly selected unique color to be used in place of a picture
   * @param newName the nickname to identify the profile
   * @param isAdmin whether the profile has administrator privileges
   * @return [DataProvider] that represents the result of the update operation
```

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -1278,6 +1279,126 @@ class ProfileManagementControllerTest {
...
+
+    val updateProvider = profileManagementController.updateProfileType(
+      PROFILE_ID_0,
+      ProfileType.ADDITIONAL_LEARNER
+    )
+    monitorFactory.waitForNextSuccessfulResult(updateProvider)
+  }
+
+  @Test
+  fun testUpdateProfile_updateProfileType_newDefaultProfile_checkUpdateSucceeded() {
```
BenHenning: These new tests should also verify that the corresponding fields of the profile were successfully updated.

scripts/assets/file_content_validation_checks.textproto
BenHenning: Should be a fine exemption to use. I generally don't mind tests using ``Locale`` (and in fact, maybe we should just default allow all tests to avoid needing to update this in the future, but I don't feel strongly one way or the other).

testing/src/main/java/org/oppia/android/testing/profile/ProfileTestHelper.kt
```diff
@@ -76,6 +76,16 @@ class ProfileTestHelper @Inject constructor(
     }
   }
 
+  /** Creates one admin profile with default values for all fields. */
+  fun createDefaultProfile() {
```
BenHenning: Perhaps instead: ``createDefaultAdminProfile``?

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
```diff
@@ -102,6 +102,17 @@ class ProfileTestHelperTest {
...
 
+  @Test
+  fun testAddDefaultProfile_createDefaultProfile_checkProfileIsAdded() {
+    profileTestHelper.createDefaultProfile()
+    testCoroutineDispatchers.runCurrent()
+    val profilesProvider = profileManagementController.getProfiles()
+    testCoroutineDispatchers.runCurrent()
+
+    val profiles = monitorFactory.waitForNextSuccessfulResult(profilesProvider)
+    assertThat(profiles).hasSize(1)
```
BenHenning, line 113: Probably should check that it's an admin, as well.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
```diff
@@ -107,19 +142,102 @@ class CreateProfileFragmentPresenter @Inject constructor(
...
+          }
+          is AsyncResult.Failure -> {
+            createProfileViewModel.hasErrorMessage.set(true)
+
+            val errorMessage = when (result.error) {
+              is ProfileManagementController.ProfileNameOnlyLettersException ->
+                appLanguageResourceHandler.getStringInLocale(
+                  R.string.add_profile_error_name_only_letters
+                )
+              else -> result.error.localizedMessage
```
BenHenning: Where does this come from? I'm a bit concerned about a human-readable message being sourced from the domain layer (vs. passing an enum value that can be switched to retrieve the correct translated string to display to the user).

BenHenning (changes requested):
Thanks @adhiamboperes! I took a first pass on everything except 2 of the test files. PTAL.

Separately, two things:
1. The video files seem to be a bit broken. They keep flashing between different screens and it's quite hard to follow. Perhaps they need to be re-recorded?
2. I didn't check OnboardingFragmentTest and AudioLanguageFragmentTest yet, but broadly please make sure to ensure all error case scenarios for all new UI code are covered in tests if you haven't already. It may also be worth double checking the navigation pathways for potential cases where the user can be completely stalled due to a failure preventing continued navigation (& adding corresponding tests to protect against such situations).

app/src/main/java/org/oppia/android/app/onboarding/AudioLanguageFragmentPresenter.kt
adhiamboperes: finishAffinity clears the backstack, which is the desired outcome once onboarding is complete.  I have added a comment explaining this.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivityPresenter.kt
adhiamboperes: Thanks, this was leftover from some changes. It's surprising that ktlint didn't catch this.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
adhiamboperes: Updated

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: Would you suggest setting the default language or showing an error message?

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: Done, and updated `getSupportedLanguages ` to `retrieveSupportedLanguages`

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/options/AudioLanguageSelectionViewModel.kt
adhiamboperes: I have made a best effort attempt at reordering these in various files.

adhiamboperes (commented):
These comments ended up showing as a review instead of single replies, so I'll go along with it 🤨

app/src/main/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdapters.java
adhiamboperes: I've created a new test suite. Seems like the test file presence check does not check .java files.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: I opted to finish the activity/close the app.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: Set to true always.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
adhiamboperes: This is a bit difficult to do, I am making multiple comparisons between the enums, and the only way to do it is via their human readable names. For example, when I am computing the `languagePreselectionLiveData`, I can either return an OppiaLanguage, or an AudioLanguage. It was easiest to return the string value rather than something like a HashMap. My aim is for the presenter to only be aware of one representation of the languages.

app/src/main/java/org/oppia/android/app/options/AudioLanguageSelectionViewModel.kt
adhiamboperes: I have also renamed the helper functions to avoid get/set as much as possible.

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
adhiamboperes: Updated as suggested.

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
adhiamboperes: No, this test specifically tests that the dropdown works. Renamed to `testFragment_languageSelectionChanged_selectionIsUpdated` to keep things clear.

Testing the domain layer change is tied with the navigation to homescreen. Similar to the app language selection, the langauge is set on clicking the next button.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
adhiamboperes: Fixed

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
adhiamboperes: Fixed

testing/src/main/java/org/oppia/android/testing/profile/ProfileTestHelper.kt
adhiamboperes: Updated

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
adhiamboperes: This is from the domain layer. It is a fallback; with this flow the probably error scenario is the `ProfileNameOnlyLettersException`.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
adhiamboperes: I don't understand. I have tests like this for the api that only updates one field, and a seperate api+corresponding tests that update multiple tests

adhiamboperes:
@BenHenning, PTAL.

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
```diff
@@ -104,13 +104,14 @@ class ProfileTestHelperTest {
...
   fun testAddDefaultProfile_createDefaultProfile_checkProfileIsAdded() {
-    profileTestHelper.createDefaultProfile()
+    profileTestHelper.createDefaultAdminProfile()
     testCoroutineDispatchers.runCurrent()
     val profilesProvider = profileManagementController.getProfiles()
     testCoroutineDispatchers.runCurrent()
 
     val profiles = monitorFactory.waitForNextSuccessfulResult(profilesProvider)
     assertThat(profiles).hasSize(1)
+    assertThat(profiles.first().isAdmin).isEqualTo(true)
```
BenHenning:
```suggestion
    assertThat(profiles.first().isAdmin).isTrue()
```

Small nit.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
```diff
@@ -0,0 +1,295 @@
...
+  @Test
+  fun testBindingAdapters_setSelection_filterDisabled_doesNotAllowKeyboardInput() {
+    launchActivity().use { scenario ->
+      scenario?.onActivity { activity ->
+        val testView: AutoCompleteTextView = activity.findViewById(R.id.test_autocomplete_view)
+        TextInputLayoutBindingAdapters.setSelection(testView, "English", false)
+        onView(withId(R.id.test_text_input_view)).perform(click())
+        testCoroutineDispatchers.runCurrent()
+        onView(withId(R.id.test_autocomplete_view)).perform(KeyboardShownAction())
+        assertThat(KeyboardShownAction().isKeyboardShown).isFalse()
```
BenHenning: Have you double checked that this test actually fails if the filter is enabled? Just wondering if the default value of ``isKeyboardShown`` is false for this object or if it's actually performing the check that we need it to.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -1278,6 +1279,126 @@ class ProfileManagementControllerTest {
...
+      -11235672,
+      "John",
+      isAdmin = true
+    )
+    monitorFactory.waitForNextSuccessfulResult(updateProvider)
+
+    val profileProvider = profileManagementController.getProfile(PROFILE_ID_0)
+    val profile = monitorFactory.waitForNextSuccessfulResult(profileProvider)
+
+    assertThat(profile.avatar.avatarImageUri).isEqualTo("")
```
```suggestion
    assertThat(profile.avatar.avatarImageUri).isEmpty()
```

Small suggestion. Ditto elsewhere in this test file (& PR if relevant).

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
```diff
@@ -1278,6 +1279,126 @@ class ProfileManagementControllerTest {
...
+  @Test
+  fun testUpdateProfile_updateProfileType_existingNonAdminProfile_checkUpdateSucceeded() {
+    setUpTestApplicationComponent()
+    addNonAdminProfileAndWait(name = "Rajat", pin = "01234")
+
+    val updateProvider = profileManagementController.updateProfileType(
+      PROFILE_ID_0,
+      ProfileType.ADDITIONAL_LEARNER
+    )
+    monitorFactory.waitForNextSuccessfulResult(updateProvider)
```
BenHenning, line 1585: Should use ``ensureNextResultIsSuccess`` for asserting rather than 'wait' since it's more intentional.

Or, did you mean to actually verify the changed state? Probably relates to my comment in the other conversation thread.

Ditto for all other new tests in this file that aren't checking state.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: Ah I see. So the outcome part of the tests is confusing: it should probably be clear on what is 'successful'. It seems that some tests are verifying that the operation doesn't fail (such as this one), and others are verifying that the actual profile was changed on disk (``testUpdateProfile_updateMultipleFields_checkUpdateIsSuccessful``). I suggest renaming these tests such that their expected outcomes are clearer on what, exactly, should happen rather than using 'succeeded/successful' since those terms are ambiguous here.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
BenHenning: But can it still actually show up? Under what circumstances?

We use ``ProfileNameOnlyLettersException`` specifically so that we can retrieve a localized (and hopefully actionable) error message for the user. Why can't we introduce another or multiple exception types for other possibilities, as well?

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
BenHenning: I think something has to pay the cost of multiple representations, the question is simply what ultimately does that. I think there's one major advantage to performing the conversion at the model layer: it ensures everything up to that layer only ever handles a single representation (the proto version). The alternative is having a data pipeline that switches midway from one representation to another, and this can make things awkward to track (or to perform later operations, hence the need here to map back from the human-readable version).

If there aren't many places that need to display the language, you could also introduce a custom view or binding adapter that allows you to bind the enum value and perform the switching to string there. I think that avoids needing to switch representations anywhere in the data pipeline, it keeps the view model "pure" (in that it only needs to track the proto representation), and you shouldn't need a hashmap anywhere (since you can just perform a ``when`` during binding time to retrieve the corresponding language name).

Caveat: it probably needs to be a custom view since I think you'll need access to injectable resources (this class) to compute the localized display name.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
BenHenning: Hmm there's not a perfect solution, but perhaps we should crash rather than just close the app (since then it at least gets special treatment from analytics systems and hopefully gets escalated to us more rapidly than filtering among non-fatal error logs).

I think you can just wrap ``result.error`` in a new exception with context and then throw it here. Definitely a good scenario to add a test for if that's possible.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivityPresenter.kt
BenHenning: It unfortunately doesn't catch certain types of imports. Might be something they've fixed (since we're on a very old version of ktlint), but I'd rather move over to ktfmt, anyway. :)

scripts/assets/file_content_validation_checks.textproto
BenHenning: Going ahead and resolving this since it seems addressed.

BenHenning (changes requested):
Thanks @adhiamboperes! Overall the PR is looking quite good, there are just a few fringe aspects that should probably be finalized before this is merged.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
adhiamboperes: Done, could only find instances in this file for this PR.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
adhiamboperes: I have renamed the updateProfileType tests, and also added assertions for the corresponding fields in the tests for the multiple fields api.

Also, I have added a check that results into a failure if the profileType is not specified to ensure the developer using the api does not fail to pass the ProfileType.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
adhiamboperes: I have changed the test assertions a bit, so that I retrieve the profile and assert the value of its profileType field instead. I thinnk that is explicit as well.

testing/src/test/java/org/oppia/android/testing/profile/ProfileTestHelperTest.kt
adhiamboperes: Fixed

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
adhiamboperes: I have created a binding adapter that binds OppiaLanguage and does the conversion there. We do however still need to perform the string to proto conversion only after selection from dropdown(since the dropdown items are an array of strings). The view model only handles protos now,  and I also switched over the audio language functionality to the TranslationController. This required me to make changes to the AudioFragment, AudioController and finally AudioLanguageFragment.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: Modified flow to allow progres regardless of saving outcome.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
adhiamboperes: Reverted the TranslationController, AudioFragment, AudioController changes since they have been addressed in #5487.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
adhiamboperes: I checked and found out from manual testing that the keyboard is never shown either way, since I disabled raw input in the code. I also found out that the filter does not affect they input. Rather, the filter being disabled ensures that once an option is seleted, the full list is still displayed on subsequent expansions. Conversely, if filtering is enabled, then the list items will be reduced to only the selected item. This behaviour turned out to be incredibly difficult to test and I tried(and failed) to set up the tests to reproduce the behaviour as follows:
1. Using the test fragment, I bound a list to the autocompletetextview via an adapter.
2. In the test, I attempted to click on the dropdown (this step fails with an error implying that an adapter class was not found.  However, this setup works in the AudioLangaugeFragmentTest, where I test the dropdown).
3. I then intend to click on an item from the list to select it
4. Click again on the dropdown menu. The list should be asserted as filtered or not.

I am however unable to verify step 3 and 4 as the test always fails on step 2.

In general, the matcher is nolonger relevant and has been moved. As are the tests as I figured out that from the binding adapter perspective, the filter is not being used in the api under test, but is passed to Android's ListAdapter.

adhiamboperes:
@BenHenning, PTAL.

adhiamboperes:
The coverage comment on this PR looks strange.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
adhiamboperes: I created a new exception, `UnknownProfileTypeException, ` which I determined to be the likely scenario when ProfileType was not passed. But because the when branch needs to be exhaustive, I also added a default message `An error occurred while creating a profile.` for any unknown error types.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
```diff
@@ -154,73 +146,12 @@ class TextInputLayoutBindingAdaptersTest {
     launchActivity().use { scenario ->
       scenario?.onActivity { activity ->
         val testView: AutoCompleteTextView = activity.findViewById(R.id.test_autocomplete_view)
-        TextInputLayoutBindingAdapters.setSelection(testView, "English", true)
+        TextInputLayoutBindingAdapters.setLanguageSelection(testView, OppiaLanguage.ENGLISH, true)
```
BenHenning, line 149: It might be nice to also test a non-English language (just 1 test, no need to repeat filter verification) just to ensure that the implementation is actually generalized for the language parameter.

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
```diff
@@ -1059,6 +1083,12 @@ class ProfileManagementController @Inject constructor(
             "Profile cannot be an admin"
           )
         )
+      ProfileActionStatus.PROFILE_TYPE_UNKNOWN ->
+        AsyncResult.Failure(
+          UnknownProfileTypeException(
+            "ProfileType must be set"
+          )
+        )
```
BenHenning:
```suggestion
        AsyncResult.Failure(UnknownProfileTypeException("ProfileType must be set."))
```

Small reformatting suggestions.

domain/src/test/java/org/oppia/android/domain/profile/ProfileManagementControllerTest.kt
BenHenning: I think the changes are much clearer--thanks!

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -589,6 +594,34 @@ class CreateProfileFragmentTest {
     }
   }
 
+  @Test
+  fun testFragment_profileTypeArgumentMissing_showsUnknownProfileTypeError() {
+    val intent = CreateProfileActivity.createProfileActivityIntent(context)
+    intent.apply {
```
BenHenning: Can we do this without ``apply``? Ditto below. I think it hurts the code readability as it's not obvious why ``apply`` is being used here.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingAppLanguageViewModel.kt
```diff
@@ -0,0 +1,28 @@
...
+import javax.inject.Inject
+
+/** ViewModel for managing language selection in [OnboardingFragment]. */
+class OnboardingAppLanguageViewModel @Inject constructor() : ObservableViewModel() {
+  /** The selected app language displayed in the language dropdown. */
+  val languageSelectionLiveData: LiveData<OppiaLanguage> get() = _languageSelectionLiveData
+  private val _languageSelectionLiveData = MutableLiveData<OppiaLanguage>()
+
+  /** Get the list of app supported languages to be displayed in the language dropdown. */
+  val supportedAppLanguagesList: LiveData<List<String>> get() = _supportedAppLanguagesList
```
BenHenning: Why isn't this a list of ``OppiaLanguage``s? I saw your other comment on this (the thread is now closed since it was focused on the other property in this view model)--is it possible to keep this as a list of ``OppiaLanguage``s for consistency and perform the conversion at binding time like we do for the language selection?

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
BenHenning: Thanks--the latest changes look much cleaner I think.

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
```diff
@@ -186,11 +186,33 @@ class AppLanguageResourceHandler @Inject constructor(
...
```
BenHenning: Can this be removed now?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
BenHenning: Ah okay--thanks for digging into it and providing the context. I think it's fine if we skip covering these very specific tests.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
BenHenning: I think this is much clearer--thanks!

BenHenning (commented):
Thanks @adhiamboperes! PTAL. Just had a few follow-ups.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
adhiamboperes: Done

domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt
adhiamboperes: Fixed

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
adhiamboperes: Done here, although below was a simplification of adding multiple params to the intent(modified now).

app/src/main/java/org/oppia/android/app/translation/AppLanguageResourceHandler.kt
adhiamboperes: Removed now

app/src/main/java/org/oppia/android/app/onboarding/OnboardingAppLanguageViewModel.kt
adhiamboperes: I initially didn't update this viewmodel(the other moldifications were in the AudioLanguageViewModel) since I didn't need a binding adapter here(only one source of truth). But per this comment I have updated to OppiaLanguage for consistency.

adhiamboperes:
@BenHenning, PTAL.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
```diff
@@ -36,13 +80,198 @@ class OnboardingFragmentPresenter @Inject constructor(
...
+      )
+
+      onboardingLanguageDropdown.apply {
+        setRawInputType(EditorInfo.TYPE_NULL)
+
+        onItemClickListener =
+          AdapterView.OnItemClickListener { _, _, position, _ ->
+            adapter.getItem(position).let { selectedItem ->
+              selectedItem?.let {
+                val localizedNameMap = OppiaLanguage.values().associateBy { oppiaLanguage ->
```
BenHenning: I'm not sure that this is the correct way to handle this. The enum includes an unspecified constant, plus it may contain languages not yet ready for full release in the app (which is why we have language configurations).

Might ``getSupportedAppLanguages`` be a better fit here? It should already do the necessary work for filtering valid languages:

https://github.com/oppia/oppia-android/blob/4730edb0263135e88dab08ab0f6dc724a1585a67/domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt#L114

I'm not sure if the returned list's contents and order are stable--you'll need to double check that.

BenHenning (changes requested):
Thanks @adhiamboperes! The PR LGTM except one comment on correctness for sourcing the list of valid languages. PTAL.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: When passing the language list to the adapter, I have correctly passed the list retrieved from `translationController. getSupportedAppLanguages()`.

The list needed to be converted to a a string array, in order to display the localized names:

| Binding OppiaLanguage |Binding the computeLocalizedDisplayName(language) |
|---|---|
|Screenshot_1726066627|Screenshot_1726068086|

So here, I compare the selected item with the enum.

BenHenning: How does the mapping work, exactly? This is iterating through the enum, but if some of the languages in the enum are excluded from the list returned by ``getSupportedAppLanguages`` won't the selected index be potentially wrong (and thus set the wrong language)?

In fact, I think this would be broken on production builds (e.g. ``//:oppia_beta``). ``oppia_dev`` uses the 'all supported languages' config file, but the production one uses this:

https://github.com/oppia/oppia-android/blob/dfb9a301280b9a46526cb2f5ca6329532fec6bf0/config/src/java/org/oppia/android/config/productionlanguages/supported_languages.textproto#L1-L105

When compared with the language proto:

https://github.com/oppia/oppia-android/blob/dfb9a301280b9a46526cb2f5ca6329532fec6bf0/model/src/main/proto/languages.proto#L9-L69

Hindi isn't supported, so presumably there would be a selection issue on production builds.

Or, am I misunderstanding still?

BenHenning (commented):
Thanks @adhiamboperes. Had one follow-up on the existing conversation--PTAL.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
adhiamboperes: You're right, since I'm relying on the position from the dropdown to fetch the selected language, the index of the dropdown item could no longer match the index of the language in the enum. This mismatch could cause the selected language to be wrong. So I have refactored the code to rely on the supported app language list, and not the OppiaLanguage enum. I have also tested with the beta build.

[device-2024-10-01-144818.webm](https://github.com/user-attachments/assets/da0fe694-2e74-4c1e-ad86-6f96aa9022f6)

adhiamboperes:
@BenHenning, I have addressed your comment, PTAL.

app/src/main/java/org/oppia/android/app/onboarding/OnboardingFragmentPresenter.kt
BenHenning: Ah--thanks for digging into this more! Latest solution LGTM.

BenHenning (approved):
Sorry for the delayed review--thanks @adhiamboperes! This LGTM!

PR #5459, opened 2024-07-06
Fix part of #5343: Enable Code Coverage Analysis for many to one test target files

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 2.1 of Project 4.1]

### Changes Made
- Enabling code coverage analysis for many-to-one relationship test target files by aggregating the coverage reports of each related test target into a single report using the `calculateAggregateCoverageReport()` function.
- Using the calculated aggregated coverage report to generate Markdown and HTML files.
- Properly re-implemented the asynchronous flow as previous implementation though has async calls was still handling it sequentially as targets are provided one by one and made to wait until the coverage report is generated and the previous one is returned back.

**Working test asynchronous flow screenshot**
With static test targets provided: https://github.com/oppia/oppia-android/pull/5459/commits/2ac7b041dcab8611bea77602b2f22e47ed97d43d

image

**Low impact changes:**

Not related to the current PR implementation but added few fixes along with it,
- Restyled the HTML report slightly since the previous version appeared somewhat vague and empty 😶‍🌫️.
- Renamed sample test name from `TwoSum` to `AddNums`, felt more appropriate and clear.

### Note
I couldn't find any actual working app module file with both shared and local tests to fully test this. Instead, I mocked the behavior with a sample coverage report and tested it with test cases, assuming that both tests for the same source file would have identical **line_number**, **lines_hit**, and **lines_found** data.

___

### Screenshot of current HTML report
Screenshot (1460)

**Responsive layouts**
zero contributions (48)

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

[...]

PR #5461, opened 2024-07-10
Fix part of #5343: Enable Coverage Generation for a list of files

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 2.2 of Project 4.1]

### Changes Made
- **File Path List:** The implementation now enables doing coverage analysis for a list of provided files (which used to be just a single file). Coverage analysis runs sequentially for each file.

**Updated command:**
```
bazel run //scripts:run_coverage -- $(pwd) <path/to/file1.kt> <path/to/file2.kt> <path/to/file3.kt> --format=MARKDOWN --processTimeout=15
```

**Handled Edge cases:**
- **Non-kotlin Files:**
  - Since when retrieving changed files list other non-kotlin files (resource files - .xml, .txt, .md, .pb, .json, etc.) could be included, this script will handle the analysis by only filtering in the required kotlin files.
  - The kotlin files which are test files are mapped to their appropriate source files and then provided for coverage analysis.
  - So the provided list of files: [utility/.../MathTokenizer.kt, app/.../layout.xml, scripts/.../TestFileCheckTest.kt] will only do coverage analysis for the files "utility/.../MathTokenizer.kt", "scripts/.../TestFileCheck.kt".

- **Input variations:**
  - Developers would be able to provide list of files in all below variations
  - [file1.kt, file2.kt, file3.kt]
  - ["file1.kt", "file2.kt", "file3.kt"]
  - file1.kt file2.kt file3.kt
  - file1.kt, file2.kt, file3.kt

- **Input format:**
  - Included the option to use the short form 'md' for the markdown format (just to enhance user experience).
```
bazel run //scripts:run_coverage -- $(pwd) <path/to/file1.kt> <path/to/file2.kt> --format=md
```

- **Coverage Results Display:** Based on the discussions, it was decided to display only the failure cases (i.e., below the minimum threshold). However, I included success cases in a hidden dropdown. This allows developers to see the coverage checks and percentages of other files, enabling them to improve their test scenarios if needed. (yet to discussed if that's ok)
- **Error Handling:** Error statements in the coverage flow have been replaced with the introduction to Failure and Exception files having their own types to identify them.
**New Proto Structure:**
```
message CoverageReport {
  oneof report_status {
    // Coverage report details when coverage data retrieval succeeds.
    CoverageDetails details = 1;
    // Coverage report failure when coverage data retrieval fails.
    CoverageFailure failure = 2;
    // Coverage exemption when the source file is exempted from having a test file.
    CoverageExemption exemption = 3;
  }
}
```
- **Handling Anomaly Cases:** There are scenarios where files could be exempted, coverage cannot be retrieved, or coverage does not exist. Previously, an error in one file could halt the entire process. Instead of throwing an error, the flow continues while displaying the encountered cases.
- **PASS / FAIL:** While still the coverage analysis continues with failure / anomaly cases with any case other than a PASS would be considered FAIL and checked at the end to throw an error at the end of the entire execution completion, helping with both local dev and ci checks.
- **Format:** Defaulting the report generation format to **HTML** considering priority format for local development. While with both .md and .html reports a text report will be logged to the console, so a quick glance is provided and that will make .md obsolete for local development.

(Updated) The Logged Report is now updated to only log reports that Fail (either a hasFailure case or details that fail below threshold or if exempted under overridden threshold)

Sample log while having min threshold as 99% and exempted percentage for MathModel.kt as 101% (for testing purposes)
image

**Logged Report Text:**

image

- **Final Markdown Report:** The final Markdown report is generated in the CoverageReporter.kt file, with the entire list of coverage data proto collected from each file's coverage analysis. The report template is inspired from [Oppia Web's Codecov report](https://github.com/oppia/oppia/pull/9618#issuecomment-648962538) and includes a dropdown for better readability (provided below). Though the original template was to have it as a list of dropdowns (the new template is yet to be confirmed).

- The updated template also has the exempted percentage included to make it clear and not cause any if any exempted cases are being included in the analysis.

### Results
Number of files assessed: 5
Overall Coverage: **94.26%**
Coverage Analysis: **FAIL** :x:
##

### Failure Cases

| File | Failure Reason |
|------|----------------|
| File.kt | No appropriate test file found for File.kt. |

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| <details><summary>RunCoverage.kt</summary>scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt</details> | 51.38% | 148 / 288 | :x: | 70% |
| <details><summary>MathModel.kt</summary>utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt</details> | 77.78% | 7 / 9 | :x: | 80% _*_ |
| <details><summary>MultipleChoiceInputModule.kt</summary>domain/src/main/java/org/oppia/android/domain/classify/rules/multiplechoiceinput/MultipleChoiceInputModule.kt</details> | 10.00% | 1 / 10 | :x: | 30% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds

### Passing coverage

<details>
<summary>Files with passing code coverage</summary><br>

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| <details><summary>MathTokenizer.kt</summary>utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt</details> | 94.26% | 197 / 209 | :white_check_mark: | 70% |
</details>

### Exempted coverage
<details><summary>Files exempted from coverage</summary> <br>
<details><summary>ActivityComponent.kt</summary>app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt</details>
<details><summary>TestExemptedFile.kt</summary>app/src/main/java/org/oppia/android/app/activity/TestExemptedFile.kt</details>
<details><summary>SourceIncompatible.kt</summary>app/src/main/java/org/oppia/android/app/activity/SourceIncompatible.kt</details>
</details>

---

### To be discussed:
- The `shards_count` for RunCoverageTest.
- The `MIN_THRESHOLD` that need to be set.
- The final report having Success cases table included.
- On how to handle the anomaly cases (should they be considered FAIL case for coverage analysis)
- To discuss cases when the coverage data are unavailable for source targets (files) (ie SF: doesn't point to source files)

### Todo
- **[Done]** Use dynamic filename list taken from the arg (required for both ci and developer workflow)
- **[Done]** Figure out a way on how to handle the delivery of the final md report and check status generated
  - The final md report will be used as an uploadable comment (considering saving it to $(pwd)/finalreport.md (to access the same file in ci) 
  - The check status will be used to set the ci run fail/pass case
- **[Done]** Update tests
- **[Done]** add min threshold for exempted files in the table to let developers know them.
- **[Done]** Add links for anomaly case file paths too.
- **[Done]** Remove Asynchronous flow and have it Sequential
- **[Done]** The test files too need to be taken into account while considering coverage analysis (mapping test files to their source files and provide it to the script run, also these files need to be figured out even in the compute changed files utility as that will help with building the targets)

With this doing a script run of 
```
bazel run //scripts:run_coverage -- $(pwd) 
utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt 
scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageRunnerTest.kt 
data/src/test/java/org/oppia/android/data/backends/gae/JsonPrefixNetworkInterceptorTest.kt 
app/src/sharedTest/java/org/oppia/android/app/story/StoryFragmentTest.kt 
app/src/test/java/org/oppia/android/app/application/alpha/AlphaBuildFlavorModuleTest.kt test.xml 
app/src/test/java/org/oppia/android/app/home/HomeActivityLocalTest.kt
```
will do coverage analysis on files:
```
Running coverage analysis for the files: [
utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt, 
scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt, 
data/src/main/java/org/oppia/android/data/backends/gae/JsonPrefixNetworkInterceptor.kt, 
app/src/main/java/org/oppia/android/app/story/StoryFragment.kt, 
app/src/main/java/org/oppia/android/app/application/alpha/AlphaBuildFlavorModule.kt, 
app/src/main/java/org/oppia/android/app/home/HomeActivity.kt
]
```

-  **[Done]** The Error statements need to have a clear indication of why it failed.
-  **[Done]** Every fail case is a Coverage Status "FAIL' only a pass is 'PASS'
-  **[Done]** Re-work on the final md generation [use a centralized proto collection way, move the existing md generation script and have it with the CoverageReporter just for md, have the exisiting html generation for individual files]
- **[Done]** Add tests with addition to including test file's source files.

## Essential Checklist

<details><summary>Old Template</summary>

**FOR TESTING PURPOSE**
The below provided data was tested while having a min percentage of 99% and having the test_file_exemption as
```
test_file_exemption {
  exempted_file_path: "utility/src/main/java/org/oppia/android/util/logging/ConsoleLogger.kt"
  override_min_coverage_percent_required: 20
}
test_file_exemption {
  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImpl.kt"
  override_min_coverage_percent_required: 70
}
```

## Coverage Report

- Number of files assessed: 5
- Coverage Analysis: **FAIL** :x:

| File | Coverage | Lines Hit | Status | Min Required |
|-----|:----------:|---------:|:-------:|:--------------:|
| [MathTokenizer.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt) | 94% | 197 / 209 | :x: | 99% |
|Exempted :small_red_triangle_down:|
| [FirebaseAuthWrapperImpl.kt](https://github.com/oppia/oppia-android/tree/develop/domain/src/main/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImpl.kt) | 31% | 5 / 16 | :x: | 70% |

<details>
<summary>Succeeded Coverages</summary><br>

| File | Coverage | Lines Hit | Status | Min Required |
|-----|:----------:|---------:|:-------:|:--------------:|
| [MathModel.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt) | 100% | 19 / 19 | :white_check_mark: | 99% |
|Exempted :small_red_triangle_down:|
| [ConsoleLogger.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/logging/ConsoleLogger.kt) | 54% | 30 / 55 | :white_check_mark: | 20% |
</details>

### Test File Exempted Cases
- [ActivityComponent.kt](https://github.com/oppia/oppia-android/tree/develop/app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt)

[Todo] Add report with failure case and coverage check status.

<details>
<summary>Old template</summary>

---

## Coverage Report

- Total covered files: 5
- Coverage Status: **FAIL**
### Failed Coverages
Min Coverage Required: 40% _// the set value is 10 (MIN_THRESHOLD), Using 40 for just testing._
| Covered File | Percentage | Line Coverage | Status |
|--------------|------------|---------------|--------|
|[FirebaseAuthWrapperImpl.kt](https://github.com/oppia/oppia-android/tree/develop/domain/src/main/java/org/oppia/android/domain/auth/FirebaseAuthWrapperImpl.kt)|31.25%|5 / 16|:x:|

<details>
<summary>Succeeded Coverages</summary><br>

| Covered File | Percentage | Line Coverage | Status |
|--------------|------------|---------------|--------|
|[NumericExpressionEvaluator.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/math/NumericExpressionEvaluator.kt)|86.36%|19 / 22|:white_check_mark:|
|[MathTokenizer.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt)|94.26%|197 / 209|:white_check_mark:|
|[RealExtensions.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/math/RealExtensions.kt)|89.73%|201 / 224|:white_check_mark:|
</details>

### Anomaly Cases
- The file: [ActivityComponent.kt](https://github.com/oppia/oppia-android/tree/develop/app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt) is exempted from having a test file; skipping coverage check.

---

</details>
</details>


Rd4dev:
@BenHenning, could you PTAL at the coverage analysis implementation for a list of files? 
Does it aligns with the expectations outlined in the comments: https://github.com/oppia/oppia-android/pull/5443#pullrequestreview-2138039817, https://github.com/oppia/oppia-android/pull/5443#discussion_r1652385633.

Rd4dev:
@BenHenning, unassigning for now as I still need to work on this PR to handle errors, manage sequential flow, and collect the protos for final markdown generation. I’ll make the necessary modifications and request a re-review once done.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -3,56 +3,101 @@ package org.oppia.android.scripts.coverage
...
 
+/** ANSI escape codes for colors. */
+/** Green text. */
+const val GREEN = "\u001B[32m"
+/** Red text. */
+const val RED = "\u001B[31m"
+/** Default text. */
+const val RESET = "\u001B[0m"
+/** Bold text. */
+const val BOLD = "\u001B[1m"
```
Rd4dev, line 26: @BenHenning, the color codes are being added to enhance user experience and provide visual cues. 

**FAILED**
image

**PASSED**
image

Their inclusion is subject to further discussion. If approved, they will be moved to a more appropriate location; if not, they will be removed.

Rd4dev:
@BenHenning, I've restructured the coverage data protos to be saved and passed as a single container for report generation and updated the description to cover most changes made in this PR. Could you PTAL?

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -10,18 +10,13 @@ import org.oppia.android.scripts.common.CommandExecutorImpl
 import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
 import org.oppia.android.scripts.testing.TestBazelWorkspace
 import org.oppia.android.testing.assertThrows
-import java.io.ByteArrayOutputStream
 import java.io.File
-import java.io.PrintStream
 import java.util.concurrent.TimeUnit
 
 /** Tests for [RunCoverage]. */
 class RunCoverageTest {
```
Rd4dev, line 22: ### Todo:
to add test for
**[Done]** saving proto
**[Done]** exemptions check
**[Done]** multiple hits on a same line

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -56,13 +61,18 @@ class CoverageRunner(
     val sfStartIdx = coverageData.indexOfFirst {
       it.startsWith("SF:") && it.substringAfter("SF:").substringAfterLast("/") == extractedFileName
     }
-    require(sfStartIdx != -1) {
-      "Coverage data not found for the file: $extractedFileName"
-    }
+    if (sfStartIdx == -1) return generateFailedCoverageReport(
```
BenHenning: Here & below: please add braces for the statement.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -5,9 +5,26 @@ package proto;
 option java_package = "org.oppia.android.scripts.proto";
 option java_multiple_files = true;
 
-// Coverage Report that contains the bazel coverage data retrieved from the
-// Bazel coverage execution.
+// Represents the list of Coverage Report of all analysed files.
```
BenHenning:
```suggestion
// Represents the complete execution of a series of code coverage runs.
```

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
```diff
@@ -5,9 +5,26 @@ package proto;
...
+message CoverageReportContainer {
+  // Stores the list of all Coverage Reports.
```
BenHenning:
```suggestion
  // The list of all CoverageReports detailing the coverage results.
```

scripts/src/javatests/org/oppia/android/scripts/coverage/BUILD.bazel
```diff
@@ -8,9 +8,10 @@ kt_jvm_test(
     name = "RunCoverageTest",
     size = "large",
     srcs = ["RunCoverageTest.kt"],
-    shard_count = 4,
+#    shard_count = 24,
```
BenHenning: Temporary change--please revert.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -3,56 +3,101 @@ package org.oppia.android.scripts.coverage
...
  * Example with custom process timeout:
- *     bazel run //scripts:run_coverage -- $(pwd)
- *     utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt processTimeout=15
+ *    bazel run //scripts:run_coverage -- $(pwd)
+ *    utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt --processTimeout=15
  *
+ * Example with output path to save the collected coverage proto:
+ *    bazel run //scripts:run_coverage -- $(pwd)
+ *    utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
+ *    --protoOutputPath=/tmp/coverage_report.proto64
```
BenHenning: Perhaps something that might be cleaner is just to introduce a proto64 format (as a choice other than HTML/MD)? That would also simplify the script a bit since you shouldn't need to bother with MD/HTML output if the only thing that's cared about is the proto.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -124,60 +200,68 @@ class RunCoverage(
...
-        parentFile?.mkdirs()
-        writeText(reportText)
+  private fun combineCoverageReports(coverageResultList: List<CoverageReport>):
+    CoverageReportContainer {
+      val containerBuilder = CoverageReportContainer.newBuilder()
+      coverageResultList.forEach { report ->
+        containerBuilder.addCoverageReport(report)
       }
+      return containerBuilder.build()
+    }
```
BenHenning:
```suggestion
  private fun combineCoverageReports(
    coverageResultList: List<CoverageReport>
  ): CoverageReportContainer {
    val containerBuilder = CoverageReportContainer.newBuilder()
    coverageResultList.forEach { report ->
      containerBuilder.addCoverageReport(report)
    }
    return containerBuilder.build()
  }
```

The correct format wrapping for this situation.

Rd4dev:
@BenHenning also adding a reference to the requested test case with #5459 - [testRunCoverage_withMultipleTestsHittingSameLine_calculatesCoverageReportCorrectly](https://github.com/oppia/oppia-android/pull/5461/files#diff-5a779ea4a8aa2caf5249ced27465f76b8ac19bed5cf250475f6033dfad464908R1598)

The line Number 7 was hit from both the test targets on the same source file, representing the total coverage as **FULL**

---

Apart from the future review comments, addition of test cases on the exemptions for overriding the coverage percentage need to be added. (working on them)

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: I'm totally fine with including TTY ANSI control codes here for colorizing--it's a nice touch. I don't have a strong opinion on centralizing the color locations, though I have a PR that I hope to revive one day that introduces a console class for all scripts which includes colorizing support (I suspect that would be the eventual place that these escape codes would live in the long-term).

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -168,173 +193,1169 @@ class RunCoverageTest {
...
+
+    RunCoverage(
+      "${tempFolder.root}",
+      filePathList,
+      ReportFormat.MARKDOWN,
+      longCommandExecutor,
+      scriptBgDispatcher
+    ).execute()
+
+    for (file in filePathList) {
```
BenHenning: Here & below (I think in one other test): prefer directly asserting on exact state rather than iterating the list. Control flow in tests is difficult to verify.

I think you'd probably benefit from leveraging some helper logic here, too--a lot of the tests are doing redundant checks (we generally want to keep tests as specific as possible to verifying whatever should fail if their conditions are broken and rely on other tests to verify related behavior rather than checking the same things many times, though this isn't always feasible).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -2,27 +2,34 @@ package org.oppia.android.scripts.coverage
 
 import org.oppia.android.scripts.proto.Coverage
 import org.oppia.android.scripts.proto.CoverageReport
+import org.oppia.android.scripts.proto.CoverageReportContainer
+import org.oppia.android.scripts.proto.TestFileExemptions
 import java.io.File
 
+/** Minimum coverage percentage required. */
+const val MIN_THRESHOLD = 10 // to be decided
```
BenHenning: Reminder to finalize this.

scripts/src/javatests/org/oppia/android/scripts/coverage/BUILD.bazel
Rd4dev: ### Will be reverted!
`shard_count = 24` will be set

(Had it for local dev purposes and accidentally checked in with the upcoming exempted tests commit this will be reverted)

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -31,200 +38,535 @@ class CoverageReporter(
...
+        val exemption = testFileExemptionList[filePath]
+        if (coveragePercentage < MIN_THRESHOLD) {
+          if (exemption != null) {
+            val ovveriddenMinCoverage = exemption.overrideMinCoveragePercentRequired
+            if (coveragePercentage < ovveriddenMinCoverage) {
+              return CoverageCheck.FAIL
+            }
+          } else {
+            return CoverageCheck.FAIL
+          }
```
BenHenning: Can probably be simplified by doing something like:

```kotlin
val threshold = testFileExemptionList[filePath]?.overrideMinCoveragePercentRequired ?: MIN_THRESHOLD
if (coveragePercentage < threshold) return CoverageCheck.FAIL
```

Also: typo on ovveridden.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -31,200 +38,535 @@ class CoverageReporter(
...
+  val fileWithoutExtension = filePath.substringBeforeLast(".")
+  val defaultFilename = when (reportFormat) {
+    ReportFormat.HTML -> "coverage.html"
+    ReportFormat.MARKDOWN -> "coverage.md"
+  }
+  return "$repoRoot/coverage_reports/$fileWithoutExtension/$defaultFilename"
+}
+
+private fun getFilenameAsLink(filePath: String): String {
+  val oppiaDevelopGitHubLink = "https://github.com/oppia/oppia-android/tree/develop"
```
BenHenning: This is a nice idea, but it's also a bit tricky because the actual file under test will be specific to the PR being run or the current branch. I'm torn here--we either include a link that is most often wrong, or we exclude it entirely. What do you think?

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
```diff
@@ -1,78 +1,291 @@
...
+    assertThat(outputReportText).isEqualTo(expectedMarkdown)
   }
 
   @Test
-  fun testCoverageReporter_noLinesFound_generatesZeroCoverageRatio() {
-    val expectedZeroCoverageRatio = 0F
-    // to check divided by zero error doesn't occur
-    val report = validCoverageReport.toBuilder().setLinesFound(0).build()
-    reporter = CoverageReporter(
+  fun testGenerateMarkDownReport_withFailureCoverageReportDetails_generateMarkdownTable() {
```
BenHenning:
```suggestion
  fun testGenerateMarkDownReport_withFailureCoverageReportDetails_generatesMarkdownTable() {
```

Note the verb tense for the "expected outcome" portion of the test name. This is the verb tense we should be using. Ditto for all other tests in all other test files in this PR.

BenHenning (changes requested):
Thanks @Rd4dev! I think the final markdown report looks really nice!

Some thoughts:
- Is the "exempted arrow" row of the tables meant to demarcate the tests which have minimum coverage overrides? If so, perhaps cut that and add something like a '*' to the required percentages of tests with overrides, then add text below the tables that says something like `'*' represents tests with custom overridden pass/fail coverage thresholds`. Any thoughts?
- If this isn't what the row means, what does it mean?
- I suggest using 'Files with passing code coverage' rather than 'Succeeded Coverages'.
- Remove the bullet list for the number of files tested and the overall pass/fail--just have these as distinct lines under a new sub-section titled "Results".
- Add subsection headers for "Failing coverage" and "Passing coverage" for the other portions of the report.
- To the new "results" section mentioned above, please also add a line that lists the overall code coverage of all tests run (i.e. the average among all covered reports; exempted are ignored).
- Rename 'Test File Exempted Cases' to 'Files Exempted from Coverage'.

Please make sure to update the PR description accordingly, and update this PR post #5459 being merged.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks for confirming. I’ll keep the color codes here until a centralized place is decided in the future.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Done. Added them.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Updated as suggested.

scripts/src/java/org/oppia/android/scripts/proto/coverage.proto
Rd4dev: Updated as suggested.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Ah, yeah! Thanks, Ben. I've introduced the **PROTO** report format, with that, it skips the report generation executions.

Rd4dev: Added / modified 2 test cases to test this behaviour,
- testRunCoverage_withProtoReportFormat_savesCoverageContainerProto
- testRunCoverage_withProtoReportFormat_skipsCoverageGeneration

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Should have realized when they appeared a bit unusual. Thanks for the suggestion. Updated it.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Ah! Thanks. Updated the same.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
Rd4dev: Ah Got it! Updated over here, will look out for any other test cases too.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Got it - removed the iterating logics from the test cases.

Edit: moved the readOutputText for html and md to its own helper functions.

---
Ah makes sense. Few places I can think of utilizing the helper logics would be to get the test and source content since they're being initialized redundantly in many places, already have helper functions for acquiring common html and md reports and will try to add any other wherever possible and I will also try if possible to may be have/split the existing test cases to be more specific like these ones I guess,
- testRunCoverage_withMultipleFiles_generatesOverallCoveragePercentage ->
```
...
assertThat(outputReportText).contains("Overall Coverage: **62.50%**")
```
- testRunCoverage_withProtoReportFormat_skipsCoverageGeneration
```
...
// Since report Format is PROTO no report generation would have been initiated
// resulting in no coverage report directory or files being created.
assertThat(directory.isDirectory).isFalse()
```

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Ahh I did not actually realize that these file links should correspond to their respective PRs diff files. 

Uh... we can either try to find the feature branch name and use the same source path to actually locate the pr file - https://github.com/oppia/oppia-android/blob/code_coverage_list_of_files/scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt

```
[base link]-[feature-branch]-[source-path]
```
But I guess this will only work if the branch is in upstream (which makes this possibility obsolete) 

Edit: If it's from not from upstream if possible we can point to the contributor's origin's feature-branch then to its source path.

eg: https://github.com/Rd4dev/oppia-android/blob/workflow_canceller_concurrency/.github/workflows/unit_tests.yml 

but would we be able to differentiate, and also would it be more complicated?

---

Don't know if these links would be really useful to have it as a mandatory thing but I liked it as links and still do (as it was better to rather have it as short filename links than as a long text of its path), will look if there are any possibility of at least grabbing the pr's diff files. 

Will update if I can find anything.

Edit:
Doing the above doesn't seem to be straightforward too - https://github.com/orgs/community/discussions/55764#discussioncomment-5940162

scripts/src/javatests/org/oppia/android/scripts/coverage/BUILD.bazel
Rd4dev: Done, reverted it.

Rd4dev: Resolving since reverted

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Thanks for providing the min threshold, now its set to 70.

Rd4dev:
@BenHenning, updated the md template as suggested, but still pending on how I can proceed with the File names and links. Thought to confirm and proceed (just to prevent reworking on the tests).

Can you confirm the template whenever possible?

This is the full template, and the Filename is subject to change (few templates are provided right after this).

# MD Report Template
## Coverage Report

### Results
Number of files assessed: 3
Coverage Analysis: **FAIL** :x:
Overall Coverage: 94.73%
##

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| [MathModel.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt) | 100.00% | 19 / 19 | :x: | 101% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds
### Passing coverage

<details>
<summary>Files with passing code coverage</summary><br>

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| [MathTokenizer.kt](https://github.com/oppia/oppia-android/tree/develop/utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt) | 94.26% | 197 / 209 | :white_check_mark: | 1% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds
</details>

### Files Exempted from Coverage
- [ActivityComponent.kt](https://github.com/oppia/oppia-android/tree/develop/app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt)

---

The Filename which was earlier added as links should be modified to just have them as plain text. Can you confirm which of these 2 or any other suggestion should the way?

## Template 1

(Dummy data just for template)

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| MathModel.kt | 0.00% | 4 / 19 | :x: | 80% _*_ |
| MultipleChoiceInputModule.kt | 10.00% | 4 / 19 | :x: | 80% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds

---
## Template 2

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt | 100.00% | 4 / 19 | :x: | 80% _*_ |
| domain/src/main/java/org/oppia/android/domain/classify/rules/multiplechoiceinput/MultipleChoiceInputModule.kt | 10.00% | 4 / 19 | :x: | 80% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds

---
## Template 3

or can we have like this...

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| <details><summary>MathModel.kt</summary>utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt</details> | 100.00% | 19 / 19 | :x: | 101% _*_ |
| <details><summary>MultipleChoiceInputModule.kt</summary>domain/src/main/java/org/oppia/android/domain/classify/rules/multiplechoiceinput/MultipleChoiceInputModule.kt</details> | 100.00% | 19 / 19 | :x: | 101% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds

---

Also addressed all the review comments, can you PTAL?

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
```diff
@@ -419,22 +434,31 @@ class CoverageReporter(
...
+      report.details?.let {
+        totalLinesFound += it.linesFound
+        totalLinesHit += it.linesHit
+      }
+    }
+
+    return totalLinesFound.takeIf { it > 0 }
+      ?.let { totalLinesHit.toFloat() / it * 100 }
+      ?: 0.0f
+  }
```
BenHenning:
```suggestion
    val reports = coverageReportContainer.coverageReportList
    val totalLinesFound = reports.sumOf { it.linesFound }.toFloat()
    val totalLinesHit = report.sumOf { it.linesHit }.toFloat()
    return if (totalLinesFound > 0) (totalLinesHit * 100.0f) / totalLinesFound else 0.0f
```

More compact and idiomatic (no need to mess with mutable state).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -36,23 +36,23 @@ class CoverageRunner(
...
-  private fun retrieveCoverageResult(
-    bazelTestTarget: String
-  ): List<String>? {
-    return bazelClient.runCoverageForTestTarget(bazelTestTarget)
+    return coverageResults
+      .takeIf { it.isNotEmpty() }
+      ?.map { singleCoverageDatFileLines ->
+        parseCoverageDataFileLines(singleCoverageDatFileLines, bazelTestTarget)
+      }
+      ?: listOf(
```
```suggestion
      .map { singleCoverageDatFileLines ->
        parseCoverageDataFileLines(singleCoverageDatFileLines, bazelTestTarget)
      }.takeIf { it.isNotEmpty() } ?: listOf(
```

Maybe a bit easier to parse when reading this.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -84,8 +83,10 @@ fun main(vararg args: String) {
   val reportFormat = when (format) {
     "HTML" -> ReportFormat.HTML
     "MARKDOWN", "MD" -> ReportFormat.MARKDOWN
+    "PROTO" -> ReportFormat.PROTO
     else -> throw IllegalArgumentException("Unsupported report format: $format")
   }
+  println("format: $reportFormat")
```
BenHenning: If you're keeping this, I suggest making it a bit cleaner, e.g. ``"Using format: $reportFormat."``.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -56,8 +55,8 @@ const val BOLD = "\u001B[1m"
  *
  * Example with output path to save the collected coverage proto:
  *    bazel run //scripts:run_coverage -- $(pwd)
- *    utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
- *    --protoOutputPath=/tmp/coverage_report.proto64
+ *    utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt --format=PROTO
+ *    --protoOutputPath=/tmp/coverage_report.pb
```
BenHenning: Part of the reason I suggested using a PROTO output format is that you can avoid needing to pass in a path (the only time I can think of that we *must* pass in a path is when writing custom Bazel rules, so that doesn't apply here).

You could, instead, make some sort of predictable output path for all of the different report types. For example:

coverage_reports/utility/src/main/.../MathModel.kt/coverage_report.[pb/md/html]

This makes it easier when using the script as a user, but it also creates a predictable location for other scripts to read/retrieve these files.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -213,14 +240,15 @@ class RunCoverage(
...
-      return containerBuilder.build()
+  private fun combineCoverageReports(
+    coverageResultList: List<CoverageReport>
+  ): CoverageReportContainer {
+    val containerBuilder = CoverageReportContainer.newBuilder()
+    coverageResultList.forEach { report ->
+      containerBuilder.addCoverageReport(report)
     }
+    return containerBuilder.build()
+  }
```
BenHenning:
```suggestion
    return CoverageReportContainer.newBuilder().apply {
      addAllCoverageReport(coverageResultList)
    }.build()
  }
```

Another approach (optional).

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
```diff
@@ -530,11 +630,55 @@ class CoverageReporterTest {
...
+        .build()
+
+      val testExemptionPb = "test_exemption.pb"
+      val coverageTestExemptionTextProto = tempFolder.newFile(testExemptionPb)
+      coverageTestExemptionTextProto.outputStream().use { outputStream ->
+        testFileExemptions.writeTo(outputStream)
+      }
+
+      val testFileExemptionsFromFile =
+        TestFileExemptions.parseFrom(coverageTestExemptionTextProto.inputStream())
```
BenHenning: Not much value in parsing it--I suggest just using the exact structure that was assembled here, instead.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
```diff
@@ -530,11 +630,55 @@ class CoverageReporterTest {
...
+
+  private fun readFinalMdReport(): String {
+    return File(
+      "${tempFolder.root}" +
+        "$coverageDir/CoverageReport.md"
+    ).readText()
+  }
+
+  private fun createTestFileExemptionTextProto():
+    Map<String, TestFileExemptions.TestFileExemption> {
```
BenHenning: This should fit on one line--I suggest moving this up so that the method body can be indented like others in the class (for consistency).

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
BenHenning: Per discussion I think we want to remove these now, right? You're correct that it's really not simple since local vs. develop vs. a PR will all have different link contexts, and it's quite difficult to get it correct in each circumstance. :)

BenHenning (commented):
Thanks @Rd4dev! I think I'm fully up-to-date on reviewing the PR. Just had a few more comments--PTAL.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
Rd4dev: Having it in a single time caused max line length exceeded Lint check error or is there any other better way I can handle this?

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Yeah. this is the one thing that is left to do in this PR (apart from stuff with the recent review), will modify the template and update here (and PR description) with the most recent version of the report template.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Sorry if I’m off track, but I was considering introducing an output path because we’ll need it for CI. My initial plan was to run coverage in CI as shards, and with each shard's results saved as `coverage_report_shard***.pb` (most probably planning to have the encoded shard names as suffix - Although this approach hasn’t been implemented yet and might change if it doesn’t work as expected, would be implementing this today), the idea is that once all shards have finished running,

```mermaid
graph LR
    A[Coverage Run Shard 1] --> B1[Coverage Report Shard 1.pb]
    A2[Coverage Run Shard 2] --> B2[Coverage Report Shard 2.pb]
    A3[Coverage Run Shard 3] --> B3[Coverage Report Shard 3.pb]

    B1 --> C[Report Combiner Script]
    B2 --> C
    B3 --> C

    C --> D[Coverage Reporter]

    style A fill:#ccf,stroke:#333,stroke-width:2px
    style A2 fill:#ccf,stroke:#333,stroke-width:2px
    style A3 fill:#ccf,stroke:#333,stroke-width:2px
    style B1 fill:#cfc,stroke:#333,stroke-width:2px
    style B2 fill:#cfc,stroke:#333,stroke-width:2px
    style B3 fill:#cfc,stroke:#333,stroke-width:2px
    style C fill:#ff9,stroke:#333,stroke-width:2px
    style D fill:#f9f,stroke:#333,stroke-width:4px
```

we would collect all the .pb files and use a separate job with a script to merge them into a single coverage report container. This combined report would then be sent to the coverage reporter to generate a Markdown report.

I’m wondering if we can skip the output path, how can we determine the file path for CI. For local development, we could specify a predictable path with its own individual file path (similar to how HTML reports are handled), but for CI, what would be the preferred method for saving these files?

Also, for md, we would be having just one common report so chose a common location for it and (in future pr for uploading the comments (called as script from ci)) will be passing a location for it too to save it in ci runs and later use that to upload it as comment. It's only for HTML (local dev purposes) we have distinct paths.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Sure, updated with the same.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Okay. updated as suggested.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Ah! Got it, updated with the same logic.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks, updated with the same.

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
Rd4dev: Yeah, got it. Removed the parsing step.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
Rd4dev: Updated the md report replacing the links with details-summary tags:

**Template:**

## Coverage Report

### Results
Number of files assessed: 5
Overall Coverage: **94.26%**
Coverage Analysis: **FAIL** :x:
##

### Failure Cases

| File | Failure Reason |
|------|----------------|
| File.kt | No appropriate test file found for File.kt. |

### Failing coverage

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| <details><summary>RunCoverage.kt</summary>scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt</details> | 51.38% | 148 / 288 | :x: | 70% |
| <details><summary>MathModel.kt</summary>utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt</details> | 77.78% | 7 / 9 | :x: | 80% _*_ |
| <details><summary>MultipleChoiceInputModule.kt</summary>domain/src/main/java/org/oppia/android/domain/classify/rules/multiplechoiceinput/MultipleChoiceInputModule.kt</details> | 10.00% | 1 / 10 | :x: | 30% _*_ |

>**_*_** represents tests with custom overridden pass/fail coverage thresholds

### Passing coverage

<details>
<summary>Files with passing code coverage</summary><br>

| File | Coverage | Lines Hit | Status | Min Required |
|------|:--------:|----------:|:------:|:------------:|
| <details><summary>MathTokenizer.kt</summary>utility/src/main/java/org/oppia/android/util/math/MathTokenizer.kt</details> | 94.26% | 197 / 209 | :white_check_mark: | 70% |
</details>

### Exempted coverage
<details><summary>Files exempted from coverage</summary> <br>
<details><summary>ActivityComponent.kt</summary>app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt</details>
<details><summary>TestExemptedFile.kt</summary>app/src/main/java/org/oppia/android/app/activity/TestExemptedFile.kt</details>
<details><summary>SourceIncompatible.kt</summary>app/src/main/java/org/oppia/android/app/activity/SourceIncompatible.kt</details>

Rd4dev:
@BenHenning, thanks for the review, and addressed all the follow-up review comments, can you PTAL?

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
BenHenning: Ah, I thought it would fit. The other way is to directly import ``TestFileExemption`` so that it can be directly referenced instead of via ``TestFileExemptions``.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: The common MD comes from generating a report from the combined list proto, right? The MD suggestion here is just for when a developer selects to generate Markdown when running the script locally.

I was thinking that you could collect all the pb files by just iterating in the generated coverage_reports directory for all *.pb files--this is straightforward to do for both shell scripts (e.g. using ``find``) and in Kotlin.

scripts/src/java/org/oppia/android/scripts/coverage/CoverageReporter.kt
BenHenning: This looks so nice! Thanks for making the changes.

BenHenning (commented):
Thanks @Rd4dev! I just had one nit left, but other than that there's the open comment about output format for ``RunCoverage``. I think if you already have a very good idea on how to collect the pb files, I'm happy to approve this. If not, I think the simplification of predictably placing all the pb, md, and html files per test based on format might be preferable just to avoid having any special casing with proto files (and then you can collect the files as suggested in my other comment).

scripts/src/javatests/org/oppia/android/scripts/coverage/CoverageReporterTest.kt
Rd4dev: Thanks, now I've adapted the approach followed with the #5480 to handle the exemptionList, and (removed the old implementation) now they fit right within the max length.

```
private fun createTestFileExemptionsProtoFile(testFileExemptions: TestFileExemptions): String {}
```

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Got it, thanks for the suggestion @BenHenning, implemented saving the protos for each file in their respective folders based on their filepaths.

**with --format=PROTO**
```kotlin
app/src/main/java/org/oppia/android/app/activity/ActivityComponent.kt

/* would save it to */
./coverage_reports/app/src/main/java/org/oppia/android/app/activity/ActivityComponent/coverage_report.pb

```

And also updated the test to test this - testRunCoverage_withProtoReportFormat_savesCoverageReportProto

(Additionally, I am still exploring ways to collect the protos from all the matrix runs, and this suggestion would super helpful. Thanks!)

#

>The common MD comes from generating a report from the combined list proto, right?

"**Yes**", the common MD is for the final combined proto.

>The MD suggestion here is just for when a developer selects to generate Markdown when running the script locally.

But still the need to save md reports based on file name paths is quite unclear to me. Do we need to save each and every md report? Would that be useful to have something like:

```
| File.kt | 13% | 10 /  54 | :x: | 70% |
```

in a separate file, as the above will be the md report for a single file run, and thought md wouldn't add much value to local dev (even if we save it in a complete table form) and skipped saving them as individual reports and only made effort to save them in the final stage (for ci and also for local dev debugging) (probably I will add a print statement to the log the final md report path - which is right now missing (html reports do log them)).

Edit: Added log to display the generated file path

```
MARKDOWN report for the coverage analysis is generated at: /mnt/c/Users/Baskaran/AndroidStudioProjects/oppia-android/coverage_reports/CoverageReport.md
```

Rd4dev:
@BenHenning, updated the PR to include changes with #5480 and also added required test cases and adjustments, also removed the output path for the proto format and let the script decide the proto storage file. The thing still left is to address the need for [md report storage](https://github.com/oppia/oppia-android/pull/5461#discussion_r1712604026) for individual files (if mandatory will add them asap), can you PTAL?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: I actually don't think there's a strong imperative to match the pathing for the MD file other than consistency since it's only for local development. The additional path output line was a nice idea to add; I think the current state of the script is in a good state for local development so I don't have any strong feelings to make additional changes here.

BenHenning (approved):
Thanks @Rd4dev! This LGTM. I don't think any additional work is needed on the MD pathing (see my other comment).

Please hold from merging this until I finish my review pass on #5465 (will follow up here).

PR #5465, opened 2024-07-19
Fix part of #5343: Introduce new CI workflow for Code Coverage

Rd4dev:
## Explanation

Fixes part of #5343 

### Project
[PR 2.3 of Project 4.1]

### Changes Made
- Replicated the compute affected test utility to compute changed files to acquire the list of kotlin files changed based on changes in the local Oppia Android Git repository, which will be organized into shards. It categorizes files by their paths into buckets (app, data, domain, etc.) and applies sharding strategies (large, medium, small).
- Modified BazelClient to run coverage commands based on shard configurations.
- Introducing a new CI workflow replicating the unit_test.yml implementation.
- The Coverage workflow should **run only after** the unit test (bazel) workflow is **successfully completed**.
- With **[RunAllTests]** added to the title this should perform code coverage on every file
- This PR is intended only to run coverages and the storing of protos and uploading of comments are yet to be figured out with the upcoming PR.

### Todo:
- **[Done]** Add test cases with RunAllTests setting.
- When a test file is provided the corresponding source file need to be mapped and provided for Build Oppia File (to pre build it before code coverage).
- Need to look after the runs and verify them.
  - To run only if unit tests pass:
  - Below image represents the ci check where the unit test fails causing the coverage runs to skip but still throw a fail status.
image

## Essential Checklist


.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,345 @@
...
+          echo "Changed files (note that this might be all files if configured to run all or on the develop branch): $FILE_BUCKET_LIST"
+          echo "::set-output name=matrix::{\"changed-files-bucket-base64-encoded-shard\":[$FILE_BUCKET_LIST]}"
+          if [[ ! -z "$FILE_BUCKET_LIST" ]]; then
+            echo "::set-output name=can_skip_files::false"
+          else
+            echo "::set-output name=can_skip_files::true"
+            echo "No files are affected by this change. If this is wrong, you can add '[RunAllTests]' to the PR title to force a run."
+          fi
+
+  code_coverage_run:
```
Rd4dev, line 105: ### Todo:
To only run after bazel run tests pass

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
```diff
@@ -176,3 +176,4 @@ class AdministratorControlsActivity :
     administratorControlsActivityPresenter.handleOnSaveInstanceState(outState)
   }
 }
+
```
Rd4dev: ### Todo:
Revert these new line changes.
Introduced for running and analysing coverage check runs.

Rd4dev: ### Revert / Remove
- New line changes
- shard_count for MathModel
- processTimeout 
- debugging statements

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
```diff
@@ -0,0 +1,784 @@
...
+    // to verify that System.exit() is called, but it's helpful.
+    assertThat(exception).hasMessageThat().contains("System.exit()")
+    assertThat(pendingOutputStream.toString()).contains("Usage:")
+  }
+
+  @Test
+  fun testUtility_invalidArguments_printsUsageStringAndExits() {
+    for (argCount in 0..3) {
+      val args = Array(argCount) { "arg${it + 1}" }
+      val exception = assertThrows<SecurityException> { main(arrayOf(*args)) }
```
Rd4dev: ### Todo
per https://github.com/oppia/oppia-android/pull/5461#discussion_r1706338158 assert directly instead of using any logic.

Rd4dev:
@BenHenning @adhiamboperes the PR is still in an intermediate Stage A - C (needs few refactoring and verifying the workflows) but would it be possible to have a 1st pass to check if the implementation and flows are done right. 

For the most part it replicates the existing unit test functionalities but with files.

scripts/src/java/org/oppia/android/scripts/ci/ComputeAffectedTests.kt
```diff
@@ -127,8 +127,6 @@ class ComputeAffectedTests(
     } else computeAffectedTargetsForNonDevelopBranch(gitClient, bazelClient, rootDirectory)
 
     val filteredTestTargets = filterTargets(affectedTestTargets)
-    println()
-    println("Affected test targets:")
```
BenHenning: Why was this removed?

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
```diff
@@ -74,6 +74,7 @@ class CoverageRunner(
     )
 
     val fileSpecificCovDatLines = coverageData.subList(sfStartIdx, sfStartIdx + eofIdx + 1)
+    println("File specific Coverage data line for $bazelTestTarget: $fileSpecificCovDatLines")
```
BenHenning: Why is this needed now?

scripts/src/javatests/org/oppia/android/scripts/ci/BUILD.bazel
```diff
@@ -31,3 +31,33 @@ kt_jvm_test(
         "//third_party:org_jetbrains_kotlin_kotlin-test-junit",
     ],
 )
+
+kt_jvm_test(
+    name = "ComputeChangedFilesTest",
+    size = "large",
+    srcs = ["ComputeChangedFilesTest.kt"],
+#    shard_count = 24,
```
BenHenning: Can this be removed now or is sharding needed?

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -0,0 +1,330 @@
...
+
+    val searchFiles = RepositoryFile.collectSearchFiles(
+      repoPath = pathToRoot,
+      expectedExtension = ".kt",
+      exemptionsList = testFileExemptionList
+    )
+
+    return searchFiles
+      .filter { it.name.endsWith(".kt") && !it.name.endsWith("Test.kt") }
+      .map { rootDirectory.toPath().relativize(it.toPath()).toString() }
```
BenHenning:
```suggestion
    return searchFiles
      .filter { it.extension == "kt" && !it.nameWithoutExtension.endsWith("Test") }
      .map { it.toRelativeString(rootDirectory) }
```

Some Kotlin-specific simplifications.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  private fun computeChangedFilesForNonDevelopBranch(
+    gitClient: GitClient,
+    rootDirectory: File
+  ): List<String> {
+    return gitClient.changedFiles
+      .map { File(rootDirectory, it) }
+      .filter { it.exists() }
+      .map { rootDirectory.toPath().relativize(it.toPath()).toString() }
```
BenHenning: Ditto here (can use ``toRelativeString``, instead).

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  private fun filterFiles(files: List<String>) : List<String> {
+    // Filtering out files that need to be ignored.
```
BenHenning: Why is this filtering necessary? The comment also needs more context since it's not clear why these specific files should be ignored.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -0,0 +1,330 @@
...
+
+  private fun computeAllFiles(
+    rootDirectory: File,
+    pathToRoot: String
+  ): List<String> {
+    val testFileExemptiontextProto = "scripts/assets/test_file_exemptions"
+    val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptiontextProto)
+      .testFileExemptionList
+      .filter { it.testFileNotRequired }
+      .map { it.exemptedFilePath }
```
BenHenning: I'm actually thinking perhaps we should ignore exemptions in these two new scripts and just go ahead and try to run coverage for exempted tests. It should just result in them being compiled in the final report as exempted, right? Seems a bit cleaner than filtering them out (even if it means a longer CI run; it's at least more explicit to the user why their changed file or test doesn't have new coverage being run).

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: Also, is this actually what we want? Shouldn't we be mapping tests back to their source files since changing tests in isolation also means that they need to be re-run for coverage?

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+    else -> {
+      listOf(filePath.replace("/main/", "/test/").replace(".kt", "Test.kt"))
+    }
+  }
+
+  return possibleTestFilePaths
+    .map { File(rootDirectory, it) }
+    .filter(File::exists)
+    .map { it.relativeTo(rootDirectory).path }
```
BenHenning: Can use ``toRelativeString`` here, as well.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
```diff
@@ -0,0 +1,127 @@
...
+      writer.println(changedFilesBucket.changedFilesList.joinToString(separator = " "))
+    }
+
+    fileTestTargetsListOutputFile.printWriter().use { writer ->
+      writer.println(changedFilesTestTargetWithoutSuffix.joinToString(separator = " "))
+    }
+  }
+}
+
+private fun findTestFile(rootDirectory: File, filePath: String): List<String> {
```
BenHenning, line 98: Why is this needed? Shouldn't we be processing source files up to the actual coverage runner script?

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -144,13 +144,37 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
     )
-    return parseCoverageDataFilePath(coverageCommandOutputLines)?.let { path ->
-      File(path).readLines()
+
+    val hasShardCount = buildRule.any { "shard_count" in it }
+    if (hasShardCount) {
+    executeBazelCommand(
+        "test",
+        "--collect_code_coverage",
+        "--combined_report=lcov",
```
BenHenning: Can we perhaps just use this attribute for all runs instead of only when there's a sharded test?

BenHenning (changes requested):
Skipped reviewing the workflow & tests for now; gotta go to a meeting and I wanted to get my initial comments out--PTAL.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
Rd4dev: The reason the source files are mapped to find the test files is to get the test targets that can be Built prior to running coverages on the tests. **(for Build Oppia Tests)**

**More context:**

(Correct me if wrong),
Though we provide source files as inputs to our run coverage scripts, the coverages are actually performed running the bazel commands on the test targets. 

In the workflow job order we set up bazel -> build the targets (to help with runs not timing out) -> run the script

So, I thought it might be helpful to have the targets built prior to running, as running coverages by itself would be expensive and having it built and run in one single job might cause timeouts. So, I am mapping these source files to test files, retrieving their test targets and passing them to the BUILD OPPIA TESTS job. 

am I doing it right?

Also, I was additionally thinking of adding one more functionality to 
if we have changed files as `**Test.kt` 
then to acquire its test target too. If the above implementation is right would proceed else will skip it then.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: The targets that belonged to the `//instrumentation/src/javatests/org/oppia/android/instrumentation/player` path were skipped in the **[ComputeAffectedTests.kt](https://github.com/oppia/oppia-android/blob/develop/scripts/src/java/org/oppia/android/scripts/ci/ComputeAffectedTests.kt#L129)**, 

```kotlin
private fun filterTargets(testTargets: List<String>): List<String> {
    // Filtering out the targets to be ignored.
    return testTargets.filter { targetPath ->
      !targetPath
        .startsWith(
          "//instrumentation/src/javatests/org/oppia/android/instrumentation/player",
          ignoreCase = true
        )
    }
  }
```

since coverages is also essentially running tests but collecting coverages additionally, thought to just follow along with whatever set up was done with the existing utility.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: The reason for the first part of the question is addressed https://github.com/oppia/oppia-android/pull/5465#discussion_r1707500694 here.

And reg mapping tests to sources, yes, we will be mapping tests to sources (eg if TestFileCheckTest.kt is changed coverage runs will be run on TestFileCheck.kt with run coverage) and that was handled in the run coverage script PR 2.2 as provided in the [PR description](https://github.com/oppia/oppia-android/pull/5461#issue-2399902488) -

> The kotlin files which are test files are mapped to their appropriate source files and then provided for coverage analysis.
So the provided list of files: [utility/.../MathTokenizer.kt, app/.../layout.xml, scripts/.../TestFileCheckTest.kt] will only do coverage analysis for the files "utility/.../MathTokenizer.kt", "scripts/.../TestFileCheck.kt".

scripts/src/java/org/oppia/android/scripts/coverage/CoverageRunner.kt
Rd4dev: Apologies for the oversight. was in a rush to get the PR out and didn’t handle the debugging lines properly.

scripts/src/java/org/oppia/android/scripts/ci/ComputeAffectedTests.kt
Rd4dev: Apologies, I was removing the debugging lines that I added and accidentally removed these lines thinking it was my change. Will re add it in the next commit.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: I was initially a bit hesitant with using this (`bazel test --collect_code_coverage --combined_report=lcov`) method since it was saving it to one common dir and though we handle things sequentially now, I'm still unsure of what all exceptions or edge cases, coverage runs would have for different files.

Just in a scenario with 10 files running in a shard if file1 ran successfully and file2 too ran successfully but with no .dat produced or any unknown exception, now trying to extract the .dat from the common location would result in mis representation of coverage.

The right way is,
This shouldn't happen at all or should be handled if this ever occurs, but I'm just not sure if it would or might have the possibility. So to be just always sure we have **coverage.dat created specific to its own files/targets** printed right in the terminal for its own run would have the advantages or 
1. to figure out instantly if .dat is not produced making it explicit to handle any unknown cases that might occur 
2. to be sure each run corresponds to its own .dat 

ultimately even shard configured coverage runs too need to be fool proof in this scenario, but what kind of edge case could occur is just unknown to me at this point. 

What would you suggest @BenHenning should we stick to just one common command bazel test --collect_code_coverage or as 2 separate commands.

scripts/src/javatests/org/oppia/android/scripts/ci/BUILD.bazel
Rd4dev: Ah, apologies. I initially thought sharding might be needed, but with just file changes, it ran like a lightning.  wouldn't be needing sharding, removed it.

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
```diff
@@ -31,3 +31,34 @@ kt_jvm_library(
...
+    srcs = [
+        "ComputeChangedFiles.kt",
+    ],
+    visibility = ["//scripts:oppia_script_binary_visibility"],
+    deps = [
+        "//scripts/src/java/org/oppia/android/scripts/common:git_client",
+        "//scripts/src/java/org/oppia/android/scripts/common:proto_string_encoder",
+        "//scripts/src/java/org/oppia/android/scripts/common:repository_file",
+        "//scripts/src/java/org/oppia/android/scripts/proto:changed_files_java_proto",
+        "//scripts/src/java/org/oppia/android/scripts/proto:script_exemptions_java_proto",
```
Rd4dev: ### Todo:
To remove script exemption deps with removal of test file exemptions

scripts/BUILD.bazel
```diff
@@ -247,6 +247,22 @@ kt_jvm_binary(
     ],
 )
 
+kt_jvm_binary(
```
Rd4dev, line 250: ### Todo:
Update with removal of test file exemptions

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Got it. Removed it from the ComputeChangedFiles.kt as that would decide what all files need to be computed, but with the RetrieveChangedFiles.kt it uses the testFileExemption just to retrieve the targets for 'building'

```kotlin
val changedFilesTestFiles = changedFilesBucket.changedFilesList.flatMap { changedFile ->
  val exemption = testFileExemptionList[changedFile]
  if (exemption != null && exemption.testFileNotRequired) {
    emptyList()
  } else {
    findTestFile(rootDirectory, changedFile)
  }
}
```

can I keep it over here, to exempt the file from searching for a test file?

This is just for 'Build Oppia Tests' and wouldn't be affecting the computed files.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Done, thanks.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
Rd4dev: Done, thanks.

Rd4dev:
Thank you so much @BenHenning, addressed the existing review comments (most of them are just explaining the reasons and asking for suggestions how to proceed), can you PTAL whenever possible?

.github/workflows/code_coverage.yml
Rd4dev: Done verified with the previous runs, and code coverage only starts after unit tests successfully get completed. 
If unit tests check fails, then it skips coverage but still fails the code coverage check.

scripts/BUILD.bazel
Rd4dev: removed the data attribute for `compute_changed_files`, will update for retrieve_changed_files as needed as per https://github.com/oppia/oppia-android/pull/5465#discussion_r1707454378

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: Done, now the tests are performed as separately without any logic.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Thanks Ben, with this fix - https://github.com/oppia/oppia-android/pull/5480/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR158, now it just has one single command to execute the bazel coverage.

```
executeBazelCommand(
  "coverage",
  bazelTestTarget,
  "--instrumentation_filter=$computeInstrumentation"
)
```

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
Rd4dev: Removed it for the compute_changed_files_lib, yet to be decided for retrieve_changed_files. - https://github.com/oppia/oppia-android/pull/5465#discussion_r1707454378

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: Ah. In that case, I suggest updating the comment like so:

``Filter out instrumentation files since code coverage only runs on unit tests.``

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,315 @@
...
+        env:
+          BAZEL_CACHE_DIR: ${{ env.CACHE_DIRECTORY }}
+        run: |
+          EXPANDED_BAZEL_CACHE_PATH="${BAZEL_CACHE_DIR/#\~/$HOME}"
+          echo "Using $EXPANDED_BAZEL_CACHE_PATH as Bazel's cache path"
+          echo "build --disk_cache=$EXPANDED_BAZEL_CACHE_PATH" >> $HOME/.bazelrc
+        shell: bash
+
+      # See explanation in bazel_build_app for how this is installed.
+      - name: Install git-secret (non-fork only)
```
BenHenning: You can remove:
- git-secret step
- Decrypt secrets
- All 'with caching' steps

They aren't used anymore--they've been disabled for >2 years now I believe since they require too much bandwidth. It'll be nicer to keep this workflow on the smaller side.

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,315 @@
...
+jobs:
+#  check_unit_tests_completed:
+#    name: Check unit test completed
+#    runs-on: ubuntu-latest
+#    steps:
+#      - name: Wait for unit tests to checks
+#        uses: ArcticLampyrid/action-wait-for-workflow@v1.2.0
+#        with:
+#          workflow: unit_tests.yml
+#          sha: auto
```
BenHenning: Is this disabled for temporary testing?

scripts/src/javatests/org/oppia/android/scripts/ci/RetrieveChangedFilesTest.kt
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  @After
+  fun tearDown() {
+    // Reinstate test output redirection.
+    System.setOut(PrintStream(pendingOutputStream))
+
+    // Print the status of the git repository to help with debugging in the cases of test failures
+    // and to help manually verify the expect git state at the end of each test.
+    println("git status (at end of test):")
```
BenHenning, line 50: Git status isn't being printed out--can this be removed?

scripts/src/javatests/org/oppia/android/scripts/ci/RetrieveChangedFilesTest.kt
```diff
@@ -0,0 +1,290 @@
...
+
+    // Bazel catches the System.exit() call and throws a SecurityException. This is a bit hacky way
+    // to verify that System.exit() is called, but it's helpful.
+    assertThat(exception).hasMessageThat().contains("System.exit()")
+    assertThat(pendingOutputStream.toString()).contains("Usage:")
+  }
+
+  @Test
+  fun testUtility_invalidBase64_throwsException() {
+    assertThrows<IllegalArgumentException>() {
```
BenHenning: Can the cause message of the exception be verified?

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
```diff
@@ -0,0 +1,826 @@
...
+    val reportedFiles = runScript()
+
+    // An empty workspace should yield no files.
+    assertThat(reportedFiles).isEmpty()
+  }
+
+  @Test
+  fun testUtility_developBranch_returnsAllFiles() {
+    initializeEmptyGitRepository()
+    switchToFeatureBranch()
```
BenHenning: Aren't this & the other 'develop' tests actually running on a feature branch? Is the script actually behaving correctly?

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
```diff
@@ -0,0 +1,826 @@
...
+    initializeEmptyGitRepository()
+    createAndCommitFile("First", "Second", "Third", subPackage = "app")
+    switchToFeatureBranch()
+    createFiles("NewUntrackedFile", subPackage = "data")
+
+    val reportedFiles = runScript()
+
+    // Only the first file should be reported since the file itself was changed & staged.
+    assertThat(reportedFiles).hasSize(1)
+    assertThat(reportedFiles.first().changedFilesList).containsExactly("data/NewUntrackedFile.kt")
```
BenHenning: This seems backwards from what the test name is suggesting.

I worry a bit about overly matching ComputeAffectedTests--a lot of those tests are verifying that the script always only considers fully checked-in files for deltas (since it's meant to run in CI). If you plan to replicate that behavior, this test seems to be behaving wrong since untracked files should not be part of the 'changed files' result.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
```diff
@@ -0,0 +1,826 @@
...
+
+    val reportedFiles = runScript()
+
+    assertThat(reportedFiles).hasSize(1)
+    assertThat(reportedFiles.first().changedFilesList).doesNotContain(
+      "instrumentation/src/javatests/org/oppia/android/" +
+        "instrumentation/player/InstrumentationFile.kt"
+    )
+    assertThat(reportedFiles.first().changedFilesList).contains(
+      "instrumentation/src/javatests/org/oppia/android/instrumentation/app/Robolectric.kt"
```
BenHenning, line 320: Isn't this supposed to be ignored (per the expected behavior & the name of the test)?

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -0,0 +1,315 @@
...
+    pathToRoot: String
+  ): List<String> {
+    val searchFiles = RepositoryFile.collectSearchFiles(
+      repoPath = pathToRoot,
+      expectedExtension = ".kt",
+    )
+
+    return searchFiles
+      .filter { it.extension == "kt" && !it.nameWithoutExtension.endsWith("Test") }
+      .map { it.toRelativeString(rootDirectory) }
```
BenHenning, line 148: It seems worth it to me to also add tests for ensuring that changes to test and non-Kotlin files are excluded from results.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: I think the latest version is probably the cleanest (one per file/shard and then we combine them when necessary). It gives us a lot more control rather than relying on Bazel behavior (which may not always be entirely understandable when it comes to merging the dat files).

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
BenHenning: So just to check, your plan is to:
- Collect all changed source files (to be passed into run_coverage).
- Collect test targets associated with those source files.
- Use the test targets to build tests ahead of running coverage for improved performance.
- Run coverage for each source file.

Is that correct? If so, I have no concerns with that.

However, it seems to be that the script still needs to map from test file back to source file. If a test is changed in a PR in isolation, we also need to re-run coverage for its corresponding source file. That flow seems missing in the current implementation (unless I'm missing something). This is also a very important case to cover, so we should definitely have tests for it (as we make isolated changes to tests all the time, and that could cause changes in coverage that we want to surface to the PR comment thread).

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: Yep I think it definitely makes sense to keep here.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: But aren't we filtering out the tests here from the changed list (so they won't ever be passed to run_coverage)? I also followed up on the other conversation thread.

scripts/src/javatests/org/oppia/android/scripts/ci/BUILD.bazel
BenHenning: A got it. Does it still need the "large" size setting if it now runs quickly?

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
BenHenning: Are these all addressed now (i.e. can this conversation thread now be resolved)?

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
BenHenning: Is this addressed now (i.e. can this conversation thread now be resolved)?

scripts/BUILD.bazel
BenHenning: Is this addressed now (i.e. can this conversation thread now be resolved)?

BenHenning (changes requested):
Thanks @Rd4dev! Took a full pass on the PR now, and had some follow-up comments. There doesn't actually seem to be that much to address, I think; it's mostly smallish items. PTAL.

app/src/main/java/org/oppia/android/app/administratorcontrols/AdministratorControlsActivity.kt
Rd4dev: Yeah, will resolve it now.

scripts/src/java/org/oppia/android/scripts/ci/BUILD.bazel
Rd4dev: Yeah, removed the dependencies for compute changed files while leaving it to stay as needed for retrieve changed files. 
Will resolve this comment.

scripts/BUILD.bazel
Rd4dev: Yeah, will resolve it as appropriate changes are made.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Actually, the code this comment was pointing to is under **computeAllFiles**. 
We don't filter out the test files under **computeChangedFilesForNonDevelopBranch**.

**More context:**

- computeAllFiles: which is only collected if the title includes **[RunAllTests]** would probably be getting all the source files so it wouldn't be needing the test files to map them back for run coverage. So excluded them here.

- computeChangedFilesForNonDevelopBranch: is the one that will run on a usual PR run, which collects the changed files - including (file.kt, fileTest.kt, file.xml, etc), then we filter it to only include the `.kt` (both source and tests) files which is then used to segregate them as buckets.

so, the final computed list of files will have source and test files. and as suggested in the linked follow up comment,
>However, it seems to be that the script still needs to map from test file back to source file. If a test is changed in a PR in isolation, we also need to re-run coverage for its corresponding source file.

the test file needs to be mapped to it source file. _(will also update the flow there)_

The actual place where the **mapping of test files -> source file** happens is in the **`RunCoverage.kt`** script.

### RunCoverage.kt
```kotlin

val filePathList = args.drop(1)
.takeWhile { !it.startsWith("--") }
.map { it.trim(',', '[', ']') }
.flatMap { filePath ->
  when {
    filePath.endsWith("Test.kt") -> {
      findSourceFile(File(repoRoot).absoluteFile, repoRoot, filePath)
    }
    filePath.endsWith(".kt") -> listOf(filePath)
    else -> emptyList()
  }
}
```

since I thought instead of having it filtered / mapped in 2 places (compute changed files and local dev script run) it might be a better to have it one central spot. (in RunCoverage.kt)

---

But now I have different opinions,

1. I could actually filter out the test files here itself **after mapping them to their sources**, since that was handled by the script, I did not bother to do that earlier but now I think that would be a cleaner way to do here too.

---

2. Also, just while answering this I bumped into an edge case that if we provide both the source and the test files (or compute changed files detects source and its related test file change (to script run) then it kind of runs it separately for the single source file each time (one with source file input, another with test mapped to source input), 

eg:
```
bazel run //scripts:run_coverage -- $(pwd) 
utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt
utility/src/test/java/org/oppia/android/util/parser/math/MathModelTest.kt
```

is doing

```
Running coverage analysis for the files: 
[utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt,
utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt]
```

which should be:

```
Running coverage analysis for the files: 
[utility/src/main/java/org/oppia/android/util/parser/math/MathModel.kt]
```

This could be resolved if we filter out the tests right in compute affected tests itself for ci and for local dev, I will probably check if the source exists in the list of files and filter that out. 

would like to know if there are any suggestions with this in mean time and will update this comment later with the changes made.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
Rd4dev: Following up with the review comment - https://github.com/oppia/oppia-android/pull/5465/files/4e4c6877a76af9f1e2fdef8413a92268a9becd90#r1711077577

To specifically answer this ->

The flow is almost right, 
- Collect **all changed source and test files** (to be passed into run_coverage) - all `.kt` files.
- step 2 (as above)
- step 3 (as above)
- Should be able to provide source and test files to Run coverage.
- The Run coverage script maps the test files to source files.
- Run coverage for each source file.

The mapping of test to sources happens in the RunCoverage.kt script (run_coverage) as mentioned in the PR 2.2 description as introduced with enabling support for list of files: #5461  

>[Done] The test files too need to be taken into account while considering coverage analysis (mapping test files to their source files and provide it to the script run, also these files need to be figured out even in the compute changed files utility as that will help with building the targets)

This can be seen in action in the last ci run - running code coverage check (code_coverage.yml) [[stack trace]](https://github.com/oppia/oppia-android/actions/runs/10311895866/job/28546379396?pr=5465#step:17:37)

image

---

### Changes I am planning to make

Reason for having the mapping of test files to source file in the run_coverage script is to have it in a centralized spot as both ci will need to have the mapping and if locally a developer tries to provide a test file (even then) this mapping needs to be done, so had it there.

But now I am planning to also have the mapping in the ci too
1. this could help with also retrieving the source's related test target - to help with build
2. could avoid duplicate coverage run as mentioned in the other comment
3. also would have a cleaner approach for ci

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Added mapping of test to source files in the recent changes

Few things handled are:
- Not all source and test files are properly positioned in their respective packages, while this was handled earlier while mapping test files int the run coverage this also need to be handled with mapping sources, so here for any test file that does not have or could not find its respective path the logic returns a null value and later filtered out. (The appropriate source files will indeed be exempted as incompatible source files)
- Used distinct to remove any duplicate entries in the list of files
- The mapping of test to source files is included in the compute changed files (also to help with acquiring its respective test targets for building).

Tests for this is yet to be added and this needs to be verified on the ci runs.

Edit:
Added the following tests:
- testUtility_featureBranch_withChangesToTestFile_mapsTestFileToSourceFile
- testUtility_featureBranch_withChangesToSourceAndTestFile_computesSingleSourceFile

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Done, updated with the same.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
Rd4dev: Added logic: https://github.com/oppia/oppia-android/pull/5465#discussion_r1711545482

.github/workflows/code_coverage.yml
Rd4dev: Yeah, it was hard to verify while waiting for unit tests to pass, will make sure to revert it once all coverage workflows are verified.

I did confirm that it only gets triggered after the unit tests are successfully completed.

.github/workflows/code_coverage.yml
Rd4dev: Got it, removed all the mentioned steps.

scripts/src/javatests/org/oppia/android/scripts/ci/BUILD.bazel
Rd4dev: Sorry for overlooking that; I've removed it now.

scripts/src/javatests/org/oppia/android/scripts/ci/RetrieveChangedFilesTest.kt
Rd4dev: I actually did not understand why the status was printed even in `RetrieveAffectedTestsTest` (since those tests too didn't involve any git operations, I guess) but just went along with the same.

But as suggested removed it here.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: The filter only filters out files in the path - "instrumentation/src/javatests/org/oppia/android/instrumentation/player", so since instrumentation/src/javatests/org/oppia/android/instrumentation/app/Robolectric.kt follows a different path, it gets included.

I realize the test name is misleading and maybe I can split them as 2 test cases specifically for their own conditions.

Edit: updated tests, 
- testUtility_developBranch_instrumentationModuleChanged_filteredFilesAreIgnored
- testUtility_developBranch_instrumentationModuleChanged_unfilteredFilesAreComputed
- testUtility_featureBranch_instrumentationModuleChanged_filteredFilesAreIgnored
- testUtility_featureBranch_instrumentationModuleChanged_unfilteredFilesAreComputed

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: Apologies, mistakenly switched to feature branch with the test `testUtility_developBranch_returnsAllFiles` fixed that now.
And can I know if any other develop test are running in the feature branch, I can't find anything else (am I missing anything).

Reg the behaviour: It retrieves to check if all the committed files are computed and that is the intended result we are looking for. If I'm not wrong? can I get more context if that was not the expected script behaviour.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Got it added the following test cases:
- testUtility_featureBranch_computeAllFiles_filtersOutNonKotlinFiles
- testUtility_featureBranch_computeAllFiles_filtersOutTestFiles

scripts/src/javatests/org/oppia/android/scripts/ci/RetrieveChangedFilesTest.kt
Rd4dev: Sure, added the following:

```kotlin
// Indicates the Proto data is improperly formatted or corrupted.
assertThat(exception).hasMessageThat().contains(
  "Last unit does not have enough valid bits"
)
```

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: If the question is on **"why the even after switching to feature branch it returned the list of files?"** (making the test case PASS) - (in case) to address that, -> I had committed the files only after switching to the feature branch, so it collected all the files in the feature branch and computed it.

#

If we had tried to commit the files before switching it would have thrown an error -

image

The updated test now handles it correctly in the develop branch itself.

---

But the comment mentioned here is yet to be addressed - it makes sense, https://github.com/oppia/oppia-android/pull/5465/files#r1710639970 and I will probably need to remove untracked files from the changed files list.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: Oh! Thank you so much for clarifying that, I was wondering if my wsl and git are functioning wrong as it was printing out a bunch of other files (while ci didn't), seems as you said I was doing it all wrong. And also, I only had a very high-level understanding of computing affected test targets and tried replicating it.

So, now I changed the script implementation to acquire **only the committed files** by using the existing method,

```
private fun retrieveChangedCommittedFiles(): List<String> {
    return executeGitCommand("diff --name-only ${computeCommitRange()}")
  }
```

hope now the script should be functioning in the way it should if not will adjust it as per feedback asap.

---

with this will also modify the tests and update here.

Edit: Updated the test as necessary.

Rd4dev:
@BenHenning thanks for the review, I addressed all the comments, can you PTAL?

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
```diff
@@ -296,28 +289,33 @@ class ComputeChangedFilesTest {
...
-    createFiles(
+    createAndCommitFiles(
       "InstrumentationFile",
       subPackage = "instrumentation/src/javatests/org/oppia/android/instrumentation/player"
     )
-    createFiles(
+
+    val reportedFiles = runScript()
+
+    assertThat(reportedFiles).hasSize(0)
```
BenHenning:
```suggestion
    assertThat(reportedFiles).isEmpty()
```

Ditto elsewhere in the test file.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
BenHenning: That helps--thanks!

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
```diff
@@ -167,6 +184,44 @@ class ComputeChangedFiles(
...
+              filePath.replace("/test/", "/main/").replace("Test.kt", ".kt"),
+              filePath.replace("/test/", "/main/").replace("LocalTest.kt", ".kt")
+            )
+          }
+          else -> emptyList()
+        }
+      }
+      else -> {
+        listOf(filePath.replace("/test/", "/main/").replace("Test.kt", ".kt"))
+      }
```
BenHenning, line 214: Are all of these different cases covered in tests? Please add tests to cover these since a regression here could be quite problematic (we'd drop re-checking code coverage for those files, leading to a potential eventual failure in the develop branch once the PR is merged).

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,258 @@
...
+        env:
+          BAZEL_CACHE_DIR: ${{ env.CACHE_DIRECTORY }}
+        run: |
+          EXPANDED_BAZEL_CACHE_PATH="${BAZEL_CACHE_DIR/#\~/$HOME}"
+          echo "Using $EXPANDED_BAZEL_CACHE_PATH as Bazel's cache path"
+          echo "build --disk_cache=$EXPANDED_BAZEL_CACHE_PATH" >> $HOME/.bazelrc
+        shell: bash
+
+      - name: Build Oppia Tests (without caching, or on a fork)
+        if: ${{ env.ENABLE_CACHING == 'false' || ((github.ref != 'refs/heads/develop' || github.event_name != 'push') && (github.event.pull_request.head.repo.full_name != 'oppia/oppia-android')) }}
```
BenHenning:
```suggestion
      - name: Build Oppia Tests
```

More caching changes that can be removed.

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,258 @@
...
+          done
+          # Capture the error code of the final command run (which should be a success if there isn't a real build failure).
+          last_error_code=$?
+          # Reenable exit-on-first-failure.
+          set -e
+          # Exit only if the most recent exit was a failure (by using a subshell).
+          (exit $last_error_code)
+
+      - name: Run Oppia Coverage (without caching, or on a fork)
+        if: ${{ env.ENABLE_CACHING == 'false' || ((github.ref != 'refs/heads/develop' || github.event_name != 'push') && (github.event.pull_request.head.repo.full_name != 'oppia/oppia-android')) }}
```
BenHenning:
```suggestion
      - name: Run Oppia Coverage
```

More caching changes that can be removed.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
BenHenning: I think the committed files probably makes sense as the source of truth. Is there anything else that needs to be addressed here? It seems you fixed the discrepancies in the latest changes unless I missed something.

scripts/src/javatests/org/oppia/android/scripts/ci/RetrieveChangedFilesTest.kt
BenHenning: Ah. It was done to provide some additional context when tests fail in specific tests that highly rely on Git environment state (like ComputeAffectedTests). It *can* be helpful to debug failures, but isn't strictly necessary.

scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
BenHenning: Ah I see--thanks for clarifying & making the changes!

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: Hmm I do think this approach is cleaner. I actually didn't even realize (or forgot) that run_coverage supported being passed test files until this PR, but I think it makes more sense to run the CI pipeline on all affected source files to keep things as simple as possible. I think it's also fine to keep the test->source file mapping in run_coverage since that's probably something a developer might try when using it locally.

BenHenning (commented):
Thanks @Rd4dev! Just had a few follow-ups, but otherwise the PR is looking quite good to me.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: Done, updated as suggested.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
Rd4dev: >Is there anything else that needs to be addressed here?

Sorry, forgot to update it here.

>But the comment mentioned here is yet to be addressed - it makes sense, https://github.com/oppia/oppia-android/pull/5465/files#r1710639970 and I will probably need to remove untracked files from the changed files list.

The above mentioned changed was implemented by only considering committed files. Now the scripts function as expected.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: It did not, added the following test cases now:
- testUtility_featureBranch_appTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_appLocalTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_appSharedTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_scriptTestChanges_mapsToRespectiveSourceFile

And for other modules,
- testUtility_featureBranch_domainModuleTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_dataModuleTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_utilityModuleTestChanges_mapsToRespectiveSourceFile
- testUtility_featureBranch_testingModuleTestChanges_mapsToRespectiveSourceFile

---

If possible, can I understand what the following specifically means,
> to cover these since a regression here could be quite problematic (we'd drop re-checking code coverage for those files, leading to a potential eventual failure in the develop branch once the PR is merged).

.github/workflows/code_coverage.yml
Rd4dev: Done, and should this also involve removing the environment variable?
```
ENABLE_CACHING: false
```

.github/workflows/code_coverage.yml
Rd4dev: Done, removed it.

Rd4dev:
@BenHenning, thanks for the review, addressed the remaining comments, with just 2 requests for clarification,
1. To understand the [comment](https://github.com/oppia/oppia-android/pull/5465#discussion_r1712388166) mentioning 
>to cover these since a regression here could be quite problematic (we'd drop re-checking code coverage for those files, leading to a potential eventual failure in the develop branch once the PR is merged).
2. If `ENABLE_CACHING: false` too needs to be removed?

Can you PTAL?

.github/workflows/code_coverage.yml
```diff
@@ -237,12 +236,11 @@ jobs:
...
           (exit $last_error_code)
 
-      - name: Run Oppia Coverage (without caching, or on a fork)
-        if: ${{ env.ENABLE_CACHING == 'false' || ((github.ref != 'refs/heads/develop' || github.event_name != 'push') && (github.event.pull_request.head.repo.full_name != 'oppia/oppia-android')) }}
+      - name: Run Oppia Coverage
         env:
           CHANGED_FILES: ${{ env.CHANGED_FILES }}
         run: |
-          bazel run //scripts:run_coverage -- $(pwd) $CHANGED_FILES --format=PROTO --processTimeout=15
+          bazel run //scripts:run_coverage -- $(pwd) $CHANGED_FILES --format=MARKDOWN --processTimeout=15
```
BenHenning, line 242: Is this correct? Shouldn't it be PROTO?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -183,26 +183,6 @@ class RunCoverage(
...
-      val testFilePaths = findTestFiles(rootDirectory, repoRoot, filePath)
-      if (testFilePaths.isEmpty()) {
-        return CoverageReport.newBuilder()
-          .setFailure(
-            CoverageFailure.newBuilder()
-              .setFilePath(filePath)
-              .setFailureMessage("No appropriate test file found for $filePath.")
-              .build()
-          ).build()
-      }
```
BenHenning: Why is this being removed now?

.github/workflows/code_coverage.yml
BenHenning: Ah good catch; that can be removed, too.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
BenHenning: Ah thanks--these are great tests to have!

What I meant by my earlier message is essentially the following scenario:
- Other developers change this script (or utilities it depends on in the future) causing it to no longer map from test to source file correctly for, say, local tests in app module. Because there weren't unit tests, everyone fails to notice the regression and it gets checked in.
- A developer in the future performs a refactor on one of the shared tests and inadvertently reduces code coverage for the source file corresponding to that test.
- Due to the regression in this script, code coverage analysis isn't re-run for that refactored shared test and everyone misses that the code coverage dropped.
- If the coverage dropped below the enforcement threshold, then once the PR gets merged in the develop CI will start failing (since it always runs code coverage on all source files).
- The recourse here is to revert the PR that was merged and fix the code coverage gap that (eventually) caused the develop branch CI to fail.

These tests help prevent this scenario from happening.

scripts/src/javatests/org/oppia/android/scripts/ci/ComputeChangedFilesTest.kt
BenHenning: Ah excellent--it seemed like the changes were completed but I wanted to make sure. Thanks!

BenHenning (commented):
Thanks @Rd4dev! Basically LGTM, but just a couple of follow-ups. PTAL.

Also, I added some clarity to your question on why the new tests you added in the most recent changes are really important (though it's in a resolved conversation thread).

Please also update the branch changes now that #5461 has been merged.

.github/workflows/code_coverage.yml
Rd4dev: Actually, this doesn't yet handle the case for proto - (it would be PROTO with the future pr (for ci) (upload comment), so when proto is declared, (with the current implementation alone) it would stop / exit after storing the protos without doing any status check, (this was done because I don't want to stop execution and creation of reports if one of the coverage reports is a failure case). I wanted the check to happen only at the very end. 

The ci would have proper failure with status check only if md report is used as of now.
(with proto it would skip coverage status check)

So locally -> we collect all protos, pass them to report generation and once everything is done, we do a status check and if pass : print a success statement else throw error message (while having all report generated)

With ci -> we collect all protos and return without executing the coverage reporter or status checks, - (with upcoming pr we upload and download the artifacts then finally pass the list of pb files to the coverage reporter script to generate the reports and then do a status check, so we have the report and output the status as pass or fail with providing the generated report)

Does that clarify things? or did I confuse?

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: They do exist but with additional checks for IncompatibleSources, I thought it would be better to have them in when cases so refactored them, 

[**RunCoverage.kt**](https://github.com/oppia/oppia-android/pull/5465/files#diff-62c81482e7c4b720da2d4efc73a48d10369eee097e0bf3439def6648ff150c23R188)
```
return when {
      exemption?.testFileNotRequired == true -> {
        CoverageReport.newBuilder()
          .setExemption(
            CoverageExemption.newBuilder()
              .setFilePath(filePath)
              .setExemptionReason(
                "This file is exempted from having a test file; " +
                  "skipping coverage check."
              )
              .build()
          ).build()
      }
      exemption?.sourceFileIsIncompatibleWithCodeCoverage == true -> {
        CoverageReport.newBuilder()
          .setExemption(
            CoverageExemption.newBuilder()
              .setFilePath(filePath)
              .setExemptionReason(
                "This file is incompatible with code coverage tooling; " +
                  "skipping coverage check."
              )
              .build()
          ).build()
      }
      else -> {
        val testFilePaths = findTestFiles(rootDirectory, repoRoot, filePath)
        when {}
     }
....
}
```

.github/workflows/code_coverage.yml
Rd4dev: Done, removed it.

Rd4dev:
@BenHenning, addressed the review comments, but not sure if I was clear enough on explaining on why proto is not used (for now), can you PTAL and see if that was well communicated?

.github/workflows/code_coverage.yml
BenHenning: Ah I see--this seems reasonable then until the next PR is merged. Thanks for clarifying!

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Ah thanks for clarifying!

BenHenning (commented):
Thanks @Rd4dev. This LGTM!

BenHenning (approved):

BenHenning:
Going ahead and merging this since everything's resolved.

scripts/src/java/org/oppia/android/scripts/ci/ComputeChangedFiles.kt
Rd4dev: Ahh got it! Thank you so much for taking time and explaining it.

PR #5466, opened 2024-07-23
Fix #2711 & part of #5343: Workflow cancellation support in CI using concurrency

### LINKED ISSUE #2711: Workflow canceller doesn't seem to work for dynamic jobs

Rd4dev:
## Explanation

Fix #2711 
Fixes part of #5343 

### Project
[PR 2.5 of Project 4.1]

### Changes Made
- The existing workflow canceller utilized the [cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) by syfle.
- Since GitHub introduced native [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) support on April 19, 2021, we can now utilize this built-in feature for better integration.
- This implementation leverages the [concurrency property](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-and-the-default-behavior) to manage workflows, using the configuration:
```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```
- This setup groups workflows by their name and reference, allowing GitHub to automatically cancel any in-progress jobs in the workflow when a new workflow run is triggered.
- This retires the existing Automatic workflow canceller (workflow_canceller.yml).
- This approach is now in place for the following workflows: 
  - Build Tests 
  - Static Checks 
  - Unit Tests (Robolectric - Gradle)
  - Unit Tests (Robolectric - Bazel)

### Triggered Cancellation
(PR Check Triggers) -> https://github.com/oppia/oppia-android/pull/5466/checks

### Screenshots of workflow cancelled
**Workflows Running:**

Screenshot (1536)

**Workflows in previous run cancelling after latest commit**

Screenshot (1538)

**Note:** It is observed that the Non-App Module runs take a bit of time to cancel but they eventually get cancelled.

image

image

## Essential Checklist


Rd4dev:
@adhiamboperes @BenHenning this is a part of #5343 (but a standalone fix), can you PTAL.

adhiamboperes (approved):
Thanks @Rd4dev! This looks good to me.

Seperately, do you have insights on why the non-app module workflows take long to cancel?

Rd4dev:
> Thanks @Rd4dev! This looks good to me.
> 
> Seperately, do you have insights on why the non-app module workflows take long to cancel?

Thanks @adhiamboperes, 

Reg Non-app module cancellation delays, I haven't really dug into why it takes a lot of time. However, based on observation, it predominantly occurs during the Non-app modules' **Build App** step, which seems to wait for certain tasks to finish before the cancellation trigger is hit or recognized. 

I may try to trigger the cancellation once the Build App step is complete to see if other stages cancel immediately without delay (which I believe is the case based on a previous observation, but I will confirm with another set of runs to identify all causes).

Rd4dev:
@adhiamboperes, based on the observation it seems that every job step that uses the expression `if: ${{ always() }}` exhibit the delay in their cancellation.

Since, the expression [`always`](https://docs.github.com/en/actions/learn-github-actions/expressions#always) 
> Causes the step to always execute, and returns true, even when canceled.

### Example case:
with [Static Checks](https://github.com/oppia/oppia-android/blob/develop/.github/workflows/static_checks.yml) workflow we have:

```yml
jobs:
  linters:
    name: Lint Tests
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2

      .............

      - name: Java lint check
        run: |
          bash /home/runner/work/oppia-android/oppia-android/scripts/checkstyle_lint_check.sh $HOME

      - name: Kotlin lint check
        run: |
          bash /home/runner/work/oppia-android/oppia-android/scripts/ktlint_lint_check.sh $HOME

      - name: Feature flag checks
        run: |
          bash /home/runner/work/oppia-android/oppia-android/scripts/feature_flags_check.sh $HOME

      .............

  script_checks:
    name: Script Checks
    runs-on: ubuntu-20.04
    env:
      CACHE_DIRECTORY: ~/.bazel_cache
    steps:
      - uses: actions/checkout@v2

      .............

      - name: Regex Patterns Validation Check
        if: always()
        run: |
          bazel run //scripts:regex_pattern_validation_check -- $(pwd)

      - name: XML Syntax Validation Check
        if: always()
        run: |
          bazel run //scripts:xml_syntax_check -- $(pwd)

      - name: Testfile Presence Check
        if: always()
        run: |
          bazel run //scripts:test_file_check -- $(pwd)

      - name: Accessibility label Check
        if: always()
        run: |
          bazel run //scripts:accessibility_label_check -- $(pwd) scripts/assets/accessibility_label_exemptions.pb app/src/main/AndroidManifest.xml

      - name: KDoc Validation Check
        if: always()
        run: |
          bazel run //scripts:kdoc_validity_check -- $(pwd) scripts/assets/kdoc_validity_exemptions.pb

      .............

```

### Screenshot:
[[Stacktrace]](https://github.com/oppia/oppia-android/actions/runs/10130346409/job/28011558544?pr=5466)

Screenshot (1546)


The **Lint Tests** steps didn't have the expression `if: ${{ always() }}` causing it to stop immediately when cancelled.

With the **Script Checks** steps having the `if: ${{ always() }}` expression set, makes the following steps to execute even after getting cancelled. And only after a certain time they get hit by the cancel trigger and actually stop their execution.

I will investigate further to see if there could be any workaround to stop them immediately and update them.

---

**Edit:**
### Workaround

Seems like there is a workaround and [as per docs](https://docs.github.com/en/actions/learn-github-actions/expressions#always), it is suggested to use the expression **`if: ${{ !cancelled() }}`**, using this allows it to still run a job or step regardless of its success or failure while allowing it to get cancelled. So, I replaced all the `if: ${{ always() }}` with `if: ${{ !cancelled() }}` and also confirmed that this doesn't interrupt the job still being called even if it failed (with the unit test workflow).

> Warning: Avoid using always for any task that could suffer from a critical failure, for example: getting sources, otherwise the workflow may hang until it times out. If you want to run a job or step regardless of its success or failure, use the recommended alternative: if: ${{ !cancelled() }}

[[Related Discussion]](https://github.com/orgs/community/discussions/26303#discussion-4252664)

Screenshot (1552)

Thanks for highlighting the issue, with this now the workflows get cancelled immediately when a new run is triggered.

Screenshot (1558)

.github/workflows/static_checks.yml
```diff
@@ -151,39 +155,39 @@ jobs:
         shell: bash
 
       - name: Regex Patterns Validation Check
-        if: always()
+        if: ${{ !cancelled() }}
```
BenHenning, line 160: We should probably have a comment for each of these lines explaining why it's needed (that it's basically a cancellation-compliant alternative to "if always()" per your findings).

BenHenning (changes requested):
This is really cool to see @Rd4dev and it looks much more robust than using an external canceler.

I had one in-line comment, but I'm also curious about one thing: per your PR description, it doesn't actually seem like https://github.com/styfle/cancel-workflow-action is being used or am I missing something? It seems that we're relying on GitHub Actions itself to handle cancellation here.

Rd4dev:
>  but I'm also curious about one thing: per your PR description, it doesn't actually seem like https://github.com/styfle/cancel-workflow-action is being used or am I missing something?

Uhm.. Sorry if the description was unclear. And I am not sure which one was mentioned here.

If that was the PR description mentioning - "The existing workflow canceller utilized the [cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) by syfle."

By that I was mentioning that the currently existing [workflow_canceller.yml](https://github.com/oppia/oppia-android/blob/develop/.github/workflows/workflow_canceller.yml#L20) workflow file is using the [styfle/cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) (to just reference the existing cancellation method) which is now being replaced with GitHub's concurrency. Am I pointing the same or have I mentioned it incorrectly elsewhere?

.github/workflows/static_checks.yml
Rd4dev: Sure, added the comment - 
```
# The expression if: ${{ !cancelled() }} runs a job or step regardless of its success or failure while responding to cancellations, 
serving as a cancellation-compliant alternative to if: ${{ always() }} in concurrent workflows.
```
for most of the occurrences, while updating the comments that existed as:
```diff
-if: ${{ always() }} # IMPORTANT: Upload reports regardless of status
+if: ${{ !cancelled() }} # IMPORTANT: Upload reports regardless of success or failure status
```

Would these phrasing be clear enough or would it require additional reference to any discussions?

Rd4dev:
Thanks for the review, @BenHenning. I've added the suggested in-line comments to each instance, can you PTAL.

.github/workflows/static_checks.yml
BenHenning: These seem really good--thanks!

BenHenning (approved):
Thanks @Rd4dev! This LGTM. Super great to see this fixed!

> > but I'm also curious about one thing: per your PR description, it doesn't actually seem like https://github.com/styfle/cancel-workflow-action is being used or am I missing something?
> 
> Uhm.. Sorry if the description was unclear. And I am not sure which one was mentioned here.
> 
> If that was the PR description mentioning - "The existing workflow canceller utilized the [cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) by syfle."
> 
> By that I was mentioning that the currently existing [workflow_canceller.yml](https://github.com/oppia/oppia-android/blob/develop/.github/workflows/workflow_canceller.yml#L20) workflow file is using the [styfle/cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) (to just reference the existing cancellation method) which is now being replaced with GitHub's concurrency. Am I pointing the same or have I mentioned it incorrectly elsewhere?

Ah, I think I simply misunderstood when reading your PR description. Thanks for clarifying!

BenHenning:
Going ahead and merging this since it shouldn't have upstream dependencies, and it'd be nice to immediately begin benefiting from the fix. :)

PR #5480, opened 2024-08-08
Fix part of #5343: Exempt targets incompatible with code coverage

BenHenning:
## Explanation
Fixes part of #5343

This PR prepares for enabling code coverage support in CI by:
- Introducing a new source file test exemption property: incompatibility with code coverage (see below for more insight into why this is needed).
- Updating the existing test file exemption textproto to include all of the current source files whose corresponding tests do not run correctly with the recently introduced code coverage tooling.
- Adding several test declarations that were missed in previous PRs (future Bazel work may include a script that checks for test file declarations to try and reduce the likelihood of this happening in the future).

At a high-level, there are several reasons why a test may fail to run in coverage:
- It isn't modularized (this fails for reasons not yet fully understood--see investigation findings in #5343), though some unmodularized tests do seem to pass (looking at you, ``SpotlightFragmentTest``).
- Its source file is either missing from offline instrumentation, or it simply fails to be resolved. This either results in a runtime failure during the test, or the source file being missing from its coverage.dat file.
- Other JaCoCo runtime issues that may imply an incorrect instrumentation filter due to attempting to reinstrument an already instrumented dependency class. A multi-hour search for related problems and an analysis of code coverage support in both Bazel and rules_kotlin did not yield any additional insight into why this situation can happen.
- Test files that are incorrectly named or in the wrong location (this PR doesn't make an effort to move these).
- And yet other unknown failures that result in instrumentation failure or a failure to generate a coverage.dat file.

There are probably many investigations needed to fix all of these problems, so it makes sense to at least denylist all tests that that are for sure known to fail. This at least sets up the team to get some value from CI-run code coverage, though hard enforcement probably can't be enabled until these stability problems are addressed.

Note that a custom script was written for the purpose of actually performing this analysis, and that can be seen here:

```kotlin
package org.oppia.android.scripts.coverage

import com.google.protobuf.TextFormat
import org.oppia.android.scripts.common.CommandExecutor
import org.oppia.android.scripts.common.CommandExecutorImpl
import org.oppia.android.scripts.common.RepositoryFile
import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
import org.oppia.android.scripts.proto.TestFileExemptions
import java.io.File
import java.io.FileInputStream
import java.util.concurrent.TimeUnit

  val repoRoot = File(args[0]).absoluteFile.normalize()
  // Skip non-Kotlin files, tests, this file, and already exempted files.
  val testFileExemptionList = loadTestFileExemptionsProto()
    .testFileExemptionList
    .filter { it.testFileNotRequired || it.sourceFileIsIncompatibleWithCodeCoverage }
    .mapTo(mutableSetOf()) { it.exemptedFilePath }
  val sourceFiles = RepositoryFile.collectSearchFiles(repoPath = repoRoot.path, expectedExtension = ".kt").filterNot { it.nameWithoutExtension.endsWith("Test") }.filterNot { it.nameWithoutExtension == "RunCoverageForAll" }.distinct()
  val nonExemptedSourceFiles = sourceFiles.filter { it.toRelativeString(repoRoot) !in testFileExemptionList }.sorted()
  println("Found ${sourceFiles.size} source files on which to run tests, but only ${nonExemptedSourceFiles.size} need to be checked (the others should be exempted).")
  val startTime = System.currentTimeMillis()
  val coverageResults = ScriptBackgroundCoroutineDispatcher().use { scriptBgDispatcher ->
    val commandExecutor = CommandExecutorImpl(scriptBgDispatcher, processTimeout = 10, processTimeoutUnit = TimeUnit.MINUTES)
    nonExemptedSourceFiles.withIndex().associate { (index, sourceFile) ->
      val sourcePath = sourceFile.toRelativeString(repoRoot)
      val timeSpentMs = (System.currentTimeMillis() - startTime).toFloat()
      val averageTimeMs = timeSpentMs / index.toFloat()
      val estRemainingTimeMs = averageTimeMs * (nonExemptedSourceFiles.size - index)
      if (index > 0) {
        print("- $sourcePath (${index + 1}/${nonExemptedSourceFiles.size})...~${estRemainingTimeMs / 60000.0f}m left")
      } else print("- $sourcePath (${index + 1}/${nonExemptedSourceFiles.size})...")
      sourcePath to commandExecutor.runCoverageForSourceFile(repoRoot, sourceFile).also { result ->
        println("\r- $sourcePath (${index + 1}/${nonExemptedSourceFiles.size}) -> ${result.summary}")
      }
    }
  }
  println()

  println("Found coverage results for ${coverageResults.size} files. High-level results:")
  val exemptions = coverageResults.values.filter { it == RunCoverageResult.Exempted }
  val failures = coverageResults.filter { (_, result) -> result.isFailure }
  val successes = coverageResults.values.filterIsInstance<RunCoverageResult.RanWithCoverage>()
  val averageCoverageRatio = successes.map { it.ratio }.average()
  println("- ${exemptions.size} source files had additional exemptions for code coverage.")
  println("- ${failures.size} source files failed to compute coverage. By category:")
  failures.values.groupBy { it.summary }.forEach { (summary, failuresGroup) ->
    println("  - ${summary.replaceFirstChar { if (it.isLowerCase()) it.titlecase() else it.toString() }}: ${failuresGroup.size}/${failures.size} failures.")
  }
  println("- Average coverage precentage is (for ${successes.size}/${coverageResults.size} passing tests): ${(averageCoverageRatio * 100.0).toFloat()}%.")
  println()

  println("Recomputed test file exemptions textproto:")
  println()
  println()
  val newExemptions = failures.map { (sourcePath, _) ->
    TestFileExemptions.TestFileExemption.newBuilder().apply {
      this.exemptedFilePath = sourcePath
      this.sourceFileIsIncompatibleWithCodeCoverage = true
    }.build()
  }
  val oldExemptionsProto = loadTestFileExemptionsProto()
  val allExemptions = (oldExemptionsProto.testFileExemptionList + newExemptions).sortedBy { it.exemptedFilePath }
  val updatedExemptionsProto = oldExemptionsProto.toBuilder().apply {
    clearTestFileExemption()
    addAllTestFileExemption(allExemptions)
  }.build()
  TextFormat.printer().print(updatedExemptionsProto, System.out)
  println()
}

private fun CommandExecutor.runCoverageForSourceFile(repoRoot: File, sourceFile: File): RunCoverageResult {
  val result = executeCommand(repoRoot, "bazel", "run", "//scripts:run_coverage", repoRoot.path, sourceFile.toRelativeString(repoRoot), "format=MARKDOWN")
  val outputLines = result.output
  return when {
    result.exitCode == 1 && outputLines.any { "Coverage data not found for the file" in it } -> RunCoverageResult.CoverageNotGenerated
    result.exitCode == 1 && outputLines.any { "No appropriate test file found" in it } -> RunCoverageResult.MissingTestFile
    result.exitCode == 1 && outputLines.any { "Cannot process instrumented class" in it } -> RunCoverageResult.FailsToInstrument
    result.exitCode == 1 && outputLines.any { "tried to access private method" in it } -> RunCoverageResult.CannotAccessJacocoInit
    result.exitCode == 0 && outputLines.any { it.startsWith("This file is exempted from having a test file") } -> RunCoverageResult.Exempted
    result.exitCode == 0 && outputLines.any { it.startsWith("Computed Coverage Ratio") } -> {
      val computedRatioLine = outputLines.single { it.startsWith("Computed Coverage Ratio") }
      val ratioStr = computedRatioLine.substringAfter(":").trim()
      RunCoverageResult.RanWithCoverage(ratioStr.toDouble())
    }
    else -> {
      println()
      println("Failed to run command (exit code ${result.exitCode}):")
      println("  ${result.command.joinToString(separator = " ")}")
      println()
      println("Output:")
      println("#".repeat(80))
      println()
      result.output.forEach(::println)
      println()
      println("#".repeat(80))
      println()
      error("Failed.")
    }
  }
}

private fun loadTestFileExemptionsProto(): TestFileExemptions {
  val protoBinaryFile = File("scripts/assets/test_file_exemptions.pb")
  val builder = TestFileExemptions.getDefaultInstance().newBuilderForType()

  // This cast is type-safe since proto guarantees type consistency from mergeFrom(),
  // and this method is bounded by the generic type T.
  @Suppress("UNCHECKED_CAST")
  val protoObj: TestFileExemptions =
    FileInputStream(protoBinaryFile).use {
      builder.mergeFrom(it)
    }.build() as TestFileExemptions
  return protoObj
}

private sealed class RunCoverageResult {
  abstract val isFailure: Boolean
  abstract val summary: String

  object Exempted : RunCoverageResult() {
    override val isFailure = false
    override val summary = "exempted"
  }

  // Seems to happen to some unmodularized tests.
  object CoverageNotGenerated : RunCoverageResult() {
    override val isFailure = true
    override val summary = "coverage.dat was not generated"
  }

  // Test files needs to be added, renamed, or moved.
  object MissingTestFile : RunCoverageResult() {
    override val isFailure = true
    override val summary = "missing test file"
  }

  // Seems to happen to some unmodularized tests.
  object FailsToInstrument : RunCoverageResult() {
    override val isFailure = true
    override val summary = "fails to instrument"
  }

  // Unclear what the root cause is for these cases; possibly an incorrect instrumentation_filter is
  // being used.
  object CannotAccessJacocoInit : RunCoverageResult() {
    override val isFailure = true
    override val summary = "cannot access private method jacocoInit in unnamed module"
  }

  data class RanWithCoverage(val ratio: Double) : RunCoverageResult() {
    override val isFailure = false
    override val summary: String get() = "coverage ran (${(ratio * 100.0).toFloat()}%)"
  }
}
```

With its corresponding library Bazel declaration:

```bazel
kt_jvm_library(
    name = "run_coverage_for_all_lib",
    testonly = True,
    srcs = [
        "RunCoverageForAll.kt",
    ],
    visibility = ["//scripts:oppia_script_binary_visibility"],
    deps = [
        "//scripts/src/java/org/oppia/android/scripts/common:repository_file",
        "//scripts/src/java/org/oppia/android/scripts/proto:script_exemptions_java_proto",
    ],
)
```

and its binary declaration:

```bazel
kt_jvm_binary(
    name = "run_coverage_for_all",
    testonly = True,
    data = TEST_FILE_EXEMPTION_ASSETS,
    main_class = "org.oppia.android.scripts.coverage.RunCoverageForAllKt",
    runtime_deps = ["//scripts/src/java/org/oppia/android/scripts/coverage:run_coverage_for_all_lib"],
)
```

This script was a significant help in simplifying the work to determining the list of exempted files. It also seems stable: running it with the latest exemption list correctly only runs valid tests.

The script also produced some high-level outputs which are interesting to note:
- 1274 total source files exist for potential code coverage (excludes non-Kotlin files, tests, and the ``RunCoverageForAll`` script itself).
- After omitting exemptions (i.e. those tests which the team has exempted from requiring test files), 387 remaining source files were expected to have 1 or more corresponding tests.
- Of those, 230 failed to have code coverage computed (and thus needed to be exempted in this PR). Here's the failure category break-down:
  - 115/230 failures occurred from coverage.dat missing after running ``bazel coverage``.
  - 32/230 failures occurred from the test file not actually existing where it was expected to be.
  - 81/230 failures occurred from a failure to instrument one or more of the classes needed for runtime.
  - 2/230 failures occurred from a JaCoCo access issue (likely indicating some significant build configuration issue somewhere in the toolchain).
- The average code coverage among the remaining 157 properly passing source files was: 85.15078%. This isn't fully representative since ~60% of source files couldn't be properly tested for coverage, but it's still much higher than expected.

**Note to reviewers**: ``scripts/assets/test_file_exemptions.textproto`` was automatically generated using the script above, and it was sorted by file path to make it a bit cleaner.

Finally, please note that ``DirectoryManagementUtilTest`` had one source file change due to a compile-time build warning that wasn't caught before (since the test wasn't being built in Bazel with Kotlin warnings-as-errors enabled).

## Essential Checklist

## For UI-specific PRs only
N/A -- This is only affecting build and test infrastructure.

### DIFF
--- a/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
+++ b/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
@@ -133,16 +133,15 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor: 
   /**
    * Runs code coverage for the specified Bazel test target.
    *
-   * An empty list being returned typically occurs when the coverage command fails to generate any
-   * 'coverage.dat' file. This can happen due to tests failures or a misconfiguration that prevents
-   * the coverage data from being properly generated.
+   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
+   * This can happen due to: Test failures or misconfigurations that prevent the coverage data
+   * from being generated properly.
    *
    * @param bazelTestTarget Bazel test target for which code coverage will be run
-   * @return the generated coverage data as a list of list of strings (since there may be more than
-   *     one file corresponding to a single test target, e.g. in the case of a sharded test), or an
-   *     empty list if no coverage data was found while running the test
+   * @return the generated coverage data as a list of strings
+   *     or null if the coverage data file could not be parsed
    */
-  fun runCoverageForTestTarget(bazelTestTarget: String): List<List<String>> {
+  fun runCoverageForTestTarget(bazelTestTarget: String): List<String>? {
     val instrumentation = bazelTestTarget.split(":")[0]
     val computeInstrumentation = instrumentation.split("/").let { "//${it[2]}/..." }
     val coverageCommandOutputLines = executeBazelCommand(
@@ -150,21 +149,22 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor: 
       bazelTestTarget,
       "--instrumentation_filter=$computeInstrumentation"
     )
-    return parseCoverageDataFilePath(bazelTestTarget, coverageCommandOutputLines).map { path ->
+    return parseCoverageDataFilePath(coverageCommandOutputLines)?.let { path ->
       File(path).readLines()
     }
   }
 
-  private fun parseCoverageDataFilePath(
-    bazelTestTarget: String,
-    coverageCommandOutputLines: List<String>
-  ): List<String> {
-    // TODO: RD to upstream this & update tests accordingly.
-    // Use the test target as the base path for the generated coverage.dat file since the test
-    // itself may output lines that look like the coverage.dat line (such as in BazelClientTest).
-    val targetBasePath = bazelTestTarget.removePrefix("//").replace(':', '/')
-    val coverageDatRegex = "^.+?testlogs/$targetBasePath/[^/]*?/?coverage\\.dat$".toRegex()
-    return coverageCommandOutputLines.filter(coverageDatRegex::matches).map(String::trim)
+  private fun parseCoverageDataFilePath(coverageCommandOutputLines: List<String>): String? {
+    val regex = ".*coverage\\.dat$".toRegex()
+    for (line in coverageCommandOutputLines) {
+      val match = regex.find(line)
+      val extractedPath = match?.value?.substringAfterLast(",")?.trim()
+      if (extractedPath != null) {
+        println("Raw Coverage Data: $extractedPath")
+        return extractedPath
+      }
+    }
+    return null
   }
 
   /**
--- a/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
+++ b/scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
@@ -133,15 +133,16 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor: 
   /**
    * Runs code coverage for the specified Bazel test target.
    *
-   * Null return typically occurs when the coverage command fails to generate the 'coverage.dat' file
-   * This can happen due to: Test failures or misconfigurations that prevent the coverage data
-   * from being generated properly.
+   * An empty list being returned typically occurs when the coverage command fails to generate any
+   * 'coverage.dat' file. This can happen due to tests failures or a misconfiguration that prevents
+   * the coverage data from being properly generated.
    *
    * @param bazelTestTarget Bazel test target for which code coverage will be run
-   * @return the generated coverage data as a list of strings
-   *     or null if the coverage data file could not be parsed
+   * @return the generated coverage data as a list of list of strings (since there may be more than
+   *     one file corresponding to a single test target, e.g. in the case of a sharded test), or an
+   *     empty list if no coverage data was found while running the test
    */
-  fun runCoverageForTestTarget(bazelTestTarget: String): List<String>? {
+  fun runCoverageForTestTarget(bazelTestTarget: String): List<List<String>> {
     val instrumentation = bazelTestTarget.split(":")[0]
     val computeInstrumentation = instrumentation.split("/").let { "//${it[2]}/..." }
     val coverageCommandOutputLines = executeBazelCommand(
@@ -149,22 +150,21 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor: 
       bazelTestTarget,
       "--instrumentation_filter=$computeInstrumentation"
     )
-    return parseCoverageDataFilePath(coverageCommandOutputLines)?.let { path ->
+    return parseCoverageDataFilePath(bazelTestTarget, coverageCommandOutputLines).map { path ->
       File(path).readLines()
     }
   }
 
-  private fun parseCoverageDataFilePath(coverageCommandOutputLines: List<String>): String? {
-    val regex = ".*coverage\\.dat$".toRegex()
-    for (line in coverageCommandOutputLines) {
-      val match = regex.find(line)
-      val extractedPath = match?.value?.substringAfterLast(",")?.trim()
-      if (extractedPath != null) {
-        println("Raw Coverage Data: $extractedPath")
-        return extractedPath
-      }
-    }
-    return null
+  private fun parseCoverageDataFilePath(
+    bazelTestTarget: String,
+    coverageCommandOutputLines: List<String>
+  ): List<String> {
+    // TODO: RD to upstream this & update tests accordingly.
+    // Use the test target as the base path for the generated coverage.dat file since the test
+    // itself may output lines that look like the coverage.dat line (such as in BazelClientTest).
+    val targetBasePath = bazelTestTarget.removePrefix("//").replace(':', '/')
+    val coverageDatRegex = "^.+?testlogs/$targetBasePath/[^/]*?/?coverage\\.dat$".toRegex()
+    return coverageCommandOutputLines.filter(coverageDatRegex::matches).map(String::trim)
   }
 
   /**
--- a/scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
+++ b/scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
@@ -85,12 +85,12 @@ class RunCoverage(
   private val reportFormat: ReportFormat,
   private val reportOutputPath: String,
   private val commandExecutor: CommandExecutor,
-  private val scriptBgDispatcher: ScriptBackgroundCoroutineDispatcher
+  private val scriptBgDispatcher: ScriptBackgroundCoroutineDispatcher,
+  private val testFileExemptionTextProtoPath: String = "scripts/assets/test_file_exemptions.pb"
 ) {
   private val bazelClient by lazy { BazelClient(File(repoRoot), commandExecutor) }
 
   private val rootDirectory = File(repoRoot).absoluteFile
-  private val testFileExemptionTextProto = "scripts/assets/test_file_exemptions"
 
   /**
    * Executes coverage analysis for the specified file.
@@ -104,13 +104,20 @@ class RunCoverage(
    *     the file is exempted from having a test file, an empty list is returned
    */
   fun execute() {
-    val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptionTextProto)
-      .testFileExemptionList
-      .filter { it.testFileNotRequired || it.sourceFileIsIncompatibleWithCodeCoverage }
-      .map { it.exemptedFilePath }
+    val testFileExemptions = loadTestFileExemptionsProto(testFileExemptionTextProtoPath)
+    val filesNotNeedingTests =
+      testFileExemptions
+        .testFileExemptionList.filter { it.testFileNotRequired }.map { it.exemptedFilePath }
+    val filesIncompatibleWithCodeCoverage =
+      testFileExemptions
+        .testFileExemptionList
+        .filter { it.sourceFileIsIncompatibleWithCodeCoverage }
+        .map { it.exemptedFilePath }
 
-    if (filePath in testFileExemptionList) {
-      println("This file is exempted from having a test file; skipping coverage check.")
+    if (filePath in filesNotNeedingTests || filePath in filesIncompatibleWithCodeCoverage) {
+      if (filePath in filesIncompatibleWithCodeCoverage) {
+        println("This file is incompatible with code coverage tooling; skipping coverage check.")
+      } else println("This file is exempted from having a test file; skipping coverage check.")
     } else {
       val testFilePaths = findTestFiles(repoRoot, filePath)
       check(testFilePaths.isNotEmpty()) {
@@ -119,12 +126,7 @@ class RunCoverage(
 
       val testTargets = bazelClient.retrieveBazelTargets(testFilePaths)
 
-      // TODO: RD to add this check in an upstream PR & add a test for it.
-      check(testTargets.isNotEmpty()) {
-        "Missing test declaration(s) for existing test file(s): $testFilePaths."
-      }
-
-      val coverageReports = testTargets.flatMap { testTarget ->
+      val coverageReports = testTargets.map { testTarget ->
         CoverageRunner(rootDirectory, scriptBgDispatcher, commandExecutor)
           .retrieveCoverageDataForTestTarget(testTarget.removeSuffix(".kt"))
       }
@@ -223,8 +225,8 @@ private fun getReportOutputPath(
   return "$repoRoot/coverage_reports/$fileWithoutExtension/$defaultFilename"
 }
 
-private fun loadTestFileExemptionsProto(testFileExemptiontextProto: String): TestFileExemptions {
-  return File("$testFileExemptiontextProto.pb").inputStream().use { stream ->
+private fun loadTestFileExemptionsProto(testFileExemptionProtoPath: String): TestFileExemptions {
+  return File(testFileExemptionProtoPath).inputStream().use { stream ->
     TestFileExemptions.newBuilder().also { builder ->
       builder.mergeFrom(stream)
     }.build()
--- a/scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
+++ b/scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
@@ -106,7 +106,7 @@ class RunCoverage(
   fun execute() {
     val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptionTextProto)
       .testFileExemptionList
-      .filter { it.testFileNotRequired }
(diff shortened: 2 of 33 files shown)


scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
```diff
@@ -133,38 +133,38 @@ class BazelClient(private val rootDirectory: File, private val commandExecutor:
...
-        println("Raw Coverage Data: $extractedPath")
-        return extractedPath
-      }
-    }
-    return null
+  private fun parseCoverageDataFilePath(
+    bazelTestTarget: String,
+    coverageCommandOutputLines: List<String>
+  ): List<String> {
+    // TODO: RD to upstream this & update tests accordingly.
```
BenHenning: @Rd4dev can you please bring this into one of your PRs & add corresponding tests?

This properly fixes the fact that sharded tests can produce multiple coverage.dat files.

This should include all the changes in this PR in BazelClient.kt, BazelClientTest.kt, CoverageRunner.kt, CoverageRunnerTest.kt, and RunCoverage.kt (except for the exemption filter update since that's specific to this PR).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -106,20 +106,25 @@ class RunCoverage(
...
       val testFilePaths = findTestFiles(repoRoot, filePath)
       check(testFilePaths.isNotEmpty()) {
-        "No appropriate test file found for $filePath"
+        "No appropriate test file found for $filePath."
       }
 
       val testTargets = bazelClient.retrieveBazelTargets(testFilePaths)
 
-      val coverageReports = testTargets.map { testTarget ->
+      // TODO: RD to add this check in an upstream PR & add a test for it.
```
BenHenning: @Rd4dev can you please bring this into one of your PR & add a corresponding test?

It addresses the specific problem of preventing a failure in the single() call below if there's a valid test file for a given source file, but that test has no declared Bazel target.

scripts/src/java/org/oppia/android/scripts/proto/script_exemptions.proto
```diff
@@ -49,6 +49,11 @@ message TestFileExemptions {
 
       // Overrides the minimum coverage percent required for the given file.
       int32 override_min_coverage_percent_required = 3;
+
+      // A medium-term state to indicate that something about this source file's build config, or
+      // that of its test(s), are not compatible with the current code coverage tooling and thus
+      // should be ignored.
+      bool source_file_is_incompatible_with_code_coverage = 4;
```
BenHenning, line 58: @Rd4dev can you please file an issue to track fixing the tests exempted per the failures outlined in the PR description, and link that back here? I'll add a code TODO for that issue once it exists.

BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @adhiamboperes for codeowners and @Rd4dev for code coverage project interoperability.

Note that this PR has a few changes that @Rd4dev should bring into her PRs, so this PR will have a few additional clean-ups before it's ready to submit. I will ask for re-review once those are done, but please do review it before then if you can.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -106,20 +106,25 @@ class RunCoverage(
   fun execute() {
     val testFileExemptionList = loadTestFileExemptionsProto(testFileExemptionTextProto)
       .testFileExemptionList
-      .filter { it.testFileNotRequired }
+      .filter { it.testFileNotRequired || it.sourceFileIsIncompatibleWithCodeCoverage }
```
BenHenning: @Rd4dev can you please advise on how I can best test this? If there's a specific test that I can copy in order to verify that source files are correctly filtered when this exemption is enabled, that would be really helpful.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
```diff
@@ -106,20 +106,25 @@ class RunCoverage(
...
       .map { it.exemptedFilePath }
 
     if (filePath in testFileExemptionList) {
       println("This file is exempted from having a test file; skipping coverage check.")
```
BenHenning: Note to self: I should update this to have a different warning when exempted due to incompatibility (to be clearer and not incorrect since the actual reason for exemption is different in those cases).

scripts/src/java/org/oppia/android/scripts/proto/script_exemptions.proto
Rd4dev: Sure, just filed the tracking issue: #5481

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Until PR 2.1 (which this PR is tested on), I used the available `test_file_exemptions.textproto` data to test exempted files, it was with PR 2.2 I introduced providing **custom test exemption list** as a parameter to test them (as to verify overridden percentages).

I think may be having a custom list would be a way to test in this scenerio.

I added the implementation and the test case for exempting incompatible source files: [**[implementation code]**](https://github.com/Rd4dev/oppia-android/pull/7/commits/8c9cdd44b6b1b7f2fc7e67115c39fa6e72634aa4#diff-5a779ea4a8aa2caf5249ced27465f76b8ac19bed5cf250475f6033dfad464908R222), can you please take a look at the above test case and see if that could be appropriate?

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Woah, yeah! Thanks, Ben! I'll include the changes with the list of files in the PR 2.2 and update here once it's done.

adhiamboperes (commented):
Thanks @BenHenning, I took a pass and I don't have concerns that you haven't documented.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
Rd4dev: Thanks for the fixes @BenHenning, updated the PR 2.2 with all the modifications.

Links to specific parts: [CoverageRunner.kt](https://github.com/oppia/oppia-android/pull/5461/files#diff-e1015bdd1aafa12e3e95de3b408525733c702992881ece1d2399b57ded6a44fdR42), [RunCoverage.kt](https://github.com/oppia/oppia-android/pull/5461/files#diff-62c81482e7c4b720da2d4efc73a48d10369eee097e0bf3439def6648ff150c23R213), [BazelClient.kt](https://github.com/oppia/oppia-android/pull/5461/files#diff-1e798d9f6f6e6059ff0d747cc961c02e0ba9bf4bf588dfe9db710be512d0c63dR158) and added tests for the same - [BazelClientTest.kt - Shard Configured Tests](https://github.com/oppia/oppia-android/pull/5461/files#diff-e50d7ab1968e374641f7298e27dca62ec182a42dc06dac4c47857cd7526de00fR463), [RunCoverageTest.kt - Missing Test Declaration](https://github.com/oppia/oppia-android/pull/5461/files#diff-5a779ea4a8aa2caf5249ced27465f76b8ac19bed5cf250475f6033dfad464908R413).

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
Rd4dev: Thanks Ben, added them -  https://github.com/oppia/oppia-android/pull/5480/files/46a0b0471447c1be80ef77efa8bce04c05516e29#r1710529201

Rd4dev:
Thanks, @BenHenning. I've incorporated all the changes into #5461 and added tests for them. I've also tested this PR with a few incompatible sources, and they were exempted as expected. Other than adjusting the exemption message, Todo addition and adding the additional test case (everything you mentioned), this looks good. I can take another look and approve once the refactoring and test additions are done if needed.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Done in latest changes.

scripts/src/java/org/oppia/android/scripts/proto/script_exemptions.proto
BenHenning: Thanks! Added a TODO here.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Thanks! Removed the code now that it's part of your PR.

scripts/src/java/org/oppia/android/scripts/common/BazelClient.kt
BenHenning: Ah, thanks for moving it! Glad it's working. Went ahead and removed this code here since it's now part of your PR.

scripts/src/java/org/oppia/android/scripts/coverage/RunCoverage.kt
BenHenning: Thanks! I wasn't expecting an actual test to copy, but it gave me a good place to start on getting everything set up. Test changes are included in the latest changes.

BenHenning (commented):
Self-reviewed changes.

Thanks again @Rd4dev and @adhiamboperes for the reviews and follow-ups. The latest changes should be code complete from my perspective, so PTAL and let me know if any additional changes are needed.

adhiamboperes (approved):

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -262,7 +296,8 @@ class RunCoverageTest {
       ReportFormat.MARKDOWN,
       markdownOutputPath,
       longCommandExecutor,
-      scriptBgDispatcher
+      scriptBgDispatcher,
+      testFileExemptionTextProtoPath = createTestFileExemptionsProtoFile()
```
Rd4dev: Just for my understanding: Since we have a default value, is it necessary to pass an empty exception list, or can we skip it? For cases where we call the main function directly, the default file path is used. 

```
main(
  "${tempFolder.root}",
  filePath,
)
```
Or is it preferable to adhere strictly to isolated test values whenever possible?

**If this is blocking the merge, please resolve it, and I’ll revisit it later.**

BenHenning: Hmm I was thinking that it would be better to have a predictable exemptions file when running the tests by default, but it now occurs to me that there's two different default states depending on whether ``execute()`` is called directly or ``main()`` is used, so it doesn't really matter. Reworked this slightly to remove the overriding behavior and hopefully make it a bit more consistent and clear.

BenHenning (commented):
Self-reviewed latest changes.

Thanks @adhiamboperes & @Rd4dev for the review.

@Rd4dev I followed up with a change per your comment. PTAL and resolve the thread if it looks good to you. Otherwise, please let me know if any follow-up changes are needed.

I'm also enabling auto-merge on this PR since it seems like everything is addressed at this point, and it's needed for RD's downstream PRs.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: Thanks for clarifying, @BenHenning. This understanding will assist me with future implementations.

PR #5487, opened 2024-08-15
Fixes part of #4938: Use TranslationController as the source of truth for the audio language setting

BenHenning:
## Explanation
Fixes part of #4938

This primarily updates ``ProfileManagementController`` to use ``TranslationController`` as the source of truth for audio language (rather than using the audio language property stored within the ``Profile`` proto). This has some noteworthy advantages:
- It allows for proper translation setting fallback behaviors to be enabled for the audio language setting without needing to migrate UI code over to ``TranslationController``.
- It isolates changes to the domain layer (with one exception: reading the audio language setting now uses a new getter in ``ProfileManagementController`` rather than reading the profile directly).

Some peripheral changes were also needed as part of this:
- A bunch of tests needed to be disabled in Gradle now that different options UIs (including for reading text) may depend on ``TranslationController`` (which is only fully configured in Bazel builds).
- The ``Profile`` proto was updated to remove its audio language setting. This means that existing users will revert back to whichever default ``TranslationController`` decides for them (most likely English, but it depends on several factors). This is considered a reasonable regression since most users are unlikely to depend on the automatic audio language setting, and the app is currently in a beta state so regressions like this should be expected.
- French and Chinese were removed from the list of ``AudioLanguage``s since they are not currently supported by the Oppia Android app (per ``OppiaLanguage`` which, plus the configured supported language textproto files, determine for which languages the app is guaranteeing support).
- ``ProfileManagementControllerTest`` was updated to have much more thorough testing around audio language.
- ``TranslationController`` was updated to use a ``PersistentCacheStore`` backing for audio and written translation languages (in addition to app language which was already supported).
- As part of the previous change, ``TranslationController.updateAppLanguage()`` and its related tests were updated to verify the _previous_ language is returned, not the current (for consistency with voiceover and written translations).

Long-term, ``AudioLanguage`` should be removed (along with its corresponding functionality in ``ProfileManagementController``) in favor of using ``TranslationController`` and ``OppiaLocale`` as the bases for managing language functionality in the UI layer.

Note that the Gradle version of the app will have increased degraded functionality in the options menu due to no supported language configuration being included in the build for that app (see the corresponding comment thread in this PR for more context). This is considered a reasonable medium-term gap as developers ought to be using the Bazel build of the app, anyway, as the Gradle version has significant functional limitations for all aspects of language selection and management (due to the lack of language configuration).

## Essential Checklist

## For UI-specific PRs only

Options screen (without changes):

image

Options screen (with changes):

image

I also verified setting a profile audio setting (Portuguese) on a ``develop`` branch build and then upgrading to a build from this branch. There are no crashes or stability issues, and (per my device setup) the audio language does revert back to English as expected.

I also verified that, like before, the audio setting persists across app instances when changed and is distinct between multiple profiles.

### DIFF
--- a/app/src/main/java/org/oppia/android/app/options/OptionControlsViewModel.kt
+++ b/app/src/main/java/org/oppia/android/app/options/OptionControlsViewModel.kt
@@ -5,8 +5,8 @@ import androidx.databinding.ObservableField
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.Transformations
-import androidx.lifecycle.ViewModel
 import org.oppia.android.app.fragment.FragmentScope
+import org.oppia.android.app.model.AudioLanguage
 import org.oppia.android.app.model.OppiaLanguage
 import org.oppia.android.app.model.Profile
 import org.oppia.android.app.model.ProfileId
@@ -20,7 +20,8 @@ import org.oppia.android.util.data.DataProviders.Companion.combineWith
 import org.oppia.android.util.data.DataProviders.Companion.toLiveData
 import javax.inject.Inject
 
-/** [ViewModel] for [OptionsFragment]. */
+private const val OPTIONS_ITEM_VIEW_MODEL_APP_AUDIO_LANGUAGE_PROVIDER_ID =
+  "OPTIONS_ITEM_VIEW_MODEL_APP_AUDIO_LANGUAGE_PROVIDER_ID"
 private const val OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID =
   "OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID"
 
@@ -67,11 +68,14 @@ class OptionControlsViewModel @Inject constructor(
   }
 
   private fun createOptionsItemViewModelProvider(): DataProvider<List<OptionsItemViewModel>> {
+    val appAudioLangProvider =
+      translationController.getAppLanguage(profileId).combineWith(
+        profileManagementController.getAudioLanguage(profileId),
+        OPTIONS_ITEM_VIEW_MODEL_APP_AUDIO_LANGUAGE_PROVIDER_ID
+      ) { appLanguage, audioLanguage -> appLanguage to audioLanguage }
     return profileManagementController.getProfile(profileId).combineWith(
-      translationController.getAppLanguage(profileId),
-      OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID,
-      ::processViewModelList
-    )
+      appAudioLangProvider, OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID
+    ) { profile, (appLang, audioLang) -> processViewModelList(profile, appLang, audioLang) }
   }
 
   private fun processViewModelListsResult(
@@ -93,12 +97,13 @@ class OptionControlsViewModel @Inject constructor(
 
   private fun processViewModelList(
     profile: Profile,
-    oppiaLanguage: OppiaLanguage
+    appLanguage: OppiaLanguage,
+    audioLanguage: AudioLanguage
   ): List<OptionsItemViewModel> {
     return listOfNotNull(
       createReadingTextSizeViewModel(profile),
-      createAppLanguageViewModel(oppiaLanguage),
-      createAudioLanguageViewModel(profile)
+      createAppLanguageViewModel(appLanguage),
+      createAudioLanguageViewModel(audioLanguage)
     )
   }
 
@@ -117,12 +122,14 @@ class OptionControlsViewModel @Inject constructor(
     )
   }
 
-  private fun createAudioLanguageViewModel(profile: Profile): OptionsAudioLanguageViewModel {
+  private fun createAudioLanguageViewModel(
+    audioLanguage: AudioLanguage
+  ): OptionsAudioLanguageViewModel {
     return OptionsAudioLanguageViewModel(
       routeToAudioLanguageListListener,
       loadAudioLanguageListListener,
-      profile.audioLanguage,
-      resourceHandler.computeLocalizedDisplayName(profile.audioLanguage)
+      audioLanguage,
+      resourceHandler.computeLocalizedDisplayName(audioLanguage)
     )
   }
 }
--- a/domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
+++ b/domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
@@ -490,7 +490,7 @@ class TranslationControllerTest {
   }
 
   @Test
-  fun testUpdateAppLanguage_uninitializedToSystem_returnsDefaultSelection() {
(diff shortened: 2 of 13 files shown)


BenHenning:
@adhiamboperes can you please confirm whether this approach will work for unblocking this part of onboarding (per our discussion earlier in the week)? The PR description just needs to be finalized before this can go into review, but I wanted to check that it will actually work for you before doing that.

I also didn't test it with the app, only via unit tests.

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
```diff
@@ -551,7 +551,7 @@ class TranslationControllerTest {
     // The previous selection was English.
     val selection = monitorFactory.waitForNextSuccessfulResult(updateProvider)
     assertThat(selection.selectionTypeCase).isEqualTo(SELECTED_APP_LANGUAGE)
-    assertThat(selection.selectedLanguage).isEqualTo(BRAZILIAN_PORTUGUESE)
+    assertThat(selection.selectedLanguage).isEqualTo(ENGLISH)
```
adhiamboperes, line 554: I don't understand why this is english instead of portuguese.

adhiamboperes (changes requested):
Thanks for taking this up @BenHenning! I have left a comment which may be tied to the implementation. PTAL.

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
BenHenning: Voiceover and written translations, when updated, return the previous setting not the current. I don't know why app language was configured differently, so I updated it & its tests to be previous for consistency.

BenHenning (commented):
Self reviewed changes.

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
adhiamboperes: This is confusing from a usage perspective. So, If a user wants to change languages, what happens to the new selection, if the API returns the previous selection? I believe I'm missing something.

BenHenning:
I think this is now ready for full review. PTAL @adhiamboperes.

app/src/main/java/org/oppia/android/app/options/OptionControlsViewModel.kt
```diff
@@ -67,11 +68,14 @@ class OptionControlsViewModel @Inject constructor(
...
+        profileManagementController.getAudioLanguage(profileId),
+        OPTIONS_ITEM_VIEW_MODEL_APP_AUDIO_LANGUAGE_PROVIDER_ID
+      ) { appLanguage, audioLanguage -> appLanguage to audioLanguage }
     return profileManagementController.getProfile(profileId).combineWith(
-      translationController.getAppLanguage(profileId),
-      OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID,
-      ::processViewModelList
-    )
+      appAudioLangProvider, OPTIONS_ITEM_VIEW_MODEL_LIST_PROVIDER_ID
+    ) { profile, (appLang, audioLang) -> processViewModelList(profile, appLang, audioLang) }
```
adhiamboperes, line 78: There's an issue here which causes an exception to be thrown, and therefore the OptionsFragment is blank in Gradle builds. Despite the recent changes, OptionsFragment should still display options, and only the AudioLanguageSelectionFragment should be blank, similar to the behavior in App Language.

I suspect that this is because the language defaults to `LANGUAGE_UNSPECIFIED`, for which https://github.com/oppia/oppia-android/blob/f9106d91297abd09b24da25d5485deb8473b0125/domain/src/main/java/org/oppia/android/domain/locale/LocaleController.kt#L355 does not provide a reasonable fallback.

adhiamboperes: Screenshot_1723810848

domain/src/test/java/org/oppia/android/domain/translation/TranslationControllerTest.kt
adhiamboperes: Ok I have tested this, and it works as expected so I am resolving this comment.

adhiamboperes (changes requested):
Thanks @BenHenning, I have one final comment. PTAL.

app/src/main/java/org/oppia/android/app/options/OptionControlsViewModel.kt
BenHenning: It's definitely related to ``LANGUAGE_UNSPECIFIED``, but more broadly we don't have any languages to display in Gradle builds since there's no (not empty, but missing) languages configuration to load.

I know it's a bit unexpected for OptionsFragment to have this issue, but it's ultimately because its view models indirectly depend on the current language (probably to show the preview). It's actually quite difficult to set this up with error tolerance since data providers aren't really meant to be chained in that way (if one fails, the whole chain fast-fails rather than trying to recover). This failure happens very early in the pipeline (somewhere in TranslationController due to the missing supported languages configuration).

As I see it, we have four options here:
1. Ignore it (which means a degraded Gradle experience for developers).
2. Remove the snippet from OptionsFragment so that it doesn't have this dependency (functional regression for users, so not ideal).
3. Introduce new data provider chaining functionality that allows for errors to be caught and recovered. Adding new functions is extremely time consuming due to how carefully they need to be tested (which is why #4484 still isn't merged yet), and I'm not yet sure that we would actually want this behavior from a conceptual perspective.
4. Introduce fault tolerance in translation controller (or wherever it's hard failing with supported languages missing). 'Fault tolerance' here means either an error propagation (rather than an exception) or defaulting the supported languages proto (which may require other analysis since certain assumptions may be made about the structure, such as 'English' always being present).

I lean toward (1) for simplicity, but (4) is possibly better for the long-term (though it's not obvious to me if the Bazel builds benefit much from (4) since we have fairly strong guarantees and checks that the correct supported languages structure is included in different builds, and the 'fail fast' approach helps ensure this).

What are your thoughts?

adhiamboperes: I would go with (1) for quick resolution. Re: 4 "wherever it's hard failing"- I tracked down the failure to LocaleController(linked in my earlier comment)

adhiamboperes:
This LGTM, please feel free to merge this at your convenience.

app/src/main/java/org/oppia/android/app/options/OptionControlsViewModel.kt
BenHenning: Ah apologies, I missed that. I'm fine with no action here then, if you are (i.e. opting for (1) over the other choices).

BenHenning:
Thanks @adhiamboperes! I appreciate the review. :)

BenHenning:
Actually quick follow-up: @adhiamboperes could you perhaps approve this, too? :) I think you may have forgotten to with your previous 'LGTM' message.

adhiamboperes (approved):
Approved 😅

PR #5483, opened 2024-08-10
Fix #5486 & part of #5343: Introducing new wiki page for code coverage usage and limitations

Rd4dev:
## Explanation
Fixes part of #5343 
Fixes #5486 

### Project
[PR 2.6 of Project 4.1]

### Changes Made
- This PR introduces 2 new wiki pages:
  - Oppia Android Code Coverage
  - Writing tests with Good Behavioural Coverage
- Fix to the comment upload feature permission issues with PRs created from the forked branches.
  - Split the code coverage workflow into 2
    1.  Core code coverage workflow to handle 
        - Collection of changed files 
        - Bucket partitioning
        - Run coverage in matrices
    2.  Comment upload workflow to handle
        - evaluation of reports 
        - generation of md reports 
        - uploading comments
        - Coverage Status Checks
      (as the later required `pull_request_target`)
- Fix to #5486 
  - The issue should have arisen as the pr got merged with the branch being deleted while the publish comment job still running, finding it hard to fetch the pr-issue number to publish a comment.
  - Now the Coverage Check Status was made to be dependent on the comment uploader ie. the final Coverage check job occurs only after the Evaluation and Comment jobs are done, so it will always have the pr reference)
```
  check_coverage_results:
    name: Check Code Coverage Results
    needs: [ evaluate-code-coverage-reports, comment_coverage_report ]
```

#

### Reasons for splitting the code_coverage workflow 

The single code_coverage workflow was split into 
1. **code_coverage** (to run coverages)
2. **coverage_report** (to generate and publish reports)

### Separating the comment upload job

- The primary reason is the need to have ability to upload comments from PRs opened from a fork branch.
  - [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target): For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission, even when it is triggered from a fork.
  - Workflows triggered by `pull_request_target` events are run in the context of the base branch. Since the base branch is considered trusted, workflows triggered by these events will always run, regardless of approval settings. [[GitHub Docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks)]

### Separating the evaluation / generation of report job

[...]

Issue #5506, opened 2024-08-22
[BUG]: Code coverage fails to capture execution flow interruptions


Rd4dev:
### Describe the bug

It is observed that lines of code involving termination statements - such as `exitProcess(1)`, assertion failures - such as `assertThat(it).isNotPending()` and flow interruption statements - such as `break` are not marked as covered even though they are executed. 

- This issue is a tracking item of the Code Coverage Enforcement feature - https://github.com/oppia/oppia-android/issues/5343.

This appears to be a known issue with JaCoCo. According to JaCoCo documentation, source code lines that involve exceptions or abrupt termination sequences show no coverage because the control flow is interrupted. [JaCoCo FAQ Page](https://www.eclemma.org/jacoco/trunk/doc/faq.html)

As per documentation:

> Source code lines with exceptions show no coverage. Why?

> JaCoCo determines code execution with so called probes. Probes are inserted into the control flow at certain positions. Code is considered as executed when a subsequent probe has been executed. In case of exceptions such a sequence of instructions is aborted somewhere in the middle and the corresponding lines of source code are not marked as covered.

### Steps To Reproduce

1. **Termination statements:**

Consider the [RetrieveChangedFiles.kt](https://github.com/oppia/oppia-android/blob/f9106d91297abd09b24da25d5485deb8473b0125/scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt#L45) file with the following termination statement:

```kotlin
fun main(args: Array<String>) {
  if (args.size < 5) {
    println(
      "Usage: bazel run //scripts:retrieve_changed_files --" +
        " <encoded_proto_in_base64> <path_to_bucket_name_output_file>" +
        " <path_to_file_list_output_file> <path_to_test_target_list_output_file>"
    )
    println("Exiting...")
    exitProcess(1)
  }
  ...
}
```

Generate code coverage report for the file RetrieveChangedFiles.kt:

```sh
bazel run //scripts:run_coverage -- $(pwd) scripts/src/java/org/oppia/android/scripts/ci/RetrieveChangedFiles.kt
```

Review the generated HTML code coverage report for RetrieveChangedFiles.kt:

![image](https://github.com/user-attachments/assets/bf5cd35f-4bbd-4c68-a886-aa0262c6b7f7)

2. **Assertion failures:**

[...]

PR #5511, opened 2024-08-24
Fix part of #5343: Update Incorrect Link for the Oppia Android Code Coverage Wiki Page

Rd4dev:
## Explanation

Fixes Part of #5343 

### This PR includes

- Updated the incorrect link reference in the #5483 for the Oppia Android Code Coverage Page
  - Current Broken link: https://github.com/oppia/oppia-android-workflow/wiki/Oppia-Android-Code-Coverage
  - Correct link to wiki:  https://github.com/oppia/oppia-android/wiki/Oppia-Android-Code-Coverage
- Fixed an incorrect SKIP status that triggered even after 'Unit Tests - Bazel' failure causing it to miscalculate the pb file list as zero thereby posting a skip comment.
  - Solved by adding an additional condition for the evaluation job.
```
if: ${{ !cancelled() && needs.check_unit_tests_completed.result == 'success'}}
```
- The code coverage comment got triggered even after assignment changes due to the presence of 'assigned' in the on `pull_request_target` triggered, fixed it by removing the 'assigned' trigger.
- Utilized 'HtmlEscapers' to escape HTML characters from source code lines in CoverageReporter.kt as strings those already included html tags overlapped / conflicted with the report html templates.
- Added limitations section to the 'Oppia-Android-Code-Coverage' wiki page
- Filed all possible coverage gaps that exist right now based on the available pass cases.

## Essential Checklist

## For UI-specific PRs only


Rd4dev:
@BenHenning, I incorrectly referenced the wrong link in PR #5483; updated to the correct link here, can you PTAL?

Oh, and also should we include other coverage gaps to the "[Limitations sections](https://github.com/oppia/oppia-android/wiki/Oppia-Android-Code-Coverage#limitations-of-the-code-coverage-tool)"?

BenHenning:
Thanks @Rd4dev. I think adding other limitations sounds like a good idea.

Rd4dev:
Seems like I also had included assigned to the triggers,

```
on:
  pull_request_target:
    types: [assigned, opened, synchronize, reopened]
```

comment posted on assignment: https://github.com/oppia/oppia-android/pull/5511#issuecomment-2308493626
and its commenting on every assignment too :|, will add that fix too here ...

---

### Coverage Gaps

1. MathExpressionParser.kt -> As I don't understand the entire functionality just wanted to confirm if this is a missing parameter declaration causing the coverage miss or a jaCoCo coverage gap?

The lines are

```kotlin
  fun parseAlgebraicEquation(
    rawExpression: String,
    allowedVariables: List<String>,
    errorCheckingMode: ErrorCheckingMode = ErrorCheckingMode.ALL_ERRORS
  ): MathParsingResult<MathEquation> {
    println("Hitting...")
    return createAlgebraicParser(
      rawExpression, isPartOfEquation = true, allowedVariables, errorCheckingMode
    ).parseGenericEquationGrammar().map { it.stripParseInfo() }
  }
``` 

The test is:

```kotlin
  @Test
  fun testParseAlgebraicEquation_basicEquation_doesNotFail() {
    expectSuccessWhenParsingAlgebraicEquation("y = 2x + 3")
  }

    private fun parseAlgebraicEquation(expression: String): MathParsingResult<MathEquation> {
      return MathExpressionParser.parseAlgebraicEquation(
        expression, allowedVariables = listOf("x", "y", "z"), ALL_ERRORS
      )
    }
```

The report says:

image

This was a similarly observed with methods: parseAlgebraicExpression() and parseNumericExpression().

#

2. **Abstract methods:**
Same as above I don't completely understand the inner working but the `TestCoroutineDispatcher.kt` has few abstract methods:

```kotlin
abstract fun hasPendingTasks(): Boolean

abstract fun getNextFutureTaskCompletionTimeMillis(timeMillis: Long): Long?

abstract fun runCurrent(
  timeout: Long = DEFAULT_TIMEOUT_SECONDS,
  timeoutUnit: TimeUnit = DEFAULT_TIMEOUT_UNIT
)
```

and these are implemented in _TestCorotineDispatcherRobolectricImpl.kt_, _TestCoroutineDispatchersRobolectricImpl.kt_, _TestCoroutineDispatchersEspressoImpl.kt_, etc.

But as per [jaCoCo documentation](https://www.jacoco.org/jacoco/trunk/doc/faq.html#:~:text=Abstract%20methods%20do%20not%20contain,non%2Ddefault%20methods%20in%20interfaces.):

>Abstract methods do not contain code, therefore code coverage cannot be evaluated. Indeed code coverage is recorded for subclasses implementing these methods

So, should the actual abstract methods in the TestCoroutineDispatcher.kt be uninstrumented as we only acquire coverage only from their respective test cases?

While most abstract methods are uninstrumented (highlighted as grey), the runCurrent shows as not covered (highlighted in red). Is this a coverage gap?

The html report:

image
image

#

3. **break:**

The finally block was reported as uncovered in `SurveyProgressController.kt`

```kotlin
    is ControllerMessage.FinishSurveySession -> {
      try {
        controllerState.completeSurveyImpl(message.callbackFlow)
      } finally {
        println("Finally...")
        // Ensure the actor ends since the session requires no further message processing.
        break
      }
    }
```

The report with break statement:

image

But I think this will also align with the same exception gap, and the presence of break statement causes the partial coverage and commenting the break statement covers the finally block:

```kotlin
    } finally {
      println("Finally...")
      // Ensure the actor ends since the session requires no further message processing.
     // break
    }
```

The report without break statement: 

image

If confirmed I will add it to the exception / termination statement #5506 issues.

#

4. **last curly brace:**

While this one is not observed in most cases, but with `CpuPerformanceSnapshotter.kt` the generated report highlights the last curly brace as uncovered:

The code:

```diff
    @OptIn(ObsoleteCoroutinesApi::class)
    private fun createCommandQueueActor(): SendChannel<CommandMessage> {
      return coroutineScope.actor(capacity = Channel.UNLIMITED) {
        for (message in channel) {
         ...
        }
-     }
    }
```

The generated report:

image

Similarly observed in CommandExecutorImpl.kt

The generated report:

image

also with TodoOpenCheck.kt

The generated report:

image

One related issue that can be found is: https://github.com/jacoco/jacoco/issues/1233

While I'm unsure if this is related to the test implementation or something that demands a configuration change or an actual coverage gap? If it is a coverage gap, then it will require further investigation on what specific cases they occur.

#

5. classes with just companion object

It is observed with `RepositoryFile.kt` that the class line was reported as uncovered, the generated report:

image

and I doubt if they are related to :
- https://www.jacoco.org/jacoco/trunk/doc/faq.html#:~:text=If%20execution%20data%20is%20missing,from%20instrumentation%20or%20not%20executed.
- https://stackoverflow.com/questions/42348077/class-name-is-marked-as-not-covered-while-running-code-coverage-in-android-stu
- https://stackoverflow.com/questions/46477696/code-coverage-does-not-reach-class-declaration

same with TodoCollector.kt

image

---

- There seems to be a timeout issue with my local machine run so will fix than and rerun and try updating the results for GenerateMavenDependenciesList.kt, MavenDependenciesListCheck,  BazelClient.kt (if possible).
- And I am even afraid to attempt running RunCoverage.kt, so I’m going to pass on that
- Tried ComputeAffectedTestsTest, but I think I am skipping that too.
If mandatory I can try running coverage on the above files later.

I think I ran coverages on most if not all of the Passing cases (not on exempted overridden files as most are reported as 0 or very low and figuring out gaps would be not straightforward) and based on the other reports it can be seen that most of the uncovered lines are based on unhit else or other branch cases, overrides and I tried adding tests to see if they can be hit for few of them and they do seem to get covered. So, I think once all filed and (the above-mentioned gaps) to be filed issues are addressed, we can achieve move towards 100% code coverage. 

While I’m not sure if we’ll uncover more gaps once the source file incompatibility issues are resolved, that’s something we’ll only know once it’s done. Can you take a look at the above cases and direct if those can be filed as coverage gap or if any are uncovered with missing tests.

Once everything is filed, I will add them to the wiki limitations in one go.

---

### Other issues (Not coverage gaps)

HTML reports with source code containing html tags / other tags mess up the reports. For eg: With CoverageReporter.kt, LiTagHandler.kt, TodoCollector.kt reports, since it already have tags as strings within them, when the report is generated it mix matches the report structure. One possible way is to move these html strings as templates to a separate place and refer them from there so the source code report generation would still preserve the format.

eg: chopped - TodoCollector.kt

image

formatted CoverageReport.kt

image

LiTagHandler.kt

image

BenHenning:
Thanks @Rd4dev. The code changes LGTM, but I'll need to process your last message a bit before I have any follow-up thoughts/suggestions.

BenHenning:
(1) is probably a real gap. It looks like the tests never verify the default parameter since they always provide one.

(2) is touching on several different things:

- We sometimes test abstract or interface classes as part of verifying their contracts (this is a case-by-case basis, and I don't think applies directly here).
- JaCoCo is making Java-specific assumptions (which are actually not correct anymore since interfaces in Java can have default methods). Its conclusion isn't really correct per modern Java or Kotlin.
- This is again, like (1), a case of default arguments seeming to not be verified so it's a real coverage gap on our end (assuming JaCoCo properly observes the coverage when it *is* tested).

(3) I'm not exactly sure if this is #5506 as-is, but I think it's reasonable to extend that issue to include this case. It definitely seems like a JaCoCo issue to me.

(4) Worth filing an issue for this case with the specific examples you've found. I suspect these are either coming from:

- A case where the code is simply not finishing in the test (in which case we need to add a properly configured test for each unit to verify that it "cleans up" correctly).
- Dead code being generated by the Kotlin compiler (similar to the Groovy problem mentioned in the JaCoCo issue that you linked).

Both will require investigation, and we need to rule out the first case before looking into the second one (which would require a bytecode analysis on the production code being instrumented).

(5) I suspect this is a case of the constructors not being called (since Kotlin generates public default constructors for classes) as we don't need to. If that's the case, we should just be turning the classes into public objects at the top-level if they don't need to be instances. This seems to be a coverage gap on our side, not a tooling issue.


-----

At a high level:
- I think skipping the tests you mentioned seems reasonable. The analysis you've performed so far looks very thorough and I'm happy to defer to your findings.
- It looks like one new issue should be filed (per (4)) and one issue updated (per (3)).
- The report generation can be fixed by escaping HTML tags (see https://stackoverflow.com/a/26572556/3689782 for a Guava-compatible way to do this without needing any additional dependencies). I think this is something that should be fixed, if possible, with a basic test in this PR.
- The wiki page should still be updated to include the coverage gaps once the issues above are filed.

Rd4dev:
@BenHenning, can you PTAL? and would it require any other verification on the Coverage Checks?

Ok, I found another case - https://github.com/oppia/oppia-android/pull/5522/checks, seems like evaluate runs irrespective of unit test failure and throwing "SKIP" status - https://github.com/oppia/oppia-android/pull/5522#issuecomment-2310730618, will also adress that over here.

Log:

- When Unit tests check fails now the Evaluate job gets skipped causing no misleading SKIP coverage report comment publication - https://github.com/oppia/oppia-android/actions/runs/10576551236/job/29304525595?pr=5511
- Added additional Unit test check pass to the Coverage Check Result job (to make sure coverage checks fails on unit test fail) - https://github.com/oppia/oppia-android/actions/runs/10576551236/job/29304525864?pr=5511
- Coverage SKIP to still get posted if no .kt files are changed - https://github.com/oppia/oppia-android/pull/5511#issuecomment-2312379269
- Comment being posted when everything goes well - https://github.com/oppia/oppia-android/pull/5511#issuecomment-2313034208

---

Thanks for the detailed information on each case. 
- I’ve filed a new issue as suggested #5523 and updated #5506 to include the relevant details. 
- I also applied the escape fix, and now the reports render right. For the testing I ended up using a direct conversion method
```kotlin
val unescapedOutputReportText = outputReportText
  .replace("&quot;", "\"")
  .replace("&amp;", "&")
  .replace("&lt;", "<")
  .replace("&gt;", ">")
```
not sure if there is anything to unescape (and escaping expected report caused issues with other tags).

Edit: <details><summary>Escaped CoverageReporter.html</summary>

image


</details>

- I've also added the newly filed issues to the limitations section of the wiki.

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
```diff
@@ -317,7 +317,13 @@ class RunCoverageTest {
 
     val expectedResult = getExpectedHtmlText(kotlinFilePath)
 
-    assertThat(readHtmlReport(kotlinFilePath)).isEqualTo(expectedResult)
+    val unescapedHtmlReport = readHtmlReport(kotlinFilePath)
```
BenHenning, line 320: Could you maybe use Guava to unescape the HTML here & for other tests in the PR?

BenHenning (changes requested):
Thanks @Rd4dev! Generally LGTM, but had one comment--PTAL.

BenHenning:
@Rd4dev can you also please make sure the PR description is updated to include all of the new changes (since there have been a few iterations on this PR?).

BenHenning:
Hmm also I think the coverage report comment is reposting after PR event changes like assignees--are its triggers correct?

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
Rd4dev: @BenHenning, Uhm... I'm not sure if Guava has methods to unescape the same?

[Guava Doc](https://guava.dev/releases/19.0/api/docs/com/google/common/escape/Escaper.html#escape(java.lang.String):~:text=Function%3CString,given%20literal%20string.)  |  [stackoverflow thread](https://stackoverflow.com/a/34091458)

or does it? or is there any other better way to do it than direct conversion.

Rd4dev:
> Hmm also I think the coverage report comment is reposting after PR event changes like assignees--are its triggers correct?

Yeah, that shouldn't be triggering and spotted it earlier https://github.com/oppia/oppia-android/pull/5511#issuecomment-2308497331 and included a fix for that too here - [PR Changes](https://github.com/oppia/oppia-android/pull/5511/files#diff-8666350cec48d3bb6effe6965f6d25893c7729dfecf890f6a29ace6411b9c7d0R10)

Ah, I’ve updated the PR description with the details—hopefully, they’ll stand out better now and won’t get lost in the mix.

@BenHenning, can you please confirm on this https://github.com/oppia/oppia-android/pull/5511#discussion_r1733655294 or other approaches that could be followed. 

#

Edit: 
Seems like [Apache Commons](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html#escapeHtml4-java.lang.String-) does have unescape functionalities, but I think that will require additional dependency?

scripts/src/javatests/org/oppia/android/scripts/coverage/RunCoverageTest.kt
BenHenning: Good call, I suppose this approach seems reasonable then. I'd rather not introduce another dependency if we can avoid it, even for tests.

BenHenning (approved):
Ah, I forgot about the fact that the old version of the comment workflow is running since it's pull_request_target--thanks for pointing out that it should now be fixed @Rd4dev.

Since there's nothing left to resolve here, approving and merging. Thanks!

PR #5525, opened 2024-09-04
Fix #1730 : Prevent binary files from being checked in using pre-commit hook

### LINKED ISSUE #1730: Prevent binary files from being checked in
Checking in binary files will significantly increase the repository size. We need a pre-commit check to prohibit binary files from being added, since we should never add them. In cases when binary files are needed, we should use Git LFS or an alternate solution.

Rd4dev:
## Explanation

Fixes #1730 

### The PR includes
- A pre-commit hook that identifies binary files in both 
  - staged changes (for local checks before committing)
  - committed files (for CI checks after pushing to a PR)
- If binary files are found, the commit/check is blocked, and the offending files are listed for removal.
- The pre-commit hook is integrated via a setup script.
- The same script is utilized for the CI pipeline.
- The 'Pass' statement is only included in the CI checks to keep the local commit process cleaner.

### Local block as pre-commit hook when a binary file is detected
image

### CI re-check if the PR includes a binary - [ Fail - [stack trace](https://github.com/Rd4dev/oppia-android/actions/runs/10715972847/job/29712474511#step:7:15) ] [ Pass - [stack trace](https://github.com/Rd4dev/oppia-android/actions/runs/10716040065/job/29712695824?pr=11#step:7:10) ]
image

## Essential Checklist

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -213,7 +213,7 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           bash /home/runner/work/oppia-android/oppia-android/scripts/pre-commit.sh
-          echo "No binary files found"
+          echo "No binary files found in commit"
           echo "BINARY FILES CHECK PASSED"
 
   # Note that caching is intentionally not enabled for this check since licenses should always be
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -107,6 +107,8 @@ jobs:
       CACHE_DIRECTORY: ~/.bazel_cache
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Set up Bazel
         uses: abhinavsingh/setup-bazel@v3
@@ -204,6 +206,16 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           bazel run //scripts:string_resource_validation_check -- $(pwd)
+
+      - name: Binary files check
+        run: |
+          bash /home/runner/work/oppia-android/oppia-android/scripts/pre-commit.sh
+          exit_code=$?
+
+          if [ $exit_code -eq 0 ]; then
+            echo "No binary files found"
+            echo "BINARY FILES CHECK PASSED"
+          fi
 
   # Note that caching is intentionally not enabled for this check since licenses should always be
   # verified without any potential influence from earlier builds (i.e. always from a clean build to
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -208,14 +208,13 @@ jobs:
           bazel run //scripts:string_resource_validation_check -- $(pwd)
 
       - name: Binary files check
+        # The expression if: ${{ !cancelled() }} runs a job or step regardless of its success or failure while responding to cancellations,
+        # serving as a cancellation-compliant alternative to if: ${{ always() }} in concurrent workflows.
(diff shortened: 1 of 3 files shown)


Rd4dev:
@BenHenning @adhiamboperes, would this be an appropriate way to prevent the binary files check in? Can you PTAL!

Also, this https://github.com/oppia/oppia-android/issues/1730#issuecomment-765098904 mentions

> We may run into an issue with the CI check if it runs across the entirety of develop since there are some binary files in mainline, and we can't force push change those. We may need to build in exceptions for those files.

What does that exactly mean?

.github/workflows/static_checks.yml
```diff
@@ -205,6 +207,16 @@ jobs:
         run: |
           bazel run //scripts:string_resource_validation_check -- $(pwd)
 
+      - name: Binary files check
```
Rd4dev, line 210: **Self-reminder:** Add if (!cancelled) check upon approval.

.github/workflows/static_checks.yml
```diff
@@ -205,6 +207,16 @@ jobs:
...
+        run: |
+          bash /home/runner/work/oppia-android/oppia-android/scripts/pre-commit.sh
+          exit_code=$?
+
+          if [ $exit_code -eq 0 ]; then
```
BenHenning: Should we fail this if the check fails? It seems like it always passes unless I'm missing something.

If the script failure already triggers the failure of the workflow, perhaps we can simplify this:

```sh
bash /home/runner/work/oppia-android/oppia-android/scripts/pre-commit.sh

# If the script above doesn't fail, then there are no binary files.
echo "No binary files found in commit."
```

.github/workflows/static_checks.yml
BenHenning: Why not do that now?

BenHenning (changes requested):
Thanks @Rd4dev!

> @BenHenning @adhiamboperes, would this be an appropriate way to prevent the binary files check in? Can you PTAL!
> 
> Also, this [#1730 (comment)](https://github.com/oppia/oppia-android/issues/1730#issuecomment-765098904) mentions
> 
> > We may run into an issue with the CI check if it runs across the entirety of develop since there are some binary files in mainline, and we can't force push change those. We may need to build in exceptions for those files.
> 
> What does that exactly mean?

I think this was originally assuming that we'd make a Kotlin script that runs across all files in the codebase to ensure only the binary files we're aware of are ever checked in. That being said, I think what you're introducing here is a sufficient way to prevent binary files from being introduced.

.github/workflows/static_checks.yml
Rd4dev: I missed adding that earlier, but it's included now.

_Also resolving since - done_

.github/workflows/static_checks.yml
Rd4dev: Thanks! Yes, the script is causing the failure. I've removed the redundant check for the pass statement.

Rd4dev:
@BenHenning, thanks for the review. Linking the [stack trace](https://github.com/Rd4dev/oppia-android/actions/runs/10715972847/job/29712474511?pr=11#step:7:15) for the working ci check from fork (updated in PR desc). 

Can you PTAL!
#
### Minor change
Moved file printing into the iteration section to properly render them with color codes in the CI log.

Earlier:
image

Now:
image

BenHenning (approved):
Thanks @Rd4dev! This LGTM. Glad to see this go in. :)

Issue #5535, opened 2024-09-12
[Feature Request]: Target SDK 34


BenHenning:
### Is your feature request related to a problem? Please describe.

Per https://developer.android.com/google/play/requirements/target-sdk and https://apilevels.com/ apps are required to target at least SDK 34 by 31 August 2024 in order to continue deploying using the Play Store.

Since Oppia Android is targeting SDK 33, we are blocked from releasing new versions of the app.

### Describe the solution you'd like

Once #5473 is addressed, we should follow the process and document:
- The list of changes from SDK 33 to SDK 34 relevant to Oppia Android.
- What (if any) follow-up changes are needed to the app before the SDK upgrade can occur.

Then, those changes will need to be implemented and verified, then the app updated to SDK 34 and manually tested.

### Describe alternatives you've considered

The upgrade to SDK 34 could happen directly, but the audit is very important to ensure no new Android functionality breaks the app (which is something I've seen happen before on other apps), and it makes sense to finish #5473 first to ensure the process of doing this is written down since we essentially need to upgrade SDK versions every year.

### Additional context

#5137 demonstrates the process that was conducted for upgrading the app to target SDK 33.

theMr17:
### Known Problems
- System now supports non-linear font scaling up to 200% for accessibility. The app needs to be tested for larger font sizes.


### Require Additional Verification
- ~All dynamically-loaded files must be marked as read-only, or the system will throw an exception. Is it a concern, considering the asset files we load?~

### Interest for Future Work
- ~Android 14 introduces partial access to photos and videos, allowing users to grant apps access to specific images and videos. This needs to be considered if we decide to switch to a different method for selecting user profile pictures.~

### Behavior changes: all apps
https://developer.android.com/about/versions/14/behavior-changes-all contains the high-level behavior changes in SDK 34. Items of note:

- Permission is now required to schedule exact alarms.
- The system can queue context-registered broadcasts while the app is in a cached state.
- Apps can no longer kill the background processes of other apps using `killBackgroundProcesses()`. This function can now only terminate the app's own processes.
- Bluetooth stack enforces an MTU (Maximum Transmission Unit) of (512 + 5) bytes for the first GATT (Generic Attribute Profile) client request.
- Move apps to the restricted standby bucket if their jobs repeatedly cause ANR errors, with tracking via the `UsageStatsManager` API.
- The `mlock()` limit is reduced from 64 MB to 64 KB per process.
[...]

PR #5538, opened 2024-09-17
Fix #5473: Add wiki guide for upgrading the app's compile/target SDK versions

### LINKED ISSUE #5473: [Feature Request]: Provide instructions for updating the app's target API level and verifying compatibility.
### Describe the bug

We are getting regular emails from the Play Store roughly once a year that say:

"[Action required] Your app is affected by Google Play's target API level requirements"

These emails tell us that the app is targeting an old version of Android, and that we need to "meet target API level requirements".

The aim of this issue is threefold:

- [x] Write a wiki page that gives instructions on how to do this in the future (including the data-gathering process), so that any developer on the team can do it. (@BenHenning to do this.) 
- ~Fix the current app to correct the target API level requirements. Per the Google Play email, this must be done by 31 Aug 2024.~
- ~Establish an annual cadence for making these updates, such that they are completed well before we are warned about them by the Play Store.~

### Steps To Reproduce

N/A

### Expected Behavior

We should hopefully stop getting these emails if the app API level requirements are always correctly up-to-date.

### Screenshots/Videos

_No response_

N/A

N/A

N/A

### Additional Context

_No response_

BenHenning:
## Explanation

Fixes #5473

This PR introduces a new wiki page that describes the process by which we can carefully upgrade the app to a newer compile/target SDK version (with a specific emphasis on the target SDK version).

This is largely meant to capture past processes we have followed since those weren't documented in an easily reproducible way. The intent is for this now-documented process to be trialed by upgrading the app to SDK 34 in order to address #5535 (though that technically means that part (1) of the documented process has already been completed).

This PR does not:
- Address exactly when the compile version should be updated, except when required by the target SDK version. It's expected that the team will perform an out-of-band analysis for cases when the compile SDK actually needs to be upgraded sooner than the target SDK.
- Specify a strongly enforced cadence for upgrading the target SDK since the Google Play Console and Google Play app policy seems to already perform this function, though it does add a recommendation for checking for updates periodically (July/August every year).

## Essential Checklist

## For UI-specific PRs only
N/A -- This is introducing a new wiki page and doesn't change the app itself.


BenHenning (commented):
Self-reviewed changes (brief skim).

BenHenning:
PTAL @adhiamboperes for code owners.

PTAL @Rd4dev for just general thoughts on whether the wiki page is well-organized and well-written, and if you have any follow-up suggestions. Please feel free to consider this review optional. If you have the time, I'd appreciate any thoughts you have on this page.

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Upgrade Process
+
+```mermaid
+flowchart TD
+    A(Part 1: Identifying the need to upgrade) -->|Tracking issue exists| B
+    B(Part 2: Auditing the Android OS changelog) -->|Audit complete| C
+    C(Part 3: Testing the app & filing problems) -->|Testing finished & issues filed| D
+    D(Part 4: Fixing and stabilizing support) -->|All issues fixed| E
+    E(Part 5: Submitting the upgrade) -->|Compile/target SDK PR submitted| F
```
Rd4dev: It looks like the illustration content might be getting cut off due to the text length. Could using a `<br>` tag help with this? (or other approach) 

```mermaid
flowchart TD
  E(Part 5: Submitting the upgrade) -->|Compile/target SDK PR submitted| F(Block)
```

with `<br>` tag:

```mermaid
flowchart TD
  E(Part 5: Submitting<br>the upgrade) -->|Compile/target SDK<br>PR submitted| F(Block)
```

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+### Part 1: Identifying the need to upgrade
+
+There are generally three signals that may indicate the team should consider upgrading to a newer Android SDK target:
+1. https://developer.android.com/google/play/requirements/target-sdk indicates an upgrade mandate and deadline. We also get this reminder via the Google Play Console.
+2. A new version of Android has released to users and we perform a periodic check for compatibility (usually around July/August). We may decide to upgrade even if there isn't a mandate.
+3. A feature requires a newer version of Android (note this is unlikely since we generally want to design features to work for all of our users).
+
+If it's deemed that there's a new SDK version to target and the app isn't yet targeting it, a new feature request should be filed (similar to #5535) as long as there isn't an existing tracking issue for this work.
```
Rd4dev: >(similar to #5535)

Would it be preferable to include direct links than issue numbers, as MD files don’t seem to link them directly?

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Part 2: Auditing the Android OS changelog
+
+All new Android OS functionality changes (both those tied to ``targetSdkVersion`` and those not) should be analyzed for potential areas of testing. These can be found on the Android developers site, for example for SDK 34:
+- Changes affecting all apps: https://developer.android.com/about/versions/14/behavior-changes-all.
+- Changes tied to changing target SDK version: https://developer.android.com/about/versions/14/behavior-changes-14.
+
+Any concerning changes or functionality that could be beneficial to Oppia Android should be noted in the tracking issue for the SDK upgrade (see https://github.com/oppia/oppia-android/issues/5137#issuecomment-1815241974 for a good example of this).
+
+Note that narrowing down these categories isn't a process that can be easily described as a set of steps since new OS features may not even be predictable ahead of time. However, here are some tips that might help:
```
Rd4dev, line 41: Would splitting them into separate subheadings, for eg. "High-Level Considerations" and "Post Findings," improve structure and enhance readability? - [sample md](https://github.com/Rd4dev/Oppia-Android-Fork-from-Fork/blob/Rd4dev-patch-3/wiki/up.md#part-2-auditing-the-android-os-changelog)

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Look for changes in permissions. This could either be an old permission that's now more restricted, or existing SDK functionality that's now blocked by a new permission (both have occurred in past Android OS updates). If the functionalty and/or permission relates to Oppia Android, it should be noted. Note that some things may be tied to permissions and access control that's not obvious such as:
```
Rd4dev: Nit: Spelling of functionality
```suggestion
- Look for changes in permissions. This could either be an old permission that's now more restricted, or existing SDK functionality that's now blocked by a new permission (both have occurred in past Android OS updates). If the functionality and/or permission relates to Oppia Android, it should be noted. Note that some things may be tied to permissions and access control that's not obvious such as:
```

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+1. One list for areas with known problems (e.g. API deprecations). Note that each of these problems should be filed as separate bugs in the Oppia Android issue tracker and their issue numbers noted as part of this list.
+2. One list for areas that require additional verification to ensure compatibility.
+3. One list for areas that could be of interest for future work.
+
+The new comment should include an explicit indication of whether the audit was completed, or if additional analysis on Android SDK documentation is needed.
+
+### Part 3: Testing the app & filing problems
+
+Ensuring compatibility with a target SDK upgrade requires two things:
+1. A local Bazel production build of the app (``//:oppia_beta``) with the target SDK temporarily upgraded (see part (5) below) but not checked in.
```
Rd4dev: Would it be helpful to include somewhere a link to the [Bazel Build instructions wiki](https://github.com/oppia/oppia-android/wiki/Oppia-Bazel-Setup-Instructions) here for building the local Bazel production app, or can we assume developers are already familiar with the process?

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+2. A list of focus areas to specifically test (per the audit completed in part (2) above).
+
+Important caveats:
+- Some features may require production assets, but usually local developer lesson assets are sufficient for testing target SDK changes.
+- In some cases, a real device may need to be used instead of an emulator since certain features change behavior on an emulator (such as the drag and drop interaction).
+- If the app isn't already using the new version of Android as its compile SDK version, then it may fail to build. Any build failures should be filed as issues on the issue tracker and fixed before this part of the process can continue.
+
+The local version of the app should be deployed to an emulator (or a real device if one is available) and tested as such:
```
Rd4dev: Would it be beneficial to add subheadings such as 
- Deployment 
- User Behavior Testing 
- General Analysis
- Filing Issues

[sample md](https://github.com/Rd4dev/Oppia-Android-Fork-from-Fork/blob/Rd4dev-patch-3/wiki/up.md#part-3-testing-the-app--filing-problems)

or would it be preferable to maintain the current single flow?

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+Any breakages should be noted, and then checked against a version of the app without the target SDK. From there:
+- If the breakage still occurs, file a new bug noting the problem and mention in the 'additional context' section that it was found during target SDK testing but was determined as unrelated.
+- If the breakage does not occur on the non-upgraded version of the app, file a new bug in the issue tracker and mention that it's specific to the new target SDK version and is a blocking issue.
+
+Please note all found blocking issues with their issue numbers in a follow-up reply to the tracking issue. Any other thoughts or findings during testing can also be noted in the tracking issue (similar to the comments in #5137). Please also note in the tracking issue when testing has concluded.
+
+### Part 4: Fixing and stabilizing support
+
+Work on fixing all identified problems from parts (3) and (4) (either by directly fixing the problems via code changes, or via coordination with other members of the team).
```
Rd4dev: Same clarification as above, if adding subheadings for better structure would be helpful (may be)? 

- Address Identified Problems
- Verification of Fixes
- Final Confirmation

[sample md](https://github.com/Rd4dev/Oppia-Android-Fork-from-Fork/blob/Rd4dev-patch-3/wiki/up.md#part-4-fixing-and-stabilizing-support)

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+- #5222 is an example of a PR that performs both steps in one, though it includes a few additional code changes that were needed as a result of the compile SDK change.
+- Please note the CI results for this upgrade change. Any failures are likely problems that will need to be fixed within the upgrade PR (if small, e.g. the change in #5222) or filed as a separate bug that will need to be fixed before the upgrade can be submitted (see part (4) above).
+- Updating the compile-time SDK version may require updating the build tools version. This is **not** a simple change and may cause difficult-to-fix breakages due to subtle compatibility issues between third-party dependencies and the build system configurations. If you suspect a build tools version upgrade is needed, please file an issue to track it and contact the developer workflow team lead to discuss next steps.
+- Updating tests to use a newer version of Android can be exceptionally complicated, and thus this is considered a completely separate exercise from upgrading production code. See part (6) below for more specifics.
+
+#### Part 5.1: Updating the compiled SDK version
```
Rd4dev: I'm not sure if this is needed or if I’m following the structure correctly, but even though Part 3 (Testing) refers to Part 5 (Upgrading), would it be useful to restructure and include a **'How to Upgrade'** section before the testing section (Part 3).

As I believe testing typically follows the upgrade process, providing the 'How to Upgrade' section earlier might create a clearer path. 

1. Identifying the need to upgrade
2. Auditing the Android OS changelog
3. **How to upgrade**
4. Testing the app & filing problems
5. Fixing and stabilizing support
6. Submitting the upgrade
7. Future work items and upgrading Robolectric

Or is the current structure designed to ensure that all necessary testing procedures are addressed before proceeding with the upgrade?

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+This code change essentially requires replacing the old SDK version number (e.g. 31) with the new one (e.g. 33), but only for compile-time behaviors. All needed changes are detailed below:
+1. [``.github/actions/set-up-android-bazel-build-environment/action.yml``](https://github.com/oppia/oppia-android/blob/dfb9a301280b9a46526cb2f5ca6329532fec6bf0/.github/actions/set-up-android-bazel-build-environment/action.yml#L75-L78) needs to be updated to install the correct SDK version (via the ``sdkmanager --install`` command).
+2. Bazel [``build_vars.bzl``](https://github.com/oppia/oppia-android/blob/dfb9a301280b9a46526cb2f5ca6329532fec6bf0/build_vars.bzl#L1) changes to ``BUILD_SDK_VERSION``.
+3. Gradle ``compileSdkVersion`` changes (e.g. for [``app/build.gradle``](https://github.com/oppia/oppia-android/blob/dfb9a301280b9a46526cb2f5ca6329532fec6bf0/app/build.gradle#L8)). Note that all module ``.gradle`` files will need to be updated in this way.
+
+#### Part 5.2: Updating the target SDK version
```
Rd4dev: Would the WORKSPACE file also require an update in this case? (apologies if I overlooked)

Rd4dev (commented):
Thanks @BenHenning, would just like to understand two minor things:

1. Would it be useful to restructure/include a 'How to Upgrade' section?
2. Would the WORKSPACE file need an update?

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: What exactly is being cut off for you?

Here's what I see:

image

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: I thought about doing that, and it's a good suggestion. Done everywhere on the page.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Added new sections as suggested, though I went with 'tips for auditing' and kept the actual high-level instructions in the top-level section.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Nice catch! Fixed.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Linking to the Bazel documentation seems reasonable--done.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Good question--it shouldn't anymore now that we've moved relevant build configurations to build_vars.bzl. It does still need to be updated for the target SDK, but that'll go away soonish (once we remove the APKs targets--they are rarely used and we should prefer using the AAB builds even for local development).

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: The goal is the latter--we don't actually want to create and submit an upgrade PR until it's time for that to be merged (which is when auditing and testing has completed). I was also a bit concerned with this organizational approach since it leads to lots of jumping around. Some possible options we have:
- Keep it as-is since the overall process is fairly short.
- Introduce an appendix with sections detailing specific tasks that need to be done and how, and update the process to point to these steps (this keeps the process itself linear but still results in jumping to/from the appendix, maybe even more than the current writing).
- Split this into two sections as you have it (though my concern with that is 'how to upgrade' doesn't actually have as strong of a completion indication compared to the other sections).

I slightly lean toward (1) since the other two _seem_ more complicated, but I'm not sold on any of the approaches. Do you have any thoughts?

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Since this is a very short section, I'm not sure the subheadings really add much value. Do you find the current section process difficult to follow such that additional organizing might help?

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: I'm not entirely sold on the specific organizational approach suggested, however I do agree that the overall flow of this section is difficult to follow and, being it's probably the most important section, needed a slight rewrite. I've attempted to reorganize it to have some sub-structure, and to have less redundancy. Is it clearer now or do you think there are additional changes that should be made?

BenHenning (commented):
Self-reviewed latest changes.

BenHenning:
Thanks @Rd4dev! I found your comments very helpful! Could you PTAL at my follow-ups and please let me know if you think any additional changes would be useful?

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: I viewed the page using View File and it renders as below:

image
 
Am I viewing it correctly?

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks.

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks.

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Ah, thanks for the clarification!

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Yeah, makes sense. The current version by itself is easy to follow, so I'll resolve it. Thanks!

wiki/Upgrading-Target-Sdk-Guide.md
```diff
@@ -0,0 +1,155 @@
...
+
+Any breakages should be noted, and then checked against a version of the app without the target SDK. From there:
+- If the breakage still occurs, file a new bug noting the problem and mention in the 'additional context' section that it was found during target SDK testing but was determined as unrelated.
+- If the breakage does not occur on the non-upgraded version of the app, file a new bug in the issue tracker and mention that it's specific to the new target SDK version and is a blocking issue.
+
+Please note all found blocking issues with their issue numbers in a follow-up reply to the tracking issue. Any other thoughts or findings during testing can also be noted in the tracking issue (similar to the comments in [#5137](https://github.com/oppia/oppia-android/issues/5137)). Please also note in the tracking issue when testing has concluded.
+
+### Step 4: Fixing and stabilizing support
+
+Work on fixing all identified problems from parts (3) and (4) (either by directly fixing the problems via code changes, or via coordination with other members of the team).
```
Rd4dev: can these too be updated as steps?

```suggestion
Work on fixing all identified problems from steps(3) and (4) (either by directly fixing the problems via code changes, or via coordination with other members of the team).
```

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: ah, maybe due to varied zoom?

https://github.com/user-attachments/assets/e74f2c87-c4d3-4ef6-b988-fe8f396cbf16

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks!

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks for adding the sub sections, those are helpful!

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Makes sense. I was considering whether that might create a gap in the flow..., but as you pointed out, sticking with option 1 seems like the better approach (for now).

Rd4dev (commented):
Thanks @BenHenning for adding the subsections, and reg. the 'How to upgrade' section, as you mentioned it might be better to stick with the original approach for now and revisit later if needed. 

The graphs seem to render differently at various zoom levels. Please resolve any comments if they're blocking the merge.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: This is actually really interesting. What browser and browser version are you using? This might be a browser-specific GitHub and/or Mermaid bug as zoom doesn't affect things at all for me. From inspecting the graph, it's a generated SVG being rendered.

I'm using Chrome 129.

BenHenning (commented):
Checked latest changes.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Good catch! Fixed.

BenHenning (commented):
Thanks @Rd4dev! I followed up to both comments--PTAL.

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks.

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: I initially tested it on Edge 128, but now I also tested it with
- Chrome 120
- Edge 128
- Chrome 129 (Updated)

seems like on my end, they were still reproducible.

https://github.com/user-attachments/assets/967ae445-7159-4abf-86b3-9039787f6964

Rd4dev:
Thanks, @BenHenning! The only thing left is the zoom levels. I've attached a video reference as I still see the zoom level affecting them. Can you PTAL!

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Could any of the OS specific rendering make these differences, I tested it by installing Mint on VM and resembles as you mentioned but the same with Windows 11 renders it differently. Also, this issue - https://github.com/typora/typora-issues/issues/3291#issuecomment-680117090 seems to mention a similar issue for MacOs.

https://github.com/user-attachments/assets/e9c581cb-29c6-4012-822f-506cb1df96fe

# 

But Firefox renders it properly on both VM Mint OS and Windows 11 (which might suggest the earlier observations are less relevant.).

BenHenning: I'm not 100% sure honestly. Browser rendering can get very complicated, so I can only guess on potential compatibility. I tried searching the Mermaid GitHub repo for similar issues and didn't find a perfect match (though there were other text cut-off issues for separate reasons that were addressed). The issue you linked is much closer to what we're observing here than anything I found.

I'm not opposed to adding breaks, but I have little confidence that will fix the problem correctly since a separate display situations might still result in it being cut off. Or, put another way, how much text can we actually fit on a line before it's likely to be cut off? This is further complicated by us not using a monospaced font so line width won't be exactly correlated to number of characters.

Do you have any suggestions on what we could do here? I'm wondering if a different diagram format or maybe just trying to shorten the text would be best (though I thought of that and couldn't come up with good alternatives at the time).

Rd4dev: I tried applying the `text-wrap` property to the node, and it appears to be working. Could this be a viable / consistent workaround?

Tested on few browsers and added them below, could you PTAL!

Snippet:

```
classDef textWrap text-wrap: wrap;

  A(Part 1: With Text Wrap property set to the node):::textWrap -->|Tracking issue exists| B
  B(Part 1: Without Text Wrap property set to the node) -->|Tracking issue exists| C
```

>Tested on Edge 128 

https://github.com/user-attachments/assets/6147dcd5-6824-4940-9747-aea832323d2c


<details open><summary>Chrome 129</summary>

image

</details>

<details><summary>Edge128</summary>

image

</details>

<details><summary>Firefox 130</summary>

image

</details>

<details><summary>Brave 1.70</summary>

image

</details>

adhiamboperes (approved):
Thanks @BenHenning! I think this guide is thorough and well structured. LGTM!

wiki/Upgrading-Target-Sdk-Guide.md
adhiamboperes: Just mentioning that I have not experienced any cutoff issues on chrome or safari even when zooming.

BenHenning (commented):
Self-reviewed latest changes.

wiki/Upgrading-Target-Sdk-Guide.md
BenHenning: Very nice suggestion! Applied the suggested changes. PTAL @Rd4dev.

BenHenning (commented):
Thanks @Rd4dev and @adhiamboperes!

@Rd4dev I think I addressed your open comment--can you please confirm?

wiki/Upgrading-Target-Sdk-Guide.md
Rd4dev: Thanks, @BenHenning.

Rd4dev:
Thanks @BenHenning for the update. I’ve resolved all my open comments.

BenHenning:
Thanks @Rd4dev!

Issue #4302, opened 2022-04-14
Introduce a JUnit annotation & rule for managing & enumerating platform parameter states


BenHenning:
Arranging tests to have the correct platform parameter state is essential for testing app behavior in cases that are potentially very difficult to test manually (particularly because one needs to manually set up a platform parameter module, override all platform parameters needed by downstream dependencies, and then change the current parameter before injection).

#4204 lays the foundation for much better approach: one module is shared for all tests and can provide case-by-case overriding for each parameter. While this makes the situation much better, it still has the drawback of not technically setting up the parameters in parity with production. Production versions of the app guarantee that a parameter cannot change for the lifetime of the app instance, whereas post-injection parameter overriding does allow for this possibility. This can introduce state inconsistencies that at best lead to bugs in tests and at worst a false sense of correctness when tests that rely on such behavior pass.

I think the best path forward is to build upon the foundation from #4204 by:
1. Introducing a JUnit rule & annotation(s) for overriding platform parameters (see below for some specific thoughts on this). It's expected that this mechanism be built such that the _production_ parameter module changes its behavior (probably by swapping out the provider for the parameters with a test overridable mechanism).
2. Updating OppiaTestRule to automatically incorporate the above rule.
3. Updating all tests to use the production module instead of the test or a custom module, and adding OppiaTestRule/test annotations where needed to ensure overrides are still happening.
4. Removing the test platform parameter module and ensuring that the production module properly scopes parameters to be singletons.
5. Maybe: Introducing additional JUnit functionality to help with testing.

It's possible to also directly build on the new functionality introduced in #4204 by instead overriding values in the test module via a JUnit annotation/rule, and use the test module everywhere. However, we should try to minimize test-specific modules & dependencies where possible as this introduces better parity with production (by reducing test-specific behaviors that could diverge from production behavior). Note also that both cases will help with the hilt migration (#1720) since it normalizes one module for all tests.

Re: (1), something like the following syntax might be nice to implement:

```kotlin
@OverrideStringParameter(name = "some_str_param", value = "new_value")
@OverrideBoolParameter(name = "enable_my_feature", value = true)
fun testSomething_whenSomething_isSomething() {
  // Test dependencies will only ever see the overridden parameter value, and the test itself can inject it.
}

@OverrideBoolParameter(name = "enable_my_feature", value = true)
fun test...
```

[...]

PR #5565, opened 2024-11-04
Fixes part of #4302 - Introducing a JUnit annotation & rule for overriding Platform Parameters and Feature Flags

Rd4dev:
## Explanation

Fixes part of #4302 
> [Not fixed in this PR: `Maybe: Introducing additional JUnit functionality to help with testing.`]

This PR includes:

### Feature Flags Enhancements:
- Introduced annotations to enable, disable, and reset feature flags.
- Add support for overriding platform parameters.
- Annotations work at both the class and method levels (-) reset annotation.
- Annotations are repeatable.

### Refactoring Test Module Constants:
- Moved test module constants to a dedicated Platform Parameters Constants Module.
- This module is used exclusively for testing purposes.

### Test File Migration:
- Migrated all test files to production usage.
- Removed the Test Platform Parameter Module.
- Refactored all Platform Parameter Providers away from `TestModule`.

### Oppia Test Rule Validation:
- Implemented a regex-based file content validation to enforce the inclusion of `@get:Rule val oppiaTestRule = OppiaTestRule()` in test files.
- Identified (~)298 test files missing the Oppia Test Rule.
- Declared Oppia Test Rule in every test file (-) scripts and few domain modules
- [To Complete] Analyze incompatible test rule declaration files. [mostly script files]

### Wiki Page Updated
https://github.com/oppia/oppia-android/blob/bcc9d4a9a670673a623240876192b44f5ddc750f/wiki/Platform-Parameters-%26-Feature-Flags.md#how-to-write-tests-related-to-platform-parameters

#### Todo:
- Analyze incompatible test rule declaration files. [mostly script files].

## Essential Checklist


Rd4dev:
@BenHenning @adhiamboperes Would it be possible to provide a draft review or approval on the implementation to ensure things are on the right track before I proceed with the rest? I would like to confirm if the [mentioned ](https://github.com/oppia/oppia-android/issues/4302#:~:text=Updating%20all%20tests%20to%20use%20the%20production%20module%20instead%20of%20the%20test%20or%20a%20custom%20module%2C%20and%20adding%20OppiaTestRule/test%20annotations%20where%20needed%20to%20ensure%20overrides%20are%20still%20happening.)

> "Updating all tests to use the production module instead of the test or a custom module, and adding OppiaTestRule/test annotations where needed to ensure overrides are still happening." 

is done right. If not then will rework based on the feedback.

BenHenning:
Reviewed during CLaM meeting today and feedback was left in real-time.

BenHenning:
> @BenHenning @adhiamboperes Would it be possible to provide a draft review or approval on the implementation to ensure things are on the right track before I proceed with the rest? I would like to confirm if the [mentioned ](https://github.com/oppia/oppia-android/issues/4302#:~:text=Updating%20all%20tests%20to%20use%20the%20production%20module%20instead%20of%20the%20test%20or%20a%20custom%20module%2C%20and%20adding%20OppiaTestRule/test%20annotations%20where%20needed%20to%20ensure%20overrides%20are%20still%20happening.)
> 
> > "Updating all tests to use the production module instead of the test or a custom module, and adding OppiaTestRule/test annotations where needed to ensure overrides are still happening."
> 
> is done right. If not then will rework based on the feedback.

Everything looks right except the overriding mechanism. Rather than having an override built into the production module, we should be overriding the value via OkHttp and the platform parameter syncing. This might be a bit tricky to set up, so it may well be worth getting this working before migrating all uses over to the new system.

Rd4dev:
Removing stale label, will work on this next week.

Rd4dev:
@BenHenning, @adhiamboperes PTAL, and I also wanted to confirm a few requirements:

1.  What should be the scope of `OppiaTestRule` inclusion?

    Is it required in every module's test file, or are there exceptions?

    The concern is that `OppiaTestRule` attempts to acquire `Build.FINGERPRINT` to determine the Test Platform. While this works for other modules that access Android classes, the script module test cases throw a `NoClassDefFoundError`. Although the issue can be mitigated by redirecting to use Robolectric as the Test Platform, the previous error logs interfere with current assertion throw cases, leading to test failures. I wanted to clarify the scope and requirements for including the rule before investigating other workarounds.

2. Codeowners Check Failure

    Are there any specific changes I need to address to resolve the [CODEOWNERS check failure](https://github.com/oppia/oppia-android/actions/runs/12678024040/job/35334689300?pr=5565#step:4:16)?

3. Reset Annotation

    As discussed in a prior CLaM meeting, it was suggested to remove the `ResetPlatformParameter` annotation since `overridePlatformParameter` was deemed sufficient. I just have few questions:

    - If `ResetPlatformParameter` was removed for platform parameters, should it also be removed for feature flags? (I remember it being mentioned to not to) but the same functionality (resetting to a default value) could be achieved by utilizing `EnableFeatureFlag` or `DisableFeatureFlag`.
    - If keeping a reset annotation is advisable, would maintaining it for both feature flags and platform parameters be helpful to ensure parity and avoid confusion?
    - Alternatively, would keeping the reset annotation exclusively for feature flags suffice?

testing/src/main/java/org/oppia/android/testing/OppiaTestRule.kt
```diff
@@ -133,5 +234,34 @@ class OppiaTestRule : TestRule {
     private fun <T> Class<T>.areAccessibilityTestsEnabledForClass(): Boolean {
       return getAnnotation(DisableAccessibilityChecks::class.java) == null
     }
+
+    /**
+     * Extracts all feature flag annotations of the specified type from a collection of annotations.
+     *
+     * @param annotations a collection of annotations to be checked for feature flag annotations.
```
Rd4dev: Todo: To remove ending periods from all KDocs.
```suggestion
     * @param annotations a collection of annotations to be checked for feature flag annotations
```

adhiamboperes:
Thanks @Rd4dev! Reviewing this today.

Rd4dev:
> Thanks @Rd4dev! Reviewing this today.

Thanks @adhiamboperes, appreciate the update!

testing/src/main/java/org/oppia/android/testing/EnableFeatureFlag.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Repeatable test class or method annotation for enabling the feature flag for tests of the
+ * class or the specific method may run on. The feature flag names are provided by the
+ * FeatureFlagConstants.kt.
+ *
+ * Note that this annotation only works if the test also has an [OppiaTestRule] hooked up.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+@Repeatable
+annotation class EnableFeatureFlag(val name: FeatureFlag)
```
BenHenning, line 14: For both this and the disable variant, it would be good to explain here how cascading cases or repeated cases are handled. For instance, consider the following setup:

```kotlin
@EnableFeatureFlag(FEATURE_1)
@DisableFeatureFlag(FEATURE_1)
class MyConfusingTest {

  @Test
  fun test_case1() {}

  @Test
  @EnableFeatureFlag(FEATURE_1)
  fun test_case2() {}

  @Test
  @DisableFeatureFlag(FEATURE_1)
  fun test_case3() {}

  @Test
  @EnableFeatureFlag(FEATURE_1)
  @DisableFeatureFlag(FEATURE_1)
  fun test_case4() {}

  @Test
  @DisableFeatureFlag(FEATURE_1)
  @EnableFeatureFlag(FEATURE_1)
  fun test_case5() {}
}
```

What is the expected state of ``FEATURE_1`` in each test case? Is it an error or does latest win?

Personally, I would expect that:
- More than one declaration for a single feature at the class or method level results in an error (i.e. ``MyConfusingTest``, ``test_case4``, and ``test_case5`` above should all fail).
- It's allowed to have at most 2 overrides for a single feature for a given test: one at the class level, and one at the method level, where the method annotation always takes precedence. This allows us to enable/disable a feature for the whole class, then selectively enable/disable it on a per-test basis as needed.

testing/src/main/java/org/oppia/android/testing/OverrideStringParameter.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Repeatable test class or method annotation for overriding string platform parameter values
+ * for tests of the class or the specific method may run on. The platform parameter names are provided
+ * by the PlatformParameterConstants.kt.
+ *
+ * Note that this annotation only works if the test also has an [OppiaTestRule] hooked up.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+@Repeatable
+annotation class OverrideStringParameter(val name: PlatformParameter, val value: String)
```
BenHenning, line 14: For this & the other override classes: see my comment in ``EnableFeatureFlag`` as I think the same consideration is needed here on how cascading and overriding scenarios are handled.

testing/src/main/java/org/oppia/android/testing/ResetFeatureFlagToDefault.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Repeatable test method annotation for resetting the feature flag for tests of the
+ * specific method may run on. The feature flag names are provided by the
+ * FeatureFlagConstants.kt.
+ *
+ * Note that this annotation only works if the test also has an [OppiaTestRule] hooked up.
+ */
+@Target(AnnotationTarget.FUNCTION)
+@Repeatable
+annotation class ResetFeatureFlagToDefault(val name: FeatureFlag)
```
BenHenning, line 14: Similar to my comment in ``EnableFeatureFlag``, cascading/duplication needs to be considered here. I'd expect any cases where an enable/disable happens and reset at the same level to be an error (i.e. a test method can only reset, not enable/disable and then reset).

For strictness, I'd also expect that reset throws an error if there's no explicit override of the feature at the class level (i.e. there's no reason to reset). While resetting a non-overridden flag is a no-op, allowing it to be checked in could result in false assumptions and a fundamental misunderstanding of what the test and surrounding test suite is actually doing.

BenHenning (commented):
Thanks @Rd4dev! This is really exciting to see progress on.

I only looked at the top-level override classes for now, and still need to dig deeper into the PR. However, I do have one thought that I wanted to run by you.

I've had an idea for some time to leverage code generation for the feature flags to minimize the amount of Dagger binding boilerplate that needs to be done each time a new flag is added. Originally, I thought that this would require a custom annotation processor. However, now that Gradle is gone, we actually could do this with protobuf. Consider:
- A new textproto file that lists every feature flag and platform parameter with server string names and default values. I suspect this would be a twofold definition: explicit parameters in a proto file, and server names and default values in textproto (so that we have strong typing for the separate flags).
- Automatic compilation to binary proto (like we do with the language configuration under ``config/``).
- Some file loading logic similar to that which is needed for the languages configuration.
- Updating Dagger bindings that just generate the entire set of bound flags/parameters by loading the textproto file.\*

\* Note that this is a little trickier than it sounds because we can't technically do file I/O within Dagger providers without incurring some performance issues. However, something that we _could_ do is automatically bind implementation classes based on the proto structure (which is available at compile time since the individual flags would be defined within a proto file), and then have the implementations support asynchronously loading their state (either from the local file cache or from their local file configuration). It's possible we don't even need this second bit since I think the platform parameter system is technically supposed to be designed to fully load its asynchronous state during the app startup so that all flag values are fixed for the lifetime of the app process.

While a long explanation, I suspect that the delta might be relatively small and also be a pretty substantial improvement to how flags and parameters are interacted with in both tests and production code. There's probably more that could be done with real code generation, but that starts getting a bit scary in scope here.

Is this something you'd want to take on as part of this PR? If not, no worries. I just wanted to ask since you've already spent a lot of time thinking about the mechanisms. :)

app/src/sharedTest/java/org/oppia/android/app/home/HomeActivityTest.kt
```diff
@@ -1985,7 +1916,6 @@ class HomeActivityTest {
 
   @Test
   fun testHomeActivity_onBackPressed_soleLearnerProfile_exitsApp() {
-    setUpTestWithOnboardingV2Enabled()
```
adhiamboperes, line 1988: This test and 3 below require the feature flag to be enabled, but they are missing the annotation. Do they pass?

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -324,8 +338,10 @@ class TopicFragmentTest {
   }
 
   @Test
+  @EnableFeatureFlag(FeatureFlag.SPOTLIGHT_UI)
+  @EnableFeatureFlag(FeatureFlag.EXTRA_TOPIC_TABS_UI)
```
adhiamboperes: Should this be disabled?

domain/src/test/java/org/oppia/android/domain/survey/SurveyGatingControllerTest.kt
```diff
@@ -280,6 +286,7 @@ class SurveyGatingControllerTest {
   }
 
   @Test
+  @EnableFeatureFlag(FeatureFlag.ENABLE_NPS_SURVEY)
```
adhiamboperes, line 289: Since this test class is valid  only for the feature flag enabled situation, I think it is sufficient to enable the flag at class level.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -103,31 +113,39 @@ class PlatformParameterModule {
...
-      EDIT_ACCOUNTS_OPTIONS_UI
-    ) ?: PlatformParameterValue.createDefaultParameter(
-      ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE
-    )
+    return overriddenFeatureFlags[FeatureFlag.EDIT_ACCOUNTS_OPTIONS_UI]?.let {
+      PlatformParameterValue.createDefaultParameter(it)
+    } ?: platformParameterSingleton.getBooleanPlatformParameter(EDIT_ACCOUNTS_OPTIONS_UI)
+      ?: PlatformParameterValue.createDefaultParameter(
+        ENABLE_EDIT_ACCOUNTS_OPTIONS_UI_DEFAULT_VALUE
+      )
```
adhiamboperes, line 138: I don't understand why this is a combination of feature flag and platform parameter here and below.

adhiamboperes (changes requested):
Thanks @Rd4dev! I have left a few comments, PTAL.

PR #5286, opened 2023-12-23
Fix #4859, #4530: Images in 'li' tags display in inline rather than block mode.

Vishwajith-Shettigar:
## Explanation

Fixes #4859, wrapped images inside div tag to display images in block mode.
Fixes #4530, passing (maxAvailableWidth - maxContentItemPadding) instead of passing maxAvailableWidth to calculate InitialMargin helps prevent the image from being cut off.
## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
| Before| After|
|--------|--------|
| imagleik| fixedroundingnumbers|
|WhatsApp Image 2023-12-23 at 11 53 30 AM|fixed|
| | imageliscreenshot  |
| | imageli2 |
| | tabletimageli |
| | tabletimageli2 |

| RTL|
|--------|
| rtlimage| 

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #5212, opened 2023-10-28
Fix #5032 Images in Arabic (RTL) lessons are right-aligned rather than center-aligned.

Vishwajith-Shettigar:
## Explanation
Fixes  #5032,  To make the image center-aligned, we are adding styling to the HTML content image and removing  the left margin by setting drawableLeft bound to 0 in RTL. 

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

### Before
LTR Potrait

beforeLTRpotrait

LTR Landscape

LTRlandscape

RTL Potrait

beforeRTLpotrait

RTL Landscape 

RTLlandscape

### After

RTL Potrait

Potrait_after

RTL Landscape

landscapeafter

LTR Potrait

LTRpotraitafter

LTR Landscape

LTRLandscape

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #5573, opened 2024-11-07
Fix #5039 : Align policy text and symbols to the left, partial fix for list items

### LINKED ISSUE #5039: [BUG]: Terms of Service and Privacy Policy content should still be left-aligned, even if the app is in an RTL language.
**Describe the bug**

In RTL languages, we are currently keeping our Terms of Service and Privacy policy content in English (so that we can have a canonical version). However, the **English** text should be left-aligned.

(Thanks to @KolliAnitha for the repro and images, which I took from https://github.com/oppia/oppia-android/issues/5028.)

**To Reproduce**
1. Go to 'Home' - Options
2. Select App language as Arabic
3. Go back and click on Help - Terms of Service/ Privacy policy
4. See that the content is right-aligned despite being in English.

**Expected behavior**
The English content should be left-aligned (but not the note at the bottom of the page linking to the up-to-date policy).

**Demonstration**
SS

**Environment**
Device/emulator being used: Infinix SMART 5
Android or SDK version (e.g. Android 5 or SDK 21): Android 11
App version (you can get this through system app settings or via the admin controls menu in-app): 0.11-beta-8c81c98d8b

### 1. Update ForcedActivityLanguageMode Proto

Update the `ForcedActivityLanguageMode` enum in `languages.proto` to include a new option:

```proto
USE_ENGLISH = 3;
```

This will allow activities to explicitly request English-only mode.

### 2. Create InjectableAutolocalizedEnglishOnlyActivity

Create a new `InjectableAutolocalizedEnglishOnlyActivity.kt` following the same pattern as [InjectableAutoLocalizedAppCompatActivity](https://github.com/oppia/oppia-android/blob/develop/app/src/main/java/org/oppia/android/app/activity/InjectableAutoLocalizedAppCompatActivity.kt)

This new activity class will serve as the base class for activities that need to be displayed exclusively in English.

### 3. Add New Function to [TranslationController](https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt)

Introduce a new function in [TranslationController](https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt):

```kotlin
fun getLocaleFor(language: OppiaLanguage): DataProvider<OppiaLocale.DisplayLocale>
```

This function will be used in [AppLanguageWatcherMixin.kt](https://github.com/oppia/oppia-android/blob/469fe9c1a6b949299f8f4d58051388fed47fba83/app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt#L71-L75) to instantiate the `InjectableAutolocalizedEnglishOnlyActivity`.
### 4. Modify AppLanguageWatcherMixin

Update [AppLanguageWatcherMixin.kt](https://github.com/oppia/oppia-android/blob/469fe9c1a6b949299f8f4d58051388fed47fba83/app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt#L71-L75) to handle the new `USE_ENGLISH` mode and integrate with the TranslationController changes.

### 5. Modify InjectableSystemLocalizedAppCompatActivity

In [InjectableSystemLocalizedAppCompatActivity.kt](https://github.com/oppia/oppia-android/blob/develop/app/src/main/java/org/oppia/android/app/activity/InjectableSystemLocalizedAppCompatActivity.kt)
- Replace the boolean `shouldOnlyUseSystemLanguage = true` with an enum
- Add enum fields for:
  - `shouldOnlyUseSystemLanguage`
  - `shouldOnlyUseEnglishLanguage` (needed for the PoliciesActivity)

### 6. Modify PoliciesActivity

Update the `PoliciesActivity` to extend the new base class and override the `initializeMixin()` method:

```kotlin
override fun initializeMixin(appLanguageWatcherMixin: AppLanguageWatcherMixin) {
  appLanguageWatcherMixin.initialize(ForcedActivityLanguageMode.USE_ENGLISH)
}
```

### 7. String Content Processing

Extract raw string content (e.g., `val policyDescription`) using regex to:
- Isolate specific tags (such as `li`)
- Apply necessary formatting
- Set the modified string to the TextView

**Reference implementations:**
- [PR #5286 Files](https://github.com/oppia/oppia-android/pull/5286/files)
- [PR #5212 Files](https://github.com/oppia/oppia-android/pull/5212/files)
- [PR #5573](https://github.com/oppia/oppia-android/pull/5573)

### 8. UI Testing

Create UI tests to verify the implementation:

#### Test Requirements:
- **Text Content Verification**: Ensure all text content is displayed in English
- **Text Alignment Verification**: Confirm text is left-aligned using:

```kotlin
assertThat(effectiveGravity and Gravity.HORIZONTAL_GRAVITY_MASK).isEqualTo(Gravity.LEFT)
```
- To properly validate the `USE_ENGLISH` mode, force the system locale to be set to a non-English language (e.g., Swahili) during the test.
- This change will require updating:
    - The system language tests to verify Swahili behavior.
    - The new app language tests to confirm that USE_ENGLISH correctly overrides the system locale.

## Files to Modify/Create

1. **Modify**: `languages.proto` - Add `USE_ENGLISH = 3` enum option
2. **Create**: `InjectableAutolocalizedEnglishOnlyActivity.kt` - New base activity class
3. **Modify**: `TranslationController.kt` - Add `getLocaleFor()` function
4. **Modify**: `AppLanguageWatcherMixin.kt` - Handle new English-only mode
5. **Modify**: `InjectableSystemLocalizedAppCompatActivity.kt` - Replace boolean with enum
6. **Modify**: `PoliciesActivity.kt` - Extend new base class and override initialization
7. **Modify**: `PoliciesFragmentPresenter.kt` - Extract raw string content, apply formatting and set it to textView.
8. **Create**: New UI test cases for validation

TanishMoral11:
Fix #5039: 
### PR Description:

This PR fixes issue #5039, where the Terms of Service and Privacy Policy content was incorrectly right-aligned in RTL (Right-to-Left) languages despite being in English. The goal is to keep the English text left-aligned, even when the app is set to an RTL language like Arabic.

### Key Changes:
1. **Updated `PoliciesActivity`**:
   - Changed `PoliciesActivity` to extend `InjectableAutoLocalizedAppCompatActivity`.
   - Added an override for `initializeMixin` to use `ActivityLanguageSource.USE_ENGLISH`, forcing the Policies screen to always use English and ensuring left-to-right alignment.

2. **Modified `InjectableAutoLocalizedAppCompatActivity`**:
   - Replaced the `shouldOnlyUseSystemLanguage = false` boolean with `ActivityLanguageSource.USE_APP_LANGUAGE` for consistency with the new enum-based approach.

3. **Modified `InjectableSystemLocalizedAppCompatActivity`**:
   - Updated to use `ActivityLanguageSource.USE_SYSTEM_LANGUAGE` instead of the old boolean `shouldOnlyUseSystemLanguage = true`.

4. **Enhanced `AppLanguageWatcherMixin`**:
   - Changed the `initialize` method to accept an `ActivityLanguageSource` parameter instead of a boolean.
   - Added support for `USE_ENGLISH` to enforce English localization when specified, alongside `USE_SYSTEM_LANGUAGE` and `USE_APP_LANGUAGE`.

5. **Added `getLocaleFor` in `TranslationController`**:
   - Introduced a new method to fetch the English locale explicitly using `OppiaLanguage.ENGLISH`, supporting the `USE_ENGLISH` option.

6. **Defined `ActivityLanguageSource` Enum in `languages.proto`**:
   - Created an enum with `USE_SYSTEM_LANGUAGE`, `USE_APP_LANGUAGE`, and `USE_ENGLISH` to control activity localization flexibly.

7. **Updated Tests in `AppLanguageWatcherMixinTest`**:
   - Adjusted test cases to use the new `ActivityLanguageSource` enum and added tests for `USE_ENGLISH` to verify English localization.

### Impact:
- The Policies screen now displays English content that stays left-aligned, even in RTL language settings (e.g., Arabic).
- This improves readability and consistency for English text in legal pages, regardless of the app’s language.

## After Changes : 

https://github.com/user-attachments/assets/0112f850-ed6a-46f8-a22c-fb6389bf26da

## Essential Checklist
- [x] The PR title and explanation start with "Fix #5039".
- [x] Changes to any scripts or assets are justified.
- [x] The PR follows the [coding style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] Unnecessary code changes from Android Studio are not included.
- [x] The branch is up-to-date and not named `develop`.
- [x] The PR is assigned to relevant reviewers.


TanishMoral11:
Hey @adhiamboperes ,

I have created a draft PR addressing #5039 , which focuses on ensuring that policy text and symbols (such as bullet points and question marks) are aligned to the left in an LTR format. While this change improves the overall text alignment, there remains an outstanding issue with the alignment of `<li>` (list item) tags when rendered within parsed HTML content.

### Issue Summary:
Although most text and individual symbols are now correctly aligned, list items created using the `<li>` tag within the policy descriptions are not fully aligned to the left. The current span-based approach successfully formats standard text and symbols but does not affect `<li>` elements, leading to inconsistencies in the displayed content.

### Request for Guidance:
I kindly ask for your insights or recommendations on potential solutions for handling the alignment of `<li>` tags in this context. Specifically, any advice on:
- Custom parsing strategies that could be employed for list elements.
- Advanced spans or Android-specific methods to ensure `<li>` tags align correctly within parsed text.

Thanks !!

TanishMoral11:
Hey @adhiamboperes PTAL

adhiamboperes:
Thanks @TanishMoral11! The solution for bullets look good. As for the items in <li> tags, One approach attempted by @Vishwajith-Shettigar was to take the raw string, like `val policyDescription` and use regex to extract the reguired tag (in this case `li`) and apply the necessary formatting, then setting the mofidified string to the textview. E.g. in https://github.com/oppia/oppia-android/pull/5286/files and https://github.com/oppia/oppia-android/pull/5212/files.

TanishMoral11:
Hey @adhiamboperes , PTAL

https://github.com/user-attachments/assets/0124c4ed-7ef6-429d-9d6b-ba9a6cc74c97

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -228,43 +224,22 @@ class HtmlParser private constructor(
...
-        entityId = "",
-        imageCenterAlign = false,
-        consoleLogger = consoleLogger,
+        context,
+        urlImageParserFactory,
+        gcsResourceName,
+        entityType,
+        entityId,
+        imageCenterAlign,
+        consoleLogger,
```
adhiamboperes: 1. Per the KDocs for these two create functions, they were seperated for a reason. I do not think you should combine them without providing sufficient reason.
2. We do not mix named and unamed arguments, and in a situation where there are 3 or more arguments like this, we use named arguments to improve readability. Please revert to named arguments as originally present.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -123,17 +125,11 @@ class HtmlParser private constructor(
     }
 
     val imageGetter = urlImageParserFactory?.create(
-      htmlContentTextView,
-      gcsResourceName,
-      entityType,
-      entityId,
-      imageCenterAlign
+      htmlContentTextView, gcsResourceName, entityType, entityId, imageCenterAlign
```
adhiamboperes: Nit: Formatting change not required.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -123,17 +125,11 @@ class HtmlParser private constructor(
...
     )
 
     val htmlSpannable = CustomHtmlContentHandler.fromHtml(
-      htmlContent,
-      imageGetter,
-      computeCustomTagHandlers(supportsConceptCards, htmlContentTextView)
+      htmlContent, imageGetter, computeCustomTagHandlers(supportsConceptCards, htmlContentTextView)
```
adhiamboperes: Nit: Formatting change not required.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
```diff
@@ -289,8 +266,9 @@ class HtmlParser private constructor(
...
         customOppiaTagActionListener = null,
         policyOppiaTagActionListener = policyOppiaTagActionListener,
-        displayLocale = displayLocale
+        displayLocale = displayLocale,
+        supportLtr = supportLtr
       )
     }
   }
-}
+}
```
adhiamboperes: Insert new line at EOF.

utility/src/main/java/org/oppia/android/util/parser/html/LiTagHandler.kt
```diff
@@ -291,4 +314,4 @@ class LiTagHandler(
     private fun <T : Mark<*>> Spannable.addMark(mark: T) =
       setSpan(mark, length, length, Spanned.SPAN_MARK_MARK)
   }
-}
+}
```
adhiamboperes: Insert new line at EOF.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -179,4 +181,4 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
 
     override fun getLeadingMargin(first: Boolean) = computedLeadingMargin
   }
-}
+}
```
adhiamboperes: Insert new line at EOF.

adhiamboperes (changes requested):
Thanks @TanishMoral11.

While the solution appears correct, a number of cleanup changes are required to get the PR merge ready.

Also, the video you have uploaded in the comments does not play. Please re-upload it to the PR description and ensure it can play.

TanishMoral11:
Hey @adhiamboperes PTAL ,

Thank you for your detailed review and valuable feedback. 

I have addressed all your requested adjustments, including:

- **Retaining the separation of the `create` methods** as per your guidance to preserve their distinct purposes.
- **Reverting to named arguments** for improved readability, following your recommendation.
- Ensuring **minor formatting issues** have been resolved, including inserting new lines at EOF where necessary.

Additionally, I have re-uploaded the video demonstration to the PR description, ensuring it is now accessible and plays correctly.

Thank you again for your guidance. I look forward to your review of these updates. Please let me know if there are any further adjustments needed.


https://github.com/user-attachments/assets/73008d0b-b100-4c80-95f4-72fc81339ed8

app/src/main/java/org/oppia/android/app/policies/PoliciesFragmentPresenter.kt
```diff
@@ -86,4 +90,4 @@ class PoliciesFragmentPresenter @Inject constructor(
         (activity as RouteToPoliciesListener).onRouteToPolicies(PolicyPage.TERMS_OF_SERVICE)
     }
   }
-}
+}
```
adhiamboperes: Missing newline at EOF.

adhiamboperes (changes requested):
Thanks @TanishMoral11! There is one pending fix, PTAL.

TanishMoral11:
Hey @adhiamboperes PTAL,

Thank you for pointing that out! I’ve added the missing newline at the end of the file. 
Moving forward, I will ensure that this issue does not occur again.
Thanks!

adhiamboperes:
@TanishMoral11, Please check the CI report for failing tests.

TanishMoral11:
Hey @adhiamboperes ,
I reviewed the CI report but am having trouble understanding why the tests are failing. 
Would you be able to provide some guidance or insights on what might be causing this?

adhiamboperes:
@TanishMoral11  all the tests are failing with this error. Please check that changes to the MathTagHandler did not conflict with the original usage.
Screenshot 2024-11-18 at 14 54 24

TanishMoral11:
Hey @adhiamboperes Could You PTAL ,

### Explanation of Implementation

I’ve made updates to handle the left alignment of text, especially bullet points, in the Policies page for both RTL and LTR layouts.

1. **Left-Aligned Bullet Points:**  
   A custom span (`LeftAlignedSymbolsSpan`) was added to ensure that bullet points are always left-aligned, regardless of text direction.

2. **Text Alignment Adjustments:**  
   The text in the policy description was explicitly set to left-aligned for both layout direction (`View.LAYOUT_DIRECTION_LTR`) and text direction (`View.TEXT_DIRECTION_LTR`), ensuring consistent presentation.


Test Passed Screen Shot :
Screenshot 2024-12-10 211515


After Changes Video : 
https://github.com/user-attachments/assets/db3a82db-4a71-4aeb-b483-9eb822da7080

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -39,14 +38,13 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
   ) : ListItemLeadingMarginSpan() {
     private val resources = context.resources
     private val bulletRadius = resources.getDimensionPixelSize(R.dimen.bullet_radius)
-    private val spacingBeforeText = resources.getDimensionPixelSize(R.dimen.spacing_before_text)
-    private val spacingBeforeBullet = resources.getDimensionPixelSize(R.dimen.spacing_before_bullet)
 
     private val bulletDiameter by lazy { bulletRadius * 2 }
+    private val baseMargin = (16f * context.resources.displayMetrics.density).toInt()
```
adhiamboperes: Here and below, why are we using a hardcoded value of 16f here?

utility/src/main/java/org/oppia/android/util/locale/LeftAlignedSymbolsSpan.kt
```diff
@@ -0,0 +1,40 @@
+package org.oppia.android.util.locale
+
+import android.graphics.Canvas
+import android.graphics.Paint
+import android.text.style.ReplacementSpan
+
+// Custom span to force LTR alignment for symbols as well
+// Custom span to force LTR alignment for all text including symbols
```
adhiamboperes: This should be a KDOC rather than a comment.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -153,30 +143,24 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
       val isFirstCharacter = startCharOfSpan == start
 
       if (isFirstCharacter) {
+        // Force left alignment
+        paint.textAlign = Paint.Align.LEFT
+
         val textWidth = Rect().also {
```
adhiamboperes: Looks like this variable is nolonger used in the new code.

adhiamboperes (changes requested):
Thanks @TanishMoral11!

I have some clarifications, and a few comments to solve the CI failures.
1. Please fix the unused attribute error `utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt:149:13: warning: variable 'textWidth' is never used`
2.  [LeftAlignedSymbolsSpan.kt](https://github.com/oppia/oppia-android/pull/5573/files#diff-629cfdb57266c5bf09ed4b58981c9004970b425495035e4eef5655f5cef1d374) is a new class, but it does not seem to have been added to a Bazel BUILD file, so Bazel may not be able to find it during build.
3. Is `locale` the correct package for `LeftAlignedSymbolsSpan`? Please verify that that is the same package where similar utilities are created.

Lastly, please do a spot check in a few lessons/revision cards to verify that the bullets behavior remains unaffected.

utility/src/main/java/org/oppia/android/util/locale/LeftAlignedSymbolsSpan.kt
TanishMoral11: Done.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
TanishMoral11: Thanks @adhiamboperes , I’ve removed this unused variable.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
TanishMoral11: Apologies for the oversight. I’ve replaced the hardcoded value to improve maintainability.

For **Unordered List (UL)**:
``` 
val baseMargin = context.resources.getDimensionPixelSize(R.dimen.spacing_before_bullet)
```
---
For **Ordered List (OL)**:
``` 
val baseMargin = context.resources.getDimensionPixelSize(R.dimen.spacing_before_number_prefix)
```

TanishMoral11:
Thanks @adhiamboperes , PTAL !!

> 2\. `[LeftAlignedSymbolsSpan.kt](https://github.com/oppia/oppia-android/pull/5573/files#diff-629cfdb57266c5bf09ed4b58981c9004970b425495035e4eef5655f5cef1d374)` is a new class, but it does not seem to have been added to a Bazel BUILD file, so Bazel may not be able to find it during build.

I Added `LeftAlignedSymbolsSpan.kt` To The Bazel BUILD File But Encountered Test Failures. Could You Help Me Understand Which Tests Are Failing And Any Specific Integration Requirements?

> 3\. Is `locale` the correct package for `LeftAlignedSymbolsSpan`? Please verify that that is the same packake where similar utilities are created.

Yeah, You're Right About The Package Location. I Moved `LeftAlignedSymbolsSpan` To `org.oppia.android.util.parser.html` Since It's An HTML Parsing Utility.

> Lastly, please do a spot check in a few lessons/revision cards to verify that the bullets behavior remains unaffected.

I've Tested The App, And As Yet No Misbehavior Was Detected Related To These Changes.

utility/src/main/java/org/oppia/android/util/parser/html/BUILD.bazel
```diff
@@ -46,6 +46,7 @@ kt_android_library(
     name = "list_item_leading_margin_span",
     srcs = [
         "ListItemLeadingMarginSpan.kt",
+        "LeftAlignedSymbolsSpan.kt",
```
adhiamboperes: @TanishMoral11, to answer your question regarding adding this class to the build file:
1. Sometimes a kt lib can have multiple source files like you have added here, and for `tag_handlers`, but it depends on usecase. E.g., if you want to use `LeftAlignedSymbolsSpan` in a different part of the codebase, you will need to add a dependency on the lib, `list_item_leading_margin_span`, which will provide both `ListItemLeadingMarginSpan` and `LeftAlignedSymbolsSpan`. On the other hand, each class can have it's own lib if they are only used in a standalone way, like most of the other utility classes in this package and their corresponding libs in this BUILD file.

adhiamboperes (changes requested):
Apologies for taking a while to review this, @TanishMoral11.

A few points to note on Bazel:
1. I have taken a look at your CI failures. The latest error is `app/policies/PoliciesFragmentPresenter_updated.kt:78:56: warning: parameter 'lineIndex' is never used, could be renamed to _
    parsedHtmlDescription.split("\n").forEachIndexed { lineIndex, line ->`. Unless you fix this error, you won't be able to see if there are additional errors. 
2. You can read [this wiki page](https://github.com/oppia/oppia-android/wiki/Interpreting-CI-Results) for information on how to read Bazel CI errors. A youtube video is also linked.
3. You should set aside some time to set up the project build in Bazel, so that you can run Bazel builds locally. Additionally, some more advanced issues you'll encounter will require you to have Bazel set up. The discussions section will be helpful in searching for solutions to some common setup issues.

TanishMoral11:
Hey @adhiamboperes , PTAL !!

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -39,14 +38,13 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
   ) : ListItemLeadingMarginSpan() {
     private val resources = context.resources
     private val bulletRadius = resources.getDimensionPixelSize(R.dimen.bullet_radius)
-    private val spacingBeforeText = resources.getDimensionPixelSize(R.dimen.spacing_before_text)
-    private val spacingBeforeBullet = resources.getDimensionPixelSize(R.dimen.spacing_before_bullet)
 
     private val bulletDiameter by lazy { bulletRadius * 2 }
+    private val baseMargin = context.resources.getDimensionPixelSize((spacing_before_bullet))
+
     private val isRtl by lazy {
```
adhiamboperes: This variable appears unused.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -123,17 +119,12 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
     private val displayLocale: OppiaLocale.DisplayLocale
   ) : ListItemLeadingMarginSpan() {
     private val resources = context.resources
```
adhiamboperes: This and `displayLocale` above are nolonger used.

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
```diff
@@ -123,17 +119,12 @@ sealed class ListItemLeadingMarginSpan : LeadingMarginSpan {
...
+      context.resources.getDimensionPixelSize((R.dimen.spacing_before_number_prefix))
 
     // Try to use a computed margin, but otherwise guess if there's no guaranteed spacing.
     private var computedLeadingMargin =
-      2 * longestNumberedItemPrefix.length + spacingBeforeText
-
-    private val isRtl by lazy {
-      displayLocale.getLayoutDirection() == ViewCompat.LAYOUT_DIRECTION_RTL
-    }
+      2 * longestNumberedItemPrefix.length + baseMargin
```
adhiamboperes: Even though you have refactored the computation of this, I don't see any current usages.

adhiamboperes (changes requested):
Thanks @TanishMoral11! I only have comments for one file, PTAL.

I have updated the branch from develop too re-trigger CI so we can see the CI failures, the logs had already expired.

TanishMoral11:
Hey @adhiamboperes , PTAL !!

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
TanishMoral11: Done  .

utility/src/main/java/org/oppia/android/util/parser/html/ListItemLeadingMarginSpan.kt
TanishMoral11: Done .

adhiamboperes:
@TanishMoral11, it is not clear to me which changes are in the latest version of the PR. Could you please ensure all checks are passing, or let us know how we can help

TanishMoral11:
Hey @adhiamboperes , PTAL !!

Thanks.

TanishMoral11:
Hey @adhiamboperes , PTAL !!
This PR is ready for review.
Thanks.

adhiamboperes:
IMG_0041

@TanishMoral11, did you need help identifying the test failures, or understanding why the tests are failing?

TanishMoral11:
Hi @adhiamboperes , 


> did you need help identifying the test failures, or understanding why the tests are failing?

Yeah , can you please guide me on how to pass all the CI checks? I tried everything and double-checked, but I don’t know why 6 tests failed.

TanishMoral11:
Hi @Rd4dev,

In the first approach, I added a `supportLtr` flag to the `HtmlParser` and `LiTagHandler` classes to force left-alignment for the Terms/Privacy text even in RTL. This flag overrides the default RTL alignment for specific text by adjusting the `isRtl` logic. I also updated the bullet/list rendering in `ListItemLeadingMarginSpan` to respect this flag.

I tested manually by running the app in both Arabic (RTL) and English (LTR) languages. All screens worked correctly.

However, I am facing the following issue:

`Expected imageRetriever to be not null.`

image

That’s why I changed my approach. In my second approach, I also tested all screens and the complete app in both Arabic (RTL) and English (LTR). This approach works fine for the entire app.

However, by using this method, 6–10 tests are affected. If you think the primary approach is better, no problem, I’ll continue with that approach.

Could you please help me resolve the `Expected imageRetriever to be not null.` issue? I’ve tried everything on my side, but unfortunately, none of the methods worked. I really need some help.

Thanks!

Rd4dev:
@TanishMoral11, that's nice to explore different approaches, the reason why I feel the second approach may not be ideal is, it completely removes the RTL language support by ignoring the check for `isRtl` support in `ListItemLeadingMarginSpan` across all existing pages, regardless of locale which shouldn't be the case. This can be visually observed if navigated to `First Test Topic` -> `Revision tab` -> `Test subtopic card`.

| Existing RTL list rendering | Approach2 affecting RTL list render |
|:---:|:---:|
|  |  |
|  |  |

While the policy and terms page seem to work as expected with the implementation applied, it negatively affects the handling of lists across other pages.

Additionally the `LeftAlignedSymbolsSpan` doesn't seem to have any effect since the parsed text doesn't include ("•") as assumed; the texts are parsed from `<li>`, `<ul>`, and `<ol>` tags.

If possible and haven't yet gone through this blog - https://medium.com/swlh/making-nested-lists-with-android-spannables-in-kotlin-4ad00052912c, do check this out, it explains the essential workings of the functionality for further context.

<h1></h1>

I also tried the PR #5296 implementation locally (which I believe corresponds to Approach 1), and that seems like a valid approach to dynamically support or force LTR rendering for list bulleting and numbering. However, please confirm this with @adhiamboperes.

I ran the tests locally, and they passed fine, but I’m curious about the nature of the test failures. Could you try the first implementation and share what the errors are and if there are any other reasons why approach 1 didn't work? That would be very helpful for understanding the issues.

TanishMoral11:
Hi @adhiamboperes , 
Thank you so much for highlighting this PR in today’s meeting. Could you kindly share Ben’s suggestions for resolving this? I’d really appreciate your guidance.

adhiamboperes:
@TanishMoral11, I have added the proposed solution to the main issue body, PTAL.

TanishMoral11:
Hi @BenHenning @adhiamboperes , PTAL

Thank you so much for the suggestion! I really love the approach you recommended. I made all the changes, and the implementation is working absolutely fine.

After Changes : 

https://github.com/user-attachments/assets/5258c7a8-338f-4603-b072-b22e4e221231

app/src/main/java/org/oppia/android/app/translation/LanguageSource.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+enum class LanguageSource {
+  /**
+   * Indicates that the system language (device default) should be used for localization.
+   */
+  SYSTEM_LANGUAGE,
+
+  /**
+   * Indicates that the app-specific language selected by the user should be used for localization.
+   */
```
adhiamboperes: These look like they could be single line KDocs

domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt
```diff
@@ -532,4 +532,10 @@ class TranslationController @Inject constructor(
       }
     }
   }
+  /**
+   * Returns a data provider for the display locale corresponding to the specified language.
+   */
```
adhiamboperes: Insert newline above. Also, this KDoc looks like it can fit on a single line. Could you please check that?

app/src/main/java/org/oppia/android/app/translation/LanguageSource.kt
adhiamboperes, line 1: I'm thinking about whether this should be a proto enum or a Kotlin enum. Looking at existing implementation patterns, we only use standalone Kotlin enum classes when there are other functions in the class.

app/src/main/java/org/oppia/android/app/activity/InjectableAutolocalizeEnglishOnlyActivity.kt
adhiamboperes, line 1: Rename: -> InjectableAutolocalizedEnglishOnlyActivity.kt (Added d to localize to make it an adjective/descriptive)

adhiamboperes (changes requested):
Thanks for making these changes @TanishMoral11! I'm happy to see that the solution is working well.

I have left some comments, and please update your PR description and recordings in the main PR body as well.

Do also work on adding tests for the classes that are failing coverage.

app/src/main/java/org/oppia/android/app/translation/LanguageSource.kt
TanishMoral11: Sure ..

domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt
TanishMoral11: Done.

app/src/main/java/org/oppia/android/app/activity/InjectableAutolocalizeEnglishOnlyActivity.kt
TanishMoral11: Thanks @adhiamboperes , for pointing that out. Yeah, it makes more sense now. I have renamed the file and updated all references accordingly, as it needed to be.

TanishMoral11:
Hi @adhiamboperes , PTAL.

Thanks !!

TanishMoral11:
Hi @adhiamboperes , PTAL. 
This pr is ready for review.
Thanks!

scripts/assets/test_file_exemptions.textproto
```diff
@@ -34,6 +34,10 @@ test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/activity/InjectableAutoLocalizedAppCompatActivity.kt"
   test_file_not_required: true
 }
+test_file_exemption {
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/activity/InjectableAutolocalizedEnglishOnlyActivity.kt"
```
adhiamboperes: Looks like this activity has been deleted in lates changes ?

adhiamboperes (commented):
Thanks @TanishMoral11! 

The changes in the latest commits have been pretty significant compared to the time of my last review.

I am assigning this to @BenHenning for a full pass.

scripts/assets/test_file_exemptions.textproto
TanishMoral11: Apologies, I just missed it. I have now removed it.

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
```diff
@@ -39,7 +41,7 @@ class AppLanguageWatcherMixin @Inject constructor(
    *
    * @param shouldOnlyUseSystemLanguage whether only the system language should be used
```
BenHenning: Please update this KDoc now that the method signature has changed.

model/src/main/proto/languages.proto
```diff
@@ -37,6 +37,32 @@ enum OppiaLanguage {
...
+// the app's selected language, or enforce a specific language such as English.
+enum ActivityLanguageSource {
+  // Corresponds to an unspecified or default language source. When this value is used, the app may
+  // fall back to a default behavior, such as using the system language or another predefined
+  // default, depending on the context in which it is applied.
+  ACTIVITY_LANGUAGE_SOURCE_UNSPECIFIED = 0;
+
+  // Indicates that the activity should use the system's default language, as configured by the user
+  // in the device's settings. This aligns with the system's locale settings and reflects the user's
+  // preferred language on the device.
```
BenHenning: Suggest adding something like:

```
Note that this may be a necessary mode in cases when an activity cannot load user settings (such as in certain test or early app lifecycle scenarios).
```

(Needs wrapping)

This helps clarify when to use this setting over `USE_APP_LANGUAGE`.

model/src/main/proto/languages.proto
```diff
@@ -37,6 +37,32 @@ enum OppiaLanguage {
...
+  USE_SYSTEM_LANGUAGE = 1;
+
+  // Indicates that the activity should use the language selected within the Oppia app. This may
+  // differ from the system's language if the user has chosen a specific language through the app's
+  // settings, providing flexibility for app-specific localization.
```
BenHenning, line 57: Suggest adding:

```
Note that this should generally be the default mode used.
```

(May need wrapping)

Since we essentially never want anyone to use the other two unless they are actually necessary.

model/src/main/proto/languages.proto
```diff
@@ -37,6 +37,32 @@ enum OppiaLanguage {
   NIGERIAN_PIDGIN = 8;
 }
 
+// Specifies the source of the language to be used for localizing activities in the Oppia Android app.
+// This enum determines whether an activity should derive its language from the system settings,
+// the app's selected language, or enforce a specific language such as English.
+enum ActivityLanguageSource {
```
BenHenning: Perhaps instead: `ForcedActivityLanguageMode`? This is actually a very 'strong' setting in the sense that it's overriding default Android OS behavior for the activity (in order to make the in-app language selection work correctly).

domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt
```diff
@@ -532,4 +532,9 @@ class TranslationController @Inject constructor(
       }
     }
   }
+
+  /** Returns a data provider for the display locale corresponding to the specified language. */
+  fun getLocaleFor(language: OppiaLanguage): DataProvider<OppiaLocale.DisplayLocale> {
+    return localeController.retrieveAppStringDisplayLocale(language)
```
BenHenning: Could we perhaps use the `LocaleController` directly for this situation? We have good reason to always go through `TranslationController` but this is a special one-off case, and it seems simpler than expanding the API of `TranslationController` (which would require additional, possibly redundant tests to be added in `TranslationControllerTest`).

app/src/test/java/org/oppia/android/app/translation/AppLanguageWatcherMixinTest.kt
```diff
@@ -229,10 +241,21 @@ class AppLanguageWatcherMixinTest {
     }
   }
 
+  @Test
+  fun testMixin_initialized_withEnglish_setsEnglishLocale() {
```
BenHenning, line 246: Here & everywhere else in the test: could you maybe force the system locale to be something other than English? Otherwise it's hard to tell whether the `USE_ENGLISH` mode is actually working correctly.

I suggest picking something none of the tests are verifying such as Swahili.

This would lead to a change in the system language test below, as well as the new app language test you added above, but this seems fine since it essentially will demonstrate exactly that the system language is being selected (Swahili) except for this case where 'USE_ENGLISH' forces the system locale to be ignored.

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
```diff
@@ -68,10 +70,18 @@ class AppLanguageWatcherMixin @Inject constructor(
...
+      ActivityLanguageSource.USE_APP_LANGUAGE -> {
+        if (currentUserProfileId != null) {
+          translationController.getAppLanguageLocale(currentUserProfileId)
+        } else {
+          translationController.getSystemLanguageLocale()
+        }
+      }
+      ActivityLanguageSource.USE_ENGLISH ->
+        translationController.getLocaleFor(OppiaLanguage.ENGLISH)
+      else -> translationController.getSystemLanguageLocale()
```
BenHenning, line 88: Probably worth adding a test for this case (the unspecified mode situation).

app/src/main/java/org/oppia/android/app/policies/PoliciesActivity.kt
```diff
@@ -33,6 +35,10 @@ class PoliciesActivity :
     )
   }
 
+  override fun initializeMixin(appLanguageWatcherMixin: AppLanguageWatcherMixin) {
+    appLanguageWatcherMixin.initialize(ActivityLanguageSource.USE_ENGLISH)
+  }
```
BenHenning: Is it possible to test this? Maybe by verifying that the content is in English and left aligned even when the system locale is Arabic?

It would be very nice to ensure this test fails without this initialization override in place, and passes with it.

BenHenning (changes requested):
Thanks @TanishMoral11! This is _really_ well done. I have a bunch of comments to clarify or improve testing (mostly because this is a very core and complex system, so it's good to try and be very thorough), but the foundation in the latest PR version is quite good. This seems like a really solid approach.

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
TanishMoral11: Done.

model/src/main/proto/languages.proto
TanishMoral11: Done.

model/src/main/proto/languages.proto
TanishMoral11: Done.

model/src/main/proto/languages.proto
TanishMoral11: Yeah, that makes more sense. Thank you for the suggestion!

domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt
TanishMoral11: Done.

app/src/test/java/org/oppia/android/app/translation/AppLanguageWatcherMixinTest.kt
TanishMoral11: Hi @BenHenning,

I’ve tested setting the system locale to Swahili as you suggested. The `testMixin_initialized_withEnglish_setsEnglishLocale` test passes, confirming that `USE_ENGLISH` correctly overrides the system locale from Swahili to English.

<details>
<summary>Test Details</summary>

Image

</details>

However, switching the system locale to Swahili causes three existing tests to fail:

1. `testMixin_initialized_noAppLanguageChange_doesNothing`: The recreation count is 1 (expected 0).
2. `testMixin_initialized_withAppLanguageChange_sameLanguage_doesNotRecreateActivity`: The recreation count is 1 (expected 0).
3. `testMixin_initialized_withAppLanguageChange_newLanguage_updatesLocale`: The locale remains English (expected Brazilian Portuguese).

These failures seem to stem from how `USE_APP_LANGUAGE` interacts with the system locale change. Given this, should I revert the `DefineAppLanguageLocaleContext` back to English to maintain test stability, or do you recommend another approach to handle these issues?

Thanks!

domain/src/main/java/org/oppia/android/domain/translation/TranslationController.kt
TanishMoral11: Done.

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
TanishMoral11: Done.

app/src/main/java/org/oppia/android/app/policies/PoliciesActivity.kt
TanishMoral11: Done.

TanishMoral11:
Hi @BenHenning , PTAL.
Thanks!

TanishMoral11:
Hi @BenHenning , PTAL.
Thanks!

app/src/sharedTest/java/org/oppia/android/app/policies/PoliciesActivityTest.kt
```diff
@@ -208,6 +209,30 @@ class PoliciesActivityTest {
     }
   }
 
+  @Test
+  fun testPoliciesActivity_withArabicSystemLocale_contentIsInEnglishAndLeftAligned() {
```
BenHenning, line 214: This test doesn't actually seem to be setting up an Arabic system locale. Did you verify that this fails correctly with your fix not in place (i.e. reverting the change to `PoliciesActivity` such that it uses the system language instead of forcing English)?

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
```diff
@@ -37,9 +39,12 @@ class AppLanguageWatcherMixin @Inject constructor(
...
    * situations.
    *
-   * @param shouldOnlyUseSystemLanguage whether only the system language should be used
+   * @param languageSource specifies the source of the language to use for this activity:
+   *   - `USE_SYSTEM_LANGUAGE`: Use the system's default language, as set in the device's settings.
+   *   - `USE_APP_LANGUAGE`: Use the language selected in the app's settings, if available; otherwise, fall back to the system language.
+   *   - `USE_ENGLISH`: Always use English, overriding both system and app language settings.
    */
-  fun initialize(shouldOnlyUseSystemLanguage: Boolean) {
+  fun initialize(languageSource: ForcedActivityLanguageMode) {
```
BenHenning: Nit: should be 'languageMode' per the proto rename. Ditto for the KDoc reference.

app/src/test/java/org/oppia/android/app/translation/AppLanguageWatcherMixinTest.kt
BenHenning: We definitely should use a non-English system locale here, so please go ahead and make that change.

As for the failing tests, I'll need to see the full stack traces for each of the failures which I can take a look at once CI runs with the follow-up changes.

BenHenning (commented):
Thanks @TanishMoral11! Just a few follow-up comments, though it seems that you may need some help with investigating new failures after changing the default system locale for the mixin tests. I do recommend continuing to try and fix those after you update the PR in case you find a fix before I'm able to look at the CI results next.

app/src/main/java/org/oppia/android/app/translation/AppLanguageWatcherMixin.kt
TanishMoral11: Done.

PR #5614, opened 2024-12-22
Fix #4848: Ensure Complete Content Descriptions for Custom HTML Tags in CustomHtmlContentHandler

### LINKED ISSUE #4848: Fix HTML generation for content descriptions (particularly for answers)
This was discovered as part of #4846. We currently generate content descriptions by parsing HTML and then converting that to a string (which strips the spans from the generated HTML). The problem with this approach is it leads to the custom tag text also being stripped out, which leads to incomplete strings like this:

```
Remember that two halves, when added together, make one whole. Click on this .
```

(Notice the lack of context before the '.').

The correct version would actually be:

```
Remember that two halves, when added together, make one whole. Click on this test_skill_id_1 concept card.
```

We probably need to introduce a custom generation pipeline for content descriptions in CustomHtmlContentHandler to ensure these cases can be properly covered. We also should be checking for other link cases, including anchors, or any case when text can appear between tags.

manas-yu:
## Explanation

Fix #4848: This PR resolves the issue with incomplete content descriptions being generated for custom HTML tags. The existing implementation stripped out spans and custom tag content, leading to incomplete descriptions.

The updated implementation introduces a custom generation pipeline for content descriptions in `CustomHtmlContentHandler`. It ensures that all custom tag content is correctly processed and appended, resulting in complete and accurate descriptions.

### Key Changes:
- Added a `StringBuilder` to accumulate content descriptions and a `MutableMap` to handle custom tag descriptions.
- Implemented `getContentDescription()` to combine default text with custom tag content descriptions, ensuring seamless integration.
- Introduced a `ContentDescriptionProvider` interface for handling custom tags.
- Updated relevant tag handlers (`ImageTagHandler`, `ConceptCardTagHandler`, `LiTagHandler`, `PolicyPageTagHandler` and `MathTagHandler`) to implement the new interface.
- Verified the implementation with appropriate test cases.

## Essential Checklist


adhiamboperes:
@manas-yu, please assign to a reviewer if ready for a pass.

manas-yu:
> @manas-yu, please assign to a reviewer if ready for a pass.

Hey @adhiamboperes PTAL, the PR is now ready for review. I've made the necessary code changes and would appreciate your feedback on the approach. Let me know if you have any suggestions or if any improvements are needed. Thanks!

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
```diff
@@ -101,4 +101,11 @@ class ImageTagHandler(
       "Failed to parse $CUSTOM_IMG_CAPTION_ATTRIBUTE"
     )
   }
+
+  override fun getContentDescription(attributes: Attributes): String? {
+    val altValue = attributes.getJsonStringValue("alt-with-value")
```
adhiamboperes:
```suggestion
    val altValue = attributes.getJsonStringValue(CUSTOM_IMG_ALT_TEXT_ATTRIBUTE)
```

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
```diff
@@ -101,4 +101,11 @@ class ImageTagHandler(
...
+    return if (altValue != null) {
```
adhiamboperes: Prefer using `isNullOrBlank()`, which is a more strict check and will prevent issues like https://github.com/oppia/lessons-team/issues/719 from re-curring.

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
```diff
@@ -101,4 +101,11 @@ class ImageTagHandler(
...
+      "Image illustrating $altValue"
+    } else null
```
adhiamboperes: Would it be better to return an empty string?

utility/src/test/java/org/oppia/android/util/parser/html/ConceptCardTagHandlerTest.kt
```diff
@@ -116,6 +116,16 @@ class ConceptCardTagHandlerTest {
     val clickableSpans = parsedHtml.getSpansFromWholeString(ClickableSpan::class)
     assertThat(clickableSpans).hasLength(1)
   }
+  @Test
```
adhiamboperes, line 120: Insert new line between tests

utility/src/test/java/org/oppia/android/util/parser/html/CustomHtmlContentHandlerTest.kt
```diff
@@ -298,6 +298,107 @@ class CustomHtmlContentHandlerTest {
...
+      imageRetriever = mockImageRetriever,
+      customTagHandlers = mapOf(
+        "first-tag" to firstHandler,
+        "second-tag" to secondHandler
+      )
+    )
+
+    assertThat(contentDescription).isEqualTo("Start First one middle Second two end")
+  }
+  @Test
```
adhiamboperes, line 334: Please format this tes class correctly.

utility/src/test/java/org/oppia/android/util/parser/html/PolicyPageTagHandlerTest.kt
```diff
@@ -120,6 +120,20 @@ class PolicyPageTagHandlerTest {
 
     assertThat(policyTypeCaptor.value).isEqualTo(PolicyType.PRIVACY_POLICY)
   }
+  @Test
```
adhiamboperes, line 124: Insert newline above

utility/src/main/java/org/oppia/android/util/parser/html/PolicyPageTagHandler.kt
```diff
@@ -83,4 +83,14 @@ class PolicyPageTagHandler(
...
+
+  override fun getContentDescription(attributes: Attributes): String? {
+    val linkType = attributes.getJsonStringValue("link") ?: return null
+
+    return when (linkType) {
+      TERMS_OF_SERVICE_PAGE -> "Link to "
+      PRIVACY_POLICY_PAGE -> "Link to "
+      else -> null
+    }
+  }
```
adhiamboperes: Could be simplified to:
```suggestion
 override fun getContentDescription(attributes: Attributes): String? {
    return when (attributes.getJsonStringValue("link")) {
        TERMS_OF_SERVICE_PAGE, PRIVACY_POLICY_PAGE -> "Link to "
        else -> null
    }
}
```

adhiamboperes (changes requested):
Thanks @manas-yu! PTAL at some suggestions that I have left.

utility/src/main/java/org/oppia/android/util/parser/html/ImageTagHandler.kt
manas-yu: Thanks for pointing it out. Returning an empty string instead of a null one is safer and avoids null checks, so I'll make the change.

manas-yu:
Hey @adhiamboperes, PTAL, I have made the requested changes. Let me know if there is anything else.

adhiamboperes (approved):
Thanks @manas-yu! This LGTM!

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -25,9 +25,17 @@ class CustomHtmlContentHandler private constructor(
...
+  private val tagContentDescriptions = mutableMapOf<Int, String>()
+  private var isInListItem = false
+  private var pendingNewline = false
+  private val blockTags = setOf("p", "ol", "ul", "li", "oppia-ul", "oppia-ol", "oppia-li", "div")
 
   override fun endElement(uri: String?, localName: String?, qName: String?) {
     originalContentHandler?.endElement(uri, localName, qName)
+    if (localName in blockTags) {
+      isInListItem = false
+    }
```
adhiamboperes, line 38: This could have been:
```kotlin
val tagName = qName ?: localName  // Fallback to localName if qName is null
   if (tagName in blockTags) {
        isInListItem = false
    }
```

**qName** (qualified name) is generally more reliable unless your parser enforces namespaces.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -45,6 +53,11 @@ class CustomHtmlContentHandler private constructor(
 
   override fun characters(ch: CharArray?, start: Int, length: Int) {
     originalContentHandler?.characters(ch, start, length)
+    if (pendingNewline) {
+      contentDescriptionBuilder.append('\n')
+      pendingNewline = false
+    }
+    ch?.let { contentDescriptionBuilder.append(String(it, start, length)) }
```
adhiamboperes, line 60: `    ch?.let { contentDescriptionBuilder.appendRange(it, start, start + length) }`

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -25,9 +25,17 @@ class CustomHtmlContentHandler private constructor(
   private var originalContentHandler: ContentHandler? = null
   private var currentTrackedTag: TrackedTag? = null
   private val currentTrackedCustomTags = ArrayDeque<TrackedCustomTag>()
+  private val contentDescriptionBuilder = StringBuilder()
+  private val tagContentDescriptions = mutableMapOf<Int, String>()
+  private var isInListItem = false
+  private var pendingNewline = false
```
adhiamboperes, line 31: I'm not sure what pending new line here means, maybe a rename or adding a comment might clear things up. I had a bit of trouble understanding how it is used in `override fun characters` below.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -196,6 +249,24 @@ class CustomHtmlContentHandler private constructor(
   }
 
   companion object {
+    /**
+     * Returns the content description for the HTML content, processing all custom tags that implement
+     * [ContentDescriptionProvider].
+     */
+    fun <T> getContentDescription(
```
adhiamboperes, line 256: I noticed that while this function was created, it was not used as required by the issue description. The only present usage is in `HintViewModel`, which means descriptions will be geneerated for Hints only. I believe that other content viewmodels need to use this too, in order to correctly generate descriptions.

According to https://github.com/oppia/oppia-android/issues/4848, we should ensure descriptions are generated for answers,  link cases, including anchors, or any case when text can appear between tags. This means this new function should be used with `ContentViewModel` and `SelectionInteractionContentViewModel`.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
```diff
@@ -196,6 +249,24 @@ class CustomHtmlContentHandler private constructor(
...
+      imageRetriever: T?,
+      customTagHandlers: Map<String, CustomTagHandler>
+    ): String where T : Html.ImageGetter, T : ImageRetriever {
+      val handler = CustomHtmlContentHandler(customTagHandlers, imageRetriever)
+      HtmlCompat.fromHtml(
+        "<init-custom-handler/>$html",
+        HtmlCompat.FROM_HTML_MODE_LEGACY,
+        imageRetriever,
+        handler
+      )
```
adhiamboperes, line 267: This code appears unused, what does it do?

adhiamboperes (commented):
Did a post-merge review. I have some questions. PTAL @manas-yu.

utility/src/main/java/org/oppia/android/util/parser/html/LiTagHandler.kt
```diff
@@ -23,7 +24,7 @@ const val CUSTOM_LIST_OL_TAG = "oppia-ol"
 class LiTagHandler(
   private val context: Context,
   private val displayLocale: OppiaLocale.DisplayLocale
-) : CustomHtmlContentHandler.CustomTagHandler {
+) : CustomHtmlContentHandler.CustomTagHandler, CustomHtmlContentHandler.ContentDescriptionProvider {
```
adhiamboperes, line 27: Since `getContentDescription` doesn't do anything here, I don't think
`CustomHtmlContentHandler.ContentDescriptionProvider` should be implemented here.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
manas-yu: The code is not unused, It triggers the HTML parsing process, allowing `CustomHtmlContentHandler` to intercept tags, extract content, and populate `contentDescriptionBuilder` to generate the final content description.

This approach leverages Android’s built-in HTML parsing system, ensuring compatibility with the framework’s expectations, and allows the reuse of the same parsing logic for both content description and display purposes.
 
I'll add comments explaining this and the `pendingNewline` in the PR to resolve:

> This means this new function should be used with ContentViewModel and SelectionInteractionContentViewModel.

utility/src/main/java/org/oppia/android/util/parser/html/CustomHtmlContentHandler.kt
manas-yu: It acts as a flag that indicates whether a newline character should be inserted before the next text content. It's used when dealing with block-level HTML elements

Issue #2133, opened 2020-11-20
Input Keyboard Issues


mschanteltc:
- [ ] When the input is selected, scrolling up when the input bar is no longer visible deselects it. Tapping on the keyboard does not enter numbers anywhere but selected numbers on the keypad become bolder. (Image below)[[Video](https://drive.google.com/file/d/1f_1geqB8qaa95xwuKwSU8ya6aMQHiM8v/view?usp=sharing)]

- [x] When input is entered, and the screen is rotated, the entered input is deleted. [[Video](https://drive.google.com/file/d/1f0GYDh1RtaaYhspmj8d9IwXB85Lgnx0E/view?usp=sharing)] Fixed in https://github.com/oppia/oppia-android/pull/5617

- [x] When creating a 3-Digit Input, tapping "Next" after entering the first PIN does not change the input to the second PIN input. Instead, it changes to an "Enter" button that does nothing. [[Video](https://drive.google.com/file/d/1gQ2bznLJtQx15Ii32PQFx-njqqDA8UJs/view?usp=sharing)] Fixed

![Screenshot_20201120-005025_Oppia](https://user-images.githubusercontent.com/22177674/99841268-bb86fb00-2b22-11eb-9a1f-669a8776d95b.jpg)

Device: Might be Samsung device

krishnakeshan:
Could you specify the steps to navigate to the screen you mentioned in the issue?

rt4914:
> Could you specify the steps to navigate to the screen you mentioned in the issue?

@krishnakeshan Welcome to oppia. 
Actually this is something that you won't be able to check on current code. Therefore I am assigning this to myself.

Also, before we start please checkout our [wiki](https://github.com/oppia/oppia-android/wiki) first. This will help you in setting up the project and once you sign the CLA and fill the survey form we can assign you issues to work on.

krishnakeshan:
Hi @rt4914 thanks! Oh alright got it, no worries.

I've signed the CLA and filled out the survey, and have a build running on a device. I'm checking out the wiki right now!

rt4914:
> Hi @rt4914 thanks! Oh alright got it, no worries.
> 
> I've signed the CLA and filled out the survey, and have a build running on a device. I'm checking out the wiki right now!

@krishnakeshan Can you please check if the CLA and the survey was filled successfully because I am not able to find you details in our records.

krishnakeshan:
> > Hi @rt4914 thanks! Oh alright got it, no worries.
[...]

PR #5458, opened 2024-07-06
Fix part of #4470, Fix #4471, Fix 4474: Handle configuration change using onSavedInstance.

Vishwajith-Shettigar:
## Explanation
Fix part of #4470. 
Fixes #4471.
Fixes #4474

This PR enables the retention of input when the device configuration changes using onSavedInstance.

List of interactions covered in this PR:
1. FractionInputInteraction
2. NumericInputInteraction
3. TextInputInteraction
4. MathExpressionInteraction
5. RatioExpressionInteraction

List of interactions not covered in this PR (Covered here #5478):
1. Image Region selection interaction.
2. Drag and Drop interaction.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

Issue #4986, opened 2023-05-17
Move all fragment arguments, activity intent extras, and saved instance state over to protos


BenHenning:
**Is your feature request related to a problem? Please describe.**
Bundles do not provide strong typing or presence guarantees, and require unique strings to identify per-fragment and per-activity values.

**Describe the solution you'd like**
Protos help alleviate this problem and can be serialized directly into bundles. Protos also allow for more trivial composition (e.g. when passing data from fragments to activities), and eventually can simplify both data processing (#1051) and on-disk storage (e.g. in the case of #2571), though the latter is less important since Android already serializes bundles on-disk. This will also help with the mechanisms needed for #309 to be possible.

The solution here is to move *all* fragment arguments, activity extras, and saved instance state bundle usage over to using protos entirely.

**Describe alternatives you've considered**
The only other obvious solution is to continue with bundles and not solve the original problem that's mentioned. There are other possible serialization intermediaries we could consider (such as JSON), but protos make the most sense given that they're already used broadly in the codebase.

**Additional context**
None.

PR #5617, opened 2024-12-24
Fixes a part of #2133: Prevent vanishing of input on configuration change.

theayushyadav11:
Fixes a part of #2133 and a part of #5818 .Prevent vanishing of input on configuration change from `PinPasswordActivity`
This PR addresses **#2133** and **#5818**and fixes the issue where the entered PIN would disappear after a configuration change (like rotating the device) in `PinPasswordActivity`, also modifies the logic for the function `appendText` in `EditTextInputAction`.

When the device was rotated, the PIN input would vanish. This happened because:

- The `android:maxLength` was bound to `viewModel.profile.pin.length()`.
- Since `profile` is a `LiveData`, on rotation, it would reset, causing `pin.length()` to evaluate to 0.
- This resulted in the `maxLength` being set to 0, and the input field was cleared.

#### Solution

- Introduced a new observable field `maxPinLength` with a default value of 5:
  ```kotlin
  val maxPinLength = ObservableField<Int>(5)
  ```
Here the default value is 5 so no input will be removed or trimmed off ehether it is `adminPin` or `userPin`.
By decoupling maxLength from profile, it prevents the input field from clearing when the configuration changes.
#####  Manual Testing:
Entered a PIN, rotated the device multiple times, and ensured that the input persisted and was not cleared for both user and admin.
#####   Automated Testing:
Added new tests to verify the correct working of the changes made.
Added a new test to PinPasswordActivityTest to verify that the entered PIN is correctly persisted across configuration changes. (PinPasswordActivityTest.kt)
[app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt](https://github.com/oppia/oppia-android/pull/5617/files#diff-d8f7fb43f0747dfb7cf7c49b873a2043017b25b64dce48dd74a52d7115259878R1217-R1234)
#### Screenshot of the passing tests:
Screenshot from 2025-04-20 02-43-50
#### Before

https://github.com/user-attachments/assets/a422ca5e-eb1a-4e27-86ad-b81947000fbd

#### After

https://github.com/user-attachments/assets/58000383-45b3-4889-8cb6-88784dbbbbef

## Essential Checklist


theayushyadav11:
@adhiamboperes PTAL.

app/src/main/AndroidManifest.xml
```diff
@@ -154,6 +154,7 @@
       android:windowSoftInputMode="adjustResize" />
     <activity
       android:name=".app.profile.PinPasswordActivity"
+      android:configChanges="screenSize|orientation"
```
adhiamboperes: I'm not sure whether this manifest change is required. Could you please explain the need?

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -247,6 +249,14 @@ class PinPasswordActivityPresenter @Inject constructor(
       exitProcess(0)
     }
   }
+  fun handleSaveInstanceState(outState: Bundle) {
+    val enteredPin = binding.pinPasswordInputPinEditText.text.toString()
+    outState.putString("enteredPin", enteredPin)
+  }
```
adhiamboperes: There are a couple of changes needed here in order to conform with Oppia Android coding patterns:
1. We are nolonger using bundle functions such as getString(See https://github.com/oppia/oppia-android/issues/4986). Instead, we are moving over to protos. e.g:
 `outState.putProto(
      PINPASSWORD_ACTIVITY_STATE_KEY,
      args
    )`
2. We prefer using constants, such as PINPASSWORD_ACTIVITY_STATE_KEY, instead of string literals like "enteredPin".
3. Go to `arguments.proto`, and add a new proto message for `PinPasswordActivityStateBundle` please look ad the code in that file for examples.

To retrieve the state, use:
`outState.getProto(
        PINPASSWORD_ACTIVITY_STATE_KEY,
        pinPasswordActivityStateBundle.getDefaultInstance()
      )`

adhiamboperes (changes requested):
Thanks @theayushyadav11. PTAL at the changes I have requested.

adhiamboperes:
Hi @theayushyadav11, could you please provide an update on how this is going?

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Thanks @adhiamboperes , I'll implement the changes.

app/src/main/AndroidManifest.xml
theayushyadav11: Thanks @adhiamboperes , I have tested without changing the manifest file but on configuration change changes does not persist as mentioned in the issue. I have tested many approaches like  LiveData with the ViewModel but only this method could solve the issue.
Can you please suggest me any better approach so I can implement the same.

theayushyadav11:
Hi @adhiamboperes , I was a bit stuck in the setup on the new device so I couldn't work on the issue.

adhiamboperes:
@theayushyadav11, will you be continuing to work on this issue or should we reassign to someone else?

theayushyadav11:
Hey @adhiamboperes , I'm currently closing this PR and unassigning myself as I'm working on other issues and not available to actively work on this one at the moment. If someone else takes it up, that would be great, but if it's still open later, I can revisit it when I have more bandwidth .

theayushyadav11:
Hi  @adhiamboperes , @Rd4dev In this approach I have used `onSaveInstanceState` to save the pin on configuration change using `proto`. The value is being saved successfully but it is not able to reflect in `pinPasswordInputPinEditText`. Can you PTAL and help me out in this.
Thanks

#### Successful toast:
https://github.com/user-attachments/assets/7f8f1b56-9b20-48b5-9432-d82dc33c3b97

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -68,6 +75,16 @@ class PinPasswordActivityPresenter @Inject constructor(
       (activity as PinPasswordActivity).finish()
     }
 
+    outState?.getProto(
+      PINPASSWORD_ACTIVITY_STATE_KEY,
+      PinPasswordActivityStateBundle.getDefaultInstance()
+    )?.inputPin?.let { savedInputPin ->
+      if (savedInputPin.isNotEmpty()) {
+        Toast.makeText(activity, savedInputPin.toString(), Toast.LENGTH_SHORT).show()
+        binding.pinPasswordInputPinEditText.setText(savedInputPin)
```
Rd4dev: The presenter should only receive the restored data. The activity should handle the `savedInstanceState` and only pass in the pin using handleOnCreate(pin). Refer to: [MarkTopicsCompletedFragement](https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/java/org/oppia/android/app/devoptions/marktopicscompleted/MarkTopicsCompletedFragment.kt#L19)

https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/java/org/oppia/android/app/devoptions/marktopicscompleted/MarkTopicsCompletedFragment.kt#L56-L69

model/src/main/proto/arguments.proto
```diff
@@ -947,6 +947,11 @@ message ProfileChooserFragmentArguments {
   ProfileType profile_type = 1;
 }
 
+// The bundle of properties that are saved on configuration change in PinPasswordActivity.
+message PinPasswordActivityStateBundle {
+  string input_pin = 1;
```
Rd4dev: Include a brief comment for the field.

Rd4dev (commented):
@theayushyadav11 -- You can make use of data binding — for example, see how it's implemented in [numeric_input_interaction_item](https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/res/layout/numeric_input_interaction_item.xml#L29). 

https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/res/layout/numeric_input_interaction_item.xml#L29-L34

For this, the `PinPasswordViewModel` would need to introduce a new field. This field might need to be set from the presenter after the **view is fully laid out** to prevent it from being overwritten. You can also use `TextWatcher` to listen for text changes (eg: [`AdminAuthViewModel`](https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/java/org/oppia/android/app/profile/AdminAuthViewModel.kt#L12C7-L12C25)` and update the value accordingly which can also be used to save states.

You can also refer to PR #5458, which handled data retention for most other interaction types

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -68,6 +75,16 @@ class PinPasswordActivityPresenter @Inject constructor(
       (activity as PinPasswordActivity).finish()
     }
 
+    outState?.getProto(
+      PINPASSWORD_ACTIVITY_STATE_KEY,
+      PinPasswordActivityStateBundle.getDefaultInstance()
+    )?.inputPin?.let { savedInputPin ->
+      if (savedInputPin.isNotEmpty()) {
+        Toast.makeText(activity, savedInputPin.toString(), Toast.LENGTH_SHORT).show()
```
adhiamboperes: We don't use toasts at all in the project. If the savedstate is null, then the pin textview should just be empty.

adhiamboperes (changes requested):
Thanks @theayushyadav11!

Please see inline for comments. Additionall,
- Please add UI tests for these changes
- Update the PR description, it seems that some of the content is out of date.

model/src/main/proto/arguments.proto
theayushyadav11: Done.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Changed.

theayushyadav11:
> @theayushyadav11 -- You can make use of data binding — for example, see how it's implemented in [numeric_input_interaction_item](https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/res/layout/numeric_input_interaction_item.xml#L29).
> 
> https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/res/layout/numeric_input_interaction_item.xml#L29-L34
> 
> For this, the `PinPasswordViewModel` would need to introduce a new field. This field might need to be set from the presenter after the **view is fully laid out** to prevent it from being overwritten. You can also use `TextWatcher` to listen for text changes (eg: [`AdminAuthViewModel`](https://github.com/oppia/oppia-android/blob/d67f6dcff48b571bbd132e7e1bb4176a1d96446f/app/src/main/java/org/oppia/android/app/profile/AdminAuthViewModel.kt#L12C7-L12C25)` and update the value accordingly which can also be used to save states.
> 
> You can also refer to PR #5458, which handled data retention for most other interaction types

Hi @Rd4dev I have implemented the same but the input still doesn't persists. PTAL at the current code.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -28,6 +28,7 @@ import kotlin.system.exitProcess
 
 private const val TAG_ADMIN_SETTINGS_DIALOG = "ADMIN_SETTINGS_DIALOG"
 private const val TAG_RESET_PIN_DIALOG = "RESET_PIN_DIALOG"
+const val PINPASSWORD_ACTIVITY_STATE_KEY = "PINPASSWORD_ACTIVITY_STATE_KEY"
```
Rd4dev: This should be defined in the Activity.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -43,8 +44,9 @@ class PinPasswordActivityPresenter @Inject constructor(
   private var profileId = ProfileId.getDefaultInstance()
   private lateinit var alertDialog: AlertDialog
   private var confirmedDeletion = false
+  private lateinit var binding: PinPasswordActivityBinding
```
Rd4dev: Nit: Nothing concerning, but could you move it below the other lateinit definition? Just to keep things a bit more organized.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -249,6 +257,9 @@ class PinPasswordActivityPresenter @Inject constructor(
     }
   }
 
+  fun getInputPin() =
```
Rd4dev: Can you explicitly specify the return type for `getInputPin()`? 

>While type inference works here, being explicit could help with readability, refactoring safety, dev experience in IDE and maintaining consistency across the codebase.

app/src/main/res/layout/pin_password_activity.xml
```diff
@@ -122,7 +122,8 @@
                 android:layout_height="wrap_content"
                 android:imeOptions="actionNext"
                 android:inputType="numberPassword"
-                android:maxLength="@{viewModel.profile.pin.length()}"
+                  android:text="@={viewModel.inputPin}"
+                  android:maxLength="@{viewModel.profile.pin.length()}"
```
Rd4dev: Nit: Indentation

```suggestion
                android:text="@={viewModel.inputPin}"
                android:maxLength="@{viewModel.profile.pin.length()}"
```

model/src/main/proto/arguments.proto
```diff
@@ -947,6 +947,12 @@ message ProfileChooserFragmentArguments {
   ProfileType profile_type = 1;
 }
 
+// The bundle of properties that are saved on configuration change in PinPasswordActivity.
+message PinPasswordActivityStateBundle {
```
Rd4dev: Can you place it next to `PinPasswordActivityParams` to maintain a clear structure?

model/src/main/proto/arguments.proto
Rd4dev: Please add spacing after `//`.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivity.kt
```diff
@@ -51,6 +58,17 @@ class PinPasswordActivity :
...
 
+  override fun onSaveInstanceState(outState: Bundle) {
+    super.onSaveInstanceState(outState)
+    val args = PinPasswordActivityStateBundle.newBuilder()
+      .setInputPin(pinPasswordActivityPresenter.getInputPin())
+      .build()
+    outState.putProto(
+      PINPASSWORD_ACTIVITY_STATE_KEY,
+      args
+    )
```
Rd4dev: Could this be a single line?

```suggestion
    outState.putProto(PINPASSWORD_ACTIVITY_STATE_KEY,args)
```

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivity.kt
```diff
@@ -40,7 +43,11 @@ class PinPasswordActivity :
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     (activityComponent as ActivityComponentImpl).inject(this)
-    pinPasswordActivityPresenter.handleOnCreate()
+    val savedPin = savedInstanceState?.getProto(
+      PINPASSWORD_ACTIVITY_STATE_KEY,
+      PinPasswordActivityStateBundle.getDefaultInstance()
+    )?.inputPin
```
Rd4dev: Instead of passing a nullable value, consider setting a default empty string as a fallback.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -67,6 +69,12 @@ class PinPasswordActivityPresenter @Inject constructor(
     binding.pinPasswordToolbar.setNavigationOnClickListener {
       (activity as PinPasswordActivity).finish()
     }
+    savedPin?.let { pin ->
```
Rd4dev: Ensure this is set after the views are fully laid out.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -67,6 +69,12 @@ class PinPasswordActivityPresenter @Inject constructor(
...
+      if (pin.isNotEmpty()) {
```
Rd4dev: I believe it should be safe to skip the empty check.

Rd4dev (commented):
> Hi @Rd4dev I have implemented the same but the input still doesn't persists. PTAL at the current code.

Thanks for trying out the approach! Were you able to explore why the update didn’t reflect? Knowing what you’ve already inspected could help guide things better.  

In particular, did you get a chance to look into this point: 
> *“This field might need to be set from the presenter after the view is fully laid out to prevent it from being overwritten.”*  

I suspect, the value might be getting set too early — before the view is fully laid out, especially after a configuration change. We’ll likely need to defer setting it until the layout is complete.  You can refer to a similar handling for TextView - `topicDescriptionTextView` in `TopicInfoFragmentPresenter`.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Updated.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivity.kt
theayushyadav11: Updated.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivity.kt
theayushyadav11: Updated.

model/src/main/proto/arguments.proto
theayushyadav11: Updated.

app/src/main/res/layout/pin_password_activity.xml
theayushyadav11: Updated.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Updated.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Updated.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
theayushyadav11: Updated.

theayushyadav11:
Thanks @Rd4dev for the review, input is being successfully retained now and tests are passing PTAL.

Rd4dev:
I tried to look into the root cause and observed the following:

EditTexts should generally retain their state across configuration changes. However in [pin_password_activity.xml](https://github.com/oppia/oppia-android/blob/develop/app/src/main/res/layout/pin_password_activity.xml), the `pin_password_input_pin_edit_text` fails to do so due to the interference of the `maxLength` property. 

https://github.com/oppia/oppia-android/blob/545dc0fae9daf65d871b37269890d073bb54b361/app/src/main/res/layout/pin_password_activity.xml#L118-L126

Since `android:maxLength` is dynamically set from `@{viewModel.profile.pin.length}`, where profile is a livedata being populated asynchronously might not be immediately available, leading to `maxLength` being set to 0 initially `<attr name="maxLength" format="integer" min="0" />`, clearing the 'EditText' content on recreation.

Since this seems to be the root cause, I believe we might _not even need to use `onSaveInstanceState`_ to restore the states, especially considering that the layout gets laid out before restoration. This requires use of _post-layout view callbacks_, which is not advised unless absolutely necessary. 

I'm thinking we could introduce a **BindingAdapter** instead to set a custom max length and adjust the input filters based on that, ignoring null, zero, or negative values. This approach would bypass the need for post-layout callbacks and extra save state logic. The default 'EditText' behavior would handle the restoration (as it works fine with [`add_profile_activity.xml`](https://github.com/oppia/oppia-android/blob/545dc0fae9daf65d871b37269890d073bb54b361/app/src/main/res/layout/add_profile_activity.xml#L147), [`profile_reset_pin_fragment`](https://github.com/oppia/oppia-android/blob/545dc0fae9daf65d871b37269890d073bb54b361/app/src/main/res/layout/profile_reset_pin_fragment.xml#L88), etc.). Or would it be more reliable to still proceed with `onSaveInstanceState`? 

@adhiamboperes I wanted to confirm the right approach before moving forward. @theayushyadav11, if you have any suggestions or the best approach, please feel free to share them as well.

<details><summary>Potential alternatives that might not be viable</summary>

- Setting it based on the `isAdmin` field to either 5 or 3, as done in `profile_reset_pin_fragment`, might seem like a possible fix. 

https://github.com/oppia/oppia-android/blob/545dc0fae9daf65d871b37269890d073bb54b361/app/src/main/res/layout/profile_reset_pin_fragment.xml#L94

However, in that scenario, the `profile` is accurately set before being passed to the fragment, ensuring the correct value is reflected. Unfortunately, this approach isn't feasible in our case, as the `profile` is immediately set to `null` in the XML, making it impossible to reliably determine the value until the `profile` is available.

- Defaulting the `maxLength` to 5 would be accurate for admins, as done in [`admin_auth_activity.xml`](https://github.com/oppia/oppia-android/blob/545dc0fae9daf65d871b37269890d073bb54b361/app/src/main/res/layout/admin_auth_activity.xml#L106), but it would apply the same max length for regular users, which is not ideal since their limit is different.

- Removing `maxLength` works as intended since the text listeners still handle the logic to display an error message and clear the input field when the correct pin length is reached, but it allows more than the required number of characters to be inputted.

</details>

theayushyadav11:
Thanks a lot @Rd4dev for the thorough investigation — this was really insightful and clearly outlines the core issue with how `maxLength` is currently being handled.

You're absolutely right in pointing out that directly using  
```xml
android:maxLength="@{viewModel.profile.pin.length()}"
```  
in the XML is problematic. Since `profile` is populated asynchronously, there's a high chance it’s `null` at the time of layout inflation (especially during configuration changes), which causes `maxLength` to be incorrectly evaluated as 0. This inadvertently disables input and leads to unexpected clearing of the EditText content — a critical UX issue.

From what I see, we have two possible paths forward:

### **Approach 1: Introduce a Custom `BindingAdapter` for `maxLength`**

- **Why it works well:**
  - Avoids premature layout evaluation by deferring the logic until a valid value is available.
  - Ensures better input stability during lifecycle changes.
  - Maintains the EditText’s native state restoration, eliminating the need for any manual saving/restoring logic.

###  **Approach 2: Use `@{viewModel.isAdmin ? 5 : 3}`**

- **Why it's not ideal here:**
  - While this works in fragments where `isAdmin` is synchronously available (e.g., passed in via arguments), it’s not reliable in this case.
  - `isAdmin` depends on `profile`, which is a `LiveData`, and may not be initialized in time — leading to race conditions or incorrect UI behavior.

Given the lifecycle and data flow constraints in `pin_password_activity`, I believe **Approach 1 is the most robust and scalable solution** for our use case.

 @adhiamboperes — would love to hear your thoughts before proceeding with the implementation.

Thanks again, and looking forward to your feedback!

BenHenning:
@theayushyadav11 I think that's a sound analysis, but do you have any evidence of approach 2's drawback (that there will be a synchronization issue)? My understanding is that databinding will generate an observer for LiveData variables used in databinding expressions to automatically recompute the layout (and reevaluate expressions) when subscribed LiveData variables change. This observable in a lot of places in code.

There _are_ cases when it fails (it's been a few years, but I recall the case where 2 LiveData variables were used in the same expression had problems resolving), but this case seems simple enough that it shouldn't run into problems. I'd be very interested to know if you are seeing issues with that solution, and what they may be.

Separately, I recommend the following in cases like this:
- If you can identify real solutions, go ahead and pick one and ask for confirmation. It's a good idea to list alternatives and why you ruled them out, but there's no need to wait on an answer if you've already found one that could work. :)
- When you identify more than one solution, pick the one most conformant with existing code. While we do use adapters (approach 1), note that they are specifically for cases where we _can't_ use automatic bindings which _probably_ isn't this case.
- If you've picked a solution that _should_ work (i.e. it's exactly the same as similar solutions, such as your approach 2) but it doesn't, that's a great time to dig in and ask for additional help (such as with a debugging document). Noticing cases like this is what we would consider "code smell" detection: a code pattern that works elsewhere, but not here. Sometimes we may be tempted to pick a different solution that does work, but really we should only rule out the more common pattern when we understand precisely why it doesn't apply in this case. Based on your explanations, approach 2 would be more conformant unless there's a problem with it, and there isn't enough details yet to make that call (at least from the limited context I have on this PR thread).

theayushyadav11:
Thanks so much for the detailed and thoughtful response, @BenHenning — I really appreciate the clarification and your perspective on how to approach these kinds of situations.

I have implemented both the approaches and manually tested them and after manually testing both approaches, I can confirm that they both behave correctly in practice. My initial hesitation with Approach 2 stemmed from the fact that `isAdmin` is an `ObservableField`, and its value is derived from `profile`, which is a `LiveData`. I was concerned that this layered dependency might introduce a synchronization issue.

However, your explanation makes it clear that `DataBinding` does, in fact, observe `LiveData`  variables and re-evaluates expressions as expected, even when they’re indirectly referenced like in this case. That definitely helps clarify things.

Given that:
 - Both approaches are functionally correct after testing.
 - Approach 2 aligns more closely with the current codebase patterns.

I’ll proceed with **Approach 2** for this implementation. It’s great to have your input to help distinguish between an actual issue and a suspected one, and I now better understand when it's appropriate to move forward confidently vs. when deeper investigation is needed.

Thanks again for your guidance — especially the reminder to favor convention when multiple options are valid, and to dig deeper when something breaks an otherwise consistent pattern. That insight will definitely help me think more critically and systematically going forward.

theayushyadav11:
PTAL @adhiamboperes , @Rd4dev .

Rd4dev:
Thanks, @theayushyadav11, for exploring different approaches, and @BenHenning for the helpful insights. 

@theayushyadav11, I had included this use of expression (with current setup)
>android:maxLength="@{viewModel.isAdmin ? 5 : 3}"

 under the "Potentially non-viable" section due to an edge case that will come up.

Here’s the scenario:
1. Log in as an admin.
2. Enter a 4-digit PIN (e.g., 1234).
3. Rotate the device.
4. The 4th digit gets chopped off (to 123).

This happens because profile is initially null during recreation, which causes `isAdmin` to default to `false`. As a result, `maxLength` falls back to 3, and anything beyond the 3rd digit is trimmed—an incorrect behavior for admin PINs.

But I agree with Ben that moving forward with data binding aligns well with the existing codebase, and with the right workaround, the considered option should be achievable.

Could you try testing this flow on your end to see if you can reproduce and share your thoughts?

theayushyadav11:
Thanks @Rd4dev for pointing out this breakage.
I am able to reproduce this issue and this happens because of the same reason as you mentioned above 
> This happens because profile is initially null during recreation, which causes `isAdmin` to default to `false`. As a result, `maxLength` falls back to 3, and anything beyond the 3rd digit is trimmed—an incorrect behavior for admin PINs.
> 
##### Video demo of breakage:

https://github.com/user-attachments/assets/d9b5a780-4fcf-4126-ae5f-e38ccfab4202

 
This was my suspection with this approach that it may fail sometime.
However using `BindingAdapter` this works fine.
```java
@BindingAdapter("safeMaxLength")
public static void setSafeMaxLength(EditText editText, Integer maxLength) {
  if (maxLength != null && maxLength > 0) {
    editText.setFilters(new InputFilter[]{ new InputFilter.LengthFilter(maxLength) });
  } else {
    editText.setFilters(new InputFilter[]{});
  }
}
```
```xml
app:safeMaxLength="@{viewModel.profile.pin.length()}"

```
##### Video demo: 

https://github.com/user-attachments/assets/40f11a9d-7d50-4cb7-9f9b-cc25f102ba79

PTAL.

Rd4dev:
Thanks, @theayushyadav11 — appreciate the prompt effort in exploring different approaches (especially the binding adapter solution), and sharing the attached assets. They were super helpful in analyzing the options thoroughly.

Regarding the use of `viewModel.isAdmin ? 5 : 3`, as mentioned earlier, this was the exact behavior I was cautious about. That said, I’m now considering, as Ben suggested, whether we can still leverage data binding in line with existing practices. However, this would likely require a couple of adjustments to the `ViewModel`.

**Option 1: Modify `isAdmin` to `true` by default (not recommended)**  :x: 
If we default `isAdmin` to `true`, This should work as it would initially allow a higher pin length (5), and later recompute correctly to either 3 or 5. But this feels highly unsafe — especially if in future `isAdmin` is accessed prematurely during activity creation, it could unintentionally grant elevated privileges to non-admin users, which is a significant risk. (Just flagging this — as it might seem like a potential solution, but it's highly risky.)

**Option 2: Introduce a new `maxPinLength` field**  
Instead of relying directly on `isAdmin`, we could introduce a separate `maxPinLength` field initialized to a safe max default (say, 5). Once the `profile` is available, we update this field based on the pin length from the profile:  
```kotlin
val maxPinLength = ObservableField<Int>(5) 

private fun processGetProfileResult() {
  // ...
  maxPinLength.set(profile.pin.length)
} 
```
This seems like a safer workaround — we avoid exposing `isAdmin` prematurely and still benefit from a reasonable default until the actual value is loaded.

I understand there’s been quite a bit of back-and-forth exploring different approaches but I believe this should hold up well without edge case issues, but feel free to flag any concerns if you come across them.

theayushyadav11:
> Thanks, @theayushyadav11 — appreciate the prompt effort in exploring different approaches (especially the binding adapter solution), and sharing the attached assets. They were super helpful in analyzing the options thoroughly.
> 
> Regarding the use of `viewModel.isAdmin ? 5 : 3`, as mentioned earlier, this was the exact behavior I was cautious about. That said, I’m now considering, as Ben suggested, whether we can still leverage data binding in line with existing practices. However, this would likely require a couple of adjustments to the `ViewModel`.
> 
> **Option 1: Modify `isAdmin` to `true` by default (not recommended)** ❌ If we default `isAdmin` to `true`, This should work as it would initially allow a higher pin length (5), and later recompute correctly to either 3 or 5. But this feels highly unsafe — especially if in future `isAdmin` is accessed prematurely during activity creation, it could unintentionally grant elevated privileges to non-admin users, which is a significant risk. (Just flagging this — as it might seem like a potential solution, but it's highly risky.)
> 
> **Option 2: Introduce a new `maxPinLength` field** Instead of relying directly on `isAdmin`, we could introduce a separate `maxPinLength` field initialized to a safe max default (say, 5). Once the `profile` is available, we update this field based on the pin length from the profile:
> 
> ```kotlin
> val maxPinLength = ObservableField<Int>(5) 
> 
> private fun processGetProfileResult() {
>   // ...
>   maxPinLength.set(profile.pin.length)
> } 
> ```
> 
> This seems like a safer workaround — we avoid exposing `isAdmin` prematurely and still benefit from a reasonable default until the actual value is loaded.
> 
> I understand there’s been quite a bit of back-and-forth exploring different approaches but I believe this should hold up well without edge case issues, but feel free to flag any concerns if you come across them.


Thanks @Rd4dev for providing a detailed overview of the solution.
I have implemented the **Option 2** and it works fine. I believe this approach may solve all the concerns without leaving any edgecases where it may fail.
PTAL.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,42 @@ class PinPasswordActivityTest {
     }
   }
 
+  @Test
+  fun testPinPassword_withAdmin_inputPin_configChange_pinIsPersisted() {
```
Rd4dev: Nit: Slight naming suggestion to align with other test cases.
```suggestion
  fun testPinPassword_withAdmin_inputPin_configChange_inputPinIsPersisted() {
```

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,42 @@ class PinPasswordActivityTest {
...
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("1234"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .check(matches(withText("1234")))
+    }
+  }
+
+  @Test
```
Rd4dev, line 1235: Could you also include test cases for the following scenarios 
- For both user and admin
    - The PIN input should restrict entry to the correct **max length**.
    - After a config change, the entered PIN should still **persist and allow transition** to the next screen. 
      - e.g., for admin: 
        - enter "123" 
        - rotate 
        - then enter "45" 
        - it should transition to HomeActivity / ClassroomActivity
      (This case should be covered for both classroom enabled and disabled states.)

These tests will help ensure the fixes introduced in this PR are properly verified and protected against future regressions.

Rd4dev (changes requested):
Thanks @theayushyadav11. This looks good! Additionally, I've requested a few more test cases that should help make this ready for a review pass and could you also update the PR description so it focuses more on the **intent behind the PR**, rather than a breakdown of code changes?

As a general guide, the description should summarize everything the _code diff can’t explain_ — mainly the **“why”** rather than the **“what.”** That doesn’t mean you can’t reference code; feel free to point to important files, functions, or lines if they help convey the purpose. Just try to avoid listing every change line by line — that’s what the diff is for.

Some helpful things to include:

- **A clear summary of the changes and their impact**  
- **The goal or problem being solved** — this helps reviewers understand the need for the PR  
- **Any relevant design decisions or reasoning** that may not be obvious from the code  
- **Links to related issues or discussions**  
- **Test and verification details** — what was tested and how (manual or automated)  
- **Future scope or intentionally omitted work** — anything planned for a later PR  
- **For UI changes**, include before/after screenshots or GIFs / Videos if possible (which you already did, and was helpful)

Try to structure with bullets or headers for readability (which you already do well!). Also, it's good practice to keep the description updated as the PR evolves — this helps ensure the final version accurately reflects what was merged.

Think of the PR description as something that sticks around even after comments and commits fade into the background — it becomes the go-to reference for future contributors trying to understand the context behind the changes. So its good to aim to write it with your future self or teammates in mind — could someone unfamiliar with this PR understand what it did and why, just by reading the description?

You can refer to #4757, #5208, #2980, or #2984 for some examples

model/src/main/proto/arguments.proto
Rd4dev: Outdated, as the solution approach evolved.

model/src/main/proto/arguments.proto
Rd4dev: Outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
Rd4dev: Outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
Rd4dev: This could have been moved to Activity -- but resolving since outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivity.kt
Rd4dev: Outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
Rd4dev: Outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
Rd4dev: Outdated, as the solution approach evolved.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
Rd4dev: Outdated, as the solution approach evolved.

theayushyadav11:
PTAL @Rd4dev , additionally to attain the testing of the following test case:
1.enter 123
2. change configuration
3. enter 45
I have to change the behavior of `appendText` as it used `setText` which replaces the previous input, so I have used `.append()` instead. [Ref](https://github.com/oppia/oppia-android/pull/5617/files#diff-dddddad2a891d378f09c055edbf29fcefea16a2e027e462472e1388338c584c2)

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Added.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,155 @@ class PinPasswordActivityTest {
...
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("12"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("43"), closeSoftKeyboard())
+      testCoroutineDispatchers.runCurrent()
+      intended(hasComponent(HomeActivity::class.java.name))
+    }
+  }
+  @Test
```
Rd4dev, line 1350: Here and elsewhere add new line after individual test cases.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,155 @@ class PinPasswordActivityTest {
...
+  fun testPinPassword_enableClassroom_withUser_inputConfigChangeInput_opensClassroomListActivity() {
```
Rd4dev: Maybe something like
```suggestion
  fun testPinPassword_enableClassroom_withUser_retainPinOnRotation_opensClassroomListActivity() {
```

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,155 @@ class PinPasswordActivityTest {
...
+        .perform(editTextInputAction.appendText("123"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("45"), closeSoftKeyboard())
+      testCoroutineDispatchers.runCurrent()
+      intended(hasComponent(ClassroomListActivity::class.java.name))
+    }
+  }
+
+  fun testPinPassword_withUser_inputPartialPin_configChange_inputFullPin_opensHomeActivity() {
```
Rd4dev: Add test annotation.

Rd4dev: Let similar test cases share the same pattern to make it easier to compare consecutive tests of the same method that have slightly different conditions with divergent outcomes. [Oppia Android Testing Wiki](https://github.com/oppia/oppia-android/wiki/Oppia-Android-Testing#naming-convention)

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,155 @@ class PinPasswordActivityTest {
...
+        adminPin = adminPin,
+        profileId = userId
+      )
+    ).use {
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("12"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("43"), closeSoftKeyboard())
```
Rd4dev:
```suggestion
        .perform(editTextInputAction.appendText("3"), closeSoftKeyboard())
```

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
```diff
@@ -50,7 +50,7 @@ class EditTextInputAction @Inject constructor(
       override fun perform(uiController: UiController?, view: View?) {
         // Appending text only works on Robolectric, whereas Espresso needs to use typeText().
         if (Build.FINGERPRINT.contains("robolectric", ignoreCase = true)) {
-          (view as? EditText)?.setText(text)
+          (view as? EditText)?.append(text)
```
Rd4dev: Nice catch — it seems like the original intention was to append text, but got modified along the way. It’s positive that the transitive test runs passed, showing this didn’t affect any other cases. However, looking at the input action, it appears that both `appendText` and `replaceText` perform the same action for Robolectric. This could be a good opportunity to properly separate their functionality to support both as the fix now only supports appending while losing support for replacements. Could you update it accordingly? 

Discussed it earlier with Ben and it was advised that utility fixes should generally be done in a separate PR, but if it’s a small change, it can be included in the same PR itself.

Rd4dev (changes requested):
@theayushyadav11, the included test cases look good. Other than few nits on the naming -- requested to also update the modified `EditTextInputAction` to include support for replaceText too. PTAL.

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
theayushyadav11: Thanks for the feedback! I've updated EditTextInputAction to correctly differentiate between appendText() and replaceText() in the Robolectric case.
Specifically:
   ` appendText()` now uses EditText.append(text) for Robolectric.
    `replaceText()` now uses EditText.setText(text) for Robolectric.

For non-Robolectric cases, both actions continue to delegate to the base Espresso ViewAction (typeText and replaceText respectively).

This ensures that both appending and replacing text behave correctly in tests based on the intended action.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: done

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Added

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Added.

theayushyadav11:
PTAL @Rd4dev .

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
     }
   }
 
+  @Test
+  fun testPinPassword_withAdmin_inputPin_configChange_inputPinIsPersisted() {
```
adhiamboperes:
```suggestion
  fun testPinPassword_adminUser_inputFiveDigitPin_configChange_inputIsPersisted() {
```

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+        adminPin = adminPin,
+        profileId = adminId
+      )
+    ).use {
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("1234"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .check(matches(withText("1234")))
```
adhiamboperes: Admin PIN is always 5 digits. It is ok, but not necessary, to test for 4 digits because the next test already checks for 3 digits. The tests we are interested in are:
1. Empty input(I think it is already covered in existing tests)
2. inputs under 3 digits
3. Inputs over 5 digits
4. 3 digits 
5. 5 digits

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+    }
+  }
+
+  @Test
+  fun testPinPassword_withUser_inputPin_configChange_inputPinIsPersisted() {
```
adhiamboperes:
```suggestion
  fun testPinPassword_nonAdminUser_inputThreeDigitPin_configChange_inputIsPersisted() {
```

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("123"), closeSoftKeyboard())
+      onView(isRoot()).perform(orientationLandscape())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .check(matches(withText("123")))
+    }
+  }
+
+  @Test
+  fun testPinPassword_withAdmin_inputPin_exceedMaxLength_textIsTrimmed() {
```
adhiamboperes:
```suggestion
  fun testPinPassword_adminUser_inputPinExceedsFive_textIsTrimmedToFive() {
```

This sets clearer expectaions of the test setup and outcome

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("1234567"), closeSoftKeyboard())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .check(matches(withText("12345")))
+    }
+  }
+
+  @Test
+  fun testPinPassword_withUser_inputPin_exceedMaxLength_textIsTrimmed() {
```
adhiamboperes:
```suggestion
  fun testPinPassword_nonAdminUser_inputPinExceedsThree_textIsTrimmedToThree() {
```

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+      testCoroutineDispatchers.runCurrent()
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .perform(editTextInputAction.appendText("1234567"), closeSoftKeyboard())
+      onView(withId(R.id.pin_password_input_pin_edit_text))
+        .check(matches(withText("123")))
+    }
+  }
+
+  @Test
+  fun testPinPassword_withAdmin__retainPinOnRotation_opensHomeActivity() {
```
adhiamboperes:
```suggestion
  fun testActivity_multipleClassroomsDisabled_adminUser_inputPin_changeConfig_opensHomeActivity() {
```
Covers all the test conditions. Use activity here to stay within line length limit.

Ditto the tests below.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
```diff
@@ -1214,6 +1214,160 @@ class PinPasswordActivityTest {
...
+        .perform(editTextInputAction.appendText("45"), closeSoftKeyboard())
+      testCoroutineDispatchers.runCurrent()
+      intended(hasComponent(HomeActivity::class.java.name))
+    }
+  }
+
+  @Test
+  fun testPinPassword_enablesClassroom_withAdmin_retainPinOnRotation_openClassroomListActivity() {
+    TestPlatformParameterModule.forceEnableMultipleClassrooms(true)
+    ActivityScenario.launch<PinPasswordActivity>(
```
adhiamboperes: If you import `ActivityScenario` directly, you will avoid redudantly mentioning it in every test.
```suggestion
   launch<PinPasswordActivity>(
```

adhiamboperes (changes requested):
Thanks @theayushyadav11! This is an interesting bug. 

I have left some comments to clean up the tests further. 

Thanks @Rd4dev for the thorough guidance and review.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Done.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

app/src/sharedTest/java/org/oppia/android/app/profile/PinPasswordActivityTest.kt
theayushyadav11: Updated.

theayushyadav11:
PTAL @adhiamboperes .

adhiamboperes (approved):
Yhanks @theayushyadav11! LGTM! I have no further comments. 

Let's look into why the code coverage run is failing.

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
Rd4dev: Thanks for fixing this properly. Could you also file an issue describing the bug, with the following two points:

1. The issue itself (Once merged, mark as done and refer to the merged PR) 
2. Adding a corresponding tests

I think we can have a test class similar to the one available for `TextInputAction` to ensure changes like this are properly tested. For this PR, it should be sufficient to just add a reference as “Fixes part of #Issue-number” in the description for (1). As for (2), it can be treated separately as it's already out of scope for this PR.

Rd4dev (approved):
Thanks, @theayushyadav11, for the work — this LGTM! -- Just requested to file the issue and reference it here — once that’s done, I’ll resolve it too.

Regarding the [failure with coverage comment](https://github.com/oppia/oppia-android/actions/runs/14721519638/job/41317323915#step:7:52), it seems like it was a flake.

Thanks, @adhiamboperes, for retriggering the coverage workflow -- I also retriggered the comment report workflows, and they report the latest version now.

theayushyadav11:
Thanks @Rd4dev for the review, I have filed the [issue](https://github.com/oppia/oppia-android/issues/5818) and updated the description as requested. PTAL

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
Rd4dev: for ref: #5818

Rd4dev:
Thanks @theayushyadav11, for helping with documenting the issue. I've resolved the related open comment. @adhiamboperes can you help with any further actions or merging.

adhiamboperes:
@theayushyadav11, the PR looks ready to merge, but it is not clear to me which parts of  https://github.com/oppia/oppia-android/issues/2133 and https://github.com/oppia/oppia-android/issues/5818 have not been addressed. I would just like clarity on whether this PR resolves these issues fully, or parts still need work in fuure PRs.

theayushyadav11:
> @theayushyadav11, the PR looks ready to merge, but it is not clear to me which parts of #2133 and #5818 have not been addressed. I would just like clarity on whether this PR resolves these issues fully, or parts still need work in fuure PRs.

Hi @adhiamboperes , this PR solves the part of the #2133, as it fixes the second part mentioned [here](https://github.com/oppia/oppia-android/issues/2133#issuecomment-2535222536).
For the #5818 this PR solves its first part for clarifying the logic for `appendText()` and `replaceText()`. 
PTAL.

Rd4dev:
Hi @theayushyadav11, could you update the issue to reflect that the first part of issue #5818 was addressed by this PR? You can list both parts as check boxes—mark the first one as completed with a reference to this PR, and leave the second one unchecked to help track progress more effectively.

theayushyadav11:
Sure will do it.

PR #5629, opened 2024-12-28
Fix part of #2747: Disable Gradle in CI

BenHenning:
## Explanation
Fix part of #2747.

In order to unblock #5604, this PR disables Gradle in CI completely. This is an extreme mitigation to the issues discovered as part of that PR, but please see https://github.com/oppia/oppia-android/pull/5628#issuecomment-2564494780 and that PR's main description for a detailed account of the attempt to migrate to the latest versions of Gradle. It seems that the best path forward both for this upcoming release and long-term is to outright remove Gradle, even though Bazel is not completely developer ready. The alternative (keeping a broken version of Gradle for several months as we migrate dependencies) seems far too disruptive, and not ideal in the long-term since we want to move away from having two build systems, anyway.

Please note that I will remove the required Gradle CI checks once the PR is approved (so that it can be merged).

This change and the eventual removal of Gradle has been announced in the CLaM chat. We'll need to reach out to specific newer team members in the new year once we actually perform the removal and update corresponding documentation.

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [ ] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
This is an infrastructure only change.

PR #3910, opened 2021-10-07
Fix #3803, #3804, #3853, #3669, #3383: Add basic KitKat support in Bazel

BenHenning:
## Explanation

Fix #3803
Fix #3804
Fix #3853
Fix #3669
Fix #3383

Introducing basic KitKat support (that is, support for the app to open and be minimally stable during usage) required the following work:
1. Introducing a permission needed by AndroidX WorkManager
2. Introducing proper legacy multidex support (with Bazel-specific issues--see below & #3803)
3. Adding support for vector graphics (which is API 21+ normally -- #3804)
4. Removing/disabling OkHttp (which doesn't have KitKat support -- #3853)
5. Fixing some Java 8 incompatibilities
6. Properly setting up the build for KitKat due to build-specializations

Each of these are covered in more detail in later sections. Note that this PR continues the work from #3740 and #3723 which provided the basis for addressing the OkHttp dependency issue, and parts of the dexing issues.

### New permission
Per https://developer.android.com/reference/androidx/work/package-summary AndroidX WorkManager requires the WAKE_LOCK permission before SDK 23 in order to properly handle background service/keep-device-awake scenarios for triggering workflows depending on certain conditions. This isn't KitKat specific, but it was a prerequisite for the app to run on KitKat devices.

### Legacy multidex support in Bazel
Bazel supports three types of multidex modes: legacy, native, and manual. Native relies on the native ART environment to support reading from multiple dex fiiles at runtime, and works without issue.

Legacy assumes that the application extends AndroidX's multidex application to handle scenarios where multidex is needed (which it is for Oppia since we have 3 dex files compiled for the size of the developer flavor of the app). However, a big part of this is computing what files should go into the main dex file list for the app. Bazel uses some mechanism to compute this list (which I think uses D8 via a compatibility bridge in Android buiol-tools), but unlike Gradle it seems to have issues computing a list that can actually fit within the main dex list. This is a problem because the main dex list needs to contain everything necessary to load the main application class so that multidex can kick in & initialization can start (more on this later). It's not clear to me why Gradle works and Bazel fails (as in, will fail with a compiler error when trying to use "legacy"), but I suspect it has to do with Gradle using the new R8 dexer (which isn't yet available via Bazel per #3748). Finally, this isn't an issue for the alpha version of the app since Proguard brings the dex file list much lower (down to 2), so Bazel seems fine computing a proper main dex list.

[...]

PR #2884, opened 2021-03-11
[RunAllTests] Remove caching for Bazel CI workflows

BenHenning:
Bazel caching is currently not cost effective: it only caches 25-30% of actions, and we're not seeing worthwhile performance benefits from it. This PR makes caching configurable for both unit tests and binary builds, and disables it for CI workflows. #1861 is tracking improving this over time, including upstream patching Bazel itself to fix some of the caching issues. We may revisit 
reenabling caching at that time.

Runs verifying that caching is now disabled:
- [Builds](https://github.com/oppia/oppia-android/pull/2884/checks?check_run_id=2084344439)
- [Unit tests](https://github.com/oppia/oppia-android/pull/2884/checks?check_run_id=2084348177)

*Builds*
image

*Unit tests*
image

PR #5630, opened 2024-12-29
Fix #5012: Remove KitKat support

### LINKED ISSUE #5012: Deprecate support for Android KitKat

We have decided not to pursue support for Android KitKat. This was an aspirational goal that we had wanted to attempt because we knew of tablets owned by some partners that still run on KitKat, but as the years go by we expect that the number of such devices will dwindle.

Additionally, there are significant barriers to supporting KitKat, due to this requiring third-party dependencies that cannot be upgraded and have security issues. So we have decided to not attempt to support it, and remove any extra code that is special-cased for KitKat in order to simplify the codebase and development experience.

Remove any code in the Android codebase that is there only to provide support for KitKat.

**Describe alternatives you've considered**
We could leave the KitKat code in, but this does not seem useful since we aren't planning to support it in the long run. It seems better to clean up the code in order to reduce the maintenance burden and make it easier for developers to follow it.

BenHenning:
## Explanation
Fixes #5012

Removes references and code corresponding to KitKat (SDK version 19) since the app has been set to a minimum version of Lollipop (SDK 21) since #3910 (roughly--technically KitKat was still supported but we no longer shipped a KitKat version). Because of this, the changes in this PR are largely a no-op from a build perspective.

More broadly, this change is motivated by a desire to decrease CI time (which was already reduce considerably in the recent merge of #5629) by removing extraneous app builds being done in CI:
- ``//:oppia`` since we should really only use ``//:oppia_dev`` moving forward.
- Dev and alpha KitKat builds (the latter of which takes a long time since only the alpha job was building 2 proguarded builds of the app and thus taking much more time than the beta and GA build jobs).

Separately, ``build_tests.yml`` and ``unit_tests.yml`` were updated to no longer support caching since we disabled this behavior a while back (#2884) and are unlikely to reintroduce it due to high storage costs.

Some other noteworthy changes:
- The optional providing of ``Retrofit`` and Retrofit services has been reverted since it's no longer necessary (and corresponding tests have been removed since there's no longer any condition to verify).
- Code that was gated to run below Lollipop has been removed since it can't execute except on KitKat devices.
- Support for a main dex target list has been removed as we're unlikely to need it with native multidex support (which wasn't an option for KitKat builds), though we may choose to reintroduce it to speed up cold app starts since it _can potentially_ help improve time-to-splash app startup performance.
- Updated manifest min SDK values to avoid potential developer confusion on the min SDK version supported (though, strictly speaking, this doesn't technically matter with the way the app builds).
- This updates the default min version supported for OS deprecation. No additional work is needed to actually activate the deprecation notice on the KitKat version of the app since we no longer deploy it, and OS deprecation wasn't added to the last version that was deployed.
- Three file content pattern failure messages were updated to no longer reference KitKat (though the value of these checks mostly still seems realized, so I think it's fine to keep them in).
- Some additional licenses were added due to tooling around the dev AAB (even though these aren't dependencies that ship with the app). I think it's fine including extra licenses in this case.
- The protobuf license was corrected to BSD 3-Clause as declared for the dependency and in the GitHub repository for the dependencies.
- This bumps version codes due to removing two flavors (for KitKat dev and KitKat alpha).
- ``NetworkModuleTest`` has had its tests redesigned and better filled in in order to pass the code coverage minimum requirement (since old tests were removed). Note that it's not quite perfect in what it's testing, but it is at least verifying the complete configuration of the ``Retrofit`` instance, and the singleton properties of the two services currently supported (verifying that the services are set up would require a lot more testing that seems outside the direct scope of this test, so it seems okay to ignore here).
- As part of the previous item's work, ``NetworkLoggingInterceptor`` had an issue fixed where it could cause an OkHttp exception to be thrown when trying to process logs (due to reading the response body twice). Reverting this change will now cause breakage failures in ``NetworkModuleTest``.
- ``NetworkModule``'s initialization order was changed for slightly better network request logging (see the new comment in that file for more context).

Note that there are some workflow job removals and renames in this PR. Any required CI checks will be correspondingly updated once this PR is approved and ready to merge (to avoid blocking any other in-flight PRs).

## Essential Checklist

## For UI-specific PRs only
This shouldn't have an side effects on non-KitKat UIs, and KitKat support itself is being removed in this PR (so there's nothing to demonstrate).


BenHenning (commented):
Self-reviewed.

BenHenning (commented):
Self-reviewed latest change.

BenHenning:
PTAL @adhiamboperes. This should be ready for review now--there was only 1 failing test and it's been fixed. I'll double check the CI run when it completes, but I expect everything to pass now.

adhiamboperes (approved):
Thanks @BenHenning! This LGTM, although more coverage is required for NetworkModule.

BenHenning (commented):
Self-reviewed latest changes.

BenHenning:
> Thanks @BenHenning! This LGTM, although more coverage is required for NetworkModule.

Thanks @adhiamboperes! I've addressed this--PTAL since there are quite a few new tests being added (and some code generally being changed). I've also added 3 new bullet items to the PR description (which are copied below) if that context is helpful.

> * `NetworkModuleTest` has had its tests redesigned and better filled in in order to pass the code coverage minimum requirement (since old tests were removed). Note that it's not quite perfect in what it's testing, but it is at least verifying the complete configuration of the `Retrofit` instance, and the singleton properties of the two services currently supported (verifying that the services are set up would require a lot more testing that seems outside the direct scope of this test, so it seems okay to ignore here).
> * As part of the previous item's work, `NetworkLoggingInterceptor` had an issue fixed where it could cause an OkHttp exception to be thrown when trying to process logs (due to reading the response body twice). Reverting this change will now cause breakage failures in `NetworkModuleTest`.
> * `NetworkModule`'s initialization order was changed for slightly better network request logging (see the new comment in that file for more context).

@kkmurerwa I'm curious whether the last 2 bullet items above may be relevant to Retrofit calls not being logged in #5243? If there's an exception occurring during the actual call pipeline, it will fail before we even have an opportunity to log the issue (which is what I was observing in tests). Can you confirm whether the changes in this PR actually fix that issue?

@seanlip PTAL for codeowners change.

seanlip (approved):
Thanks @BenHenning, LGTM for codeowners.

kkmurerwa:
@BenHenning I will run the app on this branch and see if the retrofit issue is fixed.

adhiamboperes (approved):
Thanks @BenHenning, this LGTM.

kkmurerwa:
@BenHenning The issue is not fixed yet.

image

BenHenning:
> @BenHenning The issue is not fixed yet.
> 
> image

@kkmurerwa I don't think that stack trace actually indicates anything. See https://stackoverflow.com/a/59636900. What needs to be tested here is whether the job workers are actually correctly trying to connect to Oppia web via Retrofit after this PR's changes (since they would silently fail before due to the incorrect Retrofit configuration).

BenHenning:
Thanks @adhiamboperes and @seanlip for the reviews. Enabling auto-merge on this and removing the KitKat required checks.

Edit: Actually, there may not need to be any workflow renames. Waiting to see.

adhiamboperes:
@BenHenning, there seems to be a merge confict in the PR, PTAL.

BenHenning:
Ah. Will resolve that momentarily, thanks @adhiamboperes.

I also now see what needs to be done for CI. We removed the workflow for verifying the APK builds since those targets were removed. Yay for slightly more efficient CI!

BenHenning (commented):
Double checked the changes in `build_flavors.bzl` since there were conflicts to resolve from #5664. The resolution seemed to be clean and was simple, so I don't think another review check is needed here.

PR #5647, opened 2025-01-09
Fix #4170: Adding LatexImageSpan for vertical alignment of locally rendered and cached LaTeX

### LINKED ISSUE #4170: Fix vertical alignment for locally rendered and cached LaTeX
#4068 introduced support for locally rendering and caching LaTeX expressions. Unfortunately, the local cached LaTeX uses an image span for rendering a bitmap representation of the LaTeX and doesn't seem to be center-alignable. Android only supports bottom and baseline alignment in ImageSpans, and custom translation doesn't seem to fix it. The closest I was able to get was using a negative top bound with a translation, but it's not quite correct:

cached_latex_rendering

The expected result is (when rendered with caching disabled):

direct_render_latex

Note that this only applies to inline LaTeX, not block LaTeX since that computes bounds such that the image is centered.

One attempt to fix it:

Because we're drawing the LaTeX as a bitmap in an ImageSpan, the text aligns the image either with its bottom line or its baseline. This introduces the vertical center alignment issue as shown in the first image above. The previous image shows one attempt at fixing this by eliminating unnecessary upper space by using a combination of vertical translation and negative bounds. Due to the quirk that no whitespace can be below the text, we probably need to modify the text's font metrics (specifically the bottom/top/ascent/descent properties).

manas-yu:
## Explanation

Fix #4170 

The PR implements `LatexImageSpan`, which calculates the appropriate ascent, descent, and alignment offsets based on the font metrics and drawable height. This ensures that LaTeX equations appear properly aligned.

## Before
before

## After 
after

## Essential Checklist


manas-yu:
Hey @adhiamboperes PTAL, Thanks!

utility/src/main/java/org/oppia/android/util/parser/html/MathTagHandler.kt
```diff
@@ -144,3 +147,73 @@ class MathTagHandler(
...
+    text: CharSequence,
+    start: Int,
+    end: Int,
+    x: Float,
+    top: Int,
+    y: Int,
+    bottom: Int,
+    paint: Paint
+  ) {
+    val d = drawable
```
adhiamboperes: Can we change to a more descriptive variable name?

adhiamboperes (changes requested):
Thanks @manas-yu!

This is a nice solution to this issue! 

Could you please check the text scaling behavior of the rendered latex?

Also, I think that the implementation behaviour assumes that the Drawable is already scaled to the correct size. But I think this is a pretty safe assumption since [UrlImageParser](https://github.com/oppia/oppia-android/blob/68f20cbfedc411a7fd26e359600a4572293acaea/utility/src/main/java/org/oppia/android/util/parser/image/UrlImageParser.kt#L142-L143) advises against resizing the latex.

manas-yu:
Hi @adhiamboperes, 

I've updated the variable names to make them more descriptive. Additionally, the implementation supports text scaling effectively. PTAL and let me know your thoughts!

## Small Text Size  
Small Text Size

## Extra Large Text Size  
Extra Large Text Size 1  
Extra Large Text Size 2

Let me know if there's anything else you'd like me to address!

adhiamboperes:
@manas-yu, I have noticed that at the bottom of the screen, the denominator `9` is somewhat cutoff. Why is this happening?

manas-yu:
Hi @adhiamboperes, PTAL.

After some investigation, it seems that the cutting-off issue was specific to `LaTeX` rendered at the end of the `TextView`. This was an interesting behavior since the truncation did not correlate with the height of the `LaTeX` content. During testing, I observed that the same `LaTeX` rendered correctly in other positions within the `TextView`, except at the end.

To address it, I’ve made changes to the current implementation, and it now resolves the problem effectively.

ref: https://stackoverflow.com/questions/25628258/align-text-around-imagespan-center-vertical, https://issuetracker.google.com/issues/36934789 

Below are the updated SSs of the current implementation:
## Small
small
## Extra Large
extra_large1
extra_large2

adhiamboperes (approved):
Thanks @manas-yu! LGTM!

Issue #4991, opened 2023-05-19
Add Bazel support for Espresso tests


BenHenning:
**Is your feature request related to a problem? Please describe.**
The current codebase setup doesn't support running Bazel UI tests in Espresso, only locally via Robolectric.

**Describe the solution you'd like**
We likely need a custom instrumentation setup to make this work as there are compatibility issues with android_test (as found through the end-to-end testing project).

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #4748, opened 2022-11-19
Update Robolectric tests to API 34 [Blocked: #4119, #4746, #5137]


BenHenning:
#4746 is tracking updating the codebase to SDK 31, but unfortunately Robolectric tests can't be included as part of this because:
- Robolectric starts supporting SDK at version 4.7 (see https://github.com/robolectric/robolectric/releases/tag/robolectric-4.7).
- Robolectric version 4.7 requires a minimum of JDK 15 (from trying to build the tests--see Gradle CI runs in #4747).
- While JDK 15 is supported at our current Gradle version (6.8), it's not at our current AGP version (3.6).
- The first version of AGP supporting JDK 15 (via 11+ support) is 7.0 per https://issuetracker.google.com/issues/139013660#comment7.
- AGP 7.0 requires upgrading both Gradle to 7.0 and Kotlin to 1.5 (https://developer.android.com/studio/releases/past-releases#expandable-9), hence the block on #4119.

Update as of 1 Nov 2023: Per #5137 this now needs to involve updating Robolectric tests to use SDK 33.

Upgrading to Kotlin 1.5 will affect production code, and per the actual runtime failures seen from trying to upgrade Robolectric tests in #4747 it's possible other dependencies will need to be updated, too. I think we'll instead:
1. Update to target SDK 31 without moving over tests
2. Update the codebase properly to Kotlin 1.6 (per #4120)
3. Attempt again to move over the tests to SDK 31, updating dependencies as needed

Per (1) I think we'll try to prioritize getting as close to a final solution as possible (e.g. updating everything *except* Robolectric to SDK 31, including the Java runtime versions for CI).

BenHenning:
Updating to SDK 34 per #5535 and #5604.

Issue #1617, opened 2020-08-11
Remove genrules once Gradle has been removed


miaboloix:
The genrules in app module's BUILD file are necessary only for building Oppia Android in both Gradle and Bazel. Similarly, the custom macro in the model module (`format_import_proto_library.bzl`) will need to be removed once Gradle is out of the picture. For both genrules, removing them will require changing some import statements (Ex. `import org.oppia.app.R` -> `import org.oppia.app.vm.R`).

PR #5652, opened 2025-01-16
Fix part of #2747, #1532: Remove support for Gradle

### LINKED ISSUE #1532: Change Bazel BUILD file names post-Gradle
Bazel convention suggests that all Bazel BUILD files should be named `BUILD`. However, as we gradually move to Bazel while still building with Gradle, using this convention cause naming conflicts with existing Gradle files. Thus, all Bazel BUILD files are named `BUILD.bazel`. Although this is still acceptable to Bazel, it is not proper convention and should not be used post-Gradle in Oppia Android. 

Once Oppia Android has been moved off of Gradle and is building completely with Bazel, we should refactor all Bazel build files to be named `BUILD` instead of `BUILD.bazel`

### LINKED ISSUE #1640: Move binary manifest to top-level package post-Gradle [Blocked: #6010]

### LINKED ISSUE #2747: Move off of Gradle [Blocked: #2746]

BenHenning:
## Explanation
Fixes part of #2747
Fixes #1532

This PR is doing a lot to address the vast majority of #2747--removing support for Gradle. It doesn't address everything, however--some items will need to be addressed in future PRs (largely new cleanup opportunities that are unlocked with the removal of Gradle). However, it does remove the entirety of Gradle definition infrastructure.

This PR specifically:
- Removes all Gradle build definitions (``.gradle`` files, ``settings.gradle``, the ``gradlew`` wrappers, and the ``gradle/wrapper`` directory). As a side benefit, this removes some of the last binary files from the codebase. :)
- Removes a bunch of Gradle-specific files not actually used by the Bazel builds, including proguard-rules.pro files (Bazel has separate definitions under ``config/proguard/``), some manifest files, and some Robolectric-specific definitions.
- Deletes the ``main.yml`` workflow (which was disabled in #5629). This is entirely a Gradle-specific workflow, so there's nothing that needs to be moved over as all of what's being tested is covered more efficiently and thoroughly in other specific workflow files.
- Updates to ``.gitignore`` per the Gradle file removals (fewer files/directories need to be ignored).
- Removed all TODOs on #1532 and marking it as fixed as part of this PR rather than performing the file rename. See https://github.com/oppia/oppia-android/issues/1532#issuecomment-2610933772 for an explanation on why.
- Some comments have been updated rather than addressed where it made sense that addressing the problem was not exactly specific to Gradle support (there's a desire to remove all references to 'Gradle' as a clear way to ensure everything Gradle-specific has been successfully removed from the codebase--this has largely been done).
- The infrastructure and definitions for segmenting tests between Bazel and Gradle has been completely removed (resulting in broad changes in the codebase) since Gradle-specific behaviors can't be tested anymore, and Bazel-specific behaviors are now the default.
- Some of the waiting logic in ``StateFragmentTest`` and ``ExplorationActivityTest`` has been kept but had its removal TODO removed. This is because the code seems to have changed (or the original problem misunderstood). These waits seem actually necessary due to relying on real image loading, so the fix is more complex than simply "migrate to Bazel & move off of Gradle." There's real test infrastructure work needed to fix these tests properly, and it doesn't seem sufficiently important to prioritize (though note it did seem possible to remove the waits outright in ``ExplorationActivityTest`` without issue). The longer-term infrastructure work is tracked in #1523.
- ``LoadLessonProtosFromAssets`` has been made default true (since it can't ever be false with Gradle removed). See the note below on removing JSON assets for the work to fully clean this up.
- The removal of ``OppiaTestRunner`` as it was only used for Gradle builds to help it set up state correctly for orchestrated tests (i.e. Espresso). We may need to reintroduce something like this in the future, but for now there's no value in keeping an unused class.
- Some simplifications in ``RepositoryFile`` and test removals in ``RepositoryFileTest`` now that Gradle build directories no longer need to be filtered.
- ``TestEnvironmentConfig`` was removed since there's no longer a need to test which environment tests are currently running in.
- Some Espresso documentation was removed since it's clearly no longer needed, but more documentation work is still needing to be completed (this was noticed due to the removal of ``OppiaTestRunner``).
- ``ApiMockLoader`` was simplified now that test config artifacts are in a single location during testing time.
- ``InitializeDefaultLocaleRule`` can now have nearly all of its reflection removed since it can be directly built to reference the classes that it's interacting with (which wasn't possible in the Gradle world, see the TODO in that class). Some new tests were added, as well, which brings its test coverage to 100%.
- Updates CODEOWNERS to reflect all of the many file changes from the above.
- Broad removal of the word 'module' in **many** locations and documentation. This term is generally replaced with the more architectural term 'layer' as to not overload 'module' from a Gradle, Bazel, and Dagger perspective (plus others). This required a lot of changes specifically in the Android Lint tool implementation and tests to try and keep the term 'module' specific to cases where it still makes sense (other than Dagger which is the bulk of the existing reference).
- `TransformAndroidManifest` was updated to support adding top-level comments in manifest files (which are TODOs for #6010 to move or remove manifest files--see below). Tests were also added to test the new functionality.
- A bunch of files were changed simply because my local editor removed trailing spaces and they were included in those various files that I happened to change (largely for 'module' reference renames and/or 'Gradle' reference removals/rewordings).

Note also that this work opens up more things that still need to be, or can now be, addressed, including:
- Removal of JSON lesson loading per #5663. We're no longer testing this behavior correctly as domain level tests would previously verify JSON loading logic in Gradle builds and proto logic in Bazel builds. This means there's a strong likelihood of breakages occurring if JSON loading isn't removed.
- Ensuring Espresso tests can actually run with Bazel (see #4991). We no longer can verify Espresso behaviors now that Gradle is removed, which also subsequently means all test gating on Robolectric is redundant and all tests gated on Espresso never run.
- It's possible #4748 is now made redundant with the removal of Gradle-specific ``robolectric.properties`` files, though I'm not entirely certain at this point which SDK version is used in Bazel tests for Robolectric (other than it being tied to Robolectric version and that we pull in all supported SDKs so that tests can selectively opt into specific SDKs for testing).
- Possibly some of the language configuration bits can be simplified now that we don't need to gracefully fail if the configuration is missing (since it never should be), but there's no plan to audit or track work for this.
- Reorganization of the codebase now that Gradle no longer imposes arbitrary modules (see https://github.com/oppia/oppia-android/issues/59#issuecomment-528217795) as well as fixing testing artifact isolation (multiple issues tracking parts of this, and ideas possibly not yet captured in issues).
- `genrule`s can now be removed to improve symbol analysis in Android Studio (and general build system complexity). See #1617.
- Some documentation can potentially be improved, but with the replacement of the terms 'Gradle' and 'module' (plus past documentation changes that happened as a result of moving the team over to Bazel exclusively) much of the existing documentation seems solid.
- It may now be possible to update to newer versions of Kotlin and Robolectric without worrying about AGP incompatibilities.
- Moving the app `AndroidManifest.xml` from the app directory to the top-level directory to address #1640. Unfortunately this can't be done as part of this PR (as previously attempted) due to #6010 and required changes to the Android Lint script. Note that the addressing of #6010 also allows other manifests to potentially be removed and, hopefully, a reduction of manifests to the bare minimum (the top-level plus any asset-holding libraries--at least for now).

## Essential Checklist

## For UI-specific PRs only
There are no known UI effects from removing Gradle support, but build infrastructure is complex and this is _not_ a no-op change. It's expected that the Bazel builds are relatively the same as before, however (see the APK & AAB differences analysis), so there shouldn't be an UI or UX changes as a result of this work.


app/src/sharedTest/java/org/oppia/android/app/walkthrough/topiclist/WalkthroughTopicListFragmentTest.kt
```diff
@@ -218,38 +214,6 @@ class WalkthroughTopicListFragmentTest {
     }
   }
 
-  @Test
-  fun testWalkthroughTopicListFragment_topicCard_lessonThumbnailIsCorrect() {
```
BenHenning, line 226: Try to re-add this. It seems wrong to remove it, actually.

BenHenning (commented):
Self-reviewed all changes.

app/src/sharedTest/java/org/oppia/android/app/walkthrough/topiclist/WalkthroughTopicListFragmentTest.kt
BenHenning: Tried enabling this in Bazel and am seeing the following failure:

```
1) testWalkthroughTopicListFragment_topicCard_lessonThumbnailIsCorrect(org.oppia.android.app.walkthrough.WalkthroughTopicListFragmentTest)
androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'with drawable from resource id: <2131230995>' doesn't match the selected view.
Expected: with drawable from resource id: <2131230995>[lesson_thumbnail_graphic_duck_and_chicken]
     Got: "LessonThumbnailImageView{id=2131297518, res-name=walkthrough_topic_thumbnail_image_view, visibility=VISIBLE, width=360, height=270, has-focus=false, has-focusable=false, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=androidx.constraintlayout.widget.ConstraintLayout$LayoutParams@6361c171, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0}"

        at java.base/java.lang.Thread.getStackTrace(Thread.java:1602)
        at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:16)
        at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:36)
        at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:103)
        at androidx.test.espresso.ViewInteraction.check(ViewInteraction.java:31)
        at org.oppia.android.app.walkthrough.WalkthroughTopicListFragmentTest.testWalkthroughTopicListFragment_topicCard_lessonThumbnailIsCorrect(WalkthroughTopicListFragmentTest_updated.kt:237)
        ... 19 trimmed
Caused by: junit.framework.AssertionFailedError: 'with drawable from resource id: <2131230995>' doesn't match the selected view.
Expected: with drawable from resource id: <2131230995>[lesson_thumbnail_graphic_duck_and_chicken]
     Got: "LessonThumbnailImageView{id=2131297518, res-name=walkthrough_topic_thumbnail_image_view, visibility=VISIBLE, width=360, height=270, has-focus=false, has-focusable=false, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=androidx.constraintlayout.widget.ConstraintLayout$LayoutParams@6361c171, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0}"

        at androidx.test.espresso.matcher.ViewMatchers.assertThat(ViewMatchers.java:17)
        at androidx.test.espresso.assertion.ViewAssertions$MatchesViewAssertion.check(ViewAssertions.java:15)
        at androidx.test.espresso.ViewInteraction$SingleExecutionViewAssertion.check(ViewInteraction.java:10)
        at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:11)
        at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:2)
        at android.os.Handler.$$robo$$android_os_Handler$handleCallback(Handler.java:938)
        at android.os.Handler.handleCallback(Handler.java)
        at android.os.Handler.$$robo$$android_os_Handler$dispatchMessage(Handler.java:99)
        at android.os.Handler.dispatchMessage(Handler.java)
        at org.robolectric.shadows.ShadowPausedLooper$IdlingRunnable.run(ShadowPausedLooper.java:342)
        at org.robolectric.shadows.ShadowPausedLooper.executeOnLooper(ShadowPausedLooper.java:271)
        at org.robolectric.shadows.ShadowPausedLooper.idle(ShadowPausedLooper.java:89)
        at org.robolectric.android.internal.LocalControlledLooper.drainMainThreadUntilIdle(LocalControlledLooper.java:15)
        at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:99)
        ... 22 more
```

This may tie into #5663, actually, so I'm going to temporarily ignore the test and add a TODO on that issue to fix it.

BenHenning: Actually, a better approach may be to just enable JSON loading for this test.

BenHenning:
TODO: Consider removing references to 'module' as well since that's Gradle specific in most uses and may add confusion.

app/src/sharedTest/java/org/oppia/android/app/walkthrough/topiclist/WalkthroughTopicListFragmentTest.kt
BenHenning: Re-added in 6f064dc6a7ac67cf08abb4aca0e6c5285a8caea1.

BenHenning:
> TODO: Consider removing references to 'module' as well since that's Gradle specific in most uses and may add confusion.

Done. Also fixed the code coverage issue and brought this up-to-date with develop.

Still need to update the PR description with the latest changes, and finish cleaning up the wiki page changes. After those & any lingering CI problems are addressed, this PR should be ready for review.

BenHenning:
> > TODO: Consider removing references to 'module' as well since that's Gradle specific in most uses and may add confusion.
> 
> Done. Also fixed the code coverage issue and brought this up-to-date with develop.
> 
> Still need to update the PR description with the latest changes, and finish cleaning up the wiki page changes. After those & any lingering CI problems are addressed, this PR should be ready for review.

I need to recheck this now that I'm trying to bring this PR into review again.

Edit: Looks like the Android Lint tool will need to be updated, too, given that the manifest has moved locations.

BenHenning:
I started working on fixing the manifest issues. As far as I can tell there are two:
1. The app module manifest has moved to be top-level. This is an easy fix as it's just updating some path and test setup references and I was able to get beyond that failure in local changes.
2. In the case of the data 'module' its manifest is being removed since it isn't needed. As far as I can tell this is incompatible with the lint tool as it expects all Gradle 'modules' to have manifest files.

This begs a bigger question of actually continuing with this approach of simulating Gradle modules for linting when we're very explicitly moving away from that concept.

@manas-yu I'm curious if you have any suggestions how we should proceed here. Is there a way we can adapt the Android lint tool's design to work in a less "Gradle-module"-esque way?

manas-yu:
As discussed the best way forward considering the priority of current tasks would be to delay the refactoring of the Android lint script, and to unblock this PR and unmoving the Android Manifest files for now(This would be handled in the PR related to the refactoring of the Android Lint Script).

The Lint script is currently:
- Structured in a Gradle project structure(this was done to maintain/improve the accuracy of the lint reports as lint liked a Gradle line environment)
- Analyses the codebase based on modules app, data, domain, util, and testing and finds there manifest files at <module>/src/main/manifest.xml
- Utilises bazel cquery command in the BazelClient based on the target module tags

To migrate the android lint script to execute on a project level instead of depending on modules:
- Remove the dependency of the script on [Modules](https://github.com/oppia/oppia-android/blob/b799ed3db7009c67aded18a25a02a7930a2ebe6c/scripts/src/java/org/oppia/android/scripts/lint/LintCommon.kt#L11) 
- Treat the entire project as one large module passing in the src, resource, test files
- For AAR/JAR files use a bazel command like `bazel cquery "deps(//:project)" --output=files`
- Check for any hardcoded module based paths like <module>/src/main/manifest.xml and migrate them to a project level context
- Remove [Module Deps](https://github.com/oppia/oppia-android/blob/b799ed3db7009c67aded18a25a02a7930a2ebe6c/scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt#L376) as they wont be needed.
- Make changes in the testing setup for the script for relevant testing.

Note: The refactoring can't guarantee similar accuracy to the current version as changes to the project description affects the reports majorly so the scripts accuracy would need to be tested again thoroughly.

BenHenning:
Thanks for that write-up @manas-yu! I've filed #6010 to track the tool-specific work here and will revert the changes that are breaking it.

BenHenning (commented):
Self-reviewed everything from past changes (more are still coming).

BenHenning (commented):
Self-reviewed latest (quick pass).

BenHenning (commented):
Spot check exemption update.

BenHenning:
At this point I expect tests to pass unless they flake due to the ongoing zlib issue.

BenHenning:
PTAL for codeowners @seanlip and @adhiamboperes.

PTAL specifically for infrastructure, documentation, and lint changes @manas-yu.

@adhiamboperes and @manas-yu you are probably both broadly interested in all of the changes here since they are so far-reaching.

/cc @subhajitxyz as well for FYI.

wiki/Overview-of-the-Oppia-Android-codebase-and-architecture.md
```diff
@@ -14,75 +14,75 @@ The Oppia Android codebase makes use of Kotlin and XML and can be built using An
 <img width="750" alt="Screenshot 2020-01-13 at 1 21 25 PM" src="https://user-images.githubusercontent.com/9396084/72245148-e60c4b80-3615-11ea-8a28-73b2b2bfc656.png">
 
 
-The Oppia project follows a standard Gradle project layout with five modules. Each subdirectory in a module contains files that are specific to a particular feature or purpose.
+The Oppia project follows an approximate Gradle project layout with five layers. Each subdirectory in a layer contains files that are specific to a particular feature or purpose.
```
seanlip, line 17: Do you still want the reference to Gradle here?

wiki/_Sidebar.md
```diff
@@ -52,7 +52,7 @@
   * [Triaging Process](https://github.com/oppia/oppia-android/wiki/Triaging-process)
   * Bazel
     * [Exploring Bazel](https://github.com/oppia/oppia-android/wiki/Bazel-User-Guide)
-    * [Gradle Bazel Migration Best Practices and FAQ](https://github.com/oppia/oppia-android/wiki/Gradle--Bazel-Migration-Best-Practices-and-FAQ)
+    * [Bazel Migration Best Practices and FAQ](https://github.com/oppia/oppia-android/wiki/Gradle--Bazel-Migration-Best-Practices-and-FAQ)
```
seanlip: Do you want to change the title of this page to not include Gradle?

seanlip (approved):
Thanks @BenHenning! LGTM for codeowners, left 2 comments.

wiki/_Sidebar.md
BenHenning: I really didn't want to rename the file, but I can.

wiki/Overview-of-the-Oppia-Android-codebase-and-architecture.md
BenHenning: Yep--I'm trying to communicate that the current project structure is roughly mimicking a Gradle project structure. I think this is the last or one of the last Gradle references left in the wiki after this PR.

wiki/_Sidebar.md
BenHenning: This has been done in the latest changes.

adhiamboperes (commented):
Apologies for the dely @BenHenning. I should be able to complete this review later in the day. Things are however looking very meticulous.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -4289,7 +4281,7 @@ test_file_exemption {
 }
 test_file_exemption {
   exempted_file_path: "testing/src/main/java/org/oppia/android/testing/junit/InitializeDefaultLocaleRule.kt"
-  override_min_coverage_percent_required: 38
+  override_min_coverage_percent_required: 100
```
adhiamboperes, line 4304: If we require full coverage for this file, shouldn't we just skip this override altogether like with the rest of the project?

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
```diff
@@ -18,7 +18,7 @@ class RegexPatternValidationCheckTest {
   private val outContent: ByteArrayOutputStream = ByteArrayOutputStream()
   private val originalOut: PrintStream = System.out
   private val activitiesPlacementErrorMessage =
-    "Activities cannot be placed outside the app or testing module."
+    "Activities cannot be placed outside the top-level app or testing directories."
```
adhiamboperes, line 21: After thinking about this(per CLaM meeting discussion), I think this isn't as confusing as I earlier thought.

wiki/End-to-End-Testing-Guide.md
```diff
@@ -10,12 +10,12 @@
...
 ## Overview
 
-End-to-End tests test the app from an end user’s experience by simulating the real user scenario and validating the system under test and its components for integration and data integrity. 
+End-to-End tests test the app from an end user’s experience by simulating the real user scenario and validating the system under test and its components for integration and data integrity.
 
 These tests play a major role in publishing the app. They run on a real or emulated device to make sure that our code interacts with the Android environment as expected, providing confidence in the final application or a feature when it's finished.
 
 End-to-End tests in Oppia-android are written using [UiAutomator](https://developer.android.com/training/testing/ui-automator).
-These tests are written under the instrumentation module and don’t have Gradle support.
+These tests are written under the instrumentation directory.
```
adhiamboperes, line 18: Should this maybe be - 

```suggestion
These tests are written in a directory under the instrumentation layer.
```

adhiamboperes (approved):
Finished reviewing this @BenHenning, nothing blocking. Generally LGTM.

Apologies for the delay!

scripts/assets/test_file_exemptions.textproto
BenHenning: Technically this surpasses the minimum which is currently set to 70%:

https://github.com/oppia/oppia-android/blob/27d80bdeceefd28051ccfaf8b6139132b326c0fb/scripts/src/java/org/oppia/android/scripts/coverage/reporter/CoverageReporter.kt#L11

scripts/src/javatests/org/oppia/android/scripts/regex/RegexPatternValidationCheckTest.kt
BenHenning: Ack--we can definitely iterate on terminology if people do get confused moving forward though.

wiki/End-to-End-Testing-Guide.md
BenHenning: I don't think so. The instrumentation tests are not really a layer (at least not an app architectural layer which is what we've been using 'layer' to mean in this PR) which is actually why I kept it to directory here.

BenHenning:
I think I can address both of your comments @seanlip and @adhiamboperes, thanks for the reviews! Note that I didn't change anything from your comments @adhiamboperes (see my replies). Please let me know if there is something you'd like me to change--I'm happy to consider that for a follow-up PR as needed.

BenHenning (commented):
Checked merge commit & earlier reviewer address commit.

BenHenning:
This should be good to merge once CI is passing. Hopefully it won't take too many re-runs.

PR #5627, opened 2024-12-28
Fix #4097: Adding tests for math utils

manas-yu:
## Explanation

Fix #4097
 This PR adds test files for the `FractionSubject`, `RealSubject`, `MathExpressionSubject`, and `MathEquationSubject` classes and includes a `BUILD.bazel` file to enable building and running these tests.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

PR #5655, opened 2025-01-17
Fix #4097, Fix #4848, Fix #4170: Resolving TODOs

### LINKED ISSUE #4097: Add tests for math subjects: FractionSubject, RealSubject, MathExpressionSubject, MathEquationSubject
#4047 introduces four new custom Truth test subjects: FractionSubject, RealSubject, MathExpressionSubject, MathEquationSubject. Each of these would be advantageous to add thorough tests for to ensure that:
- Each of the subject & comparator methods correctly allow users to test specific properties of the protos each subject helps verify
- Each of the subject & comparator methods are verified to fail when expected (i.e. when used with an incompatible proto property) with the correct failure reason being verified

### LINKED ISSUE #4170: Fix vertical alignment for locally rendered and cached LaTeX

### LINKED ISSUE #4838: Introduce support for admins to mark lesson completions for profiles

### LINKED ISSUE #4848: Fix HTML generation for content descriptions (particularly for answers)

manas-yu:
## Explanation

Fixes #4097, Fix #4848  and Fix #4170
This PR resolves the TODOs left in #5627, #5614 and #5647.

1. **Integrating `getContentDescription` in `HintViewModel`:**
   - Replaced `fromHtml` with `getContentDescription` introduced in #4848.
   - Added `ConceptCardTagHandler` to `customTagHandlers` for handling `CUSTOM_CONCEPT_CARD_TAG`.
   - Injected `ConceptCardLinkClickListener` into the `HintsAndSolutionViewModel.Factory`.

2. **Resolving TODOs for #4097:**
   - Removed TODOs and text-file-exemptions(in scripts/assets) for `FractionSubject`, `MathExpressionSubject`, `RealSubject`, and `MathEquationSubject`.

3. **Resolving TODOs for #4170 :**
    - Removed TODO comment from `MathTagHandler`.
## Essential Checklist


domain/src/main/assets/test_exp_id_2.json
```diff
@@ -524,7 +524,7 @@
           "hints": [{
             "hint_content": {
               "content_id": "hint_6",
-              "html": "<p>Remember that two halves, when added together, make one whole.</p><p>Click on this <oppia-noninteractive-skillreview skill_id-with-value=\"&amp;quot;test_skill_id_1&amp;quot;\" text-with-value=\"&amp;quot;test_skill_id_1 concept card&amp;quot;\"></oppia-noninteractive-skillreview>.</p>"
```
manas-yu: In [ConceptCardTagHandler.kt](https://github.com/oppia/oppia-android/pull/5614/files#diff-68dca4c94553feb9ac44cb42824ce02993af00396a111a84a26a674b1bb7b6e5), would it be more appropriate to use "$text concept card $skillId", or would simply using "text" be sufficient?

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParserEntityTypeModule.kt
```diff
@@ -1,11 +1,21 @@
 package org.oppia.android.util.parser.html
 
+import android.view.View
 import dagger.Module
 import dagger.Provides
+import javax.inject.Singleton
 
 /** Provides Html parsing entity type dependencies. */
 @Module
 class HtmlParserEntityTypeModule {
```
manas-yu, line 9: Would it be better to create a separate module for `TagHandlers`, or could we make the `HtmlParserEntityTypeModule` more generic to include TagHandlers as well?

manas-yu:
Hey @adhiamboperes, PTAL. I've left a few comments where I’d like a second opinion. Thanks!

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParserEntityTypeModule.kt
adhiamboperes: Adding a new dagger module with our configuration is a lot of work, so I would avoid it as much as possible.

domain/src/main/assets/test_exp_id_2.json
adhiamboperes: Could you please provide examples?

adhiamboperes (changes requested):
Thanks @manas-yu! In addition to my inline comments, Please a PR description explaining your changes briefly

domain/src/main/assets/test_exp_id_2.json
manas-yu: If `text-with-value` contains the skill ID (e.g., ["test_skill_id_1 concept card"](https://github.com/oppia/oppia-android/blob/f98c410e64bb4968a73aa428eef8f1d498e8af9a/domain/src/main/assets/test_exp_id_2.json#L527)), using `"$text concept card $skillId"` results in `"test_skill_id_1 concept card concept card test_skill_id_1"`, which is redundant. In this case, using `"$text"` alone would generate `"test_skill_id_1 concept card"`, which is more appropriate.

However, if text-with-value does not contain the skill ID (e.g., ["refresher lesson"](https://github.com/oppia/oppia-android/blob/f98c410e64bb4968a73aa428eef8f1d498e8af9a/utility/src/test/java/org/oppia/android/util/parser/html/ConceptCardTagHandlerTest.kt#L43-L57)), using `"$text concept card $skillId"` results in `"refresher lesson concept card skill_123"`, which provides necessary context, while using just "text" would not give enough context.

Overall, if text-with-value already refers to the skill ID, including both the text and the skill ID in the description is unnecessary and can be avoided.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParserEntityTypeModule.kt
manas-yu: To resolve the `Dagger/MissingBinding` issue for the `ConceptCardLinkClickListener` in [ExplorationActivityTest](https://github.com/oppia/oppia-android/pull/5655/files#diff-f5dcd6d5f1dea40aa1f0414cb8839d3fa80d5c8efc3781be6dd6ded20450fa7a), I explored three potential approaches:

1. **Current Approach**: Providing the listener interface directly.

2. **Direct Handler Injection**: Injecting the handler directly into the dependent class.

3. **Handler Factory Approach**:
   - Create a factory interface for each tag handler (e.g., `ConceptCardTagHandler` in this case).
   - Implement a base factory to manage all custom tag handlers within `CustomHtmlContentHandler`.
   - Update the Dagger module to provide these factories as dependencies.

After reviewing the [Dagger Wiki](https://github.com/oppia/oppia-android/wiki/Dagger), it seems that the **Handler Factory Approach** is the best solution.

Before proceeding, I wanted to seek a second opinion on the approach to confirm if this aligns with the project's architectural principles. Additionally, it would be helpful to reference any existing implementations within the codebase that address similar issues, so I can use them as a guide for this solution. Let me know your thoughts!

manas-yu:
Hi @adhiamboperes PTAL, I have added some comments and updated the PR description.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParserEntityTypeModule.kt
adhiamboperes: You're correct in that the factory pattern is the preferred way for handling this. So instead of overriding the interface here, you will instead provide/return an instance of the factory. Take a look at `DebugLogReportingModule > providePerformanceMetricsEventLogger` for example.

domain/src/main/assets/test_exp_id_2.json
adhiamboperes: It sounds better to use text only, and isn't this the existing pattern?

manas-yu: The current implementation was based on `"$text concept card $skillId."`([ConceptCardTagHandler.kt](https://github.com/oppia/oppia-android/pull/5614/files#diff-68dca4c94553feb9ac44cb42824ce02993af00396a111a84a26a674b1bb7b6e5)), I will update it to use just the `text` for now. Thanks for the clarification!

manas-yu:
Hi @adhiamboperes PTAL, Thanks!

utility/src/main/java/org/oppia/android/util/parser/html/ConceptCardTagHandler.kt
```diff
@@ -52,10 +53,18 @@ class ConceptCardTagHandler(
...
-    return if (!skillId.isNullOrBlank() && !text.isNullOrBlank()) {
-      "$text concept card $skillId"
+    return if (!text.isNullOrBlank()) {
+      text
     } else ""
   }
+
+  /** Application-injectable factory for creating [ConceptCardTagHandler]s). */
+  class Factory @Inject constructor() {
+    fun createConceptCardLinkClickListener(): ConceptCardLinkClickListener {
```
adhiamboperes, line 71: I believe this should have a KDoc since it is a public function. KDoc check seems to not have caught this.

adhiamboperes (changes requested):
Thanks @manas-yu! Only one minor comment left.

manas-yu:
Hi @adhiamboperes PTAL, Thanks!

adhiamboperes (approved):
LGTM, thanks @manas-yu!

PR #5664, opened 2025-01-24
Fix #5660: Ensure mobile-install works

### LINKED ISSUE #5660: [Feature Request]: Add support for mobile-install

This discussion came up from https://github.com/oppia/oppia-android/discussions/5643#discussioncomment-11874290 and a discussion during the CLaM team meeting. The codebase currently doesn't support mobile-install (or at least, not well per that comment thread and #2610). The codebase is currently configured with both APK and AAB binary declarations, but we plan to remove the APK ones entirely since it simplifies things: we'd only be building AABs (including for uploading to the Play Console).

There are two problems that need to be solved:
1. Figure out a way to declare APK targets that use the AABs (rather than being separately defined as they are today). This is probably not super tricky. We use [``bundletool``](https://developer.android.com/tools/bundletool) for both assembling the AAB and generating a device-compatible version for installation. We don't actually need to do that latter, however--we could automatically declare "universal" APKs generated from the AAB which would be device agnostic (and thus could be declared via a custom Bazel rule). Doing so would provide a valid APK target that could be directly installed (i.e. with ``adb install``).
2. Figure out how to make ``bazel mobile-install`` work with the above (or at all). It's not clear whether specifically ``android_binary`` needs to be used, or if the target just needs specific Bazel [aspects](https://bazel.build/extending/aspects) to be configured in order to work. My last chat with the Bazel Android team about mobile-install (a few years ago) was that its support is very limited and not currently well maintained, so this may require fixes to mobile-install itself in Bazel to ultimately be made to work.

The alternative is continuing the status quo of using a separate script for installing a device-compatible APK (e.g. using ``//:install_oppia_dev``), or by using separate APK targets (which isn't ideal since we want to keep parity between development and uploading binaries).

### Additional context

_No response_

BenHenning:
## Explanation

Fixes #5660

I've verified with local testing that ``bazel mobile-install`` works with the APK generated for each AAB target (it ends in ``_binary``, e.g. ``oppia_dev_binary``). Thus, documentation has been updated to recommend this as the main way to install the local development version of the app as ``mobile-install`` provides substantial performance benefits (see https://bazel.build/docs/mobile-install).

The ``install_*`` targets (e.g. ``install_oppia_dev``) have been removed since ``mobile-install`` should be used, instead. The existing Starlark rule & macro has been repurposed to generating a universal APK which, in turn, will be useful for specific situations where ``mobile-install`` breaks down or is not used (such as for emulator tests).

## Essential Checklist

## For UI-specific PRs only
N/A -- this changes nothing about the build of the app, only installation pathways.


BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @adhiamboperes.

oppia_android_application.bzl
```diff
@@ -158,61 +158,57 @@ def _package_metadata_into_deployable_aab_impl(ctx):
...
+        "--bundle=%s" % input_aab_file.path,
         "--output=%s" % apks_file.path,
         "--ks=%s" % debug_keystore_file.path,
         "--ks-pass=pass:android",
         "--ks-key-alias=androiddebugkey",
         "--key-pass=pass:android",
+        "--mode=universal",
     ]
 
+    # bundletool only generats an APKs file, so the universal APK still needs to be extracted.
```
adhiamboperes: Nit:
```suggestion
    # bundletool only generate an APKs file, so the universal APK still needs to be extracted.
```

oppia_android_application.bzl
```diff
@@ -158,61 +158,57 @@ def _package_metadata_into_deployable_aab_impl(ctx):
...
         outputs = [apks_file],
-        inputs = ctx.files.input_file + ctx.files.debug_keystore,
+        inputs = ctx.files.input_aab_file + ctx.files.debug_keystore,
         tools = [ctx.executable._bundletool_tool],
         executable = ctx.executable._bundletool_tool.path,
-        arguments = generate_apks_arguments,
-        mnemonic = "BuildApksFromDeployAab",
-        progress_message = "Preparing AAB deploy to device",
+        arguments = generate_universal_apk_arguments,
+        mnemonic = "GenerateUniversalAPk",
```
adhiamboperes:
```suggestion
        mnemonic = "GenerateUniversalAPK",
```

oppia_android_application.bzl
```diff
@@ -158,61 +158,57 @@ def _package_metadata_into_deployable_aab_impl(ctx):
...
+    mv universal.apk "$(pwd)/{1}"
+    """.format(apks_file.path, output_apk_file.path)
+
+    # Reference: https://docs.bazel.build/versions/main/skylark/lib/actions.html#run_shell.
+    ctx.actions.run_shell(
+        outputs = [output_apk_file],
+        inputs = [apks_file],
+        tools = [],
+        command = command,
+        mnemonic = "ExtractUniversalAPk",
```
adhiamboperes:
```suggestion
        mnemonic = "ExtractUniversalAPK",
```

oppia_android_application.bzl
```diff
@@ -390,30 +394,30 @@ def oppia_android_application(name, config_file, proguard_generate_mapping, **kw
...
         )
 
-def declare_deployable_application(name, aab_target):
+def generate_universal_apk(name, aab_target):
     """
-    Creates a new target that can be run with 'bazel run' to install an AAB file.
+    Creates a new installable universal APK target for the provided AAB target.
+
+    This new target is installable via 'adb install'. This should generally never be used for direct
+    development installs as it will be far less performant than 'bazel mobile-install'.
```
adhiamboperes: I find this ambiguous, it is not clear to me whether the warning is against using 'adb install' or `bazel mobile-install //:oppia_prod_universal_apk`

adhiamboperes (changes requested):
Thanks @BenHenning! I have a few comments. PTAL.

Rd4dev:
@BenHenning, @adhiamboperes, 

Just wanted to request testing for compatibility with Windows regarding the `mobile-install`. Earlier, I tested it on Windows, and the app crashed immediately after installation when using mobile-install, 

>when installed with `bazel mobile-install //:oppia_dev_binary`




https://github.com/user-attachments/assets/f064887b-46b2-4f54-9fbc-4a16626dd6fb





while it worked fine when installed via `//:install_oppia_dev` (as mentioned in the CLaM chat thread). 


I’ve recently switched to Linux, and `bazel mobile-install //:oppia_dev_binary` works well here, so I’m just seeking clarity on whether the post-PR merge would still allow Windows developers to install it successfully, since the `//:install_oppia_dev` would no longer be available, or if the PR merge would address this issue.


<h1></h1>


It might be something just with my Windows setup, so thought would add it to the discussion to see if any other Windows users face the same.

adhiamboperes:
> @BenHenning, @adhiamboperes,
> 
> Just wanted to request testing for compatibility with Windows regarding the `mobile-install`. Earlier, I tested it on Windows, and the app crashed immediately after installation when using mobile-install,
> 
> > when installed with `bazel mobile-install //:oppia_dev_binary`
> 
>  screen-20250126-194259.mp4 
> while it worked fine when installed via `//:install_oppia_dev` (as mentioned in the CLaM chat thread).
> 
> I’ve recently switched to Linux, and `bazel mobile-install //:oppia_dev_binary` works well here, so I’m just seeking clarity on whether the post-PR merge would still allow Windows developers to install it successfully, since the `//:install_oppia_dev` would no longer be available, or if the PR merge would address this issue.
> 
> It might be something just with my Windows setup, so thought would add it to the discussion to see if any other Windows users face the same.


Thanks for flagging this @Rd4dev, I tested this `bazel mobile-install //:oppia_dev_binary` on my Mac, and recorded the stacktrace:

```
FATAL EXCEPTION: main        
 Process: org.oppia.android, PID: 9508
    java.lang.RuntimeException: Unable to instantiate application com.google.devtools.build.android.incrementaldeployment.StubApplication package org.oppia.android: java.lang.SecurityException: Writable dex file '/data/local/tmp/incrementaldeployment/org.oppia.android/dex/incremental_classes4.dex' is not allowed.
     at android.app.LoadedApk.makeApplicationInner(LoadedApk.java:1466)
     at android.app.LoadedApk.makeApplicationInner(LoadedApk.java:1395)
     at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6959)
     at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2236)
     at android.os.Looper.loopOnce(Looper.java:205)
     at android.os.Looper.loop(Looper.java:294)
     at android.app.ActivityThread.main(ActivityThread.java:8177)
     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
    Caused by: java.lang.SecurityException: Writable dex file '/data/local/tmp/incrementaldeployment/org.oppia.android/dex/incremental_classes4.dex' is not allowed.
```

This only occured when I used an API 34 emulator. The app works fine when built on API 29. I encountered this error during my initial setup(https://github.com/oppia/oppia-android/discussions/5643)

BenHenning:
Thanks @Rd4dev and @adhiamboperes, this is really helpful.

@Rd4dev per Adhiambo's comment, which SDK version were you installing to, and are you using WSL2 or building with native Windows? (I'm assuming WSL2 since native Windows builds still do not work for me).

Also, it is the intent that we maintain current support for all three platforms, so I won't submit this PR until we can confirm it working in at least the places we know it works today with the install script.

oppia_android_application.bzl
BenHenning: Fixed--nice catch!

oppia_android_application.bzl
BenHenning: Fixed.

oppia_android_application.bzl
BenHenning: Fixed.

oppia_android_application.bzl
BenHenning: Ah, yeah this is just very confusing. I decided to remove the 'adb install' bit since it's not important to know, and tried to make the other parts of the documentation here a bit clearer. I did add a new 'adb install' section to cover a detail I didn't think about originally--there are cases when mobile-install can actually be detrimental to use. It's mainly useful for development since it can substantially decrease install time, but it runs different code than a real production APK so it's less good for testing production builds.

Does this look better to you?

BenHenning:
Updated to latest develop & addressed comments. PTAL @adhiamboperes.

Regarding the failure, if you were using an SDK 34 device also @Rd4dev then it seems very likely both you and @adhiamboperes ran into https://github.com/bazelbuild/bazel/issues/20456. I'm not actually sure what we should do here. I lean toward continuing with this PR even though it means SDK 34+ devices can't use mobile-install. We also have the option to use adb install for those, and we can add a caveat to the wiki accordingly until we migrate to the Starlark rules_android (which I've attempted to do a few weekends ago--it's not an easy change). WDYT?

BenHenning (commented):
Self-reviewed changes & followed up to reviewer comments.

Rd4dev:
@BenHenning, it seems my handset was running API 14, and after testing with a Linux setup on SDK 34 emulator, the app crashed with the same stack trace @adhiamboperes recorded. Most solutions for the API 14 behavior issue suggested making the DEX files read-only, which I believe was addressed in the bazel incremental script, though I wasn’t sure if that's possible or how to implement that directly in the codebase. Adding ADB support seems reasonable for SDK 34+ and production builds. 

Separately, since the setup now uses `mobile-install`, would it be helpful to include the optional `--start_app` flag in the wiki documentation for development?

adhiamboperes (approved):
LGTM, thanks @BenHenning.

oppia_android_application.bzl
adhiamboperes: This is clear now.

adhiamboperes:
> Updated to latest develop & addressed comments. PTAL @adhiamboperes.
> 
> Regarding the failure, if you were using an SDK 34 device also @Rd4dev then it seems very likely both you and @adhiamboperes ran into [bazelbuild/bazel#20456](https://github.com/bazelbuild/bazel/issues/20456). I'm not actually sure what we should do here. I lean toward continuing with this PR even though it means SDK 34+ devices can't use mobile-install. We also have the option to use adb install for those, and we can add a caveat to the wiki accordingly until we migrate to the Starlark rules_android (which I've attempted to do a few weekends ago--it's not an easy change). WDYT?


@BenHenning, since this is not a permanent issue, and would be eliminated in future upgrades of Bazel(8+), or migrating to the Starlark rules_android, I am fine with accepting this PR with documented restrictions.

BenHenning:
> @BenHenning, it seems my handset was running API 14, and after testing with a Linux setup on SDK 34 emulator, the app crashed with the same stack trace @adhiamboperes recorded. Most solutions for the API 14 behavior issue suggested making the DEX files read-only, which I believe was addressed in the bazel incremental script, though I wasn’t sure if that's possible or how to implement that directly in the codebase. Adding ADB support seems reasonable for SDK 34+ and production builds.
> 
> Separately, since the setup now uses `mobile-install`, would it be helpful to include the optional `--start_app` flag in the wiki documentation for development?

I haven't tested this with `--start_app` actually, does it work correctly? If so, we could also update `.bazelrc` and have this flag be default for whenever someone uses `mobile-install`. Do you have any thoughts on this @Rd4dev or @adhiamboperes?

BenHenning (commented):
PTAL @adhiamboperes and @Rd4dev at latest wording.

If you can, can you double check that both mobile-install and the `adb install` command work as expected @Rd4dev? For the former, I want to verify that `start_app` actually works and, for the latter, that the exact `adb install` command correctly installs a workable version of the app on SDK 34. If you can't, no worries--I can check this at home later as well.

Rd4dev:
@BenHenning, yes, I checked the commands earlier and just double-checked them now for 
- SDK 28 (using `mobile-install` with `--start_app` in the command line and also including it in the `.bazelrc`)  
- SDK 34, 35 (performing a rebuild + running `adb install bazel-bin/oppia_dev_binary.apk`). 
 
All the above mentioned runs did install a workable version of the app. 
Including `--start_app` in `.bazelrc` would be very convenient, but I’d like to confirm if it works on Mac, as there are discussions raised here - https://github.com/bazelbuild/bazel/issues/21574. 

Also, these are tested only on Linux setup.

BenHenning:
Ah, I actually forgot that I enabled auto-merge. Hmm. :)

Thanks so much @Rd4dev!

@adhiamboperes can you please double check the wording in the latest changes, and maybe double check `mobile-install` on OS X per @Rd4dev's latest comment? I can address any issues in a follow-up.

adhiamboperes:
> Ah, I actually forgot that I enabled auto-merge. Hmm. :)
> 
> Thanks so much @Rd4dev!
> 
> @adhiamboperes can you please double check the wording in the latest changes, and maybe double check `mobile-install` on OS X per @Rd4dev's latest comment? I can address any issues in a follow-up.


Double checked, everything looks good, the commands work as expected.

BenHenning:
Awesome, thanks much for checking again @adhiamboperes!

Issue #5645, opened 2025-01-08
[BUG]: Flicker in Option Section During Drag and Drop with Merging Tiles [Maybe blocked: #171]


manas-yu:
### Describe the bug

While dragging the tiles in the "Drag and Drop" feature with the merging tiles, the option section exhibits a flickering effect. This occurs when the box containing the options briefly changes its height after the drop action. This issue does not happen during regular drag and drop questions but only when the merge button is available.

### Steps To Reproduce

1. build the application in developer environment
2. scroll down and select the "First Test Topic" topic
3. Select the Prototype Exploration
4. Play till the 9th question (drag and drop with merge tile option)
5. After you drag and drop a tile, the options section flickers

### Expected Behavior

The options section should not flicker.

### Screenshots/Videos

[screen_flicker.webm](https://github.com/user-attachments/assets/fa3d68e3-c4ac-4658-a040-90c050a99675)

### What device/emulator are you using?

Nexus 5X API 30

### Which Android version is your device/emulator running?

Android 11

### Which version of the Oppia Android app are you using?

0.14-beta-17f2ef3044

Tejas-67:
Hi @adhiamboperes @manas-yu I would like to work on this issue. Please assign this to me.

TanishMoral11:
Hey @Tejas-67  , we’d love to hear your proposed solution. If it sounds good, we’ll assign it to you and you can create a PR.

Neer-rn:
Hi @adhiamboperes, I would like to work on this issue. I actually already spent quite some time investigating this before commenting so let me share what I found.
                                                                                                                                                                                              
  So first off, I tried reproducing this on my physical device and honestly I could not see any flicker at all. I was really confused, I tried multiple times, dragged and dropped tiles in question 9 of Prototype Exploration, but everything looked smooth. I was like okay maybe this is already fixed? But the issue is still open so that didn't make sense.

  Then I noticed the video in the issue was recorded on Nexus 5X API 30 emulator. So I downloaded the exact same emulator config and tested again. And there the flicker was clearly visible. Every time I dropped a tile, the options container would briefly shrink and expand back. So the issue definitely exists, it's just not visible on physical devices.

[...]

Issue #1612, opened 2020-08-11
Enable proper testing of drag & drop interaction


BenHenning:
After some investigation, I realized that drag & drop can't be correctly tested on Robolectric due to a couple of issues:
1. Robolectric shadows ViewGroup's ``requestDisallowInterceptTouchEvent`` and ``onInterceptTouchEvent``, both of which are essential for ``ItemTouchHelper`` to process a stream of touch events simulating drag. I don't think this issue is hit with DragDropTestActivityTest (which does pass in Robolectric) because the latter doesn't use data-binding, and the view hierarchy is much simpler. The solution is to update ``ShadowViewGroup`` to call forweard to the real intercept touch event methods.
2. With larger views (such as drag & drop in StateFragment), moving these around can trigger ItemTouchHelper to think it should enable auto-scroll. Auto-scrolling results in Robolectric entering an infinite loop.

(1) requires a fix to Robolectric, whereas (2) can be solved by introducing our own simple Shadow:
```java
import org.robolectric.annotation.Implementation;
import org.robolectric.annotation.Implements;

@Implements(className = "androidx.recyclerview.widget.ItemTouchHelper",
isInAndroidSdk = false, minSdk = 19, maxSdk = 29)
public final class ShadowTouchItemHelper {

  @Implementation
  protected boolean scrollIfNecessary() {
    // Never scroll while dragging since this can put Robolectric into an infinite loop. Prefer to
    // let the test fail, instead.
    return false;
  }
}
```

yurilev:
Possibly relevant: https://github.com/oppia/oppia-android/issues/1769#issuecomment-738056830

PR #5677, opened 2025-01-29
Fix #2974: Reducing Long Press Time in Drag-Drop

### LINKED ISSUE #2974: Drag & drop is slow to respond on slow devices
**Describe the bug**
When using drag & drop on a slow device (e.g. Nexus 5X), the user has to tap & press for several seconds before getting the haptic feedback they've selected an item and can  now drag it.

**To Reproduce**
1. Use a slow device (such as a Nexus 5X)
2. Navigate to the prototype exploration in a local developer build
3. Play to the drag & drop interaction
4. Try to drag items quickly
5. Observe that they don't move at all due to a long tap & hold timeout

**Expected behavior**
Items should respond faster, or we should have instructions to make it clear that the user needs to tap and hold to drag. The latter may be more general-purpose, and mitigate even slower devices having a longer-than-expected delay. The key problem here is that it's unclear to the user that they're actually doing something correctly when trying to drag items.

**Screenshots**
N/A

**Device**
- Device being used: Nexus 5X
- SDK version: 27

**Additional context**
This was found while testing alpha MR2.

manas-yu:
## Explanation

Fix #2974
This PR reduces the long press duration(from the default `500ms` to `300ms`) by bypassing the default ItemTouchHelper long-press detection and implementing custom logic for triggering drag events.

- The `RecyclerView.OnItemTouchListener`, overrides the default behavior of ItemTouchHelper by handling `MotionEvent` actions directly.
- The `ItemTouchHelper` is still used to perform the actual drag-and-drop functionality.

(recordings are from Nexus 5X API 30)
## Before
https://drive.google.com/file/d/1_igIDW5y8GtCgGgzE1s7ICCnXw4eLNaA/view?usp=sharing

## After
https://drive.google.com/file/d/1760fz4orlX5ZIzclZTkQwkW-UWlYlGts/view?usp=sharing

(note: the flickering issue is being tracked in #5645)

## Essential Checklist


manas-yu:
Hi @adhiamboperes, could you PTAL at this approach? The workaround is functioning well based on manual testing, but I’d love to get your insights on its performance and architectural impact. Let me know your thoughts. Thanks!

After some investigation, it appears that `ItemTouchHelper's` long press duration is a fixed gesture with no straightforward way to optimize drag-and-drop behavior. As a workaround, intercepting touch events in `DragDropSortInteractionView` seemed like a valid approach. Ref: https://stackoverflow.com/questions/7919865/detecting-a-long-press-in-android?rq=3.

adhiamboperes (commented):
@manas-yu, overall, the implemantation appears solid. 

The "after" video does not play. Could you please take a look?

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
```diff
@@ -45,10 +51,68 @@ class DragDropSortInteractionView @JvmOverloads constructor(
   private lateinit var dataList: List<DragDropInteractionContentViewModel>
   private lateinit var onDragEnd: OnDragEndedListener
   private lateinit var onItemDrag: OnItemDragListener
+  private var itemTouchHelper: ItemTouchHelper? = null
+  companion object {
+    private const val LONG_PRESS_TIMEOUT_MS = 30L
```
adhiamboperes: This is extremely short for a long press. Users might not even notice it. Consider increasing it to up to 300ms.

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
```diff
@@ -45,10 +51,68 @@ class DragDropSortInteractionView @JvmOverloads constructor(
...
+  }
+
+  private val touchListener = object : OnItemTouchListener {
+    private var startX = 0f
+    private var startY = 0f
+    private val touchSlop = ViewConfiguration.get(context).scaledTouchSlop
+    private var currentTimer: LiveData<Any>? = null
+
+    override fun onInterceptTouchEvent(rv: RecyclerView, e: MotionEvent): Boolean {
+      val fragment = FragmentManager.findFragment<Fragment>(this@DragDropSortInteractionView)
```
adhiamboperes: This is done inside touch events, meaning it’s called frequently. Consider for example, 
```
private val fragment: Fragment by lazy {
    FragmentManager.findFragment<Fragment>(this@DragDropSortInteractionView)
}
```

adhiamboperes: Can use the same variable for `cancelCurrentTimer` below.

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
```diff
@@ -45,10 +51,68 @@ class DragDropSortInteractionView @JvmOverloads constructor(
...
+    }
+
+    private fun cancelCurrentTimer() {
+      val fragment = FragmentManager.findFragment<Fragment>(this@DragDropSortInteractionView)
+      currentTimer?.let { timer ->
+        fragment.viewLifecycleOwner.let { lifecycleOwner ->
+          timer.removeObservers(lifecycleOwner)
+        }
+      }
+      currentTimer = null
```
adhiamboperes, line 121: This lookup happens every time you cancel the timer. If the viewLifecycleOwner changes (e.g., during fragment recreation), it may cause issues.

```suggestion
    private var viewLifecycleOwner: LifecycleOwner? = null

init {
    fragment.viewLifecycleOwnerLiveData.observe(fragment) { owner ->
        viewLifecycleOwner = owner
    }
}

private fun cancelCurrentTimer() {
    viewLifecycleOwner?.let { owner ->
        currentTimer?.removeObservers(owner)
    }
    currentTimer = null
}
```

Now, viewLifecycleOwner is always up to date. Prevents possible crashes when accessing viewLifecycleOwner.

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
```diff
@@ -45,10 +51,68 @@ class DragDropSortInteractionView @JvmOverloads constructor(
...
+          fragment.let { fragmentOwner ->
+            currentTimer = lifecycleSafeTimerFactory.createTimer(LONG_PRESS_TIMEOUT_MS).apply {
+              observe(fragmentOwner.viewLifecycleOwner) {
+                findChildViewUnder(startX, startY)?.let { childView ->
+                  findContainingViewHolder(childView)?.let { viewHolder ->
+                    itemTouchHelper?.startDrag(viewHolder)
+                  }
+                }
+                removeObservers(fragmentOwner.viewLifecycleOwner)
+              }
```
adhiamboperes: This creates and attaches an observer every time a long press is detected.

```suggestion
      currentTimer = lifecycleSafeTimerFactory.createTimer(LONG_PRESS_TIMEOUT_MS).apply {
    observeOnce(viewLifecycleOwner ?: return) {
        findChildViewUnder(startX, startY)?.let { childView ->
            findContainingViewHolder(childView)?.let { viewHolder ->
                itemTouchHelper?.startDrag(viewHolder)
            }
        }
    }
}
```

where:

```
private fun <T> LiveData<T>.observeOnce(owner: LifecycleOwner, observer: Observer<T>) {
    observe(owner, object : Observer<T> {
        override fun onChanged(t: T?) {
            observer.onChanged(t)
            removeObserver(this)
        }
    })
}
```
Observer automatically removes itself after execution and prevents unnecessary re-attachment of observers.

adhiamboperes (changes requested):
Thanks @manas-yu! Left some suggested improvements. Please note that the code suggestions are untested, but meant to provide a general guidance.

manas-yu:
> The "after" video does not play. Could you please take a look?

The 'after' video was playing fine on my end, but I've replaced it with the Drive link just in case. I've also demonstrated how the frequency of hold-release has changed in both recordings.

manas-yu:
Just an update for the PR, the implementation is working fine with the requested changes in manual testing, and I'm currently trying to make the respective tests pass as it seems to not work with the current test env.
also tried https://github.com/nifux12/oppia-android/pull/20/files#diff-0b2672e38b1e7b9882816d25565fe60a854aac0f6bc46a0b667785103a3dc689

manas-yu:
Hi @adhiamboperes, could you PTAL?

I'm currently unable to replicate and test the `OnItemTouchListener` in `DragDropTestActivityTest`, even though it works fine in manual testing. 

Previously, the tests relied on `ItemFacilitator`, but now, due to `OnItemTouchListener`, the long press is handled in the view itself, while drag-and-drop is still managed by `ItemFacilitator`. 

So far, I've attempted two main approaches for testing:

1. https://github.com/nifux12/oppia-android/pull/20/files#diff-0b2672e38b1e7b9882816d25565fe60a854aac0f6bc46a0b667785103a3dc689
2. https://github.com/oppia/oppia-android/pull/5677/files

Would appreciate if you could point out where I'm going wrong or provide any advice like if the implementation needs to be changed or the approach with the testing is incorrect. Thanks!

manas-yu:
@adhiamboperes @BenHenning PTAL, After further investigation, I'm mostly certain that the failing tests (all in `DragDropTestActivityTest` and 2 in `StateFragmentTest`) are related to #1612. Below are the reasons why I believe so:  

### **StateFragmentTest**  

> With larger views (such as drag & drop in `StateFragment`), moving these around can trigger `ItemTouchHelper` to think it should enable auto-scroll. Auto-scrolling results in Robolectric entering an infinite loop.  

- In the `develop` branch, all tests utilizing `fun dragAndDropItem` are being tracked in #1612, except for the two `errorMessageIsReset` tests, which are failing to the current implementation.  
- Both tests fail to perform the drag-drop action in `develop` branch, which I verified through [assertion](https://github.com/oppia/oppia-android/blob/628d8adcfc19b6e2907f320fddea36649106349d/app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt#L1112-L1118). However, they do successfully replicate the `LongPress` action.  
- Since the `LongPress` is being replicated, the error message resets as expected—even without the actual drag-drop occurring. However, because `LongPress` is now handled in `onInterceptTouchEvent`, the tests fail even after using `testCoroutineDispatchers` in the tests.  
- While debugging through print statements for these tests, the `OnItemTouchListener` seemed to log correctly while the `ItemTouchHelper` did not work.

### **DragDropTestActivityTest**  

> I don't think this issue affects `DragDropTestActivityTest` (which does pass in Robolectric) because it doesn't use data-binding, and its view hierarchy is much simpler.  

- Previously, drag-drop functionality was entirely dependent on `ItemFacilitator`.  
- However, due to `onInterceptTouchEvent` being in the View, neither `LongPress` nor drag-drop is being successfully replicated, even when using `testCoroutineDispatchers` tests are failing and items are not being dragged.  
- While debugging through print statements for these tests, both the `OnItemTouchListener` and `ItemTouchHelper` failed to log.
---
I have manually tested the current implementation(updated the "after" video too) for the failing tests, and the drag-drop is working fine. The error message is also resetting as expected. 

For now, I can add `@RunOn(TestPlatform.ESPRESSO)` to the failing tests to make the CI pass. However, if my reasoning is incorrect or if there is a workaround, please let me know.  

Thanks!

manas-yu:
Hi @adhiamboperes, PTAL. The tests should now pass.

The issue was due to `isLongPressDragEnabled` being set to `false` (thanks @BenHenning for pointing it out). I initially disabled it, assuming it might interfere with long-press handling in `OnItemTouchListener`, given that long-press handling in `OnItemTouchListener` was independent of `ItemTouchHelper`'s long-press handling and shouldn't have required it to be enabled in the current implementation.

However, after enabling it, the tests are passing, and the final UI result remains unchanged. Let me know if this implementation is sufficient or if anything else is required. Thanks!

adhiamboperes (approved):
Thanks @manas-yu! Seems like everything has been resolved.

PR #5696, opened 2025-02-10
Fix #5143: Update robolectric to 4.7 and update espresso-core to 3.5-alpha07

### LINKED ISSUE #5143: Update Robolectric to 4.7+
### Describe the bug

Since Robolectric 4.7, there have been major updates and improvements including:
- Native support for M1 Macs
- Native support for ShadowSql for both M1 Macs and Linux.
- Improved timing and performance

We are unable to update to this on Bazel  because of a compatibility issue introduced by the required incidental update of the Espresso-Core dependency (Required minimum is 3.5.0-alpha03), likely due to the use of kotlin 1.5.

### Steps To Reproduce

1. In `third_party/versions.bzl`:
Change roboloctric to 4.7
Update espresso-core to  3.5.0-alpha03

2. In the `WORKSPACE` file"
Update the robolectric http_archive to download version 4.7.

3. Follow the Wiki [instructions](https://github.com/oppia/oppia-android/wiki/Updating-Maven-Dependencies) on updating dependencies on bazel.

Run bazel

### Expected Behavior

Project builds successfully.

### Screenshots/Videos

Here is the stacktrace:

`merged_manifests\debug\AndroidManifest.xml:23: AAPT: error: unexpected element <queries> found in <manifest>.`

_No response_

_No response_

_No response_

### Additional Context

I found similar error reports for Gradle, but none for Bazel. Assuming that the root cause is the same, updating to a newer version of Bazel should resolve this. The cause being newer versions of dependencies introduce a <query> element to manifest, but older build systems are unaware of this.

https://stackoverflow.com/a/62969918/11396524

https://stackoverflow.com/questions/62969917/how-can-i-fix-unexpected-element-queries-found-in-manifest-error

https://forum.ionicframework.com/t/how-do-i-fix-unexpected-element-queries-found-in-manifest/206645

jainv4156:
## Explanation
Fixes: #5143

## Essential Checklist

## Testing 

- data  
image

- testing
image

- utility
image

- domain 
image

- app

before 
image

image

after 
image

image


jainv4156:
@adhiamboperes, @BenHenning PTAL
Thanks!

third_party/versions.bzl
```diff
@@ -102,7 +102,7 @@ MAVEN_TEST_DEPENDENCY_VERSIONS = {
     "androidx.compose.ui:ui-test-junit4": "1.1.1",
     "androidx.test.espresso:espresso-accessibility": "3.1.0",
     "androidx.test.espresso:espresso-contrib": "3.1.0",
-    "androidx.test.espresso:espresso-core": "3.2.0",
+    "androidx.test.espresso:espresso-core": "3.5.0-alpha05",
```
adhiamboperes: Could we use the stable 3.5.0? At the time of creating the issue we were using an older version of Kotlin which was not compatible, but now it might work. Please try that and let me know.

```suggestion
    "androidx.test.espresso:espresso-core": "3.5.0,
```

adhiamboperes (changes requested):
Thanks @jainv41561. PTAL at my comment.

Additional note - There is no need to run all tests locally since the CI will run all the tests(faster than your PC) when you push to a PR.  We typically only need to run the test files affected by changes directly related to the PR,

jainv4156:
hey @adhiamboperes,
I attempted to upgrade Espresso to the latest version, but I'm encountering additional errors. Could you please review them to determine if they can be resolved on our end? If so, then we migh be able to use stable version 3.5.0

adhiamboperes:
@jainv4156, the build fails with pro guard errors, meaning there are dependency incompatibilities. Based on the maven-install.json, it’s clear that 3.5.0 introduces kotlin 1.7, while the version we support is 1.6.21. 

According to  https://mvnrepository.com/artifact/androidx.test.espresso/espresso-core, the newest version to use kotlin 1.6.21 is 3.5.0-alpha07. I believe we should use that instead.

third_party/versions.bzl
jainv4156: done

adhiamboperes (approved):
LGTM, thanks!

BenHenning (approved):
Thanks @jainv4156! It's actually nice to see that this update seems relatively straightforward without needing to worry about Gradle. :)

BenHenning:
Also FWIW the errors in the PR description for Jetifier incompatibilities are unrelated to the Robolectric & Espresso upgrades, and those upgrades won't fix the Jetifier issues. It's caused by a non-hermetic state in the build system leaking out sometimes, especially during large multi-target builds.

PR #5700, opened 2025-02-12
Fix #5699: Update wiki deployment permissions

### LINKED ISSUE #5699: [BUG]: Wiki changes can't push currently
### Describe the bug

Attempts to update the wiki are failing during wiki deployment:

```
...
Switched to a new branch 'master'
remote: Permission to oppia/oppia-android.wiki.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/oppia/oppia-android.wiki.git/': The requested URL returned error: 403
```

From https://github.com/oppia/oppia-android/actions/runs/13187896020/job/36818438965#step:3:264.

### Steps To Reproduce

Make a change to the wiki and submit the PR, or manually trigger a wiki deployment via the actions workflow.

### Expected Behavior

Deployment should succeed.

### Screenshots/Videos

_No response_

N/A

N/A

N/A

### Additional Context

This is very likely tied to a policy update made at the Oppia organization level recently where ``GITHUB_TOKEN`` was restricted to only have "read repository contents and packages" permissions. Oppia web and Android have different approaches for deployment which is why we're only seeing this for Android:
- Oppia web has an explicit process for fetching the auth token using a secret: https://github.com/oppia/oppia-web-developer-docs/blob/4b4e27ab12da2f582b87e2ffb7e70cdbb5a4a7b5/.github/workflows/deploy.yml#L15-L18 and https://github.com/oppia/get-github-app-token.
- Oppia Android relies on ``GITHUB_TOKEN`` since the workflow isn't running cross-repository: https://github.com/oppia/oppia-android/blob/a4317ba22be260c313b6c7d9a80abd87908d8441/.github/workflows/wiki.yml#L38-L53

https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token describes the permission levels for ``GITHUB_TOKEN``. I suspect all we need to do is update the required permissions for the workflow so that it has the correct permissions per https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions.

BenHenning:
## Explanation
Fixes #5699

This attempts to fix #5699 by updating the job permissions for wiki deployment to include access to content writing (which is now off by default at the Oppia organization level--see #5699). It's not clear whether this will fix the issue since GitHub's documentation is unclear as to which permissions are required in order to be able to push to the wiki, but it seems likely that 'contents' is the correct permission per https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview. Unfortunately, the fix can't be verified until after the PR is merged as the deployment job only runs manually or upon a PR merging.

Note that the permission is intentionally being set job-specific to avoid over-subscribing permissions at the workflow level (since the other job and potentially future jobs in this workflow do not need content write access).

## Essential Checklist

## For UI-specific PRs only
N/A -- this is an infrastructure-only change that also only affects wiki deployment and verification workflows.


BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @adhiamboperes.

.github/workflows/wiki.yml
```diff
@@ -1,5 +1,8 @@
 name: Deploy to Wiki
 
+permissions:
```
seanlip: @BenHenning Generally (at least on Web), we are moving towards a system where all yml files specify "permissions: {}" at the top of the workflow file, and individual jobs are assigned higher permission levels. This allows us to isolate elevated permissions to specific jobs rather than an entire workflow.

See https://github.com/oppia/oppia/blob/develop/.github/workflows/full_stack_tests.yml#L298 for an example (I know it's permissions: read-all at the top but we do want to change that to {} as well).

Should we adopt a similar approach on the Android team?

seanlip (changes requested):

.github/workflows/wiki.yml
BenHenning: I'm completely fine with making the permission at a job-level.

I don't see value in marking the workflow as permissions {}, though. Simply omitting it will set all of the jobs to the organization-level default which we've deemed is reasonable. I'm generally not keen on the human element of having to remember to include the "reset permissions" line (though I am a fan of restricting access, so I'll go ahead and make the corresponding change here).

BenHenning (commented):
Self-reviewed change.

seanlip (approved):
OK, thanks @BenHenning -- LGTM.

adhiamboperes (approved):
Thanks @BenHenning, this LGTM.

.github/workflows/wiki.yml
BenHenning: Marking as resolved since I think you're fine with this per your latest comment.

BenHenning:
Thanks @adhiamboperes & @seanlip! Enabling auto-merge after syncing. Will verify the deployment once the PR is merged and update the original issue per the results.

Issue #1683, opened 2020-08-19
android_local_test doesn't work with data-binding. 


miaboloix:
At the moment, certain necessary data binding processing is not being handled by kt_android_library and android_local_test, rules causing `DataBinderMapperImpl.java` not to be generated. The workaround for this is manually creating `DataBinderMapperImpl.java` and adding it to the source file list for each test target.

BenHenning:
It seems that properly setting up AndroidX in Bazel fixes this issue. My latest changes to add AndroidX support mitigate the need for the additional mapper file we've included in the codebase, though I'm not fully sure why.

BenHenning:
Correction: that's not actually right. I was misled by this being a runtime, not compile-time, error. Additional work in Bazel will be necessary.

BenHenning:
I think I have a fix for Bazel, but it does lead to one limitation: we can't do any databinding generation within the test. I don't think this will actually cause any problems, and it's possible to work around by changing the package of the base kt_android_library that contains the actual test code.

We will also need to update oppia_android_test to enable data binding on the native android_local_test rather than the kt_android_library (and it has to be _instead of_ since they share the same package).

BenHenning:
The needed changes have been checked into Bazel: https://github.com/bazelbuild/bazel/commit/cbdb2c74d747ed32ccb748be9c1633926e963b0f & https://github.com/bazelbuild/bazel/commit/f4430e4e7ec9e832e2b6d0e7f1c108256563bd23 (the latter is just so that we can move off of our custom copy of Bazel).

There are still a few changes need in Oppia Android, but we need to wait to make these until after a version of Bazel containing the above changes becomes available.

BenHenning:
Bazel 4.0 is out, and in conjunction with https://github.com/oppia/oppia-bazel-tools this seems to work quite well.

PR #5702, opened 2025-02-13
Fix #1617, #2745: Remove source code genrules

### LINKED ISSUE #1617: Remove genrules once Gradle has been removed

### LINKED ISSUE #2745: Databinding has visual errors in Android Studio with Bazel
When syncing the project in Android Studio using Bazel (rather than Gradle), all databinding mapping files are shown as if they don't exist despite the project working correctly. This needs investigation & potential fixes to the plugin,

BenHenning:
## Explanation

Fixes #1617
Fixes #2745

This PR removes all ``genrule``s (https://bazel.build/reference/be/general#genrule) that generate Kotlin or Java source code in the repository. This corresponds to almost every app/ directory test and production file, including the few Java files used for binding adapters. While not every file was being regenerated, the vast majority were.

This change has fairly substantial implications:
- Symbolization in Android Studio with the IntelliJ Bazel plugin should now work correctly if the corresponding targets are included in one's ``.bazelproject`` file. This didn't work before because the ``genrule``s would create a new file either with a different name or different path, so the plugin didn't know to associate the two during sync.
- All old crash stack traces in crash dashboards (i.e. Crashlytics and the Play Console) _may_ no longer match to the same crashes in versions of the app that will include this change if those stack traces include any affected app classes (since the names are now different). This isn't certain however, since the class names are the same (only the filenames are different, and it's possible the ``SourceFile`` attribute will be stripped or not included in corresponding Dex files that are deployed with the app).
- The build will be a bit faster and take up fewer resources. While the ``genrule``s themselves ran very quickly, the bigger win is the large decrease in targets that need to be analyzed during a large build.
- Imports will now appear fairly inconsistent across contexts as we now sometimes import ``R`` locally, from a common package, or from a very specific package (such as for tests). Databinding imports will face a similar issue. With symbolization in Android Studio, I'm hopeful this isn't too confusing to developers. In the long-term, we will try to remove the need for explicit imports entirely (and do away with databinding), but this sort of work will be on the order of years, not weeks or months, to complete. This will be the medium- to long-term state for development moving forward.

This change also includes a few noteworthy simplifications:
- ``app_test`` has been removed entirely in favor of using ``oppia_android_test`` directly. This is the direction we want to go in general for simpler test definitions, so it's nice to see this was actually possible with the removal of ``test_with_resources``. Using this for resource cases _is_ a bit awkward right now due to needing to include the ``DataBinderMapperImpl.java`` bridge class for databinding, but hopefully this will be fixable in the future (#1683).
- A bunch of structures were simplified in ``app/BUILD.bazel`` due to the extra resources-specific splits of different libraries being no longer necessary. The notion of migrated targets were also removed since they were actually never needed with how Bazel's ``glob`` behaves.
- Non-modularized app tests were able to be consolidated to one statement for both ``test/`` and ``sharedTest/`` tests (rather than having two), which is cleaner.

There was one wiki change needed since ``app_test`` is no longer available to use, and it has been clarified to better explain how to set up a test for resources vs. not.

For compatibility, this further cements Gradle no longer being able to be used for builds, but that's no longer a concern since we can't build with Gradle, anyway.

It's important to note that all of the updated import statements are reflecting the existing reality of the Bazel build configuration (i.e. these are exactly the import statements that Bazel was using via the ``genrule`` targets). Android resource processing is very complex, and even more so in the Bazel world. This means a lot of the ``R`` symbolic references are actually duplicated many times in the build pipeline, so there are other ``R`` files that can be used. This was an approach I took a few months ago to try and remove the ``genrule``s while maintaining Gradle compatibility. I didn't take that approach here for simplicity (it was more important to get this change in), but it's quite likely the import statements will see more iterations as we continue with modularization work. One useful goal may be to try and consolidate to local package ``R`` files to avoid needing an import at all (and thus remove potential confusion that there _are_ many ``R`` files generated throughout the build graph), but there are other possibilities that can be explored as well.

#5747 is also tracking some future clean-up work made possible after this change in order to simplify `R` imports throughout the codebase.

## Essential Checklist

## For UI-specific PRs only
N/A -- It's expected that this infrastructure-only change has no real impact on UIs, and this is assumed to be the case as per the passing tests.


BenHenning:
Going to skip self-reviewing on this PR since it's quite a large number of files, and the way I did this change relied heavily on automation and verifying builds (so I'm rather confident in its correctness).

PTAL @adhiamboperes and feel free to delegate as needed since I know this is a large PR.

app/src/sharedTest/java/org/oppia/android/app/databinding/BUILD.bazel
```diff
@@ -2,48 +2,18 @@
...
-    outs = ["TextViewBindingAdaptersTest_updated.kt"],
-    cmd = """
-    cat $(SRCS) |
-    sed 's/import org.oppia.android.R/import org.oppia.android.app.databinding.R/g' |
-    sed 's/import org.oppia.android.app.databinding.TextViewBindingAdapters./import org.oppia.android.app.databinding.TextViewBindingAdapters_updated./g' > $(OUTS)
-    """,
-)
-
-app_test(
+oppia_android_test(
```
adhiamboperes, line 7: Noticed in build warnings that `TextInputLayoutBindingAdaptersTest` is missing a test target here. It seems I forgot to create one when I added the test suite in a previous PR.

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt
```diff
@@ -40,6 +39,7 @@ import org.oppia.android.app.devoptions.DeveloperOptionsModule
 import org.oppia.android.app.devoptions.DeveloperOptionsStarterModule
 import org.oppia.android.app.player.state.itemviewmodel.SplitScreenInteractionModule
 import org.oppia.android.app.shim.ViewBindingShimModule
+import org.oppia.android.app.test.R
```
adhiamboperes, line 42: Import does not appear to be used (greyed out in IDE, and test ran fine when removed). It seems that the test is using  local package R file instead.
Screenshot 2025-02-14 at 21 00 43

adhiamboperes: Ditto all tests in this package.

adhiamboperes (changes requested):
I have reviewed about 50% of these files, and mostly look good. I ran the code locally and my observation was:
1. It appears that the Bazel plugin does not handle conflict resolution well, e.g.
 - when there are 2 versions of the same dependency, such as version 2.20 and 2.3.1, then it shows an error wherever classes of that dependency are used.
 - There is a possible conflic when multiple R classes are available, and I wonder if it has something to do with how `DataBinderMapperImpl` handles generated bindings, or perharps how the plugin interprets the bindings.
 ```kotlin
  public List<DataBinderMapper> collectDependencies() {
    ArrayList<DataBinderMapper> result = new ArrayList<DataBinderMapper>(5);
    result.add(new org.oppia.android.app.databinding.DataBinderMapperImpl());
    result.add(new org.oppia.android.app.databinding.adapters.DataBinderMapperImpl());
    result.add(new org.oppia.android.app.ui.DataBinderMapperImpl());
    result.add(new org.oppia.android.app.view.models.DataBinderMapperImpl());
    result.add(new org.oppia.android.app.views.DataBinderMapperImpl());
    return result;
  }
```
It seems that the generated R classes are prioritized based on the order of packages in the resulting list, such that the plugin can only resolve `import org.oppia.android.app.databinding.R` but not `import org.oppia.android.app.ui.R`, `import org.oppia.android.app.view.models.R` or `import org.oppia.android.app.views.R`.

2. It is likely that `import org.oppia.android.app.ui.R` can safely be replaced by `org.oppia.android.app.databinding.R`

PTAL, thanks @BenHenning!

**Comment by BenHenning** [comment]
> I have reviewed about 50% of these files, and mostly look good. I ran the code locally and my observation was:
> 
> 1. It appears that the Bazel plugin does not handle conflict resolution well, e.g.
> 
> * when there are 2 versions of the same dependency, such as version 2.20 and 2.3.1, then it shows an error wherever classes of that dependency are used.
> * There is a possible conflic when multiple R classes are available, and I wonder if it has something to do with how `DataBinderMapperImpl` handles generated bindings, or perharps how the plugin interprets the bindings.
> 
> ```kotlin
>  public List<DataBinderMapper> collectDependencies() {
>    ArrayList<DataBinderMapper> result = new ArrayList<DataBinderMapper>(5);
>    result.add(new org.oppia.android.app.databinding.DataBinderMapperImpl());
>    result.add(new org.oppia.android.app.databinding.adapters.DataBinderMapperImpl());
>    result.add(new org.oppia.android.app.ui.DataBinderMapperImpl());
>    result.add(new org.oppia.android.app.view.models.DataBinderMapperImpl());
>    result.add(new org.oppia.android.app.views.DataBinderMapperImpl());
>    return result;
>  }
> ```
> 
> It seems that the generated R classes are prioritized based on the order of packages in the resulting list, such that the plugin can only resolve `import org.oppia.android.app.databinding.R` but not `import org.oppia.android.app.ui.R`, `import org.oppia.android.app.view.models.R` or `import org.oppia.android.app.views.R`.
> 
> 2. It is likely that `import org.oppia.android.app.ui.R` can safely be replaced by `org.oppia.android.app.databinding.R`
> 
> PTAL, thanks @BenHenning!

Thanks @adhiamboperes! Usually, we should never have multiple versions (`rules_jvm_external` makes a strong effort to ensure this), but you're correct that there are cases with both databinding and Android resources code generation that could result in duplicated symbols.

I'm also not sure if the error you're describing is due to duplication--are you seeing an actual sync failure or broken symbols in Oppia Android code? Dependency management is much more complex in the Bazel world than Gradle since there are thousands of targets to build rather than ~6, so it's possible for multiple R files to exist with the same symbols but not be available for a given source file (if the corresponding `kt_android_library` is not in the `deps` for that target).

Separately, do you want me to make changes before you finish reviewing the PR? I'm assuming it will be much easier to review incremental changes after a full review since then your marked review files will correctly diff against later changes.

File: app/src/sharedTest/java/org/oppia/android/app/databinding/BUILD.bazel (hunk 1 above)
Comment (BenHenning) [reply app/src/sharedTest/java/org/oppia/android/app/databinding/BUILD.bazel:7]: Good catch--I'll add this on the next pass for changes. I *think* I might've caught this in my modularization PR chain, though I'm actually surprised this wasn't caught by `TestFileCheck`. I'm assuming it might be because that script is only verifying against production Kotlin files and not Java. :) I suppose I could fix the check in this PR, but ideally we move away from Java long-term, anyway, so it may not actually be worth the effort.

I think I had some success migrating these Java files to Kotlin in the past, as well...

File: app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt (hunk 2 above)
Comment (BenHenning) [reply app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt:42]: This is actually really surprising to me. I can try to remove the imports and re-run the tests. It may be the case that a LOT of code can technically rely on the locally produced R files (it's a really significant problem with `android_library` in my opinion as it adds a ton of unnecessary build complexity and time, but we don't have an obvious mitigation for now).

This is a great catch. My goal for the PR was reaching code parity with the actual build today, but I welcome simplifications like this.

**Comment by adhiamboperes** [comment]
This PR also fixes https://github.com/oppia/oppia-android/issues/2745 or https://github.com/oppia/oppia-android/issues/2745 can be closed as duplicate.

File: app/src/main/java/org/oppia/android/app/utility/math/MathExpressionAccessibilityUtil.kt [hunk 3]
```diff
@@ -1,6 +1,6 @@
 package org.oppia.android.app.utility.math
 
-import org.oppia.android.R
+import org.oppia.android.app.R
```
adhiamboperes, line 3: This R import appears to be from app context so it does not conform to other patterns. Ditto the SpotlightFragment one.

adhiamboperes (approved):
Thanks @BenHenning!

As far as removing genrules and updating the resulting imports, everything looks correct. 

I believe my comments regarding imports appearing fairly inconsistent across contexts have been addressed in the PR description, and should not block this PR.

adhiamboperes:
> Thanks @adhiamboperes! Usually, we should never have multiple versions (`rules_jvm_external` makes a strong effort to ensure this), but you're correct that there are cases with both databinding and Android resources code generation that could result in duplicated symbols.
> 
> I'm also not sure if the error you're describing is due to duplication--are you seeing an actual sync failure or broken symbols in Oppia Android code? Dependency management is much more complex in the Bazel world than Gradle since there are thousands of targets to build rather than ~6, so it's possible for multiple R files to exist with the same symbols but not be available for a given source file (if the corresponding `kt_android_library` is not in the `deps` for that target).

The error is a syntax higlighting error, while sync finishes, the highlighting is not fixed. Bazel build also works fine.


> Separately, do you want me to make changes before you finish reviewing the PR? I'm assuming it will be much easier to review incremental changes after a full review since then your marked review files will correctly diff against later changes.

I have reviewed the PR in full as of now.

app/src/main/java/org/oppia/android/app/utility/math/MathExpressionAccessibilityUtil.kt
BenHenning: What do you suggest should be used, instead? At the moment, I think we (mostly) need to use the generated R class corresponding to the package that the source file is in (e.g. how the view models are set up).

BenHenning:
@adhiamboperes PTAL at my reply, and please resolve the earlier comments if they've been addressed.

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt
adhiamboperes: I found out from reviewing therest of the PR that this is because the databinding package has its own R used widely elsewhere in the project. So even though these are tests, the package R takes priority.

app/src/main/java/org/oppia/android/app/utility/math/MathExpressionAccessibilityUtil.kt
adhiamboperes: In general, I think it is best to use package local R.

adhiamboperes:
> @adhiamboperes PTAL at my reply, and please resolve the earlier comments if they've been addressed.


Done, please feel free to resolve the 2 open comments once addressed.

app/src/main/java/org/oppia/android/app/utility/math/MathExpressionAccessibilityUtil.kt
BenHenning: Hmm, I'm actually not sure if that'll work, and it can possibly introduce runtime issues. Are you okay with me filing an issue to try and remove all R imports and do that in a follow up PR in case it needs to be reverted? The current code here has almost exact parity with what the generated code from the genrules produced before.

adhiamboperes: Sure, that sounds good.

BenHenning: Thanks! Filed #5747 accordingly.

Also, marking this as resolved since the action was just to file an issue, and that issue is now filed.

app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerPaddingBindingAdaptersTest.kt
BenHenning: Got it--it seems that there's no more action to take here, then, so resolving this.

PR #5715, opened 2025-02-24
Fixes #5556: Feature flags failing to log correctly

### LINKED ISSUE #5556: [BUG]: Feature flags are failing to log correctly
### Describe the bug

Analyzing the 'feature_flag_list' event in production is showing that feature_flag_names and feature_flag_sync_statuses are both appearing as 'null'. This essentially renders the event unusable since there's no way to correspond event enabled states with the corresponding event.

### Steps To Reproduce

Run the following query on the developer Firebase project:

```sql
SELECT
(SELECT value.string_value FROM UNNEST(event_params) WHERE key = "uuid") AS uuid,
(SELECT value.string_value FROM UNNEST(event_params) WHERE key = "app_session_id") AS app_session_id,
(SELECT value.string_value FROM UNNEST(event_params) WHERE key = "feature_flag_names") AS feature_flag_names,
(SELECT value.string_value FROM UNNEST(event_params) WHERE key = "feature_flag_enabled_states") AS feature_flag_enabled_states,
(SELECT value.string_value FROM UNNEST(event_params) WHERE key = "feature_flag_sync_statuses") AS feature_flag_sync_statuses,
FROM `oppia-android-dev.analytics_235566504.events_*`
WHERE event_name = "feature_flag_list"
```

See sample results:

```json
[{
  "uuid": null,
  "app_session_id": "277ed911-efef-3144-b0fa-d37cdd87ac86",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,false",
  "feature_flag_sync_statuses": null
}, {
  "uuid": null,
  "app_session_id": "f2ed5da7-bd3d-3ec3-9044-13ddf766d8fe",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,true",
  "feature_flag_sync_statuses": null
}, {
  "uuid": null,
  "app_session_id": "a41e3a56-4c4a-3a3b-b820-a6337dd5797c",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,true",
  "feature_flag_sync_statuses": null
}]
```

### Expected Behavior

Flag names and sync statuses should successfully log.

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

_No response_

kkmurerwa:
## Explanation
Fixes #5556: Feature flags failing to log correctly

### Cause of the bug
The primary cause of the bug was [Google Analytics character limits](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.html#logEvent(java.lang.String,android.os.Bundle)) on log event parameter values. Google Analytics (GA), particularly GA 4, imposes the following limits on log events and their parameters;

- **25 Parameters** - Each app event cannot exceed 25 individual parameters. This number is inclusive of default GA parameters making the limit slightly lower than that number.
- **40 Characters** - An event parameter name cannot exceed 40 characters.
- **100 Characters** - Each event parameter can have a value not exceeding 100 characters. If a key or name exceeds the limit, the event is dropped and a `firebase_error` parameter is added to the event and a `firebase_error_value` parameter is added to indicate the name of the parameter with the invalid value. In the case of exceeding the character limit, the error code is usually `4` as can be seen from the list of [GA4 errors](https://firebase.google.com/docs/analytics/errors).

**Note**: These limits only apply for free tier users. Google Analytics 360 users are allowed up to `100 parameters`  per event and `500 characters` for event parameter values.

### Fix implemented
To fix this issue, three changes were made to the `EventBundleCreator`;

- **Feature flag names** - A new converter, the `FeatureFlagNameToIntegerNameConverter` was introduced. This converter will keep a list of flag names supported by the app and their integer representations. Each feature flag name should have a unique integer representation and no two feature flag names should have the same integer representation, even when one of them is no longer supported.
- **Feature flag sync status** - Each feature flag sync status is parsed into an integer representation based on the number it has in the `platform_parameters.proto` that defines the various sync statuses as an enum. The following are the current sync status and their integer representations: `SYNC_STATUS_UNSPECIFIED` as `0`, `NOT_SYNCED_FROM_SERVER` as `1`, and `SYNCED_FROM_SERVER` as `2`.
- **Feature flag enabled state** - Since enabled states are either `true` or `false`, `true` is converted to `1` and `false` to `0`.

By making these changes, it is possible to theoretically log **36 feature flags** in one event. The breakdown for this is as shown;

**Feature flags with 1 digit** - 1 digit each and a comma seperator = _**9 maximum feature flags taking 18 characters**_
**Feature flags with 2 digits** - 2 digits each and a comma seperator = _**27 maximum feature flags taking 81 characters**_
**Total loggable feature flags** = _**9 + 27 = 36 maximum feature flags taking 99/100 characters**_

### Logs comparison
Previously, the feature flags list used to appear like this;

```json
{
    "uuid": null,
    "app_session_id": "277ed911-efef-3144-b0fa-d37cdd87ac86",
    "feature_flag_names": null,
    "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,false",
    "feature_flag_sync_statuses": null
}
```

With the new implementation, the feature flags list appears like this;
```json
{
    "uuid": null,
    "app_session_id": "277ed911-efef-3144-b0fa-d37cdd87ac86",
    "feature_flag_sync_statuses": "1,1,1,1,1,1,1,1,1,1,1,1,1",
    "feature_flag_enabled_states": "0,0,0,0,0,0,0,0,0,0,1,0,1",
    "feature_flag_names": "1,2,3,4,5,6,7,8,9,10,11,12,13"
}
```

## Essential Checklist


utility/src/main/java/org/oppia/android/util/logging/FeatureFlagSyncStatusToIntegerConverter.kt
```diff
@@ -0,0 +1,24 @@
...
+   * @param syncStatus The flag name to convert.
+   * @return An integer representation of the event name.
+   */
+  fun convertToInteger(syncStatus: SyncStatus): Int {
+    return when (syncStatus.toString()) {
+      "SYNC_STATUS_UNSPECIFIED" -> 1
+      "NOT_SYNCED_FROM_SERVER" -> 2
+      "SYNCED_FROM_SERVER" -> 3
+      else -> -1
+    }
```
BenHenning: This should just `when` on the enum directly since then no conversion happens (and it can be exhaustive so `else` can be omitted).

BenHenning: Also, could we maybe just use the SyncStatus's number or index field that proto generates? That is a stable integer representation of the enum.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+      LOGGING_LEARNER_STUDY_IDS -> 5
+      EDIT_ACCOUNTS_OPTIONS_UI -> 6
+      ENABLE_PERFORMANCE_METRICS_COLLECTION -> 7
+      SPOTLIGHT_UI -> 8
+      INTERACTION_CONFIG_CHANGE_STATE_RETENTION -> 9
+      APP_AND_OS_DEPRECATION -> 10
+      ENABLE_NPS_SURVEY -> 11
+      ENABLE_ONBOARDING_FLOW_V2 -> 12
+      ENABLE_MULTIPLE_CLASSROOMS -> 13
+      else -> -1
```
BenHenning: Let's use zero to avoid negatives for the error case.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
```diff
@@ -688,12 +688,19 @@ class EventBundleCreator @Inject constructor(
...
+
+        val featureFlagNames = featureFlagsList.map {
+          featureFlagNameConverter.convertToInteger(it.flagName)
+        }
+
+        val featureFlagSyncStatuses = featureFlagsList.map {
+          featureFlagSyncStatusConverter.convertToInteger(it.flagSyncStatus)
+        }
+
         val featureFlagEnabledStates = featureFlagsList.map { it.flagEnabledState }
```
BenHenning: As discussed, this should be replaced with 1s and 0s for trues and falses.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.oppia.android.util.platformparameter.INTERACTION_CONFIG_CHANGE_STATE_RETENTION
+import org.oppia.android.util.platformparameter.LEARNER_STUDY_ANALYTICS
+import org.oppia.android.util.platformparameter.LOGGING_LEARNER_STUDY_IDS
+import org.oppia.android.util.platformparameter.SPOTLIGHT_UI
+import javax.inject.Inject
+
+/**
+ *
+ */
+class FeatureFlagNameToIntegerConverter @Inject constructor() {
```
BenHenning: Doesn't need to be injectable and can just be an object to avoid instantiation.

BenHenning: Also, let's use 'NumericId' instead of 'Integer' here and below.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagSyncStatusToIntegerConverter.kt
kkmurerwa: NTS: Use `syncStatus.number` to get the index

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
kkmurerwa: NTS: Use `syncStatus.compareTo(false)`

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagSyncStatusToIntegerConverter.kt
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
kkmurerwa: Done

kkmurerwa:
@adhiamboperes this is ready for another pass. PTAL

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.oppia.android.util.platformparameter.FAST_LANGUAGE_SWITCHING_IN_LESSON
+import org.oppia.android.util.platformparameter.INTERACTION_CONFIG_CHANGE_STATE_RETENTION
+import org.oppia.android.util.platformparameter.LEARNER_STUDY_ANALYTICS
+import org.oppia.android.util.platformparameter.LOGGING_LEARNER_STUDY_IDS
+import org.oppia.android.util.platformparameter.SPOTLIGHT_UI
+import javax.inject.Inject
+
+/**
+ *
+ */
```
adhiamboperes: Incomplete KDoc.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+class FeatureFlagNameToIntegerNameConverter @Inject constructor() {
+  /**
+   * Converts a feature flag name to an integer.
+   *
+   * @param flagName The flag name to convert.
+   * @return An integer representation of the event name.
```
adhiamboperes:
```suggestion
    * @param flagName the string constant flag name to convert.
   * @return an integer representation of the flag name.
```

adhiamboperes: Per last weeks discussion, we don't want to re-use the integers. We should make a note of that in the KDoc.

utility/src/test/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverterTest.kt
```diff
@@ -0,0 +1,154 @@
...
+    "flagName=$ENABLE_ONBOARDING_FLOW_V2",
+    "expectedValue=12"
+  )
+  @Iteration(
+    "enable_multiple_classrooms",
+    "flagName=$ENABLE_MULTIPLE_CLASSROOMS",
+    "expectedValue=13"
+  )
+  fun testConvertToIntegerName_returnsCorrectIntegerForEach() {
+    val nameInteger = converter.convertToInteger(flagName)
```
adhiamboperes: Maybe?
```suggestion
    val integerName = converter.convertToInteger(flagName)
```

adhiamboperes (changes requested):
Thanks @kkmurerwa! Implementation looks good, but the PR description could benefit from a thorough write up since we havent documented the character length issues elsewhere (AFAIK) and why we chose this solution. The description could also benefit from sample logs as copied over from println debug and Firebase logs.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: Done

utility/src/test/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverterTest.kt
kkmurerwa: I have implemented this suggestion.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: I have added text here. I have also updated the Wiki so that contributors can know to add new feature flags to the flag name to integer converter.

kkmurerwa:
@adhiamboperes I have updated the PR description and addressed the other comments. PTAL

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -16,14 +16,19 @@ import org.oppia.android.util.platformparameter.SPOTLIGHT_UI
 import javax.inject.Inject
 
 /**
- *
+ * Converter for feature flag names into integers for more compact logging to address Google
```
adhiamboperes:
```suggestion
 * Converter for feature flag names to integers for more compact logging to address Google
```

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -16,14 +16,19 @@ import org.oppia.android.util.platformparameter.SPOTLIGHT_UI
...
+ * Analytics character limits. GA4 limits the characters permitted in a log event parameter value to
+ * a maximum of 100 characters as of March 2025.
```
adhiamboperes: I believe that when we include information like this in KDocs, the standard pattern is to add the url(e.g. https://github.com/oppia/oppia-android/blob/9404ea41d91d0a2a62ca8d42de29d60d234fd02c/app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt#L45-L47)

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -16,14 +16,19 @@ import org.oppia.android.util.platformparameter.SPOTLIGHT_UI
...
  */
 class FeatureFlagNameToIntegerNameConverter @Inject constructor() {
   /**
    * Converts a feature flag name to an integer.
    *
-   * @param flagName The flag name to convert.
-   * @return An integer representation of the event name.
+   * These numbers are supposed to be increment indefinitely and the integer representation of
```
adhiamboperes:
```suggestion
   * These numbers are supposed to be incremented indefinitely and the integer representation of
```

adhiamboperes (changes requested):
Thanks @kkmurerwa! I left some further suggestions for KDocs, but minor. Also, it would be great to have a sample log of the current results e.g:

```json
[{
  "uuid": null,
  "app_session_id": "277ed911-efef-3144-b0fa-d37cdd87ac86",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,false",
  "feature_flag_sync_statuses": null
}, {
  "uuid": null,
  "app_session_id": "f2ed5da7-bd3d-3ec3-9044-13ddf766d8fe",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,true",
  "feature_flag_sync_statuses": null
}, {
  "uuid": null,
  "app_session_id": "a41e3a56-4c4a-3a3b-b820-a6337dd5797c",
  "feature_flag_names": null,
  "feature_flag_enabled_states": "false,false,false,false,false,false,false,false,false,false,true,false,true",
  "feature_flag_sync_statuses": null
}]
``` 

Since as reviewers we don't have other means of verifying the wxpected output.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: I wanted to do this but the link is too long and exceeds the 100 character limit.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: Done

kkmurerwa:
@adhiamboperes I have updated the PR description to include the logs. I have also added the changes suggested with the exception of the link which is too long and exceeds the character limit. PTAL

adhiamboperes (approved):
Thanks @kkmurerwa! I have no further comments. Assigning @BenHenning for verification and resolving open comments.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
```diff
@@ -688,12 +688,15 @@ class EventBundleCreator @Inject constructor(
...
-        val featureFlagSyncStatuses = featureFlagsList.map { it.flagSyncStatus }
-        val featureFlagEnabledStates = featureFlagsList.map { it.flagEnabledState }
+        val featureFlagNameConverter = FeatureFlagNameToIntegerNameConverter()
+
+        val featureFlagNames = featureFlagsList.map {
+          featureFlagNameConverter.convertToInteger(it.flagName)
+        }
+
+        val featureFlagSyncStatuses = featureFlagsList.map { it.flagSyncStatus.number }
+        val featureFlagEnabledStates = featureFlagsList.map { it.flagEnabledState.compareTo(false) }
```
BenHenning: `compareTo` is an interesting approach, though it's possibly going to be hard for people to understand why it works. It also supports a theoretical tri-state: less than (negative), equal (0), greater than (positive). I suggest just doing a basic integer conversion, instead:

```suggestion
        val featureFlagEnabledStates = featureFlagsList.map { if (it) 1 else 0 }
```

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
BenHenning: You should be able to include it without triggering a lint failure. It's better to include the link for context even if it goes over 100 characters.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
```diff
@@ -688,12 +688,15 @@ class EventBundleCreator @Inject constructor(
       value: FeatureFlagListEventContext
     ) : EventActivityContext<FeatureFlagListEventContext>(activityName, value) {
       override fun EventLog.FeatureFlagListContext.storeValue(store: PropertyStore) {
-        val featureFlagNames = featureFlagsList.map { it.flagName }
-        val featureFlagSyncStatuses = featureFlagsList.map { it.flagSyncStatus }
-        val featureFlagEnabledStates = featureFlagsList.map { it.flagEnabledState }
+        val featureFlagNameConverter = FeatureFlagNameToIntegerNameConverter()
```
BenHenning: Either inject the converter, or convert it to an object (not a class). In neither case should we be instantiating it here.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+class FeatureFlagNameToIntegerNameConverter @Inject constructor() {
+  /**
+   * Converts a feature flag name to an integer.
+   *
+   * These numbers are supposed to be incremented indefinitely and the integer representation of
+   * each feature flag name should not be reused even after the feature flag is no longer in use.
+   *
+   * @param flagName the string constant flag name to convert.
+   * @return an integer representation of the flag name.
```
BenHenning:
```suggestion
   * @param flagName the string constant flag name to convert
   * @return an integer representation of the flag name
```

Nit: we don't end at-clauses with punctuation per style guide.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -0,0 +1,51 @@
...
+      FAST_LANGUAGE_SWITCHING_IN_LESSON -> 4
+      LOGGING_LEARNER_STUDY_IDS -> 5
+      EDIT_ACCOUNTS_OPTIONS_UI -> 6
+      ENABLE_PERFORMANCE_METRICS_COLLECTION -> 7
+      SPOTLIGHT_UI -> 8
+      INTERACTION_CONFIG_CHANGE_STATE_RETENTION -> 9
+      APP_AND_OS_DEPRECATION -> 10
+      ENABLE_NPS_SURVEY -> 11
+      ENABLE_ONBOARDING_FLOW_V2 -> 12
+      ENABLE_MULTIPLE_CLASSROOMS -> 13
```
BenHenning: Can you please update these to number each feature based on when its flag was first created (i.e. they should be date sorted)? This will make it seem as though these were always used (as if we had used the system since the first feature flag was added), and ensures there's always a correspondence between the integer value and the age of the feature (i.e. that a larger number always indicates a newer feature).

I'll be using these for the new feature flag indexes in #5725 (which will allow us to remove this class actually since then we can just use the proto enum indexes).

wiki/Platform-Parameters-&-Feature-Flags.md
```diff
@@ -217,6 +217,28 @@ Besides the feature-flag logger, the `FeatureFlagLoggerTest` located at `domain/
...
+  DOWNLOADS_SUPPORT -> 1
+  ...
+  ENABLE_MULTIPLE_CLASSROOMS -> n // Current maximum integer
+  YOUR_NEW_FEATURE_FLAG_NAME -> n + 1  // Your new flag name and integer representation
+  else -> 0
+  }
+}
+```
+
+**DO NOT** reuse any of the digits that have already been used. The numbers are supposed to increment to infinity.
```
BenHenning:
```suggestion
**DO NOT** reuse any of the digits that have already been used.
```

Not sure the second bit adds anything. We could alternatively say "The upper bound is unexpected to be hit." or something, but I'm not sure that's really needed either. Am I maybe missing something?

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
BenHenning: Why are these changes not leading to failures in `EventBundleCreator`? It seems that perhaps we are missing coverage--can you please make sure these changes are properly tested?

BenHenning (changes requested):
Thanks @kkmurerwa! I was thinking that we should also add a top-level converter test to verify the count of flags, but I'm not too bothered since #5725 will actually largely simplify this PR's approach since proto enums will provide the exact same monotonic property without needing to actually test it in the same way.

Separately, can you add an `EventBundleCreator` test that fails if any of the feature flag parameter values exceed 100 characters? And, please verify that this fails without your fixes in place. Not only does it prove the fix, it also provides a gate for when we inevitably hit this in the future when we have more than ~30 active features at once (or ~25 if we hit 3 digit numbers).

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: I have included it.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
kkmurerwa: I have changed the approach to this.

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
kkmurerwa: I converted it to an object

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: I tracked the order in which they were created and documented it in this [Google Sheet](https://docs.google.com/spreadsheets/d/1MbW-O3toEUQu1LUXozDkYZCfBi8kDn2jN86nUwTWWLA/edit?usp=sharing). I have updated this block of code based on that list. Let me know if I should include feature flags that have already been removed.

wiki/Platform-Parameters-&-Feature-Flags.md
kkmurerwa: Done

utility/src/main/java/org/oppia/android/util/logging/EventBundleCreator.kt
kkmurerwa: I have added some tests to the EventBundleCreatorTest. I chose to go with a list of feature flag item contexts instead of parameterized tests because it allows us to better test the logs in the exact format they will be in.

kkmurerwa:
@BenHenning I have addressed the comments. PTAL

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
```diff
@@ -848,6 +852,27 @@ class EventBundleCreatorTest {
     assertThat(bundle).string("topic_id").isEqualTo(TEST_TOPIC_ID)
   }
 
+  @Test
+  fun testFillEventBundle_featureFlagsContext() {
```
BenHenning:
```suggestion
  fun testFillEventBundle_fillsFeatureFlagsContext() {
```

The expected outcome should be documented as an action.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
```diff
@@ -848,6 +852,27 @@ class EventBundleCreatorTest {
...
+    val typeName = eventBundleCreator.fillEventBundle(eventLog, bundle)
+    assertThat(typeName).isEqualTo("feature_flag_list")
+    assertThat(bundle).hasSize(14)
+    assertThat(bundle).longInt("timestamp").isEqualTo(TEST_TIMESTAMP_1)
+    assertThat(bundle).string("priority").isEqualTo("essential")
+    assertThat(bundle).integer("event_type").isEqualTo(FEATURE_FLAG_LIST_CONTEXT.number)
+    assertThat(bundle).integer("android_sdk").isEqualTo(TEST_ANDROID_SDK_VERSION)
+    assertThat(bundle).string("app_version_name").isEqualTo(TEST_APP_VERSION_NAME)
+    assertThat(bundle).integer("app_version_code").isEqualTo(TEST_APP_VERSION_CODE)
+    assertThat(bundle).string("feature_flag_names").isEqualTo("0,0,0,7")
```
BenHenning: This feels a bit fake--we should never realistically logging the same flag multiple times, right? Can you please make the test match a more realistic case (i.e. different flags)?

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
```diff
@@ -13,35 +13,36 @@ import org.oppia.android.util.platformparameter.INTERACTION_CONFIG_CHANGE_STATE_
...
+      SPOTLIGHT_UI -> 4
+      EXTRA_TOPIC_TABS_UI -> 5
+      INTERACTION_CONFIG_CHANGE_STATE_RETENTION -> 6
+      DOWNLOADS_SUPPORT -> 7
+      APP_AND_OS_DEPRECATION -> 8
+      FAST_LANGUAGE_SWITCHING_IN_LESSON -> 9
+      LOGGING_LEARNER_STUDY_IDS -> 10
       ENABLE_NPS_SURVEY -> 11
       ENABLE_ONBOARDING_FLOW_V2 -> 12
       ENABLE_MULTIPLE_CLASSROOMS -> 13
```
BenHenning: This order still doesn't seem right to me, so I did an analysis just to double check and to also account for platform parameters (for my PR).

- 19 Jul 2024: remove ENABLE_LANGUAGE_SELECTION_UI_DEFAULT_VALUE
- 28 May 2024: add multiple classrooms (feature)
- 06 Feb 2024: add onboarding v2 (feature)
- 25 Jan 2024: add nps (feature)
- 16 Nov 2023: remove continue button animation
- 22 Jun 2023: add npcsurveygraceperiod (parameter), npssurveyminimaggregatelearningtimeinatopic (parameter), 
- 05 Jun 2023: add fast language switching (feature), logging learner study ids (feature)
- 26 May 2023: add os deprecation (feature), app update version code (parameter), force update version code (parameter), lowest api level (parameter)
- 19 Nov 2022: add continue button animation (platform parameter, but probably actually a feature)
- 10 Nov 2022: rename AutomaticUpdateTopicSetting to EnableDownloadsSupport
- 18 Oct 2022: add EnableInteractionConfigChangeStateRetention (feature)
- 03 Oct 2022: add AutomaticUpdateTopicSetting (feature)
- 27 Sep 2022: add extra topics UI (feature)
- 24 Sep 2022: add spotlights (feature)
- 18 Jul 2022: add enable edit account options (feature)
- 08 Jun 2022: enable performance metrics collection (feature), perfmetcollectionuploadtime (parameter), perfmetricscollectionhighfreq (parameter), perfmetricscollectionlowfreq (parameter)
- 26 Mar 2022: add cachelatexrendering (parameter)
- 24 Jan 2022: add learner study analytics (feature)
- 06 Oct 2021: add enable language selection UI (feature)
- 20 Aug 2021: add sync up worker time period (parameter)
- 08 Jul 2021: add splash screen welcome message (parameter)

Complete list of parameters (from oldest to newest):
- `SplashScreenWelcomeMsg`
- `SyncUpWorkerTimePeriodHours`
- `CacheLatexRendering`
- `PerformanceMetricsCollectionUploadTimeIntervalInMinutes`
- `PerformanceMetricsCollectionHighFrequencyTimeIntervalInMinutes`
- `PerformanceMetricsCollectionLowFrequencyTimeIntervalInMinutes`
- `OptionalAppUpdateVersionCode`
- `ForcedAppUpdateVersionCode`
- `LowestSupportedApiLevel`
- `NpsSurveyGracePeriodInDays`
- `NpsSurveyMinimumAggregateLearningTimeInATopicInMinutes`

Complete list of features (from oldest to newest, deleted ones indicated):
- REMOVED: `EnableLanguageSelectionUi`: expected numeric value of 1
- `EnableLearnerStudyAnalytics`: expected numeric value of 2
- `EnablePerformanceMetricsCollection`: expected numeric value of 3
- `EnableEditAccountsOptionsUi`: expected numeric value of 4
- `EnableSpotlightUi`: expected numeric value of 5
- `EnableExtraTopicTabsUi`: expected numeric value of 6
- `EnableDownloadsSupport` (previously: `AutomaticUpdateTopicSetting`): expected numeric value of 7
- `EnableInteractionConfigChangeStateRetention`: expected numeric value of 8
- REMOVED: `EnableContinueButtonAnimation`: expected numeric value of 9
- `EnableAppAndOsDeprecation`: expected numeric value of 10
- `EnableFastLanguageSwitchingInLesson`: expected numeric value of 11
- `EnableLoggingLearnerStudyIds`: expected numeric value of 12
- `EnableNpsSurvey`: expected numeric value of 13
- `EnableOnboardingFlowV2`: expected numeric value of 14
- `EnableMultipleClassrooms`: expected numeric value of 15

Looks like it was close; downloads and interaction config change were swapped (probably due to downloads having a different name originally), and 2 removed flags weren't accounted for (which they need to be because we'll be reserving those old values so that they specifically don't get reused or, rather, ever used).

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
BenHenning: We should account for the removed ones, and I did an analysis as well (before seeing this link, though I also need to know platform parameters for my PR). Left a new comment accordingly.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
BenHenning: Marking this as resolved since I've verified it (feel free to reopen this if you'd like @adhiamboperes).

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
BenHenning: The 'NumericId' bit is still unresolved--please fix.

BenHenning (changes requested):
Thanks @kkmurerwa. Left a few new comments, though almost everything is resolved.

However, I think you didn't fully address this part of my previous review's top-level comment:

> Separately, can you add an `EventBundleCreator` test that fails if any of the feature flag parameter values exceed 100 characters? And, please verify that this fails without your fixes in place. Not only does it prove the fix, it also provides a gate for when we inevitably hit this in the future when we have more than ~30 active features at once (or ~25 if we hit 3 digit numbers).

Please make sure to.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerConverter.kt
kkmurerwa: This should be resolved now.

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
kkmurerwa: Done

utility/src/test/java/org/oppia/android/util/logging/EventBundleCreatorTest.kt
kkmurerwa: I have updated this to include actual feature flag names as they appear. Let me know if this is better.

utility/src/main/java/org/oppia/android/util/logging/FeatureFlagNameToIntegerNameConverter.kt
kkmurerwa: I have swapped the `EnableInteractionConfigChangeStateRetention` and `EnableDownloadsSupport` flag names to  the new numeric IDs and renumbered everything else to reflect the suggestion.

adhiamboperes:
Updated branch to retrigger checks - did not find any stacktraces from the stuck run.

adhiamboperes:
@kkmurerwa, aside from the failing code coverage failures, You have failing tests: https://github.com/oppia/oppia-android/actions/runs/14720814671/job/41314244218?pr=5715#step:12:328

kkmurerwa:
@BenHenning this is now ready for another pass. PTAL

BenHenning (approved):
Thanks @kkmurerwa. Really sorry for the long review delay here, and I appreciate your patience.

This LGTM!

Issue #5676, opened 2025-01-29
[BUG]: Platform parameter syncing races against SplashActivity


BenHenning:
### Describe the bug

Currently, platform parameter syncing relies on a application-scope singleton to be fully initialized before its properties will be correctly retrieved from the local data store. See:

https://github.com/oppia/oppia-android/blob/4f58be9d399c70e1d1cd241495280ae913afbf07/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt#L50

And

https://github.com/oppia/oppia-android/blob/4f58be9d399c70e1d1cd241495280ae913afbf07/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt#L88-L89

Yet, the singleton gracefully fails:

https://github.com/oppia/oppia-android/blob/4f58be9d399c70e1d1cd241495280ae913afbf07/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterSingletonImpl.kt#L47-L49

And ``SplashActivityPresenter`` relies on app-critical features during construction:

https://github.com/oppia/oppia-android/blob/4f58be9d399c70e1d1cd241495280ae913afbf07/app/src/main/java/org/oppia/android/app/splash/SplashActivityPresenter.kt#L78-L79

I have not tested this, but I strongly suspect that if the failure logic were to be updated above to fail with an exception that splash activity would always fail (which means it will never have a synced state). This, in turn, means that OS and app deprecation will never work correctly in the current implementation.

### Steps To Reproduce

I haven't devised actual steps to repro the issue.

### Expected Behavior

State needs to be synced such that even splash activity can have platform parameters upon starting.

### Additional Context

I think that this could probably be solved by introducing a hard block on ``PlatformParameterController`` during startup. Normally, this would happen in ``SplashActivity`` but it seems that may not be an option since ``SplashActivity`` is doing a lot of other stuff.

This is actually rather tricky: we want the Dagger injections to be non-blocking (which means there does need to be some state syncing ahead of injection), but we need a good point in the app startup cycle to perform this logic. My current thinking is that we should:
- Make invalid states throw an exception (so trying to inject unsynced parameters should always crash the app, including in splash).
- Update splash to use lazy injection for platform parameters.
- Introduce a proper syncing flow (such as adding a ``PlatformParameterController.syncAppState()`` function which force-refreshes the local state used by providers).

[...]

Issue #5345, opened 2024-02-17
[Feature Request]: Introduce platform parameter dashboard (GSoC'25 4.2)


BenHenning:
### Is your feature request related to a problem? Please describe.

Feature flags are a special type of configurable [platform parameter](https://github.com/oppia/oppia-android/wiki/Platform-Parameters-&-Feature-Flags#introduction) which allows the team to stage features behind remotely configurable flags until they're ready to be launched. This allows features to be developed across multiple releases without users seeing part of the feature (or app stability issues when the feature is enabled), ensuring the team releases high-quality features and doesn't hurt the overall quality and performance of the app. Broadly, platform parameters allow the team to configure the overall app (which can be useful both for feature flags, as described above, and safety 'knobs' such as controlling rate limits to remote APIs to help reduce the chance of server outages).

This project entails introducing a developer-only UI (as part of the developer options section of the app) which displays all platform parameters and feature flags in the app, their current enabled/disabled status (for feature flags) or values (for platform parameters), and their sync status (i.e. whether they're being synced from the server or using a local developer default). It also allows an explicit manual override to force the feature on or off, or to override the platform parameter's value.

Note that implementations should be based on the new version of platform parameters being introduced in https://github.com/oppia/oppia-android/pull/5725.

#### Suggested Milestones:
- **Milestone 1**:
  - Display a list of platform parameters and feature flags from a Developer Options menu, along with their current values and sync statuses.
  - Set up initial tests to demonstrate that the UI displays correctly.
  - The new UI correctly shows all platform parameters and feature flags, and their correct values and sync statuses.
  - Support for overriding feature flag values (requires manual restart).
- **Milestone 2**:
  - Support for overwriting and resetting both feature flags and platform parameters back to their default values, including force-restarting the app upon navigating away from the menu (so that the changes can take effect).
  - Updated UI tests for the new functionality.
  - The new screen fully supports overriding various platform parameters and feature flag values.
  - Pending overrides should 'survive' device rotations (that is, rotating the device should not cause any temporary state in the UI to be lost).
  - Support for force downloading flags from Oppia web (i.e. by calling PlatformParameterController.downloadRemoteParameters() and asking for an app restart). 
  - The indicator for downloading state should 'survive' device rotations.
[...]

Issue #5307, opened 2024-01-16
[BUG]: ActivityScenarioRule causes issues with platform parameter initialization


BenHenning:
### Describe the bug

During the development of #5138 it was discovered that ActivityScenarioRule can cause issues with any module-override static initialization (such as platform parameters). This is most likely to happen with platform parameters, and it manifests as the parameters not properly setting for some of the tests (due to the scenario being set up before the module override can occur in the test).

### Steps To Reproduce

Trying to run specific tests in ProfileAndDeviceIdFragmentTest can fail due to the platform parameters needed for the test not being properly set up ahead of the test running (running the whole suite doesn't fail since earlier tests don't depend on the parameter and give the module time to be initialized before the scenario actually sets up a UI which _does_ require the parameters).

### Expected Behavior

All tests should be able to be run independently of any particular execution order, especially with platform parameters.

All uses of ActivityScenarioRule should be removed and replaced with manual ActivityScenario management, as shown in ProfileAndDeviceIdFragmentTest (after #5138 is merged).

### What device/emulator are you using?

N/A -- test environment

### Which Android version is your device/emulator running?

N/A -- test environment

### Which version of the Oppia Android app are you using?

N/A -- test environment

### Additional Context

Multiple contributors can work on replacing the ActivityTestRule with ActivityScenarioRule at the same time. Create small PRs changing tests in one package at a time. See comment below for the list of test classes to be fixed.

BenHenning:
From #5138 the classes that need to be fixed:

- [x] app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileAndDeviceIdActivityTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/AppCompatCheckBoxBindingAdaptersTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/CircularProgressIndicatorAdaptersTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/DrawableBindingAdaptersTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/ImageViewBindingAdaptersTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/MarginBindingAdaptersTest.kt
- [x] app/src/sharedTest/java/org/oppia/android/app/databinding/StateAssemblerMarginBindingAdaptersTest.kt
[...]

PR #5725, opened 2025-02-28
Introduce new platform parameter system infrastructure

BenHenning:
## Explanation
TODO: Finish

## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [ ] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".

## For UI-specific PRs only


BenHenning:
@Rd4dev and @adhiamboperes I'd really like your thoughts on this initial approach for a reinventing of the platform parameter system. It's as much as I could get done today, but I'm really hoping it will serve as a significant foundation for both the testing work being done in #5565 and what we need to do for the rest of #5345.

At a high-level, the plan is:
- Isolate the singleton properly for state--it's currently confusing and messy existing in two different layers of the app.
- Introduce first-class support for feature flags.
- Leverage textproto definitions to replace the constants classes for both feature flags and platform parameters to make the definitions much cleaner and less error-prone (they now have strong typing guarantees!).
- Introduce code generation for the platform parameter module (not done yet) to automatically convert the textproto definitions to nice bindings (see the `FeatureFlag` and `PlatformParameter` qualifier files for examples).
- New injection styles for parameters and flags (see above).
- Proper loading ordering for parameters and flags (so that it doesn't race against splash).
- Consolidation of all things parameters/flags into one controller (vs. splitting it up with the platform parameter worker).
- Abstracting the controller so that multiple implementations can be provided (nice for testing and dev options).

A lot of files will eventually need to be updated, plus the old system torn out. Would be great to get your thoughts of progress so far.

Note that for testing I plan to update `OppiaTestRule` to sort of simulate an app restart by creating a completely separate application class in order to 'download' parameters and save them to disk, then load the parameters ahead of the test running. This will be used in conjunction with a new PlatformParameterTestOrchestrator class which will provide functionality for faking calls to the remote endpoint to simulate remotely synced parameters and flags. My hope is that this is relatively easy to hook into the work you're doing @Rd4dev.

Rd4dev:
@BenHenning, actually it took me some time to understand the sync race issue, but I was able to get a better picture along the way.

The restructured implementation would be very helpful to have a more organized approach, with works #5565, #5345. And I’m in favor of having a dedicated debug implementation.

>Note that for testing I plan to update OppiaTestRule to sort of simulate an app restart by creating a completely separate application class in order to 'download' parameters and save them to disk, then load the parameters ahead of the test running. This will be used in conjunction with a new PlatformParameterTestOrchestrator class which will provide functionality for faking calls to the remote endpoint to simulate remotely synced parameters and flags.

But I was not able to fully understand the testing plan for simulating an app restart. Is it intended for all platform parameter test scenarios? or is it specifically for validating the new system’s cache storage and retention?

If it's the latter, then what I was considering might not be relevant. But if not, I thought we could update the database before Dagger initialization in the test rule’s apply() method, based on test annotations, ensuring the updated values would already be picked up in the platform parameter controller during initialization. Wouldn't that be possible?

Would it be possible to clarify the testing overrides a bit more, particularly how the app restart simulation should be carried forward in #5565?

BenHenning:
Thanks @Rd4dev, I appreciate the read-through and follow-up. A lot has changed since that comment as I worked through trying to make this work. :) It is, admittedly, rather complex to get the initialization flows correct (which affects both the app and all tests that interact with code which use parameters/flags, or need to override them).

I think I figured out a working model. Please look at the following files in the latest changes:
- `TestPlatformParameterConfigRetriever`: This is the class that provides static override support for parameters and flags. An example of this working is `EventBundleCreatorTest` which performs overrides and utilizes the initializer. It also performs some recreation logic that was _very_ difficult to fix earlier today. :)
- `PlatformParameterTestInitializer`: This is the class that is needed in order to ensure that parameters and flags are loaded early enough for the Dagger injection bits to work. See `LogReportingModuleTest` for an example test that uses this without overrides. Basically we just need to inject it at the beginning of the test to make things work.

Note that application initialization has been hacked into `AbstractOppiaApplication` for now and needs to be redone. We really should be loading parameters in `SplashActivityPresenter` but we have a few parameters being used in the work management system. These need to be either carefully changed to `Provider`s or outright removed if they're initialized too early in the application lifecycle. We'll run into problems like this on occasion (including in `Splash*` classes), but it's sort of definitionally unavoidable. We need to have sane state and load that asynchronously _somehow_.

As for how this affects testing, my thoughts are:
- The static state managed by the config retriever can be easily bridged by your new test rules. This isn't much different than your existing/previous implementation (if I'm remembering correctly), but it's probably a lot cleaner/safer with this approach.
- We may be able to eliminate the initializer in favor of a rule-based approach, but I'm personally not exactly sure how. One idea that may have merit: introduce an injector for PlatformParameterController and an injector provider for it, then have the rule cast the current application class to the injector provider to retrieve the injector. The injector can be used to retrieve the controller and perform the same initialization steps that are being done today. Something similar may have to be done with the monitor, as well (or we can just avoid it to keep dependencies simpler). Unfortunately, this does require updating a _lot_ of `TestApplication`s and their components to implement the new interfaces.

This PR still needs a lot of work to get tests passing and adding new tests (beyond the initialization revamp mentioned above), but I think that this at least proves the viability of the new approach. I think we can use this as the basis now for both finishing your test PR and building a more clearly defined foundation for the dashboard project.

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingProfileTypeFragmentTest.kt
```diff
@@ -389,7 +388,7 @@ class OnboardingProfileTypeFragmentTest {
       ExplorationStorageModule::class, NetworkModule::class, NetworkConfigProdModule::class,
       NetworkConnectionUtilDebugModule::class, NetworkConnectionDebugUtilModule::class,
       AssetModule::class, LocaleProdModule::class, ActivityRecreatorTestModule::class,
-      PlatformParameterSingletonModule::class,
+      PlatformParameterTestModule::class,
```
adhiamboperes, line 419: Nit: Duplicate entry.

app/src/sharedTest/java/org/oppia/android/app/options/AudioLanguageFragmentTest.kt
```diff
@@ -623,7 +627,7 @@ class AudioLanguageFragmentTest {
       ExplorationStorageModule::class, NetworkModule::class, HintsAndSolutionProdModule::class,
       NetworkConnectionUtilDebugModule::class, NetworkConnectionDebugUtilModule::class,
       AssetModule::class, LocaleProdModule::class, ActivityRecreatorTestModule::class,
-      NetworkConfigProdModule::class, PlatformParameterSingletonModule::class,
+      NetworkConfigProdModule::class, PlatformParameterTestModule::class,
```
adhiamboperes: Nit: duplicate entry.

app/src/sharedTest/java/org/oppia/android/app/options/OptionsFragmentTest.kt
```diff
@@ -632,8 +631,8 @@ class OptionsFragmentTest {
   @Singleton
   @Component(
     modules = [
-      TestPlatformParameterModule::class,
-      RobolectricModule::class, PlatformParameterSingletonModule::class,
+      PlatformParameterTestModule::class,
+      RobolectricModule::class, PlatformParameterTestModule::class,
```
adhiamboperes: Nit: duplicate entry.

app/src/sharedTest/java/org/oppia/android/app/player/exploration/ExplorationActivityTest.kt
```diff
@@ -2877,7 +2876,7 @@ class ExplorationActivityTest {
   @Component(
     modules = [
       RobolectricModule::class,
-      TestPlatformParameterModule::class, PlatformParameterSingletonModule::class,
+      PlatformParameterTestModule::class, PlatformParameterTestModule::class,
```
adhiamboperes: Nit: duplicate entry.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -6178,7 +6205,7 @@ class StateFragmentTest {
       ExplorationStorageModule::class, NetworkConnectionUtilDebugModule::class,
       NetworkConnectionDebugUtilModule::class, NetworkModule::class, NetworkConfigProdModule::class,
       AssetModule::class, LocaleProdModule::class, ActivityRecreatorTestModule::class,
-      PlatformParameterSingletonModule::class,
+      PlatformParameterTestModule::class,
```
adhiamboperes, line 6283: Nit: duplicate entry.

app/src/sharedTest/java/org/oppia/android/app/onboarding/OnboardingFragmentTest.kt
```diff
@@ -1134,7 +1135,7 @@ class OnboardingFragmentTest {
   @Component(
     modules = [
       RobolectricModule::class,
-      TestPlatformParameterModule::class, PlatformParameterSingletonModule::class,
+      PlatformParameterTestModule::class, PlatformParameterTestModule::class,
```
adhiamboperes: Nit: Duplicates.

adhiamboperes (commented):
@BenHenning, I took a highlevel look at the approach, and would just like two clarifications:
1. It looks like we are planning to have the `TestPlatformParameterConfigRetriever` replace the annotation system introduced in https://github.com/oppia/oppia-android/pull/5565, since they perform the same ovveride function.
2. In app module tests like `OnboardingFragmentTest`, `TestPlatformParameterConfigRetriever.setFlagOverride()` is used without first injecting `PlatformParameterTestInitializer` like in `EventBundleCreatorTest` for example. Is this by design or general work still needing to be updated?

BenHenning:
> @BenHenning, I took a higlevel look at the approach, and would just like two clarifications:
> 
> 1. It looks like we are planning to have the `TestPlatformParameterConfigRetriever` replace the annotation system introduced in [Fixes part of #4302 - Introducing a JUnit annotation & rule for overriding Platform Parameters and Feature Flags #5565](https://github.com/oppia/oppia-android/pull/5565), since they perform the same ovveride function.
> 2. In app module tests like `OnboardingFragmentTest`, `TestPlatformParameterConfigRetriever.setFlagOverride()` is used without first injecting `PlatformParameterTestInitializer` like in `EventBundleCreatorTest` for example. Is this by design or general work still needing to be updated?

Thanks for the high-level review @adhiamboperes, and I'm aware of the duplicate references--they're the least of the compiler failures to fix in a few dozen tests. :) Slowly getting through those...

1. Nope, this does not replace @Rd4dev's project. It's replacing the test module we use today, plus several other hackier approaches we've used to override flags. If anything, this streamlines the annotation system as it provides an easier hook-in for the annotation system to use. Ideally, we'd also figure out how to replace the new test initializer so that tests don't need it (and can instead rely on a rule). That's something RD may want to figure out after this PR gets merged.
2. This is intentional. Flags must be overwritten statically ahead of the initializer preparing for parameter and flag injection. That's actually why this pairs nicely with annotations since those are also, effectively, static (since they are processed before the test starts being run). Annotations also offer an even stronger safety net since they guarantee one override per test (whereas the test retriever can be called multiple times, though it will throw an error if done).

BenHenning:
Going to close this for now since I'll be pushing a lot of commits and I don't want to overload CI.

BenHenning:
Alas, I have discovered new problems. None of the app tests are going to work by default because they rely on InjectableAppCompatActivity which injects AppLanguageWatcherMixin which injects ProfileManagementController which relies on feature flags, and all of this happens before setUp() if using ActivityScenarioRule. :)

The thing is, #5307 could also be solved by using a test rule (a la #5565) that properly sets up config parameters before ActivityScenarioRule can run. Currently, there are 27 tests to update to not use ActivityScenarioRule, so I'm uncertain how best to proceed. Perhaps it would be preferable just to remove using the rule (even if it can be made workable in the future) as a prerequisite to this PR.

Edit: There are also 49 ActivityTestRule cases that would need to be updated, too.

model/src/main/proto/platform_parameter.proto
```diff
@@ -5,36 +5,152 @@ package model;
...
+  NPS_SURVEY_GRACE_PERIOD_IN_DAYS = 10;
+  NPS_SURVEY_MINIMUM_AGGREGATE_LEARNING_TIME_IN_A_TOPIC_IN_MINUTES = 11;
+}
+
+// Represents a forever-unique reference to a specific feature flag.
+enum FeatureFlagId {
+  // Corresponds to a feature flag whose ID is not known (e.g. invalid or undefined).
+  FEATURE_FLAG_ID_UNSPECIFIED = 0;
+
+  // TODO: Finish documentation & update the indexes based on creation date of each.
```
BenHenning: Match the values in #5715 if that's merged first.

model/src/main/proto/platform_parameter.proto
```diff
@@ -5,36 +5,152 @@ package model;
...
-    int32 integer = 3;
-    // Indicates a string-typed platform parameter.
-    string string = 4;
-  }
+// Represents a forever-unique reference to a specific platform parameter.
+enum PlatformParameterId {
+  // Corresponds to a platform parameter whose ID is not known (e.g. invalid or undefined).
+  PLATFORM_PARAMETER_ID_UNSPECIFIED = 0;
+
+  // TODO: Finish documentation & update the indexes based on creation date of each.
```
BenHenning: See https://github.com/oppia/oppia-android/pull/5715#discussion_r2032090170.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
```diff
@@ -0,0 +1,330 @@
...
+      processState.initializeFeatureFlags(flagStatesById)
+      processState.initializeFeatureFlagSyncStatuses(statusesById)
+
+      // Erase the data provider's value so that callers cannot inadvertently depend on the actual
+      // list of parameters available.
+      return@transform Unit
+    }
+  }
+
+  override fun downloadRemoteParameters(): DataProvider<Any?> {
```
Rd4dev, line 87: Following up on the discussion reg. "Force Download Remote Parameters' Status Indication" —requesting to include `Pending` and `Failure` results to support the status indicators planned with #5345.

model/src/main/proto/platform_parameter.proto
```diff
@@ -5,36 +5,152 @@ package model;
...
+message RemotePlatformParameter {
+  // The unique ID of this platform parameter which corresponds to a PlatformParameterDefinition.
+  PlatformParameterId id = 1;
+
+  // The remotely synced value of this parameter, or absent/default if never yet synced.
+  PlatformParameterValue remote_value = 2;
+
   // Indicates the sync status of the platform parameter.
-  SyncStatus sync_status = 5;
+  SyncStatus sync_status = 3;
```
Rd4dev, line 74: Following up to confirm if we should proceed with removing the `sync_status` field from `RemotePlatformParameter`, since a fetched remote value could be treated as default `SYNCED_FROM_SERVER` (assuming the current scope).

BenHenning:
NB This should fix #5676.

BenHenning:
This PR is going to be split up in the effort of time. Specifically, my plan is:
- Introduce the new platform parameter controller directly and hook it up to the existing injection & testing approaches for platform parameters and feature flags. This means building interop.
- Introduce the new injection approach for feature flags and platform parameters.
  - This may be done in 2 parts: one for introducing the new generation script, and one for actually hooking it up.
- Introduce the new testing approach for feature flags and platform parameters.
  - This may require much of the ongoing module cleanup work to continue (or, rather, significantly benefit from it).
  - This may also require some forward work around adding test application classes to all existing tests. I may opt for code generation for this as well.

BenHenning:
#5835 is tracking the remaining work for this PR. I plan to use this PR as the final bit of addressing that issue once the preliminary pieces are done.

Closing for now since this won't be completely finished for several weeks yet.

Issue #5571, opened 2024-11-06
[Feature Request]: Support practice questions on the Android app


seanlip:
### Is your feature request related to a problem? Please describe.

We implemented a Practice tab in the topic pages of the Android app, but turned it off because, at the time, practice questions didn't yet exist for the topics we were providing. Now that practice questions have been created for our core math topics ([example](https://www.oppia.org/learn/math/place-values/practice)), we would like to turn it on again, because it is helpful for learners to be able to practise skills after learning a new topic so that they can develop mastery. 

The aim of this issue is to turn on this functionality in the Android app and make any necessary adjustments to it so that it can be launched.

### Describe the solution you'd like

Roughly, bringing this functionality to users involves 4 steps:

- [x] Turn on the existing practice questions functionality on Android
        - We have created the feature flag(#5729)
- [ ] do a verification (by developers) that the functionality works and that it is accessible and playable with the developer assets.

https://github.com/user-attachments/assets/85301a5b-d7f0-4f71-8b85-537cd718fce5

- [ ] do a verification (by developers) that the functionality works and that it is accessible and playable with production assets. - blocked on the download script. -- tracked by #6072
- [ ] Launch this functionality on alpha so that the product team can audit it (both the entry-point to practice questions and the functionality itself).
- [ ] The product team should then do a gap analysis to determine what other improvements are needed before the feature can be launched, filing design/implementation issues for specific bugs as needed.
- [ ] Once all prerequisites have been completed for the feature to launch, launch the feature.

**Important Note**: We should also verify that it is possible to turn practice questions on and off on a per-topic basis. On Web, there is a toggle in the topic editor page that determines whether to show practice questions for that topic or not, and that toggle can be used to determine whether or not to show the Practice tab for that topic in the Android app. This functionality is necessary because, from previous experience, practice question development can sometimes lag the creation of new topics.

adhiamboperes:
After merging https://github.com/oppia/oppia-android/pull/5729, I verified the feature and found that it’s not fully working yet. Additional coding is needed for the app to correctly parse the downloaded questions.

Since #5729 was created before https://github.com/oppia/oppia-android/pull/4885/commits/5f65a7f0880d8c8e181adbe2e54a58a2cfa43b33, the question structure still needs to be defined and properly wired on Android.

Edit - questions work with the dev assets. Per the CLaM meeting discussion today, there is further script work needed to verify that the scripts work with the prod assets too.

BenHenning:
[...]

PR #5729, opened 2025-03-02
Fix part of #5571: Support practice questions

theMr17:
## Explanation

Fixes part of #5571

Currently, the single feature flag `EXTRA_TOPICS_TAB_UI` controls the visibility of both the Info and Practice tabs on Topic screens. However, #5571 requires enabling only the Practice tab, which is not possible with the current setup.

This PR introduces two separate feature flags (`TOPIC_INFO_TAB` & `TOPIC_PRACTICE_TAB`) to independently control the visibility of the Info and Practice tabs.

## Screenshots

|Practice tab enabled|Info tab enabled|Both tabs enabled|
|--|--|--|
||Screenshot_20250908_211225|Screenshot_20250908_211046|

## Essential Checklist


theMr17:
Hey @adhiamboperes, please review the gating logic using the two feature flags. Once confirmed, I will proceed with implementing the ability to enable or disable practice questions on a per-topic basis.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
```diff
@@ -185,7 +186,8 @@ class TopicActivityTest {
   @Test
   @RunOn(TestPlatform.ROBOLECTRIC) // TODO(#3858): Enable for Espresso.
   fun testTopicActivity_startPracticeSession_questionActivityStartedWithProfileId() {
-    TestPlatformParameterModule.forceEnableExtraTopicTabsUi(true)
+    TestPlatformParameterModule.forceEnableTopicInfoTab(true)
+    TestPlatformParameterModule.forceEnableTopicPracticeTab(true)
```
adhiamboperes: Can this test work with only the `forceEnableTopicPracticeTab`? It looks like it is only testing the practice tab.

app/src/main/java/org/oppia/android/app/topic/TopicTab.kt
```diff
@@ -7,34 +7,44 @@ import org.oppia.android.app.ui.R
 /** Enum to store the tabs of [TopicFragment] and get tab by position. */
 enum class TopicTab(
   val positionWithTwoTabs: Int,
+  val positionWithThreeTabsWithInfo: Int,
+  val positionWithThreeTabsWithPractice: Int,
   val positionWithFourTabs: Int,
   @StringRes val tabLabelResId: Int,
   @DrawableRes val tabIconResId: Int,
   @StringRes val contentDescriptionResId: Int
```
adhiamboperes:
```suggestion
enum class TopicTab(
  @StringRes val tabLabelResId: Int,
  @DrawableRes val tabIconResId: Int,
  @StringRes val contentDescriptionResId: Int,
  private val positions: Map<TabConfig, Int>
)
```
We could remove the redundant positionWithTwoTabs, positionWithThreeTabsWithInfo, etc., and replace them with a `getPosition` function(see comment below), while also encapsulating tab configurations.

```kotlin
 /** Enum representing different tab configurations. */
  private enum class TabConfig {
    TwoTabs, ThreeTabsWithInfo, ThreeTabsWithPractice, FourTabs;

    companion object {
      fun getConfig(enableInfo: Boolean, enablePractice: Boolean) = when {
        enableInfo && enablePractice -> FourTabs
        enableInfo -> ThreeTabsWithInfo
        enablePractice -> ThreeTabsWithPractice
        else -> TwoTabs
      }
    }
  }
  
    /** Returns the tab position based on enabled tabs. */
  fun getPosition(enableInfo: Boolean, enablePractice: Boolean): Int? {
    val config = TabConfig.getConfig(enableInfo, enablePractice)
    return positions[config]
  }
```

Then

```kotlin
  INFO(
    tabLabelResId = R.string.info,
    tabIconResId = R.drawable.ic_info_icon_24dp,
    contentDescriptionResId = R.string.info_tab_content_description,
    positions = mapOf(
      TabConfig.ThreeTabsWithInfo to 0,
      TabConfig.FourTabs to 0
    )
  ),
  ```

File: app/src/main/java/org/oppia/android/app/topic/TopicTab.kt [hunk 3]
```diff
@@ -44,20 +54,38 @@ enum class TopicTab(
...
+          position == if (enableTopicInfoTab && enableTopicPracticeTab) {
             it.positionWithFourTabs
-          } else it.positionWithTwoTabs
+          } else if (enableTopicInfoTab) {
+            it.positionWithThreeTabsWithInfo
+          } else if (enableTopicPracticeTab) {
+            it.positionWithThreeTabsWithPractice
+          } else {
+            it.positionWithTwoTabs
+          }
```
adhiamboperes: Based on my suggestion above, this could be condensed into:
```suggestion
       fun getTabForPosition(position: Int, enableInfo: Boolean, enablePractice: Boolean): TopicTab {
      return values().find { it.getPosition(enableInfo, enablePractice) == position }
        ?: error("No tab corresponding to position: $position")
    }

**Review by adhiamboperes — CHANGES_REQUESTED** [review]
Thanks @theMr17! I have left some suggestions, PTAL and let me know your thoughts.

File: app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt (hunk 1 above)
Comment (theMr17) [reply app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt]: ~Yes that should work, removed ` TestPlatformParameterModule.forceEnableTopicInfoTab(true)`~

Edit: We need to enable info tab as well since we are checking for the `topic_name_text_view` (present in `topic_name_text_view.xml`) in the `launchTopicActivity()` function.

File: app/src/main/java/org/oppia/android/app/topic/TopicTab.kt (hunk 2 above)
Comment (theMr17) [reply app/src/main/java/org/oppia/android/app/topic/TopicTab.kt]: Thanks for the suggestion! This makes the code much cleaner.

File: app/src/main/java/org/oppia/android/app/topic/TopicTab.kt (hunk 3 above)
Comment (theMr17) [reply app/src/main/java/org/oppia/android/app/topic/TopicTab.kt]: Done.

**Comment by theMr17** [comment]
Thanks for the review @adhiamboperes! Addressed the comments and moving ahead with the implementation of turning practice questions on/off on a per-topic basis.

**Comment by theMr17** [comment]
Hey @adhiamboperes! I've introduced an `enable_practice_tab` field in the JSON/proto assets. This field is read by the `TopicController`, passed to the `TopicViewModel`, and then forwarded to the presenter. In the presenter, I'm combining this boolean with the `ENABLE_TOPIC_PRACTICE_TAB` feature flag using `&&` to control the visibility of the practice tab on the topic screen.

I'm a bit unsure about how to update the existing tests to reflect this change--would appreciate any suggestions there.

Could you please take a look at the approach? I'd like to hear your thoughts.

You can check out the changes [here](https://github.com/oppia/oppia-android/pull/5729/files/ae940acfecb22aaa1c7e3480c51d8bb9f86da1e8..e932376516d27b586d8b8eec9b696f83661f505c).

File: domain/src/main/assets/test_topic_id_2.textproto [hunk 4]
```diff
@@ -1,4 +1,5 @@
 id: "test_topic_id_2"
+enable_practice_tab: false
```
adhiamboperes: Maybe, `has_practice_questions` would be more representative?

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -485,6 +485,7 @@ class TopicController @Inject constructor(
               availableToPlayNow = true
             } else availableToPlayInFuture = true
           }.build()
+          enablePracticeTab = topicRecord.enablePracticeTab
```
adhiamboperes:
```suggestion
          hasPracticeQuestions = topicRecord.enablePracticeTab
```

The controller should not need context on what we intend to do with this variable, the viewmodel will make the decision. here, we only need to know whether the topic has practice questions.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -76,7 +78,10 @@ class TopicFragmentPresenter @Inject constructor(
     viewModel.setTopicId(topicId)
     binding.viewModel = viewModel
 
-    setUpViewPager(viewPager, classroomId, topicId, isConfigChanged)
+    viewModel.enablePracticeTab.observe(fragment) { enablePracticeTab ->
+      setUpViewPager(viewPager, classroomId, topicId, isConfigChanged, enablePracticeTab)
```
adhiamboperes: This LiveData should only be observed if `enableTopicPracticeTab.value` is `true`. It's best to apply gating logic at the highest level possible to simplify testing and improve code clarity. 

Additionally, since the topic's support for practice questions is a static property that doesn't change at runtime, using LiveData isn't appropriate in this case. A simple `val` or a computed getter function would be more suitable.

app/src/main/java/org/oppia/android/app/topic/TopicTab.kt
```diff
@@ -6,58 +6,104 @@ import org.oppia.android.app.ui.R
...
+    /**
+     *  Returns the number of active tabs considering [enableTopicInfoTab] and
+     *  [enableTopicPracticeTab].
+     */
+    fun getTabCount(enableTopicInfoTab: Boolean, enableTopicPracticeTab: Boolean) =
+      if (enableTopicInfoTab && enableTopicPracticeTab)
+        values().size
+      else if (enableTopicInfoTab || enableTopicPracticeTab)
+        values().size - 1
+      else values().size - 2
```
adhiamboperes: Making use of the new structures introduced:
```suggestion
   fun getTabCount(enableTopicInfoTab: Boolean, enableTopicPracticeTab: Boolean): Int {
  val config = TabConfig.getConfig(enableTopicInfoTab, enableTopicPracticeTab)
  return values().count { it.positions[config] != -1 }
}
```

This is less brittle/more flexible.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
adhiamboperes: I would have approached this by creating a boolean global variable, `enableTopicPracticeTab`, with a default value of `false`. 

Then I would update it's value based on  

```kotlin
enableTopicPracticeTab = enableTopicPracticeTab.value && viewModel.hasPracticeQuestions
```

I would then use the variable everywhere else, to avoid computing `enableTopicPracticeTab.value && enablePracticeTab` every time.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -120,14 +125,15 @@ class TopicFragmentPresenter @Inject constructor(
...
+    return tab.getPosition(enableTopicInfoTab.value, enableTopicPracticeTab.value)
   }
 
   private fun setUpViewPager(
     viewPager2: ViewPager2,
     classroomId: String,
     topicId: String,
-    isConfigChanged: Boolean
+    isConfigChanged: Boolean,
+    enablePracticeTab: Boolean
```
adhiamboperes: It is somewhat inconsistent to pass `enablePracticeTab` boolean as an argument, and then call `enableTopicInfoTab` from the top level declaration. the function would be cleaner if you passed both variables in the function parameter list.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
```diff
@@ -136,25 +142,36 @@ class TopicFragmentPresenter @Inject constructor(
...
+        enableTopicPracticeTab.value && enablePracticeTab
       )
     viewPager2.adapter = adapter
     TabLayoutMediator(tabLayout, viewPager2) { tab, position ->
-      val topicTab = TopicTab.getTabForPosition(position, enableExtraTopicTabsUi.value)
+      val topicTab = TopicTab.getTabForPosition(
+        position,
+        enableTopicInfoTab.value,
+        enableTopicPracticeTab.value && enablePracticeTab
+      )
```
adhiamboperes, line 166: I would like you to review this old code, and determine whether we still need the code under `TabLayoutMediator`, since we have a function for returning tab position.

app/src/main/java/org/oppia/android/app/topic/TopicViewModel.kt
```diff
@@ -78,6 +78,12 @@ class TopicViewModel @Inject constructor(
     }
   }
 
+  val enablePracticeTab: LiveData<Boolean> by lazy {
+    Transformations.map(topicLiveData) { ephemeralTopic ->
+      ephemeralTopic.topic.enablePracticeTab
+    }
+  }
```
adhiamboperes: Would this work instead?

`val enablePracticeTab: Boolean = topicLiveData.value.ephemeralTopic.topic.enablePracticeTab`

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
adhiamboperes: You response is not clear with regards to my original question. Is the visibility of `topic_name_text_view` gated behind both flags? why?

adhiamboperes: If we are in the practice session, the status of the info tab should be irrelevant. Otherwise it indicates a problem with the implementation.

adhiamboperes:
> Hey @adhiamboperes! I've introduced an `enable_practice_tab` field in the JSON/proto assets. This field is read by the `TopicController`, passed to the `TopicViewModel`, and then forwarded to the presenter. In the presenter, I'm combining this boolean with the `ENABLE_TOPIC_PRACTICE_TAB` feature flag using `&&` to control the visibility of the practice tab on the topic screen.
> 
> I'm a bit unsure about how to update the existing tests to reflect this change--would appreciate any suggestions there.
> 
> Could you please take a look at the approach? I'd like to hear your thoughts.
> 
> You can check out the changes [here](https://github.com/oppia/oppia-android/pull/5729/files/ae940acfecb22aaa1c7e3480c51d8bb9f86da1e8..e932376516d27b586d8b8eec9b696f83661f505c).



Are you looking for suggestions for testing the controller, or the fragment?

adhiamboperes (changes requested):
Thanks @theMr17! 

The implementation still appears to need more streamlining, so we could hold off on the tests for now. The legacy implementaion appears tangled, and it would be useful to take a step back and see which parts could be improved first, in order to integrate well with the new solution.

domain/src/main/assets/test_topic_id_2.textproto
theMr17: This sounds much more appropriate. Updated.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
theMr17: Done.

app/src/main/java/org/oppia/android/app/topic/TopicTab.kt
theMr17: Updated.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
theMr17: > You response is not clear with regards to my original question. Is the visibility of topic_name_text_view gated behind both flags? why?

The `topic_name_text_view` is gated behind `ENABLE_TOPIC_INFO_TAB` only. Note that `topic_name_text_view` is a text element inside the info tab. The test launches the `TopicActivity` and expects the info tab to be viewed. Then clicks on the practice tab and navigates away from the info tab to the practice tab.
```
onView(withText("Practice")).perform(click())
```

Considering the above setup of the test, we need to have both the feature flags enabled.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
theMr17: I took a look through the code. `getTabForPosition()` will still be required here because it's used here to map a position (0, 1, 2,...) to the current TopicTab enum value. This is then used to set tab label, icon, and accessibility description. It could be technically replaced with a hardcoded list of tab labels, but then it would be less maintainabile.

app/src/main/java/org/oppia/android/app/topic/TopicViewModel.kt
theMr17: This approach won't work because `topicLiveData` is initially `null` until it receives data from the controller. Therefore, `enablePracticeTab` will need to be a `LiveData`, similar to the existing `numberOfChaptersCompletedLiveData`.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
theMr17: Instead of passing both the variables in the function parameter, I have created global variables for both info and practice tab to maintain consistency.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
theMr17: Following the later suggestion, I have added a global variable for `enableTopicPracticeTab`. It is updated by observing the `hasPracticeQuestions` live data from the viewmodel. As mentioned in the other comment, it needs to be a live data because `hasPracticeQuestions` is getting its value from the `topicLiveData`, which is null initially untill it gets the data from the controller. So, a simple `val` or getter will not work.

theMr17:
Hey @adhiamboperes, thanks for the review! It took a bit longer than expected, but all comments have been addressed now, PTAL.

theMr17:
Hi @adhiamboperes , PTAL at this PR. Once, the current approach is approved. I will go ahead with tests.

adhiamboperes:
Hi @theMr17, there are quite a number of files here with conflicts. Please address that, so that we can review the PR.

theMr17:
> Hi @theMr17, there are quite a number of files here with conflicts. Please address that, so that we can review the PR.

All conflicts are resolved now, thanks!

subhajitxyz:
@theMr17 , Please update the PR description with the intended output, along with screenshots.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
subhajitxyz: Since we are using separate feature flag for info tab and practice tab. I think we should remove this line `onView(withId(R.id.topic_name_text_view)).check(matches(isDisplayed()))` from fun `launchTopicActivity`. So that we can properly test with both feature flag on and off.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -128,21 +128,34 @@ import org.oppia.android.util.networking.NetworkConnectionUtilDebugModule
...
+
+// When both extra tabs (info and practice) are disabled
 private const val LESSON_TAB_POSITION_EXTRA_TABS_DISABLED = 0
 private const val REVISION_TAB_POSITION_EXTRA_TABS_DISABLED = 1
 
+// When only the info tab is enabled
+private const val LESSON_TAB_POSITION_INFO_ENABLED_PRACTICE_DISABLED = 1
+private const val REVISION_TAB_POSITION_INFO_ENABLED_PRACTICE_DISABLED = 2
+
+// When only the practice tab is enabled
```
subhajitxyz: Apply to others also.

```suggestion
// When only the practice tab is enabled.
```

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagBindingModule.kt
```diff
@@ -95,6 +91,17 @@ class FeatureFlagBindingModule {
...
+  @EnableTopicInfoTab
+  fun provideEnableTopicInfoTab(processState: PlatformParameterProcessState) =
+    processState.retrieveFeatureFlag(FeatureFlagId.TOPIC_INFO_TAB)
+
+  @Provides
+  @EnableTopicPracticeTab
+  fun provideEnableTopicPracticeTab(processState: PlatformParameterProcessState) =
+    processState.retrieveFeatureFlag(FeatureFlagId.TOPIC_PRACTICE_TAB)
+
+
```
subhajitxyz: Remove this extra line space.

subhajitxyz:
Hi @theMr17 ,the approach looks great. I’ve left some review comments.
One more thing: do we actually need separate feature flags for the **info tab** and the **practice tab**? From what I see, these two were previously controlled by a single feature flag.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
theMr17: Done.

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagBindingModule.kt
theMr17: Done.

theMr17:
> @theMr17 , Please update the PR description with the intended output, along with screenshots.

Added the description and the screenshots.



> Hi @theMr17 ,the approach looks great. I’ve left some review comments.
> One more thing: do we actually need separate feature flags for the **info tab** and the **practice tab**? From what I see, these two were previously controlled by a single feature flag.

Addressed both the review comments. And mentioned in the PR description, why 2 separate feature flags are needed. Thanks @subhajitxyz, PTAL!

subhajitxyz:
Hi @theMr17 , Please correct the failing tests.

utility/src/main/java/org/oppia/android/util/platformparameter/FeatureFlagConstants.kt
```diff
@@ -189,3 +177,27 @@ const val ENABLE_FLASHBACK_SUPPORT = "android_enable_flashback_support"
...
+ * Qualifier for the feature flag corresponding to enabling the topic practice tab.
+ */
+@Qualifier
+annotation class EnableTopicPracticeTab
+
+/** Name of the feature flag that controls whether to enable the topic practice  tab. */
+const val ENABLE_TOPIC_PRACTICE_TAB = "enable_topic_practice_tab"
+
+/** Default value for the feature flag corresponding to [EnableTopicPracticeTab]. */
+const val ENABLE_TOPIC_PRACTICE_TAB_DEFAULT_VALUE = false
```
adhiamboperes, line 203: I actually don't think this constatnts file is being used anymore since we moved to using `feature_flags.textproto`. PTAL.

adhiamboperes (commented):
@theMr17, the overall approach LGTM, please proceed to wrap up the PR.

utility/src/main/java/org/oppia/android/util/platformparameter/FeatureFlagConstants.kt
theMr17: Tried removing these changes, but seems like the annotations are still needed.

theMr17:
Hey @adhiamboperes, addressed your comment and fixed/added tests. This PR is now ready for a complete review, PTAL!

utility/src/main/java/org/oppia/android/util/platformparameter/FeatureFlagConstants.kt
adhiamboperes: @theMr17, this was missed in a previous PR but will be removed soon.

app/src/sharedTest/java/org/oppia/android/app/testing/TopicTestActivityForStoryTest.kt
adhiamboperes, line 1: In this suite, shouldn't we ensure coverage for the new tab configurations?

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
adhiamboperes: Yes, I agree with @subhajitxyz on this. We should make the `launchTopicActivity` helper function as lean and reusable as possible, so as not to impact the ability to write proper tests.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
adhiamboperes, line 1: Would it be useful to add tests for these changes?

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
```diff
@@ -418,47 +437,89 @@ class TopicFragmentTest {
...
+            enableTopicInfoTab = true,
+            enableTopicPracticeTab = true
+          ).name
+        )
+      ).check(matches(isDescendantOfA(withId(R.id.topic_tabs_container))))
+    }
+  }
+
+  @Test
+  fun testTopicFragment_disableExtraTabs_isNotDisplayedInTabLayout() {
```
adhiamboperes: The test name doesn't highlight which tab(or if both) are not expected to be displayed.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
adhiamboperes, line 1: I believe we also need tests to verify the condition for the practice tab not being shown when a topic does not have practice questions.

adhiamboperes (changes requested):
Thanks @theMr171 This PR is close to being complete. 

I have some thoughts regarding tests. PTAL.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
theMr17: The condition is being verified by the following test:
https://github.com/oppia/oppia-android/blob/768bca748d6d47b23302b776a4b43a39395b1299/app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt#L921-L939

app/src/sharedTest/java/org/oppia/android/app/topic/TopicActivityTest.kt
theMr17: Thanks for the suggestions! I have updated the helper function accordingly.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
theMr17: Yup! added tests to ensure that correct values of the `hasPracticeQuestions` field are loaded from the dev assets.

app/src/sharedTest/java/org/oppia/android/app/topic/TopicFragmentTest.kt
theMr17: Done.

app/src/sharedTest/java/org/oppia/android/app/testing/TopicTestActivityForStoryTest.kt
theMr17: `TopicTestActivityForStory` uses `TopicActivityPresenter`, which in turn hosts the original `TopicFragment`. Since `TopicFragmentTest` already provides full coverage across all tab configurations, those cases are effectively covered for `TopicTestActivityForStoryTest` as well.

Similarly, the original `TopicActivityTest` also depends on `TopicFragment` to verify tab behavior, so tab-related scenarios are already exercised there.

Given this overlap, I don't believe additional tests are necessary for this suite. Please let me know if you have any concerns.

theMr17:
Thanks for the review @adhiamboperes! The comments are now addressed, PTAL.

adhiamboperes (approved):
LGTM, thanks @theMr17!

app/src/sharedTest/java/org/oppia/android/app/testing/TopicTestActivityForStoryTest.kt
adhiamboperes: sg.

adhiamboperes:
Will merge this tomorrow after cutting the release branch. In the meantime, please feel free to resolve all of my super old commnets - they take a bit of time to find in the old Github UI.

theMr17:
> Will merge this tomorrow after cutting the release branch. In the meantime, please feel free to resolve all of my super old commnets - they take a bit of time to find in the old Github UI.

All comments are now resolved, thanks!

theMr17:
https://github.com/oppia/oppia-android/commit/a45a694c2ce7c0044667b7affeb364d4b6da73e7 was accidentally pushed to the wrong branch 🙂. It's been reverted now, so everything should be fine.

config/src/java/org/oppia/android/config/platform/feature_flags.textproto
```diff
@@ -68,3 +63,13 @@ feature_flag_definition {
   remote_server_name: "android_enable_flashback_support"
   default_is_enabled: true
 }
+feature_flag_definition {
+  id: TOPIC_INFO_TAB
+  remote_server_name: "enable_topic_info_tab"
```
BenHenning: The server names need to include 'android_'.

config/src/java/org/oppia/android/config/platform/featureoverrides/alpha/feature_flags_overrides.textproto
```diff
@@ -14,3 +14,7 @@ feature_flag_definition {
   id: ONBOARDING_FLOW_V2
   default_is_enabled: true
 }
+feature_flag_definition {
+  id: TOPIC_PRACTICE_TAB
+  default_is_enabled: true
```
BenHenning: We probably can't do this yet--we need to first validate that the script can successfully import working questions. I've verified they can download but not that they work yet.

model/src/main/proto/topic.proto
```diff
@@ -679,6 +682,9 @@ message TopicRecord {
   // Whether this topic is published/available to be played.
   bool is_published = 6;
 
+  // Whether practice questions are available for this topic.
+  bool has_practice_questions = 11;
```
BenHenning, line 686: For this & the above--why are these needed exactly? Trying to understand the use case for why topics need to specify this. Does it come from Oppia web?

BenHenning (approved):
This is fantastic @theMr17! Really excited to see this move along. I had some questions that I think need answers before I can approve this. PTAL.

BenHenning (changes requested):
Correcting review to be 'request changes'--accidentally approved in my previous submission.

model/src/main/proto/topic.proto
theMr17: Some topics may not yet have practice questions available. This field indicates whether a topic is ready with practice content. If the topic has practice questions and the `ENABLE_PRACTICE_TAB` flag is enabled, the practice tab will be shown on the topic screen. Briefly, this field helps us to toggle the visibility of the practice tab, on a per topic basis.

Yes, this field should come from the web along with other topic content.

config/src/java/org/oppia/android/config/platform/feature_flags.textproto
theMr17: Updated both the server names.

config/src/java/org/oppia/android/config/platform/featureoverrides/alpha/feature_flags_overrides.textproto
theMr17: I have removed this override from the alpha file.

theMr17:
Thanks for the review @BenHenning! The comments are addressed, PTAL.

model/src/main/proto/topic.proto
BenHenning: Got it--thanks. We'll need to make sure to update the lesson download script for this new field, as well, which will require a change to the Oppia proto API work-in-progress branch. Let me know when that needs to happen and I can make the changes--it looks relatively simple.

BenHenning (approved):
Thanks @theMr17! This LGTM per the specific codeowner changes (I didn't review the UI pieces in detail). Note that I did leave a follow-up comment (now resolved) about updating the download script that we will want to make sure happens before we can test questions with real lesson data.

PR #5755, opened 2025-03-16
Fix #5307: Remove all uses of ActivityScenarioRule.

### LINKED ISSUE #5307: [BUG]: ActivityScenarioRule causes issues with platform parameter initialization

BenHenning:
## Explanation
Fixes #5307

This migrates tests over to managing their `ActivityScenario` directly, instead. This change does not affect existing uses of `ActivityTestRule`.

The existing exemptions have been removed to ensure there are now zero tests that are using `ActivityScenario`.

There were some additional clean-ups done, including:
- Removing extra lines for injections in a bunch of files.
- Adding a test target for `TextInputLayoutBindingAdaptersTest` since, as a Java test, it isn't covered by the test file static check.
- The error for using `ActivityScenarioRule` was updated to drop the platform parameters bit since, while true, it won't be as relevant after #5725 is completed since it will become impossible to use pre-setup orchestration of an activity.

Some other notes:
- This work is being done as a prerequisite to #5725 per https://github.com/oppia/oppia-android/pull/5725#issuecomment-2727017259.
- `HtmlParserTest` required an extra bit of logic to reset a mock state in `testHtmlContent_withConceptCard_noLinkSupport_clickSpan_callsTagListener` which is unexpected. Some other tests had slightly changed behaviors due to the activity being initialized later, but this one actually had a functional difference.
- Similarly, `testObserver_onFirstActivityResume_logsStartupLatency` in `ApplicationLifecycleObserverTest` had an event order change, but this seems less surprising since the change from rule to direct management is definitely likely to impact some arrangement ordering as they are not functionally equivalent.
- In a few cases, `ActivityScenarioRule` was being arranged but not used (or ignored since a later explicit `launch` was used, anyway).
- I renamed `runWithActivity` to `onActivityWithResult` in a bunch of places as a clearer, new pattern. The 'runWith' notion is really for the `ActivityScenario`, now, and the actual operation being performed is an `onActivity` with some sort of extraction (hence the 'withResult').
- Some extra work was needed in `MathExpressionAccessibilityUtilTest` to avoid majorly changing a lot of tests (since the activity was needed within a Truth assertion method). The change came out cleanly, though something like Kotlin context receivers would be even cleaner (or their new replacement).
- `ApplicationLifecycleObserverTest` had a bunch of module changes since it was migrated over to using a more appropriate test activity fixture (specifically `TestActivity`) that that required a lot more of the app to be included as part of the build.
- `ApplicationLifecycleObserver` was exempted from code coverage since it seems that the new modules being brought into the test are triggering some of the duplicate instrumentation issues that we see with a lot of app layer tests.
- Despite the length of the PR, much of these changes are space-only due to new indentations being added in a few hundred tests.

## Essential Checklist

## For UI-specific PRs only
N/A -- This only affects tests.


BenHenning (commented):
Did a quick self-review to catalog noteworthy changes.

BenHenning:
Looks like there may be some code coverage adjustment (likely due to the number of lines in tests increasing which may reduce the % covered). I will update the thresholds instead of actually changing coverage since this PR is not meant to impact test behaviors beyond the specific improvements of self-managing `ActivityScenario` over `ActivityScenarioRule`.

BenHenning:
PTAL @adhiamboperes or delegate as you see fit. I'm confident in this change based on how I performed the refactor, so it could be a good learning review.

Separately, apologies for taking one of the starter issues. Both `ActivityScenarioRule` and `ActivityTestRule` are likely to block #5725 :(.

adhiamboperes (approved):
Thanks @BenHenning! I do not have any comments, LGTM!

BenHenning:
Thanks @adhiamboperes! Going to go ahead and merge this, then.

Issue #3830, opened 2021-09-21
Introduce base general-purpose test activity & fragment [Blocked #1720]


BenHenning:
I couldn't find an issue tracking this already, so filing a new one.

We can avoid the pattern of needing to introduce custom test activities & fragments throughout the codebase by introducing a general-purpose test activity & fragment that can be trivially added to stage different UI scenarios.

BenHenning:
Note that while we can introduce a TestActivity today, it's a bit painful to use broadly since it'll require introducing a bunch of custom injectors throughout the codebase. We may want to wait until #1720 is introduced so that injectors can be automatically generated (via entrypoints) which will simplify stuff a lot.

BenHenning:
Note that #147 is probably also a blocker since most cases when ``TestActivity`` can't be used is due to needing the hosting activity to implement certain event listeners.

Issue #3602, opened 2021-07-28
[A11y] Add label for all non-test activities


rt4914:
[A11y] Add label for all non-test activities

[Reference](https://github.com/oppia/oppia-android/pull/3352/files/9c944564734be9aa46bf029cf14e6789ced95a7c#r672844473)

- [x] HomeActivity (Assigned to @vrajdesai78)
- [x] MyDownloadsActivity (Assigned to @bhaktideshmukh) 
- [x] ProfilePictureActivity
- [x] ProfileEditActivity(open)
- [x] SplashActivity (Assigned to @bkaur-bkj )
- [x] WalkthroughActivity (Assigned to @priyatanu)

Example PR: https://github.com/oppia/oppia-android/pull/3227/files

rt4914:
@vrajdesai78 Can you work on this issue? If yes, suggest picking one of the mentioned files and creating PR for that first.

vrajdesai78:
@rt4914 I want to work on this issue. I will create label for HomeActivity

deepanshu731:
@rt4914 I want to create label for ProfilePictureActivity.

vrajdesai78:
@rt4914 I am receiving static checks failed. Will I need to do anything else than adding label for HomeActivity in android manifest file and creating value in strings.xml

rt4914:
@vrajdesai78 I have commented on PR

bkaur-bkj:
@rt4914 sir I will make labels for SplashActivity pls assign that part to me.

bkaur-bkj:
@rt4914 PTAL

bkaur-bkj:
> @rt4914 sir PTAL

priyatanu:
@rt4914 I want to work on this WalkthroughActivity activity

rt4914:
@priyatanu Sure. Thanks.

priyatanu:
> @priyatanu Sure. Thanks.

Hi @rt4914 , @BenHenning , kindly review my PR - https://github.com/oppia/oppia-android/pull/3983

vivaansahai:
Hi, Can I work on `RecentlyPlayedActivity`?

ankita240796:
Hi @rt4914, can I pick `MyDownloadsActivity`? Thanks!

Rishavgupta12345:
can i work on RecentlyPlayedActivity

priyatanu:
Hi @rt4914 , Can I work on Profile edit Activity?

[...]

PR #5759, opened 2025-03-18
Fix #5308: Remove uses of ActivityTestRule.

### LINKED ISSUE #5308: [BUG]: ActivityTestRule is deprecated
### Describe the bug

ActivityTestRule has been deprecated in favor of ActivityScenarioRule.

### Steps To Reproduce

See documentation: https://developer.android.com/reference/androidx/test/rule/ActivityTestRule.

### Expected Behavior

While ActivityScenarioRule is preferable to ActivityTestRule, direct ActivityScenario management should be done, instead, per #5307.

### Screenshots/Videos

_No response_

N/A -- test environment

N/A -- test environment

N/A -- test environment

### Additional Context

Multiple contributors can work on replacing the ActivityTestRule with ActivityScenarioRule at the same time. Create small PRs changing tests in one package at a time. See comment below for the list of test classes to be fixed.

BenHenning:
## Explanation
Fixes #5308

Similar to #5755, this removes all uses of `ActivityTestRule`.

Broadly, this was done by:
- Replacing `ActivityTestRule` uses with equivalent 'launch' calls with `ActivityScenario`. This was doable in many places, and generally required little to no functional differences in the test (just some `onActivity` calls).
- Replacing all uses of `ActivityScenario` and `ActivityTestRule` with a 'runWith*' pattern similar to `ProfileAndDeviceIdFragmentTest`: https://github.com/oppia/oppia-android/blob/b68f993733cc9a9c830dc034c3fc9a649bd32bd5/app/src/sharedTest/java/org/oppia/android/app/administratorcontrols/learneranalytics/ProfileAndDeviceIdFragmentTest.kt#L932-L947 (also similar to what was done in #5755). Note that this was done in a few test suites where it had the most value (or where there was zero use of `ActivityScenario` already).
- Continuing the pattern of relying on `TestActivity` when testing fragments and views (where possible) rather than an actual activity since this better validates the fragment's API rather than the activity's behaviors. Callback listeners are verified with custom extensions of `TestActivity` that delegate their calls to mocks that can be verified using Mockito's utilities.

There were separate changes made or things to note:
- Some of the unnecessary casting to parent activities in fragments were removed when finishing the activity. This is something we should probably clean up broadly, but that wasn't done in this PR (generally, it was only done were necessary or where noticed when investigating test failures that might have been related).
- `TestActivity` was updated to include some helpers for registering test activities without requiring an `AndroidManifest.xml` registration. The issue is documented at https://github.com/robolectric/robolectric/pull/4736 and this will help us actually remove all of our test activities in the future (per #3830).
- Test injections and rule declarations were simplified and condensed for better readability.
- `LessonThumbnailImageViewTest` had its tests largely redone to avoid the need for a custom `LessonThumbnailImageViewTestActivity` (and test fragment). These classes aren't being removed in this PR, but they should be now obsolete.
- In some cases, the screen name tests were reorganized a bit to make them easier to read and to inline, where possible, "create intent" helpers since they were largely not needed and could be simplified (especially when launching using `ActivityScenario`).
- In most cases the 'runWith*' functions have a `testCoroutineDispatchers.runCurrent()` before running the test except for `TopicFragmentTest` and `ExplorationActivityTest` since some tests were highly sensitive to arranging specific state before running coroutines. In these cases the previous test behaviors were attempted to be retained with regards to background task and state synchronization.
- `setUpTestApplicationComponent` was moved below the tests where it was noticed as not already there for consistency with the vast majority of test suites.
- `RecentlyPlayedActivity` has a new test suite that has moved over tests from `RecentlyPlayedFragmentTest` since it was verifying activity-specific behavior. Test names have been renamed accordingly.
- `TopicFragmentTest` has some behavior differences in on the test fragment is being set up (large differences in that it's no longer hosted in `TopicActivity` for these tests). Some tests had to be set up to not pass a story ID in order to be made to work, but most were able to be moved over to that approach for a bit more setup consistency.
- Most of the changes in `ExplorationActivityTest` are spacing related, but due to the specific diffing behavior many of the lines are showing as re-written. This is likely due to `startPlayingNewExploration` being folded into a new "launch activity & start exploration" helper to reduce on the redundancy in a lot of the tests.

Finally, similar to #5755 this change is also a prerequisite to #5725.

## Essential Checklist

## For UI-specific PRs only
N/A -- This is only affected tests and should have no impact on user-facing flows.


app/src/main/java/org/oppia/android/app/testing/activity/TestActivity.kt
```diff
@@ -74,5 +87,33 @@ open class TestActivity : InjectableAutoLocalizedAppCompatActivity() {
...
+      }
+      registerActivityInfo(context, activityInfo)
+    }
+
+    /**
+     * Registers the specified [ActivityInfo] with [context]'s Robolectric shadow [PackageManager].
+     * Generally, this shouldn't be used directly; it's made accessible for the sake of
+     * [registerWithPackageManager].
+     */
+    fun registerActivityInfo(context: Context, activityInfo: ActivityInfo) {
```
BenHenning, line 112: Add tests for this & above.

BenHenning (commented):
Self-reviewed changes.

app/src/main/java/org/oppia/android/app/testing/activity/TestActivity.kt
BenHenning: Done in latest changes.

BenHenning (commented):
Reviewed merge (no conflicts) and new tests for `TestActivity`.

BenHenning:
PTAL @adhiamboperes or delegate as needed. :)

app/src/sharedTest/java/org/oppia/android/app/ongoingtopiclist/OngoingTopicListActivityTest.kt
adhiamboperes, line 1: This might be an instance where we want to move setUpTestApplicationComponent to the bottom of the test suite.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileChooserActivityTest.kt
```diff
@@ -124,10 +108,6 @@ class ProfileChooserActivityTest {
     Intents.release()
   }
 
-  private fun setUpTestApplicationComponent() {
-    ApplicationProvider.getApplicationContext<TestApplication>().inject(this)
-  }
-
```
adhiamboperes, line 130: This activity test doesn’t have a label test. Is that something we’d want to add in this PR?

adhiamboperes (approved):
Thanks @BenHenning! This is nicely done.

I don’t have any blocking comments, please feel free to resolve the 2 that I have left.

app/src/sharedTest/java/org/oppia/android/app/ongoingtopiclist/OngoingTopicListActivityTest.kt
BenHenning: Nice catch, will add in a follow-up commit.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileChooserActivityTest.kt
BenHenning: Ah good catch. Looks like this was missed in #3602. I think it's an easy thing to add, so I'll go ahead and do so in a follow-up commit.

app/src/sharedTest/java/org/oppia/android/app/ongoingtopiclist/OngoingTopicListActivityTest.kt
BenHenning: Verifying that this has been addressed in the latest changes.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileChooserActivityTest.kt
BenHenning: Verifying that this has been addressed in the latest changes.

BenHenning (commented):
Double-checked latest changes and verified that @adhiamboperes's two comments have been addressed.

BenHenning:
Thanks @adhiamboperes! I appreciate the review. :)

Enabling auto-merge since everything's addressed and CI checks should pass.

PR #5795, opened 2025-04-01
Fix #5793: Update workflows to Ubuntu 24.04

### LINKED ISSUE #5793: [Feature Request]: Update workflows to Ubuntu 24.04

Per https://github.com/actions/runner-images Ubuntu 20.04 is deprecated and we recently experienced a brownout since 20.04 will be removed on 15 April.

All workflows should be migrated to Ubuntu 24.04.

We could alternatively upgrade to 22.04, but it seems prudent just to move to latest now.

### Additional context

Note that this may not be a trivial task as environment changes can have significant impact on action setup and Bazel behavior.

Rd4dev:
Fixes #5793 

## Explanation

Updates workflows to use `ubuntu-24.04`

Took this to test run and see if it could help unblock other PRs. 

- ~~Updated the workflows to use the `ubuntu-latest` version rather setting a hard-coded version to stay updated with current versions and to minimize manual updates over time.~~ Edit: Updated to `ubuntu-24.04` as just realized that latest version might introduce breaking changes or future compatibility issues, while hard-coded version being well-tested would be more stable.
- Noticed that upgrading caused the runner to default to JDK 17, [ignoring the JDK 11 setup](https://github.com/oppia/oppia-android/actions/runs/14203094048/job/39795027843?pr=5795#step:7:31). Instead of forcing a fix, removed the JDK setups since Bazel now handles it remotely.
```
Java home: /usr/lib/jvm/temurin-17-jdk-amd64
openjdk version "17.0.14" 2025-01-21
OpenJDK Runtime Environment Temurin-17.0.14+7 (build 17.0.14+7)
OpenJDK 64-Bit Server VM Temurin-17.0.14+7 (build 17.0.14+7, mixed mode, sharing)
Error: Process completed with exit code 1.
```
- Used the matrix strategy for test builds to simplify the setup and reduce duplication—it's not directly related to this fix, but I’d been meaning to try it for a while, and this PR seemed like a good opportunity to do so.

## Essential Checklist


Rd4dev:
@BenHenning, I'm not sure about the expected checks — I believe they are set in the gating rules with old configurations, or is there something that needs to be fixed? I also tried upgrading to the `ubuntu-24.04` -- PTAL.

Also will revert the changes made to module files as they were just introduced to check run tests and run coverages, once the coverage report is generated to confirm it.

BenHenning:
Can you move back to 24.04 @Rd4dev? I don't think we want to run against Ubuntu latest (since it introduces potential breakages later on if GitHub introduces a new OS that creates CI incompatibilities).

Expected will depend on the OS type unfortunately, but I'll change that right before merging so that merging is unblocked (I can't do it sooner since it'll then cause all other PRs to become unmergeable).

Edit: Ah, I think you changed it back to 24.04 exactly when I submitted this comment. :)

Rd4dev:
>I don't think we want to run against Ubuntu latest (since it introduces potential breakages later on if GitHub introduces a new OS that creates CI incompatibilities).

Yeah, that just hit me - moved every workflow to utilize `24.04` now.

.github/actions/set-up-android-bazel-build-environment/action.yml
```diff
@@ -11,27 +10,11 @@ description: Sets up the Android & Bazel build environments for Oppia Android
...
-          exit 1
-        fi
-      shell: bash
-
     - name: Verify Bazel version
       run: |
         bazel --version | grep -q 6.5.0
         HAS_CORRECT_BAZEL_VERSION=$(echo $?)
-        if [[ "$HAS_CORRECT_JAVA_VERSION" == 1 ]] ; then
+        if [[ "$HAS_CORRECT_BAZEL_VERSION" == 1 ]] ; then
```
BenHenning, line 17: Wow nice catch--this apparently was always wrong before.

BenHenning (approved):
This is excellent, thanks @Rd4dev! I like the matrix change for the build flavors, too--I think that makes a lot of sense.

#4886 is when we changed Bazel to use a deterministic version of Java. I had tried removing the Java setup in CI (or upgrading it, I forget which) but it caused sdkmanager failures, so it's interesting to see that things are working correctly here. Either way, passing CI is passing CI. :)

Thanks much for the really fast turnaround after the issue was filed, as well!

.github/actions/set-up-android-bazel-build-environment/action.yml
BenHenning: (Resolving since there's nothing actually actionable to do here).

BenHenning:
Ah I also just realized that by removing the OS matrix from the build jobs we no longer will need to update the expected checks on future OS upgrades...at least for now. I mean to add OSX and Windows to those at some point. :)

BenHenning:
Going ahead and merging this since I had to change the expected checks which will cause other PRs to be unmergeable.

PR #5805, opened 2025-04-03
Fix part of #59: Introduce bundle modules [Blocked: #5804]

BenHenning:
## Explanation

Fixes part of #59

TODO: Finish. Explain the following modules being ignored, and the general strategy for why this is being done, limitations, and roughly how the bundles were decided (plus the strategy for implementing them):

Ignored for test sharing
  - ./domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterAlphaModule.kt (not used)
  - ./domain/src/main/java/org/oppia/android/domain/platformparameter/syncup/PlatformParameterSyncUpWorkerModule.kt (not used)
  - ./app/src/main/java/org/oppia/android/app/activity/ActivityModule.kt (not used)
  - ./app/src/main/java/org/oppia/android/app/application/dev/DeveloperBuildFlavorModule.kt (only used to test itself)
  - ./app/src/main/java/org/oppia/android/app/application/ga/GaBuildFlavorModule.kt (only used to test itself)
  - ./app/src/main/java/org/oppia/android/app/application/beta/BetaBuildFlavorModule.kt (only used to test itself)
  - ./app/src/main/java/org/oppia/android/app/application/alpha/AlphaBuildFlavorModule.kt (only used to test itself)
  - ./app/src/main/java/org/oppia/android/app/application/ApplicationModule.kt (to be replaced by a general purpose TestModule)
  - ./app/src/main/java/org/oppia/android/app/translation/ActivityRecreatorProdModule.kt (unused)
  - ./app/src/main/java/org/oppia/android/app/fragment/FragmentComponentBuilderModule.kt (unused)
  - ./utility/src/main/java/org/oppia/android/util/accessibility/AccessibilityProdModule.kt (unused)
  - ./utility/src/main/java/org/oppia/android/util/caching/testing/AssetTestNoOpModule.kt (only used when necessary)
  - ./utility/src/main/java/org/oppia/android/util/caching/CachingModule.kt (not used, prefer to use CachingTestModule instead)

## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [ ] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
[...]

PR #5804, opened 2025-04-03
Fix part of #59: Consolidate to common test modules [Blocked: #5823]

BenHenning:
## Explanation

Fixes part of #59

TODO: Finish.

Document the following module updates & explain why this PR exists:

Ignored but requires test updates (round 1):
  - ./domain/src/main/java/org/oppia/android/domain/onboarding/ExpirationMetaDataRetrieverProdModule.kt (prefer using ExpirationMetaDataRetrieverTestModule)
  - ./utility/src/main/java/org/oppia/android/util/networking/NetworkConnectionUtilProdModule.kt (prefer to use debug unless not possible)
  - ./utility/src/main/java/org/oppia/android/util/logging/SyncStatusProdModule.kt (prefer to use test unless not possible)
  - ./domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterProdModule.kt (prefer to use PlatformParameterTestModule unless not possible)
  - ./utility/src/main/java/org/oppia/android/util/parser/image/ImageLoaderProdModule.kt (prefer to use ImageLoaderTestModule, instead)
  - ./utility/src/main/java/org/oppia/android/util/locale/LocaleProdModule.kt (prefer to use LocaleTestModule instead)
  - ./domain/src/main/java/org/oppia/android/domain/hintsandsolution/HintsAndSolutionDebugModule.kt (prefer using prod module)
  - ./domain/src/main/java/org/oppia/android/domain/exploration/ExplorationStorageProdModule.kt (prefer to use ExplorationStorageTestModule)
  - ./utility/src/main/java/org/oppia/android/util/threading/DispatcherProdModule.kt (prefer to use DispatcherTestModule)
  - ./data/src/main/java/org/oppia/android/data/backends/gae/NetworkConfigProdModule.kt (prefer to use NetworkConfigTestModule instead).
  - ./utility/src/main/java/org/oppia/android/util/logging/firebase/LogReportingDebugModule.kt (prefer to use LogReportingTestModule unless not possible)
  - ./utility/src/main/java/org/oppia/android/util/logging/firebase/LogReportingProdModule.kt (prefer to use LogReportingTestModule)
  - ./utility/src/main/java/org/oppia/android/util/system/OppiaClockModule.kt (prefer using Fake)
  - ./utility/src/main/java/org/oppia/android/util/logging/performancemetrics/PerformanceMetricsAssessorProdModule.kt (prefer using PerformanceMetricsAssessorTestModule)

## Essential Checklist

[...]

Issue #5333, opened 2024-02-06
Android App - User Feedback Reporting


rolandbellerud:
### Is your feature request related to a problem? Please describe.

To effectively deliver to the needs of our learners, we will need to understand what issues or concerns arise with the app, what suggestions learners might have to improve the app, and what is working well for them. This feedback guides improvements of the learner experience and can empower positive change in the app — even (and especially) when it's negative.

### Describe the solution you'd like

Please see "Summary of the proposed solution" in [this PRD](https://docs.google.com/document/d/1ssV3eh0RtAR7YlatxdQoLG2KOJ-Qri33trYJr0AIw-o/edit#heading=h.mremwtv0zllq).

See also the mocks [here](https://www.figma.com/file/d2zD9RGnIbnX8q0uOKYNVD/Final---App-feedback?type=design&node-id=0%3A1&mode=design&t=aVm0CDccpkfD1yJn-1).

### Describe alternatives you've considered

There is currently no option to leave feedback directly in the Oppia Android app. The current (beta) version of the app provides a CTA when opening the app which prompts users to share issues with Oppia via email, but there is no direct path to submit feedback within the app. 

### Additional context

Please see https://docs.google.com/document/d/1ssV3eh0RtAR7YlatxdQoLG2KOJ-Qri33trYJr0AIw-o/edit

seanlip:
Note: For design clarifications, please contact @jiwonc26. (For reference, the original design thread is [here](https://github.com/oppia/design-team/issues/57).)

Issue #2951, opened 2021-03-19
Retrofit needs to have proper threading controls


BenHenning:
Retrofit needs:
1. To use our threading dispatchers
2. Needs to have proper adaption with DataProvider (preferably in a way where no one ever interacts with Retrofit directly to avoid accidentally introducing synchronous calls)

BenHenning:
I dug into this a bit more last night and the findings are mixed.

**Good news**
Retrofit apparently leans on OkHttp for asynchronous threading management, but does perform callback invocation in background threads (which is configurable by passing in a custom ExecutorService). OkHttp lets us configure the threads used for asynchronous processing by passing in a custom dispatcher class (which uses ExecutorService under the hood). Fortunately, we also leverage ExecutorServices as the base concurrency primitives for all of our coroutine dispatchers making it easy to centralize our threading across OkHttp, Retrofit, and the rest of the app (only Glide won't be part of this since they intentionally use their own fine-tuned ExecutorServices).

**Not-so-good news**
We expect all threading controls to go through Kotlin coroutines which means we have no way currently to synchronize threading across the Java world with ExecutorServices and the Kotlin world with coroutines. There are two options to fix this:
1. Introduce a means to use coroutines in Java (e.g. make our existing CoroutineExecutorService used for Glide interop production-ready)
2. Change our test synchronization mechanism to instead synchronize on ExecutorService and allow the rest of the app to have access to the underlying ExecutorServices used in the coroutine dispatchers

(1) seems particularly scary since coroutines are difficult to make block, and ExecutorService has blocking APIs that must be implemented (see warning at the top of [CoroutineExecutorService](https://github.com/oppia/oppia-android/blob/1be8ccde14c25d763c10dd8902ec7f5a2fa6f99e/testing/src/main/java/org/oppia/android/testing/CoroutineExecutorService.kt#L56) & the implementation for a lot more details on this). This has resulted in very challenging situations that aren't obvious to fix. Moving toward an ExecutorService-based synchronization model will make a lot more sense (since then we're coordinating all threading to a single ExecutorService rather than via a wrapper ExecutorService through corotouines to an ExecutorService). This will also nicely let us replace the existing CoroutineExecutorService when testing Glide.

While (2) seems nice, it has some drawbacks:
- It exposes two ways to perform concurrency in the app. So long as we encourage developers to continue using DataProviders then they will automatically be forced to use Kotlin coroutines which are preferred due to their lightweight performance and better concurrency management.
[...]

PR #5806, opened 2025-04-05
Fix #77, part of #59: Revise testability of Retrofit services

### LINKED ISSUE #77: Testing Strategy for GAE Abstraction [Blocked #76]
Implement the testing strategy for all the retrofit interfaces. 
Fake versions of each retrofit interface will be introduced which can check the corresponding interface. Fakes should be able to check the retrypolicy, XSSI_PREFIX removal and correct deserialisation of response.
This is blocked on #5 , #76

BenHenning:
## Explanation

Fixes #77
Fixes part of #59 (as part of an ongoing PR chain, see #5804)

This PR does two batches of work:
1. It splits `NetworkModule` into distinct modules: `RetrofitModule` and `RetrofitServiceModule`.
2. It replaces `MockPlatformParameterService`, `MockFeedbackReportingService`, and `MockRetrofit` with orchestrator alternatives that leverage OkHttp's `MockWebServer`, instead.

(1) is needed for the ongoing PR chain to introduce bundle modules and consolidate on common test modules and configurations (leading up to #5804 and #5805). It was discovered that decoupling the different dependencies for OkHttp and Retrofit was a bit too cumbersome to be practical in those PRs, so this one was introduced, instead.

As part of (1), it became clear that managing what appeared to be two distinct means of testing HTTP endpoints (a mock Retrofit service and `MockWebServer`) was tricky and inconsistent, so this PR pushes everything over to just using `MockWebServer` as that provides the best integration pathway (it ensures that the generated Retrofit services and corresponding Moshi data objects are actually set up correctly, and that calling code into OkHttp is correct--which is especially important for resolving #2951).

From a design perspective, the new orchestrators are meant to simulate the Oppia web backend by fixing specific responses with an assumed deterministic order to depending on them. `PlatformParameterIntegrationTest` does a good job showing this working all the way to a UI that depends on the parameters. There's some repetition between the two orchestrators that might be worth solving with a common base class in the future (or leveraging some sort of composition or mixin pattern), but this seems like a reasonable start. The orchestrators were designed to keep relative parity with the mock services that they are essentially replacing.

Some things to note:
- `SplashTestActivity` was updated to use `TestActivity` and self-register itself (which means that it could be removed from the main app manifest & its accessibility exemption removed). I didn't actually intend to use programmatic registration, but at first I didn't understand where this failure was coming from. I eventually discovered that the package manager setup in `NetworkConfigTestModule` was resetting already registered activities and later fixed this (plus added a test for it), but I decided to keep the self-registration for `SplashTestActivity` since that's in-line with long-term test activity goals.
- There are some `post()` exemptions added for the new tests--this is an unfortunate side effect of using simple regex to detect bad patterns. These aren't actually the same `post()` calls that we want to try and prevent, but the exemptions are needed regardless.
- `NetworkApiKey` was moved to `network_config_annotations` since it's an annotation (it probably should have been in that library originally).
- The Dagger `NetworkApiKey` provider is now part of `NetworkConfigProdModule` so that its counterpart `NetworkConfigTestModule` can provide a non-prod version of the key (and ditto for the instrumentation app).
- `Moshi` is now being injected as well since it has additional configurations (due to the introduction of `GaePlatformParameterValue`). This is mainly for the benefit of tests, but it wouldn't be harmful if other parts of the codebase depended on this object if they needed to. We may want to qualify it to be GAE-specific in the future.
- `ApiMockLoader` was removed in favor of just loading mocks directly (since Gradle is no longer supported). Similarly, due to the mock services being removed everything under `testing/src/main/java/org/oppia/android/testing/network` was able to be removed (which is a nice win to move things out of the top-level testing directory).
- Note that the change to load mocks directly required some small differences in asset management, i.e. an introduction of a test-only assets library for the data layer.
- Some field names in `GaeUserSuppliedFeedback` were updated to match their mock JSON counterparts (these aren't actually used yet since feedback reporting hasn't been completed--see #5333).
- `GaePlatformParameterValue` was introduced to simplify parameter deserialization (since Moshi can handle this quite cleanly with a custom `JsonAdapter` rather than needing to declare an `Any` type and risk bugs or crashes at the time of using the data). A test was added for this class to help establish the pattern of testing Moshi objects moving forward.
- There's some extra Robolectric initialization happening in `NetworkConfigTestModule` to simplify test arrangement (which can be seen several times in the PR when similar-looking package manager setup code has been removed). This may not end up being the permanent home for this, but it's a very reasonable place for it to live for now.
- There's some risk in the time-based orchestrator tests. These didn't have a better option for now (at least until #2951 is solved) since testing for a disconnect basically requires having a short response timeout, and the request is always blocking since OkHttp isn't yet configured to be thread cooperative with the rest of the app.
- Some `MockRetrofit` comments were removed in tests since they weren't actually using `MockRetrofit` (as intended, and as we want long-term). Note that `ApplicationLifecycleObserverTest` _was_ updated to use the new `NetworkConfigTestModule` (in-line with this idea of centralizing test modules).

## Essential Checklist

## For UI-specific PRs only
N/A -- This should only affect testing infrastructure and test behaviors.


data/src/test/java/org/oppia/android/data/backends/gae/RetrofitModuleTest.kt
```diff
@@ -266,17 +236,17 @@ class NetworkModuleTest {
...
   class TestApplication : Application() {
     private val component: TestApplicationComponent by lazy {
-      DaggerNetworkModuleTest_TestApplicationComponent.builder()
+      DaggerRetrofitModuleTest_TestApplicationComponent.builder()
         .setApplication(this)
         .build()
     }
 
-    fun inject(networkModuleTest: NetworkModuleTest) {
+    fun inject(networkModuleTest: RetrofitModuleTest) {
```
BenHenning: Fix the inject name here & above, and in the other new module test file.

BenHenning: This was fixed in the latest changes.

app/src/sharedTest/java/org/oppia/android/app/help/HelpActivityTest.kt
```diff
@@ -165,7 +166,8 @@ class HelpActivityTest {
       HintsAndSolutionConfigModule::class, HintsAndSolutionProdModule::class,
       FirebaseLogUploaderModule::class, FakeOppiaClockModule::class,
       DeveloperOptionsStarterModule::class, DeveloperOptionsModule::class,
-      NetworkModule::class, ExplorationStorageModule::class, NetworkConfigProdModule::class,
+      RetrofitModule::class, RetrofitServiceModule::class, ExplorationStorageModule::class,
```
BenHenning, line 168: TODO: Remove extra duplication.

build_flavors.bzl
```diff
@@ -34,61 +34,61 @@ _PRODUCTION_PROGUARD_SPECS = [
...
 # Note to developers: keys of this dict should follow the order of AVAILABLE_FLAVORS.
 _FLAVOR_METADATA = {
-    "dev": {
-        "manifest": "//app:src/main/AndroidManifest.xml",
-        "min_sdk_version": 21,
-        "target_sdk_version": 34,
-        "multidex": "native",
-        "proguard_specs": [],  # Developer builds are not optimized.
-        "production_release": False,
+    "alpha": {
```
BenHenning: TODO: Revert this file.

domain/src/main/java/org/oppia/android/domain/platformparameter/syncup/PlatformParameterSyncUpWorker.kt
```diff
@@ -109,10 +114,13 @@ class PlatformParameterSyncUpWorker private constructor(
         }
         Result.success()
       } else {
+        println("@@@@@ WAS NULL")
         oppiaLogger.e(TAG, "Failed to fetch platform parameters (no network stack available)")
         Result.failure()
       }
     } catch (e: Exception) {
+      println("@@@@@ CAUGHT EXCEPTION")
+      e.printStackTrace()
```
BenHenning: Remove this & other debug lines.

oppia_android_application.bzl
```diff
@@ -262,11 +262,11 @@ _restrict_languages_in_raw_module_zip = rule(
 
 _bundle_module_zip_into_deployable_aab = rule(
     attrs = {
-        "input_file": attr.label(
+        "config_file": attr.label(
```
BenHenning: Revert this file.

BenHenning (commented):
Self-review whole PR. There are some code comments I noticed separately that I also want to address.

oppia_android_application.bzl
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/syncup/PlatformParameterSyncUpWorker.kt
BenHenning: Done in latest.

build_flavors.bzl
BenHenning: Done in latest.

app/src/sharedTest/java/org/oppia/android/app/help/HelpActivityTest.kt
BenHenning: Done in latest.

BenHenning (commented):
Self-reviewed latest changes, noted specifics, and confirmed that previous self-review comments have been addressed.

BenHenning:
PTAL @adhiamboperes or delegate as you see fit. I don't think this is a particularly controversial PR, and most of it is rather routine. It is one of several large-scale module refactors coming, however, with the actual long-term goal of eliminating most of those codebase-wide changes that we see with modules (that's the goal, anyway).

adhiamboperes:
@theayushyadav11, PTAL at the `app/test` files as discussed.

app/src/test/java/org/oppia/android/app/testing/PlatformParameterIntegrationTest.kt
```diff
@@ -5,8 +5,6 @@ import android.content.Context
 import androidx.appcompat.app.AppCompatActivity
```
theayushyadav11, line 5: This file contains other changes as well.

app/src/test/java/org/oppia/android/app/translation/BUILD.bazel
```diff
@@ -57,7 +57,7 @@ oppia_android_test(
         "//app/src/main/java/org/oppia/android/app/application/testing:testing_build_flavor_module",
         "//app/src/main/java/org/oppia/android/app/translation/testing:test_module",
         "//config/src/java/org/oppia/android/config:all_languages_config",
-        "//data/src/main/java/org/oppia/android/data/backends/gae:prod_module",
+        "//data/src/main/java/org/oppia/android/data/backends/gae:retrofit_module",
```
theayushyadav11, line 60: This file just changes the dependency name.

theayushyadav11 (commented):
Hi @adhiamboperes , I took a look at all the files under `app/test` and all the files have same changes as expected (ie replacing `NetworkModule` with `RetrofitModule` and `RetrofitServiceModule` . Except the file `PlatformParameterIntegrationTest.kt` which have other changes as well. PTAL!

subhajitxyz:
Hi @adhiamboperes , I've reviewed all the files under packages A–O (from administratorcontrols to options). Everything looks good to me.

app/src/test/java/org/oppia/android/app/translation/BUILD.bazel
subhajitxyz: Good observation! Please check that prod_module has been replaced with retrofit_module in data/src/main/java/org/oppia/android/data/backends/gae/BUILD.bazel

manas-yu:
Hi @adhiamboperes,
I've completed a review of all files under the packages  `P–W (parser–walkthrough)`  within `app/sharedTest`, and everything looks good to me. 

Files include the following updates:
- **Test Files:** `NetworkModule` has been replaced with `RetrofitModule` and `RetrofitServiceModule`, including the corresponding import changes.
- **Build Files:** `prod_module` has been replaced with `retrofit_module`.

adhiamboperes:
Pending review for ~13 files -- should finish up later today.

app/src/test/java/org/oppia/android/app/translation/BUILD.bazel
adhiamboperes: That's right. It is intentional, therefore not a cause for concern. Thanks for flagging!

app/src/test/java/org/oppia/android/app/testing/PlatformParameterIntegrationTest.kt
adhiamboperes: Verified the formatting changes and injection of the orchestrator (for configuring [MockWebServer] to properly respond to requests via PlatformParameterService)

adhiamboperes (approved):
Thanks @BenHenning! changes LGTM! 

Thanks @manas-yu, @theayushyadav11, @subhajitxyz for additional verifications.

adhiamboperes:
@seanlip, PTAL at the codeowners file.

seanlip (approved):

seanlip:
@adhiamboperes LGTM. Should this be merged?

adhiamboperes:
@seanlip yes, I’m going ahead and merging it.

BenHenning:
Thanks everyone for the reviews! :) Will have more PRs coming soonish.

Issue #5812, opened 2025-04-19
[BUG]: Fix coverage script missing incorrectly placed tests on incremental runs


BenHenning:
### Describe the bug

https://github.com/oppia/oppia-android/issues/5780#issuecomment-2752627473 explains the specifics in detail, but essentially it's possible for the code coverage script to incorrectly miss files whose tests are either missing or in the wrong directory.

### Steps To Reproduce

Either move a test to the wrong directory (like the one mentioned in #5780) or find one, and modify it to have a change. Then, run the incremental code coverage process and verify that the test wasn't run.

### Expected Behavior

Files with tests that aren't exempted and essentially missing (due to being in the wrong test directory) should trigger an incremental code coverage failure so that it's impossible to check in a broken test as happened in #5780.

### Additional Context

#5780 has significant context on where this issue is found, and possible mitigations.

Issue #3436, opened 2021-07-07
Develop a mechanism for permanently exempting files which do not ever need tests


Sparsh1212:
Currently, we are just checking that for every `.kt` file, a corresponding Filename+"Test" file must be present.
As per this, we also have to add those files to the exemptions list which do not ever need to have any tests.
For example: Currently, we have to add:
https://github.com/oppia/oppia-android/blob/develop/scripts/src/java/org/oppia/android/scripts/common/CommandResult.kt
manually to the exemption list. However, it should ideally be automatically be exempted by the script for the check.

**Criteria to be permanently exempted for the check**: Kotlin files that do not have any interface, annotation, function, or class

We need to develop a proper mechanism to permanently exempt those files which lie in the above criteria.

BenHenning:
@Sparsh1212 if you were to implement this, what would your approach be for the mechanism?

Sparsh1212:
@BenHenning 
This is the approach that comes to my mind:

We first have to parse the file to a `KtFile`. This allows us to iterate over the AST elements in an easy manner.
Once, we have got the `ktFile` for this file, we will iterate over each and every element to check if it belongs to any of:
`KtClass`, `KtNamedFunction`, and similar equivalents for other members. If every element, does not match any of those, then it's guaranteed that the file does not contain any class, function, interface, or annotation. 

Hence, we can skip this file for the check and will not include it in the `prodFilesList`.

Issue #2532, opened 2021-01-21
Move all tests under app layer to test/


BenHenning:
We have local tests under the app module test folder that only run on Robolectric. Now that we have the RunOn annotation, we should prefer keeping all tests under app module's sharedTest folder & instead use the RunOn animation. This issue is tracking:
- Moving all app module tests from the test folder to sharedTest counterparts
- Annotating the tests OR test class with RunOn(ROBOLECTRIC)
- Removing the Gradle/Bazel code around the app module test folder

Update 19 Apr 2025: We should actually isolate to a common test/ folder rather than sharedTest/ and remove sharedTest/ as there's no value to differentiate in the post-Gradle world.

dipu989:
@BenHenning Would like to work upon this! Can I self assign?

rt4914:
> @BenHenning Would like to work upon this! Can I self assign?

@dipu989 Yes you can self-assign.

dipu989:
@rt4914 @BenHenning 
How I am thinking about going with this issue is - 
Viewing the Project structure, we have sharedTest and test directories. 
I have to move the tests from "test" directory to the "sharedTest" directory in their respective counterparts and refactor if needed.

I have few doubts here - 
1) I need to add @RunOn(ROBOELECTRIC) annotation in every test.kt file which are already using @RunWith(AndroidJunit4::class). Do these Junit annotations need to be replaced with Roboelectric ones above every class declaration? 

2) Can you elaborate a bit more on the point
   'Removing the Gradle/Bazel code around the app module test folder'.

BenHenning:
@dipu989 good questions. I hope this will help clarify:

1. RunOn and RunWith are independent of one another, so we may need both in some cases (but we always need RunWith)
2. I looked into this a bit more--it seems that nothing needs to change in Gradle, but our Bazel file will have a bunch of test references that need to be updated once the tests are moved ([here](https://github.com/oppia/oppia-android/blob/develop/app/BUILD.bazel#L722) is one example)

anandwana001:
@dipu989 any update on this?

dipu989:
@anandwana001 I might get this done by the beginning of the next week. Sorry for blocking this for quite a long time.

dipu989:
@BenHenning @anandwana001 Local tests are passing but the instrumentation ones are giving this error. Any pointers to resolve this? Running on Realme device as well as emulator gives this error.

[...]

PR #5814, opened 2025-04-22
Fix #5780: Fix develop CI and improve test file check robustness

### LINKED ISSUE #5780: [BUG]: CI is currently failing for code coverage
### Describe the bug

Code coverage is failing on develop with an indicator that the test for `RecentlyPlayedActivity` is currently missing:

```
  COVERAGE FAILURE REPORT:
  -----------------------
  Coverage Report Failure:
  ------------------------
  Test Target: 
  Failure Message: No appropriate test file found for app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedActivity.kt.
```

This was first introduced in #5759 due to a new test being added: `RecentlyPlayedActivityTest`. What's odd is that this was successfully checked in meaning that code coverage analysis passed incrementally, and the test file exemption check didn't trigger. It's unclear why there's a behavior discrepancy, but notably the test was put in the wrong package (it should be under `home/recentlyplayed` not `home/`). I would, however, have expected it to fail for both incremental and non-incremental checks.

### Steps To Reproduce

Either try running code coverage for the specific `RecentlyPlayedActivity` or run the whole suite (I haven't verified either repro case).

### Expected Behavior

Since there's a valid test, this issue should not be triggered.

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

We need to understand why this failed in CI before we can be confident in a fix.

BenHenning:
## Explanation
Fixes #5780

#5780 was ultimately caused by two things:
1. An issue in how the code coverage utility propagates certain errors for incremental runs.
2. Specific repository issues that can cause failures.

This PR addresses (2) which mitigates (1) in its entirety (for this particular class of issues, that is, incorrectly placed and/or named tests). (1) is to be addressed in #5812.

In order to ensure this problem was fully addressed, `TestFileCheck` has been completely reworked to have a much stricter verification of production and test files. Specifically, it:
- Now knows the production & test directory layouts for each top-level layer package.
- Checks _both_ production and test files to ensure they have corresponding matches.
- Introduces an exemption for tests which should exist but not match their production file names (such as extra tests). We may want to consolidate these in the future, but it's not pressing.
- Checks that exemptions actually match real files--a bunch of the existing exemptions point to non-existent files.
- Presents "could be" suggestions when it encounters a test or production file that might have a match which is in the wrong directory.

Some specific notes:
- The new matching approach has led to a _lot_ of files that have been moved or otherwise fixed:
  - A bunch of admin control, dev options, and help tests weren't in as specific of directories as their production files (or were in the wrong higher-level package).
  - `FaqListActivityTest` didn't match the name of its production file and thus needed to be renamed.
  - `ImageRegionSelectionInteractionViewTest`'s new location is modularized, so it needed a new test declaration. Note that this was largely based on `StateFragmentTest`'s test declaration, so it likely contains redundant dependencies (I didn't check them thoroughly). There's tooling coming in the modularization work to clean these up, so I think it's probably fine.
  - #5806 set the wrong packages for some of the new orchestrator classes (due to the Bazel plugin not knowing where these files are supposed to go), and this was unfortunately missed (yet still built). This has been corrected and the `TestFileCheck` changes will act as a guard against these situations in the future.
  - Some tests had an extra 'rule' in them: `DragDropSortInputHasElementXAtPositionYRuleClassifierProviderTest` and `DragDropSortInputHasElementXBeforeElementYRuleClassifierProviderTest`.
  - `SurveyEventsLoggerTest` and its test target have been moved to the correct `oppialogger` subpackage (which is new).
  - A number of `app` and `domain` tests were moved to the utility layer:
    - `FakeAccessibilityServiceTest`
    - `StringExtensionsTest`
    - Most (all?) of the HTML parsing tests: `CocneptCardTagHandlerTest`, `CustomHtmlContentHandlerTest`, `HtmlParserTest`, `ImageTagHandlerTest`, `ListItemLeadingMarginSpanTest`, `LiTagHandlerTest`, `MathTagHandlerTest`, and `PolicyPageTagHandlerTest`.
    - `PlatformParameterSingletonTest`.
  - `HtmlParserTest` and `ListItemLeadingMarginSpanTest` both depended on bespoke test activities, and trying to make `utility` tests depend on these was challenging. Instead, these have been moved over to using `TestActivity` (for which there's some precedent in non-`app` tests) and the bespoke activities have been removed. This helps with the effort of #3830.
- Instrumentation tests currently need to be marked as 'extra' tests. Fixing this properly will probably require addressing #3436 (that is, having proper mechanisms for detecting permanent exemptions though the mechanism may be different than that issue in particular).
- There are some Java-specific bits built in that can be removed when #1639 is resolved.
- #2532 will simplify some of the extra test directory management. We may want to take this further in the future and remove all top-level layer directories in favor of just UI, service, and scripts. At that point we could also move away from the Maven directory structure, but more thought is needed on that.
- TODO exemptions needed to be updated to wiki page adjustments.
- `UserAppHistoryViewModel` was a file that was introduced a long time ago, but has always been empty. I've been meaning to remove it for a while, but now we finally had a good reason: the new script requires `package` statements in Java and Kotlin files in order to verify that the file was in its correct location (relative to its test/production file), and since this file is empty it has no `package`. It's now been removed.
- `TestFileCheck`'s tests have been completely rewritten. Note that they now focus on testing the utility in isolation rather than using the real exemption script because the new check of verifying that the files exist present a prohibitively difficult restriction on testing `main` directly (i.e. by requiring all these files exist).
- The valid exemption cases for missing tests have been clarified in the Static Analysis Checks wiki page.
- `RecentlyPlayedActivityTest` has been exempted as it was missed originally (which was what ultimately caused #5780) since it is currently incompatible with code coverage.

I recommend looking at the test changes for `TestFileCheckTest` to see the new check output in different failure cases since it better demonstrates what failures this check can now find and report.

This is being verified by both running all tests (which simulates being merged into `develop` and an incremental run after the complete run finishes), specifically:
- Full check: https://github.com/oppia/oppia-android/pull/5814/checks?sha=040baf1e0c72368a03d8123aaf0b4c346e5f297a.
- Incremental check: https://github.com/oppia/oppia-android/pull/5814/checks?sha=67d5f905ed0e67ff4428fc8948c7236fe234d6b7.

## Essential Checklist

## For UI-specific PRs only
N/A -- Infrastructure and refactor changes only.


scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
```diff
@@ -9,7 +9,7 @@ import java.io.FileInputStream
  * Script for ensuring that all production files have test files present.
  *
  * Usage:
- *   bazel run //scripts:test_file_check -- <path_to_directory_root>
+ *   bazel run //scripts:test_file_check -- <path_to_directory_root> <path_to_exemptions>
```
BenHenning: This is wrong--I never ended up changing the arguments.

BenHenning (commented):
Self-reviewed changes.

BenHenning:
The full run was a success as it didn't run into the same failure as develop. There's one test that requires an exemption which will be added in a subsequent commit.

scripts/src/java/org/oppia/android/scripts/testfile/TestFileCheck.kt
BenHenning: Fixed in latest.

BenHenning (commented):
Self-reviewed latest. I also manually checked locally that `RecentlyPlayedActivity` is properly exempted now, as well.

BenHenning:
I expect this to pass now, so I think it can be reviewed.

PTAL @adhiamboperes or delegate as needed. This is pretty important as it should stop develop CI from failing.

seanlip (approved):
LGTM for owners (thanks @BenHenning)!

wiki/Static-Analysis-Checks.md
```diff
@@ -138,8 +141,10 @@ exempted_file_path: "app/src/main/java/org/oppia/android/app/activity/ActivityCo
 2. Add an explanation to your PR description detailing why this exemption is correct.
 
 Following are the cases where its valid to have test file exemptions:
-1. Interface files
-2. Files with only constants defined (no logic)
+1. Interface files.
+2. Files with only constants defined (no logic).
+3. Cases when tests fail to pass code coverage.
+4. Cases when a test is verifying additional functionality beyond what can be tested in its usual test suite (for these cases use the 'is_extra_test' property in the exemption structure).
```
adhiamboperes: It might be worth adding an example here as well. It took me a while to understand why the math parser tests were exempted.

adhiamboperes (changes requested):
Thanks @BenHenning! This LGTM, just had one comment.

I am happy to see these improvements - especially catching exemptions that don't correspond to actual files, and improved specificity of the failure messages.

wiki/Static-Analysis-Checks.md
BenHenning: Good point--added an example and made some other small clean-ups in this file.

BenHenning (commented):
Thanks @seanlip & @adhiamboperes!

I addressed your comment @adhiamboperes, PTAL.

Also, self-reviewed the latest changes.

BenHenning:
Odd CI flake--restarting everything.

Also, I'm enabling auto-merge since I think the main comment to verify should be non-controversial, but I'm happy to make additional changes as needed. :)

BenHenning:
Ah, unfortunate. TODOs need to be regenerated for the updated wiki page.

adhiamboperes (approved):
LGTM, thanks!

BenHenning:
Thanks @adhiamboperes! Just waiting on a passing CI now.

BenHenning (commented):
Self-reviewed TODO exemptions update due to wiki page changes.

BenHenning:
GitHub is having a LOT of spurious failures today. Most of these are coming from trying to download `Buf` from a GitHub release URL.

Issue #89, opened 2019-09-02
Properly handle background thread processing in AndroidX tests


BenHenning:
tl;dr: The project needs a way to consistently test operations which require background processing in a way that's consistent with production environments,

UI-centric AndroidX tests can run on both Robolectric and Espresso. Both Robolectric and Espresso have their own ways to block on background tasks completing.

Robolectric running in legacy mode can deterministically advance the system clock and looper to execute pending tasks. This, combined with Kotlin's experimental [runBlockingTest](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-blocking-test.html) provides a nice way to at least test background tasks that are blocking on coroutines. However, this isn't fool-proof, and has its own issues (e.g. best practice is to use [InstantTaskExecutorRule](https://developer.android.com/reference/android/arch/core/executor/testing/InstantTaskExecutorRule) which actually forces synchronous execution).

Espresso, on the other hand, has the choice of [idling resources](https://developer.android.com/training/testing/espresso/idling-resource). These allow the app to communicate to Espresso when a non-UI bound task has completed which pairs well with Espresso's internal idling mechanism. This helps improve test determinism for Espresso.

Having two solutions isn't viable for shared AndroidX tests, and Robolectric's approach results in synchronizing background processing which reduces the chance of discovering race conditions. Instead, Oppia Android should have a way to run background tasks similar to production (i.e. by not synchronizing them or forcing them to run on the main thread), and ensure both testing frameworks can consistently wait for completion (probably by using idling resource since Robolectric should support them, or using either idling resource or a Robolectric-specific solution). This should also be done in conjunction with running Robolectric in a [paused looper mode](http://robolectric.org/blog/2019/06/04/paused-looper/) to have better parity with Android and try to detect real timing regressions that could occur in production.

Edit: Jetpack Compose also needs to be considered (see comment below).

BenHenning:
This is actually blocked on #59 since this partially requires support for testing libraries in order to have shared testing infrastructure.

BenHenning:
NB: https://github.com/Kotlin/kotlinx.coroutines/issues/242 might be an interesting solution to this if it becomes available.

BenHenning:
NB: this probably blocks exploration fragment tests since the exploration takes a while to load, and we need Espresso to properly idle until all the background processing is complete.

BenHenning:
[...]

PR #5825, opened 2025-05-23
Fix Part of #4938: Redesign Profile login and PIN Reset(1/12)

adhiamboperes:
## Explanation
Fixes Part of #4938 
This PR introduces the following changes:

- redesigns the profile login screen 
- ensures that profiles with pin will login via this new screen when the onboardingv2 feature flag is enabled
- handles PIN reset when the user has forgotten their PIN
- expands support for Jetpack Compose
- adds tests for all these changes

### Use of Jetpack Compose
We agreed that we can use Jetpack Compose for new screens being introduced, complete with its own activity and fragment. Although there is an existing login screen, the UI and flows are a little different from the new one, so I opted to create an new screen. The legacy screen has TODOs added to earmark it for removal.

The PR introduces Compose UIs for the screen, as well as the admin pin reset dialogs, since these are alert dialogs. The non-admin pin reset dialog reuses the existing dialogs, since they are standalone dialog fragments.

The Composable functions are private because they are not being tested directly, like you would in in a purely Compose project. The two admin PIN reset dialogs are however in their own files because it made sense to isolate them at that level, being that they are independent components.

Finally, I added the `"androidx.compose.runtime:runtime-livedata": "1.1.1",` dependency to be able to observe livedata objects from within composables.

### Test Design
Because the screen contains both Compose and non-compose components, some of the tests use both Compose nodes and Espresso Matchers and ViewInteractions. I am pleased with how well the dialogs tested.

I used `createEmptyComposeRule()` as the test rule because I found out previously in https://github.com/oppia/oppia-android/pull/5387 that it is the most appropriate rule to use when there is need to launch a new activity for every test, which in our case is useful for testing with different feature flag states.

### `maven_install.json`
Due to conflicts with develop, I had to regenerate this file since we can't directly edit it.

### Misc
There are no dark mode screenshots since the design team is still working on those(https://github.com/oppia/design-team/issues/192), and we will need to include them in a cleanup PR.

## Essential Checklist

## For UI-specific PRs only
||Admin| Non-Admin|
|---|---|---|
|Screenshot|Screenshot_2025-05-23-21-59-15-096_org oppia android|Screenshot_2025-05-23-21-59-20-605_org oppia android|
|Happy Path|https://github.com/user-attachments/assets/e5201d13-cd8f-476f-b3f8-35a830aa0074|https://github.com/user-attachments/assets/53ee4e37-f889-4bd6-9b08-0d3ef60b6c44|
|Wrong PIN|https://github.com/user-attachments/assets/91613b98-4b67-4f89-87dd-e32a306e96e8|https://github.com/user-attachments/assets/4d705cd9-4821-443b-9ce9-47d1a8a61989|
|Reset PIN|https://github.com/user-attachments/assets/eee181f0-4aa7-4b2d-88fa-8bd42fc39a69|https://github.com/user-attachments/assets/e061c064-8382-4afe-99ba-84e4101cf0ed|


deonwaju (approved):

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+      )
+    },
+    text = {
+      Text(stringResource(R.string.admin_confirm_app_wipe_message, appName))
+    },
+    properties = DialogProperties(
+      dismissOnClickOutside = false,
+      dismissOnBackPress = false
+    ),
+    onDismissRequest = {
```
deonwaju: You can simplify this with a function reference, its cleaner and prevents unnecessary lambda call: (Optional)

`onDismissRequest = onDismissRequest`

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+      onDismissRequest()
+    },
+    dismissButton = {
+      TextButton(
+        onClick = {
```
deonwaju: Same as above — consider refactoring : (Optional)
TextButton(onClick = onDismissRequest)

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+          onDismissRequest()
+        }
+      ) {
+        Text(stringResource(R.string.admin_confirm_app_wipe_negative_button_text))
+      }
+    },
+    confirmButton = {
+      TextButton(
```
deonwaju: Same as above — consider refactoring : (Optional)
TextButton(onClick = deleteAppData)

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
```diff
@@ -0,0 +1,58 @@
...
+      Text(stringResource(R.string.profile_login_forgot_pin_dialog_title))
+    },
+    text = {
+      Text(stringResource(R.string.profile_login_forgot_pin_dialog_message, appName))
+    },
+    properties = DialogProperties(
+      dismissOnClickOutside = false,
+      dismissOnBackPress = false
+    ),
+    onDismissRequest = {
```
deonwaju: Same as above — consider refactoring : (Optional)
`onDismissRequest = onDismissRequest`

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
```diff
@@ -0,0 +1,58 @@
...
+      onDismissRequest()
+    },
+    dismissButton = {
+      TextButton(
```
deonwaju: Same as above — consider refactoring : (Optional)
TextButton(onClick = onDismissRequest)

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
```diff
@@ -0,0 +1,58 @@
...
+        onClick = {
+          onDismissRequest()
+        }
+      ) {
+        Text(stringResource(R.string.profile_login_forgot_pin_dialog_cancel_button))
+      }
+    },
+    confirmButton = {
+      TextButton(
```
deonwaju: Same as above — consider refactoring : (Optional)
TextButton(onClick = onConfirmation)

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
```diff
@@ -147,7 +147,7 @@ class PinPasswordActivityPresenter @Inject constructor(
           activity.supportFragmentManager.beginTransaction().remove(previousFrag).commitNow()
         }
         val dialogFragment = AdminSettingsDialogFragment
```
deonwaju: Consider using null checks instead of Using !! (not-null assertions),it is risky because if `adminPin` or `name.get()` is null, the app will crash at runtime. 

```
val name = pinViewModel.name.get()
if (adminPin != null && pinViewModel.name.get() != null) {
  val dialogFragment = AdminSettingsDialogFragment.newInstance(adminPin, profileId, pinViewModel.name.get())
  dialogFragment.show(activity.supportFragmentManager, TAG_ADMIN_SETTINGS_DIALOG)
}
```

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
adhiamboperes, line 1: This change should not be part of this PR.

app/src/main/java/org/oppia/android/app/profile/PinPasswordActivityPresenter.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
adhiamboperes: I am unable to remove the lambdas because the `onClick` function expects a unit, which needs the lambda. What I have done, however, is to refactor the lambda to a single line to improve readability.

Screenshot 2025-07-01 at 13 12 19

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
adhiamboperes: Ditto here

app/src/main/java/org/oppia/android/app/profile/ForgotAdminPinDialog.kt
adhiamboperes: Ditto here

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
adhiamboperes: Same comment regarding the Unit.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
adhiamboperes: Ditto here.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
adhiamboperes: Ditto here.

utility/src/main/java/org/oppia/android/util/parser/html/HtmlParser.kt
adhiamboperes: Fixed

adhiamboperes:
Thanks @deonwaju! I have addressed your comments. Please resolve if you have the permissions, otherwise let me know and I will resolve them myself.

app/src/main/res/layout/profile_login_activity.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
BenHenning: Are we ever adding anything to this layout dynamically? If not, can we just make the `FrameLayout` the root, instead? It helps reducing the layout nesting.

app/src/main/res/layout/profile_login_fragment.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:background="@color/component_color_shared_screen_primary_background_color"
+        android:gravity="center">
+
+        <androidx.compose.ui.platform.ComposeView
```
BenHenning: Similar to my comment for the `profile_login_activity` layout, can the `ComposeView` possibly be the top-level one here? I'm assuming the `gravity` doesn't matter since the `ComposeView` has a `match_parent` width/height, and we should be able to set the background of the `ComposeView` (I'm assuming).

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
BenHenning, line 1: For this & the other class: we need to start thinking about naming standards for these files since they're a new type of file altogether that we haven't really had before.

https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#naming-unit-composable-functions-as-entities explains this, but it's honestly quite a bit more complicated than I was expecting. I suspect we'll need to think about this and have a lot more examples before we can settle on naming conventions that make sense to us.

I don't think there's an action item here, but I did want to get your thoughts on the matter, if you have any.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
```diff
@@ -0,0 +1,81 @@
...
+        .add(
+          R.id.profile_login_fragment_placeholder, profileLoginFragment, TAG_PROFILE_LOGIN_FRAGMENT
+        )
+        .commitNow()
+    }
+  }
+
+  /** Handles showing the [ResetPinDialogFragment]. */
+  fun handleRouteToResetPinDialog(profileId: ProfileId, profileName: String) {
+
```
BenHenning: Nit: please remove empty line for convention.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
```diff
@@ -0,0 +1,81 @@
...
+    resetPinDialog.showNow(fragmentManager, TAG_ADMIN_RESET_PIN_DIALOG)
+  }
+
+  /** Handles showing the reset pin success dialog. */
+  fun handleRouteToSuccessDialog() {
+    val resetPinDialog = fragmentManager.findFragmentByTag(TAG_ADMIN_RESET_PIN_DIALOG)
+      as DialogFragment
+    resetPinDialog.dismiss()
+
+    fragmentManager.executePendingTransactions()
```
BenHenning, line 63: For here & above: why is this needed?

BenHenning (changes requested):
Thanks @adhiamboperes! I haven't looked at the bulky bits of the PR yet (the new `Composable`s, the new fragment presenter, and the fragment test), but did look at everything else and left some comments. Besides those, could you also fix CI & look at the replies below?

> ### Other Design Choices
> **Dialog State Management:**
> 
> * When opening two dialogs in sequence in compose, I chose to display the second dialog without dismissing the first one. This causes the second dialog to appear on top of the first. When the second dialog is dismissed, the first remains open and must be manually closed.
> * On the flip side, if the first dialog is configured to close when the positive button is clicked(to open the next dialog), it dismisses correctly, but then the second dialog doesn’t appear at all.
> 
> Neither approach is ideal, but keeping the first dialog open while showing the second is the more acceptable of the two.

This shouldn't be necessary, I think. In the non-Compose world you can check if the fragment has already been added to the activity's fragment manager (via its tag) and replace it with a new one (or not open a new one at all) to de-duplicate the dialogs. Shouldn't there be a way to do the same via Compose?

> ### Misc
> There are no dark mode screenshots since the design team is still working on those([oppia/design-team#192](https://github.com/oppia/design-team/issues/192)), and we will need to include them in a cleanup PR.

Is there a tracking issue for this that includes the details for updating this & any other affected flows in particular?

> ## Essential Checklist
> ...

Can you please fill this out?

app/src/main/res/layout/profile_login_activity.xml
adhiamboperes: Not really. I have flattened the layout.

app/src/main/res/layout/profile_login_fragment.xml
adhiamboperes: Changed the top-level view to be the ComposeView

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
adhiamboperes: Since there is a series of dialogs leading to one another, forcing all pending transactions to complete immediately helps prevent stacking and errors.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
adhiamboperes: Generally, I believe that naming composable entities should follow the same conventions as the existing files in the project. We typically have FeatureNameActivity/Fragment/Dialog, and for custom views, we have CustomView. Composable entities will likely fall into 3 categories:
1. the composable functions that represent small view units, and will be written inside a presenter(or a larger Composable class, e.g. https://github.com/oppia/oppia-android/blob/43a08262fa9d5abddb9f0c5848aacbe6ca9af680/app/src/main/java/org/oppia/android/app/classroom/ClassroomListFragmentPresenter.kt#L207
2. composables that represent screens, standalone componets such as dialogs
3. and finally, custom views, e.g e.g. [AllClassroomsHeaderText.kt](https://github.com/oppia/oppia-android/blob/develop/app/src/main/java/org/oppia/android/app/classroom/classroomlist/AllClassroomsHeaderText.kt)

The list is not exhaustive, but will form a bulk of Composables that we write.

Combining our present naming convention with the Noun imperative from the docs, it shouldn't be too complex.

- `ComingSoonTopicList.kt` would be renamed to `ComingSoonTopicListView.kt`
- This dialog can be renamed from `ConfirmDataResetDialog.kt` to `DataResetConfirmationDialog`, 
- and the smaller unit composables can be named explicitly after the view they represent, e.g `NameErrorText()`

adhiamboperes:
> > ### Other Design Choices
> > **Dialog State Management:**
> 
> This shouldn't be necessary, I think. In the non-Compose world you can check if the fragment has already been added to the activity's fragment manager (via its tag) and replace it with a new one (or not open a new one at all) to de-duplicate the dialogs. Shouldn't there be a way to do the same via Compose?

Compose Navigation replaces Fragment transactions and tags entirely, but has state. For hiding/showing Composables, we use `remember { mutableStateOf(false) }`, and set the value to true when we want the Composable to be added to the tree, and false when we want it to be removed. I did some refactor, and found a way to manage these dialogs correctly.



> > ### Misc
> > There are no dark mode screenshots since the design team is still working on those([oppia/design-team#192](https://github.com/oppia/design-team/issues/192)), and we will need to include them in a cleanup PR.
> 
> Is there a tracking issue for this that includes the details for updating this & any other affected flows in particular?

#4938 should track all the work related to this project.



> > ## Essential Checklist
> > ...
> 
> Can you please fill this out?


Done

app/src/main/java/org/oppia/android/app/profile/AdminPinRecoveryDialog.kt
```diff
@@ -0,0 +1,52 @@
...
+import androidx.compose.material.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.window.DialogProperties
+import org.oppia.android.app.ui.R
+
+/** Test tag for the admin's forgot pin dialog. */
+const val ADMIN_PIN_RECOVERY_DIALOG_TEST_TAG = "TEST_TAG.admin_forgot_pin"
```
BenHenning: This seems a bit gross to me--it's generally a bad practice to customize something that affects the real production behavior of a component for test-specific reasons. Is this truly the cleanest way to do this?

We never ran into this problem pre-Compose since most fragments would have tags that had a legitimate use case for transactions.

Another thought: could we inspect that _a_ dialog is open, and that it has the expected text, rather than validating against the specific tag? This is more akin to what an end-to-end test (that is, a user manually verifying it) would do.

Additional thought after reviewing the tests: I definitely am not keen on this approach. It places far too much coupling between the test logic and the UI arrangement. As far as I can tell, it's probably possible to show _something_ with the tag that isn't at all the thing we actually care about. That's why we largely validate using strings in most places, but sometimes IDs (which is admittedly not much better than a test-only tag, especially if we only add the ID for the purpose of testing). Perhaps this is the "new way" of doing things and we need to rely more heavily on future screen shot tests for correct layout and text display...but I'd rather we explore other options before hard committing to that outcome.

app/src/main/java/org/oppia/android/app/profile/DataResetConfirmationDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.material.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.window.DialogProperties
+import org.oppia.android.app.ui.R
+
+/** Test tag for the confirm reset app data dialog. */
+const val DATA_RESET_CONFIRMATION_DIALOG_TEST_TAG = "TEST_TAG.admin_reset_app_data"
```
BenHenning: Ditto for this test tag.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
BenHenning: Got it--suggest adding a line comment explaining the need for the function.

Do tests cover it missing? (It may not be easy to test--that's fine, it just means that the line comment is more important to guard against unintentional regressions in the future).

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -0,0 +1,447 @@
...
+    }
+    return showError
+  }
+
+  private fun loginToProfile(profileId: ProfileId) {
+    profileManagementController.loginToProfile(profileId).toLiveData()
+      .observe(fragment) {
+        if (it is AsyncResult.Success) {
+          activity.startActivity(
+            if (enableMultipleClassrooms.value)
```
BenHenning: Nit: `if`-`else` here needs braces since it's wrapping to multiple lines.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -0,0 +1,447 @@
...
+    Text(
+      text = resourceHandler.getStringInLocaleWithWrapping(
+        R.string.profile_login_activity_enter_pin_prompt
+      ),
+      textAlign = TextAlign.Center,
+      style = TextStyle(
+        fontSize = 16.sp,
+        color = colorResource(id = R.color.component_color_profile_login_primary_text_color)
+      ),
+      modifier = Modifier.testTag(PROMPT_TEST_TAG)
```
BenHenning: Ditto for this (any any other) test tag (see my earlier comment in the PR).

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -0,0 +1,447 @@
...
+    val maskedChar = transformedPin.text.getOrNull(index)?.toString() ?: ""
+    val isFocused = pinValue.length == index
+
+    Box(
+      modifier = Modifier
+        .size(48.dp)
+        .aspectRatio(0.7F)
+        .border(
+          width = if (isFocused) 2.dp else 1.dp,
+          color = if (isFocused)
```
BenHenning: Ditto here (`if`-`else` needs braces). Ditto elsewhere in the PR (where needed).

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -0,0 +1,447 @@
...
+          setShakeOffset(-10f)
+          delay(50)
+          setShakeOffset(10f)
+          delay(50)
+        }
+
+        setShakeOffset(0f)
+
+        // Clear the incorrect PIN and hide the error message after a 2s delay.
+        delay(2000)
```
BenHenning, line 362: Does this & the above block the UI thread, or is Compose properly handing back control? Otherwise we risk an ANR (which is why we use the lifecycle-safe timer utility for such cases).

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+    profileTestHelper.addOnlyAdminProfile()
+    scenario = launch(ProfileLoginActivity::class.java)
+    testCoroutineDispatchers.runCurrent()
+
+    composeRule.onNodeWithTag(GREETING_TEST_TAG)
+      .assertTextContains(context.getString(R.string.profile_login_activity_greeting_text, "Admin"))
+      .assertIsDisplayed()
+    composeRule.onNodeWithTag(PROMPT_TEST_TAG).assertIsDisplayed()
+      .assertTextContains(context.getString(R.string.profile_login_activity_enter_pin_prompt))
+      .assertIsDisplayed()
```
BenHenning: Checking that it's displayed twice?

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+  fun tearDown() {
+    TestPlatformParameterModule.reset()
+    Intents.release()
+  }
+
+  @Test
+  fun testFragment_onLaunch_allTextViewsHaveCorrectContent() {
+    setUpTestApplicationComponentWithFeatureFlags()
+    profileTestHelper.addOnlyAdminProfile()
+    scenario = launch(ProfileLoginActivity::class.java)
```
BenHenning: This seems...ill-advised. We probably should be closing the scenario--is three a reason you don't approach it that way here (like we do in most/all other UI tests)?

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+  @Test
+  fun testFragment_onLaunch_adminProfile_fivePinInputBoxesAreDisplayed() {
+    setUpTestApplicationComponentWithFeatureFlags()
+    profileTestHelper.addOnlyAdminProfile()
+    scenario = launch(ProfileLoginActivity::class.java)
+    testCoroutineDispatchers.runCurrent()
+
+    repeat(5) { index ->
+      composeRule
+        .onNodeWithTag(PIN_BOX_TEST_TAG + index, useUnmergedTree = true)
```
BenHenning: What does `useUnmergedTree` do? The documentation made it sound like it had more to do with the error output during failure, but I admittedly only looked into it a ltitle.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+    scenario = launch(ProfileLoginActivity::class.java)
+    testCoroutineDispatchers.runCurrent()
+
+    composeRule
+      .onNodeWithTag(PIN_ERROR_TEST_TAG)
+      .assertDoesNotExist()
+  }
+
+  @Test
+  fun testFragment_nonAdmin_enterTwoDigits_doesNotTriggerLogin() {
```
BenHenning, line 238: Should this also be verifying no intent is fired? That would actually be verifying that the login hasn't been triggered.

(I was expecting more consistency with the next test, including the `runCurrent` call after inputting text).

Ditto for the admin case.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+  }
+
+  @Test
+  fun testFragment_adminUser_clickForgotPin_opensAdminForgotPinDialogFlow() {
+    setUpTestApplicationComponentWithFeatureFlags()
+    profileTestHelper.addOnlyAdminProfile()
+    scenario = launch(ProfileLoginActivity::class.java)
+    testCoroutineDispatchers.runCurrent()
+
+    composeRule.onNodeWithTag(FORGOT_PIN_TEST_TAG).performClick()
```
BenHenning: This feels really fake to me, and perhaps leans more heavily into my concerns with using test-only tags.

A user cannot perform this action, can they? We normally would require them to enter the wrong pin, first, _then_ click on the 'forgot pin' button. Here, we're clicking on a button that's not even visible.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+      .assertExists()
+      .assertIsDisplayed()
+
+    composeRule
+      .onNodeWithText(
+        context.getString(R.string.admin_confirm_app_wipe_negative_button_text)
+      )
+      .performClick()
+
+    composeRule.onNodeWithTag(DATA_RESET_CONFIRMATION_DIALOG_TEST_TAG)
```
BenHenning: Suggest also verifying the app isn't closing (for consistency with the next test).

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+  @Test
+  fun testFragment_nonAdminUser_enterWrongAdminPin_showsWrongAdminPinError() {
+    setUpTestApplicationComponentWithFeatureFlags()
+    profileTestHelper.addMoreProfiles(1)
+    scenario = launch(ProfileLoginActivity::class.java)
+    testCoroutineDispatchers.runCurrent()
+
+    composeRule.onNodeWithTag(FORGOT_PIN_TEST_TAG).performClick()
+
+    composeRule.waitForIdle()
```
BenHenning, line 712: Huh. What does this do? Does it actually interact with Espresso `IdlingResource` or is it doing something else?

I foresee me having to add Jetpack Compose to #89 if there's another asynchronous mechanism that needs to settle for tests.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
```diff
@@ -0,0 +1,829 @@
...
+      .inRoot(isDialog())
+      .perform(click())
+
+    onView(withId(R.id.reset_pin_input_pin))
+      .inRoot(isDialog())
+      .check(matches(isDisplayed()))
+  }
+
+  @Test
+  fun testFragment_nonAdminUser_enterAndSubmitNewPin_opensSuccessDialog() {
```
BenHenning: Suggest perhaps adding one more test that completes the flow here and validates that the user can actually log in with the new PIN once reset (that seems like a nice completion of the set of steps being verified here). It is slightly "end-to-end-esque" but I think that's fine here.

app/src/main/res/layout/profile_login_fragment.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <androidx.compose.ui.platform.ComposeView
```
BenHenning: Nit: This should be two-spaced, I think.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
BenHenning: That makes sense, though my concerns right now:
- Traditionally in Android naming, ending something with 'Activity', 'Fragment', 'Dialog', 'View', etc. usually meant that it was a subclass of the corresponding thing. With this naming style, that's no longer true and could introduce some confusion, especially in the intermediary state before we move entirely over to Compose. Do other apps just not care about this difference?
- The smaller unit composables still seem very fuzzy to me, but if it's possible for us to limit what they can be called to select types of composables (e.g. 'Text' in your example, and others) then we could perhaps tighten the naming quite a bit. This is a case where I think we would be advantaged with a custom CI check to ensure naming consistency. I can see things quickly getting very confusing without such a check in place.

BenHenning (commented):
Thanks @adhiamboperes! Apologies again for the delayed review. I plan to be a lot more responsive for the rest of the PRs in this project.

I've taken a pass on everything now, and have a bunch of thoughts. Most of them are around how we are actually testing the new Compose layouts--I have a few specific concerns that I think need to be addressed.

Most of my remaining comments are nits or small suggestions.

PTAL.

app/src/main/res/layout/profile_login_fragment.xml
adhiamboperes: Done

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: The behaviour tested here is correct and consistent with the legacy flow. The forgot PIN button is displayed when the screen is rendered and can be clicked without first entering an incorrect PIN.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: compose tests use the [semantics tree](https://developer.android.com/jetpack/compose/testing#semantics) to look for elements on the screen and read their properties. This is the structure that accessibility services use as well, as they're meant to be read by a service such as [TalkBack](https://support.google.com/accessibility/android/answer/6283677).

In newer versions of Compose, you can use `printToLog` which will output the full semantics tree showing all the nodes in that composable. This is similar to the entire view hierachy output printed by Robolectric  when a test fails with `cannot find view in hierachy matching...`. e.g.

```php
 |-Node #3 at (l=42.0, t=105.0, r=105.0, b=168.0)px
        | Role = 'Tab'
        | Selected = 'false'
        | StateDescription = 'Not selected'
        | ContentDescription = 'Overview'
        | Actions = [OnClick]
        | MergeDescendants = 'true'
        | ClearAndSetSemantics = 'true'
```

The Semantics tree always tries to be as compact as possible, showing only the information that is relevant. The above node has property `MergeDescendants = 'true'` which means that this node had descendants, but they have been merged into it, but in some tests, we want to access all the nodes.  Passing` useUnmergedTree = true`  to the onRoot finder allows us to query the unmerged Semantics tree. All finders have the `useUnmergedTree` parameter.

Sample: https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/samples/ui/test/samples/SemanticsNodeInteractionSamples.kt;drc=03ca30d22e6ee3483142f2e4048db459cb5afb79;l=58

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: Done.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: This method waits for all `IdlingResources` to become idle before executing actions.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-test-junit4/src/jvmMain/kotlin/androidx/compose/ui/test/junit4/ComposeTestRule.jvm.kt;drc=e687da9e7221a7de5adac9d3150334975233821d;l=98

Although similar to [Espresso's Idling Resources](https://developer.android.com/training/testing/espresso/idling-resource), [Compose's Idling Resources](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/IdlingResource) is a different API.

The `waitForIdle()` function here manually synchronizes the test to wait for Compose to become idle in order for the rest of the test (which runs on Espresso) to continue executing.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: Done.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: The test cannot click on a component that is not in the semantics tree, even if using the test tag to try to access it. The test will fail.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: Done for asserting that no intent is fired in the 2 test, but `runCurrent` does not need to be called because nothing happens after this input, therefore no synchronization is required.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: I wrapped the tests in the `use {}` block like we do in other tests. Unfortunately, this has resultled into a lot of formatting changes in the file due to new indentation.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
adhiamboperes: Yes. Fixed here and elsewhere.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/profile/AdminPinRecoveryDialog.kt
adhiamboperes: `Composables` don't have IDs and you can't use the Node numbers shown in the semantics tree to match them. If matching a node with its semantics properties is impractical or impossible, such as in the case of the PIN boxes which are dynamically generated and do not contain text when created, we can use the [testTag](https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/package-summary#(androidx.compose.ui.Modifier).testTag(kotlin.String)) modifier as a last resort. For this reason, I have removed the test tags where there is an alternate, more traditional matcher available.

app/src/main/java/org/oppia/android/app/profile/DataResetConfirmationDialog.kt
adhiamboperes: Removed.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: Removed all test tags except for `PIN_INPUT_TEST_TAG` and `PIN_BOX_TEST_TAG` as explained above.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: No. Please see [documentation](https://developer.android.com/develop/ui/compose/side-effects#launchedeffect).

- `LaunchedEffect` launches a coroutine in the composition’s `CoroutineScope`, tied to the current `CompositionLifecycle`.
- The coroutine runs on the main thread (Dispatchers.Main.immediate) by default.
- Every delay(...) is non-blocking, it suspends the coroutine and returns control to the main thread’s event loop.
- The coroutine will be cancelled if LaunchedEffect leaves the composition.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
adhiamboperes: Most apps I’ve seen that have moved to Compose don’t stress too much about the inheritance mismatch. They often:

- Use `*Screen`, `*Dialog`, `*Card`, `*Row`, etc. for composables.
- Rely on package structure (e.g. ui/screen, ui/component, ui/dialog) to disambiguate.
- Accept that the suffix signals role, not class hierarchy anymore.

**Possible conventions for Oppia**
Screens (top-level composables) - represents a whole navigable screen.
Naming: `XyzScreen.kt`,  e.g `ProfileListScreen()`.

b. Dialogs & other standalone components - represents:a self-contained UI element invoked directly (dialogs, bottom sheets).
Naming: `XyzDialog.kt`,  e.g. `DataResetConfirmationDialog()`.
This keeps alignment with our current convention while signaling functionality not inheritance.

c. Small reusable UI units (building blocks)
Naming: `XyzView.kt` or `XyzSomething()` where Something is a UI role.
e.g `ErrorText()`, `TopicCard()`.

This way, Dialog and Screen still mean “top-level entry point,” while View and specific UI-role suffixes capture small building blocks.


I agree CI checks will be invaluable here. We can enforces that only approved suffixes (Screen, Dialog, View, Text, Card, etc.) allowed for composables, and that File name ↔ top-level composable name match.

To avoid confusion in the hybrid state, we can separate Compose files into dedicated packages (e.g. ui/compose/dialog vs ui/dialog).

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
BenHenning: Thanks.

Updated #89 to consider this.

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileLoginFragmentTest.kt
BenHenning: Interesting. I like that--it adds a lot more flexibility in testing.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
BenHenning: Wow. The cooperation between the main thread's coroutine scope and the main looper must be very complex to make something like that work. I...wonder if anyone's run into challenging threading issues from that mechanism. At a glance this seems very cool since it solves a rather annoying problem with Android development, but I do worry a bit what the actual costs are to make this work.

Either way, I don't have any suggestions to change this since it's obviously following best practices and the cooperative dispatcher _should_ avoid ANRs.

app/src/main/java/org/oppia/android/app/profile/ConfirmDataResetDialog.kt
BenHenning: I like where this discussion is headed--perhaps it's a good time to maybe move this into a document for further collaboration?

I don't want to over-index on what other apps are doing unless there's clear, unambiguous best practices that we can follow. It doesn't sound like there is and that people are largely following an unspecified or partly specified convention. That convention demonstrably has gaps that could lead to confusion, and I think your suggestions largely address those.

I also think there may be a bit more we can do on things that call into (c) to fully define them (maybe having a suffix like 'Unit' or something, not sure), but this is best continued out of this PR, I think, so that it doesn't block the PR itself.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
BenHenning: Reminder to please respond to this.

BenHenning (commented):
Thanks @adhiamboperes! Just had a few follow-up thoughts, but I like the changes that moved away from test tags where it's avoidable (which is, hopefully, most places as we move more of the app over to Compose).

The PR otherwise largely looks good to me.

BenHenning:
Also, it looks like there are some conflicts to address from `develop`.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginActivityPresenter.kt
adhiamboperes: Apologies -- I missed this. I have added a comment.

The tests as they are, pass with or without this function, but I am not sure how to test the posibility of stacking dialogs currently. In future, if we do end up with a pure Compose app, we might be able to test such scenarios.

adhiamboperes:
@BenHenning, addressed pending comment and resolved conflicts. Also, in the latest commit I have deleted a couple of tests that are actual duplicates of earlier tests. PTAL.

BenHenning (approved):
Thanks @adhiamboperes!

Issue #5835, opened 2025-05-31
[Feature Request]: Revise platform parameter system


BenHenning:
### Is your feature request related to a problem? Please describe.

This is the tracking issue largely covering the work from #5725.

The specific issue is projects and issues like #5345, #4302, #5676, and #4304 become exceedingly more challenging with the way the current system is designed. In particular:
- Using a singleton for managing state requires a lot more duplication to bring in feature flags as a first class item.
- Making injection cleaner in general requires large amounts of rewriting.
- Using annotations for overwriting (e.g. in tests) requires stronger typing for flags and parameters which has far-reaching impacts on the system.
- Decomposing the module for flags into multiple not only makes testing messier but also is a bit tricky with the way the system is currently wired.
- Changing initialization order is tricky under the current system (fix the syncing race condition), and very far reaching for tests.

### Describe the solution you'd like

As #5725 demonstrates, it is possible to introduce an entirely new platform parameter system that fixes the ordering issues, obviates the module splitting (by generating the parameter modules), provides support for strong typing and first-class representation for feature flags and parameters, introduces strong testing functionality, and introduces a very clean build-time configuration for platform parameter and feature flag states (to make staging features much easier).

Going into the full depth of the solution is a bit too much for this issue, so refer to the issue above.

As also indicated in #5725, this needs to be approached in multiple pieces:

This PR is going to be split up in the effort of time. Specifically, the sub-parts are:
- [x] Introduce new platform parameter controller
  - Hook it up to the existing injection & testing approaches for platform parameters and feature flags to ensure existing functionality interoperability.
- [ ] Introduce new generation script for the platform parameter modules.
- [ ] Integrate platform parameter module generation with `PlatformParameterController`.
- [ ] Introduce the new testing approach for feature flags and platform parameters.
  - This may require much of the ongoing module cleanup work to continue (or, rather, significantly benefit from it).
  - This may also require some forward work around adding test application classes to all existing tests. I may opt for code generation for this as well.
- [ ] Update the documentation for https://github.com/oppia/oppia-android/wiki/Platform-Parameters-&-Feature-Flags.
- [ ] Rethink staging and overrides for parameters and flags, and reconsider ownership (perhaps instead using a common group with web's definitions file for Android features?).

### Describe alternatives you've considered

Too much to list here, but there are certainly many alternative implementations possible.

### Additional context

Much of this was discovered during the review of proposals for #5345.

BenHenning:
[...]

PR #5836, opened 2025-06-01
Fix #5676, part of #5835: Introduce new platform parameter controller

### LINKED ISSUE #5676: [BUG]: Platform parameter syncing races against SplashActivity

BenHenning:
## Explanation
Fixes part of #5835
Fixes #5676

This PR addresses one major part of #5835: it introduces a new base platform parameter system.

Implementation details:
- `PlatformParameterController` is now an interface with a production-specific implementation. This is so that we can have a debug implementation (as needed for #5345).
- `PlatformParameterSingleton` has been replaced with a new `PlatformParameterProcessState`. This behaves much the same way as the singleton except it guarantees initialization lifecycles by prohibiting double initialization or attempting to retrieve a parameter value before it's ready.
  - This should eliminate race conditions like #5676 because attempting to access a parameter too early will actually cause a crash.
  - This also ensures that tests are always properly initialized (which has required other solutions described below).
- Feature flags are now a first-class domain concept, though they aren't yet differentiated at the injection level (see caveats below).
- Proto changes:
  - Platform parameter and feature flag IDs are now backed by proto enums rather than not being entirely defined. These enums use ordinal values based on the order the parameter or flag was introduced (including reserving removed ones). This acts as thought the system has always been in place for the app.
  - `SyncStatus` is now a top-level enum since it's used in multiple places.
  - There are distinct definition structures for both platform parameters and  feature flags. These correspond to new textproto files that define the actual defaults and remote server names for each supported parameter and flag, using configuration textproto files (see next top-level bullet below).
  - Remote parameters and flags use a separate structure for the server-provided value (to be merged by the `PlatformParameterController`).
  - Feature flags don't bother with `PlatformParameterValue` since they can only ever be enabled/disabled (and are represented as such).
- Platform parameters and feature flag defaults are no longer provided via constants but as part of text proto files that are loaded at runtime.
  - This may change in a future implementation to copy the defaults into a generated module (as per the plans in #5835).
  - This mechanism allows for a test-specific loader to be used to substitute default values for testing purposes (which makes solving #4302 eventually possible).
- Alluded to above, the new `TestPlatformParameterConfigRetriever` can be used to statically prepare platform parameters and feature flags for testing.
  - This is currently called via `TestPlatformParameterModule` for feature flags to minimize the changes needed to tests.
  - There are checks to ensure parameters are properly reset between tests, and that flags/parameters are only overwritten once and before initialization time. These all ensure that tests are behaving correctly for predictable flag initialization (to avoid trying to override a flag after a test has started only for that flag to not take effect). This **significantly** improves the reliability of overriding parameters and flags in tests, and reduces the potential human error that can occur here.
- The new configuration `textproto` files are designed to allow flavor-specific overrides which obviates the need for something like `PlatformParameterAlphaModule` (which is a really nice simplification).
  - This is especially nice to quickly see the difference in features and parameters between flavors rather than needing to scan the whole redefined list.
- Parameter initialization behaves like so: one (or more) calls to `loadParametersAsync`, then use `getParameterInitializationStatus` to observe when initialization finishes.
  - This allows non-UI code like application startup to initiate the actual parameter syncing while ensuring that the splash experience properly blocks on parameters being ready.
- **Limitation** in tests: properly initializing feature flags and platform parameters in tests is even harder than for the main app. As a temporary workaround, some very particular piping was introduced in `TestPlatformParameterModule` and a _blocking_ synchronization point to ensure that it's impossible to reference platform parameters before they have been fully loaded. With the other design checks in place that prevent later overwriting, this is a key aspect of ensuring strong predictability in test behaviors.
  - This will be removed with the solution to #4302 since a JUnit rule is the most reliable location for initializing things before injection, but that's tricky for different reasons not explored in this PR.
- A new `FeatureFlagId`-keyed map of feature flags is available for injection (mainly for logging purposes).
- New binding modules were introduced for separately providing platform parameter and feature flags injectable definitions (and the feature flags map). These modules are implicitly provided via submodules. This is a temporary solution to avoid significant changes to test modules, and will eventually go away in favor of generated modules (per #5835). It does have the added benefit of ensuring consistency between the prod and test platform parameter modules.
- `platformparameter` (as part of the domain layer) is now an actual Bazel package, so there are some build file updates accordingly.
- `PlatformParameterControllerProdImpl` exposes some package-local functions for use by the new debug controller variant as part of the implementation of #5345.
- Remote platform parameter downloading has moved to the controller (or will) and out of the sync-up worker for simplicity.

Caveats:
- Initialization for parameters is complex. The cleanest way I could figure on how to do this was a deferred initialization of the entire application startup by using `Dispatchers.Main`. I'm not keen on this as it could very significantly **introduce startup issues**. I'll need to think on how better to solve this long-term, but it's a viable approach for now.
  - Note that this is ultimately necessary because initializing work manager can synchronously construct some of the worker objects that rely on platform parameters.
  - Another approach might be to use Dagger `Provider`s for these, but it gets a bit messy (and difficult to discover when creating a new worker in the future).
- Remote platform parameter syncing has been **disabled**, though it's currently not working, anyway, so this isn't a regression to current app behavior. Fixing this correctly would require introducing a bunch more module changes, further exploding the scope of this PR. It will be reintroduced in a later PR, instead.
- Using `TestPlatformParameterConfigRetriever` is a little bit annoying right now since it requires overriding parameters before injection, and resetting them in tear-down methods. This explains many of the test changes in this PR, but it will eventually become much nicer with #4302 addressed.
  - Sometimes this gets extra annoying, such as with `AdministratorControlsActivityTest` where it needed to be expanded out simply because the previous test behavior involved disabling the override after injection (which is no longer valid). This happened in some other test suites, as well.
- Splash initialization is a bit more complicated now since `SplashActivityPresenter` cannot directly inject parameters. Instead, a new 'post loading' 'presenter' is used for this purpose, and its lifecycle is carefully managed to ensure it's not injected before parameter loading has completed.
- There was one discrepancy between `TestPlatformParameterModule`'s initial override states and the real defaults: spotlights were being enabled by defaults in tests and not prod. This has been reversed and only the necessary tests that required spotlights have been updated to override this.
- A bunch of tests were updated to use `TestPlatformParameterModule`. Due to the design limitation mentioned above, this is necessary for any test that uses platform parameters or feature flags (even if only for production code).
- Introducing the new textproto files led to a requirement for `AssetModule` in a bunch of tests.
- There are a bunch of now-deprecated or soon-to-be-deprecated old machinery for platform parameters that are not being removed in this change in order to try and keep it smaller.
- `PlatformParameterServiceTestOrchestrator` is a bit more complicated to use presently in order to maintain compatibility with existing tests and reduce changes. It doesn't allow mixing valid and invalid parameters any longer, and it's not using centrally defined test parameters (since these aren't valid--there are no test-only feature flags).

Other important things to note:
- The feature flags check was removed. It seems this is no longer needed since there's now an injected map of all feature flags to log (so there's no manual updating other than the Dagger side, and this will eventually be generated per #5835). There's also pretty good testing for the logger to ensure that all known feature flags are correctly logged.
- Multiple classrooms has been disabled by default since it's not yet production ready. However, it's been enabled for alpha builds along with spotlights and learner study analytics (keeping parity with the old `PlatformParameterAlphaModule`).
- This change attempted to try and only update tests that required updates, though some may have been updated slightly along the way. This is generally fine since it's preferable for all tests to use `TestPlatformParameterModule` in the long run.
- All of the feature flag definitions now have remote server names which are also ensured to follow the Oppia web convention for Android feature flags. Ditto for platform parameters that actually need remote syncing (not all of which do).
- Some platform parameter defaults have technically regressed: the version codes for app updates (both forced and optional). Previously, these defaults to the current version of the app. Now they default to 0 since the gate is a greater-than check (so it shouldn't change the resulting logic for deprecation notices).
- A lot of tests were performing their own parameter overriding. These have largely been removed in favor of using `TestPlatformParameterModule`, instead.
- The change to `FeatureFlagItemContext` could result in some inconsistencies for new versions of the app. While forward compatible, there will be a defaulting of the new `id` field. This is probably fine since it'll only impact logs not yet uploaded (which should be fairly rare, and the failure case is fine from a logging perspective since the team isn't yet fully relying on these logs).
  - This also included some Truth subject updates.
- `FeatureFlagNameToNumericIdConverter` has been removed (along with its test) since the `FeatureFlagId`'s `number` value can be used, instead. It should have the same value as the old converter produces.
- `LogReportingModuleTest` was a bit tricky because pulling in some of the additional modules now required for `TestPlatformParameterModule` led to injecting a production dependency from `LogReportingModule` that required Firebase to be initialized. This was technically the test not being set up correctly initially, and likely would have required this change in the future, anyway.
- Downloads is being used as a proxy for a feature flag that is off by default, and NPS survey is being used as a proxy for a feature flag that is on by default (to replace previous test-only variants that didn't really exist). This is a medium-term fix until a better long-term solution can be determined.
- **Behavior change**: `FeatureFlagsLogger` now sorts the flags that it logs for determinism in tests (and for benefits in queries later on--ordering is often nice to rely upon for these cases).
- Exemption explanations:
  - The TODO line number was changed for `static_checks.sh` due to the removal of the feature flags check.
  - Test file exemptions that are low priority to resolve (and will not be tracked) or don't ever need tests:
    - `FeatureFlagBindingModule`
    - `FeatureFlagIdKey`
    - `FeatureFlags`
    - `FeatureFlagsMapBindingModule`
    - `PlatformParameterBindingModule`
    - `PlatformParameterConfigRetriever`
  - Test file exemptions that will have tests added as part of #5835:
    - `PlatformParameterConfigRetrieverProdImpl`
    - `PlatformParameterProcessState`
    - `TestPlatformParameterConfigRetriever`
- Other tests which need to be reimplemented as part of #5835 and have been removed in this change (these were omitted due to complexity of fixing against the urgency of this change being merged):
  - `PlatformParameterControllerProdImplTest`
  - `PlatformParameterIntegrationTest`
  - `PlatformParameterModuleTest`
  - `PlatformParameterSyncUpWorkerTest`

## Essential Checklist

## For UI-specific PRs only
This is a significant infrastructural change to the app. However, if it's behaving correctly then there should be no impact on the app's UX. There are no intentional UX or UI changes introduced in this PR.


BenHenning:
/cc @Rd4dev and @ayush0402 

The tests still need to be completely fixed, plus docs and new tests finalized, but this actually works for the ~2 dozen tests I migrated + production so I think it's a viable approach.

BenHenning:
This PR is not 100% ready for review, but we need to review it anyway. PTAL @adhiamboperes and @Rd4dev.

I still have 31 known test failures to fix, plus docs, lint fixes, and PR description (along with a self-review). There are a lot of test changes being deferred to a later PR to prioritize merging this ASAP.

Apologies for not having a fully documented description; I hope that the implementation is reasonably straightforward to review without that (though it _will_ come later).

adhiamboperes:
Looking at this. Will probably finish later today.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -1,349 +1,49 @@
...
+  includes = [
+    FeatureFlagsMapBindingModule::class,
+    FeatureFlagBindingModule::class,
+    PlatformParameterBindingModule::class,
+    PlatformParameterModule.PlatformParameterProcessStateModule::class,
+    PlatformParameterModule.PlatformParameterControllerProdImplModule::class
+  ]
+)
+interface PlatformParameterModule {
+  @Binds
```
Rd4dev, line 19: Could we consider separating the controller provision from `PlatformParameterModule` and instead provide it via a specific prod module? Or is it fine to keep it this way for now since there aren’t any debug-specific modules yet, and we can separate it later when introducing the debug implementation?

domain/src/main/java/org/oppia/android/domain/platformparameter/BUILD.bazel
```diff
@@ -0,0 +1,100 @@
...
+        "PlatformParameterModule.kt",
+        "PlatformParameterSingletonModule.kt",
+    ],
+    visibility = ["//:oppia_prod_module_visibility"],
+    deps = [
+        ":annotations",
+        ":process_state",
+        ":prod_impl",
+        "//:dagger",
+        "//domain/src/main/java/org/oppia/android/domain/oppialogger:startup_listener",
```
Rd4dev: For my understanding, curious to know which module directly depends on `startup_listener`.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
```diff
@@ -0,0 +1,233 @@
...
+        val remote = this.remote ?: computeRemote()
+        return if (remoteValue != null) {
+          remote.toBuilder().apply {
+            this.remoteValue = remoteValue
+            this.syncStatus = SyncStatus.SYNCED_FROM_SERVER
+          }.build()
+        } else remote // Nothing changes.
+      }
+
+      private fun computeRemote() = RemotePlatformParameter.newBuilder().apply {
```
Rd4dev: Considering if we could rename to `buildInitialRemote` or `createDefaultRemote` or something denoting the default, as the `computeRemote` gives the impression that it's calculating a value, whereas here it's more like a fallback or default setup.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterProcessState.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.oppia.android.app.model.PlatformParameterValue
+import org.oppia.android.app.model.SyncStatus
+
+class PlatformParameterProcessState {
+  private lateinit var platformParameters: Map<PlatformParameterId, PlatformParameterValue>
+  private lateinit var featureFlags: Map<FeatureFlagId, Boolean>
+  private lateinit var featureFlagSyncStatuses: Map<FeatureFlagId, SyncStatus>
+
+  fun initializePlatformParameters(states: Map<PlatformParameterId, PlatformParameterValue>) {
+    check(!::platformParameters.isInitialized) {
```
Rd4dev, line 38: Ah, I just realized this is a one-time immutable initialization — so I guess that means we can’t really delegate from the debug implementation to the prod one and then override it again? In that case, would we have to make `loadAllParameterStates` from `prodImpl` public and recompute it along with the debug-specific values and let the `debugImpl` handle directly the loading of parameters (no delegation to `prodImpl`)? 

Just wondering if that could work—or if there might be a better approach.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt
```diff
@@ -1,96 +1,46 @@
...
+   *    actual value returned does not have any significant meaning.
    */
-  private suspend fun getDeferredResult(
-    deferred: Deferred<PlatformParameterCachingStatus>
-  ): AsyncResult<Any?> {
-    return when (deferred.await()) {
-      PlatformParameterCachingStatus.SUCCESS -> AsyncResult.Success(null)
-    }
-  }
+  fun loadParametersAsync(): Deferred<Any?>
```
Rd4dev: Does having a generic return type `Deferred<Any?>` serve a specific purpose here, or could it be explicitly defined as `Deferred<Unit>` since it doesn’t return anything?

model/src/main/proto/platform_parameter.proto
```diff
@@ -33,8 +33,154 @@ message PlatformParameter {
...
+// Corresponds to a remotely synced platform parameter.
+message RemotePlatformParameter {
+  // The unique ID of this platform parameter which corresponds to a PlatformParameterDefinition.
+  PlatformParameterId id = 1;
+
+  // The remotely synced value of this parameter, or absent/default if never yet synced.
+  PlatformParameterValue remote_value = 2;
+
+  // Indicates the sync status of the platform parameter.
+  SyncStatus sync_status = 3;
```
Rd4dev, line 168: Reiterating earlier comment from [[#5725 comment](https://github.com/oppia/oppia-android/pull/5725/files#r2094157221)] -- Following up to confirm if we should proceed with removing the `sync_status` field from `RemotePlatformParameter`, since a fetched remote value could be treated as default `SYNCED_FROM_SERVER` (assuming the current scope).

Rd4dev (commented):
@BenHenning Apologies — I couldn’t find the time to look at everything in detail, but for now I did go through few key files:  parts of (`PlatformParameterController`), `PlatformParameterProcessState`, `PlatformParameterModule`, `platform_parameter.proto`, and `platformparameter/BUILD.bazel`. I also pulled the changes locally and tried working off of them — everything seems to be working fine, so no major concerns on that front. I’ve added a couple of comments mainly for clarification to proceed with #5345. 

Apart from that a general clarification request on: will only feature flags be synced from the server? I assumed so since parameters default to `SYNC_STATUS_UNSPECIFIED` in the binding module. Some download-related pieces (like serialization) are still present, but I assume those are specifically left to help proceeding with storage implementing #5725.

Also, please resolve any comments that might block the merge — as most are just for my own understanding. If I get time to dig deeper and need further clarification, I’ll reach out. Thanks!

BenHenning:
Bringing the branch up-to-date and adding a bunch of changes to try and fix all remaining failures.

I still need to finish the PR description (with a self-review) and look at your comments @Rd4dev.

Also apologies @theayushyadav11; I should've double checked your handle earlier when cc'ing you.

Rd4dev:
@BenHenning -- Also, the binary check wrongly identifies newly introduced `.textproto` files as binary. 

image

I've filed a bug #5847 reporting this and proposed a possible quick fix. Thought would check if this is reliable, or a `.kt` script with proper testing be more sustainable to avoid regressions.

BenHenning:
Static check will pass once #5849 is merged. I'll hold off on syncing develop until then (I don't expect other changes to impact the PR, but it's definitely possible depending on what sorts of test changes have happened).

app/src/main/java/org/oppia/android/app/splash/SplashActivityPresenter.kt
```diff
@@ -75,11 +77,14 @@ class SplashActivityPresenter @Inject constructor(
   private val appLanguageLocaleHandler: AppLanguageLocaleHandler,
   private val lifecycleSafeTimerFactory: LifecycleSafeTimerFactory,
   private val currentBuildFlavor: BuildFlavor,
-  @EnableAppAndOsDeprecation
-  private val enableAppAndOsDeprecation: PlatformParameterValue<Boolean>,
   private val profileManagementController: ProfileManagementController,
```
BenHenning: This is probably wrong: everything here likely needs to be a `Provider`.

app/src/test/java/org/oppia/android/app/translation/BUILD.bazel
```diff
@@ -19,6 +19,7 @@ oppia_android_test(
         "//domain/src/main/java/org/oppia/android/domain/oppialogger:prod_module",
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/analytics:cpu_module",
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/analytics:prod_module",
+        "//domain/src/main/java/org/oppia/android/domain/platformparameter:prod_module",
```
BenHenning: Should be test module?

domain/src/test/java/org/oppia/android/domain/exploration/testing/BUILD.bazel
```diff
@@ -15,6 +15,7 @@ oppia_android_test(
         "//domain/src/main/java/org/oppia/android/domain/exploration/testing:fake_exploration_retriever",
         "//domain/src/main/java/org/oppia/android/domain/exploration/testing:test_module",
         "//domain/src/main/java/org/oppia/android/domain/oppialogger/analytics:prod_module",
+        "//domain/src/main/java/org/oppia/android/domain/platformparameter:prod_module",
```
BenHenning: Check everywhere that prod_module references are correct. Most likely are not.

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagBindingModule.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.oppia.android.util.platformparameter.EnableLoggingLearnerStudyIds
+import org.oppia.android.util.platformparameter.EnableMultipleClassrooms
+import org.oppia.android.util.platformparameter.EnableNpsSurvey
+import org.oppia.android.util.platformparameter.EnableOnboardingFlowV2
+import org.oppia.android.util.platformparameter.EnablePerformanceMetricsCollection
+import org.oppia.android.util.platformparameter.EnableSpotlightUi
+import org.oppia.android.util.platformparameter.PlatformParameterValue
+
+/** Dagger module for providing injectable bindings for feature flags. */
+// TODO(#5835): Remove this module.
```
BenHenning, line 21: Move above Kdoc.

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagsMapBindingModule.kt
```diff
@@ -0,0 +1,129 @@
...
+import org.oppia.android.util.platformparameter.EnableLoggingLearnerStudyIds
+import org.oppia.android.util.platformparameter.EnableMultipleClassrooms
+import org.oppia.android.util.platformparameter.EnableNpsSurvey
+import org.oppia.android.util.platformparameter.EnableOnboardingFlowV2
+import org.oppia.android.util.platformparameter.EnablePerformanceMetricsCollection
+import org.oppia.android.util.platformparameter.EnableSpotlightUi
+import org.oppia.android.util.platformparameter.PlatformParameterValue
+
+/** Dagger module for providing a map of feature flags, per [FeatureFlags]. */
+// TODO(#5835): Remove this module.
```
BenHenning, line 22: Ditto.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterBindingModule.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.oppia.android.util.platformparameter.OptionalAppUpdateVersionCode
+import org.oppia.android.util.platformparameter.PerformanceMetricsCollectionHighFrequencyTimeIntervalInMinutes
+import org.oppia.android.util.platformparameter.PerformanceMetricsCollectionLowFrequencyTimeIntervalInMinutes
+import org.oppia.android.util.platformparameter.PerformanceMetricsCollectionUploadTimeIntervalInMinutes
+import org.oppia.android.util.platformparameter.PlatformParameterValue
+import org.oppia.android.util.platformparameter.SplashScreenWelcomeMsg
+import org.oppia.android.util.platformparameter.SyncUpWorkerTimePeriodHours
+
+/** Dagger module for providing injectable bindings for platform parameters. */
+// TODO(#5835): Remove this module.
```
BenHenning, line 20: Ditto.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt
```diff
@@ -1,96 +1,58 @@
...
+   * Returns a [DataProvider] that will switch from a value of 'false' to 'true' when parameters
+   * have finished loading.
    *
-   * @param deferred task which needs to be executed
-   * @return async result for success or failure after the execution of deferred task
+   * This method and the returned [DataProvider] are safe to use for the lifetime of the process,
+   * even before [loadParametersAsync] is called. Calling this method will not initiate any actual
+   * initialization--[loadParametersAsync] must be used for that, instead.
+   *
+   * The returned [DataProvider] will never return a pending state, or will it represent the failure
```
BenHenning: nor will it

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
```diff
@@ -0,0 +1,238 @@
...
+      DATABASE_NAME, RemotePlatformParameterAndFeatureFlagDatabase.getDefaultInstance()
+    )
+  }
+  // Note that the 'by lazy' here guarantees thread-safe and singleton initialization.
+  private val initializationDeferred by lazy { loadParametersInternalAsync() }
+  private val areParametersLoadedFlow by lazy { MutableStateFlow(false) }
+
+  init {
+    // Ensure that parameters and flags are fully loaded ahead of a call to retrieveData() since
+    // loadParameters() guarantees exactly one non-pending result. Note that this will also
```
BenHenning: Update and check for other uses.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
```diff
@@ -0,0 +1,238 @@
...
+      val featureFlags = params.filterIsInstance<ParameterState.FeatureFlag>()
+      val platStatesById = platformParams.associate { it.definition.id to it.computeCurrentState() }
+      val flagStatesById = featureFlags.associate { it.definition.id to it.computeCurrentState() }
+      val statusesById = featureFlags.associate { it.definition.id to it.computeCurrentStatus() }
+      processState.initializePlatformParameters(platStatesById)
+      processState.initializeFeatureFlags(flagStatesById)
+      processState.initializeFeatureFlagSyncStatuses(statusesById)
+
+      // Let observers know that parameters have been initialized.
+      areParametersLoadedFlow.value = true
```
BenHenning: Nit: should probably be `parametersAreLoadedFlow`.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
```diff
@@ -1,349 +1,56 @@
...
+
+  @Binds
+  fun bindPlatformParameterConfigRetriever(
+    impl: PlatformParameterConfigRetrieverProdImpl
+  ): PlatformParameterConfigRetriever
+
+  /**
+   * Dagger module for providing the application-wide instance of [PlatformParameterProcessState].
+   */
+  // TODO(#5835): Remove this and make PlatformParameterProcessState injectable once the hack for
```
BenHenning: Ditto here & below.

BenHenning (commented):
Finished self-review of the entire PR, and documented everything that seemed important in the PR description. Most of my self findings are definitely nits.

BenHenning:
> Apart from that a general clarification request on: will only feature flags be synced from the server? I assumed so since parameters default to `SYNC_STATUS_UNSPECIFIED` in the binding module.

Thanks @Rd4dev! As to this, no, platform parameters will also be synced. The `SYNC_STATUS_UNSPECIFIED` is being set right now because we don't actually use sync status for anything except logging feature flags. I recognize this may introduce a problem with #5345, but note that sync status isn't filled in yet, anyway, due to remote parameters being disabled. We can chat further on how to mitigate this better, but I'd rater get this PR merged first.

> Some download-related pieces (like serialization) are still present, but I assume those are specifically left to help proceeding with storage implementing #5725.

I'm not sure what's meant by this, can you clarify?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
BenHenning: I would _really_ rather avoid that because it explodes the module includes in every test in the codebase.

I was envisioning that the debug variant of this module would just duplicate it (which is actually why it's so small now). Fewer bindings will be needed once some of the weird edge cases are cleaned up (mainly for test initialization), and in fact the only things that will need to be provided will be `PlatformParameterController` and `PlatformParameterConfigRetriever` so I don't see a debug copy being too onerous to write and maintain, or am I missing something?

domain/src/main/java/org/oppia/android/domain/platformparameter/BUILD.bazel
BenHenning: None. :) It was a previous attempt at introducing app startup initialization, but it didn't work because we need to parameter syncing both earlier than the listeners and before the listeners (or really most of anything) is called.

This dependency has now been removed.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
BenHenning: Will address this in my next round of changes (along with my own comments) since I agree.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterProcessState.kt
BenHenning: I was assuming we would override `loadParametersAsync` and basically reimplement it in the debug controller (i.e. call through to the production controller's helpers since they're exposed, but perform overriding locally and then finally pass the state onto `PlatformParameterProcessState`).

That basically means creating a copy and debug-specific version of this method from the production controller:

```kotlin
  private fun loadParametersInternalAsync(): Deferred<Unit> {
    return CoroutineScope(backgroundCoroutineDispatcher).async {
      val params = loadAllParameterStates()

      // Synchronize injectable parameter and flag state.
      val platformParams = params.filterIsInstance<ParameterState.PlatformParameter>()
      val featureFlags = params.filterIsInstance<ParameterState.FeatureFlag>()
      val platStatesById = platformParams.associate { it.definition.id to it.computeCurrentState() }
      val flagStatesById = featureFlags.associate { it.definition.id to it.computeCurrentState() }
      val statusesById = featureFlags.associate { it.definition.id to it.computeCurrentStatus() }
      processState.initializePlatformParameters(platStatesById)
      processState.initializeFeatureFlags(flagStatesById)
      processState.initializeFeatureFlagSyncStatuses(statusesById)

      // Let observers know that parameters have been initialized.
      areParametersLoadedFlow.value = true

      // Erase the data provider's value so that callers cannot inadvertently depend on the actual
      // list of parameters available.
    }
  }
```

We _could_ expose `loadAllParameterStates` instead of duplicating it; I'm not opposed, but I did it this way to avoid needing to expose `ParameterState`.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt
BenHenning: Nope. I had a reason a few months ago, but I forgot what it is. `Unit` seems sensible and the API contracts have been correspondingly updated.

model/src/main/proto/platform_parameter.proto
BenHenning: I think I disagree with this, but I recognize it's confusing.

The main problem I see is we need to store _somewhere_ that we synced successfully from the server (or that we failed). I think that I have a solution for this now:

We should replace this status with a new structure that more granularly tracks things. This would require introducing a structure like so:

```proto
message SyncDetails {
  // The millisecond timestamp that this parameter/flag was last successfully downloaded from the server.
  int64 last_successful_sync_timestamp_ms = 1;

  // The millisecond timestamp that the server was last contacted for syncing (regardless of success).
  int64 last_sync_attempt_timestamp_ms = 2;

  // Whether the last sync included this parameter. This will be false if the last sync failed OR if the last sync succeeded but was missing this parameter.
  bool last_sync_included_this_parameter = 3;

  // Whether the most recent attempt to contact the server for parameters failed.
  bool last_sync_failed = 4;
}
```

This proto would be absent from parameters and flags that have not yet synced.

This lets us derive the same states and likely any future states where we would want to selectively retry syncing, or change the value that we use from the server (such as for cases when the parameter goes missing from the server or we fail to sync).

I think we need to keep the sync status for now, but if you like this approach I can add it (or a variant) to #5835.

BenHenning:
PTAL @Rd4dev and @adhiamboperes for full review. I will take another coding pass on this tomorrow and ask for a final check from one of you before merging, but I'm hoping to get this merged within the next 2 days.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
BenHenning: Ah, I forgot this code has been removed. I will keep this in mind when reintroducing the remote parameter code in a later PR.

adhiamboperes (approved):
Thanks @BenHenning, these changes look good to me.

Will the documentation be updated by you or by @theayushyadav11 ?

app/src/main/java/org/oppia/android/app/splash/SplashActivityPresenter.kt
BenHenning: Fixed in latest. Decided to split the presenter in half, instead, to make things a lot more reliable. Verified that the app opens without issue and that the splash activity test still passes as expected.

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagBindingModule.kt
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterBindingModule.kt
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
BenHenning: Done in latest.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt
BenHenning: Corrected the documentation in latest since there was more to do than just updating the method reference. This was also the only reference that I found to the old `loadParameters()` name.

domain/src/main/java/org/oppia/android/domain/platformparameter/FeatureFlagsMapBindingModule.kt
BenHenning: Done in latest.

domain/src/test/java/org/oppia/android/domain/exploration/testing/BUILD.bazel
BenHenning: Updated everywhere except the 2 valid prod uses, and updated the corresponding tests to use the test module (since we want everything moving in that direction, anyway). Done in latest.

app/src/test/java/org/oppia/android/app/translation/BUILD.bazel
BenHenning: Done in latest per https://github.com/oppia/oppia-android/pull/5836#discussion_r2141286055.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterModule.kt
Rd4dev: This PR https://github.com/oppia/oppia-android/pull/5845/files#diff-e71269102192b4d7b62db633ed027f8e64e98facc556a5806042865f51c55659 does it by splitting out the controller. Would this be fine or should we have a copy of the Module?

But resolving now for time being, as it's more of a discussion that I can follow up later. Thanks.

domain/src/main/java/org/oppia/android/domain/platformparameter/BUILD.bazel
Rd4dev: Ah, thanks resolving now.

BenHenning:
Deciding to go ahead and push now even though the static check will fail mainly so that I can check against latest develop and self-review the comment changes.

Edit: Ah, just noticed #5849 is now merged so will bring this up-to-date before re-running CI so that hopefully it's fully green and this is merge-ready (aside from comment verification).

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterController.kt
Rd4dev: Thanks.

BenHenning:
> Thanks @BenHenning, these changes look good to me.
> 
> Will the documentation be updated by you or by @theayushyadav11 ?

Thanks @adhiamboperes!

The documentation around writing parameters and testing them will be updated by me, probably in two batches since there are several more changes coming yet (per #5835).

model/src/main/proto/platform_parameter.proto
Rd4dev: Thanks for explaining the details of it, this will be helpful to refer to on future discussions. I do not have full context as of now on considering fitting it with `ParameterValue` or other relevant parts (if would be related in any override portions) as needed. But resolving for now and follow up on future discussions.

BenHenning (commented):
Self-reviewed latest changes, will update PR description shortly and resolve my own review comments.

BenHenning:
PTAL @seanlip for code owners.

@adhiamboperes could you PTAL at the latest changes just to make sure they look good to you? Apologies for the splash activity changes--they are much simpler than they seem the diff just isn't very clean (I moved most of the logic into the new wrapper class and call through to it).

@Rd4dev PTAL and let me know if I can merge this (per your last remaining open comment). I can also address major issues in a follow-up PR if they don't need to block this one.

Otherwise, mainly waiting to see how CI turns out.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterProcessState.kt
Rd4dev: I was initially thinking this might lead to some duplication, especially around the remote DB retrieval logic being repeated in the `DebugImpl`. But either approach is fine. For now, I’ll proceed assuming it will be fully handled within `DebugImpl` (even if it involves duplication) and will follow up with further discussions if needed.

Rd4dev:
> @Rd4dev PTAL and let me know if I can merge this (per your last remaining open comment). I can also address major issues in a follow-up PR if they don't need to block this one.
> 
> Otherwise, mainly waiting to see how CI turns out.

@BenHenning -- thanks, I’ve resolved all my comments for now since they weren’t blockers for merging (most were discussion points rather than critical issues). I’ll follow up in future discussions if any further clarifications are needed to handle things.

Rd4dev:
> Thanks @Rd4dev! As to this, no, platform parameters will also be synced. The `SYNC_STATUS_UNSPECIFIED` is being set right now because we don't actually use sync status for anything except logging feature flags. I recognize this may introduce a problem with #5345, but note that sync status isn't filled in yet, anyway, due to remote parameters being disabled. We can chat further on how to mitigate this better, but I'd rater get this PR merged first.

Thanks, for explaining that.

> I'm not sure what's meant by this, can you clarify?

Oh, that was actually about the [piece of code](https://github.com/oppia/oppia-android/blob/f33ea4b0e94209b718df3f699acb19c9c2e3ebc0/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerProdImpl.kt#L168) that was earlier left in the `ProdImpl`, but it looks like it was cleaned up in the subsequent commits—so outdated now.

adhiamboperes (approved):
I have reviewed the latest changes -- LGTM. This is a really nice improvement on overall clarity of the class. I do expect some conflicts in one of my WIP branches 🥲.

BenHenning:
> I have reviewed the latest changes -- LGTM. This is a really nice improvement on overall clarity of the class. I do expect some conflicts in one of my WIP branches 🥲.

Ah, well fortunately all my other changes are on hold until I'm fully back in ~2 weeks so hopefully only this PR is the one that causes conflicts (and hopefully not too bad :crossed_fingers:).

seanlip (approved):
LGTM for codeowners

Issue #5734, opened 2025-03-04
[Feature Request]: Introduce custom Android Lint wrapper & fix key issue categories (GSoC'25 4.3)


BenHenning:
### Is your feature request related to a problem? Please describe.
The Android build ecosystem supports a [linting tool](https://developer.android.com/studio/write/lint) which catches a variety of Android-specific problems (beyond those that can be caught by more general-purpose linters like [ktlint](https://github.com/pinterest/ktlint)). Historically a Gradle feature, Android Lint is [not yet](https://github.com/bazelbuild/rules_android/issues/321) directly supported in the Bazel world (though there are some [alternatives available](https://github.com/Bencodes/rules_android_lint)). This project involves:
- Introducing an Oppia Android script that runs Android lint directly.
- Supporting the same exemption override support (via XML) but using a textproto interface for parity with the allow-listing and deny-listing configurations used in other Oppia Android scripts.
- Pre-populating the exemption list with existing known failures.
- Adding wiki documentation to explain how to use the new script.
- Filing and/or updating issues with findings such that all remaining Android lint issues are properly tracked in the repository.
- Updating CI to run the Android lint script.
- Fixing the following categories of lint issues:
  - All categories classified as 'errors' by the tool.
  - All of the following 'warning' categories:
    - OldTargetApi
    - UnusedAttribute
    - Typos
    - StringFormatCount
    - ObsoleteSdkInt
    - UnusedResources
    - UselessLeaf
    - UselessParent
    - UnusedIds
    - DuplicateStrings
    - SelectableText
    - LabelFor
    - RtlSymmetry
    - UnknownNullness

### Describe the solution you'd like
**Suggested Milestones:**
- **Milestone 1**:
  - Introduce a new script that can run Android lint checks with support for a local textproto file that can override specific failures to allow them to pass.
  - Check in a populated exemption list with all known exemptions.
  - Introduce a new wiki page explaining how to use the script and update the exemption list.
  - Introduce CI support for the new script so that it runs for every change to develop and for all PRs.

- **Milestone 2**:
  - Fix all of the error categories of lint issues, plus the warning categories mentioned in the project description.
  - Ensure all exemptions that aren't fixed as part of this project have corresponding tracking issues filed and spec'd with enough context for contributors to work on them.

<details>
<summary>Technical hints / guidance</summary>

- General changes explanation:
  - Background on running the linter
[...]

PR #5843, opened 2025-06-05
Fix part of #5734: Introduce LintAnalysisReporter to parse and extract XML files

manas-yu:
## Explanation

Fixes part of #5734 
### Project
[PR 1.2 of project 4.3]

This  PR includes the following updates:
- Introduce `LintAnalysisReporter` to parse and extract XML report files
- Added tests to validate XML parsing logic
## Essential Checklist


manas-yu:
Hi @bhavukJain PTAL, Thanks!

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+    require(xmlFile.exists()) { "Lint report file not found: $xmlFilePath" }
+
+    return try {
+      val docBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder()
+      val doc: Document = docBuilder.parse(xmlFile)
+      doc.documentElement.normalize()
+
+      val issueNodes: NodeList = doc.getElementsByTagName("issue")
+
+      (0 until issueNodes.length).map { index ->
```
bhavukJain: For very large XML files, this creates intermediate collections. Using `asSequence()` from a memory point of view would be better.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
+package org.oppia.android.scripts.lint
+
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+import org.w3c.dom.NodeList
+import java.io.File
+import javax.xml.parsers.DocumentBuilderFactory
+
+/** Enum representing the severity levels of lint issues. */
```
bhavukJain: Would be nice to add KDoc here. Soemthing like:
```

/**
 * Enum representing the severity levels of lint issues.
 * 
 * @property displayName The string representation used in XML reports
 * 
 * Example usage:
 * ```
 * val severity = LintSeverity.fromString("Error") // Returns LintSeverity.ERROR
 * ```
 */
```

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+        parseIssueElement(issueNodes.item(index) as Element)
+      }
+    } catch (e: Exception) {
+      throw IllegalArgumentException("Error processing file $xmlFilePath: ${e.message}")
+    }
+  }
+
+  /**
+   * Parses a single issue element and returns a LintIssue object.
+   * @throws IllegalArgumentException if the issue element is invalid or missing required attributes
```
bhavukJain, line 154: Nit: Would be nice to add enhance this: 

```
/**
 * Parses a single issue element and returns a LintIssue object.
 * 
 * @param issueElement The XML element containing issue data
 * @return A LintIssue object with all parsed data
 * @throws IllegalArgumentException if the issue element is invalid or missing required attributes
 */
 ```

File: scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt [hunk 4]
```diff
@@ -0,0 +1,147 @@
...
+   */
+  private fun parseIssueElement(issueElement: Element): LintIssue {
+    val id = issueElement.getAttribute("id")
+    val severityString = issueElement.getAttribute("severity")
+    val locations = extractLocations(issueElement)
+
+    if (id.isBlank() || severityString.isBlank() || locations.isEmpty()) {
+      throw IllegalArgumentException(
```
bhavukJain: We can make the error messages more specific. Something like: 

```
id.isBlank() -> throw IllegalArgumentException("Issue element missing required 'id' attribute")
severityString.isBlank() -> throw IllegalArgumentException("Issue element missing required 'severity' attribute")
 locations.isEmpty() -> throw IllegalArgumentException("Issue element must contain at least one location")
```

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+   * @return List of LintIssue objects representing all issues found in the report
+   * @throws IllegalArgumentException if file doesn't exist or parsing fails
+   */
+  fun parseLintReport(xmlFilePath: String): List<LintIssue> {
+    val xmlFile = File(xmlFilePath)
+    require(xmlFile.exists()) { "Lint report file not found: $xmlFilePath" }
+
+    return try {
+      val docBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder()
+      val doc: Document = docBuilder.parse(xmlFile)
```
bhavukJain: Should we add validation here to ensure compatibility? Something like 

```
val rootElement = doc.documentElement
if (rootElement.tagName != "issues") {
  throw IllegalArgumentException("Invalid lint report format: expected root element 'issues'")
}
```

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
```diff
@@ -0,0 +1,467 @@
...
+
+/** Tests for [LintAnalysisReporter]. */
+// FunctionName: test names are conventionally named with underscores.
+@Suppress("FunctionName")
+class LintAnalysisReporterTest {
+
+  @field:[Rule JvmField] val tempFolder = TemporaryFolder()
+
+  private lateinit var lintAnalysisReporter: LintAnalysisReporter
+  private val originalOut = System.out
```
bhavukJain, line 23: Can we move the common test data (warningIssue, errorIssue, etc.) to a companion object? This makes it clear that these are shared test fixtures.

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
```diff
@@ -0,0 +1,467 @@
...
+import org.oppia.android.testing.assertThrows
+import java.io.ByteArrayOutputStream
+import java.io.File
+import java.io.PrintStream
+import java.lang.IllegalArgumentException
+
+/** Tests for [LintAnalysisReporter]. */
+// FunctionName: test names are conventionally named with underscores.
+@Suppress("FunctionName")
+class LintAnalysisReporterTest {
```
bhavukJain, line 18: Would be good to add test for empty locations list handling. This is to verify that locations with empty file attributes are properly filtered out.

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
```diff
@@ -0,0 +1,467 @@
...
+    assertThat(issue.locations[1].file)
+      .isEqualTo("app/src/main/res/values-night/color_palette.xml")
+    assertThat(issue.locations[1].lineNumber).isEqualTo("159")
+  }
+
+  @Test
+  fun testParseLintReport_emptyXml_returnsEmptyList() {
+    val xmlContent =
+      """
+      <issues format="6" by="lint 7.3.1">
```
bhavukJain: Instead of repeating these, we can extract them out as constants in the companion object. 
```
companion object {
  private const val XML_HEADER = """<issues format="6" by="lint 7.3.1">"""
  private const val XML_FOOTER = """</issues>"""
}
```

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+   * Parses an XML lint report file and returns a list of lint issues.
+   *
+   * @param xmlFilePath Path to the XML lint report file
+   * @return List of LintIssue objects representing all issues found in the report
+   * @throws IllegalArgumentException if file doesn't exist or parsing fails
+   */
+  fun parseLintReport(xmlFilePath: String): List<LintIssue> {
+    val xmlFile = File(xmlFilePath)
+    require(xmlFile.exists()) { "Lint report file not found: $xmlFilePath" }
+
```
bhavukJain, line 112: Do you think we should add a file size check here? Or maybe when you experiment with project-description later on, we can add if need be. Something like:

```
  val maxFileSize = 50 * 1024 * 1024 // 50MB
  require(xmlFile.length() <= maxFileSize) { 
    "Lint report file too large: ${xmlFile.length()} bytes (max: $maxFileSize)" 
  }
```

This would help to avoid memory issues with very large XML files.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Parses an XML lint report file and returns a list of lint issues.
+   *
+   * @param xmlFilePath Path to the XML lint report file
+   * @return List of LintIssue objects representing all issues found in the report
+   * @throws IllegalArgumentException if file doesn't exist or parsing fails
+   */
+  fun parseLintReport(xmlFilePath: String): List<LintIssue> {
+    val xmlFile = File(xmlFilePath)
```
bhavukJain: Do you think it will be a good idea to add a cache here? If this parser will be used multiple times on the same files, it can be a good idea. 

```
private val cache = mutableMapOf<String, List<LintIssue>>()
```

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,147 @@
...
+    require(xmlFile.exists()) { "Lint report file not found: $xmlFilePath" }
+
+    return try {
+      val docBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder()
```
bhavukJain: While DocumentBuilder doesn't require explicit cleanup, I think we should still wrap this in a use block for consistency with other resource management patterns in the codebase (Eg: PersistentCacheStore.kt, FeedbackReportingServiceTestOrchestrator.kt, AssetRepositoryImpl.kt). This would make the code more defensive against future changes.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Updated.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Thanks, Updated.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Thanks, this would be great. I updated it in other places as well.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Updated.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Thanks updated.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Thanks, this would be great. We could also optimize it later with actual report sizes. Added.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: We can maintain the files that the script requires in the system's default temporary file directory. This would eliminate the need to regenerate the files and allow us to use caching techniques. This way, we would also have the option to set up partial-result directories for lint to support incremental analysis in lint while implementing the project description. I have updated the code to support caching of XML reports.

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
manas-yu: Added `testParseLintReport_issueWithEmptyFileAttributes_filtersOutEmptyLocations` and `testParseLintReport_issueWithOnlyEmptyFileAttributes_throwsException`

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
manas-yu: Done.

scripts/src/javatests/org/oppia/android/scripts/lint/LintAnalysisReporterTest.kt
manas-yu: Added.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Updated the kdoc. Not sure if example usage is needed since the fromString function already has its own kdoc.

manas-yu:
Hi @bhavukJain, PTAL. All comments have been addressed.

bhavukJain:
@adhiamboperes I have reviewed the PR. Can you also take a look? Thanks!

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,199 @@
...
+  WARNING("Warning"),
+  /** Represents Lint issue of severity Information. */
+  INFORMATION("Information");
+
+  companion object {
+    /**
+     * Converts a string severity to enum, case-insensitive.
+     * @param severityString the string representation of severity
+     * @return the corresponding LintSeverity enum
+     * @throws IllegalArgumentException if severity is unknown
```
adhiamboperes: Please reformat KDoc correctly.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -0,0 +1,199 @@
...
+   *
+   * @param xmlFilePath Path to the XML lint report file
+   * @return List of LintIssue objects representing all issues found in the report
+   * @throws IllegalArgumentException if file doesn't exist or parsing fails
+   */
+  fun parseLintReport(xmlFilePath: String): List<LintIssue> {
+    val xmlFile = File(xmlFilePath).absoluteFile
+    check(xmlFile.exists()) { "Lint report file not found: $xmlFilePath" }
+
+    val maxFileSize = 50 * 1024 * 1024 // 50MB
```
adhiamboperes, line 102: Curious: What happens when the max file size is exceeded? Is there a plan for a fallback?

adhiamboperes (commented):
Thanks @manas-yu! 

I have a couple of minor comments. PTAL.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Done.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: This would be a good check since the size limit is well above typical report sizes. It ensures the script fails for unusually large or corrupted files, preventing excessive memory use, or crashes during XML parsing. So makes sense to fail the script without a fallback.

manas-yu:
Hi @adhiamboperes PTAL, Thanks.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
adhiamboperes: Sounds good.

adhiamboperes (approved):
LGTM, thanks!

PR #1789, opened 2020-09-04
Fix #1788: Prepare build for production

BenHenning:
Fixes #1788.

This PR does a few things:
- Increase build resources to max 1GB of ram (needed for running the R8 optimization step)
- Enable minification using R8 for release builds
- Enable resource shrinking using AGP (see https://developer.android.com/studio/build/shrink-code)
- Add support for split APKs (see https://developer.android.com/studio/build/configure-apk-splits); we'll probably be using app bundles for the actual upload, but this lets us analyze the APK sizes people will actually install for a given device type

No additional proguard work is needed. I verified that the universal APK on my Pixel XL using a test signature (since the release flavors are not signed by default) is working.

Note that this removes two of the min sdk declarations that break both Gradle syncing & APK signing. AGP is a bit sensitive about this min sdk property even though it's valid, so I'm hoping Bazel will build without issues. Long-term we will define min SDK in the manifest when AGP is out of the picture.

PR #5874, opened 2025-06-30
Fix part of #5734: Final configurations for Lint tool and Added tests

manas-yu:
## Explanation

Fixes part of #5734 

Project
[PR 1.7 of project 4.3]

- Hook utilities to AndroidLintRunner to generate reports
- Add tests to validate the accuracy of the script in AndroidLintRunnerTest
- Updated AndroidManifest.xml files to include `minSdkVersion` and `targetSdkVersion` following #1789, since it helps with improving the Lint Reports accuracy, and we have completed Bazel migration.

## Essential Checklist


manas-yu:
Hi @bhavukJain PTAL. Thanks!

app/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
   xmlns:tools="http://schemas.android.com/tools"
   package="org.oppia.android">
-
+  <uses-sdk android:minSdkVersion="21"
+      android:targetSdkVersion="34" />
```
bhavukJain, line 6: Do we need to specify the min and targetSDK in multiple AndroidManifest.xml? Is there a way to centralize this?

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
```diff
@@ -401,6 +421,11 @@ private class ModuleConfigurationBuilder(
       lintCheckJars = dependencyResolver.extractLintCheckJars(
         dependencyResolver.resolveAarFiles(module)
       ),
+      partialResultsDir = partialResultDir,
+      annotationZips = dependencyResolver.extractAnnotationZips(
```
bhavukJain: It would be good to add error handling here for corrupted AAR files during annotation extraction. Soemthing like: 

```
annotationZips = try {
  dependencyResolver.extractAnnotationZips(
    dependencyResolver.resolveAarFiles(module)
  )
} catch (e: Exception) {
  logger.warning("Failed to extract annotation zips: ${e.message}")
  emptyList()
},
```

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
```diff
@@ -377,21 +378,8 @@ class LintModelCreator(
...
-      ?.map { it.absolutePath }
-      ?: emptyList()
-
-    return if (proguardFiles.isNotEmpty()) {
-      """proguardFiles="${proguardFiles.joinToString(",")}" """
-    } else {
-      ""
-    }.trimEnd()
+  private fun createProguardAttribute(proGuardFiles: List<String>): String {
+    return """proguardFiles="${proGuardFiles.joinToString(FILE_SEPARATOR)}" """
```
bhavukJain: Would be a good idea to filter out non-existent ProGuard files to prevent lint failures. 

```
  val existingFiles = proGuardFiles.filter { File(it).exists() }
```

scripts/src/java/org/oppia/android/scripts/lint/LintCommon.kt
```diff
@@ -46,8 +46,11 @@ data class ModuleConfig(
   val dependencies: List<String>,
   val aarFiles: List<AarFileInfo>,
   val jarFiles: List<String>,
+  val annotationZips: List<String>,
   val lintCheckJars: List<String>,
-  val lintModelDir: File? = null
+  val lintModelDir: File? = null,
+  val partialResultsDir: File,
+  val proGuardFiles: List<String>
```
bhavukJain, line 53: Would it be a good idea to add validation for directory and file existence in the init block for this? 

Something like: 
```
  init {
    require(partialResultsDir.canWrite() || partialResultsDir.mkdirs()) {
      "Cannot create or write to partialResultsDir: ${partialResultsDir.absolutePath}"
    }
   // same for proguard files
  }
}

File: scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt [hunk 5]
```diff
@@ -26,11 +30,15 @@ class AndroidLintRunnerTest {
   private lateinit var sdkPath: String
   private lateinit var jdkHome: File
   private lateinit var buildSdkVersion: String
+  private val fakeCommandExecutor by lazy { FakeCommandExecutor() }
+  private lateinit var androidLintAnalyzerWithFakeExecutor: AndroidLintAnalyzer
+  private lateinit var workingDirectory: File
 
   companion object {
+    private const val JAVA_VERSION = "11.0.6"
```
bhavukJain, line 41: Instead of hardcoded version, is it possible to do AndroidBuildSdkProperties().javaVersion?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -318,465 +353,604 @@ class AndroidLintRunnerTest {
...
-    )
+    androidLintAnalyzerWithFakeExecutor.runAnalysis()
+
+    val output = outputStream.toString()
+    assertThat(output).contains("NotifyDataSetChanged")
+    assertThat(output)
+      .contains("notifyDataSetChanged()")
+    assertThat(output).contains("Line: 13")
+    assertThat(output)
+      .contains(
```
bhavukJain, line 639: Would be good to also verify that lint rule configuration is properly loaded in test? Something like the following: 

```
  val projectDescription = File(workingDirectory, "lint-project-description.xml")
  assertThat(projectDescription.text()).contains("NotifyDataSetChanged") 
```

Feel free to modify the above for reading text from projectDescription but I hope you get the idea.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
```diff
@@ -245,13 +245,13 @@ class LintModelCreator(
     val content =
       """
         <lint-module
-            dir="${escapeXmlAttribute(relativeProjectPath.toString())}"
+            dir="${escapeXmlAttribute(modulePath.toString())}"
             name="${escapeXmlAttribute(moduleConfig.name)}"
             type="${moduleType.name}"
-            maven="__non_maven__"
             buildFolder="${escapeXmlAttribute(buildDir.toFile().absolutePath)}"
```
bhavukJain: Should we ensure build directory exists before referencing it here? 

Double check this but I think if we just do `buildDir.createDirectories().toFile()`, it will automatically create (if not present) or use the same one if present.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
```diff
@@ -377,21 +378,8 @@ class LintModelCreator(
...
-      ?.listFiles { file -> file.name.endsWith(".pro") }
-      ?.map { it.absolutePath }
-      ?: emptyList()
-
-    return if (proguardFiles.isNotEmpty()) {
-      """proguardFiles="${proguardFiles.joinToString(",")}" """
-    } else {
-      ""
-    }.trimEnd()
+  private fun createProguardAttribute(proGuardFiles: List<String>): String {
```
bhavukJain: When proGuardFiles.isEmpty, do we want to return empty string ("") or empty proguardFiles attribute?

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
```diff
@@ -43,9 +43,9 @@ class LintModelCreator(
     private const val MIN_SDK_VERSION = "21"
     private const val TARGET_SDK_VERSION = "34"
 
-    private const val PROGUARD_CONFIG_PATH = "config/proguard"
-
     private const val MODEL_CACHE_TTL_HOURS = 24L
```
bhavukJain: Where is this used?

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
```diff
@@ -511,6 +557,13 @@ private class DependencyResolver(
       if (lintJar.exists()) lintJar.absolutePath else null
     }
 
+  /** Extracts annotation zip files from the given list of AAR files. */
+  fun extractAnnotationZips(aarFiles: List<AarFileInfo>): List<String> =
+    aarFiles.mapNotNull { aarInfo ->
```
bhavukJain, line 580: Should we validate AAR extracted path exists before checking for annotations.zip?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -816,12 +990,242 @@ class AndroidLintRunnerTest {
...
+      reportFile = reportFile,
+      projectDescriptionFile = projectDescriptionFile
+    )
+  }
+
+  private fun setupFakeCommandExecutor(aarPath: String, jarPath: String) {
+    fakeCommandExecutor.registerHandler("bazel") { _, args, outputStream, _ ->
+      when {
+        args.contains("cquery") && args.contains("deps(//app:*)") -> {
+          outputStream.println(aarPath)
```
bhavukJain: Should we make fake Bazel output match real format with target labels? Something like: 

```
        outputStream.println("@maven//:$aarPath")
        outputStream.println("@maven//:$jarPath")
```

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -318,465 +353,604 @@ class AndroidLintRunnerTest {
...
+      .contains(
+        "Attribute `android:theme` is only used by `<include>` tags "
+      )
   }
 
-  private fun createLintRunner(): AndroidLintRunner {
-    val reportFile = File(tempFolder.root, "lint-report.xml")
-    val projectDescriptionFile = createProjectDescriptionFile()
+  @Test
+  fun testAndroidLintAnalyzer_withNotifyDataSetChanged_detectsIssue() {
```
bhavukJain, line 628: Should this test expect failure when lint issues are detected?

app/src/main/AndroidManifest.xml
manas-yu: It's meant to be specified in all the AndoridManifest.xml files according to the Wiki [[wiki](https://github.com/oppia/oppia-android/blob/43a08262fa9d5abddb9f0c5848aacbe6ca9af680/wiki/Upgrading-Target-Sdk-Guide.md?plain=1#L136)]

bhavukJain: Ack, sounds good!

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
manas-yu: Done.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
manas-yu: Done.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
manas-yu: It's being used in `isCacheValid` to check for TTL

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
manas-yu: It's the same thing but i have included an empty list check.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: We are using BazelInfo to get the information related to the java configs like java_home and java version which is a better approach than `AndroidBuildSdkProperties` from my understanding and `private const val JAVA_VERSION = "11.0.6" ` is just acting as a mock for `fakeCommandExecutor` so should be alright.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
manas-yu: used `buildDir.createDirectories().toFile()` to ensure the directory exists before referencing it.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Updated to match the real format the query returns relative paths starting with `bazel-bin/` or `external/` so matched the similar output.

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
manas-yu: Added check.

scripts/src/java/org/oppia/android/scripts/lint/LintCommon.kt
manas-yu: Added checks in init{}.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Failure will only happen when the severity of the issue is Error/Fatal or if the issue won't be present in the exemption Enum/mapping

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Added to tests.

manas-yu:
@adhiamboperes PTAL all comments are addressed.
Cc: @bhavukJain

manas-yu:
Will resolve conflicts and reassign.

manas-yu:
@adhiamboperes PTAL resolved conflicts. Thanks!

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
```diff
@@ -43,9 +43,12 @@ class LintModelCreator(
     private const val MIN_SDK_VERSION = "21"
     private const val TARGET_SDK_VERSION = "34"
 
-    private const val PROGUARD_CONFIG_PATH = "config/proguard"
+    // Model directories are stable and won't change frequently therefore a longer TTL
+    private const val MODEL_CACHE_TTL_HOURS = 24L // 24 hours
```
adhiamboperes, line 47: Aside from TTL, how else can the cache be deleted?

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
```diff
@@ -443,11 +474,16 @@ private class ModuleConfigurationBuilder(
...
 private class SourceFileCollector(
-  repoRoot: File,
+  private val repoRoot: File,
   module: ModuleName
 ) {
   companion object {
     private val SOURCE_EXTENSIONS = setOf("kt", "java")
+    private const val PROGUARD_CONFIG_PATH = "config/proguard"
+
+    // Lint logs that file does not appear to be in the right project location
```
adhiamboperes: This comment is not clear to me

adhiamboperes (commented):
Thanks @manas-yu. PTAL at my comments.

scripts/src/java/org/oppia/android/scripts/lint/LintModelCreator.kt
manas-yu: The cache is automatically invalidated when any input that would affect lint model generation changes. The `computeBazelInputsHash()` method would track these inputs and generate a hash. If the current hash doesn't match the stored hash, the cache is considered invalid and will be regenerated.

The cache can be deleted manually, too, by deleting the `.lint-model-cache` in the model directory.

Also, If any of the required model files are deleted then also the cache will be considered invalid.

scripts/src/java/org/oppia/android/scripts/lint/LintProjectDescription.kt
manas-yu: I have updated the comment to be more descriptive. The file is removed from the source set as the lint tool wasn't running on that file, and instead logged it to not be in the correct location, and as it was also excluded from the gradle source set, we have explicitly exempted it here as well. [[Ref](https://github.com/oppia/oppia-android/blob/925f80c2544620338da3f0d1c21dbdef091682b3/app/build.gradle#L98)]

manas-yu:
@adhiamboperes PTAL. Thanks!

adhiamboperes (commented):
@manas-yu, while these changes LGTM, I am wondering how last night’s discussion impacts the project overall, in terms of using the term modules. Ben mentioned that once modularization work is complete, we will lose the top level targets/directories.

manas-yu:
Hi @adhiamboperes,

Since the concern regarding replacing the modular approach with a single project-level module is more of a long-term consideration and would require experimentation to understand how the lint tool reacts to it, I think we could proceed with merging the current PRs. As Ben mentioned, the goal right now is to get the current approach working and iteratively improve it afterward.

Switching to the alternate approach at this point, especially with the ongoing AAR and JAR generation blocker, could delay the evaluation, which I believe is due on the 12th.

That said, I do have an idea for implementing the other approach as well. It would primarily involve changes to just one utility in the script, namely `LintProjectDescription`. Given the way the current code is structured, it should be relatively easy to experiment with once I resolve the AAR and JAR issue, which would also be a blocker in the alternate approach, and when the current PRs are merged

adhiamboperes:
Thanks for clarifying, @manas-yu !  I am happy to approve the PR.

adhiamboperes (approved):
Thanks @manas-yu! LGTM.

adhiamboperes:
@BenHenning, PTAL for owners.

BenHenning (approved):
Approved the top-level manifest change. I didn't review the PR in detail.

It is slightly amusing to me this is needed, though. It's a case where the linter is wrong, but that's fine. The final merged manifest will put in the `uses-sdk` line as part of Bazel's generation of the merged manifest (since we control these values via Bazel). I _think_ Bazel will always override them, as well, so the declarations basically don't serve a purpose other than silencing the linter warning. Perhaps we'll just disable that check in the future, but that's not important at this stage.

manas-yu:
> It is slightly amusing to me this is needed, though.

I tried passing the context of `minSdkVersion` and `targetSdkVersion` via the model files of lint tool and also even hardcoding the path of the merged manifest generated by Bazel which included the `uses-sdk` line to the lint tool,  but it still relied on the `AndroidManifest.xml` files that are passed to the modules, hence I had to make the changes.

PR #5878, opened 2025-07-02
Fix part of #5734: Resolve issues UseRequireInsteadOfGet and UseAppTint

manas-yu:
## Explanation

Fixes part of #5734 

Project
[PR 1.11 of project 4.3]

- Resolved issue UseAppTint
   - BindingAdapter for `app:tint` support  (ref: https://stackoverflow.com/questions/65058461/set-the-tint-of-an-imageview-using-databinding)
- Resolved issue UseRequireInsteadOfGet
  - Updated `SurveyFragment` with proper null checks instead of `arguments!!`
  - Updated `LogoutDialogFragment` to use `requireActivity()` instead of `activity!!`

## Essential Checklist


manas-yu:
Hi @adhiamboperes PTAL.
Cc: @bhavukJain

app/src/main/res/layout/add_profile_activity.xml
```diff
@@ -114,7 +114,7 @@
             android:layout_height="48dp"
             android:contentDescription="@string/add_profile_info_content_description"
             android:padding="12dp"
-            android:tint="@color/component_color_add_profile_activity_info_icon_color"
+            app:tint="@color/component_color_add_profile_activity_info_icon_color"
```
bhavukJain, line 117: For all XML files, it will be good to add a comment "" to clarify that this is a custom attribute handled by a BindingAdapter, not a standard Android attribute.Explanation: This helps future maintainers understand why app:tint is used instead of android:tint.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -256,6 +256,32 @@ class SurveyFragmentTest {
     }
   }
```
bhavukJain: Should we also add a test where arguments is present but missing required keys. This will help ensure `requireArguments().getProto()` and `requireArguments().extractCurrentUserProfileId()` both fail with clear error messages.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
```diff
@@ -64,12 +64,12 @@ class SurveyFragment :
...
   ): View? {
 
-    val args = arguments!!.getProto(
+    val args = requireArguments().getProto(
       SURVEY_FRAGMENT_ARGUMENTS_KEY,
       SurveyFragmentArguments.getDefaultInstance()
     )
 
-    val internalProfileId = arguments!!.extractCurrentUserProfileId().internalId
+    val internalProfileId = requireArguments().extractCurrentUserProfileId().internalId
```
bhavukJain: I think `requireArguments()` will throw if the bundle is missing, but if a single key is missing, it will result in less clear errors. Do you think we should add explicit error handling or a custom error message if the required proto or profile ID is missing from arguments?

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
```diff
@@ -35,7 +35,7 @@ class LogoutDialogFragment : InjectableDialogFragment() {
         dialog.dismiss()
       }
       .setPositiveButton(R.string.log_out_dialog_okay_button) { _, _ ->
-        val intent = ProfileChooserActivity.createProfileChooserActivity(activity!!)
+        val intent = ProfileChooserActivity.createProfileChooserActivity(requireActivity())
```
bhavukJain, line 38: If `requireActivity()` is called when the fragment is not attached, it throws an IllegalStateException with a generic message. This can be hard to trace in crash logs, especially if multiple fragments use this pattern. 

Would be good to do something like this: 
```
val intent = try {
    ProfileChooserActivity.createProfileChooserActivity(requireActivity())
} catch (e: IllegalStateException) {
    throw IllegalStateException("LogoutDialogFragment: requireActivity() failed because the fragment was not attached. Cannot start ProfileChooserActivity.", e)
}
startActivity(intent)
```

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
adhiamboperes: The most widespread approach we use is the `checkNotNull` function, e.g.


```
val profileId =
      checkNotNull(arguments?.extractCurrentUserProfileId()) {
        "Expected profileId to be included in the arguments for IntroFragment."
      }
```

app/src/main/res/layout/add_profile_activity.xml
adhiamboperes: I think adding a comment in the PR description is enough, we can always see that when viewing git history. I feel like a comment in all the xml files is not ideal.

adhiamboperes (commented):
Thanks @manas-yu, I’d be happy to approve this once you address Bhavuk’s comments.

app/src/main/res/layout/add_profile_activity.xml
manas-yu: The PR description already includes the change along with the ref so considering this addressed.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
manas-yu: Thanks Updated.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
manas-yu: Updated the `testFragment_fragmentLoaded_verifyCorrectArgumentsPassed` to include all the checkNotNull cases similar to other FragmentTests.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
manas-yu: Yup, thanks, followed the checks from other files and added explicit error messages accordingly.

manas-yu:
@adhiamboperes PTAL all comments are addressed.
Cc: @bhavukJain

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
```diff
@@ -63,15 +63,26 @@ class SurveyFragment :
     container: ViewGroup?,
     savedInstanceState: Bundle?
   ): View? {
+    val arguments = checkNotNull(arguments) {
+      "Expected arguments to be passed to StoryFragment."
```
adhiamboperes:
```suggestion
      "Expected arguments to be passed to SurveyFragment."
```

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -276,8 +276,9 @@ class SurveyFragmentTest {
         }
 
         assertThat(exception).isNotNull()
-        assertThat(exception!!.message).contains("Fragment SurveyFragment")
-        assertThat(exception.message).contains("does not have any arguments.")
+        assertThat(exception!!.message).contains(
+          "Expected arguments to be passed to StoryFragment."
```
adhiamboperes: Ditto here

adhiamboperes (changes requested):
Thanks @manas-yu, a minor correction.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
manas-yu: Ah Thanks. fixed!

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
manas-yu: fixed.

manas-yu:
@adhiamboperes PTAL.

adhiamboperes (approved):
LGTM, thanks @manas-yu!

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -530,14 +557,27 @@ class SurveyFragmentTest {
 
         val surveyFragment = activity.supportFragmentManager
           .findFragmentById(R.id.survey_fragment_placeholder) as SurveyFragment
-        val args = surveyFragment.arguments!!.getProto(
+
+        val arguments = checkNotNull(surveyFragment.arguments) {
+          "Expected arguments to be passed to SurveyFragment."
+        }
```
BenHenning: Could we use `?` instead of checking for null? It's generally a bit clearer for failures to use ``assertThat()`` for validating something is explicitly not null which means we sometimes end up with a bunch of `?.` calls, which is fine since Truth is Java-based and automatically takes care of the null failure case for us.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -256,6 +256,33 @@ class SurveyFragmentTest {
...
+  @Test
+  fun testSurveyFragment_withNullArguments_throwsIllegalStateException() {
+    val fragment = SurveyFragment()
+
+    launch<SurveyActivity>(
+      createSurveyActivityIntent()
+    ).use { scenario ->
+      scenario.onActivity { activity ->
+        val exception = try {
+          fragment.onCreateView(
```
BenHenning: This feels invalid--we shouldn't ever be calling lifecycle methods directly in tests as that can create behavioral skew with real Android. Instead, we should use the Jetpack testing library to orchestrate behaviors that lead to the same outcomes.

In this case, wouldn't simply adding the fragment to the activity trigger the exception?

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
```diff
@@ -63,15 +63,26 @@ class SurveyFragment :
...
+      "Expected profileId to be included in the arguments for SurveyFragment."
+    }
+    val internalProfileId = profileId.internalId
+
+    val topicId = checkNotNull(args.topicId) {
+      "Expected topicId to be included in the arguments for SurveyFragment."
+    }
+    val explorationId = checkNotNull(args.explorationId) {
+      "Expected explorationId to be included in the arguments for SurveyFragment."
+    }
```
BenHenning: Are these valid? `args` can't be null here, and proto never returns null for these either since it defaults. In fact, I assume only `arguments` needs to be checked for nullness.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
```diff
@@ -35,7 +35,15 @@ class LogoutDialogFragment : InjectableDialogFragment() {
         dialog.dismiss()
       }
       .setPositiveButton(R.string.log_out_dialog_okay_button) { _, _ ->
-        val intent = ProfileChooserActivity.createProfileChooserActivity(activity!!)
+        val intent = try {
+          ProfileChooserActivity.createProfileChooserActivity(requireActivity())
+        } catch (e: IllegalStateException) {
```
BenHenning: We should generally never be catching an ``IllegalStateException``. Why can't we just let the underlying `require` throw?

BenHenning (changes requested):
Thanks @manas-yu. I had some thoughts--PTAL.

BenHenning:
Also @manas-yu please make sure you're fully filling in the PR description.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
manas-yu: Updated test to use `assertThat().isNotNull()` instead.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
manas-yu: Updated the test to avoid manually calling `onCreateView()`.

app/src/main/java/org/oppia/android/app/survey/SurveyFragment.kt
manas-yu: Removed unnecessary checks, only keeping the check for `arguments`.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
manas-yu: Bhavuk suggested that the message thrown by `requireActivity()` might be generic:
 `throw new IllegalStateException("Fragment " + this + " not attached to an activity.");` therefore hard to trace in crash logs [[link to comment](https://github.com/oppia/oppia-android/pull/5878#discussion_r2183412861)]. I'm open to making changes as required, so a confirmation would be nice.

manas-yu:
@BenHenning PTAL, I've addressed all comments.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
BenHenning: I agree with the principle, but in this case the new message isn't adding anymore context. We can easily inspect the stack trace and see that this is coming from this line and `LogoutDialogFragment`, ergo, it's pretty clear what's failing.

Where messages are important is when we can add additional context, like variables, which we can't inspect from the stack trace alone.

/cc @bhavukJain

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
```diff
@@ -546,6 +554,29 @@ class SurveyFragmentTest {
...
+
+    val exception = assertThrows<IllegalStateException> {
+      launch<SurveyActivity>(
+        createSurveyActivityIntent()
+      ).use { scenario ->
+        scenario.onActivity { activity ->
+          val fragmentTransaction = activity.supportFragmentManager.beginTransaction()
+          fragmentTransaction.add(R.id.survey_fragment_placeholder, fragment, "test_fragment")
+          fragmentTransaction.commit()
+          activity.supportFragmentManager.executePendingTransactions()
```
BenHenning: Nit: can use `commitNow` instead and do this in one statement.

BenHenning (approved):
Thanks @manas-yu. Approving to unblock, but please address the last two comments and have @bhavukJain or @adhiamboperes check and resolved my comments before merging.

app/src/sharedTest/java/org/oppia/android/app/survey/SurveyFragmentTest.kt
manas-yu: Done.

app/src/main/java/org/oppia/android/app/administratorcontrols/LogoutDialogFragment.kt
manas-yu: removed try catch.

manas-yu:
@adhiamboperes PTAL.

adhiamboperes:
@manas-yu, PTAL for conflicts

adhiamboperes:
@manas-yu, is this PR meant to be opened against develop? There are a lot of conflicts for files not changed in this PR.

manas-yu:
@adhiamboperes Fixed base branch, also do take a look at the open comments from Ben. Thanks!

manas-yu:
@adhiamboperes PTAL.

adhiamboperes (approved):
Thanks @manas-yu! I have verified that the remaining comments have been addressed

Issue #5422, opened 2024-06-07
[enhancement]: Migrate codebase to jetpack compose


itsPronay:
### Is your feature request related to a problem? Please describe.
Migrate codebase to jetpack compose. 
[Jetpack Compose](https://developer.android.com/develop/ui/compose) is the modern toolkit for building native Android UIs. By adopting Compose, Kiwix can simplify UI development, improve maintainability, and ensure the app aligns with the latest Android development practices.

### Describe the solution you'd like

Migration can be carried out in 2 phrases, 
In the first migration, we can migrate screens, and then use them with views 
In the second migration, we can remove Fragments and Navigation components 

To get a full idea of how the migration will take place please view official docs on  [Migration strategy](https://developer.android.com/develop/ui/compose/migrate/strategy)

![image](https://github.com/kiwix/kiwix-android/assets/78101731/e64dc95c-70fb-4cb7-99b7-2a4d6647e59f)

Part of the migration needs to be coming up with naming conventions for classes, functions and components.
Some discussions have been going on about this in https://github.com/oppia/oppia-android/pull/5880/changes#r2719090701, https://github.com/oppia/oppia-android/pull/5825, and https://github.com/oppia/oppia-android/pull/5839

### Additional context

Benefits of Jetpack Compose

- **Simplified UI Development**: Compose allows developers to write less code, reducing boilerplate and making the UI code more intuitive and easier to maintain.

- **Improved Performance**: Compose is designed to be more efficient, which can result in better app performance and responsiveness.
- **Better Integration with Kotlin**: Compose is fully Kotlin-based, providing a seamless development experience with Kotlin features like coroutines and flows.
- **Future-Proof**: Google is actively developing and promoting Compose as the future of Android development, ensuring long-term support and innovation.

itsPronay:
@BenHenning Hey, what do you think about migrating codebase to jetpack compose ? since everyone is adopting it.

BenHenning:
Thanks for filing this @itsPronay! This has been on our list for many years, but we never actually filed an issue, so this is a great one to use as the main feature request.

At a high-level, we definitely want to move to Jetpack Compose in the long-term. Besides the benefits you listed, Jetpack Compose has very specific benefits that I think we're interested in:
- It's a more stable replacement for Android databinding.
- It consolidates code to a single language for most of the app (Kotlin).
- It has a much smaller coding footprint over the much-more-verbose XML.

However, there are some cascading challenges that I see with migrating to Compose:
[...]

PR #5880, opened 2025-07-04
Fix Part of #4938: Create Admin PIN (4/12)

adhiamboperes:
## Explanation

Fixes part of #4938

Introduces a new screen for creating an admin PIN during admin onboarding. This is the last step of admin onboarding.

The PR ensures there are validations to enforce correct PINs that are 5 digits long and that the user inputs and appropriate error messages are retained even on configuration changes.

There are additional changes that clean up the onboarding flow placeholders, and tests have been added where necessary.

## Essential Checklist

## For UI-specific PRs only

|  |  |
|--------------|--------------|
|  |  |
|  |  |
|  |  |


app/src/main/java/org/oppia/android/app/profile/PinSetupActivity.kt
adhiamboperes, line 1: Rename activity and fragment

app/src/main/java/org/oppia/android/app/profile/ProfileChooserFragmentPresenter.kt
```diff
@@ -260,6 +261,7 @@ class ProfileChooserFragmentPresenter @Inject constructor(
         )
       )
     } else {
+      // TODO(#4938): Replace with the redesigned ProfileLoginActivity.
```
adhiamboperes: Ask product for confirmation, and copy verification

adhiamboperes: Ditto above

app/src/main/res/layout/create_profile_fragment.xml
```diff
@@ -132,7 +132,6 @@
       android:id="@+id/onboarding_steps_count"
       style="@style/OnboardingStepCountStyle"
       android:layout_marginBottom="@dimen/phone_shared_margin_medium"
-      android:text="@string/onboarding_step_count_three"
```
adhiamboperes, line 137: Why was this removed - should be reverted since the onboarding steps have changed.

app/src/sharedTest/java/org/oppia/android/app/onboarding/AdminIntroFragmentTest.kt
```diff
@@ -201,7 +200,7 @@ class AdminIntroFragmentTest {
   }
 
   @Test
-  fun testIntroFragment_continueButtonClicked_launchesProfileChooserActivity() {
+  fun testIntroFragment_continueButtonClicked_launchesCreateProfileActivity() {
```
adhiamboperes: Update to point to the pin set up screen.

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -200,9 +200,26 @@ class CreateProfileFragmentTest {
...
+              withText(
+                context.getString(
+                  R.string.onboarding_step_count_four
+                )
+              )
+            )
+          )
+        )
+    }
+  }
```
adhiamboperes: Revert

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -183,7 +183,7 @@ class CreateProfileFragmentTest {
   }
 
   @Test
-  fun testFragment_stepCountText_isDisplayed() {
+  fun testFragment_learnerOnboardingFlow_stepCountThreeText_isDisplayed() {
```
adhiamboperes, line 189: Revert

model/src/main/proto/arguments.proto
```diff
@@ -939,20 +939,6 @@ message OnboardingFragmentStateBundle {
...
-  ProfileType profile_type = 1;
-}
-
-// TODO(#4938): Remove once admin profile onboarding is implemented.
-// Arguments required when creating a new ProfileChooserFragment.
-message ProfileChooserFragmentArguments {
-  // The ProfileType of the new profile as implied by the user's selection.
-  ProfileType profile_type = 1;
-}
-
```
adhiamboperes: Overwrite during conflict resolution.

adhiamboperes (commented):
Self review to refresh memory.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileFragmentPresenter.kt
```diff
@@ -170,7 +170,8 @@ class CreateProfileFragmentPresenter @Inject constructor(
             .setParentScreen(IntroActivityParams.ParentScreen.CREATE_PROFILE_SCREEN)
             .build()
 
-          val learnerIntroIntent = IntroActivity.createIntroActivity(activity, params, profileId)
+          val learnerIntroIntent =
+            IntroActivity.createIntroActivity(activity, params, profileId)
```
adhiamboperes, line 174: Revert formatting change.

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivityPresenter.kt
```diff
@@ -28,20 +27,9 @@ class ProfileChooserActivityPresenter @Inject constructor(
...
-        avatarImagePath = null,
-        colorRgb = -10710042,
-        isAdmin = true
-      )
-    } else {
-      // TODO(#482): Ensures that an admin profile is present.
-      // This can be removed once the new onboarding flow is finalized, as it will handle the creation of an admin profile.
+  fun handleOnCreate(profileId: ProfileId, parentScreen: ParentScreen) {
+    if (!enableOnboardingFlowV2.value) {
+      // In the legacy flow, the default admin account is created here.
```
adhiamboperes: This looks like it should have been updated in the previous PR

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivity.kt
```diff
@@ -30,15 +30,13 @@ class ProfileChooserActivity : InjectableSystemLocalizedAppCompatActivity() {
     super.onCreate(savedInstanceState)
     (activityComponent as ActivityComponentImpl).inject(this)
 
-    val profileChooserActivityParams = intent.getProtoExtra(
+    val parentScreen = intent.getProtoExtra(
       PROFILE_CHOOSER_PARAMS_KEY,
       ProfileChooserActivityParams.getDefaultInstance()
-    )
+    ).parentScreen
```
adhiamboperes, line 36: This looks like it should have been updated in the previous PR

app/src/sharedTest/java/org/oppia/android/app/profile/ProfileChooserFragmentTest.kt
```diff
@@ -1487,62 +1484,6 @@ class ProfileChooserFragmentTest {
     }
   }
 
-  @Test
```
adhiamboperes, line 1490: Deleted tests are a carry over from the previous PR

config/src/java/org/oppia/android/config/platform/feature_flags.textproto
```diff
@@ -56,7 +56,7 @@ feature_flag_definition {
 feature_flag_definition {
   id: ONBOARDING_FLOW_V2
   remote_server_name: "android_enable_onboarding_flow_v2"
-  default_is_enabled: false
+  default_is_enabled: true
```
adhiamboperes: Revert

model/src/main/proto/arguments.proto
```diff
@@ -962,12 +962,8 @@ message OnboardingFragmentStateBundle {
 
 // Params required when creating a new ProfileChooserActivity.
 message ProfileChooserActivityParams {
-  // TODO(#482): Remove once the default admin profile has been removed.
-  // The nickname associated with a newly created profile.
-  string profile_nickname = 1;
-
   // The screen from which the ProfileChooserActivity was opened.
-  ParentScreen parent_screen = 2;
```
adhiamboperes: This looks like it should have been updated in the previous PR

adhiamboperes (commented):
Self-reviewed latest changes.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinActivityPresenter.kt
```diff
@@ -0,0 +1,44 @@
...
+import javax.inject.Inject
+
+/** The presenter for [PinSetupActivity]. */
+@ActivityScope
+class PinSetupActivityPresenter @Inject constructor(
+  private val activity: AppCompatActivity
+) {
+  private companion object {
+    private const val TAG_PIN_SETUP_FRAGMENT = "TAG_PIN_SETUP_FRAGMENT"
+  }
```
adhiamboperes, line 18: Make static global instead of a companion.

app/src/main/java/org/oppia/android/app/profile/PinSetupFragmentPresenter.kt
```diff
@@ -0,0 +1,332 @@
...
+      setContent {
+        MaterialTheme {
+          PinSetupScreen(profileId)
+        }
+      }
+    }
+  }
+
+  @Composable
+  fun PinSetupScreen(profileId: ProfileId) {
```
adhiamboperes: Remove `profileId` and instead make it a global variable.

app/src/main/java/org/oppia/android/app/profile/PinSetupFragmentPresenter.kt
```diff
@@ -0,0 +1,332 @@
...
+        modifier = Modifier.padding(bottom = 16.dp)
+      )
+
+      Row(
+        modifier = Modifier.fillMaxWidth(),
+        horizontalArrangement = Arrangement.SpaceBetween
+      ) {
+        TextButton(
+          onClick = { activity.finish() }
+        ) {
```
adhiamboperes: Put these inside a column for grouping

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -207,8 +206,26 @@ class CreateProfileFragmentTest {
...
+              )
+            )
+          )
+        )
+    }
+  }
+
   @Test
-  fun testFragment_learner_clickContinueButton_filledNickname_launchesLearnerIntroScreen() {
+  fun testLearnerOnboarding_continueClicked_filledNickname_launchesLearnerIntroScreen() {
```
adhiamboperes, line 227: These tests should be in the previous PR.

model/src/main/proto/arguments.proto
```diff
@@ -963,7 +963,7 @@ message OnboardingFragmentStateBundle {
 // Params required when creating a new ProfileChooserActivity.
 message ProfileChooserActivityParams {
   // The screen from which the ProfileChooserActivity was opened.
-  ParentScreen parent_screen = 2;
+  ParentScreen parent_screen = 1;
```
adhiamboperes: Change should be in the previous PR.

model/src/main/proto/arguments.proto
```diff
@@ -974,7 +974,7 @@ message ProfileChooserActivityParams {
     SPLASH_SCREEN = 1;
 
     // Corresponds to the admin intro screen.
-    ADMIN_INTRO_SCREEN = 2;
+    CREATE_PIN_SCREEN = 2;
```
adhiamboperes: Update comment.

adhiamboperes (commented):
Self-reviewed

app/src/main/java/org/oppia/android/app/profile/ProfileChooserActivityPresenter.kt
adhiamboperes: merge conflict fixed.

app/src/sharedTest/java/org/oppia/android/app/profile/PinSetupFragmentTest.kt
```diff
@@ -0,0 +1,602 @@
...
+      testCoroutineDispatchers.runCurrent()
+
+      composeRule
+        .onNodeWithText(context.getString(R.string.pin_setup_activity_mismatch_error))
+        .assertIsDisplayed()
+    }
+  }
+
+  @Test
+  fun testFragment_clickContinue_withMismatchedPins_showsMismatchError() {
```
adhiamboperes: Continue button should be disabled.

adhiamboperes (commented):
Reviewed new tests

BenHenning:
Apologies--I do have a bit of limited time today due to GSoC planning. Will try to look at this tomorrow, or possibly Friday if I can't get to it sooner.

app/src/main/res/layout/pin_setup_activity.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/pin_setup_fragment_placeholder"
```
BenHenning: Here & other layouts: 2 space indentation.

app/src/main/res/values/strings.xml
```diff
@@ -737,4 +737,14 @@
...
+
+  <!-- Pin Setup Activity -->
+  <string name="pin_setup_activity_title">Create Admin PIN</string>
+  <string name="pin_setup_activity_header">Set up a unique PIN to protect the admin controls</string>
+  <string name="pin_setup_activity_message">Administrators are able to control content downloads and manage all profiles on this device</string>
+  <string name="pin_setup_activity_enter_pin_label">Enter a 5-Digit PIN</string>
+  <string name="pin_setup_activity_confirm_pin_label">Confirm the 5-Digit PIN</string>
+  <string name="pin_setup_activity_blank_error">Type a 5-Digit PIN</string>
+  <string name="pin_setup_activity_length_error">Your PIN should be 5 digits long</string>
+  <string name="pin_setup_activity_mismatch_error">Please make sure that both PINs match</string>
```
BenHenning: I suggest clarifying 'admin' in these string names.

app/src/main/java/org/oppia/android/app/profile/PinSetupActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.oppia.android.app.activity.ActivityComponentImpl
+import org.oppia.android.app.activity.InjectableAutoLocalizedAppCompatActivity
+import org.oppia.android.app.model.ProfileId
+import org.oppia.android.app.model.ScreenName.PIN_SETUP_ACTIVITY
+import org.oppia.android.util.logging.CurrentAppScreenNameIntentDecorator.decorateWithScreenName
+import org.oppia.android.util.profile.CurrentUserProfileIdIntentDecorator.decorateWithUserProfileId
+import org.oppia.android.util.profile.CurrentUserProfileIdIntentDecorator.extractCurrentUserProfileId
+import javax.inject.Inject
+
+/** Activity for displaying the PIN setup screen. */
```
BenHenning: I wonder if these class names & descriptions should perhaps be specialized for admins to emphasize that use case?

config/src/java/org/oppia/android/config/platform/feature_flags.textproto
```diff
@@ -51,7 +51,7 @@ feature_flag_definition {
...
```
BenHenning: Revert. :)

If it helps you can just enable this in the dev options menu so that it sticks, though I suppose for this particular feature you may be resetting the app data a lot which means it wouldn't stick.

model/src/main/proto/arguments.proto
BenHenning: Not done, I think.

app/src/sharedTest/java/org/oppia/android/app/onboarding/AdminIntroFragmentTest.kt
```diff
@@ -209,22 +201,15 @@ class AdminIntroFragmentTest {
...
 
       testCoroutineDispatchers.runCurrent()
 
-      val expectedParams = ProfileChooserActivityParams.newBuilder()
-        .setParentScreen(ProfileChooserActivityParams.ParentScreen.ADMIN_INTRO_SCREEN)
-        .build()
-
-      intended(hasComponent(ProfileChooserActivity::class.java.name))
-      intended(hasProtoExtra(PROFILE_CHOOSER_PARAMS_KEY, expectedParams))
-      intended(hasExtraWithKey(PROFILE_ID_INTENT_DECORATOR))
```
BenHenning, line 227: Probably still valuable to validate the activity extras payload.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,530 @@
...
+    val focusRequester = remember { FocusRequester() }
+    val keyboardController = LocalSoftwareKeyboardController.current
+    val orientation = LocalConfiguration.current.orientation
+    val isPortrait = orientation == Configuration.ORIENTATION_PORTRAIT
+    val stepCountIsVisible by remember(orientation) {
+      derivedStateOf { isPortrait }
+    }
+
+    // PinSetupUiStateSaver uses Pascal case because it's a singleton functional object, and
+    // Kotlin’s naming conventions recommend PascalCase for anything that behaves like a singleton.
```
BenHenning: This seems like a strange place to document this. I don't disagree with it, but it feels like something we should be enforcing generally.

I recommend keeping this text but filing and attaching a TODO to it that tracks figuring out how to normalize this pattern and enforce it. Specifically:
- Normalizing means determining what cases will this come up, and whether we can standardize maybe with helpers or a wrapper class to handle these more generally.
- Enforcing probably means using static checks to fail when someone doesn't use the standardized patterns.

If we expect this to happen a _lot_ then it may be hard to standardize it and probably warrants further discussion.

Also, it's fine to point this at the Jetpack Compose tracking issue and update that, instead, if you want to keep everything tied to that issue (which isn't unreasonable--we can split it up later).

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,530 @@
...
+      pin != confirmPin -> {
+        PinValidationResult(
+          isValid = false,
+          errorMessage = resourceHandler.getStringInLocaleWithWrapping(
+            R.string.pin_setup_activity_mismatch_error
+          )
+        )
+      }
+
+      else -> {
```
BenHenning: For safety I suggest reversing this: have it check correctness at the very beginning (including a non-empty check as part of that), then other specific errors, then default to the mismatch error (so that we're always defaulting to an error). It's a bit more of a 'defense-in-depth'-type pattern which is nice when dealing with sensitive paths like auth. If we make a mistake we want to try and coerce code into a failing state when possible.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,530 @@
...
+  }
+
+  companion object {
+    /** The required length for admin PINs. */
+    private const val ADMIN_PIN_LENGTH = 5
+  }
+}
+
+/** Data class to encapsulate the PIN setup UI state. */
+data class PinSetupUiState(
```
BenHenning: Do this & below actually need to be public?

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,530 @@
...
+        MaterialTheme {
+          PinSetupScreen()
+        }
+      }
+    }
+  }
+
+  @OptIn(ExperimentalComposeUiApi::class)
+  @Composable
+  fun PinSetupScreen() {
```
BenHenning: First, I think we need to revise this a bit. At the moment there's a lot of conflation between the layout structure and logic tied to the layout (such as the `onValueChange` bits). We ought to keep a clean separation between the structure (data) and behaviors (logic). Since Compose makes it easy to couple these together, for now I suggest moving everything logic-esque to separate functions and reference them rather than inline them as things are done now. I'm hoping that will lead to much leaner compositions to make it a lot easier to see the actual layout. For now let's keep the logic in this class, though (at some point I can see us introducing new top-level patterns, but I'm not yet sure what those will look like).

Second, I suggest using protos for the UI data objects here. Those replace the data classes and can also be leveraged as a custom `Saver` class (that is probably easy to generalize and eventually share as a utility for other classes that need it). Compose actually seems to make this much easier to solve than the old saved instance state method which is really nice. This hopefully simplifies the saver code below a lot.

Third, a best practice in Compose (as I'm learning it) is to actually avoid state altogether and try to make Compose as 'dumb as possible' (which has a bunch of downstream benefits of simplifying developer decision making and thus improving code maintenance). https://developer.android.com/develop/ui/compose/state#stateful-vs-stateless talks about this more. The key concept is to use state hoisting to avoid needing to remember any states at all and instead pass through state. Other resources that touch on this:
- https://proandroiddev.com/passive-views-keep-your-ui-code-simple-and-stupid-75108c789af3
- https://developer.android.com/topic/architecture/ui-layer

(I've only skimmed these for now). It would be interesting to see how those principles apply here since this is one of the more complicated Compose cases I've seen so far, and it's sufficiently isolated that it can be reworked in place to try and find some ideal patterns that can be generalized for the actual Compose migration.

BenHenning (changes requested):
Thanks @adhiamboperes! I've reviewed everything except for `CreateProfileFragmentTest` (since it seems like it shouldn't be part of this PR) and `PinSetupFragmentTest` (since I'm assuming it's not yet completed).

I have left a variety of feedback, but the most significant is on the architectural structure of `PinSetupFragmentPresenter`. I used Gemini to review the code and help me figure out what about the approach was making me uneasy, and it came down to some data/logic conflation which then led to best practice patterns that I think we should be considering here. I'd like to take another full pass on that file in particular once you refactor it.

app/src/main/java/org/oppia/android/app/profile/PinSetupActivity.kt
adhiamboperes: Done

app/src/main/java/org/oppia/android/app/profile/PinSetupActivity.kt
adhiamboperes: Yes, the previous names were pretty generic.

app/src/main/res/layout/pin_setup_activity.xml
adhiamboperes: Done

app/src/main/res/values/strings.xml
adhiamboperes: Done.

app/src/sharedTest/java/org/oppia/android/app/onboarding/AdminIntroFragmentTest.kt
adhiamboperes: Reverted the extras, but only PROFILE_ID_INTENT_DECORATOR is currently used.

model/src/main/proto/arguments.proto
adhiamboperes: Done now.

config/src/java/org/oppia/android/config/platform/feature_flags.textproto
adhiamboperes: Reverted. I change this in the code because the onboarding screens come before I can access the developer options menu.

app/src/sharedTest/java/org/oppia/android/app/onboarding/AdminIntroFragmentTest.kt
```diff
@@ -209,21 +208,21 @@ class AdminIntroFragmentTest {
...
-        .setParentScreen(ProfileChooserActivityParams.ParentScreen.ADMIN_INTRO_SCREEN)
+      val expectedParams = AdminIntroActivityParams.newBuilder()
+        .setProfileType(ProfileType.SUPERVISOR)
+        .setProfileNickname("Admin")
         .build()
 
-      intended(hasComponent(ProfileChooserActivity::class.java.name))
-      intended(hasProtoExtra(PROFILE_CHOOSER_PARAMS_KEY, expectedParams))
+      intended(hasComponent(CreateAdminPinActivity::class.java.name))
+      intended(hasProtoExtra(ADMIN_INTRO_PARAMS_KEY, expectedParams))
```
adhiamboperes: This is incorrect, CreateAdminPinActivity only has the PROFILE_ID_INTENT_DECORATOR.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: No, made PinValidationResult private.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: Fair point. The success case is now a positive assertion at the top.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: This has been refactored to `createAdminPinUiStateSaver` which is now a regular instance property on the presenter class, not a singleton functional object. It follows standard camelCase property naming and no special justification is needed. The comment has be dropped entirely. I have documented this in https://github.com/oppia/oppia-android/issues/5422.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: I refactored the entire class:
- used protos for the UI data objects, but the actual error messages are derived on read. I however would like an opinion on where to store the proto. It currently lives in `screens.proto`.
- all `onValueChange` logic extracted to named presenter functions.
- the `CreateAdminPinScreen` Composable takes state + callbacks and owns nothing

The composables below `CreateAdminPinScreen` have no internal state. However, `CreateAdminPinScreen` itself still owns state via `rememberSaveable`, which means state hoisting hasn't been applied at the top level.

adhiamboperes (commented):
Self reviewed.

app/src/sharedTest/java/org/oppia/android/app/onboarding/AdminIntroFragmentTest.kt
adhiamboperes: Fixed.

adhiamboperes:
@BenHenning, could you PTAL at this PR? 

It is ready for review with the exception of `app/src/sharedTest/java/org/oppia/android/app/profile/CreateAdminPinActivityTest.kt ` which fails with `sigterm`. So far, I have been unable to debug and fix it due to being unable to run tests locally.

adhiamboperes:
Also the issue assignment workflow is still failing despite this branch being up to date with develop.

BenHenning:
> Also the issue assignment workflow is still failing despite this branch being up to date with develop.

@adhiamboperes it seems the regex check is case sensitive. Perhaps it will pass now?

app/src/sharedTest/java/org/oppia/android/app/profile/CreateAdminPinFragmentTest.kt
```diff
@@ -0,0 +1,627 @@
...
+
+  @After
+  fun tearDown() {
+    Intents.release()
+    TestPlatformParameterModule.reset()
+  }
+
+  @Test
+  fun testFragment_onLaunch_allTextViewsHaveCorrectContent() {
+    launch(CreateAdminPinActivity::class.java).use {
```
BenHenning, line 135: Nit: it's ideal to test fragments in the context of a different activity. If it's easy I suggest using `TestActivity` for this purpose.

That might potentially help with figuring out the timeout, too, since it looks like (per CI) both the fragment and activity classes are hanging.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    var uiState by rememberSaveable(stateSaver = createAdminPinUiStateSaver) {
+      mutableStateOf(initialState)
+    }
+
+    LaunchedEffect(Unit) {
+      focusRequester.requestFocus()
+      delay(100)
+      keyboardController?.show()
```
BenHenning: Is the delay and explicit show actually necessary? Gemini seems to suggest that the delay is likely what's causing the deadlock with our coroutine dispatcher testing barrier which would cause the hanging. It seems to assume that these lines may be unnecessary because focusing the text view _ought to_ be sufficient to open the keyboard automatically.

However, if this is called too soon it may need to be wrapped in a `onGloballyPositioned` so that focus happens after layout.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,490 @@
...
+        ),
+        keyboardActions = KeyboardActions(
+          onNext = if (imeAction == ImeAction.Next) {
+            { focusManager.moveFocus(FocusDirection.Down) }
+          } else null,
+          onDone = if (imeAction == ImeAction.Done) {
+            {
+              focusManager.clearFocus()
+              onDone?.invoke()
+            }
```
BenHenning, line 262: Unnecessary extra curly braces?

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,490 @@
...
+        isValid = false,
+        errorMessage = resourceHandler.getStringInLocaleWithWrapping(
+          R.string.create_admin_pin_activity_length_error
+        )
+      )
+
+      else -> PinValidationResult(
+        isValid = false,
+        errorMessage = resourceHandler.getStringInLocaleWithWrapping(
+          R.string.create_admin_pin_activity_mismatch_error
```
BenHenning: There are some repeat references to these strings. Is it possible to cleanly refactor to avoid repeating these such that each string is only referenced once?

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,490 @@
...
+  companion object {
+    private const val ADMIN_PIN_LENGTH = 5
+    private const val UI_STATE_SAVED_INSTANCE_STATE_KEY = "CreateAdminPinFragmentPresenter.ui_state"
+  }
+
+  /** Data class for PIN validation result. */
+  private data class PinValidationResult(
+    val isValid: Boolean,
+    val errorMessage: String = ""
+  )
```
BenHenning: Optional, but you could refine this a bit by leveraging a sealed class and also using a `StringRes` so that the string lookup only needs to happen in one place for the error:

```kotlin
private sealed class PinValidationResult {
  object Valid: PinValidationResult()
  data class Invalid(@StringRes val errorMessageId: Int): PinValidationResult()
}
```

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
```diff
@@ -0,0 +1,490 @@
...
+
+/** The presenter for [CreateAdminPinFragment]. */
+@FragmentScope
+class CreateAdminPinFragmentPresenter @Inject constructor(
+  private val activity: AppCompatActivity,
+  private val fragment: Fragment,
+  private val resourceHandler: AppLanguageResourceHandler,
+  private val profileManagementController: ProfileManagementController
+) {
+  private lateinit var binding: CreateAdminPinFragmentBinding
```
BenHenning: Does this actually need to use databinding? It seems unnecessary looking at the layout file (and probably better to avoid it since I'm not exactly sure how well Compose and databinding will interoperate...).

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -205,8 +204,26 @@ class CreateProfileFragmentTest {
     }
   }
 
+  fun testFragment_supervisorOnboardingFlow_stepCountThreeText_isDisplayed() {
```
BenHenning, line 208: Missing `@Test`.

app/src/sharedTest/java/org/oppia/android/app/profile/CreateAdminPinFragmentTest.kt
```diff
@@ -0,0 +1,627 @@
...
+      testCoroutineDispatchers.runCurrent()
+
+      composeRule
+        .onNodeWithText(context.getString(R.string.onboarding_navigation_continue))
+        .performClick()
+
+      testCoroutineDispatchers.runCurrent()
+
+      composeRule
+        .onNodeWithText(context.getString(R.string.create_admin_pin_activity_blank_error))
```
BenHenning, line 204: Hmm looking at these tests, I acknowledge they are quite clean however one issue I see is that there's a bit of a risk here. I think it's theoretically possible to show every related string to the new screen at one time and all of the tests will pass (or most will, there are a few assert not visible).

This is actually why I think it's important to validate specific views but I know Compose makes that tricky. I think we do need a solution for this but I'm not precisely sure what it is.

model/src/main/proto/screens.proto
```diff
@@ -182,10 +182,28 @@ enum ScreenName {
...
+  string pin = 1;
+
+  // The confirmation PIN currently entered by the user.
+  string confirm_pin = 2;
+
+  // Whether a submit-time validation error should be displayed.
+  bool show_error = 3;
+
+  // The submit-time error message to display when show_error is true.
+  string error_message = 4;
```
BenHenning, line 208: Will this cause any problems with the user changing their system language while the activity is open?

scripts/assets/test_file_exemptions.textproto
```diff
@@ -2028,6 +2028,14 @@ test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/profile/AdminSettingsViewModel.kt"
   test_file_not_required: true
 }
+test_file_exemption {
+exempted_file_path: "app/src/main/java/org/oppia/android/app/profile/CreateAdminPinActivityPresenter.kt"
+test_file_not_required: true
```
BenHenning, line 2033: Nit: missing spacing.

BenHenning (changes requested):
Thanks @adhiamboperes! Took a full pass, PTAL. I'll also respond to my earlier long comment (but can't as part of this review due to a GitHub bug).

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
BenHenning: I think introducing a new proto isn't unreasonable here, perhaps `ui_state.proto`?

Even with the `rememberSaveable` still in place, I find this approach already **much** cleaner and easier to follow and understand. This seems very solid for us to continue building patterns for broader Compose integration. Thanks!

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: These are needed. The function expectes a unit:

```php
Error: type mismatch: inferred type is Unit? but (KeyboardActionScope.() -> Unit)? was expected
          onDone = if (imeAction == ImeAction.Done) {
                   ^
```

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: Added these to the companion object and refactored the references.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: Done, the code readability has improved a bit.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: I have made the ComposeView the root element in the layout, and refactored its reference directly:

```kotlin
val composeView = inflater.inflate(
  R.layout.create_admin_pin_fragment, container, /* attachToRoot= */ false
) as ComposeView
```

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
adhiamboperes: Fixed.

app/src/sharedTest/java/org/oppia/android/app/profile/CreateAdminPinFragmentTest.kt
adhiamboperes: I realized that I didn't quite understand your intent here.

If you mean that we should assert something about the structure or context of those strings, then we could try to ensure we assert for a few conditions at a time, e.g. all other errors hidden + button disabled + only expected error shown. The downside is that the tests can be overly verbose.

But this doesn't look unique to Compose, since the problem could still occur with Android Views.

adhiamboperes: Or did you mean:

```kotlin
composeTestRule
    .onNode(matcher).assert(hasText("Button") or hasText("Button2"))
```

model/src/main/proto/screens.proto
adhiamboperes: No, it won't.  I tested this and the string language updates correctly for all errors on activity recreation.

[Screen_recording_20260421_160640.webm](https://github.com/user-attachments/assets/1dbc1e76-f8fa-4e96-9d77-995232c90e75)

scripts/assets/test_file_exemptions.textproto
adhiamboperes: This file does not use spacing between each item.

app/src/main/java/org/oppia/android/app/profile/CreateAdminPinFragmentPresenter.kt
adhiamboperes: No, They keyboard shows correctly without the delay, so I have removed it, and there was no need to use `onGloballyPositioned`

adhiamboperes: Removing the delay does not fix the hanging test issue. I had previously spent a lot of time investigating this issue, and the closest I got was pinpointing a platform parameter load issue:

```php
at org.oppia.android.testing.platformparameter.TestPlatformParameterModule$providePlatformParameterController$1.loadParametersAsync
at org.oppia.android.app.activity.InjectableAppCompatActivity$ensurePlatformParametersAreLoaded$1.invokeSuspend
```

adhiamboperes: Found a fix for the test hang, and it had to do with the test arrangement itself  :(

app/src/sharedTest/java/org/oppia/android/app/profile/CreateAdminPinFragmentTest.kt
adhiamboperes: In Compose, a stable way to identify specific elements and their relationships is to use Modifier.testTag, but we have previously agreed to avoid extensive use of test tags.

In this case, we can rely on semantics like combining text with tree relationships to identify each field-specific error.

So a test may look like:

```
val pinField = hasText(
  context.getString(R.string.create_admin_pin_activity_enter_pin_label)
)

composeRule.onNode(
  hasText(context.getString(R.string.create_admin_pin_activity_length_error)) and
    hasAnySibling(pinField)
).assertIsDisplayed()
```

Another option is to select a field by its editable semantics and label:

```
composeRule.onNode(
  hasText(context.getString(R.string.create_admin_pin_activity_enter_pin_label)) and
    hasSetTextAction()
)
```


Tests in this suite have been refactored such that they use:

1. normal accessibility semantics for controls: label, role, click action, enabled state.
2. parent/sibling relationships to associate an error with its field.
3.  no testTags because the layout exposes enough semantics to avoid tags

Issue #5732, opened 2025-03-03
[Feature Request]: Introduce 'flashback' flow to help learners get unstuck in explorations (GSoC'25 4.1)


BenHenning:
### Is your feature request related to a problem? Please describe.

When learners make a mistake on a concept they have previously demonstrated in an earlier part of a lesson, it often makes sense to redirect them back (or to a parallel flow) to try and reinforce earlier concepts that the learner may have not fully understood. However, in the current app implementation, learners subsequently need to re-answer all the cards between the earlier state and the state they had reached, which is frustrating.

#5572 is tracking a short-term solution to this problem, whereas this issue tracks a longer-term fix.

### Describe the solution you'd like

This project aims to provide a new feature called 'flashbacks' which helps to bring the benefits of earlier redirection (i.e. reviewing an earlier concept that directly ties to the learner's likely misconception) without the frustrating experience of having to redo the all the questions up to returning back to the question that originally caused the learner to become stuck.

Additionally, this project also includes improving the general look-and-feel of submitted answers for both multiple choice and item selection interactions as these both currently rely on HTML generation rather than having a cleaner, natively rendered experience. This change will also allow them to be displayed properly in the 'flashback' experience.

**Suggested milestones:**
- **Milestone 1**:
  - The new flashback dialog is implemented and hooked up to the existing soft redirection button. (The in-line flow does not need to work at this stage.)
  - When a learner is redirected, the Flashback Dialog should appear. Upon confirmation, the learner is taken back to the most recent instance of the card without adding a duplicate to the stack. This ensures a smoother learning experience without unnecessary reattempts of intermediate questions.

- **Milestone 2**:
  - Integrate the Flashback Dialog into the learner flow with an in-line view, where the "Learn Again" button is directly attached to the incorrect answer that triggered the flashback.
  - Implement the updated designs for multiple choice and item selection interactions submitted answers to ensure a cleaner, natively rendered experience.
  - Additionally, add relevant UI tests to verify the new functionality.

<details>
<summary>Technical hints/guidelines</summary>

- General changes explanation:
  - The new flows are introducing a novel navigation pathway (being able to navigate 'back' to an earlier state, but interact with it in a completely unique way), as well as a parallel "I'm stuck" flow that's similar to hints and solutions, but more explicit (due to being largely controlled by a "I'm stuck" button). This pathway diverges for cases when the user is stuck on the first card of an exploration.
[...]

PR #5864, opened 2025-06-14
 Fix part of #5732: Implement domain logic to show flashback button

subhajitxyz:
## Explanation
Fixes part of #5732

- Add a new field to the AnswerAndResponse message in exploration.proto.
- Add a logic to detect flashbacks in ExplorationProgressController and add functions in StateDeck.
- Add relevant tests in ExplorationProgressControllerTest.
- Logic gated behind feature flag.

This PR will introduce domain logic for showing new "See example" button (Flashback button).

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

PR #5910, opened 2025-08-01
Fix part of #5732: Enhance conversation flow by adding “See example” button and feedback in the previous response section (2/5)

subhajitxyz:
## Explanation
Fixes part of #5732 

**PR 2.1**

Based on Sean’s feedback, I’ve updated the UI to match the last mock shared in the [comment](https://github.com/oppia/design-team/issues/179#issue-2874247813). In this version, we hide the submitted answer, feedback, and the “See Example” button. Only the feedback text — “Now that you’ve seen an example, let’s try again.” — is shown.

This feedback string was suggested by the UXW team.

- Introduces a boolean `flashback_viewed` in AnswerAndResponse proto. This flag indicates whether "See example" button was used by the learner to view a flashback.
- When the "**Return to Question**" button is used, flashback_viewed is set to true. This helps us determine whether to include the "See example" button and its feedback in the Previous Response section.

## Essential Checklist

### **(Outdated) Phone light mode**
https://github.com/user-attachments/assets/886b7d61-f06f-4945-aa1a-85a972ac0b3b

### Latest output
**Phone light and dark mode**

https://github.com/user-attachments/assets/7aa546e3-bf82-439e-ab3a-8a2fa5f695f8

**Tablet light mode**

[pr 2.1 - string_change.webm](https://github.com/user-attachments/assets/13d525c9-9d34-416b-952c-3e2de4c17b8e)


subhajitxyz:
PTAL, @adhiamboperes

seanlip (changes requested):
Hi @subhajitxyz, leaving a product review as requested:

- Could you please confirm the text "This doesn't seem right. Let's go back and look at the previous question and answer to understand better." with the UXW team? Ditto for the text in the popup. (I think these might need revision -- one thing to clarify when talking to UXW is that the question being shown may not be the immediately-previous question, so the use of "the" in "the previous question" might be misleading as it implies that there's only one.)
- In the flashback, could you separate out the explanation section more? Maybe bold the "Explanation" header and leave a bit of space below it.
- Why is there a "1/2" learner's answer on the bottom right of the flashback, just above "Return to question"? That should not be there.
- On returning to the original question, the text is wrong (the previous thing should be collapsed and the "Need help?" text should instead be "Now that you have ..." . See last mock here: https://github.com/oppia/design-team/issues/179#issue-2874247813 . This text should also be confirmed with UXW (but I think it looks generally OK). The previous answer/flashback-button should also be hidden.
- Do not offer the same flashback twice for a question. In general, there should only ever be one flashback experience for a particular question.

Please fix the above and let me know if you have any questions. Thanks!

Also, one question -- how are you selecting the question to flashback to?

subhajitxyz:
Hi @seanlip , 
One question -- should we count the "See example" button as part of the previous response count? (We currently show a count in the Previous Responses header.)

Here is answers for your above comment.

- I will consult the UXW team regarding the feedback texts.
- Got it.
- Regarding the "1/2" in the learner's answer — in the [mock](https://github.com/oppia/design-team/issues/179#issue-2874247813), it shows the multiple-choice submitted answer above the "Return to question" button in the flashback. Could you please clarify whether we should show the submitted answer only for multiple-choice, or for all types of submitted answers?
- It was my mistake to use the wrong string here — I’ll correct it. For the rest (hiding the flashback button and collapsing the previous section), I’ll discuss it with the mentors.
- Got it.

Answer to your last question:
When we get previous destination state for a wrong answer. we show this state as a flashback For testing purpose, I have modify test_exp_id.json [[LINK](https://github.com/oppia/oppia-android/pull/5864/files#diff-46693020da8ec7022032062529948bc1c70af597865fca3c95bdbf26f6caeede)] -- the one we saw in the screen recording.


Adding @BenHenning and @adhiamboperes for your feedback.

seanlip:
@subhajitxyz 

- Each answer/response pair is counted as 1 in the previous response count. This includes the answer/response pair that includes the "See example" button.
- Re the "1/2" -- oh, I see. No, it doesn't make sense to show the correct answer twice. An alternative would be to update the "correct solution is" section in the "Solution" box to show the three radio options instead, with the correct one highlighted, followed by the usual explanation if there is any.

I hope this helps, please ask if you have further questions.

subhajitxyz:
PTAL, @adhiamboperes

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -491,9 +494,100 @@ class StatePlayerRecyclerViewAssembler private constructor(
             previousAnswerViewModels += viewModel
           }
         }
+
+        if (playerFeatureSet.flashbackNavigationSupport && answerAndResponse.flashbackViewed) {
```
adhiamboperes: Is this condition really different from the one  below? for this condition to be met, I also expect `!answerAndResponse.stateNameToRevisit.isNullOrBlank()` to also be met.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -872,14 +972,9 @@ class StatePlayerRecyclerViewAssembler private constructor(
...
 
   private fun addFlashbackButton(
-    conversationPendingItemList: MutableList<StateItemViewModel>,
-    extraInteractionPendingItemList: MutableList<StateItemViewModel>,
     flashbackStateName: String
-  ) {
-    val targetList =
-      if (isSplitView.get()!!) extraInteractionPendingItemList else conversationPendingItemList
-
-    targetList += FlashbackButtonViewModel(
```
adhiamboperes, line 898: Could you please explain why this refactor was necessary

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -5735,15 +5744,61 @@ class StateFragmentTest {
...
       )
+    }
+  }
 
-      // Verify flashback button is visible.
-      scrollToViewType(FLASHBACK_BUTTON)
-      onView(withId(R.id.flashback_button)).check(
-        matches(withText(R.string.state_flashback_button))
+  @Test
+  fun testFlashback_CompleteFlashbackView_submitWrongAnsAgain_verifyFlashbackButtonisNotOffered() {
```
adhiamboperes:
```suggestion
  fun testFlashback_viewFlashback_submitWrongAnsAgain_verifyFlashbackButtonIsNotOffered() {
```

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -5735,15 +5744,61 @@ class StateFragmentTest {
...
+      // Verify submit button is visible.
+      scrollToViewType(SUBMIT_ANSWER_BUTTON)
+      onView(withId(R.id.submit_answer_button)).check(
+        matches(withText(R.string.state_submit_button))
       )
     }
   }
 
+  @Test
+  fun testFlashback_completeFlashbackView_submitCorrectAnswer_canSuccessfullyMoveToNextState() {
```
adhiamboperes:
```suggestion
  fun testFlashback_viewFlashback_submitCorrectAnswer_canSuccessfullyMoveToNextState() {
```

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -915,6 +922,9 @@ class ExplorationProgressController @Inject constructor(
         "Cannot navigate to a next state if an answer submission is pending."
       }
 
+      if (!explorationProgress.stateDeck.isFlashbackViewed()) {
+        explorationProgress.stateDeck.setFlashbackIsViewed()
+      }
```
adhiamboperes, line 913: Insert newline for formatting consistency.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
```diff
@@ -730,7 +730,13 @@ class ExplorationProgressController @Inject constructor(
         // Follow the answer's outcome to another part of the graph if it's different.
         val ephemeralState = computeBaseCurrentEphemeralState()
         when {
-          answerOutcome.destinationCase == AnswerOutcome.DestinationCase.STATE_NAME -> {
+          answerOutcome.destinationCase == AnswerOutcome.DestinationCase.STATE_NAME &&
+            (
+              !enableFlashbackSupport.value ||
+                answerOutcome.labelledAsCorrectAnswer ||
+                !explorationProgress.stateDeck.isFlashbackViewed()
```
adhiamboperes: Is this possibly a redudant condition?
```suggestion
                !explorationProgress.stateDeck.isFlashbackViewed()
```

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
```diff
@@ -281,4 +281,28 @@ class StateDeck constructor(
     return previousStates.find { it.state.name == stateName }
       ?: EphemeralState.getDefaultInstance()
   }
+
+  /**
+   *  Sets the `flashback_viewed` field to true for the last [AnswerAndResponse] in the
```
adhiamboperes:
```suggestion
   *  Sets the `flashback_viewed` field [of someproto] to true for the last [AnswerAndResponse] in the
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -3362,6 +3361,8 @@ class ExplorationProgressControllerTest {
     // Verify answer and response contains the flashback state name.
     assertThat(answerAndResponse.stateNameToRevisit)
       .isEqualTo("Fractions")
+    // Verify answer and response contains the flashbackViewed is false.
```
adhiamboperes:
```suggestion
    //  Verify the answer and response has the flashbackViewed field set to false.
```

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -3462,6 +3463,14 @@ class ExplorationProgressControllerTest {
 
     // Verify that there is exactly one wrong answer.
     assertThat(expectedEphemeralState.pendingState.wrongAnswerCount).isEqualTo(1)
+
+    // Access the first answer and response in the list.
+    val answerAndResponse = expectedEphemeralState.pendingState.wrongAnswerList[0]
+
+    // Verify answer and response contains the flashback state name.
+    assertThat(answerAndResponse.stateNameToRevisit).isEqualTo("Fractions")
+    // Verify answer and response contains the flashbackViewed is true.
```
adhiamboperes: Refactor per my suggestion above.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
```diff
@@ -3462,6 +3463,14 @@ class ExplorationProgressControllerTest {
...
+    assertThat(answerAndResponse.flashbackViewed).isEqualTo(true)
```
adhiamboperes, line 3477: Looking at this test, it was not set up correctly the firsttime round, and I missed that in the original review.  

In ln 3457, where     // Click on 'Return to Question' button, `moveBackToLatest()` is not being called. Instead, the test goes ahead to assert the result of `val expectedEphemeralState = moveBackToLatest()`, which will never fail.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -5714,15 +5716,22 @@ class StateFragmentTest {
...
 
       // Verify feedback is visible.
-      val expectedFeedback = "This doesn't seem right. Let's go back and look at the previous" +
-        " question and answer to understand better."
-      scrollToViewType(FEEDBACK)
+      val expectedFeedback = "Now that you’ve seen an example, let’s try again."
+
+      onView(withId(R.id.state_recycler_view)).perform(
+        scrollToPosition<RecyclerView.ViewHolder>(4)
+      )
```
adhiamboperes: Please maintain a consistent formatting style when using newlines.

adhiamboperes (changes requested):
Thanks @subhajitxyz! I have left inline comments, but I also noticed something in the screen recording that is worth looking at. In the previous responses header, after entering 3 answers, the header shows a count of 2. Could you please verify that? It would also be great to add a test for that if non exists.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
subhajitxyz: Yes, we can add the condition `!answerAndResponse.stateNameToRevisit.isNullOrBlank()`. However, I didn’t include it because `answerAndResponse.flashbackViewed` already guarantees that stateNameToRevisit is neither null nor blank. We only set flashbackViewed = true for an answerAndResponse that already contains a valid stateNameToRevisit.

I am adding this condition.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
subhajitxyz: In the previous implementation, this function adds ViewModel only in conversationPendingItemList and extraInteractionPendingItemList.

In this PR, we also need to store this ViewModel in previousAnswerViewModels which is used to show previous response section, You can see to code above, I’ve add a comment (ping section).

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -491,9 +494,100 @@ class StatePlayerRecyclerViewAssembler private constructor(
...
+        }
+
+        addFlashbackButton(
+          answerAndResponse.stateNameToRevisit
+        ).let { viewModel ->
+          if (showPreviousAnswers) {
+            pendingItemList += viewModel
+          }
+          previousAnswerViewModels += viewModel
+        }
```
subhajitxyz, line 595: Pinging for below explaination.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
subhajitxyz: This condition ensures that the flashback button will not be shown more than once after it has already been viewed. In that case, we fall into the second condition, which handles the pending state.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
subhajitxyz: In the above line, the `navigateToFlashbackState()` function is used to navigate to the Flashback screen.
After that, we call `val expectedEphemeralState = moveBackToLatest(),` which returns the latest ephemeralState  (where the user previously made a mistake.)
We then verify the attributes of this latest ephemeralState.

subhajitxyz:
> Thanks @subhajitxyz! I have left inline comments, but I also noticed something in the screen recording that is worth looking at. In the previous responses header, after entering 3 answers, the header shows a count of 2. Could you please verify that? It would also be great to add a test for that if non exists.

Previous responses header does not count the last submitted answer.(This is an existing behaviour).
I have added a test which verifies header is not displayed ,when submit one wrong answer and get flashback button.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
subhajitxyz: Updated.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
subhajitxyz: Thanks, Updated.

domain/src/main/java/org/oppia/android/domain/exploration/ExplorationProgressController.kt
subhajitxyz: Done.

domain/src/main/java/org/oppia/android/domain/state/StateDeck.kt
subhajitxyz: Done.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
subhajitxyz: Thanks, Done.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
subhajitxyz: Done.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
subhajitxyz: Done.

subhajitxyz:
PTAL, @adhiamboperes . I have updated the pr.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -447,10 +450,23 @@ class StatePlayerRecyclerViewAssembler private constructor(
...
+        answerAndResponse.flashbackViewed
+    }
+    val previousAnswerCount = if (flashbackRecentlyViewed)
+      answersAndResponses.size else answersAndResponses.size - 1
+
+    if (answersAndResponses.size > 1 || flashbackRecentlyViewed) {
       if (playerFeatureSet.wrongAnswerCollapsing) {
         PreviousResponsesHeaderViewModel(
-          answersAndResponses.size - 1,
+          previousAnswerCount,
```
subhajitxyz, line 469: Here, we handle a special case where, even if the user submits one wrong answer and completes the flashback, we show the previous response header but hide the user answer, feedback, and flashback button, displaying only the button feedback.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
adhiamboperes: Per the discussion in the meeting today, please rename `expectedEphemeralState` to `ephemeralState`.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
adhiamboperes: Sounds good.

adhiamboperes (changes requested):
Thanks @subhajitxyz, I have resolved the comments that I have verified, but please also respond to my top-level comment:

> Thanks @subhajitxyz! I have left inline comments, but I also noticed something in the screen recording that is worth looking at. In the previous responses header, after entering 3 answers, the header shows a count of 2. Could you please verify that? It would also be great to add a test for that if none exists. 

Ben left a final clarification in the group chat thread, please verify that the solution is aligned.

subhajitxyz:
For your top level comment.
This was Ben's decision

> 
> Chatted about this in the meeting today: I think the see example button should probably be above the interaction for consistency.
> 
> Separately, should the interaction and submit button be disabled to force the flow?
> 
> Finally, the answer, feedback, see example button items should be one unit for collapsing/expanding. This means a second time hitting the flashback should hide the previous see example button(s) unless previous responses are expanded. It’s fine that these buttons be duplicated in the previous responses though.
> 
> @Sean Lip do you agree with all of this or have other thoughts? Note the first two paragraphs are a change to the existing mocks.


After that Sean shared his final decision above on this PR comment.

> On returning to the original question, the text is wrong (the previous thing should be collapsed and the "Need help?" text should instead be "Now that you have ..." . See last mock here: https://github.com/oppia/design-team/issues/179#issue-2874247813 . This text should also be confirmed with UXW (but I think it looks generally OK). The previous answer/flashback-button should also be hidden.

> Do not offer the same flashback twice for a question. In general, there should only ever be one flashback experience for a particular question.


So, I chose to keep the answer, feedback, see example button items should be one unit for collapsing/expanding (As per Ben decision)
The previous thing should be collapsed  See last [mock](https://github.com/oppia/design-team/issues/179#issue-2874247813 )  (As pen Sean decision)
chose not to offer the same flashback twice for a question. (As per Sean decision)

Since, Ben's above two decision is actually a change of M1 work and that decision was not final. So, I have not implemented it in this PR.

domain/src/test/java/org/oppia/android/domain/exploration/ExplorationProgressControllerTest.kt
subhajitxyz: Done

adhiamboperes (approved):
LGTM, thanks @subhajitxyz!

adhiamboperes:
@BenHenning, could you please do a final review pass on this when you have a moment? It would be great to verify that the UI works as expected and the logic is sound.

BenHenning:
I haven't tested this on a device, but looking at the video I noticed a few things:
- The 'See Example' button needs to be repeated otherwise users need to expand previous answers in order to see it. This is confusing, especially with the feedback message saying to review things.
- As was mentioned in the chat thread, the solution box needs the following changes:
  - 'Solution' should be 'Explanation'
  - The box should only include the explanation text, not the answer.
  - If there is no explanation, the entire solution (explanation) box should be hidden.

@seanlip I think you may disagree with point (1), but per the video I find the current experience very confusing. If the user enters the same type of wrong answer, they get text to go back and see an example but no button. That's really confusing to me.

Perhaps we could hide all previous 'See Example' buttons and only show the _latest_ one that has been triggered? It would still be the same flashback experience as before, but it would remove the duplication concern and avoid logically getting stuck.

subhajitxyz:
Thanks @BenHenning , for your feedback.
For clarification on your 2nd point about the solution box. Adhiambo and I decided to first finish the M2 work, and then address these changes in upcoming PRs, since they relate to change on the M1 work.

seanlip:
@BenHenning Just a note, one stipulation I had from the product side is that we never show a flashback more than once in a card. In other words, the 'See Example' button would only ever get shown once. There is no point showing the learner the same type of feedback twice.

Does that address your concern?

BenHenning:
> @BenHenning Just a note, one stipulation I had from the product side is that we never show a flashback more than once in a card. In other words, the 'See Example' button would only ever get shown once. There is no point showing the learner the same type of feedback twice.
> 
> Does that address your concern?

No--I don't think this addresses it. The video shows the following scenario:
- User submits an incorrect answer and is prompted with a prompt saying that they need to revisit the concept. They are shown the 'See Example' button.
- User clicks on 'See Example', visits the flashback, and then returns.
- User submits another answer that triggers the same flashback and is told that they need to revisit the concept, again.

What should we do in this case? Not showing 'See Example' means showing them feedback that isn't actually actionable (go back and visit again) which is very confusing to me.

My proposal is that we still show 'See Example', but we move it to be at the latest instance of feedback rather than at each (so that there's only ever one button).

Edit: Also, I suppose we could change the feedback in these cases but what should that feedback be?

seanlip:
Hi @subhajitxyz, @adhiamboperes -- Ben and I had a long discussion about this, and here is what we think makes the most sense from a product perspective:

- The trigger for showing a flashback should be when the learner hits the default feedback **for the first time** in that card. In that case, **if** the current card is tagged with a [linked_skill_id](https://github.com/oppia/oppia/blob/develop/core/domain/state_domain.py#L3636), look at the learner's card history and find the most recent card that's tagged with the same linked skill ID. Show that card as the flashback.
- If the above condition can't be met (the current card doesn't have a linked_skill_id, or there is no linked_skill_id in the history, or this is the learner's second time encountering the default feedback, etc.), just show the default answer as normal and don't show the flashback.

Keep routing etc. the same as it is now. We will ask the lesson team to add linked skill IDs to all the relevant lesson cards, and to remove loop-backs that force the learner to go to a much earlier card. Note that, with the above mechanic, a flashback will never be offered more than once within a card.

Do you think this will work? I apologize for this being figured out a bit late -- there has always been some uncertainty about what the right trigger should be, but in both our opinions this makes the most sense from an overall product perspective.

adhiamboperes (approved):
LGTM, thanks @subhajitxyz.

adhiamboperes:
@seanlip, could you please take a look and approve the changes to unblock merge?

seanlip:
I am looking at my original comment at https://github.com/oppia/oppia-android/pull/5910#pullrequestreview-3080868148 and comparing it with the "latest output" video. I don't think that the points I mentioned were addressed, e.g. I still see the answer shown just above the "Return to question" button. (I didn't check the others.)

@subhajitxyz If you would like a follow-up review then please ensure that you address all reviewer comments beforehand and leave a comment to that effect. Thanks.

subhajitxyz:
Thanks @seanlip , This PR involves in adding only "See Example" button and feedback in Previous response section.
In next PR #5917 , I have improved the design of submitted answer.
In next PR #5918  I have addressed those changes request for flashback screen and remove confirmation dialog.

subhajitxyz:
> I am looking at my original comment at [#5910 (review)](https://github.com/oppia/oppia-android/pull/5910#pullrequestreview-3080868148) and comparing it with the "latest output" video. I don't think that the points I mentioned were addressed, e.g. I still see the answer shown just above the "Return to question" button. (I didn't check the others.)
> 
> @subhajitxyz If you would like a follow-up review then please ensure that you address all reviewer comments beforehand and leave a comment to that effect. Thanks.

@Sean, I have filed this issue #5932 for the requested changes. Here is the corresponding PR #5918 .

seanlip (approved):
Thanks @subhajitxyz for keeping track of the unaddressed items. LGTM for this PR.

PR #5912, opened 2025-08-05
Fix part of #5734: Resolve issue NewApi of severity Error

manas-yu:
## Explanation

Fixes part of #5734

- Resolve issues for Issue Category 'NewApi'
   - Call requires API level 24 (current min is 21): java.lang.Iterable#forEach: Used iterator() as a workaround. ref: [SO](https://stackoverflow.com/questions/68046599/lint-is-giving-error-if-i-use-java-lang-iterableforeach-in-android-code#:~:text=This%20issue%20link,Java8%20forEach%20function)
       - The lint tool is probably confusing Kotlin's forEach (which is an extension function) with Java 8's Iterable.forEach() method. This was also an issue in the Gradle lint tool reports.
   - android:autofillHints requires API level 26 (current min is 21): Created a `@BindingAdapter` for EditText to disable autofillHints with if-else check for API level.

## Essential Checklist


manas-yu:
Hi @bhavukJain PTAL,
We can revert the issue/changes relevant to the `forEach`, I think, as they were false reports in the gradle versions as well.

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
```diff
@@ -74,7 +74,7 @@ abstract class AbstractOppiaApplication(
           }
           WorkManager.initialize(applicationContext, workManagerConfiguration)
           val workManager = WorkManager.getInstance(applicationContext)
-          component.getAnalyticsStartupListenerStartupListeners().forEach {
+          component.getAnalyticsStartupListenerStartupListeners().iterator().forEach {
```
bhavukJain: forEach is used in a lot of other places. Are you sure you have covered all of them? For eg: https://github.com/oppia/oppia-android/blob/develop/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt#L1009

bhavukJain: Can we instead use a more traditional approach such as for (item in list) that works in API level 21? Or we can also look into @TargetApi annotations.

manas-yu: Suppressed NewApi issue related to forEach as discussed in meet. ref: https://stackoverflow.com/questions/68046599/lint-is-giving-error-if-i-use-java-lang-iterableforeach-in-android-code#:~:text=This%20issue%20link,Java8%20forEach%20function.

manas-yu:
Hi @adhiamboperes PTAL.

Cc: @bhavukJain

adhiamboperes:
@BenHenning, could you please take a look at this? I am unsure whether the correct approach for this issue is to supress the warnings. Wouldn't it be better to keep the exemptions?

BenHenning:
> @BenHenning, could you please take a look at this? I am unsure whether the correct approach for this issue is to supress the warnings. Wouldn't it be better to keep the exemptions?

This is tricky--thanks for calling it out @adhiamboperes.

@manas-yu I'm not keen on the per-line obsolete lines here since it's easy to not fully understand why we are exempting. In general, we should minimize exemptions.

Another approach that may be cleaner: introduce custom extension functions for `forEach` (e.g. `safeForEach`) that can call the correct Kotlin API with an exemption. Then we can do two things:
1. Introduce a static check to prohibit using other `forEach` calls everywhere (and thus force everything to use the central ones--now people will fail the regex check before hitting a lint issue that needs to be interpreted and exempted).
2. Add a TODO to remove those functions once the lint tool no longer falsely triggers on them.

Isolating exemptions like this is a good pattern, in general, so that people don't need to know the full context of exempting when they run into the same case in the future.

manas-yu:
Hi @BenHenning @adhiamboperes, PTAL at the updated approach.  

Instead of enforcing a strict regex check that would block all usages of `forEach`, this approach would use `falsePositiveIssues` so we only target the cases incorrectly flagged by Lint. This would also give us future flexibility to handle similar scenarios within an issue category, where we can make more controlled reports focusing on issues within an issue category.
> introduce custom extension functions for forEach (e.g. safeForEach) that can call the correct Kotlin API with an exemption. 

- `safeForEach` in IterableExtensions:  Uses a traditional `for` loop internally to safely replace the falsely reported `forEach` usage. [[ref](https://github.com/oppia/oppia-android/pull/5912#discussion_r2261962649)]

>Introduce a static check to prohibit using other forEach calls everywhere (and thus force everything to use the central ones--now people will fail the regex check before hitting a lint issue that needs to be interpreted and exempted).

- The Android Lint script enforces usage by flagging only the falsely reported cases.  
  Unlike a global regex (e.g., `"\.forEach\s*\("`), this avoids blocking every `forEach` call and keeps the workflow cleaner.
  
  Let me know if this is feasible or if using a global regex pattern check would be better, eliminating all the usage of `forEach`. Thanks!

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
```diff
@@ -513,6 +516,19 @@ class CreateProfileFragmentTest {
...
+    launchNewLearnerProfileActivity().use {
+      onView(withId(R.id.create_profile_nickname_edittext))
+        .check { view, _ ->
+          val editText = view as EditText
+          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            assertThat(editText.importantForAutofill).isEqualTo(View.IMPORTANT_FOR_AUTOFILL_NO)
+          }
+        }
+    }
+  }
```
adhiamboperes: This test will always pass because according to `robolectric.properties`, our tests always run on `sdk 30`. There is a posible [Config](https://robolectric.org/configuring/) that you can try though, and check, as well as share proof, that setting a lower sdk causes this test to fail.

adhiamboperes (commented):
Thanks @manas-yu@ This cleanup looks pretty good!

I have only one comment

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
manas-yu: Thanks for pointing it out. I’ve updated the test to cover both SDK 25 (where autofill isn’t available) and SDK 30 (where we can validate autofill). If we include the autofill assertion inside the SDK 25 branch, it results in the following stack trace:
```kotlin
 testFragment_nicknameEditText_autofillBehavior_multiSdk[25](org.oppia.android.app.onboarding.CreateProfileFragmentTest)
java.lang.NoSuchMethodError: 'int android.widget.EditText.getImportantForAutofill()'
        at org.oppia.android.app.onboarding.CreateProfileFragmentTest$testFragment_nicknameEditText_autofillBehavior_multiSdk$1$1.check(CreateProfileFragmentTest.kt:534)
        at androidx.test.espresso.ViewInteraction$SingleExecutionViewAssertion.check(ViewInteraction.java:2)
        at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:14)
        at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:1)
        at android.os.Handler.$$robo$$android_os_Handler$handleCallback(Handler.java:751)
        at android.os.Handler.handleCallback(Handler.java)
        at android.os.Handler.$$robo$$android_os_Handler$dispatchMessage(Handler.java:95)
        at android.os.Handler.dispatchMessage(Handler.java)
        at org.robolectric.shadows.ShadowPausedLooper$IdlingRunnable.run(ShadowPausedLooper.java:332)
        at org.robolectric.shadows.ShadowPausedLooper.executeOnLooper(ShadowPausedLooper.java:366)
        at org.robolectric.shadows.ShadowPausedLooper.idle(ShadowPausedLooper.java:92)
        at org.robolectric.android.internal.LocalControlledLooper.drainMainThreadUntilIdle(LocalControlledLooper.java:17)
        at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:1)
        at androidx.test.espresso.ViewInteraction.check(ViewInteraction.java:12)
        at org.oppia.android.app.onboarding.CreateProfileFragmentTest.testFragment_nicknameEditText_autofillBehavior_multiSdk(CreateProfileFragmentTest.kt:524)

``` 

Let me know if this resolves the issue. Thanks!

manas-yu:
Hi @adhiamboperes PTAL. Thanks!

adhiamboperes:
@manas-yu, did you forget to push your latest commits? I can't seem to see any new changes since my last review

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
adhiamboperes: The assertion should them be against the exception.

manas-yu:
Hi @adhiamboperes, PTAL, I forgot to push the changes and have pushed them now. Thanks for pointing it out.

app/src/sharedTest/java/org/oppia/android/app/onboarding/CreateProfileFragmentTest.kt
manas-yu: I have included assertion for the exception in the test.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
```diff
@@ -177,6 +192,16 @@ class LintAnalysisReporter {
 
     private const val PROTO_BINARY_FILE_PATH = "scripts/assets/android_lint_exemptions.pb"
     private const val EXEMPTIONS_FILE_PATH = "scripts/assets/android_lint_exemptions.textproto"
+
+    private val falsePositiveIssues: Set<FalsePositiveIssue> = setOf(
+      FalsePositiveIssue(
```
BenHenning, line 198: This seems like a solid approach, though I recommend still filing an issue and adding a TODO here that points to it to track removing this and the corresponding `safeForEach` once we upgrade to a version of the lint tool that fixes the underlying problem.

utility/src/main/java/org/oppia/android/util/extensions/IterableExtensions.kt
```diff
@@ -0,0 +1,7 @@
+package org.oppia.android.util.extensions
+
+// TODO(#5734): Remove this function once lint no longer falsely triggers on Iterable#forEach.
```
BenHenning: Per my comment above, this should be an issue specific to the workaround not the general issue (which hopefully can be closed when all the key lint categories are supported).

BenHenning:
Thanks @manas-yu, just had a few follow-up thoughts but otherwise the workaround seems solid.

scripts/src/java/org/oppia/android/scripts/lint/LintAnalysisReporter.kt
manas-yu: Created issue tracker and added TODO

utility/src/main/java/org/oppia/android/util/extensions/IterableExtensions.kt
manas-yu: Created a new issue and updated TODO

manas-yu:
Hi @adhiamboperes PTAL, addressed all comments

adhiamboperes (approved):
Thanks @manas-yu, this looks good to me.

PR #5921, opened 2025-08-17
Fix part of #5734 and Fix #5897: Suppress Issue Categories and Include timer

### LINKED ISSUE #5897: [BUG]: Android lint fails due to synthetic accessor warning for private method used within same class.
### Describe the bug

When declaring a `private` method `onBackNavigation()` in the [`PlatformParametersFragmentPresenter`](https://github.com/oppia/oppia-android/blob/b50f6e19c338c41c50b62317acaf3eebc169eb2b/app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt#L110) class (used only within the same class), the Android lint check fails with a synthetic accessor warning. Although this should ideally only be a warning (since it's just a performance suggestion), it's currently causing a build failure due to the lint step throwing an IllegalStateException.

This seems to be an unintended side effect of treating a performance warning as a hard failure, which should be relaxed or fixed in the lint pipeline.

### Steps To Reproduce

This behaviour is seen in [`PlatformParametersFragmentPresenter`](https://github.com/oppia/oppia-android/blob/b50f6e19c338c41c50b62317acaf3eebc169eb2b/app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt#L110) class .

### Expected Behavior

The Android lint test should not fail on this.

### Screenshots/Videos

Image

_No response_

_No response_

_No response_

### Additional Context

_No response_

manas-yu:
## Explanation

Fixes part of #5734
Fix #5897 

- Suppressed Issue categories  that aren't relevant in lint analysis for the codebase
   - MissingTranslation: This is kept in check via TranslateWiki and can be suppressed in Android lint reports
   - SelectableText: It is preferred to keep the TextViews non-selectable for now as being able to select the text would actively conflict with a user action
   - DuplicateStrings: some strings can get translated differently due to the context within which it is being used, and keeping duplicated strings is preferred  and is a general false positive even in other codebases' lint analysis
   - SyntheticAccessors: The fix is really tedious since it involves refactoring all the lambdas and suppressing them is preferred.
   - GradleOverrides: This is a Gradle-related issue category and isn't relevant since we have migrated to Bazel.
- Added a timer to display the script execution progress in the terminal. 
## Essential Checklist


scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -94,11 +194,20 @@ class AndroidLintAnalyzer(
...
   private val bazelClient = BazelClient(repoRoot, commandExecutor)
   companion object {
     private const val LINT_REPORT_FILE = "lint-report.xml"
+
+    private val suppressLintIssues = setOf(
+      "MissingTranslation",
+      "GradleOverrides",
+      "SyntheticAccessor",
+      "DuplicateStrings",
+      "SelectableText"
```
bhavukJain, line 223: Wouldn't suppressing this impact accessibility compliance as this warning indicate missing android:textIsSelectable="true" attributes?

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -94,11 +194,20 @@ class AndroidLintAnalyzer(
...
-  private val groupByIssueSeverity: Boolean = false
+  private val groupByIssueSeverity: Boolean = false,
+  private val timer: ElapsedTimeDisplayer? = null
 ) {
   private val bazelClient = BazelClient(repoRoot, commandExecutor)
   companion object {
     private const val LINT_REPORT_FILE = "lint-report.xml"
+
+    private val suppressLintIssues = setOf(
+      "MissingTranslation",
```
bhavukJain, line 215: Can you add a comment why these issue types are being suppressed?

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -37,47 +127,57 @@ private const val DEFAULT_PROTO_BINARY_PATH = "scripts/assets/android_lint_exemp
  *   bazel run //scripts:android_lint_check -- $(pwd) --processTimeout=15
  */
 fun main(vararg args: String) {
-  require(args.isNotEmpty()) {
-    "<path_to_repository_root argument> is required: \$(pwd)"
-  }
+  ScriptBackgroundCoroutineDispatcher().use { scriptBgDispatcher ->
```
bhavukJain, line 130: Should we move timer initialisation after successful argument validation?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -104,6 +112,51 @@ class AndroidLintRunnerTest {
     assertThat(exception).hasMessageThat().contains("Repository root path does not exist")
   }
 
+  @Test
+  fun testStart_firstTime_startsTimer() = runBlocking {
+    elapsedTimeDisplayer.start()
+
+    delay(1100)
```
bhavukJain: Often, tests using delay() and timing can be flaky in CI environments. A better approach I guess would be to use fake time providers or test schedulers for more reliable timing tests.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: I have suppressed it following the mail Thread related to the SelectableText, where the conclusion was to suppress the issue category as it is currently preferred that text remains non-selectable.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: I have avoided that since the purpose of changes like these is primarily in PR descriptions, but I can add comments if required. I would like @adhiamboperes's confirmation as well.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: Updated.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Updated tests to avoid using delay()

manas-yu:
Hi @adhiamboperes PTAL.

Cc: @bhavukJain

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -18,66 +23,195 @@ private const val DEFAULT_PROCESS_TIMEOUT_MINUTES = 10L
...
-  val processTimeout = args.find { it.startsWith("--processTimeout=") }
-    ?.substringAfter("=")
-    ?.toLongOrNull() ?: DEFAULT_PROCESS_TIMEOUT_MINUTES
+    val exemptionProtoPath = args.find { it.startsWith("--proto=") }?.let { option ->
+      val path = option.substringAfter("=")
+      require(path.endsWith(".pb")) {
+        "Invalid exemption file: $path. The file must have a .pb extension."
+      }
+      path
+    } ?: DEFAULT_PROTO_BINARY_PATH
```
adhiamboperes, line 145: It is great to check that the file is a valid pb file but any pb file could be passed here, so either the check is not strict enough or I could be missing something.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
adhiamboperes: A comment here would be great, @manas-yu.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -18,66 +23,195 @@ private const val DEFAULT_PROCESS_TIMEOUT_MINUTES = 10L
...
-    val commandExecutor = CommandExecutorImpl(
-      scriptBgDispatcher,
-      processTimeout = processTimeout,
-      processTimeoutUnit = TimeUnit.MINUTES
-    )
+    try {
+      if (timer != null) {
+        timer.printMessage(
+          "Using ${workingDirectory.absolutePath} as an intermediary working directory"
+        )
```
adhiamboperes: I am a little confused by why we need 2 seperate print functions. Why are we invoking timer's print here?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -104,6 +116,59 @@ class AndroidLintRunnerTest {
     assertThat(exception).hasMessageThat().contains("Repository root path does not exist")
   }
 
+  @Test
+  fun testStop_afterRunning_returnsCorrectElapsedTime() {
```
adhiamboperes: Here and below:
```suggestion
  fun testStopTimer_afterRunning_returnsCorrectElapsedTime() {
```

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -104,6 +116,59 @@ class AndroidLintRunnerTest {
...
+  fun testStop_withNoTimeAdvance_returnsZero() {
+    elapsedTimeDisplayer.start()
+
+    val elapsedTime = elapsedTimeDisplayer.stop()
+
+    assertThat(elapsedTime).isEqualTo(0L)
+  }
+
+  @Test
+  fun testStart_calledMultipleTimes_onlyStartsOnce() {
```
adhiamboperes:
```suggestion
  fun testStartTimer_calledMultipleTimes_onlyStartsOnce() {
```

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -104,6 +116,59 @@ class AndroidLintRunnerTest {
...
+    elapsedTimeDisplayer.start() // Should be ignored
+
+    fakeTime += 1000L
+    val elapsedTime = elapsedTimeDisplayer.stop()
+
+    assertThat(elapsedTime).isEqualTo(1000L)
+  }
+
+  @Test
+  fun testFormatDuration_variousInputs() {
```
adhiamboperes: What is the expected outcome of the test?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -104,6 +116,59 @@ class AndroidLintRunnerTest {
...
+    fakeTime += 61000L
+    elapsed = elapsedTimeDisplayer.stop()
+    assertThat(elapsed).isEqualTo(61000L)
+
+    // Test 1 hour 1 minute 1 second
+    elapsedTimeDisplayer.start()
+    fakeTime += 3661000L
+    elapsed = elapsedTimeDisplayer.stop()
+    assertThat(elapsed).isEqualTo(3661000L)
+  }
```
adhiamboperes, line 170: Which scenario in the script does this set up represent?

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -649,35 +769,14 @@ class AndroidLintRunnerTest {
   }
 
   @Test
-  fun testAndroidLintAnalyzer_withSyntheticAccessor_detectsIssue() {
+  fun testAndroidLintAnalyzer_withSyntheticAccessor_issueSuppressed() {
```
adhiamboperes:
```suggestion
  fun testAndroidLintAnalyzer_withSyntheticAccessor_issueIsSuppressed() {
```

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
```diff
@@ -469,30 +592,27 @@ class AndroidLintRunnerTest {
...
+        emptySet()
+      )
     }
 
     assertThat(exception.message).contains("JDK home path does not exist or is not a directory")
   }
 
   @Test
-  fun testAndroidLintAnalyzer_withDuplicateStringResources_detectsIssue() {
+  fun testAndroidLintAnalyzer_withDuplicateStringResources_issueSuppressed() {
```
adhiamboperes:
```suggestion
  fun testAndroidLintAnalyzer_withDuplicateStringResources_issueIsSuppressed() {
```

adhiamboperes (changes requested):
Thanks @manas-yu! PTAL at my comments.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: The proto flag doesn’t need to be passed by developers/users since it’s intended only for easier testing. Similar to other scripts (e.g., [KdocValidityCheck](https://github.com/oppia/oppia-android/blob/47b9540586758a8f089394c9b34be7ca33ca4ad1/scripts/src/java/org/oppia/android/scripts/docs/KdocValidityCheck.kt#L56)), strict proto file validation isn’t enforced here because the flag is used solely for testing purposes, not for actual usage.

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: The issue is that the ElapsedTimeDisplayer continuously updates the console with elapsed time using \r to overwrite the same line. This creates a potential race condition where:

- The timer prints "Elapsed time: 00:01" and sets needsLineClear = true
- Regular println() tries to print a message
- The output gets confusing because the timer line wasn't properly cleared

I have updated it with the below impl to make it clearer and avoid an extra print funct
```kotlin
synchronized(System.out) {
     timer?.clearLine()
     println("Using ${workingDirectory.absolutePath} as an intermediary working directory")
}
```

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: Thanks! Added.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Done.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Done.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Done.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Done.

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: Updated test name to `testStopTimer_withVariousElapsedTimes_returnsCorrectDuration`

scripts/src/javatests/org/oppia/android/scripts/lint/AndroidLintRunnerTest.kt
manas-yu: It tests the elapsed time calculation - verifying that the ElapsedTimeDisplayer correctly calculates how much time has passed between start() and stop() calls.

manas-yu:
Hi @adhiamboperes PTAL addressed comments. Thanks!

Cc: @bhavukJain PTAL at open comments.

adhiamboperes (commented):
Thanks @manas-yu. Overrall, the implementation works. I am however thrown by the use of the `synchronized` keyword. We don't have an existing use case of this pattern, so I am assigning @BenHenning to check if it is justifiable to introduce it here. It might be ok, or we might want to use `withLock`

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
```diff
@@ -18,66 +23,178 @@ private const val DEFAULT_PROCESS_TIMEOUT_MINUTES = 10L
...
+      }
+    }
+  }
+
+  /**
+   * Clears the timer line if it's currently displayed, preparing console for new output.
+   * This method ensures proper synchronization with the timer display.
+   */
+  fun clearLine() {
+    synchronized(System.out) {
```
BenHenning: Do you need to synchronize output for everything writing to `System.out` or just the cases in this file?

What's also confusing is you're synchronizing printing with flushing--that should never be behaviorally necessary. As I recall, `PrintStream` (which is what `System.out` is under the surface) already synchronizes, so you can't run into cases where multiple lines overlap each other unless you have multiple print statements which we don't here.

If there _is_ a reason we must synchronize the printing to `needsLineClear` state (which is effectively the only thing these `synchronized` blocks are doing), then you can probably get away with a lock-free solution by using a `CoroutineScope` with a single thread and just queue all writes to it. This may be a good use of `ScriptBackgroundCoroutineDispatcher` though you'd need to extend its contract to support providing a single thread `ThreadPool`, but then it should just work (and you can have it cooperatively shutdown at the end of the script to avoid thread hanging or missed output).

BenHenning:
Apologies--followed up on the `synchronized` bit. I suspect it's simply not needed at all and that everything can be simplified, but if it is needed it can be made lock-free by using a `CoroutineDispatcher` with a single-threaded `CoroutineScope` (which is probably easiest done via `ScriptBackgroundCoroutineDispatcher` with a small change to its constructor).

scripts/src/java/org/oppia/android/scripts/lint/AndroidLintRunner.kt
manas-yu: Thanks, the synchronization was unnecessary. `PrintStream` already handles thread safety, and the synchronized blocks were only protecting the `needsLineClear` flag, which doesn't actually need coordination. The timer was only conflicting with the "Using ${workingDirectory.absolutePath} as an intermediary working directory" message in one of my previous implementations, but it's working fine now without any synchronization.

manas-yu:
Hi @adhiamboperes @BenHenning  PTAL.

adhiamboperes (approved):
LGTM, thanks @manas-yu

adhiamboperes:
@manas-yu, please review the failing tests.

manas-yu:
Hi @adhiamboperes PTAL, the test should pass now.

manas-yu:
Hi @adhiamboperes PTAL, the test should pass now. Thanks!

adhiamboperes (approved):
LGTM, thanks @manas-yu!

PR #5949, opened 2025-09-04
Fixes part of #5345: Implementing Force downloading .

theayushyadav11:
Fixes part of #5345 : 
This pull request adds a new "Force Download" feature to the developer options, allowing developers to manually trigger a download of remote platform parameters and feature flags, and restart the app to apply changes. The implementation includes new UI elements, dialog fragments, presenter logic, and integration with existing controllers and activity lifecycle.
The actual downloading implementation of platform parameter is not yet done, and will be done in future. This PR just wires the UI with the controller which can be verified by logs.

**Feature Addition: Force Download Functionality**
* Added a new `ForceDownloadsButtonClickListener` interface and integrated it into `DeveloperOptionsActivity`, enabling handling of force download button clicks. 
* Implemented `ForceDownloadDialogFragment` and its presenter to show a dialog for force downloading parameters, with options to restart the app or cancel. 
* Updated `DeveloperOptionsActivityPresenter` to handle force download requests and perform a full app restart if needed.

**UI Enhancements**
* Updated `developer_options_override_app_behaviors_view.xml` to add a force download button, divider, and related layout changes for improved organization and visibility.
* Bound the button's enabled state to a new LiveData property in the view model, reflecting whether force download is currently available

**Integration with ViewModels and Controllers**
* Extended `DeveloperOptionsViewModel` and `DeveloperOptionsOverrideAppBehaviorsViewModel` to support force download logic and button state, integrating with `ForceDownloadParametersController`. 
* Registered new listener and fragment for dependency injection and activity lifecycle management. 

**Verification via logs**
The downloading behaviour can be verified using the logs:
- When download starts it logs: "Calling Force Download of remote parameters".
- When download completes successfully: "Download finished successfully."
- When we cancel download before being finished: "Cancelled ongoing remote parameter download"
- All the logs can be verified using the TAG :
```
PlatformParameterController
```

### Screenshot
image

 
## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
[...]

PR #5953, opened 2025-09-06
Fixes part of #5345: Fixing restarting of app on config change.

theayushyadav11:
Fixes part of #5345 :
This pull request improves the handling of dialog persistence in the Feature Flags and Platform Parameters developer options screens, specifically ensuring that pending changes and restart dialogs remain visible after configuration changes (such as device rotation). It introduces a new state variable to track dialog visibility, updates fragment and presenter logic to restore dialogs, and adds tests to verify this behavior. Additionally, it refines the app restart logic to avoid unnecessary restarts during configuration changes.

**Dialog Persistence and State Management**

* Added an `PendingDialogListener` and implemented in `PlatformParametersFragment` and `FeatureFlagsFragment` when the save and discard are called from `PendingChangesDialogFragment`. 

**App Restart Logic**

* Modified the app restart logic in both presenters to only restart the app if the configuration is not changing, preventing unnecessary restarts during device rotation.

**Testing**

* Added new instrumented tests for both fragments to verify that dialogs (pending changes and restart dialogs) persist after configuration changes, improving reliability and user experience.

**Code Cleanup**

* Removed extra `onBackNavigation` code in the `PlatformParametersFragmentPresenter.kt` that might have been duplicated.

These changes ensure a smoother and more predictable experience for developers using the feature flags and platform parameters screens, especially when making changes that require confirmation dialogs.

#### Screenrecording: 

https://github.com/user-attachments/assets/4642d72f-bfce-4179-9524-0e7aaa3971a9

## Essential Checklist


theayushyadav11:
PTAL @Rd4dev

Rd4dev (changes requested):
Thanks, @theayushyadav11 -- for promptly pushing this. Regarding the fix, I have quite a few concerns:
- Can you confirm if the restart functionality works well with the changes made in the PR? I tested it locally, while the fix preserves the dialog across configuration changes, it currently breaks the app restart behavior — the app does not restart correctly after changes (i.e. it closes the app but doesn't relaunch it again).

https://github.com/user-attachments/assets/c7c7c58f-43a3-4ef3-a044-83eacd708725

- The current approach manually persists a UI dialog's open state in `savedInstanceState`, but `DialogFragment`s already handle configuration changes. Persisting UI state manually is unnecessary.
- The `isPendingDialogOpen` flag seems like a workaround rather than a proper fix. It looks fragile and triggers `onBackNavigation` again guarding them with flags.

### Key point to fix the issue:
- The main issue is not UI persistence, but the premature restart caused by `fragment`/`activity` recreation. Currently, `restartRequired` is set immediately after a `Success` result, causing `handleOnDestroy` to terminate the app even if the user hasn’t confirmed a restart. 
- Ideally, `restartRequired` should only be set when the **user explicitly confirms a restart**. Please verify that the fix both survives configuration changes and still correctly restarts the app when the user confirms a restart afterward.

https://github.com/oppia/oppia-android/blob/97bbf7716d3d19eee558d8912ce7004fed69685d/app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt#L202-L206

**Pending Changes Dialog (Save / Discard Dialog):**

- Additionally, though not noted in the evaluation results, the "Pending Changes dialogs" too does not seem to survive configuration changes. I believe it's necessary to persist them as well and would be nice if you could include those fixes as well here.

Please attach video demonstrations showing the full sequence of actions, including app restart after applying changes, to help reviewers verify the behavior.

Rd4dev:
Update: I’ve removed the earlier requirement regarding persisting the Pending Changes Dialog, as it appears to survive configuration changes correctly. I’ll follow up if I notice any unexpected behavior with that later.

Edit: Ah, it was working with the changes in the PR, and not with the base implementation, and with just the 'force_downloads' PR the Pending Changes dialog doesn't persist config changes, so it's something that will need to be fixed as well.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -309,4 +302,26 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
...
+  fun markRestartRequired() {
+    restartRequired = true
+  }
+
+  /**
+   * Called when [FeatureFlagsFragment] is destroyed.
+   * Performs a fresh restart of the app to load any updated feature flag states, if required.
+   */
+  fun handleOnDestroy() {
+    if (restartRequired) {
```
theayushyadav11: Hi @Rd4dev , according to the new logic the value of restart required is comming as expected but it is not behaving, I have verified the logs it is comming true but not entering the if body.
This is observed only when pending changes dialog is called and then restart dialog is called simultaneously.
Can you please take a look and tell whats the actual issue and how to fix it?

seanlip: @theayushyadav11 You should investigate this at your end too. Where's your debugging doc for it?

theayushyadav11: Yes @seanlip , I am investigating it too and currently analysing the logs in the locgat and looking for a probable fix.

seanlip: OK, thanks. The main point I want to make is the debugging doc (see [wiki](https://github.com/oppia/oppia/wiki/Debugging-Docs)). Please complete it with the necessary info (so that all the findings are in one place) and link to it from here. Then RD can help guide you.

theayushyadav11: Hi @Rd4dev here is the [debugging doc,](https://docs.google.com/document/d/1pj3Bx8h3XPgUAWRnTtecqabFXzkMAyGdkMk4za2ezFc/edit?tab=t.0#heading=h.yz68ilctwvg) will addup more if I get something;

adhiamboperes:
I have thoughts on why the app reopen behavior is not working as expected. Thinking critically, the way restart is currently handled is problematic.

```kotlin
fun handleOnDestroy() {
  if (restartRequired) {
    val intent = Intent(activity, SplashActivity::class.java).also {
      it.action = Intent.ACTION_MAIN
      it.addCategory(Intent.CATEGORY_LAUNCHER)
    }
    activity.startActivity(intent)

    // Kill the process
    exitProcess(0)
  }
}
```

**What happens:**

1. onDestroy() runs  -  Android is destroying your Activity. This might be due to:
User pressing back,
finish() called,
Configuration change (rotation, language change),
Or system killing it for memory.

2. If restartRequired is true:
You start SplashActivity immediately.
Then you call exitProcess(0), which terminates the entire app process.

3. Result:
The process is killed before SplashActivity can actually finish launching.
Depending on timing, the new activity may never appear, because exitProcess(0) halts the VM immediately.
If it does appear, it will immediately die since the process is gone.

**Problems with this approach:**

`exitProcess(0)` is very abrupt. It doesn’t give Android time to finish launching the new activity.
On some devices it may look like the app just crashed instead of restarting cleanly.
If this happens during a configuration change, you may end up in a loop.

theayushyadav11:
> I have thoughts on why the app reopen behavior is not working as expected. Thinking critically, the way restart is currently handled is problematic.
> 
> ```kotlin
> fun handleOnDestroy() {
>   if (restartRequired) {
>     val intent = Intent(activity, SplashActivity::class.java).also {
>       it.action = Intent.ACTION_MAIN
>       it.addCategory(Intent.CATEGORY_LAUNCHER)
>     }
>     activity.startActivity(intent)
> 
>     // Kill the process
>     exitProcess(0)
>   }
> }
> ```
> 
> **What happens:**
> 
> 1. onDestroy() runs  -  Android is destroying your Activity. This might be due to:
>    User pressing back,
>    finish() called,
>    Configuration change (rotation, language change),
>    Or system killing it for memory.
> 2. If restartRequired is true:
>    You start SplashActivity immediately.
>    Then you call exitProcess(0), which terminates the entire app process.
> 3. Result:
>    The process is killed before SplashActivity can actually finish launching.
>    Depending on timing, the new activity may never appear, because exitProcess(0) halts the VM immediately.
>    If it does appear, it will immediately die since the process is gone.
> 
> **Problems with this approach:**
> 
> `exitProcess(0)` is very abrupt. It doesn’t give Android time to finish launching the new activity. On some devices it may look like the app just crashed instead of restarting cleanly. If this happens during a configuration change, you may end up in a loop.

Thanks @adhiamboperes for the describing the app behaviour cleanly.
Yes `exitProcess()` closes all the ongoing tasks in the app. But we need this because we want the processState to get reinitialise on startup, if `exitProcess()` isnt used app restarts with the previous instance only and the overriddes arn't applied to the app behaviour.
Here app restart works well in normal condition, but when config change is done then , it doesn't restarts.
This is the main issue.

theayushyadav11:
Hi @adhiamboperes  @Rd4dev  using `exitProcess(0)` directly in the dialogFragmentPresenter tends to behave well, but have doubt on the practices and better lifecycle, that it should be implemented there or not..

adhiamboperes:
Please clarify what you mean @theayushyadav11.

Rd4dev:
Thanks, @adhiamboperes, for the notes, and @theayushyadav11 — firstly, starting with why the restart doesn’t work on orientation change. The restart only fails on configuration changes, not normally. This isn’t because the logic itself is wrong, but because it never gets called in that case — it’s guarded by the `isChangingConfigurations` check. 

```kt
fun handleOnDestroy() {
    if (restartRequired && !activity.isChangingConfigurations) { // This blocks the logic on config change.
      val intent = Intent(activity, SplashActivity::class.java).also {
        it.action = Intent.ACTION_MAIN
        it.addCategory(Intent.CATEGORY_LAUNCHER)
      }
      activity.startActivity(intent)
      // App is terminated to ensure a fresh restart and kill all the current process
      // so that ProcessState can be reinitialised on the fresh restart.
      exitProcess(0)
    }
  }
```

So, the app closes due to `finishAffinity()`, but the **restart never happens**. You can test this by either removing the entire `handleOnDestroy` logic or just adding logs to confirm that it’s not actually called on config change.

<h1></h1>

That being said, my thoughts on resolving this:

1. The reason I preferred handling `exitProcess(0)` inside `onDestroy()` is to ensure the activity is fully closed before killing the app. Calling it directly on button click is a quick fix and will work, but I found it breaks lifecycle flow.
2. The actual issue is that `restartRequired` isn’t set correctly. Ideally, it should only be set when the user explicitly clicks the restart button. To achieve this, we should propagate the confirmation from the dialog’s restart button back to the fragment. The clean way is via a listener (e.g., define something like `AppRestartListener` with `fun onRestartConfirmed()`), then implement it in the fragment so the presenter can set its own `restartRequired = true` at the right moment.

For implementation you can refer to existing similar patterns (e.g., `ConceptCardListener`). 

https://github.com/oppia/oppia-android/blob/d3d6f821bc3e325335a874e5c16ec856686e1522/app/src/main/java/org/oppia/android/app/topic/conceptcard/ConceptCardListener.kt#L4-L7

https://github.com/oppia/oppia-android/blob/d3d6f821bc3e325335a874e5c16ec856686e1522/app/src/main/java/org/oppia/android/app/topic/revisioncard/RevisionCardActivity.kt#L123-L125

I believe following this, it ensures `restartRequired` is only set when restart is explicitly confirmed, lets the fragment/activity close appropriately, and still leverages `onDestroy()` to handle the restart cleanly. It also blends well with the current setup with minimal changes; we can skip all guarding flags and maintain a clean restart.

PTAL — and I’d like to hear your thoughts too, also @adhiamboperes's thoughts, on best practices here.

theayushyadav11:
> I believe following this, it ensures `restartRequired` is only set when restart is explicitly confirmed, lets the fragment/activity close appropriately, and still leverages `onDestroy()` to handle the restart cleanly. It also blends well with the current setup with minimal changes; we can skip all guarding flags and maintain a clean restart.
> 
> PTAL — and I’d like to hear your thoughts too, also @adhiamboperes's thoughts, on best practices here.

Hi @Rd4dev, Agree on `activity.isChangingConfigurations` part.
But when I removed that and used the function to directly set on `restartRequired` variable and then log the values of it, despite of being true and entering into the if block, restart doesn't occur, shared the logs in [debugging doc](https://docs.google.com/document/d/1pj3Bx8h3XPgUAWRnTtecqabFXzkMAyGdkMk4za2ezFc/edit?tab=t.0#heading=h.rpq25vez37fp).
I tried with using Listener and set the `restartRequired` value ( though settinh of its value was working fine), still it fails to restart after config changes.

Rd4dev:
> > I believe following this, it ensures `restartRequired` is only set when restart is explicitly confirmed, lets the fragment/activity close appropriately, and still leverages `onDestroy()` to handle the restart cleanly. It also blends well with the current setup with minimal changes; we can skip all guarding flags and maintain a clean restart.
> > PTAL — and I’d like to hear your thoughts too, also @adhiamboperes's thoughts, on best practices here.
> 
> Hi @Rd4dev, Agree on `activity.isChangingConfigurations` part. But when I removed that and used the function to directly set on `restartRequired` variable and then log the values of it, despite of being true and entering into the if block, restart doesn't occur, shared the logs in [debugging doc](https://docs.google.com/document/d/1pj3Bx8h3XPgUAWRnTtecqabFXzkMAyGdkMk4za2ezFc/edit?tab=t.0#heading=h.rpq25vez37fp). I tried with using Listener and set the `restartRequired` value ( though settinh of its value was working fine), still it fails to restart after config changes.

Thanks @theayushyadav11, 
```sh
2025-09-06 23:42:05.322 23561-23561 TestRestartConfig       org.oppia.android                    D  handleOnDestroy: restartRequired=false
2025-09-06 23:42:07.382 23561-23561 TestRestartConfig       org.oppia.android                    D  handleOnDestroy: restartRequired=true
2025-09-06 23:42:07.384 23561-23561 TestRestartConfig       org.oppia.android                    D  Restarting app to apply feature flag changes
```

from the above logs I can see it being logged as true, but to actually understand the trigger can you share the listener logic you implemented, it would be nice to remove the other unrelated changes or outdated changes like the guarding flags for persisting UI, in the PR to get rid of other components affecting it and focus solely on the issue here.

theayushyadav11:
Hi @Rd4dev here is the code:

FragmentPresenter:
```kotlin
 fun restartApp() {
   restartrequired = true
    activity.finishAffinity()
  }

  fun handleOnDestroy() {
    oppiaLogger.d("TestRestartConfig", "restartrequired: $restartrequired")
    if(restartrequired) {
      oppiaLogger.d("TestRestartConfig", "Restarting App")
      val intent = Intent(activity, SplashActivity::class.java).also {
        it.action = Intent.ACTION_MAIN
        it.addCategory(Intent.CATEGORY_LAUNCHER)
      }
      activity.startActivity(intent)
      // App is terminated to ensure a fresh restart and kill all the current process
      // so that ProcessState can be reinitialised on the fresh restart.
      exitProcess(0)
    }
  }
```
Fragment:
```kotlin
  override fun restartApp() {
    featureFlagsFragmentPresenter.restartApp()
  }

  override fun onDestroyView() {
    super.onDestroyView()
    featureFlagsFragmentPresenter.handleOnDestroy()
  }

```

Logs:
```
2025-09-07 05:57:09.087 20288-20288 TestRestartConfig       org.oppia.android                    D  restartrequired: false
2025-09-07 05:57:10.370 20288-20288 TestRestartConfig       org.oppia.android                    D  restartrequired: false
2025-09-07 05:57:14.025 20288-20288 TestRestartConfig       org.oppia.android                    D  restartrequired: false
2025-09-07 05:57:15.263 20288-20288 TestRestartConfig       org.oppia.android                    D  restartrequired: false
2025-09-07 05:57:17.445 20288-20288 TestRestartConfig       org.oppia.android                    D  restartrequired: true
2025-09-07 05:57:17.446 20288-20288 TestRestartConfig       org.oppia.android                    D  Restarting App
```
Edit:-
While skipping `onDestroy` works fine:
```kotlin
 fun restartApp() {
    val intent = Intent(activity, SplashActivity::class.java).also {
      it.action = Intent.ACTION_MAIN
      it.addCategory(Intent.CATEGORY_LAUNCHER)
    }
    activity.finishAffinity()
    activity.startActivity(intent)
    // App is terminated to ensure a fresh restart and kill all the current process
    // so that ProcessState can be reinitialised on the fresh restart.
    exitProcess(0)
  }
```

adhiamboperes:
@theayushyadav11, we discussed [here](https://github.com/oppia/oppia-android/pull/5949#discussion_r2328353579) that you should replace ondestroy, but the PR does not look up to date. Have you tested with the suggested changes?

theayushyadav11:
> @theayushyadav11, we discussed [here](https://github.com/oppia/oppia-android/pull/5949#discussion_r2328353579) that you should replace ondestroy, but the PR does not look up to date. Have you tested with the suggested changes?


Yes @adhiamboperes, I have replaced the `onDestroy` with `restartApp` and it works as intended.

Rd4dev:
@theayushyadav11, as I shared in the chat conversation, I might be away until the afternoon, so I’m sharing what I tried locally to quickly unblock things in the way I mentioned above.

I worked off commit https://github.com/oppia/oppia-android/pull/5949/commits/55db30976494ddce27c73a1da284482fca58360b from the ForceDownload flow since that’s the point before removing the `onDestroy` handling later.

The approach is:

1. Define an interface:

```kotlin
interface AppRestartListener {
  fun onRestartConfirmed()
}
```

2. In the dashboard, implement this and forward the callback:

```kotlin
class FeatureFlagsFragment : InjectableFragment(), RestartListener {
  override fun onRestartConfirmed() {
    featureFlagsFragmentPresenter.handleRestartConfirmation()
  }

  override fun onDestroy() {
    super.onDestroy()
    featureFlagsFragmentPresenter.handleOnDestroy()
  }
}
```

3. Update the presenter with:

```kotlin
fun handleRestartConfirmation() {
  restartRequired = true
}

fun handleOnDestroy() {
  if (restartRequired) {
    val intent = Intent(activity, SplashActivity::class.java).also {
      it.action = Intent.ACTION_MAIN
      it.addCategory(Intent.CATEGORY_LAUNCHER)
    }
    activity.startActivity(intent)
    // App is terminated to ensure a fresh restart and kill all the current process
    // so that ProcessState can be reinitialised on the fresh restart.
    exitProcess(0)
  }
}
```

4. Finally, wire up the restart button in `AppRestartDialogFragmentPresenter`:

```kotlin
binding.restartButton.setOnClickListener {
  dialog.dismiss()
  (fragment.parentFragment as? AppRestartListener)?.onRestartConfirmed()
    ?: (fragment.requireActivity() as? AppRestartListener)?.onRestartConfirmed()
  activity.finishAffinity()
}
```

I’ve tested this locally and it restarts with config changes as well.

<h1></h1>

Thanks, @adhiamboperes, for pointing to your [earlier comment](https://github.com/oppia/oppia-android/pull/5953#issuecomment-3263314042) — it was really helpful since it links to many resources related to this PR. I think the approach I used and the one you suggested are quite close in direction. The only difference is that I was planning to leverage `onDestroy` for handling the restart (since it’s also invoked after `finishAffinity`), while your suggestion was to trigger restart directly. Just wanted your thoughts on whether we should continue using `onDestroy` to follow lifecycle, or if the direct restart approach is preferred. I believe either should work but would be helpful to have your guidance on understanding the best practice here.

@theayushyadav11 — could you also take a look at the above? It would be helpful if you can just **keep the related changes aligned with the PR context** so I can pull them in quickly to test locally and see what’s happening. I’ll continue on this by afternoon if it’s still blocked, but with the above code working and Adhiambo’s direction on best practices, you can incorporate both to complete the implementation. Thanks.

Rd4dev:
@theayushyadav11 - Reiterating from the chat discussion and above threads — please keep the related changes up to date, removing old code (`isPendingDialogOpen` if it's unused) so that reviewers can go through it quickly. Even with Adhiambo’s direction, things will work fine; my earlier request was mainly about proper wiring on if we should use `onDestroy` or a direct restart call. From your discussion, though, it seems you’re mentioning that the restart itself isn’t working.

theayushyadav11:
> @theayushyadav11 - Reiterating from the chat discussion and above threads — please keep the related changes up to date, removing old code (`isPendingDialogOpen` if it's unused) so that reviewers can go through it quickly. Even with Adhiambo’s direction, things will work fine; my earlier request was mainly about proper wiring on if we should use `onDestroy` or a direct restart call. From your discussion, though, it seems you’re mentioning that the restart itself isn’t working.

@Rd4dev `isPendingDialogOpen` is still being used. As per the Adhiambo's suggestion restart functionality is working totally fine.

Rd4dev:
> > @theayushyadav11 - Reiterating from the chat discussion and above threads — please keep the related changes up to date, removing old code (`isPendingDialogOpen` if it's unused) so that reviewers can go through it quickly. Even with Adhiambo’s direction, things will work fine; my earlier request was mainly about proper wiring on if we should use `onDestroy` or a direct restart call. From your discussion, though, it seems you’re mentioning that the restart itself isn’t working.
> 
> @Rd4dev `isPendingDialogOpen` is still being used. As per the Adhiambo's suggestion restart functionality is working totally fine.

Ah, that's awesome then, sounds good for the restart — I have no concerns there. But regarding `isPendingDialogOpen`, why do we need that flag, you can utilize `DialogFragment` if it's for UI persistence?

app/BUILD.bazel
```diff
@@ -81,7 +81,9 @@ LISTENERS = [
     "src/main/java/org/oppia/android/app/administratorcontrols/RouteToLearnerAnalyticsListener.kt",
     "src/main/java/org/oppia/android/app/administratorcontrols/RouteToProfileListListener.kt",
     "src/main/java/org/oppia/android/app/administratorcontrols/ShowLogoutDialogListener.kt",
+    "src/main/java/org/oppia/android/app/devoptions/AppRestartListener.kt",
     "src/main/java/org/oppia/android/app/devoptions/ForceCrashButtonClickListener.kt",
+    "src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersButtonClickListener.kt",
```
adhiamboperes: Not sure why I can see these here when they were added in the previous PR. Please update branch with develop and resolve any conflicts.

app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragmentPresenter.kt
```diff
@@ -22,14 +22,15 @@ class AppRestartDialogFragmentPresenter @Inject constructor(
     )
     binding.lifecycleOwner = fragment
 
+    val appRestartInterface = fragment.parentFragment as AppRestartListener
     val dialog = AlertDialog.Builder(activity, R.style.OppiaAlertDialogTheme)
```
adhiamboperes, line 26:
```suggestion
    val dialog = AlertDialog.Builder(fragment.activity, R.style.OppiaAlertDialogTheme)
```

Then you can remove `private val activity: AppCompatActivity,`  from the constructor and its import

app/src/main/java/org/oppia/android/app/devoptions/AppRestartListener.kt
adhiamboperes, line 1: This also looks like a case of merge conflicts

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
```diff
@@ -77,17 +81,18 @@ class FeatureFlagsFragment : InjectableFragment() {
         .setOverriddenValue(it.value)
         .build()
     }
+    val isPendingDialogOpen = featureFlagsFragmentPresenter.isPendingDialogOpen
```
adhiamboperes: @Rd4dev is correct in flagging this, this variable doesn't do anythinng. You created a public variable in the presenter(bad practice, breaks encapsulation), calling it in the fragment and then passing it back to the presenter in `handleCreateView` above.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
```diff
@@ -77,17 +81,18 @@ class FeatureFlagsFragment : InjectableFragment() {
...
 
     val proto = FeatureFlagsFragmentStateBundle.newBuilder()
       .addAllFeatureFlagStates(featureFlagStates)
       .addAllResetFeatureFlags(resetFlags)
+      .setIsPendingDialogOpen(isPendingDialogOpen)
```
adhiamboperes: Instead of saving that PendingDialog is open, it would be better to save the actual `resetFlags` and `featureFlagStates` lists to `SavedInstanceState`,  and I can see that this is already handled, as @Rd4dev pointed out. I would suggest that you test the branch without this `setIsPendingDialogOpen`  field.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
```diff
@@ -39,6 +40,7 @@ class FeatureFlagsFragment : InjectableFragment() {
   ): View {
     val featureFlagStates: MutableMap<FeatureFlagId, Boolean> = mutableMapOf()
     val resetFlags: MutableMap<FeatureFlagId, Boolean> = mutableMapOf()
+    var isPendingDialogOpen = false
```
adhiamboperes: This needs to go away in light of the comments below.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -44,14 +44,17 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
   private lateinit var binding: FeatureFlagsFragmentBinding
   private lateinit var linearLayoutManager: LinearLayoutManager
   private lateinit var bindingAdapter: BindableAdapter<FeatureFlagItemViewModel>
-  private var restartRequired: Boolean = false
+
+  /** Indicates whether the pending changes dialog is open. */
+  var isPendingDialogOpen: Boolean = false
```
adhiamboperes: This is not needed.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -118,6 +125,7 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
   }
 
   private fun showPendingChangesDialog(overriddenFlags: List<OverriddenFeatureFlag>) {
+    isPendingDialogOpen = true
```
adhiamboperes: This is really bad design because a presenter should orchestrate UI logic, not track ad-hoc booleans about the transient UI state, and is not lifescycle safe.

Suppose the system destroys and recreates the fragment (rotation, low-memory kill).

`isPendingDialogOpen` will reset to false on recreation, even if the dialog is still showing.

Now your presenter’s idea of state is out of sync with what the user actually sees.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -136,8 +145,9 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
       dialog.dismiss()
       activity.finish()
     }
-
-    dialog.show()
+    if (!activity.isFinishing && !activity.isDestroyed) {
```
adhiamboperes: This conditional check is probably not necessary now that the fragment should be handling lifecycle states.

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragment.kt
```diff
@@ -42,7 +43,7 @@ class PlatformParametersFragment : InjectableFragment() {
     val platformParameterStates:
       MutableMap<PlatformParameterId, PlatformParameterValue?> = mutableMapOf()
     val resetParamList: MutableMap<PlatformParameterId, PlatformParameterValue> = mutableMapOf()
-
+    var isPendingDialogOpen = false
```
adhiamboperes: Same suggested imp here as the FeatureFlagsFragment.

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
```diff
@@ -53,14 +53,17 @@ class PlatformParametersFragmentPresenter @Inject constructor(
   private val invalidInputErrorText =
     resourceHandler.getStringInLocale(R.string.platform_parameter_invalid_input_error_msg)
   private val boundParamIds = mutableSetOf<PlatformParameterId>()
-  private var restartRequired: Boolean = false
+
+  /** Indicates whether the pending changes dialog is open. */
+  var isPendingDialogOpen: Boolean = false
```
adhiamboperes: Same suggested imp here as the FeatureFlagsFragmentPresenter.

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
```diff
@@ -103,6 +97,10 @@ class PlatformParametersFragmentPresenter @Inject constructor(
       platformParameterViewModel.resetParameters.value = resetParameters.toMutableMap()
     }
 
+    if (isPendingDialogOpen) {
+      onBackNavigation()
+    }
```
adhiamboperes: This should be reverted

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -82,6 +85,10 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
       featureFlagsViewModel.resetFlags.value = resetFlags.toMutableMap()
     }
 
+    if (isPendingDialogOpen) {
+      onBackNavigation()
+    }
```
adhiamboperes: This looks out of place. Why was this added?

adhiamboperes (changes requested):
@theayushyadav11, please revise and test the app per the comments, and please communicate if the proposed solution does not appear to work(it should).

Additionally:
1. Enable pre-push checks on your local repo to avoid wasting time with avoidable lint failures. Please see [wiki section ](https://github.com/oppia/oppia-android/wiki/Installing-Oppia-Android#install-oppia-android)regarding `setup.sh`.
2. Run an app and the affected test file build before pushing to catch failures before pushing to CI. Local builds are faster and will avoid wasting 20 minutes on CI.
3. I missed something in my earlier review of `PlatformParametersFragmentPresenter` and `FeatureFlagsFragmentPresenter`. Since both have:

```kotlin
    binding.apply {
      this.lifecycleOwner = fragment
      this.viewModel = featureFlagsViewModel
    }
```

Please move binding of the views into the apply block, i.e `featureFlagsRecyclerView`, ` and `saveButton` for both presenters.

adhiamboperes:
> > @Rd4dev `isPendingDialogOpen` is still being used. As per the Adhiambo's suggestion restart functionality is working totally fine.
> 
> Ah, that's awesome then, sounds good for the restart — I have no concerns there. But regarding `isPendingDialogOpen`, why do we need that flag, you can utilize `DialogFragment` if it's for UI persistence?


Yes, if  the`AlertDialog` isn't persisting despite the changes, then a `DialogFragment` should be preferred.

theayushyadav11:
Thanks @adhiamboperes @Rd4dev for your suggestions, I have now made a new fragment for the `pendingChnagesDialog`, removing the `isPendingDialogOpen` and things now works as expected.

#### Recording:

https://github.com/user-attachments/assets/645bb0a8-0e31-49f6-932a-582f7198a38f

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
theayushyadav11: Thanks for the details, removed this

app/BUILD.bazel
theayushyadav11: updated

app/src/main/java/org/oppia/android/app/devoptions/AppRestartListener.kt
theayushyadav11: updated

app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragmentPresenter.kt
theayushyadav11: `activity` is also used in :
```kotlin
 val binding = AppRestartDialogFragmentBinding.inflate(
      activity.layoutInflater,
      /* parent= */ null,
      /* attachToRoot= */ false
    )
```
and there it should not be `nullable` so utilising `activity` over `fragment.activity`.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
theayushyadav11: Yeah, removed `isPendingDialogOpen` and utilised a seperate fragment for this dialog which persists on the configuration changes.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragment.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
theayushyadav11: Removed

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragment.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: removed

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: removed this way of handling.

theayushyadav11:
PTAL @Rd4dev .

app/src/main/java/org/oppia/android/app/devoptions/PendingChangesDialogFragmentPresenter.kt
```diff
@@ -0,0 +1,42 @@
...
+      /* parent= */ null,
+      /* attachToRoot= */ false
+    )
+    binding.lifecycleOwner = fragment
+
+    val pendingChangesInterface = fragment.parentFragment as PendingChangesListener
+    val dialog = AlertDialog.Builder(activity, R.style.OppiaAlertDialogTheme)
+      .setView(binding.root)
+      .create()
+    dialog.setCanceledOnTouchOutside(false)
```
Rd4dev: The dialog should be dismissible so that users can remain on the dashboard when clicked anywhere outside.

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
```diff
@@ -36,6 +36,9 @@ import kotlin.system.exitProcess
 /** Tag for displaying [AppRestartDialogFragment]. */
 const val TAG_PLATFORM_PARAMETER_RESTART_DIALOG = "PLATFORM_PARAMETER_RESTART_DIALOG_TAG"
 
+/** Tag for displaying [PendingChangesDialogFragment]. */
+const val TAG_PLATFORM_PARAMETER_PENDING_CHANGES_DIALOG = "FEATURE_FLAG_PENDING_CHANGES_DIALOG_TAG"
```
Rd4dev:
```suggestion
const val TAG_PLATFORM_PARAMETER_PENDING_CHANGES_DIALOG = "PLATFORM_PARAMETER_PENDING_CHANGES_DIALOG_TAG"
```

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -309,4 +285,20 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
   fun getFeatureFlagStates(): Map<FeatureFlagId, Boolean> {
     return featureFlagsViewModel.featureFlagStates.value ?: mapOf()
   }
+
+  /**
+   * Called when restart is triggered by [AppRestartDialogFragment].
+   * Performs a fresh restart of the app to load any updated feature flag states, if required.
+   */
+  fun restartApp() {
```
Rd4dev: Since this now only handles the restart behavior, and we always trigger a restart when needed, the phrase ‘if required’ implies a conditional, which isn’t accurate.
```
Performs a fresh restart of the app to reload feature flag states and reinitialize the app processState.
```

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
```diff
@@ -476,19 +459,18 @@ class PlatformParametersFragmentPresenter @Inject constructor(
   }
 
   /**
-   * Called when [PlatformParametersFragment] is destroyed.
+   * Called when restart is triggered by [AppRestartDialogFragment].
    * Performs a fresh restart of the app to load any updated feature flag states, if required.
```
Rd4dev: Ditto.

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
```diff
@@ -1110,91 +1109,87 @@ class FeatureFlagsFragmentTest {
...
-          .inRoot(isDialog())
-          .perform(click())
-
-        testCoroutineDispatchers.runCurrent()
-
-        onView(withText(R.string.app_restart_dialog_title))
-          .inRoot(isDialog())
-          .perform(click())
-      }
+    launch(FeatureFlagsTestActivity::class.java).use { scenario ->
```
Rd4dev: I can see this test + a number of below tests has been modified to only verify the rendering of the dialog alert, whereas previously it was also checking whether the app actually closes (i.e., calling `System.exit()`), which seems to be missing now. Could you also share more details?

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
```diff
@@ -1110,91 +1109,87 @@ class FeatureFlagsFragmentTest {
...
+
+      onView(withId(R.id.save_button)).perform(click())
+
+      testCoroutineDispatchers.runCurrent()
+
+      onView(withText(R.string.app_restart_dialog_title))
+        .inRoot(isDialog())
+        .check(matches(isDisplayed()))
+      testCoroutineDispatchers.runCurrent()
+      scenario.close()
```
Rd4dev: Here and below as well — every test seems to have been altered. Please consider the above comment for all the following test cases, including those for Platform Parameters.

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
```diff
@@ -1246,6 +1240,102 @@ class FeatureFlagsFragmentTest {
     }
   }
 
+  @Test
+  fun testFeatureFlagsFragment__clickSave_showsRestartDialog_configChange_dialogPersists() {
```
Rd4dev: `testFeatureFlagsFragment_`
```suggestion
  fun testFeatureFlagsFragment_clickSave_showsRestartDialog_configChange_dialogPersists() {
```

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
```diff
@@ -1246,6 +1240,102 @@ class FeatureFlagsFragmentTest {
...
+      testCoroutineDispatchers.runCurrent()
+      scrollToPosition(0)
+      onView(
+        atPositionOnView(
+          recyclerViewId = R.id.feature_flags_recycler_view,
+          position = 0,
+          targetViewId = R.id.feature_flag_switch
+        )
+      ).perform(click())
+      onView(withId(R.id.save_button)).perform(click())
```
Rd4dev, line 1298: Here and elsewhere (including the tests below and the Platform Parameters tests as well):
Nit: Please add a line break to separate the two actions (toggling the switch and clicking the save button).

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
```diff
@@ -597,29 +597,6 @@ class DeveloperOptionsFragmentTest {
     }
   }
 
-  @Test
-  fun testDeveloperOptionsFragment_clickDownload_showsDialog_clickRestart_exitsApp() {
```
Rd4dev: Ditto: This removes the test to verify `force_download` completely here

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
```diff
@@ -1246,6 +1240,102 @@ class FeatureFlagsFragmentTest {
...
+      testCoroutineDispatchers.runCurrent()
+
+      onView(withText(R.string.app_restart_dialog_title))
+        .inRoot(isDialog())
+        .check(matches(isDisplayed()))
+    }
+  }
+
+  @Test
+  fun testFeatureFlagsFragment_showsPendingChangesDialog_configChange_dialogPersists() {
```
Rd4dev: Here and for Platform Parameters as well, since shows implies a result with no action defined.
```suggestion
  fun testFeatureFlagsFragment_pendingChangesDialog_configChange_dialogPersists() {
```

Rd4dev (changes requested):
Thanks, @theayushyadav11 — the logic looks good. However, regarding the test cases, I noticed most of them have been modified to accommodate the changes, and it's now no longer verifying exit. Could you explain the reason behind this and the approaches you tried? You had mentioned this briefly in the chat, but it’s always helpful to share it here as well, since other reviewers mainly Adhiambo could have assisted in the meantime.

This is particularly important as we need to wrap up this fix today to allow evaluations to proceed. In general, tests are written to safeguard new changes, and figuring out why they are failing is critical to either fixing the issue or identifying a valid workaround if unavoidable. Could you please share the details now, and also assign Adhiambo help verifying the earlier addressed comments so we can keep this moving?

theayushyadav11:
Hi @Rd4dev @adhiamboperes  , regarding the modified tests, those tests were failing with the message that "cannot perform click on the button" and so to verify the changes after saving it to database and closing the scenerio works as database updation is done when the restart button is shown. This happens as roboelectric is not able to perform click on the button.

Here is the test:
```kotlin
@Test
fun testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp() {

  val exception = assertThrows<SecurityException> {
    launch(FeatureFlagsTestActivity::class.java).use { scenerio ->

      scrollToPosition(0)
      onView(
        atPositionOnView(
          recyclerViewId = R.id.feature_flags_recycler_view,
          position = 0,
          targetViewId = R.id.feature_flag_switch
        )
      ).perform(click())

      pressBack()

      onView(withText(R.string.pending_changes_dialog_save_button_text))
        .inRoot(isDialog())
        .perform(click())

      onView(withText(R.string.app_restart_dialog_button_text))
        .inRoot(isDialog())
        .perform(click())

      scenerio.close()
    }
  }

  assertThat(exception.message).contains("System.exit()")
}
```
Error log: 
```
1) testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp
   (org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest)

java.lang.AssertionError: Expected exception of type: class java.lang.SecurityException,
not: class androidx.test.espresso.PerformException.
    at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest
        .testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp
        (FeatureFlagsFragmentTest.kt:1682)
    ... 18 trimmed

Caused by: androidx.test.espresso.PerformException: Error performing
    'single click - At Coordinates: 647, 345 and precision: 16, 16'
    on view 'an instance of android.widget.TextView and view.getText() equals
    string from resource id: <2131820652> [app_restart_dialog_button_text] value: Restart'.
    at androidx.test.espresso.PerformException$Builder.build(PerformException.java:1)
    at androidx.test.espresso.base.PerformExceptionHandler.handleSafely(PerformExceptionHandler.java:8)
    at androidx.test.espresso.base.PerformExceptionHandler.handleSafely(PerformExceptionHandler.java:9)
    at androidx.test.espresso.base.DefaultFailureHandler$TypedFailureHandler.handle(DefaultFailureHandler.java:4)
    at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:5)
    at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:8)
    at androidx.test.espresso.ViewInteraction.desugaredPerform(ViewInteraction.java:11)
    at androidx.test.espresso.ViewInteraction.perform(ViewInteraction.java:8)
    at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest
        $testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp$exception$1.invoke
        (FeatureFlagsFragmentTest.kt:1137)
    at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest
        $testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp$exception$1.invoke
        (FeatureFlagsFragmentTest.kt:1113)
    at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest
        .testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp
        (FeatureFlagsFragmentTest.kt:1679)

```

Rd4dev:
> Hi @Rd4dev @adhiamboperes , regarding the modified tests, those tests were failing with the message that "cannot perform click on the button" and so to verify the changes after saving it to database and closing the scenerio works as database updation is done when the restart button is shown. This happens as roboelectric is not able to perform click on the button.

Thanks, @theayushyadav11, for sharing this. I don’t think the issue is with `perform(click())`, and I believe closing the scenario isn’t accurately what’s helping here for our case. I ran the test locally, and the logs further down show the real problem: it’s still failing with a `SecurityException`. The stack trace confirms the click action succeeds — it reaches the call to `System.exit()`. However, that exception is intercepted and wrapped inside a `PerformException`, because now the logic to restart triggers immediately on `perform(click())` rather earlier via `onDestroy`.

```kt
There was 1 failure:
1) testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp(org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest)
java.lang.AssertionError: Expected exception of type: class java.lang.SecurityException, not: class androidx.test.espresso.PerformException.
        at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest.testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp(FeatureFlagsFragmentTest.kt:1670)
        ... 18 trimmed
Caused by: androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 647, 345 and precision: 16, 16' on view 'an instance of android.widget.TextView and view.getText() equals string from resource id: <2131820652> [app_restart_dialog_button_text] value: Restart'.
        at androidx.test.espresso.PerformException$Builder.build(PerformException.java:1)
        at androidx.test.espresso.base.PerformExceptionHandler.handleSafely(PerformExceptionHandler.java:8)

        ... 19 more
Caused by: java.lang.SecurityException: Test code should never call System.exit()
        at com.google.testing.junit.runner.util.GoogleTestSecurityManager.checkExit(GoogleTestSecurityManager.java:34)
        at java.base/java.lang.Runtime.exit(Runtime.java:114)
```

The test is failing because the `assertThrows<SecurityException>` block is not correctly catching the exception. As it can be seen in the log the `SecurityException` is intercepted and rethrown as `PerformException`, but still we can verify the `SecurityException` via `PerformException` figuring out the `cause` of exception. e.g.:

```kt
@Test
fun testCase_exitApp() {
  val performException = assertThrows<PerformException>() {
    // test logic
  }
  
  // assert that performException.cause is instance of SecurityException 
  // and has message System.exit() here
}
```

PTAL and share your thoughts.

@adhiamboperes — It would be helpful to get your thoughts on this. This isn’t a common verification pattern in the codebase, but then again neither is explicitly triggering `exitProcess(0)`. I did notice a few similar instances of verifying messages in `ExceptionsControllerTest`, so it would be helpful to confirm this in meantime, so everything is aligned well to wrap this up quickly.

theayushyadav11:
Thanks @Rd4dev for clarifying the exact situation, I also think that this can be verified via the method you suggested. I'll update the test and send it for review asap.

app/src/main/java/org/oppia/android/app/devoptions/PendingChangesDialogFragmentPresenter.kt
theayushyadav11: Yes, updated

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
theayushyadav11: updated

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: Thanks for the suggestion, Updated.

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
theayushyadav11: updated

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
theayushyadav11: Updated as per latest suggestion

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
theayushyadav11: Updated as per latest suggestion

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
theayushyadav11: Updated

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
theayushyadav11: added

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
theayushyadav11: added the test

app/src/sharedTest/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentTest.kt
theayushyadav11: updated.

theayushyadav11:
@Rd4dev, I have incorporated the changes in the tests PTAL.
Thanks

app/BUILD.bazel
```diff
@@ -84,6 +84,7 @@ LISTENERS = [
     "src/main/java/org/oppia/android/app/devoptions/AppRestartListener.kt",
     "src/main/java/org/oppia/android/app/devoptions/ForceCrashButtonClickListener.kt",
     "src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersButtonClickListener.kt",
+    "src/main/java/org/oppia/android/app/devoptions/PendingChangesListener.kt",
```
adhiamboperes: Maybe?
```suggestion
    "src/main/java/org/oppia/android/app/devoptions/SavePendingChangesDialogListener.kt",
```

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
```diff
@@ -97,20 +88,23 @@ class PlatformParametersFragmentPresenter @Inject constructor(
...
+        val overriddenParameters = computeOverriddenParameters()
+        savePendingPlatformParameters(overriddenParameters)
+      }
+      this.platformParametersToolbar.setNavigationOnClickListener {
+        onBackNavigation()
+      }
+      this.platformParametersRecyclerView.apply {
+        layoutManager = linearLayoutManager
+        adapter = bindingAdapter
+      }
```
adhiamboperes: The `this` keyword is not needed here, for any of the elements inside the `apply` block.

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
```diff
@@ -85,20 +76,23 @@ class FeatureFlagsFragmentPresenter @Inject constructor(
...
+        val overriddenFlags = computeOverriddenFlags()
+        savePendingFeatureFlags(overriddenFlags)
+      }
+      this.featureFlagsToolbar.setNavigationOnClickListener {
+        onBackNavigation()
+      }
+      this.featureFlagsRecyclerView.apply {
+        layoutManager = linearLayoutManager
+        adapter = bindingAdapter
+      }
```
adhiamboperes: The `this` keyword is not needed here, for any of the elements inside the `apply` block.

adhiamboperes:
> > Hi @Rd4dev @adhiamboperes , regarding the modified tests, those tests were failing with the message that "cannot perform click on the button" and so to verify the changes after saving it to database and closing the scenerio works as database updation is done when the restart button is shown. This happens as roboelectric is not able to perform click on the button.
> 
> Thanks, @theayushyadav11, for sharing this. I don’t think the issue is with `perform(click())`, and I believe closing the scenario isn’t accurately what’s helping here for our case. I ran the test locally, and the logs further down show the real problem: it’s still failing with a `SecurityException`. The stack trace confirms the click action succeeds — it reaches the call to `System.exit()`. However, that exception is intercepted and wrapped inside a `PerformException`, because now the logic to restart triggers immediately on `perform(click())` rather earlier via `onDestroy`.
> 
> ```kotlin
> There was 1 failure:
> 1) testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp(org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest)
> java.lang.AssertionError: Expected exception of type: class java.lang.SecurityException, not: class androidx.test.espresso.PerformException.
>         at org.oppia.android.app.devoptions.featureflags.FeatureFlagsFragmentTest.testFeatureFlagsFragment_pressBack_saveChanges_clickRestartInDialog_exitsApp(FeatureFlagsFragmentTest.kt:1670)
>         ... 18 trimmed
> Caused by: androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 647, 345 and precision: 16, 16' on view 'an instance of android.widget.TextView and view.getText() equals string from resource id: <2131820652> [app_restart_dialog_button_text] value: Restart'.
>         at androidx.test.espresso.PerformException$Builder.build(PerformException.java:1)
>         at androidx.test.espresso.base.PerformExceptionHandler.handleSafely(PerformExceptionHandler.java:8)
> 
>         ... 19 more
> Caused by: java.lang.SecurityException: Test code should never call System.exit()
>         at com.google.testing.junit.runner.util.GoogleTestSecurityManager.checkExit(GoogleTestSecurityManager.java:34)
>         at java.base/java.lang.Runtime.exit(Runtime.java:114)
> ```
> 
> The test is failing because the `assertThrows<SecurityException>` block is not correctly catching the exception. As it can be seen in the log the `SecurityException` is intercepted and rethrown as `PerformException`, but still we can verify the `SecurityException` via `PerformException` figuring out the `cause` of exception. e.g.:
> 

@Rd4dev, @theayushyadav11,   the `SecurityException` is thrown to prevent test JVM shutdown. I think the way that you have asserted that  `System.exit()` is called is enough verification that that part of the production code is executed. I am quite fine with that, but adding comments to the test explaining the situation explicitly would be best.

adhiamboperes (approved):
@theayushyadav11, the PR mostly looks good to me. I am approving this PR, merge pendng on addressing my latest comments and the test comments. 

Either @Rd4dev or @theayushyadav11, please request @seanlip  or @BenHenning to verify and resolve my comments, and merge the PR.

Please ensure that the PR is in a merge ready state - checks passing and up to date with develop.

Rd4dev (approved):
Thanks, @theayushyadav11. I reviewed the latest changes and they LGTM. Please follow Adhiambo’s last two suggestions to proceed with proposing this PR for merge.

app/BUILD.bazel
theayushyadav11: updated

app/src/main/java/org/oppia/android/app/devoptions/platformparameters/PlatformParametersFragmentPresenter.kt
theayushyadav11: updated

app/src/main/java/org/oppia/android/app/devoptions/featureflags/FeatureFlagsFragmentPresenter.kt
theayushyadav11: updated

theayushyadav11:
Thanks @adhiamboperes @Rd4dev for reviewing and the suggestions. I have applied all the suggestions.
@seanlip or @BenHenning can you please merge this PR as it is now approved.

app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragmentPresenter.kt
adhiamboperes: Oh I planned to delete the comment but forgot O_o. Sorry!

PR #5955, opened 2025-09-07
Fix #5906: Fix app language behavior when a non-admin profile is logged out

### LINKED ISSUE #5906: [BUG]: Unexpected app language behavior when a non-admin profile is logged out
### Describe the bug

After logging out of a non admin profile with a specific profile language(e.g. Portuguese), the language persists  in the logged out state when the admin tries to add another profile.

### Steps To Reproduce

1. Set the app-wide language to English
2. Log into a non-admin profile
3. Change language to Portuguese
4. Go back to profile selection screen
5. Log into a new profile for the first time (for the onboarding flow) or create a profile
6. Observe all UIs are in Portuguese instead of English until the profile is fully logged in

### Expected Behavior

The admin should see the screens in the default system language. (Confirm with product)

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

This is caused by there being no logged in profile, so the translation system fetches the latest result of app language selection. My suggestion would be to make all these screens  extend the `SystemLocalizedAppCompatActivity`, so that they are always displayed in the system language, which is the most likely the admin's preffered language.

theMr17:
## Explanation

Fixes #5906

This PR ensures that all screens shown when no profile is logged in are displayed using the system-wide language selected in the Android settings.

Additionally, it addresses the issue (mentioned in https://github.com/oppia/oppia-android/issues/5906#issuecomment-3224409093) where accessing the Admin Controls screen from Profile Selection (after entering the admin PIN) did not properly reflect the admin’s chosen language. With this fix, once the admin profile is logged in via the PIN, navigating to Home (and other screens) from Admin Controls will correctly display content in the language set for that profile.

Note: `AdminPinActivity` still extends `InjectableAutoLocalizedAppCompatActivity` since it is never launched after logging out of another profile. Instead, it is only triggered in a specific case: when the admin attempts to add the first non-admin profile and the admin profile does not yet have a PIN set.

Per https://github.com/oppia/oppia-android/pull/5955#discussion_r2379391490, refactored the `AdminAuthActivity` to use profileId decorator instead of passing the `profileId` via `AdminAuthActivityParams`.

### Before
[Recording.webm](https://github.com/user-attachments/assets/bf66739e-9ca0-495a-8894-5f70cd70fcf8)

### After
[Recording.webm](https://github.com/user-attachments/assets/130edd3d-0a58-453c-9822-352c5cffe077)

## Essential Checklist


theMr17:
Hey @adhiamboperes, this PR is now ready for review, PTAL.

Do I also need to check the onboarding screens for this bug, or have those already been verified as a part of the onboarding project?

app/src/main/java/org/oppia/android/app/profile/AdminAuthActivityPresenter.kt
```diff
@@ -86,13 +90,16 @@ class AdminAuthActivityPresenter @Inject constructor(
...
+                activity.startActivity(
+                  AdministratorControlsActivity.createAdministratorControlsActivityIntent(
+                    context, profileId
+                  )
+                )
+                activity.finish()
+              }
+            }
           }
           AdminAuthEnum.PROFILE_ADD_PROFILE.value -> {
```
adhiamboperes, line 103: In this scenario, the admin can also be logged in to see the screen in their preferred language -- I believe this is shown after the admin authentication flow.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivity.kt
```diff
@@ -13,7 +13,7 @@ import org.oppia.android.util.profile.CurrentUserProfileIdIntentDecorator.extrac
 import javax.inject.Inject
 
 /** Activity for displaying a new learner profile creation flow. */
-class CreateProfileActivity : InjectableAutoLocalizedAppCompatActivity() {
```
adhiamboperes, line 16: This is one of the onboarding screens in the v2 onboarding flow, where the user creates a nickname for their profile. If this gets system localized, then should all the other onboarding screens be too? It would be nice to do the following:
- enable the onboarding v2 feature flag
- set your device to arabic or portuguese system language
- delete app data to trigger the onboarding flow
- record the system language behavior.

This can be compared with the existing behaviour of the v1 onboarding flow.

adhiamboperes (changes requested):
Thanks @theMr17! Left some thoughts, PTAL.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivity.kt
theMr17: Updated `IntroActivity` and `OnboardingActivity` to be system localized as well.

Note: This change does not affect the onboarding screen's behavior since those screens are accessed when the `profileId` is `null`, which already defaults to the system language. The update simply ensures consistency across all onboarding-related screens.

app/src/main/java/org/oppia/android/app/profile/AdminAuthActivityPresenter.kt
theMr17: Logged in the admin profile for `PROFILE_ADD_PROFILE` case as well.

theMr17:
Hey @adhiamboperes, thanks for the review! Addressed your comments, PTAL.

app/src/main/java/org/oppia/android/app/profile/AdminAuthActivityPresenter.kt
```diff
@@ -102,12 +102,18 @@ class AdminAuthActivityPresenter @Inject constructor(
             }
           }
           AdminAuthEnum.PROFILE_ADD_PROFILE.value -> {
-            activity.startActivity(
-              AddProfileActivity.createAddProfileActivityIntent(
-                context, args?.colorRgb ?: -10710042
-              )
-            )
-            activity.finish()
+            val internalId = args?.internalProfileId ?: -1
```
adhiamboperes: We have been using the profileId decorator for passing and retrieving the profileId from intents, so this looks a little off. Could you please check and possibly do a refactor?

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivity.kt
adhiamboperes: @theMr17, do you mean that changing the device language does not impact the onboarding screens? I wonder if the screens have been translated. I need to check this

adhiamboperes (changes requested):
Thanks @theMr17!. I have one follow up. PTAL.

app/src/main/java/org/oppia/android/app/onboarding/CreateProfileActivity.kt
theMr17: @adhiamboperes Apologies for the confusion. What I meant is that whether we make the onboarding screens system-localized or auto-localized, the language will still follow the system's default language.

app/src/main/java/org/oppia/android/app/profile/AdminAuthActivityPresenter.kt
theMr17: Refactored this with the decorator.

theMr17:
Addressed your comments, PTAL @adhiamboperes.

adhiamboperes (approved):
Thanks for the changes @theMr17!

Also, apologies for the delay in review.

PR #5958, opened 2025-09-09
Fixes part of #5345: Fixes restarting of download on config change.

theayushyadav11:
Fixes part of #5345:
This pull request updates the logic for downloading remote platform parameters in `PlatformParameterControllerDebugImpl`. The main change is that logging for the download process now only occurs when a new download is actually initiated, and not every time the method is called.

Improvements to download logic and logging:

* In `PlatformParameterControllerDebugImpl.kt`, changed the placement of logging statements so that "Calling Force Download of remote parameters" and "Download finished successfully." are logged only when a new download task is started, preventing duplicate logs when a download is already in progress.
Added a check :
```kotlin
   if (ongoingDownloadTask == null || ongoingDownloadTask?.isCancelled == true) {
```
To simulate actual download behaviour and verify that on config change download doesn't restarts please repalace:
```
 // TODO(#5950): Replace with actual logic to force-download remote parameters.
```
with

```kotlin
delay(3000)
```
#### Evaluation Points:

1. Rotating the device while the downloads dialog is open and actively downloading does not causes it to restart the call from scratch. [Video](https://github.com/user-attachments/assets/40aa156d-240f-4ae7-85c9-8538db102700)
2. Rotating the device when the ‘successfully downloaded’ dialog is open does not causes the download to start again.
[Video](https://github.com/user-attachments/assets/728c30b2-16c8-476b-9063-7571b63f96c4)

</div></b>

##### Before

[Screencast from 2025-09-09 14-46-31.webm](https://github.com/user-attachments/assets/40aa156d-240f-4ae7-85c9-8538db102700)

##### After

[Screencast from 2025-09-09 09-42-16.webm](https://github.com/user-attachments/assets/95c4dd7a-6098-453f-a94e-6a4ebdb4afe8)

## Essential Checklist

### DIFF
--- a/app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
@@ -110,7 +110,6 @@ import org.oppia.android.util.parser.image.GlideImageLoaderModule
 import org.oppia.android.util.parser.image.ImageParsingModule
 import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
-import org.robolectric.shadows.ShadowLog
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -626,26 +625,6 @@ class DeveloperOptionsFragmentTest {
     val cause = performException.cause
     assertThat(cause).isInstanceOf(SecurityException::class.java)
     assertThat(cause?.message).contains("System.exit()")
-  }
-
-  @Test
-  fun testDeveloperOptionsFragment_clickDownload_configChange_downloadDoesnotRestarts() {
-    launch<DeveloperOptionsTestActivity>(
-      createDeveloperOptionsTestActivityIntent(internalProfileId)
-    ).use {
-      testCoroutineDispatchers.runCurrent()
-
-      scrollToPosition(position = 2)
-      onView(withId(R.id.force_download_button)).perform(click())
-      testCoroutineDispatchers.runCurrent()
-      onView(isRoot()).perform(orientationLandscape())
-
-      val logs = ShadowLog.getLogs()
-      val platformParamLogs = logs.filter { it.tag == "PlatformParameterController" }
-      assertThat(
-        platformParamLogs.count { it.msg.contains("Calling Force Download of remote parameters") }
-      ).isEqualTo(1)
-    }
   }
 
   private fun createDeveloperOptionsTestActivityIntent(internalProfileId: Int): Intent {
--- a/app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
@@ -110,6 +110,7 @@ import org.oppia.android.util.parser.image.GlideImageLoaderModule
 import org.oppia.android.util.parser.image.ImageParsingModule
 import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
+import org.robolectric.shadows.ShadowLog
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -625,6 +626,26 @@ class DeveloperOptionsFragmentTest {
     val cause = performException.cause
     assertThat(cause).isInstanceOf(SecurityException::class.java)
     assertThat(cause?.message).contains("System.exit()")
+  }
+
+  @Test
+  fun testDeveloperOptionsFragment_clickDownload_configChange_downloadDoesnotRestarts() {
+    launch<DeveloperOptionsTestActivity>(
+      createDeveloperOptionsTestActivityIntent(internalProfileId)
+    ).use {
+      testCoroutineDispatchers.runCurrent()
+
+      scrollToPosition(position = 2)
+      onView(withId(R.id.force_download_button)).perform(click())
+      testCoroutineDispatchers.runCurrent()
+      onView(isRoot()).perform(orientationLandscape())
+
+      val logs = ShadowLog.getLogs()
+      val platformParamLogs = logs.filter { it.tag == "PlatformParameterController" }
+      assertThat(
+        platformParamLogs.count { it.msg.contains("Calling Force Download of remote parameters") }
+      ).isEqualTo(1)
+    }
   }
 
   private fun createDeveloperOptionsTestActivityIntent(internalProfileId: Int): Intent {
--- a/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
+++ b/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
@@ -74,14 +74,16 @@ class PlatformParameterControllerDebugImpl @Inject constructor(
   }
 
   override fun downloadRemoteParameters(): DataProvider<Unit> {
-    oppiaLogger.d("PlatformParameterController", "Calling Force Download of remote parameters")
     return dataProviders.createInMemoryDataProviderAsync(DOWNLOAD_REMOTE_PARAMETERS_PROVIDER_ID) {
-      ongoingDownloadTask?.cancel()
-      ongoingDownloadTask = CoroutineScope(backgroundCoroutineDispatcher).async {
-        // TODO(#5950): Replace with actual logic to force-download remote parameters.
+      if (ongoingDownloadTask == null || ongoingDownloadTask?.isCancelled == true) {
+        oppiaLogger.d("PlatformParameterController", "Calling Force Download of remote parameters")
+        ongoingDownloadTask?.cancel()
+        ongoingDownloadTask = CoroutineScope(backgroundCoroutineDispatcher).async {
+          // TODO(#5950): Replace with actual logic to force-download remote parameters.
+          oppiaLogger.d("PlatformParameterController", "Download finished successfully.")
+        }
       }
       ongoingDownloadTask?.await()
-      oppiaLogger.d("PlatformParameterController", "Download finished successfully.")
       return@createInMemoryDataProviderAsync AsyncResult.Success(Unit)
     }
   }
--- a/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
+++ b/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
@@ -74,9 +74,7 @@ class PlatformParameterControllerDebugImpl @Inject constructor(
     }
   }
 
-  override fun downloadRemoteParameters(): DataProvider<Unit> {
-    return downloadRemoteParametersProvider
-  }
+  override fun downloadRemoteParameters() = downloadRemoteParametersProvider
 
   /** Cancels any ongoing remote parameter download. */
   fun cancelRemoteParameterDownload(): Boolean {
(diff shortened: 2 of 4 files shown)


theayushyadav11:
Hi @Rd4dev PTAL

BenHenning:
Please also add tests that you can verify fail without the fix @theayushyadav11.

theayushyadav11:
> Please also add tests that you can verify fail without the fix @theayushyadav11.

Hi @BenHenning, should I add a unit test for this or a video demonstration that download was restarting before the fix.
If unit test is to be added then we can verify fix only not before the fix, please clarify if I'm missing something.

seanlip:
@theayushyadav11 Add unit tests for the fix, and demonstrate (e.g. through video) that those same tests fail if the fix is not applied.

app/src/sharedTest/java/org/oppia/android/app/devoptions/DeveloperOptionsFragmentTest.kt
```diff
@@ -627,6 +628,26 @@ class DeveloperOptionsFragmentTest {
...
+
+      scrollToPosition(position = 2)
+      onView(withId(R.id.force_download_button)).perform(click())
+      testCoroutineDispatchers.runCurrent()
+      onView(isRoot()).perform(orientationLandscape())
+
+      val logs = ShadowLog.getLogs()
+      val platformParamLogs = logs.filter { it.tag == "PlatformParameterController" }
+      assertThat(
+        platformParamLogs.count { it.msg.contains("Calling Force Download of remote parameters") }
```
theayushyadav11: Hi @Rd4dev @adhiamboperes , this test passes here but when fix is not applied then running this test the log for restarting download is not shown in the tests but while manually verifying it shows there.
While manually verifying:
1. Click download.
2. Rotate screen.
3. Wait for some time.

The logs are :
```
2025-09-09 14:24:22.060  3147-3147  PlatformPa...Controller org.oppia.android                    D  Calling Force Download of remote parameters
2025-09-09 14:24:23.254  3147-3147  PlatformPa...Controller org.oppia.android                    D  Calling Force Download of remote parameters
2025-09-09 14:24:26.285  3147-3267  PlatformPa...Controller org.oppia.android                    D  Download finished successfully.
```
But this test does not give in the unit tests, where logs are there only for once.

Rd4dev:
Hi @theayushyadav11, the root cause I think here is that the current implementation creates a new `DataProvider` on every call, which triggers a fresh download each time. 

https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt#L76-L87

By reusing the same provider, the download can persist across orientation changes, which should solve this issue. You should be able to verify this by validating the providers where the test fails with the old implementation and passes with the updated one.

theayushyadav11:
Hi @BenHenning here is the video  test failing on the current develop branch.
[Screencast from 2025-09-09 20-11-43.webm](https://github.com/user-attachments/assets/4869d845-4d69-4e60-ab54-573e339a01d5)

theayushyadav11:
> Hi @theayushyadav11, the root cause I think here is that the current implementation creates a new `DataProvider` on every call, which triggers a fresh download each time.
> 
> https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt#L76-L87
> 
> By reusing the same provider, the download can persist across orientation changes, which should solve this issue. You should be able to verify this by validating the providers where the test fails with the old implementation and passes with the updated one.

Thanks @Rd4dev for the suggestion, i have updated the tests PTAL.

theayushyadav11:
PTAL @adhiamboperes .

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
```diff
@@ -1138,6 +1138,19 @@ class PlatformParameterControllerDebugImplTest {
     assertThat(ephemeralNewUiFlag?.currentValue)
       .isEqualTo(TEST_REMOTE_MULTIPLE_CLASSROOMS)
   }
+
+  @Test
+  fun testDownloadRemoteParameters_withMultipleProviders_isOfSameInstance() {
+    setUpTestApplicationComponent()
+    testCoroutineDispatchers.runCurrent()
```
adhiamboperes: Why is this needed?

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
```diff
@@ -258,6 +250,21 @@ class PlatformParameterControllerDebugImpl @Inject constructor(
     }
   }
 
+  private fun downloadRemoteParametersInternal(): DataProvider<Unit> {
+    return dataProviders.createInMemoryDataProviderAsync(DOWNLOAD_REMOTE_PARAMETERS_PROVIDER_ID) {
+      if (ongoingDownloadTask == null || ongoingDownloadTask?.isCancelled == true) {
```
Rd4dev: Does it need the additional `isCancelled` check? Cancellations are separately handled in `cancelRemoteParameterDownload`, which sets `ongoingDownloadTask` to `null`. In this implementation, `null` effectively covers both the “never started” and “was cancelled” cases.
```suggestion
      if (ongoingDownloadTask == null) {
```

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
```diff
@@ -258,6 +250,21 @@ class PlatformParameterControllerDebugImpl @Inject constructor(
...
+        oppiaLogger.d("PlatformParameterController", "Calling Force Download of remote parameters")
+        ongoingDownloadTask?.cancel()
```
Rd4dev: Do we need to call `cancel()` here? Since we only initiate a new download when `ongoingDownloadTask` is null.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
```diff
@@ -1138,6 +1138,19 @@ class PlatformParameterControllerDebugImplTest {
     assertThat(ephemeralNewUiFlag?.currentValue)
       .isEqualTo(TEST_REMOTE_MULTIPLE_CLASSROOMS)
   }
+
+  @Test
+  fun testDownloadRemoteParameters_withMultipleProviders_isOfSameInstance() {
```
Rd4dev: To focus the naming on the behavior than on the implementation.
```suggestion
  fun testDownloadRemoteParameters_secondInvocation_returnsSameInstance() {
```
>Ref: https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/data/src/test/java/org/oppia/android/data/backends/gae/RetrofitServiceModuleTest.kt#L59

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
```diff
@@ -1138,6 +1138,19 @@ class PlatformParameterControllerDebugImplTest {
...
+    setUpTestApplicationComponent()
+    testCoroutineDispatchers.runCurrent()
+
+    val firstProvider = platformParameterControllerDebugImpl.downloadRemoteParameters()
+    val secondProvider = platformParameterControllerDebugImpl.downloadRemoteParameters()
+    testCoroutineDispatchers.runCurrent()
+
+    assertThat(secondProvider).isSameInstanceAs(firstProvider)
```
Rd4dev: Please include a comment to describe this assertion.

Ref:
https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/data/src/test/java/org/oppia/android/data/backends/gae/RetrofitServiceModuleTest.kt#L58-L65

Rd4dev: I can see instances of `isEqualTo` being used in the codebase. While I think `isSameInstance` is preferable here, it’s better to clarify if there’s a preferred style to ensure consistency. Please let the `codeowner` decide on this and help resolve this as well.

Ref:
https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/data/src/test/java/org/oppia/android/data/backends/gae/RetrofitServiceModuleTest.kt#L49-L55

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
```diff
@@ -1138,6 +1138,19 @@ class PlatformParameterControllerDebugImplTest {
     assertThat(ephemeralNewUiFlag?.currentValue)
       .isEqualTo(TEST_REMOTE_MULTIPLE_CLASSROOMS)
   }
+
+  @Test
```
Rd4dev, line 1142: Please include the same verification for `ForceDownloadRemoteParametersController` as well.

Rd4dev (changes requested):
@theayushyadav11 -- Took a pass. Considering UI testing, please confirm with Adhiambo if anything additional is required, as I am unsure of testing this scenario directly with state components with `DialogFragment`, since I find it not so reliable but please confirm and proceed with the guidance. 

Also, make sure to include both evaluation points with their descriptions and screen recordings for both cases in the PR description, so verification is straightforward and nothing is missed. Additionally include this PR to the issue tracker milestone table as well.

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
theayushyadav11: updated

domain/src/main/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImpl.kt
theayushyadav11: yeah, we don't need to call these , removed this

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
theayushyadav11: updated

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
theayushyadav11: added

theayushyadav11: I also think `isSameInstanceAs` seems good.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
theayushyadav11: Added

theayushyadav11:
Hi @Rd4dev PTAL.

domain/src/test/java/org/oppia/android/domain/platformparameter/PlatformParameterControllerDebugImplTest.kt
Rd4dev: Resolving this to keep the PR unblocked for merging, since I don’t have concerns with `isSameInstanceAs` at the moment. However, please add a link to this comment so the `codeowner` can review it before giving final confirmation. Thanks!

Rd4dev: @theayushyadav11 -- PTAL

Rd4dev (approved):
@theayushyadav11 -- The domain changes LGTM. 

>Also, make sure to include both evaluation points with their descriptions and screen recordings for both cases in the PR description, so verification is straightforward and nothing is missed. 

Other than that, the PR description doesn’t seem updated with these — PTAL.

theayushyadav11:
>  Also, make sure to include both evaluation points with their descriptions and screen recordings for both cases in the PR description, so verification is straightforward and nothing is missed.
> 

Added

theayushyadav11:
Hi @adhiamboperes, please review this PR so that it can be merged asap.

adhiamboperes (approved):
LGTM, thanks @theayushyadav11

PR #5968, opened 2025-09-16
Fix #482: Remove default admin for onboarding V2 (3/12)

### LINKED ISSUE #482: Remove the hardcoding of the admin profile in the profile initialization stage.
Currently an admin is hard coded into the application. This needs to be removed when there is a proper admin account creation.

adhiamboperes:
## Explanation

Fix #482

In the legacy onboarding flow, the admin profile is hardcoded into the profile chooser screen. In **V2 of the onboarding flow**, an admin user creates their profile during onboarding. This makes the hardcoded profile redundant, and this PR removes it.

### Main Changes  
1. **Gating `addProfile()`**  
   - `profileManagementController.addProfile()` is now gated behind the `enableOnboardingFlowV2` flag.  
   - This ensures it is only called in the **legacy flow**.  

2. **Cleanup of `profileNickname`**  
   - The `profileNickname` argument (introduced in a previous PR to ensure the created profile had the correct display name) is no longer needed.  
   - This PR removes its usages and updates the relevant tests.  

3. **Fix for App Data Reset Edge Case**  
   - Previously, deleting all profiles during an app data reset left the app in a **bad state**: the app was considered onboarded, but no profiles existed.  
   - To address this, a new function `resetOnboardingState` has been added in `AppStartupStateController.kt` to reset the onboarding state.  

### Considered Approaches  
There were multiple possible solutions for handling the reset logic:

- **(a)  UI-Layer Reset**  
  - Update `alreadyOnboardedApp` to `false` and call the function in the UI layer.  
  -  Minimal impact, no refactor required.  
  - Requires keeping the *delete profiles DataProvider* and *app startup state DataProvider* in sync.  
  - Harder to test the reset flow in `ProfileLoginFragmentTest.kt`.

- **(b) Domain-Layer Reset**  
  - Update `alreadyOnboardedApp` to `false` inside `profileManagementController.deleteAllProfiles()`.  
  - Ensures the two providers are always in sync.  
  - Violates single responsibility principle of `deleteAllProfiles()`.

- **(c) Refactor to Remove `alreadyOnboardedApp`**  
  - Remove the field and instead rely on whether the **profiles list is empty**.  
  - Requires significant refactors across multiple areas, including handling both states of the `enableAppAndOsDeprecation` feature flag.  
  - Startup logic would need to move away from `onboardingFlowStore`, adding complexity.  

Option **(a)** was chosen because it:  
- Requires the least amount of change.  
- Minimizes side effects.  
- Keeps the implementation cleaner, even though testing and DataProvider synchronization require extra care.  

## Essential Checklist

## For UI-specific PRs only
| Before | After |
|---|---|
| https://github.com/user-attachments/assets/a2f635e9-4078-49c3-b157-374f9f6b8ba9 | https://github.com/user-attachments/assets/cce07588-912e-4b44-85b5-6007381106bc|


adhiamboperes:
@BenHenning, PTAL. This is probably the smallest PR in the entire project 😁.

model/src/main/proto/arguments.proto
```diff
@@ -962,10 +962,6 @@ message OnboardingFragmentStateBundle {
 
 // Params required when creating a new ProfileChooserActivity.
 message ProfileChooserActivityParams {
-  // TODO(#482): Remove once the default admin profile has been removed.
-  // The nickname associated with a newly created profile.
-  string profile_nickname = 1;
```
BenHenning, line 967: This should be marked as `reserved` probably, just for posterity (since the interoperability risk is low).

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -421,8 +423,17 @@ class ProfileLoginFragmentPresenter @Inject constructor(
       .showNow(activity.supportFragmentManager, TAG_VALIDATE_ADMIN_PIN_DIALOG)
   }
 
+  /**
+   * Deletes all app data by removing all profile,s resetting onboarding state, and closing the app.
```
BenHenning:
```suggestion
   * Deletes all app data by removing all profile's resetting onboarding state, and closing the app.
```

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
```diff
@@ -421,8 +423,17 @@ class ProfileLoginFragmentPresenter @Inject constructor(
...
+   *
+   * Once completed, the user will be forced to go through onboarding again to re-create their
+   * profile.
+   */
   private fun deleteAppData() {
     profileManagementController.deleteAllProfiles().toLiveData().observe(fragment) {
+      // Reset onboarding so the app starts fresh.
+      appStartupStateController.resetOnboardingState()
```
BenHenning, line 435: Perhaps worth verifying via test that this step happens?

BenHenning (changes requested):
Thanks @adhiamboperes! Just a few comments--PTAL.

BenHenning:
Reminder check the checkboxes on the PR description as well.

Other than that, super cool to see an issue from 2019 get resolved! :D

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: Fixed.

model/src/main/proto/arguments.proto
adhiamboperes: This does not need to be reserved because it has never actually been used in logs. I introduced it in one of the earlier PRs for this project, so there is no risk to removing it completely. I noticed that I should have re-numbered the fields however, which I have now done.

app/src/main/java/org/oppia/android/app/profile/ProfileLoginFragmentPresenter.kt
adhiamboperes: Added, and here is proof that the test fails if the wrong `StartupMode` is asserted.

Screenshot 2025-10-02 at 22 11 10

adhiamboperes:
@BenHenning, PTAL.

BenHenning (approved):
Thanks @adhiamboperes! This LGTM.

adhiamboperes:
Resolved conflicts, changes LGTM.

Issue #5442, opened 2024-06-24
[BUG]: Play Console warns about unspecified mutability in PendingIntents


BenHenning:
### Describe the bug

When pushing the latest release to Play Console, it provided warnings about:

```
Apps targeting Android 12 or later must specify the mutability of PendingIntents with either FLAG_IMMUTABLE or FLAG_MUTABLE. Having unspecified mutability may cause your app to crash.
```

And it links to https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability. We don't use ``PendingIntent``s directly, so this is related to dependencies. The ones detected by Play Console are (obfuscated for the latest release):
- androidx.work.impl.background.systemalarm.a.a
- bs.dL.u
- bn.dL.u
- com.google.android.gms.measurement.internal.et.u

Unfortunately, none of these seem to be part of our codebase (at least from what I could see in our proguard.map file, so they're probably coming from GMS if I was to guess (I can't be positive).

### Steps To Reproduce

N/A, though I think there might be a way to check for this using Android lint (needs more research) from looking briefly at https://stackoverflow.com/q/67045607/3689782.

### Expected Behavior

We should avoid this warning by ensuring our app is Android 31 compliant.

### What device/emulator are you using?

N/A

### Which Android version is your device/emulator running?

All

### Which version of the Oppia Android app are you using?

0.13-beta-dcde7b359e

### Additional Context

Note that Play Console warns about potential crashes, but I'm fairly certain the app never hits these scenarios. However, more testing will be needed on a newer version of Android just to be sure.

This probably ultimately requires isolating and then updating the corresponding dependencies. Updating dependencies is a difficult problem for us without more significant e2e testing, and it almost certainly will cause compatibility issues with Gradle.

adhiamboperes:
Linking this comment from @seanlip: https://github.com/oppia/oppia-android/issues/4751#issuecomment-1685210663, the app is definitely crashing.

From looking around the web, this is happening on s+ devices because of our workmanager version. See note: https://developer.android.com/jetpack/androidx/releases/work#version_27_2.

adhiamboperes:
Reported crashes logged at:

https://github.com/oppia/oppia-android/issues/5452

adhiamboperes:
@whyash8, PTAL.

whyash8:
@adhiamboperes Can you please provide more context for the issue and what will be best approach to understand and solve it . 
and also tell me how to reproudce this issue . 
thanks

[...]

Issue #5452, opened 2024-06-27
[BUG]: java.lang.IllegalArgumentException - org.oppia.android: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.


seanlip:
### Describe the bug

We are seeing this crash in Crashlytics for both 0.12 and 0.13. It happens at org.oppia.android.app.settings.profile.m.a.

This issue has 4 crash events affecting multiple users in the span of 1 day. 

<details>
  <summary>Full stacktrace</summary>

```

Fatal Exception: java.lang.IllegalArgumentException: org.oppia.android: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
       at android.app.PendingIntent.checkFlags(PendingIntent.java:408)
       at android.app.PendingIntent.getBroadcastAsUser(PendingIntent.java:688)
       at android.app.PendingIntent.getBroadcast(PendingIntent.java:675)
       at androidx.work.impl.utils.ForceStopRunnable.a(SourceFile:196)
       at androidx.work.impl.utils.ForceStopRunnable.run(SourceFile:3128)
       at androidx.work.impl.utils.i.run(SourceFile:91)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
       at java.lang.Thread.run(Thread.java:1012)

Crashlytics Exception Handler1:
       at dalvik.system.VMStack.getThreadStackTrace(VMStack.java)
       at java.lang.Thread.getStackTrace(Thread.java:1841)
       at java.lang.Thread.getAllStackTraces(Thread.java:1909)
       at cf.ae.a(SourceFile:1260)
       at cf.av.a(SourceFile:218)
       at cf.av.a(SourceFile:122)
       at cf.H.call(SourceFile:1395)
       at cf.m.a(SourceFile:106)
       at bt.r.run(SourceFile:1)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[...]

PR #6002, opened 2025-11-06
Fix #4751: Update work manager to 2.7.0 and enable it on S+ devices

### LINKED ISSUE #4751: WorkManager needs to be updated to at least 2.7.0 to start working on SDK 31
The investigation into #4750 revealed that our current version of WorkManager (2.4.0) doesn't work on SDK 31 due to changes in PendingIntent policy. 2.7.0 does work, however it requires updating to Kotlin 1.5.0 SDKs which affect the app more broadly than simply updating the work manager dependencies. Also, for reasons I'm not fully sure of, AndroidX work started depending on Kotlin coroutines jvm (instead of just Android), which leads to an issue with Android desugaring (since it depends on sun.misc classes that don't exist in the Android world). I have fixes for these, but I'll probably wait to introduce them until migrating to Kotlin 1.6 (or even 1.7).

theMr17:
## Explanation

Fixes #4751

This PR updates `WorkManager` from version 2.4.0 to 2.7.0.

The upgrade ensures compatibility with Android 12 (API level 31) and above, where background work scheduling requires newer `WorkManager` APIs. Additionally, since the project already uses Kotlin 1.6, the version bump aligns well with the current dependency ecosystem and does not introduce any major breaking changes.

The `androidx.work.impl.WorkManagerInitializer` entry was removed from the main manifest because `WorkManager` ≥ 2.6.0 no longer includes that class (initialization moved to AndroidX Startup). The `InitializationProvider` handling was added in their respective test manifests instead of the main manifest because test modules have different application IDs, dependency classpaths, and merge behavior--so each test target needs a manifest overlay that matches its classpath and avoids manifest merge conflicts.

|Device|Logs|
|--|--|
|Older|image|
|S+|image|

## Essential Checklist


theMr17:
@adhiamboperes PTAL.

app/src/main/AndroidManifest.xml
```diff
@@ -364,11 +364,6 @@
     <activity
       android:name=".app.testing.TextInputLayoutBindingAdaptersTestActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
-    <provider
-      android:name="androidx.work.impl.WorkManagerInitializer"
-      android:authorities="${applicationId}.workmanager-init"
-      android:exported="false"
-      tools:node="remove" />
```
adhiamboperes, line 371: Why was this removed from the main manifest file, but added to all other files? -- This sounds important to document in the PR description.

adhiamboperes (commented):
Thanks for the meticulous work, @theMr17! I have one question inline.

Additionally, could you please do manual tests on an S+ device and an older device to verify that logs are stil being uploaded correctly? You can do this by:
1. Enabling analyticscollection via the `AndroidManifest`
2. Checking log cat logs to see whether Firebase upload is actually succeeding.

app/src/main/AndroidManifest.xml
theMr17: I have updated the PR description with this explanation.

theMr17:
Hey @adhiamboperes, I've addressed your comment and added screenshots confirming that logs are still being uploaded correctly.

I do have one more question, though. Since `WorkManager` no longer requires `WorkManagerInitializer` and instead relies on the AndroidX Startup library, the following manifest entry should technically be removable:

https://github.com/oppia/oppia-android/blob/c47a03918b65505926f04146d858a0b256336654/instrumentation/src/java/AndroidManifest.xml#L218-L222

I'm unsure how to properly verify that the instrumentation module continues to work as expected after removing it. Do you have any suggestions on how to validate this safely?

app/src/main/AndroidManifest.xml
adhiamboperes: Sg, thanks.

adhiamboperes:
> oppia-android/instrumentation/src/java/AndroidManifest.xml

@theMr17, I believe that you can attempt to run `instrumentation/src/javatests/org/oppia/android/instrumentation/player/ExplorationPlayerTest.kt` and see if it will build with either configuration. This is a test-only module.

adhiamboperes (commented):
Other than the open question, this PR LGTM.

BenHenning (approved):
Thanks @theMr17! This is fantastic, and looks like a great and clearly explained fix. Really happy to see this get addressed.

Do we know if this fixes #5442?

theMr17:
Thanks for the review @BenHenning!

Re: Its difficult to verify if #5442 (or even #5452) got fixed. As both are production issues and have unclear repro steps. We might need to rely on future Play Console logs to confirm if this PR fixes those as well (which ideally should be the case).

theMr17:
> I do have one more question, though. Since `WorkManager` no longer requires `WorkManagerInitializer` and instead relies on the AndroidX Startup library, the following manifest entry should technically be removable:

@adhiamboperes this is now done, PTAL.

PR #6005, opened 2025-11-13
Fix #5962: Ensure parameters are initialized for all activities

### LINKED ISSUE #5962: [BUG]: Issue with Platform Parameter Initialization on app startup
### Describe the bug

Sometimes when starting the developer version of the app, it crashes before leaving the splash screen.

### Steps To Reproduce

TODO: Update when more information is available

I think this mostly happens when the app is  backgrounded for a while(not explicitly closed), then it fails to resume correctly.

This also needs to be tested on the prod versions.

### Expected Behavior

The app should not crash on startup.

### Screenshots/Videos

```php
FATAL EXCEPTION: main
Process: org.oppia.android, PID: 2433
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.oppia.android/org.oppia.android.app.profile.ProfileChooserActivity}:
java.lang.IllegalStateException: Attempting to access feature flag LOGGING_LEARNER_STUDY_IDS before initialization.
If this is a test, is it using TestPlatformParameterModule?
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3479)
    at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3651)
    at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
    at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2104)
    at android.os.Looper.loop(Looper.java:236)
    at android.app.ActivityThread.main(ActivityThread.java:7861)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:600)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:967)
Caused by: java.lang.IllegalStateException: Attempting to access feature flag LOGGING_LEARNER_STUDY_IDS before initialization.
    at org.oppia.android.domain.platformparameter.PlatformParameterProcessState.retrieveFeatureFlagState(PlatformParameterProcessState.kt:141)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule$Companion$retrieveFeatureFlag$1.<init>(FeatureFlagBindingModule.kt:103)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule$Companion.retrieveFeatureFlag(FeatureFlagBindingModule.kt:102)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule$Companion.access$retrieveFeatureFlag(FeatureFlagBindingModule.kt:98)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule.provideLoggingLearnerStudyIds(FeatureFlagBindingModule.kt:49)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.provideLoggingLearnerStudyIds(FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.java:48)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.get(FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.java:37)
    at org.oppia.android.domain.platformparameter.FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.get(FeatureFlagBindingModule_ProvideLoggingLearnerStudyIdsFactory.java:12)
    at org.oppia.android.util.logging.EventBundleCreator_Factory.get(EventBundleCreator_Factory.java:40)
    at org.oppia.android.util.logging.EventBundleCreator_Factory.get(EventBundleCreator_Factory.java:12)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.util.logging.firebase.FirebaseAnalyticsEventLogger_Factory_Factory.get(FirebaseAnalyticsEventLogger_Factory_Factory.java:41)
    at org.oppia.android.util.logging.firebase.FirebaseAnalyticsEventLogger_Factory_Factory.get(FirebaseAnalyticsEventLogger_Factory_Factory.java:13)
    at org.oppia.android.util.logging.firebase.DebugAnalyticsEventLogger_Factory.get(DebugAnalyticsEventLogger_Factory.java:31)
    at org.oppia.android.util.logging.firebase.DebugAnalyticsEventLogger_Factory.get(DebugAnalyticsEventLogger_Factory.java:10)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.util.logging.firebase.DebugLogReportingModule_ProvideDebugEventLoggerFactory.get(DebugLogReportingModule_ProvideDebugEventLoggerFactory.java:36)
    at org.oppia.android.util.logging.firebase.DebugLogReportingModule_ProvideDebugEventLoggerFactory.get(DebugLogReportingModule_ProvideDebugEventLoggerFactory.java:12)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.domain.oppialogger.analytics.AnalyticsController_Factory.get(AnalyticsController_Factory.java:103)
    at org.oppia.android.domain.oppialogger.analytics.AnalyticsController_Factory.get(AnalyticsController_Factory.java:23)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.domain.oppialogger.analytics.LearnerAnalyticsLogger_Factory.get(LearnerAnalyticsLogger_Factory.java:40)
    at org.oppia.android.domain.oppialogger.analytics.LearnerAnalyticsLogger_Factory.get(LearnerAnalyticsLogger_Factory.java:12)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.domain.profile.ProfileManagementController_Factory.get(ProfileManagementController_Factory.java:107)
    at org.oppia.android.domain.profile.ProfileManagementController_Factory.get(ProfileManagementController_Factory.java:24)
    at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
    at org.oppia.android.app.application.dev.DaggerDeveloperApplicationComponent$ActivityComponentImplImpl.getAppLanguageWatcherMixin(DaggerDeveloperApplicationComponent.java:4899)
    at org.oppia.android.app.activity.InjectableAppCompatActivity.onInitializeLocalization(InjectableAppCompatActivity.kt:73)
    at org.oppia.android.app.activity.InjectableAppCompatActivity.attachBaseContext(InjectableAppCompatActivity.kt:33)
    at android.app.Activity.attach(Activity.java:7996)
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3431)
    ... 11 more

---

FATAL EXCEPTION: main
Process: org.oppia.android, PID: 12446
java.lang.RuntimeException: Unable to create service androidx.work.impl.background.systemjob.SystemJobService:
java.lang.IllegalStateException: WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate().
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:4254)
    at android.app.ActivityThread.access$1600(ActivityThread.java:244)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1970)
    at android.os.Looper.loop(Looper.java:236)
    at android.app.ActivityThread.main(ActivityThread.java:7861)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:600)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:967)
Caused by: java.lang.IllegalStateException: WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate().
    at androidx.work.impl.background.systemjob.SystemJobService.onCreate(SystemJobService.java:73)
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:4242)

```

Redmi Note 9S

Android 11

_No response_

### Additional Context

This error can also show up when running some existing tests locally.

BenHenning:
## Explanation
Fixes #5962

This PR introduces a force synchronization of platform parameter loading state for all activities in the app. This guards against a very specific edge case: if the app is killed due to a low memory situation, reopening it will bypass the splash activity (which ensures parameter initialization in concert with `AbstractOppiaApplication`) and lead to a race between the application kicking off loading parameters, the parameter states saving, and attempting to retrieve parameter state (#5962 goes into much more depth).

In most cases the new `runBlocking` introduces a very slight thread overhead only during early activity creation (since platform parameters only load one time at the beginning of the app startup). All new app cases will first ensure parameters are loaded through `SplashActivity`, so this new `runBlocking` _only_ has work to do in specifically the low memory case (or other similar manifestations of directly reloading the old activity stack, but this is the only one I'm aware of). It also doesn't kick off the actual loading, that happens earlier during application creation. From what I can tell there's a very slim duration between initialization kicking off and the activity being created (maybe 20ms according to my emulator), so it's expected that there's only a few milliseconds of work, at most, to synchronize to at this point, one time in the duration of the app lifecycle and only for this specific edge case.

The real risk here is in the nature of `runBlocking` which could, hypothetically, result in a deadlock. The team takes huge care in avoiding main thread usage in the entirety of the domain layer, and beyond that essentially never uses non-cooperative blocking (until now). So long as that trend continues it seems very unlikely that this scenario could ever result in a real problem. However, if it did then either a timeout could be introduced (which might not actually guarantee continuation depending on the nature of the deadlock), or the app eventually gets terminated (a main thread deadlock should result in an ANR).

The new test is an integration test specifically validating that `InjectableAppCompatActivity`'s new logic does what it's supposed to, and continues to. It emphasizes the latter since it's ultimately aiming to be a regression guard against future changes (since failing to initialize will likely result in failing tests, anyway). The new test really has one primary test for validating functionality: `testLaunchInjectableAppCompatActivity_withParameterLoading_doesNotThrowException`. However, it has 3 other tests that 'bolster' this to try and guard against future changes:
- Two tests demonstrate injecting a parameter directly and prove that initialization is necessary to successfully inject it. This is needed to demonstrate that the activity needs to do the same if it wants to inject that parameter, and injecting the parameter is used later on in the test (even though it can't trigger due to feature flags actually triggering the crash earlier in the activity lifecycle; this acts as a guard if those flags or how earlier they're used change in the future).
- One test tries to demonstrate the need for the activity's initialization logic by attempting to start it when forcibly disabling the initialization pathway. This effectively neuters the initialization logic in the activity (in the same way as if it were commented out in code).

The changes to `TestPlatformParameterModule` are needed to avoid the new `runBlocking` call from deadlocking all app layer tests (since attempting to load parameters requires `TestCoroutineDispatchers.runCurrent()` since it uses test-coordinated background dispatchers). It does this by leveraging Kotlin's `Unconfined` dispatcher to essentially synchronously perform the loading logic (with the necessary test coroutine synchronization barrier) within the new `runBlocking` call, thus working around the specific limitation. This entire process may end up being refined as part of addressing #5835 since the platform parameter initialization pathways for tests is getting quite complicated and deviates heavily from production flows (hence the need for this app layer-specific workaround).

Some miscellaneous notes:
- Note that the new interfaces don't need tests since they are not introducing any logic or complex contract.
- The new test is what I consider an integration test and thus doesn't have any corresponding production file that it's validating. This means it won't count for code coverage.
- The change to `ExplorationActivityLocalTest` is due to some event reodering that's happening. Some of the learner analytics events are dependent on the initialization order and the changes in `InjectableAppCompatActivity` cause parameters to be loaded a bit earlier, so it seems to be logging one of the later events earlier. It's unclear whether this matches production (or if the relative event ordering even matters here), but it seems fine to make the change for the test to pass.
- Note that an integration test is needed because it's not really viable to directly test the abstract activity classes without a lot of scaffolding (like `TestActivity`), so it's easier just to provide a very targeted integration test suite and leverage `TestActivity` for the actual `InjectableAppCompatActivity` testing.

## Essential Checklist

## For UI-specific PRs only
N/A -- this is addressing a crash with no direct UI changes.


BenHenning:
NB: Will revisit this later once CI runs and I can assess how UI tests are affected by this change.

I also need to figure out where to put a test for this. I suspect that's going to be rather tricky.

BenHenning (commented):
Self-reviewed the changes.

BenHenning:
Looks like every app test is timing out. :) It makes sense why that might be happening, but I'll need to noodle on it a bit to try and think of how to get the initialization order right in tests without deviating too much further from prod.

BenHenning (commented):
Skimmed latest changes.

BenHenning:
Putting this in review now since I do expect all of the tests to pass.

BenHenning:
PTAL @adhiamboperes or a delegate.

app/src/test/java/org/oppia/android/app/activity/PlatformParameterInitializationIntegrationTest.kt
```diff
@@ -0,0 +1,344 @@
...
+
+  @Test
+  fun testLaunchInjectableAppCompatActivity_withoutParameterLoading_throwsException() {
+    setUpTestApplicationComponent(skipParameterLoading = true)
+
+    // Trying to launch the activity when parameter loading is disabled will throw an exception.
+    // This is a hack to try and simulate InjectableAppCompatActivity missing its extra
+    // initialization logic to demonstrate that the failure still happens without it.
+    val exception: IllegalStateException = assertThrows<IllegalStateException>() {
+      // See asdf for why the parameter fetching is happening here.
```
adhiamboperes: Is asdf here an acronym or a placeholder?

adhiamboperes (approved):
Thanks @BenHenning! Nothing blocking, PR LGTM.

app/src/test/java/org/oppia/android/app/activity/PlatformParameterInitializationIntegrationTest.kt
BenHenning:
```suggestion
      // See testLaunchInjectableAppCompatActivity_withParameterLoading_doesNotThrowException for why
      // the parameter fetching is happening here.
```

BenHenning (commented):
Reviewed in-comment change.

app/src/test/java/org/oppia/android/app/activity/PlatformParameterInitializationIntegrationTest.kt
BenHenning: Oops--a placeholder that I forgot to update. This is meant to reference one of the other tests and has been updated accordingly.

Thanks!

BenHenning:
Thanks @adhiamboperes! Needed to re-rnu CI due to the comment update, but otherwise this should be good to merge now. Enabling auto-merge.

PR #6036, opened 2025-12-08

translatewiki:
Translation updates


BenHenning:
@seanlip is Hindi now something we need to support in the app? :) It looks like the app has been at least partially translated which means we'll need to officially enable it at least in developer and alpha builds to pull those in.

BenHenning:
Actually it looks like this is a bug in the resource checker. It should be using the all languages configuration (https://github.com/oppia/oppia-android/blob/develop/config/src/java/org/oppia/android/config/alllanguages/supported_languages.textproto) rather than a fixed enum like it is now:

https://github.com/oppia/oppia-android/blob/8bfb51930963b9619956d8496f6c059e169eaebf/scripts/src/java/org/oppia/android/scripts/xml/StringResourceParser.kt#L65-L86

We'll need to fix that before we can merge this PR.

seanlip:
> @seanlip is Hindi now something we need to support in the app? :) It looks like the app has been at least partially translated which means we'll need to officially enable it at least in developer and alpha builds to pull those in.

@BenHenning To your question, the platform strings have been fully translated to Hindi on translatewiki, but the content isn't ready yet. So I don't think we are ready to launch the app in Hindi yet.

BenHenning:
We have a PR right now that will add support for the check, and a separate issue #6044 is tracking fixing this in a generic way so that we don't run into it again for new languages (or, rather, that we can just update the textproto configuration file for supported languages).

BenHenning:
Also: it will be possible for us to distribute the new Hindi strings in the alpha version of the app (I think this may happen automatically, actually, once we update the textproto configuration).

BenHenning:
From a quick glance I see the following things that need to fix get fixed:
- StringParser needs to be updated to support the new Latin American Spanish language (ES-419), but also to parse BCP 47 qualifiers (`b+`) from file names correctly.
- We need issues to enable Hindi and Latin American Spanish in our language configuration. This probably will require an update to something because of the `b+` qualifier (maybe the utility we use to string unused languages?).
- The compute affected tests target is failing because of the `+` in the new string filename--we need to escape this in the Bazel query. This should be reproducible locally and needs a tracking issue.
- There are ~70 additional typography fixes needed for both Hindi and Spanish per the Android Lint findings. We need to surface this to the translation team or fix them ourselves in the translation console.

PR #6027, opened 2025-12-02
Update repository 'vacation' mode reminders for end-of-year 2025

BenHenning:
## Explanation
Updates the repository's 'vacation mode' to send automated replies to both new issues and PRs from 19 December 2025 to 05 January 2026 due to reduced team reviewer and triage coverage during that time period.

Note that we'll still have some folks coming in and out during that time, but we won't have consistent reviewer coverage until after the new year.

## Essential Checklist
- [ ] ~The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)~
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
N/A -- This is a developer workflow change that only impacts GitHub.

PR #6055, opened 2025-12-23
Correctly enable holiday messaging

BenHenning:
## Explanation
The holiday message system apparently never worked with the old configuration file, but for some reason it wasn't removed before which caused some confusion this time in #6027. It's now being removed and the actual messaging system is being restored.

The existing workflow has also been updated so that it uses environment variables at the repository level to be enabled/disabled to make things much easier--a PR shouldn't be needed to enable or disable it anymore. Some other notes:
- The workflow has been updated to use the latest issue creation action to update it.
- It's now using `ubuntu-latest`. The risk of this causing problems seems low compared to other workflows, and this way it doesn't need to be updated regularly.
- The latest `github-script` is being used for some future proofing.
- An issue comment is being generated now to avoid the multi-line formatting from being included in the comment and it being a bit ugly.
- The new environment variables already exist in the repository and are ready to be used. These changes won't go into effect until merged (due to using `pull_request_target`), but they've been thoroughly tested (see below).
- Some additional logs have been added for the new code to help future investigations of failures since they basically provide reasonable debug diagnostics of each part of the process.
- Much of the code is derived from AI generated code when prompted to rewrite the existing workflow to work correctly with Github Actions but also to use JavaScript rather than shell code.

For testing this was verified on a fork since it required changes directly to `develop`. Here's a demonstration of it working:
- https://github.com/BenHenning/oppia-android/pull/26#issuecomment-3687891606
- https://github.com/BenHenning/oppia-android/pull/25 demonstrates the workflow successfully not running when the dates are out of range.
- https://github.com/BenHenning/oppia-android/pull/20 demonstrates the workflow failing with an error if one of the dates is incorrectly formatted.
- https://github.com/BenHenning/oppia-android/pull/28 demonstrates the workflow working with the correct permission setup (which the earlier runs were missing but didn't cause failures due to the fork's default permissions being much more permissive than Oppia Android's).

## Essential Checklist
- [ ] ~The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)~

## For UI-specific PRs only
This only affects a GitHub workflow with no UI functionality ties.


BenHenning (commented):
Self-review spot check.

BenHenning:
@seanlip PTAL for owners change (just removing something from `CODEOWNERS`).

BenHenning (commented):
Quick spot check on codeowner change.

seanlip (approved):
LGTM for changes to codeowners file

adhiamboperes:
I have re-run the script check because Lint Analysis was failing but it shouldn’t have for this PR.

.github/workflows/repository_messaging.yml
```diff
@@ -8,17 +8,56 @@ on:
...
+          console.log('Start date:', startDateTimestamp);
+          console.log('End date:', endDateTimestamp);
+          console.log('Current date:', currentDateTimestamp);
+
+          if (isNaN(startDateTimestamp) || isNaN(endDateTimestamp)) {
+            core.setFailed("Start or end date is configured incorrectly. Please use an ISO 8601 timestamp for both.");
+            return;
+          }
+
+          const holidayIsActive = currentDateTimestamp > startDateTimestamp && currentDateTimestamp < endDateTimestamp;
```
manas-yu: Should we keep the dates inclusive, or is the exclusivity intentional?

.github/workflows/repository_messaging.yml
```diff
@@ -8,17 +8,56 @@ on:
 
 jobs:
   holiday_message:
```
manas-yu, line 10: Could add a permission block with PR write and content read.

manas-yu (changes requested):
@BenHenning changes look good, just had some minor comments.

.github/workflows/repository_messaging.yml
BenHenning: It shouldn't meaningfully matter given it's a difference of milliseconds. That being said, since I needed to update the script for permissions anyway I did go ahead and make the start inclusive.

.github/workflows/repository_messaging.yml
BenHenning: Excellent call-out. I missed this and didn't discover it in my testing because my fork's default permissions were much more permissive than Oppia's. Re-running with the same repo permissions did, indeed, cause a failure in https://github.com/BenHenning/oppia-android/pull/27. Correcting the script (latest changes) resulted in it working in https://github.com/BenHenning/oppia-android/pull/28.

BenHenning (commented):
Spot check reviewer fixes.

BenHenning:
Thanks all.

@manas-yu PTAL for review again.

manas-yu (approved):
Thanks changes LGTM.

PR #6061, opened 2025-12-31
Fix #3506, #5950, #5951: Add support for remote platform parameters [Blocked: #6062]

BenHenning:
TODO: Finish one PR is actually code complete.

## Essential Checklist

- [ ] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [ ] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [ ] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [ ] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [ ] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [ ] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

Issue #3506, opened 2021-07-20
Change the Endpoint for retrieving Platform Parameters


Arjupta:
Currently the **PlatformParameterService** points to a demo endpoint for fetching platform parameters from the Oppia Backend. This url needs to be changed to a new endpoint after the platform parameter project gets completed. For doing so we need to first make the neccessary changes in the Oppia Backend to support the expected format of response needed and Afterwards we will raise a Pull request that changes the url in PlatformParameterService

BenHenning:
This has actually evolved since the original issue was filed. The web team has completed a revamping of the feature flag & platform parameter system. As of now:
- Feature flags are not exposed via any endpoints for Oppia Android to consume, so we won't be able to leverage web's new feature flag functionality. We don't plan to change this until after GA.
- We're considering leveraging platform parameters for feature flags on Android, but this is still being discussed.
- Platform parameters on web currently do not support platform-specific retrieval like they did before, and this functionality is planned to be reintroduced. The platform parameter controller was also removed, so it will need to be reintroduced.

As of now, no work can be done on this until web is updated to reintroduce its platform parameter controller with support for platform evaluation (so that Android only receives Android-specific constants).

Once that's completed, the following code changes will need to be made:
- PlatformParameterService: may need updates to call into the correct handler and with the correct values.
- PlatformParameterSyncUpWorker: may need changes if the API for the service changes.
- New platform parameters should be introduced on web to introduce support for app deprecations (see notes in #5025).

The code changes above and the end-to-end behavior will need to be verified in two stages:
- Stage 1: local verification.
  - A dev version of the web server needs to be started (with the new functionality introduced).
  - A local build of the app needs to be deployed to an emulator and configured to use the local development version of the app (see https://stackoverflow.com/a/6310592). This will require a change to NetworkConfigProdModule (to point away from oppia.org).
  - The platform parameters for app deprecation (#5025) should be tweaked via the local dev app admin panel to deprecate the local dev app.
  - The local dev app should be verified to sync by seeing the deprecation after an app restart (need to ensure the background worker runs).
- Stage 2: production verification.
  - The app deprecation feature should be enabled in production Oppia web (without any versions being deprecated).
  - We should verify in newer versions of the app that they're logging the feature as enabled via the feature list event.

BenHenning:
@adhiamboperes this is a critical issue for GA--I don't think we should be removing it.

BenHenning:
[...]

PR #6062, opened 2026-01-02
Fix #5989, #6053, #6059, #6060, #6097, #6200: Rebuild app & `WorkManager` initialization

### LINKED ISSUE #5989: [BUG]: PlatformParemeterSyncUpWorkerFactory and MetricLogSchedulingWorkerFactory are not being initialised correctly
### Describe the bug

The `createWorkerMethod` of the `PlatformParemeterSyncUpWorkerFactory` and `MetricLogSchedulingWorkerFactory` is not set up properly. This leads to instances such as duplicate factories being created and/or the worker not being initialised correctly once added to the `DelegatingWorkerFactory` on the `WorkManagerConfigurationModule`.  This needs to be fixed in a similar way to what is shown below. The fix should also extend to the `LogUploadWorkerFactory`.

Sample fix;

### LogUploadWorkerFactory
```kotlin
class LogUploadWorkerFactory @Inject constructor(
  private val workerFactory: LogUploadWorker.Factory
) : WorkerFactory() {

  override fun createWorker(
    appContext: Context,
    workerClassName: String,
    workerParameters: WorkerParameters
  ): ListenableWorker? {
    return if (workerClassName == LogUploadWorker::class.java.name) {
      workerFactory.create(appContext, workerParameters)
    } else null
  }
}
```

### PlatformParameterSyncUpWorkerFactory
```kotlin
class PlatformParameterSyncUpWorkerFactory @Inject constructor(
  private val workerFactory: PlatformParameterSyncUpWorker.Factory
) : WorkerFactory() {

  override fun createWorker(
    context: Context,
    workerClassName: String,
    workerParameters: WorkerParameters
  ): ListenableWorker? {
    return if (workerClassName == PlatformParameterSyncUpWorker::class.java.name) {
      workerFactory.create(context, workerParameters)
    } else null
  }
}
```

### MetricLogSchedulingWorkerFactory
```kotlin
class MetricLogSchedulingWorkerFactory @Inject constructor(
  private val workerFactory: MetricLogSchedulingWorker.Factory
) : WorkerFactory() {

  override fun createWorker(
    appContext: Context,
    workerClassName: String,
    workerParameters: WorkerParameters
  ): ListenableWorker? { // must be nullable!
    return if (workerClassName == MetricLogSchedulingWorker::class.java.name) {
      workerFactory.create(appContext, workerParameters)
    } else null
  }
}
```

### Steps To Reproduce

1. Uninstall the Oppia app or clear storage.
2. Open logcat and filter by the keyword "Worker".
3. Reinstall the app on a mobile device with Android 10 and below.
4. Observe the logcat.
5. You should be able to see the MetricLogSchedulingWorker and PlatformParameterSyncUpWorker reporting Failure results after they run.

### Expected Behavior

Both workers should succeed.

### Screenshots/Videos

Image

_No response_

Android 10

_No response_

### Additional Context

_No response_

### LINKED ISSUE #6053: [BUG]: Work manager integration is incorrect
### Describe the bug

I've spent the last few hours trying to debug why workers aren't running (or, more specifically, how to force a specific job to run and then finding out there are a lot of problems preventing me from doing that). With a lot of help from Gemini, I was able to narrow down to at least two very significant problems:
1. We are initializing `WorkManager` incorrectly. Apparently, there are three ways to initialize it: (1) default initialization, (2) custom configuration (which requires disabling the default initializer), or (3) manually calling `WorkManager.initialize()`. We are doing all three of these, and are also being way over-zealous in disabling startup logic (we're disabling everything in AndroidX startup right now rather than just specifically `WorkManager`). This can lead to lots of initialization and consistency issues.
2. Our custom `WorkManager` configuration is using a `DelegatingWorkerFactory` with custom worker factories, but those factories are implemented incorrectly. Currently, each factory will return exactly its instance of the worker. However, the delegating factory simply **enumerates each provided factory until one returns non-null**. We are not performing any type check on the class name which means **every worker being created is always a `LogUploadWorker`**.

I haven't actually fixed (2) locally yet to continue my debugging, but I suspect there may yet be other problems. The current implementation seems to make it impossible for all other workers to run (including for platform parameters and performance metrics). However, we definitely upload performance metrics and neither I nor Gemini can reason about why that is given the given configuration (and confirmed logcat errors about jobs not successfully starting up or quickly failing).

### Steps To Reproduce

This is very much a configuration-and-inspect-logcat problem so I can't quickly provide steps to reproduce. Essentially we need to fix the setup of `WorkManager` and validate that all of the jobs can successfully run.

### Expected Behavior

The configuration should be correct and jobs should successfully run.

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

It would also be nice to include instructions in the wiki as part of addressing this issue about how to forcibly start a specific job (such as platform parameter sync) by pulling job information from `WorkManager`'s database. Note that these instructions may deviate on different versions of `WorkManager`, but for us we can rely on the database existing under the app's `no_backup` directory.

### LINKED ISSUE #6059: [BUG]: WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate() [top crash]
### Describe the bug

This is our largest current crash cluster and it seems to have begun in 0.16 (though it's possible it's being miscategorized and actually existed earlier, but it definitely seems much worse starting in 0.16). Stack trace:

```
Fatal Exception: java.lang.RuntimeException: Unable to create service androidx.work.impl.background.systemjob.SystemJobService: java.lang.IllegalStateException: WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate().
       at android.app.ActivityThread.handleCreateService(ActivityThread.java:2953)
       at android.app.ActivityThread.access$1900(ActivityThread.java:168)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1462)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:150)
       at android.app.ActivityThread.main(ActivityThread.java:5665)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:822)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:712)

Caused by java.lang.IllegalStateException: WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate().
       at androidx.work.impl.background.systemjob.SystemJobService.onCreate(SourceFile:73)
       at android.app.ActivityThread.handleCreateService(ActivityThread.java:2940)
       at android.app.ActivityThread.access$1900(ActivityThread.java:168)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1462)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:150)
       at android.app.ActivityThread.main(ActivityThread.java:5665)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:822)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:712)
```

### Steps To Reproduce

Unknown--this is an observed ongoing crash for users.

### Expected Behavior

No crashes.

### Screenshots/Videos

_No response_

_No response_

_No response_

0.17-beta-92346c369d

### Additional Context

_No response_

### LINKED ISSUE #6060: [BUG]: java.lang.IllegalStateException - Attempting to access feature flag LOGGING_LEARNER_STUDY_IDS before initialization
### Describe the bug

This is a crash being observed in production through Firebase. Stack trace:

```
Fatal Exception: java.lang.RuntimeException: Unable to create service androidx.work.impl.background.systemalarm.SystemAlarmService: java.lang.IllegalStateException: Attempting to access feature flag LOGGING_LEARNER_STUDY_IDS before initialization. If this is a test, is it using TestPlatformParameterModule?
       at android.app.ActivityThread.handleCreateService(ActivityThread.java:3158)
       at android.app.ActivityThread.access$1900(ActivityThread.java:177)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1531)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:145)
       at android.app.ActivityThread.main(ActivityThread.java:5942)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)

Caused by java.lang.IllegalStateException: Attempting to access feature flag LOGGING_LEARNER_STUDY_IDS before initialization. If this is a test, is it using TestPlatformParameterModule?
       at iB.aF.b(SourceFile:141)
       at iB.c.<init>(SourceFile:103)
       at iB.b.a(SourceFile:1102)
       at iB.a.n(SourceFile:49)
       at iB.q.b(SourceFile:1048)
       at iS.ai.b(SourceFile:1040)
       at es.d.b(SourceFile:47)
       at iT.i.b(SourceFile:1041)
       at iT.x.b(SourceFile:1036)
       at es.d.b(SourceFile:47)
       at org.oppia.android.domain.oppialogger.analytics.E.b(SourceFile:1103)
       at es.d.b(SourceFile:47)
       at org.oppia.android.domain.oppialogger.loguploader.l.b(SourceFile:1071)
       at org.oppia.android.domain.oppialogger.loguploader.k.b(SourceFile:1030)
       at iK.b.b(SourceFile:1048)
       at es.d.b(SourceFile:47)
       at org.oppia.android.app.application.beta.f.a(SourceFile:2566)
       at org.oppia.android.app.application.AbstractOppiaApplication.a(SourceFile:92)
       at androidx.work.impl.v.a(SourceFile:147)
       at androidx.work.impl.background.systemalarm.f.<init>(SourceFile:88)
       at androidx.work.impl.background.systemalarm.f.<init>(SourceFile:76)
       at androidx.work.impl.background.systemalarm.SystemAlarmService.b(SourceFile:93)
       at androidx.work.impl.background.systemalarm.SystemAlarmService.onCreate(SourceFile:45)
       at android.app.ActivityThread.handleCreateService(ActivityThread.java:3148)
       at android.app.ActivityThread.access$1900(ActivityThread.java:177)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1531)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:145)
       at android.app.ActivityThread.main(ActivityThread.java:5942)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)
```

### Steps To Reproduce

Unknown. This is a cluster of user-faced crashes.

### Expected Behavior

No crash.

### Screenshots/Videos

_No response_

_No response_

_No response_

0.17-beta-92346c369d

### Additional Context

_No response_

### LINKED ISSUE #6097: [BUG]: It's possible for Oppia web to (eventually) permanently disable platform parameter syncing
### Describe the bug

I noticed this in passing as a theoretical possibility.

The app will eventually allow Oppia web to override its sync-up time for platform parameters. If this threshold were set to an extremely large value (like 1 million) it would effectively stop all apps that receive that value from ever syncing again.

### Steps To Reproduce

Not really necessary.

### Expected Behavior

The app probably needs to have an upper bound default, though we need to weigh this against a potential DoS protection case on the web side. Perhaps a very long default (like once per month) might be a reasonable worst-case fallback.

Separately, we may as well clamp the lower bound to 15 minutes as well. Even though work manager does this for us, it's nice to be explicit about it to make sure that there doesn't even _seem_ to be the possibility of configuring a potential denial-of-service on the web side (e.g. setting it to 0 to force the entire fleet of apps that receive the parameter to start indefinitely hitting the server).

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

_No response_

### LINKED ISSUE #6200: [BUG]: App crashes after reopening (WorkManager initialization conflict)
### Describe the bug

The app works correctly when I build and run it initially. However, after some time (around 1–2 days), when I reopen the app, it crashes immediately on launch.

If I rebuild and reinstall the app, it starts working normally again.

This behavior is consistent and reproducible after leaving the app unused for some time.

### Steps To Reproduce

1. Build and install the app using adb install
2. Use the app normally (it works fine)
3. Leave the app unused for 1–2 days
4. Open the app again
5. Observe that the app crashes on launch

### Expected Behavior

The app should open normally without crashing, even after being unused for some time.

### Screenshots/Videos

https://drive.google.com/file/d/170Ze4c0n2IeCs2GRmaQsczBdvHZW5722/view?usp=sharing

vivo T2x 5G

Android Version 15

version 0.18-dev-23771e3ef5

### Additional Context

I am using adb install to install the app on a physical device.
I captured logcat during the crash. Relevant logs:

```
--------- beginning of crash
04-05 14:36:43.899 29703 29703 E AndroidRuntime: FATAL EXCEPTION: main
04-05 14:36:43.899 29703 29703 E AndroidRuntime: Process: org.oppia.android, PID: 29703
04-05 14:36:43.899 29703 29703 E AndroidRuntime: kotlinx.coroutines.CompletionHandlerException: Exception in completion handler
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.JobSupport.notifyCompletion(JobSupport.kt:1525)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.JobSupport.completeStateFinalization(JobSupport.kt:323)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.JobSupport.finalizeFinishingState(JobSupport.kt:240)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.JobSupport.tryMakeCompleting(JobSupport.kt:863)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:100)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at android.os.Handler.handleCallback(Handler.java:1014)
04-05 14:36:43.899 29703 29703 E AndroidRuntime: Caused by: java.lang.Exception: Failed to continue application initialization.
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at org.oppia.android.app.application.AbstractOppiaApplication.onCreate
04-05 14:36:43.899 29703 29703 E AndroidRuntime: Caused by: java.lang.IllegalStateException: WorkManager is already initialized.
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at androidx.work.impl.WorkManagerImpl.initialize
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at androidx.work.WorkManager.initialize
04-05 14:36:43.899 29703 29703 E AndroidRuntime:        at org.oppia.android.app.application.AbstractOppiaApplication.onCreate

```

```
04-05 14:36:45.583 29828 29828 E AndroidRuntime: FATAL EXCEPTION: main
04-05 14:36:45.583 29828 29828 E AndroidRuntime: Process: org.oppia.android
04-05 14:36:45.583 29828 29828 E AndroidRuntime: java.lang.RuntimeException: Unable to create service androidx.work.impl.background.systemjob.SystemJobService
04-05 14:36:45.583 29828 29828 E AndroidRuntime: Caused by: java.lang.IllegalStateException:
WorkManager needs to be initialized via a ContentProvider#onCreate() or an Application#onCreate()
04-05 14:36:45.583 29828 29828 E AndroidRuntime:        at androidx.work.impl.background.systemjob.SystemJobService.onCreate
```

This suggests a possible conflict in WorkManager initialization.

BenHenning:
## Explanation

Fixes #5989
Fixes #6053
Fixes #6059
Fixes #6060
Fixes #6097
Fixes #6200

This PR fixes the app's `WorkManager` configuration by essentially rebuilding how it works on a fundamental level.

### Existing behaviors

Previously, the app attempted to provide a custom `WorkManager` `Configuration` (via `AbstractOppiaApplication`) which, per `WorkManagerConfigurationModule`, added a bunch of custom `WorkerFactory`s per a provided `DelegatingWorkerFactory`. However, there are two problems with this setup (#6053 covers both of these):
1. In order to provide a custom `Configuration` the app must remove the custom initialization provider (https://developer.android.com/develop/background-work/background-tasks/persistent/configuration/custom-configuration). Technically this is being done in the app today, however it's removing all app initialization which can affect more than just `WorkManager`.
    - This is expected to be one of the root problems behind #6059 (though it cannot be easily verified since we haven't determined a reproduction of the issue).
    - Note that the fix in the PR for this also involves setting `exported` to `false` for the `provider`. This is technically redundant since `WorkManager` already declares it as not exported, but the Android Linter is only reviewing the local `AndroidManifest.xml` and not the fully merged manifest so it's a reasonable way to silence the warning (and be a bit more explicit about safety).
3. More problematically, the custom `WorkerFactory`s are not adhering to `DelegatingWorkerFactory`'s contract. The delegating factory expects subsequent factories to return `null` if they do not match the request worker being created, but Oppia's workers _always_ return non-null. This means that the first worker always wins which is in this case the `MetricLogSchedulingWorker` (hence why we didn't notice this easily before because the majority of analytics events are successfully being uploaded, and most performance metrics are captured via direct app callbacks). This also wasn't caught in testing because of tests directly validating the jobs and not their scheduled and periodic behaviors. #5989 is tracking this issue.

Another issue with the current setup is that platform parameters cannot be safely initialized when the app starts up via a worker which has led to a top crash: #6060. That issue thread plus the documentation in `AbstractOppiaApplication` go into a lot more detail about the initialization order changes.

#### `BootstrapOppiaWorker`

The new design involves introducing a single `WorkManager` worker: `BootstrapOppiaWorker`. This introduces a very elegant improvement over the previous solution since it allows us to solve several classes of common issues every worker will face:
- Renaming the worker causes its previous runs to fail.
- The worker's factory needs to correctly identify that _it_ is the correct worker to instantiate for the particular task (the original problem leading to this PR), and in a way that doesn't interfere with `WorkManager`'s own internal workers (which run through the same factory).
- Dependency injection and needs to be correctly supported.
- Worker logic needs to be interoperative with Oppia's background dispatchers so that they can be tested.
- Worker failures need to correctly lead to periodic task cancellation.
- Rogue jobs are correctly cancelled (to handle the rename case above and to ensure that such jobs aren't kept indefinitely running by `WorkManager`; these now clear themselves). This behavior works retroactively for past Oppia workers, as well, once this code is deployed to user's devices.
- Platform parameter initialization must be done correctly.

Much of these weren't actually handled in the existing design. Using a common worker allows all of that to be solved in one place, and additional benefits could also be realized:
- No need to ever change the `Configuration` implementation since the worker itself never changes.
- Stricter controls over interacting with `WorkManager` (which actually, combined with some testing requirements, led to a complete isolation of `WorkManager` throughout the codebase).
- Strong typing, automatic subtask support, and simpler worker implementations using a nice `OppiaWorker` interface with Dagger-managed bindings. In the case of subtasks, this is mainly establishing an architectural pattern around an emergent behavior that was already the case for basically all of the current workers. Each subtask operates as a distinct top-level worker from the eyes of `WorkManager`.

Note that `BootstrapOppiaWorker` and `WorkManagerConfigurationModule` are both being very thoroughly tested to ensure everything built on them can be comfortably relied upon and to cover the specific core cases that led to needing this PR in the first place (to help guard against regressions, however unlikely with these two classes rarely needing to change).

#### New initialization logic

Beyond that, the recreation of initialization led to some other nice improvements:
- `ApplicationStartupListener` was updated to have two different callbacks depending on whether platform parameters are initialized. This actually provides significant more stability in listeners that rely on platform parameters.
  - There's also some specific trickiness around the threads used for each callback and this is documented in the listener's KDocs.
  - Note that this split was specifically critical in fixing #6060 because there was a start-up race condition between `WorkManager` trying to run a worker that relies on platform parameters and `AbstractOppiaApplication` attempting to asynchronously load platform parameters. This two-step initialization process fully eliminates that race condition specifically for the path of a worker immediately running upon scheduling it (see next point). The other part of #6060 was fixed by forcing `BootstrapOppiaWorker` to synchronously load platform parameters before proceeding with creating the actual underlying `OppiaWorker`.
- `AnalyticsStartupListener` was replaced with `StartupWorkerScheduleReadinessListener` since that's effectively what it was: an application-bound listener that wanted to run on startup to schedule periodic jobs to run. Now that's done completely in an initialization order-safe way and with a custom scheduler (described later in this description).

#### New scheduler

A new `WorkManagerScheduler` has been introduced to simplify interacting with `WorkManager` when wanting to schedule jobs by reusing common and obvious constraints across jobs and also ensuring that jobs cannot be scheduled in a way where they'd effectively never run (to address the potential of #6097 being an issue). Jobs are enforced to run no more than every 15 minutes (which is also a hard `WorkManager` rule) and no less often than every 14 days.

This scheduler works in tandem with the new `StartupWorkerScheduleReadinessListener` to provide an easy-to-use method for defining custom schedulers for a job. `StartupWorkerScheduleReadinessMonitor` is a new utility (bound itself as an `ApplicationStartupListener`) which injects all of the `StartupWorkerScheduleReadinessListener`s and calls each at an initialization order safe time to schedule jobs (shortly after platform parameters are fully initialized). A few things to note:
- The monitor is actually implemented in a way where it's safe for `StartupWorkerScheduleReadinessListener` implementations to use platform parameters directly (which is actually necessary since they may use platform parameters for configurations like job periodicity intervals as is the case for `MetricLogSchedulingWorkerScheduler`.
- This approach of making the monitor itself an `ApplicationStartupListener` helps to keep `AbstractOppiaApplication` as lean as possible. One thing found during the development of this PR was that reasoning about `AbstractOppiaApplication`'s initialization order and nuances was really challenging. Moving responsibilities out of the `onCreate()` should hopefully help make it much easier to understand and perform the necessary isolation for correctness.

One limitation to note: the current version of `WorkManager` means that `WorkManagerScheduler` must use `ExistingPeriodicWorkPolicy.KEEP` when scheduling jobs whereas `ExistingPeriodicWorkPolicy.UPDATE` would actually be better since the latter properly supports property changes being applied without needing to cancel and restart the job. Practically this isn't expected to cause any problems for a while but it could potentially in the future (such as if we tweak the periodicity of performance metric captures or platform parameter syncing). #6115 is capturing the upgrade work needed to address this and there's a TODO pointing to that issue in code to ensure the update happens once it's possible.

Finally, actual scheduling logic relies on `WorkManager`. Even though we attempt to schedule jobs upon startup they may not actually run unless it's time for them to run again (i.e. their constraints are met and they haven't run within their configured interval). In most cases, requesting to schedule a job is a no-op since it's will already be tracked by `WorkManager`.

#### `OppiaWorker`

`OppiaWorker` is described a bit above but it's meant to be an extremely straightforward way to represent scheduled workers without having to worry about the extra complexity of `WorkManager` and edge integration-specific quirks. The idea is that all a developer needs to do is:
- Create a new `OppiaWorker` implementation with whatever subtasks are desired.
- Create a new scheduler class that implements `StartupWorkerScheduleReadinessListener` to schedule the worker's subtasks upon application startup.
- Create a new module class that binds both the listener and the worker (based on the worker's defined string name).
- Add tests for each of the previous classes using the new testing functionality (described below).

And that's basically it. An `OppiaWorker` implementation and its `StartupWorkerScheduleReadinessListener` implementation can both directly depend on platform parameters and feature flags as needed and perform whatever background logic that needs to be done for the worker. The hope is this significantly simplifies implementing such workers in the future.

### Developer reference improvements

To help illustrate how to set up a new `OppiaWorker` and test it, a developer-only `DebugWorker` has been introduced along with its scheduler and module classes. This is a fully functioning worker that actually runs in `//:oppia_dev` builds of the app and can be used as both a coding and maintenance reference (see wiki bits below).

A bunch of updates were made to the `WorkManager` documentation covering not only the new implementation and testing patterns but also how to actually forcibly run workers using `adb`. Unfortunately these instructions basically only work when `adb` can be rooted which is limited for personal devices (but should be fine for emulators). Hopefully these instructions significantly help when debugging workers in the future.

### Testing infrastructure changes

Testing workers is surprisingly complex which is likely why existing tests weren't actually validating the periodicity part of workers, and were introducing a lot of "backdoors" to directly access configuration values rather than reading what was passed to `WorkManager`. A lot of investigation and iteration went into the creation of a new `OppiaWorkManagerTestDriver` utility which solves all of these (and more) problems:
- It makes it possible to actually introspect a job's characteristics even as it's running. This gives us high confidence that a job's scheduler is actually correct in what it passes `WorkManager`.
- It uses `WorkManager`'s own internal testing library for managing constraints (though `WorkManager` itself doesn't do a great job here so the utility is limited).
- It makes it possible to test periodicity. This is actually **very** complex and requires essentially simulating `WorkManager`'s own behavior in a way that's interoperable with `TestCoroutineDispatchers`. I suggest looking at the code to understand what this does more.
  - Fortunately, this can be simplified once #6115 is addressed because `USE_TIME_BASED_SCHEDULING` can be used which allows direct binding to the `FakeOppiaClock` and seems to actually "just work" with Oppia's test coroutine management (with some other tweaks).
  - Unfortunately, #6115 requires a Kotlin upgrade (possibly two) which is definitely outside the scope of this PR. I tested a forced version of 2.9.0 to come to the above conclusions, and I won't repeat how I did that since it's so bad it should really never be done even for testing purposes.
- It removed the need for `FakeLogScheduler`, `FakeLogUploader`, and `FirebaseLogUploader` (and `FirebaseLogUploaderModule`) since workers should actually just be tested directly, instead.

It's worth noting that the configuration test setup for the driver is rather complex and won't be fully described here since the documentation and code do a better job (though one thing to note is that using the driver will change the time management mode of the test away from the standard Oppia Android default). This and the other overall complexity of properly interacting with `WorkManager`, its driver, or `TestCoroutineDispatchers` is ultimately why this PR also mandates using this driver over `WorkManager` or its test utility directly. Trying to make jobs work correctly in a test environment is a complete minefield, so it makes sense to just eliminate whole classes of developer frustration and confusion by requiring using a utility that trivializes the majority of these interactions.

Some other notes:
- There is some fragility in the driver implementation (specifically that it interacts directly with `WorkManager`'s internal database) that might break in future versions, but it's the best solution for what it's being used for currently.
- `CoroutineExecutorService` is needed as part of helping to bridge the gap and ensure `TestCoroutineDispatcher` interoperability. This will continue to be the case even after #6115 is addressed, and will likely continue until #3715 is addressed.

### Changes to worker implementations & simplifications

The event log upload worker, performance metrics capture worker, and platform parameter worker all obviously require substantial reworking (along with their corresponding peer classes) to fit the new architectural patterns. Note that some attempt was also made to ensure naming consistency when it was easy (which is why `LogUploadWorker` remains inconsistent for now--updating it to be consistent would be a rather large codebase-wide change on top of an already very large PR for minimal gain). Fortunately the new architectural pattern makes renaming these classes trivial in the future.

Note that this change does migrate over the platform parameter work which means it actually should run now. However, it doesn't actually perform syncing logic anymore and that will be addressed in #6061 (which actually prompted the investigation that led to this PR).

### Miscellaneous non-test changes

There are a bunch of other changes that were made in passing:
- `ApplicationLifecycleListener` has a new KDoc update that comes from the fact that `ProcessLifecycleOwner` has built-in delays for foreground/background changes to avoid invalid "went to background"/"wen to foreground" subsequent events just for navigating activities normally (since `onStop`/`onStart` is called between navigation moments). This was found to cause some initialization order complexities in tests so the documentation is meant to help make this clear for other developers in the future.
- `ApplicationLifecycleObserver` had a bunch of changes:
  - It was split into itself and a new `ApplicationLifecycleLogger` which is responsible for the actual metric logging so that the observer can focus on activity-level state tracking.
  - The observer now  makes use a command queue to handle complex ordering: even if the app isn't done initializing state changes still need to be captured _immediately_ upon app start (actually logging events can catch up  later). This wasn't something being done before and couldn't be due to the delayed initialization for platform parameters.
    - Changes can't be immediately processed since the logger indirectly depends on platform parameters which is why it needs to be injected as a `Provider`.
    - The command queue also resulted in some additional synchronizations and broader changes in tests.
- `AbstractOppiaApplication.onCreate` was further simplified by abstracting away `AppCheckProviderFactory` to the class's constructor. This allows it to default to the default production-friendly factory but also allow `DeveloperOppiaApplication` to use `DebugAppCheckProviderFactory` as needed. This led to an overall cleanup of no longer needing the build flavor as part of the root application component anymore.
- A small change in `AndroidManifest.xml` instructions was added to clarify how to actually enable faster event logs. This isn't directly related to the PR but it was noticed in passing. The previous support article was changed over the last few years and it's much harder to find the actual `adb` command needed.
- `CpuPerformanceSnapshotter` was made injectable to simplify the Dagger graph a bit and which modules are actually needed for certain test scenarios. It's not obvious why it wasn't done this way originally from looking at code history.
- Timestamps are propagated in a few more log places than before due to changes in time management in certain tests (due to the restrictions on setting time when using `OppiaWorkManagerTestDriver`).
- `ExceptionsController` was noticed in passing that it wasn't a singleton, and this has been fixed. It needs to be to avoid potential race conditions against its `PersistentCacheStore` if/when more than one is created. This almost certainly has caused unseen problems.
- `UncaughtExceptionLoggerStartupListener` was updated in a few ways:
  - It was reworked to only try and log exceptions when it's safe (i.e. when platform parameters can be injected) since otherwise it would silently fail and lead to no exceptions being logged or propagated upon app death.
  - It was also noticed that the utility wasn't correctly swapping out the _default_ uncaught exception handler, only the main thread's. This has been corrected, and Gemini helped with writing the tests in this case to validate that it's actually correct. Many more exceptions will now be caught by the utility.

### Miscellaneous test & test infrastructural changes

There are other miscellaneous test and test infrastructure changes outside the main work manager bits:
- All `WorkManager` initialization was removed in tests that didn't actually use workers (e.g. `ProfileAndDeviceIdActivityTest`).
- `ExplorationActivityLocalTest` did have an event reordering update but I don't remember precisely why this was correct.
- `ApplicationLifecycleObserverTest` was updated to use an actual real activity to validate lifecycle integration (since before it wasn't actually testing that the utility was wired correctly). There's some messy reflection to ensure `ProcessLifecycleOwner` properly resets across tests because it uses static state internally and the current version being used doesn't allow it to be properly reset. Addressing this is being tracked in #6187.
- `CpuPerformanceSnapshotterTest` has been updated so that the snapshotter itself is being tested rather than testing happening via `ApplicationLifecycleObserver` for more correct separations of concern.
- `TestPlatformParameterModule` has become even more complex due to a realization during debugging that `TestCoroutineDispatchers.runCurrent()` cannot be safely called off the main thread (see #6116), yet it still being possible via certain worker initialization pathways for platform parameters to now be needed off the main thread. This has been addressed by specifically delegating back to the main thread and correctly blocking on it.
  - Note that the `Looper` check done is actually an SDK 21-compatible variant to `isCurrentThread` which is, unfortunately, SDK 23+. This technically doesn't matter for Robolectric tests but Android Lint is unhappy without this approach.
  - Separately, some new parameter overrides have also been added.
- The various fake event loggers `FakeAnalyticsEventLogger`, `FakeExceptionLogger`, `FakeFirestoreEventLogger`, and `FakePerformanceMetricsEventLogger` have been updaed to have similar APIs (particularly for event count and exceptions) since this was needed for a bunch of tests. The loggers' tests themselves have also been updated to reflect these changes.

### Script, pattern, an exemption updates

Other non-app changes:
- `KdocValidityCheck` was updated to also ignore `Multibinds` (which it should've done before, this was just missed since it's a rarely used annotation). A new test was added to validate this behavior change.
- A bunch of new regex checks were added to further restrict `WorkManager` library interaction as mentioned in previous sections.
- No new tests are being added for:
  - `StartupWorkerScheduleReadinessMonitor` for simplicity since it's a large PR and a relatively simple utility. #6189 is tracking adding tests for this in a future PR.
  - `ApplicationLifecycleLogger` because `ApplicationLifecycleObserver`'s tests need to be audited to determine which should move into a dedicated logger test suite, and what additional logger-specific tests should be added. This is being tracked in #6190.
  - `OppiaWorkManagerTestDriver` because it's highly complex to test and not worth it at this point because it will actually be simplified with the upgrade to 2.9.0 per #6115. Nevertheless, adding tests for this driver is being tracked by #6191.
- Some exemptions were actually removed since there are tests for them now (such as for `MetricLogSchedulingWorkerScheduler`). This revealed that the `TestFileCheck` utility doesn't correctly fail when an exemption is added but a test actually exists. #6188 is tracking fixing this.
- `DebugWorkerDebugModule` was exempted for coverage compatibility since it actually caused a failure during the finalization of the PR. I didn't dig into the specifics as to why it has an incompatibility, but this seems to be a reasonable workaround for now.

### Disclosure of LLM usage

I did use Gemini pretty heavily for diagnosing the early issues with the work manager integration and how to effectively debug them. This actually all began with #6061 when I was trying to figure out why I couldn't get the jobs to run, and pulling on that thread led (with LLM help) uncovered all of the broader problems that ended up being addressed here. Gemini did help a bit with ideating on an approach but the vast majority of the bootstrapping pattern and all of its corresponding code came from me. I used Gemini CLI a bit for some of the testing for `UncaughtExceptionLoggerStartupListener` but that was it.

## Essential Checklist

## For UI-specific PRs only
Largely this PR should have zero impact on direct UIs in the app, though it will indirectly affect the overall UX since it will enable platform parameters to actually work correctly (though not entirely, #6061 still needs to be finished). No screens are being updated as part of these changes.


BenHenning:
Note that this is probably in a good enough place for me to work off of it to make progress on #3506 (which was my original goal heading into this), and I'll need to circle back to this PR. It still needs a lot of documentation and testing work, including figuring out hopefully a better testing strategy than using fakes (I'm hoping it's possible to actually properly orchestrate kicking off the jobs but I have a lot more digging to do in order to figure that out fully).

testing/src/main/java/org/oppia/android/testing/FakeFirestoreEventLogger.kt
```diff
@@ -10,11 +10,22 @@ import javax.inject.Singleton
...
+  private var failureToThrow: Exception? = null
 
   override fun uploadEvent(eventLog: EventLog) {
+    val possibleFailure = failureToThrow
+    if (possibleFailure != null) throw possibleFailure
     eventList.add(eventLog)
   }
 
+  /**
+   * Sets an [Exception] to throw the next time(s) [logPerformanceMetric] are called, or `null` if
```
adhiamboperes: These events fall under analytics events, not Performance Metrics AFAIK.

domain/src/test/java/org/oppia/android/domain/platformparameter/syncup/PlatformParameterSyncUpWorkerModuleTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import javax.inject.Singleton
+
+/** Tests for [PlatformParameterSyncUpWorkerModule]. */
+@Suppress("FunctionName") // FunctionName: test names are conventionally named with underscores.
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config(application = PlatformParameterSyncUpWorkerModuleTest.TestApplication::class)
+class PlatformParameterSyncUpWorkerModuleTest {
+  @Inject lateinit var context: Context
+  @Inject lateinit var readyLists: Set<@JvmSuppressWildcards StartupWorkerScheduleReadinessListener>
```
adhiamboperes: Here and a few other places: I find it odd to shorten listeners to lists, because it is not intuitive given our extensive use of Lists in the repo.

adhiamboperes (commented):
Thanks @BenHenning!

I still have a few files to review, but should be done by tomorrow. So far, I haven't come across anything egregious.

BenHenning:
Not stale.

testing/src/main/java/org/oppia/android/testing/FakeFirestoreEventLogger.kt
BenHenning: Nice catch. I copied and pasted without looking too closely it seems. Fixed here & elsewhere.

domain/src/test/java/org/oppia/android/domain/platformparameter/syncup/PlatformParameterSyncUpWorkerModuleTest.kt
BenHenning: Fair point, I was also a bit uneasy with this. I think only the '`readLists`' were impacted, so I renamed all the occurrences I could find with `readinessListeners` and line-wrapped.

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
```diff
@@ -42,48 +48,40 @@ abstract class AbstractOppiaApplication(
...
-        component.getApplicationStartupListeners().forEach(ApplicationStartupListener::onCreate)
-      }.invokeOnCompletion {
-        if (it != null) {
-          throw Exception("Failed to continue application initialization.", it)
-        }
+    // Kick off a background task to finish startup initialization. Nothing at this stage should be
+    // startup-state sensitive. It's also fine for parameters to not be fully initialized at this
+    // point because each app entry point already accounts for potentially uninitialized parameters:
+    // splash, direct activity recreation, and waking up the app to kick off a worker.
+    CoroutineScope(component.getBackgroundDispatcher()).async {
```
BenHenning: Check if `launch` would correctly crash the app in this case & avoid the awkward re-throw below.

app/src/test/java/org/oppia/android/app/utility/datetime/DateTimeUtilTest.kt
```diff
@@ -206,9 +206,9 @@ class DateTimeUtilTest {
   }
 
   @Module
-  interface AnalyticsStartupListenerTestModule {
+  interface StartupWorkerScheduleReadinessListenerTestModule {
     @Multibinds
-    fun provideAnalyticsListenerSet(): Set<AnalyticsStartupListener>
+    fun provideAnalyticsListenerSet(): Set<StartupWorkerScheduleReadinessListener>
```
BenHenning: Nit: rename provider.

app/src/test/java/org/oppia/android/app/utility/datetime/DateTimeUtilTest.kt
```diff
@@ -206,9 +206,9 @@ class DateTimeUtilTest {
   }
 
   @Module
-  interface AnalyticsStartupListenerTestModule {
+  interface StartupWorkerScheduleReadinessListenerTestModule {
```
BenHenning: Is this actually still needed? Seems very out of place for this test suite.

app/BUILD.bazel
```diff
@@ -559,6 +559,7 @@ kt_android_library(
         "//third_party:androidx_databinding_databinding-common",
         "//third_party:androidx_databinding_databinding-runtime",
         "//utility",
+        "//utility/src/main/java/org/oppia/android/util/enumfilter:enum_filter_util",
```
BenHenning, line 574: Needed?

app/BUILD.bazel
```diff
@@ -678,6 +679,7 @@ kt_android_library(
         "//third_party:org_jetbrains_kotlin_kotlin-stdlib-jdk8_jar",
         "//utility",
         "//utility/src/main/java/org/oppia/android/util/extensions:bundle_extensions",
+        "//utility/src/main/java/org/oppia/android/util/logging:current_app_screen_name_intent_decorator",
```
BenHenning, line 695: Needed?

domain/src/main/java/org/oppia/android/domain/oppialogger/exceptions/UncaughtExceptionLoggerStartupListener.kt
```diff
@@ -3,25 +3,39 @@ package org.oppia.android.domain.oppialogger.exceptions
...
 
   private var existingUncaughtExceptionHandler: Thread.UncaughtExceptionHandler? = null
+  private var canLogExceptions: Boolean = false
 
-  /** Sets up the uncaught exception handler to [UncaughtExceptionLoggerStartupListener]. */
-  override fun onCreate() {
+  override fun onCreateStarted() {
+    // This should be set up immediately to try and capture exceptions that occur early, but it may
+    // not be safe to log them until after app initialization completes.
     existingUncaughtExceptionHandler = Thread.currentThread().uncaughtExceptionHandler
```
BenHenning: NB: I noticed this in a different PR or at a different time but this isn't actually capturing for more than the main thread. This should be set up to be the default handler, not just the current thread's handler.

Also, add tests for this since it should be testable.

domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessListener.kt
```diff
@@ -0,0 +1,23 @@
...
+ * execution.
+ *
+ * Note that implementation bindings should be multi-bound into a set of this listener type in order
+ * to automatically be included in application startup procedures.
+ */
+interface StartupWorkerScheduleReadinessListener {
+  /**
+   * Called early in application startup to allow periodic work to be scheduled.
+   *
+   * Note that no guarantees can be called about when in the application initialization this is
```
BenHenning:
```suggestion
   * Note that no guarantees can be called about when in the application initialization process this is
```

Reads as missing a word.

domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessMonitor.kt
```diff
@@ -0,0 +1,33 @@
...
+ * An [ApplicationStartupListener] which monitors for when it's safe to perform `WorkManager` worker
+ * scheduling using [WorkManagerScheduler].
+ *
+ * This is an automated mechanism class and should never need to be interacted with directly, and
+ * likely rarely changed.
+ */
+@Singleton
+class StartupWorkerScheduleReadinessMonitor @Inject constructor(
+  private val workManagerScheduler: WorkManagerScheduler,
+  private val listsProv: Provider<Set<@JvmSuppressWildcards StartupWorkerScheduleReadinessListener>>
```
BenHenning: Rename this, too, per review comment.

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
```diff
@@ -608,29 +574,93 @@ class ApplicationLifecycleObserverTest {
...
+    applicationLifecycleObserver.onCompletedInitialization()
+  }
+
+  private fun initializeProcessLifecycleOwner() {
+    // Hacky way to force ProcessLifecycleOwner to initialize itself and start listening for events.
+    ProcessLifecycleOwnerInitializer().also { it.attachInfo(context, /* info= */ null) }.onCreate()
+  }
+
+  private fun resetProcessLifecycleOwner() {
+    // A *VERY* hacky way to force ProcessLifecycleOwner to reset state by recreating its internal
```
BenHenning, line 592: Add a TODO here to upgrade away from needing this hack.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -3843,16 +3843,24 @@ test_file_exemption {
   exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/OppiaLogger.kt"
   source_file_is_incompatible_with_code_coverage: true
 }
+test_file_exemption {
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessListener.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessMonitor.kt"
```
BenHenning, line 3903: Add TODO

scripts/assets/test_file_exemptions.textproto
```diff
@@ -3843,16 +3843,24 @@ test_file_exemption {
...
   override_min_coverage_percent_required: 68
 }
 test_file_exemption {
-  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/AnalyticsStartupListener.kt"
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt"
   test_file_not_required: true
 }
 test_file_exemption {
-  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleListener.kt"
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleLogger.kt"
```
BenHenning, line 3919: Add TODO.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -3896,21 +3908,13 @@ test_file_exemption {
   source_file_is_incompatible_with_code_coverage: true
 }
 test_file_exemption {
-  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorkerFactory.kt"
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/logscheduler/MetricLogSchedulingWorkerScheduler.kt"
```
BenHenning: Add TODO.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -3875,6 +3883,10 @@ test_file_exemption {
   exempted_file_path: "domain/src/main/java/org/oppia/android/domain/oppialogger/analytics/LearnerAnalyticsLogger.kt"
   source_file_is_incompatible_with_code_coverage: true
 }
+test_file_exemption {
+  exempted_file_path: "domain/src/main/java/org/oppia/android/domain/workmanager/testing/OppiaWorkManagerTestDriver.kt"
```
BenHenning, line 3943: Add TODO.

testing/src/main/java/org/oppia/android/testing/threading/BUILD.bazel
```diff
@@ -86,6 +86,7 @@ kt_android_library(
     ],
     visibility = [
         "//app:__subpackages__",
+        "//domain/src/main/java/org/oppia/android/domain/workmanager/testing:__pkg__",
```
BenHenning, line 104: Still needed? Might need to document the service's use.

tools/kotlin/remove_processor_duplicates.patch
```diff
@@ -15,3 +15,7 @@
      )
 
      compiler_plugins = [
+@@ -349,2 +349,2 @@
+     # Unwrap kotlinc_options/javac_options options or default to the ones being provided by the toolchain
+-    args.add_all("--kotlin_passthrough_flags", kotlinc_options_to_flags(kotlinc_options))
++    args.add_all("--kotlin_passthrough_flags", kotlinc_options_to_flags(kotlinc_options) + ["-Xskip-metadata-version-check"])
```
BenHenning: Revert this.

testing/src/test/java/org/oppia/android/testing/firebase/BUILD.bazel
```diff
@@ -41,5 +41,6 @@ oppia_android_test(
         "//third_party:junit_junit",
         "//third_party:org_robolectric_robolectric",
         "//third_party:robolectric_android-all",
+        "//utility/src/main/java/org/oppia/android/util/logging/firebase:debug_module",
```
BenHenning, line 44: Needed?

utility/src/main/java/org/oppia/android/util/logging/firebase/BUILD.bazel
```diff
@@ -135,6 +131,7 @@ kt_android_library(
         "FirestoreInstance.kt",
         "FirestoreInstanceWrapper.kt",
     ],
+    visibility = ["//:oppia_api_visibility"],
```
BenHenning: Needed?

BenHenning (commented):
Self-reviewed PR (though kept dense code reading at a minimum). Noted a few places I need to double check, and there are CI failures to fix and a PR description to finish (I used this self-review to take notes on what to expand upon in the PR description).

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
BenHenning: Switched to launch though I want to actually test it with a device first before resolving this.

app/src/test/java/org/oppia/android/app/utility/datetime/DateTimeUtilTest.kt
BenHenning: Removed instead.

app/src/test/java/org/oppia/android/app/utility/datetime/DateTimeUtilTest.kt
BenHenning: Nope--removed.

app/BUILD.bazel
BenHenning: Yep, I'm assuming due to some sort of transitive dep change in this PR. Keeping.

app/BUILD.bazel
BenHenning: Yup. Probably also a transitive dependency change from other changes in the PR. Keeping.

domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessListener.kt
BenHenning: Updated.

domain/src/main/java/org/oppia/android/domain/workmanager/StartupWorkerScheduleReadinessMonitor.kt
BenHenning: Done (and another place that was just inconsistent but not necessarily confusing).

domain/src/test/java/org/oppia/android/domain/oppialogger/analytics/ApplicationLifecycleObserverTest.kt
BenHenning: Done (and filed #6187 to track).

scripts/assets/test_file_exemptions.textproto
BenHenning: Filed #6189.

scripts/assets/test_file_exemptions.textproto
BenHenning: This one shouldn't even be here since there is a test. Found #6188 in the process and removed this and other such cases.

scripts/assets/test_file_exemptions.textproto
BenHenning: Filed #6190.

scripts/assets/test_file_exemptions.textproto
BenHenning: Filed #6191.

testing/src/main/java/org/oppia/android/testing/threading/BUILD.bazel
BenHenning: Still needed. Added a note to PR description to explain.

tools/kotlin/remove_processor_duplicates.patch
BenHenning: Done.

testing/src/test/java/org/oppia/android/testing/firebase/BUILD.bazel
BenHenning: `DebugLogReportingModule` is needed by the test. Probably another transitive dep change. Keeping.

utility/src/main/java/org/oppia/android/util/logging/firebase/BUILD.bazel
BenHenning: Yes, but not so broad. Updated.

domain/src/main/java/org/oppia/android/domain/oppialogger/exceptions/UncaughtExceptionLoggerStartupListener.kt
BenHenning: Updated & changed tests with Gemini help.

BenHenning (commented):
Self-reviewed fixes.

PR should be up-to-date with develop again and seems to be in a pretty good shape. Just need to finish main PR description and it should be ready for full review (assuming CI passes now which I am hoping it will).

akkicodes-dev:
Hi @BenHenning 

This refactor is really impressive — I can see how introducing the Bootstrap worker and centralizing scheduling resolves the failure modes I was exploring earlier around WorkManager’s KEEP policy and permanent sync stalls. It’s much clearer now how the new architecture avoids those dead states.

I especially found the separation of scheduling concerns and the abstraction over WorkManager interactions quite elegant — it makes the system easier to reason about and extend.

I had one question while going through the changes: with the new scheduling flow, are intermittent failures (like transient network issues) primarily handled within individual workers (via retry/backoff), or is there also a mechanism at the scheduler/Bootstrap level that ensures eventual recovery?

Happy to dig deeper into related areas or help validate edge cases if that would be useful. Really enjoyed going through this — thanks for putting together such a thorough redesign!

BenHenning:
@akkicodes-dev the current implementation expects workers (that is, `OppiaWorker`s) to be responsible for handling errors that they encounter. Otherwise, failures will bubble up to the bootstrap worker and then to `WorkManager` which will handle the worker accordingly. There's not really a need for back-off at this point, and retry will happen automatically by virtue of these being periodic jobs.

akkicodes-dev:
Got it, that clarifies things — thanks!

I was thinking in terms of additional retry/backoff handling, but this makes sense given the periodic nature of the workers and existing flow.

Appreciate the explanation!

BenHenning:
PTAL @adhiamboperes. This is now ready for full review (the PR description is finally finished!).

testing/src/test/java/org/oppia/android/testing/FakeExceptionLoggerTest.kt
```diff
@@ -140,6 +172,136 @@ class FakeExceptionLoggerTest {
     assertThat(exceptionLogStatus2).isTrue()
   }
 
+  @Test
+  fun testFakeExceptionLogger_getMostRecentExceptions_two_noExceptionsLogged_returnsEmptyList() {
+    val mostRecentExceptions = fakeExceptionLogger.getMostRecentExceptions(count = 2)
+
+    assertThat(mostRecentExceptions).isEmpty()
+  }
```
adhiamboperes, line 182: It is not clear what this test is checking for.

testing/src/test/java/org/oppia/android/testing/FakeExceptionLoggerTest.kt
```diff
@@ -140,6 +172,136 @@ class FakeExceptionLoggerTest {
...
+  @Test
+  fun testFakeExceptionLogger_getMostRecentExceptions_one_twoExceptionsLogged_returnsLatest() {
+    exceptionLogger.logException(exception2)
+    exceptionLogger.logException(exception1)
+
+    val mostRecentExceptions = fakeExceptionLogger.getMostRecentExceptions(count = 1)
+
+    assertThat(mostRecentExceptions).hasSize(1)
+    assertThat(mostRecentExceptions.single()).hasMessageThat().isEqualTo("First Exception")
+  }
```
adhiamboperes, line 214: Nit on naming -- could these and adjacent tests be better as:
`testFakeExceptionLogger_getMostRecentException_twoExceptionsLogged_returnsLatest`

and

`testFakeExceptionLogger_getTwoMostRecentExceptions_twoExceptionsLogged_returnsInOrder`

domain/src/main/java/org/oppia/android/domain/workmanager/testing/OppiaWorkManagerTestDriver.kt
```diff
@@ -0,0 +1,446 @@
...
+import java.util.UUID
+import java.util.concurrent.CopyOnWriteArraySet
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * A general-purpose, all-in-one test utility when interacting with [WorkManager] in Oppia tests.
+ *
+ * This utility contains a bunch of critical setup and interaction pathways that tests needing to
```
adhiamboperes: This sentence seems a bit broken?

adhiamboperes (changes requested):
Thanks @BenHenning!  PR LGTM mostly. PTAL at my comments.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterModule.kt
```diff
@@ -57,21 +62,38 @@ class TestPlatformParameterModule {
...
-        // hence the use of the Unconfined dispatcher.
-        return CoroutineScope(Dispatchers.Unconfined).async {
-          // TODO(#5835): Remove this blocking hack to ensure tests are properly inited for params.
-          val loadResult = prodController.loadParametersAsync()
-          testCoroutineDispatchers.runCurrent()
-          check(loadResult.isCompleted) { "Expected parameter loading to have finished." }
+        // hence the use of the dispatcher. The Main dispatcher must be used due to the runCurrent()
+        // call, however, since otherwise it can contend with tests trying to synchronize state and
+        // then deadlock. Additional work must be done to ensure that the thread hop *doesn't*
+        // happen if the main thread is already being used (since that could then cause a deadlock).
```
adhiamboperes: Is this the correct place to document this? Ln 79 looks better + some of the context belongs to `loadParamsAndRunTestDispatchersOnMainThread`

BenHenning:
Not stale.

adhiamboperes:
@coderabbitai review

BenHenning:
@coderabbitai full review

BenHenning:
Don't close.

BenHenning:
Still don’t close.

testing/src/test/java/org/oppia/android/testing/FakeExceptionLoggerTest.kt
BenHenning: It's purpose is to set up the preconditions for the following tests (to ensure they're operating in exactly the context they are expected to). I don't feel strongly about keeping it, though.

testing/src/test/java/org/oppia/android/testing/FakeExceptionLoggerTest.kt
BenHenning: I agree, these are super awkward. Part of the problem is the whole test suite is using a redundant naming convention that's a bit tricky to simplify, so I decided just to update all of the tests here to be a bit more 'modern' and instead name what method they're testing.

domain/src/main/java/org/oppia/android/domain/workmanager/testing/OppiaWorkManagerTestDriver.kt
BenHenning: Indeed, but I also reworked the section broadly to hopefully make a bit more sense and read better.

testing/src/main/java/org/oppia/android/testing/platformparameter/TestPlatformParameterModule.kt
BenHenning: Good points. I've split this up and moved each section to hopefully better locations, and added additional context to hopefully make this make more sense (which will probably be especially helpful when we have to debug or update this in the future).

app/src/main/java/org/oppia/android/app/application/AbstractOppiaApplication.kt
BenHenning: Confirmed that an exception thrown within the `launch` does correctly stop the app from opening.

BenHenning:
@coderabbitai review

BenHenning (commented):
Self-reviewed latest changes. I'm not entirely sure why maven_install.json is being updoated, actually. This feels like there was some sort of npm-esque non-hermetic change somewhere which I don't think we've ever seen before. Nevertheless, it doesn't seem particuarlly scary all things considered.

BenHenning (commented):
Self-reviewed small wiki formatting change.

BenHenning:
@adhiamboperes PTAL. I've addressed all of your comments I think (or at least replied to them). I believe that there are presently 6 open threads on the PR, all started by you. Please let me know if any further work is needed on those or related to other changes I've since made (which shouldn't be many). Apologies that I did address these comments after resolving conflicts from `develop` so the review diff might have some unrelated changes as a consequence.

I've self-reviewed the remaining code, double checked that `WorkManager` is still working correctly on a local app build, and the failure case that I noted in `AbstractOppiaApplication` (the app correctly crashes & fails to start up if one of the startup listeners throws in its second initialization method). I've also needed to undo some of the wiki page formatting due to a limitation in the ToC correctness check (it doesn't handle special characters like backticks for code or question marks correctly). I decided not to actually fight it and just undid the formatting, instead. This PR is plenty large enough without more unrelated changes. :)

Please let me know if I need to do anything further here or if the remaining conversations can be resolved & the PR approved for merging.

BenHenning (commented):
Self reviewed, and brought this up-to-date with develop to pull in #6334 since that will be a big help with this PR's CI.

Note the change to `scripts/src/java/org/oppia/android/scripts/telemetry/DecodeUserStudyEventString.kt` @adhiamboperes. This was needed because #6299 likely broke this utility (since we don't have tests for it). As far as I can reason, the method call replacement seems correct per my read of https://javadoc.io/doc/com.google.protobuf/protobuf-java-util/latest/com/google/protobuf/util/JsonFormat.Printer.html.

The latest changes hopefully will bring this PR's CI into fully passing state.

adhiamboperes (approved):
Everything LGTM, thanks @BenHenning.

adhiamboperes:
@seanlip, PTAL for owners. As far as I can tell, only the `wiki/Work-Manager.md` and `AndroidManifest` files require owners approval.

BenHenning:
Thanks @adhiamboperes! Enabling auto-merge for whenever you can approve @seanlip.

seanlip (approved):

Issue #5935, opened 2025-08-26
[BUG]: Submitted answer and Feedback text briefly flash incorrect response before updating to actual feedback


subhajitxyz:
### Describe the bug

There is unexpected behavior with the submitted answer and feedback text display in the exploration player.
When submitting an answer (e.g., "Red"):

- The submitted answer initially flashes "Eagle" for a short moment before updating to the correct submitted answer.
- The feedback also initially flashes "Correct! Eagles can sustain flight" for a brief moment before updating to the correct feedback.

It appears that the player is briefly showing the previous card’s last submitted answer and feedback before updating.

### Steps To Reproduce

- Start Topic: First Test Topic
- Click Prototype Exploration
- Reach to this question "What color does the 'G' in 'RGB' correspond to?"
- Submit an answer and observe that both the submitted answer and feedback briefly flash the previous card’s response before updating.

### Expected Behavior

The player should directly show the correct submitted answer and feedback without flashing the previous card’s response.

### Screenshots/Videos

For better understanding, the attached video has been slowed down to 0.2x speed to make the issue more visible.

https://github.com/user-attachments/assets/c97bcf3c-789e-4d64-b97d-96dcf9364d75

### What device/emulator are you using?

Oppo reno 2f

### Additional Context

**Files to investigate:**

- [feedback_item.xml](https://github.com/oppia/oppia-android/blob/96af5e5f11dbf856eef5be9f0ccf264fde4491d6/app/src/main/res/layout/feedback_item.xml#L1), [submitted_answer_item.xml](https://github.com/oppia/oppia-android/blob/96af5e5f11dbf856eef5be9f0ccf264fde4491d6/app/src/main/res/layout/submitted_answer_item.xml#L1) and corresponding viewmodels.

- Check the following function implementations in [StatePlayerRecyclerView](https://github.com/oppia/oppia-android/blob/96af5e5f11dbf856eef5be9f0ccf264fde4491d6/app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt#L150): `addPastAnswersSupport()`, `addFeedbackSupport()`

rezaul-web:
Hello! As a new contributor, I would like to work on this issue. Could you please assign it to me

shrimpnaur:
Hello @subhajitxyz Can I work on this issue?

adhiamboperes:
@shrimpnaur, yes you can

rajeev-k47:
[...]

Issue #3059, opened 2021-04-08
Add tests for BindableAdapters in  ImageViewBindingAdapters


rt4914:
### Is your feature request related to a problem? Please describe.

Currently, [ImageViewBindingAdaptersTest](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/app/src/sharedTest/java/org/oppia/android/app/databinding/ImageViewBindingAdaptersTest.kt#L101) only contains tests for one bindable adapter:

```java
@BindingAdapter("playStateImage")
public static void setPlayStateDrawable(
    @NonNull ImageView imageView,
    ChapterPlayState chapterPlayState
)
```

However, other bindable adapters defined in [ImageViewBindingAdapters](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java#L19) are not yet tested. This results in incomplete coverage and may allow regressions in critical UI logic.

### Describe the solution you'd like

We should expand `ImageViewBindingAdaptersTest` to include dedicated tests for **all bindable adapters in `ImageViewBindingAdapters`**. Each test should verify that the binding adapters behave as expected under different inputs, using Espresso matchers and Robolectric where applicable.

### Tasks / Checklist

Add new tests for the following bindable adapters:

1. **`setImageDrawableCompat(ImageView, @DrawableRes int)`**

     - Verify that the correct drawable resource is set on the `ImageView`.

2. **`setImageDrawableCompat(ImageView, Drawable)`**

     - Verify that a provided `Drawable` object is applied correctly.

3. **`setProfileImage(ImageView, ProfileAvatar)`**

     - Test behavior for `AVATAR_COLOR_RGB` → default avatar with color filter applied.
     - Test behavior for `AVATAR_IMAGE_URI` → correct image loaded with placeholder.
     - Test null `ProfileAvatar` case → `ImageView` should not crash and fallback gracefully.

4. **`setTint(ImageView, int)`**

     - Verify that the correct color filter is applied to the `ImageView`.

5. **`setPlayStateDrawable(ImageView, ChapterPlayState)`**

     - Already partially tested, but ensure coverage for all possible `ChapterPlayState` values.
     - Validate that the fallback/default drawable is correctly applied when no match is found.

### Acceptance Criteria

[...]

PR #1627, opened 2020-08-12
Fix #1625, #1626: Revert "Fix #171: Introduce support for diffing subsets of data in generic RecyclerView adapter"

BenHenning:
Reverts oppia/oppia-android#1359.

Fixes #1625.
Fixes #1626.

Issue #171, opened 2019-09-20
Introduce support for diffing subsets of data in generic RecyclerView adapter


BenHenning:
The initial implementation of the generic RecyclerView BindableAdapter notifies that the whole data set is changed for each re-bind, which is sometimes wrong (the data may be the same) and highly inefficient (1 element might have changed out of hundreds). Diffing logic should be introduced to only notify that a subset of the data has changed, which also enables support for RecyclerView animations.

Later update for much more context on approaching this issue:

We've tried two attempts at fixing this:
- #1359 that was actually merged, but then reverted in #1627 due to causing #1625 and #1626. As far as I know, we never looked into the specifics of what caused the performance issues here but I have some suspicions (see below)
- #2666 there were a bunch of test failures that weren't fixed, and more manual testing was needed. It also wasn't quite implemented correctly as we want to use more granular controls than just ``notifyDataSetChanged``.

To illustrate the current state and desired outcome (for added context):
- Today, ``BindableAdapter`` has a single ``setData`` method that replaces the entire list of data: https://github.com/oppia/oppia-android/blob/4f5aec630c8b4826cd94fe63bc2ab99f0cb35ad7/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt#L42-L48 This is actually very convenient because it ties directly into databinding. However, it is very inefficient: calling [``notifyDataSetChanged``](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.Adapter#notifyDataSetChanged()) which is a bit of a sledgehammer solution since it basically forces the recycler view to reset its views (which causes issues like #4839 plus other annoyances like view flashing and prevents clean automatic built-in recycler view animations).
- Ideally, we would actually use the more granular adapter functions for notifying changes: ``notifyItemChanged``, ``notifyItemInserted``, and ``notifyItemRemoved`` (possibly also ``notifyItemRangeChanged``, ``notifyItemRangeInserted``, and ``notifyItemRangeRemoved`` though these are less important and probably harder to do) so that ``RecyclerView`` knows how to manage incoming data changes correctly.
- Since we can't easily add new ``set*()`` methods in ``BindableAdapter`` for this purpose, we need to rely on diffing to understand the difference between the current set of data bound to the recycler view and the new list that's incoming. ``DiffUtil`` performs this computation. It has a benchmark listed that we could try to beat, but I'm not sure that it's necessary. I think we should try to use ``DiffUtil`` first, but correctly.
[...]

PR #6063, opened 2026-01-02
Fix #5959 Profiles Flicker issue

### LINKED ISSUE #5959: [BUG]: Profile colors and positions change when selecting a profile in the Profile screen
### Describe the bug

When selecting a profile from the "Select your profile" screen, the profile list slightly shifts and some profile colors change unexpectedly.

### Steps To Reproduce

- Open the Oppia app.
- Go to the "Select your profile" screen.
- Set up multiple profiles (more than 5).
- Scroll down and click on any profile near the bottom of the list.
- Observe that the profiles slightly shift from their original position and profile colors change.

### Expected Behavior

Profile positions and colors should remain consistent when selecting a profile.

### Screenshots/Videos

https://github.com/user-attachments/assets/5b91a4b7-1a96-4f34-ac61-7fe639506d2e

Oppo reno 2f

Android 11

_No response_

### Additional Context

**Cause:**
When a profile is clicked, `onProfileClick() `updates the learner ID if it is absent. This causes the `LiveData` holding the profile list to emit a new list. Since the current adapter uses `notifyDataSetChanged()`, the entire RecyclerView reloads all items. As a result, the list briefly shows flickering, incorrect colors, and temporary item shifts.

**Proposed Solution:**
Introduce `DiffUtil`in `RecyclerViewBindingAdapter` for the `profile_recycler_view`. Since each profile has a unique id, we can use it to efficiently calculate differences and update only the changed items. This will eliminate flickering and keep profile colors stable during list updates.

**Files to Check:**
- [ProfileChooserFragmentPresenter](https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/app/src/main/java/org/oppia/android/app/profile/ProfileChooserFragmentPresenter.kt#L70C7-L70C38) -> funs `onProfileClick()`, `updateLearnerIdIfAbsent()`, `computeLoginRoute()`
- [ProfilemanagementController](https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/domain/src/main/java/org/oppia/android/domain/profile/ProfileManagementController.kt#L94) -> funs `initializeLearnerId()`
- [ProfileChooserViewModel](https://github.com/oppia/oppia-android/blob/fcb87dc79074450f90a658c9dddb186a12ab75f3/app/src/main/java/org/oppia/android/app/profile/ProfileChooserViewModel.kt#L22) -> fun `retrieveProfiles()`
- [RecyclerViewBindingAdapter](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/app/src/main/java/org/oppia/android/app/recyclerview/RecyclerViewBindingAdapter.java#L20)
- [BindableAdapter](https://github.com/oppia/oppia-android/blob/3ffeedf7e90795ae9b23ec600a19d7cff72f1ca4/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt#L42)

**Personal Note:**
I personally tried using `DiffUtil` for the `profile_recycler_view`, and it solved the problem. After applying it, the list updates smoothly without shifts or temporary wrong colors.

Output:

https://github.com/user-attachments/assets/64d1048c-fe56-44b8-9786-a3279350640d

itsadityaaaaa:
## Explanation
  - "Fixes #5959 " in the explanation so that GitHub can auto-close the issue
 -  Reuse view holders for items with the same stable ID, preventing full rebinds that cause UI flicker.
  - when this PR is merged - Cleaner and Smoother UI transitions during Profile Selection
  
  

## Essential Checklist

## For UI-specific PRs only
- Mobile 

https://github.com/user-attachments/assets/3e294808-df92-46ff-ab3e-6bcf2e2e3142

-Tablet

https://github.com/user-attachments/assets/9a91995f-8559-46c6-8dcf-f71065c231c5

 - BindableAdapterTest Result 
 
Screenshot from 2026-01-02 22-09-49

### DIFF
--- a/app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
+++ b/app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
@@ -7,10 +7,6 @@ import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 import androidx.databinding.BindingAdapter;
 import com.bumptech.glide.Glide;
-import com.bumptech.glide.load.DataSource;
-import com.bumptech.glide.load.engine.GlideException;
-import com.bumptech.glide.request.RequestListener;
-import com.bumptech.glide.request.target.Target;
 import org.oppia.android.app.databinding.adapters.R;
 import org.oppia.android.app.model.ChapterPlayState;
 import org.oppia.android.app.model.ProfileAvatar;
@@ -44,34 +40,16 @@ public final class ImageViewBindingAdapters {
   public static void setProfileImage(ImageView imageView, ProfileAvatar profileAvatar) {
     if (profileAvatar != null) {
       if (profileAvatar.getAvatarTypeCase() == ProfileAvatar.AvatarTypeCase.AVATAR_COLOR_RGB) {
-        Glide.with(imageView.getContext())
-            .load(R.drawable.ic_default_avatar)
-            .listener(new RequestListener<Drawable>() {
-              @Override
-              public boolean onLoadFailed(
-                  GlideException e,
-                  Object model,
-                  Target<Drawable> target,
-                  boolean isFirstResource) {
-                return false;
-              }
-
-              @Override
-              public boolean onResourceReady(
-                  Drawable resource,
-                  Object model,
-                  Target<Drawable> target,
-                  DataSource dataSource,
-                  boolean isFirstResource
-              ) {
-                imageView.setColorFilter(
-                    profileAvatar.getAvatarColorRgb(),
-                    PorterDuff.Mode.DST_OVER
-                );
-                return false;
-              }
-            }).into(imageView);
+        // Clear any pending Glide request from a recycled ViewHolder and load the default avatar
+        // synchronously.
+        Glide.with(imageView.getContext()).clear(imageView);
+        imageView.setImageResource(R.drawable.ic_default_avatar);
+        imageView.setColorFilter(
+                profileAvatar.getAvatarColorRgb(),
+                PorterDuff.Mode.DST_OVER
+        );
       } else {
+        imageView.clearColorFilter();
         Glide.with(imageView.getContext())
             .load(profileAvatar.getAvatarImageUri())
             .placeholder(R.drawable.ic_default_avatar)
--- a/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
+++ b/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
@@ -211,6 +211,7 @@ class BindableAdapter<T : Any> internal constructor(
             // Attaching lifecycleOwner before view model initialization can sometimes cause a
             // NullPointerException because data might not be attached to the views yet.
             binding.lifecycleOwner = lifecycleOwner
+            binding.executePendingBindings()
           }
         }
       }
@@ -337,6 +338,7 @@ class BindableAdapter<T : Any> internal constructor(
             // Attaching lifecycleOwner before view model initialization can sometimes cause a
             // NullPointerException because data might not be attached to the views yet.
             binding.lifecycleOwner = lifecycleOwner
+            binding.executePendingBindings()
           }
         }
       }


adhiamboperes:
@BenHenning, I understand that there have been previous attempts to use Diff Utils that caused some issues, i.e. linked PRs in https://github.com/oppia/oppia-android/issues/171 and https://github.com/oppia/oppia-android/pull/1627. Should we explore other possible solutions to this issue?

BenHenning:
Apologies--I didn't realize this was assigned to me.

We discussed this briefly in the team meeting this week @itsadityaaaaa. As @adhiamboperes mentioned we have tried diff utils multiple times in the past--it's very hard to get right particularly for the core lesson player. I'm not sure we should have suggested this as the main fix for #5959 actually, which in turn means that's not quite a starter issue.

Beyond that, I think the main problem here is we need understand _why_ the flickering is happening. The profile colors are supposed to be fixed within a user's `Profile` proto--we only generate colors upon profile creation. At a high-level this suggests a few possibilities (though I'm sure there are many more) for why the bug may be happening:
- There's some sort of race condition between the list updating, the recycler view partially re-rendering, and the page changing.
- There's some sort of corruption or mismanagement happening for the list of profiles caused by the late update during screen transition.
- Recycler view is actually binding the wrong view data (i.e. an issue in the recycler view pipeline or how we bind it rather than the actual data we're binding _to_ it).

I think we need to figure out what's actually wrong at the data level, then work backwards to understand why it's wrong. This is what I recommend you do:
1. Either step debug or use 'print/logcat debugging' to print out the exact list of profile data being bound at each step. See what the list is before selecting a profile, then compare that to the list after selecting (that ends up showing up wrong). What are the differences? This will narrow down if it's a data representation, binding, or rendering problem.
2. Once we understand which category of problem it is, we need to ask if the data is correct at that stage. If it is, we have to work to do in how we're managing it (i.e. rendering or binding). If it isn't, we need to go backwards and figure out how the data came to be that way (data management). This involves stepping or more logcat logs earlier in the dataflow for the profile list to try and figure out at which point is the wrong data being created and propagated.
3. Follow up with your findings. If you are unable to determine the _why_ then document as much of the _what_ as you can (i.e. what steps you took and the observed outcomes, and what seems wrong or right about them).

From there, one of us can probably provide further advice or specific suggestions.

Note that it may be the case that diff util is still the best way to fix the issue, but we need to understand exactly why it's broken before we can be confident in that solution. Also if that does happen, we'll want to update the binding adapter to support enabling diff utils since we'll really only want it enabled for the profile list for now (to avoid the bugs elsewhere in the app that it will cause).

itsadityaaaaa:
@BenHenning @adhiamboperes Thank you for your suggestions, I tried to use stable ID's to fix the flickering. It reuse view holders for items with the same stable ID, preventing full rebinds that cause UI flicker. I build and investigate and found It patches the flickering, Is it the correct approach or should I have to investigate further?

BenHenning:
> @BenHenning @adhiamboperes Thank you for your suggestions, I tried to use stable ID's to fix the flickering. It reuse view holders for items with the same stable ID, preventing full rebinds that cause UI flicker. I build and investigate and found It patches the flickering, Is it the correct approach or should I have to investigate further?

@itsadityaaaaa please take a closer look at my earlier message. Finding a fix without understanding the solution is an inherently risky strategy (see https://en.wikipedia.org/wiki/Root-cause_analysis for why root-cause analysis is a preferred methodology). We actually need to understand what the root cause is to be sure the diff utils solution not only is actually correct (logically, not just observationally) and that it's the _best_ solution we can take (considering it has fairly significant potential drawbacks without a fully thought-out solution).

To that end, your next steps should be to work through the items in my previous comment and write up a debugging document: https://github.com/oppia/oppia-android/wiki/Debugging-Docs.

itsadityaaaaa:
@BenHenning Thanks for your guidance I tried to find the root cause and it seems like it causes due to full rebind. Please take a look at this doc https://drive.google.com/file/d/1G9Tyi7odQenHtqqQ5DPsrjUeRYW0p2AL/view?usp=sharing, I tried to use DiffUtil as a potential solution earlier which leads to many pre-written UI tests failure.

BenHenning:
> @BenHenning Thanks for your guidance I tried to find the root cause and it seems like it causes due to full rebind. Please take a look at this doc https://drive.google.com/file/d/1G9Tyi7odQenHtqqQ5DPsrjUeRYW0p2AL/view?usp=sharing, I tried to use DiffUtil as a potential solution earlier which leads to many pre-written UI tests failure.

Thanks @itsadityaaaaa. Can you create a debugging document, start filling that out, and share it? It's important to also collaborate on the steps to debug, not just findings.

In this case I think we actually know already that this is a full rebind situation since that's how `BindableAdapter` works today. What's unclear is why the full rebind causes the specific issue with the colors changing--that's what needs to be figured out. It may well be the case that diff utils, by avoiding a full rebind, are just sidestepping the issue rather than fixing its real root cause (which we won't be able to know without understanding why exactly the full rebind is causing problems).

itsadityaaaaa:
@BenHenning Thanks again for your guidance. I’ve put together a debugging document and wanted to confirm the approach https://docs.google.com/document/d/1CkVmUMNSJeCQ0mXIPxIedZRqGKKeBSwx3FERdsn2iy0/edit?usp=sharing

 I found out BindingAdapter in ImageViewBindingAdapters.java was not resetting ImageView state, so recycled views could briefly retain stale color filters until the async Glide load completed, causing the flicker. I implemented a minimal fix that makes the adapter idempotent seems like it solves the issue.

Does this seem like the right direction, or would I have to prefer a different approach?

itsadityaaaaa:
@BenHenning Apologise for my earlier comment, I have tested it further and found out the flickering doesn't happen until we have more than 5 ID's(with or without changes). I had tested it with 5 or lesser ID's so it seems the approach work but now I think I have to find other possible causes.

BenHenning:
Thanks @itsadityaaaaa this is a really helpful investigation, and thanks especially for the doc.

Just to check: does the state resetting fix _not_ work for cases when there are more than 5 profiles? I wasn't completely clear on your most recent comment.

I think this is definitely the right direction to continue the investigation, however. It's curious if 5 is exactly the threshold where different behavior kicks in, perhaps triggering some sort of recycling logic in the `RecyclerView`? That hypothesis could be tested by temporarily disabling recycling so that the `RecyclerView` just acts like a `ListView` to see if it changes the behavior at all.

itsadityaaaaa:
@BenHenning thanks for your suggestion turning off recycler view solves the bug. Should I go ahead with this exact change, or do you want it done differently? Like this logic to live centrally  or should the reset be handled individually inside each binder?

BenHenning:
As discussed in the team meeting, you can try to disable routing to the main activity to see if this is a race condition in the rebinding flow or a bug in us not properly rebinding @itsadityaaaaa.

BenHenning:
Also note the discussion happening in #6093 @itsadityaaaaa. It's possible that there's a general fix here that solved both #5959 and #5935 (plus probably other issues).

itsadityaaaaa:
commenting to remove the stale tag

itsadityaaaaa:
Hi @BenHenning  I have further investigated the root cause of this issue. Here's my analysis:

Root Cause
The flicker (specifically the color swap between profile avatars) is caused by two layers of deferred/async rendering during notifyDataSetChanged():

1. Data-binding defers updates to the next frame
When onBindViewHolder() calls setViewModel(binding, data), Android data-binding does not execute the binding expressions immediately  it schedules them for the next layout pass. This means when a ViewHolder is recycled from position A to position B, the stale data from position A is still rendered for one frame before the new binding expressions evaluate.

2. Glide's async pipeline for local drawables
setProfileImage() in ImageViewBindingAdapters.java uses Glide to load R.drawable.ic_default_avatar (a local resource) and applies the color filter asynchronously in onResourceReady. Even from Glide's memory cache, the callback is posted to the next frame. So even after fixing (1), the color filter from the previous profile persists for one frame on a recycled ViewHolder.

Please let me known if it seems the correct cause of flickering, I have tested this fixes and no flickering was observed.

I will update this and my previous finding related this issue shortly in the debug doc.

itsadityaaaaa:
@adhiamboperes please re-run the CI and it will great if you help in review.

BenHenning:
Done (for CI) @itsadityaaaaa.

Edit: Actually no. I am repeatedly facing 500 errors trying to restart these workflows. :/

Edit 2: Actually can restart now.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
```diff
@@ -44,34 +40,16 @@ public static void setImageDrawableCompat(
...
-                imageView.setColorFilter(
-                    profileAvatar.getAvatarColorRgb(),
-                    PorterDuff.Mode.DST_OVER
-                );
-                return false;
-              }
-            }).into(imageView);
+        // Clear any pending Glide request from a recycled ViewHolder and load the default avatar
+        // synchronously.
+        Glide.with(imageView.getContext()).clear(imageView);
```
BenHenning, line 45: Is clearing necessary?

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
```diff
@@ -44,34 +40,16 @@ public static void setImageDrawableCompat(
...
+        imageView.setImageResource(R.drawable.ic_default_avatar);
+        imageView.setColorFilter(
+                profileAvatar.getAvatarColorRgb(),
+                PorterDuff.Mode.DST_OVER
+        );
       } else {
+        imageView.clearColorFilter();
```
BenHenning, line 52: Why the clear color filter?

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
```diff
@@ -44,34 +40,16 @@ public static void setImageDrawableCompat(
...
         Glide.with(imageView.getContext())
             .load(profileAvatar.getAvatarImageUri())
```
BenHenning, line 54: Have you tested when one of the avatars has a loaded avatar image (i.e. selected from their photos)? That might still cause issues potentially. It's worth checking that with and without the fixes, I think, to see how it's impacted.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
```diff
@@ -44,34 +40,16 @@ public static void setImageDrawableCompat(
...
-              }
-            }).into(imageView);
+        // Clear any pending Glide request from a recycled ViewHolder and load the default avatar
+        // synchronously.
+        Glide.with(imageView.getContext()).clear(imageView);
+        imageView.setImageResource(R.drawable.ic_default_avatar);
+        imageView.setColorFilter(
+                profileAvatar.getAvatarColorRgb(),
+                PorterDuff.Mode.DST_OVER
+        );
```
BenHenning, line 50: If this works the same way (and I don't see why it wouldn't) it _definitely_ makes sense to do this synchronously. I'm not actually sure why we used Glide originally for this bit.

app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
```diff
@@ -211,6 +211,7 @@ class BindableAdapter<T : Any> internal constructor(
             // Attaching lifecycleOwner before view model initialization can sometimes cause a
             // NullPointerException because data might not be attached to the views yet.
             binding.lifecycleOwner = lifecycleOwner
+            binding.executePendingBindings()
```
BenHenning, line 214: This is a bit tricky but I will need to hold off on approving this change until #6088 makes a bit more progress (since the two PRs are overlapping right now).

Just to check: are _both_ this & the changes in the adapters file needed? I.e. if you remove either of them then the issue regresses?

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
```diff
@@ -44,34 +40,16 @@ public static void setImageDrawableCompat(
   public static void setProfileImage(ImageView imageView, ProfileAvatar profileAvatar) {
```
BenHenning, line 40: We should, if possible, try to add tests for the fix here.

BenHenning (changes requested):
Thanks @itsadityaaaaa! I have some thoughts--PTAL.

BenHenning:
(unassigned @adhiamboperes since I think I'm fully looking at this PR now)

app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
itsadityaaaaa: @BenHenning Thanks, for the review and Yes, I tested both changes independently. Flickering occurred when the change was applied to only one file. However, the type of flickering was different depending on which file was modified either BindableAdapter or ImageViewBindingAdapter.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
itsadityaaaaa: Now, I have tested it with loaded avatar images. Here is My findings
without fix - Flickering Observed
with fix - No flickering was observed 
Note - I will add a video demonstrating no flicking with and without loaded avatar images.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
itsadityaaaaa: It is added as defensive correctness if a ViewHolder were recycled from color  URI, the stale color filter would tint the loaded photo. But I havetested it and think its not needed, and the original code didn't have it, so I will remove it to keep the diff minimal.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
itsadityaaaaa: No, However I added this as a defensive correctness. I have tested and found we actually don't require this. I will remove this.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
itsadityaaaaa: I think existing test file acknowledges missing coverage (TODO #3059), and that the specific fix (sync vs async rendering) addresses a visual timing issue that's not reproducible in Robolectric tests so i think right now it is not possible to test the flickering however if you wanted i can write test for null avatar, color avatar sets drawable + color filter

itsadityaaaaa:
@BenHenning First of all I am very sorry  I think I have ran into some conflicts on this branch earlier which leads to i have to use "force with lease " command and i think that was also not allowed on the Oppia repo. So, I don't have any option other than closing this and opening a new PR. I understand it will become difficult for you to follow-up the discussion, I am very sorry for this. I will take care of this in future.

BenHenning:
@itsadityaaaaa please add links to each of the open comment threads from this PR in a reply to the new PR so that it’s easy to go back to them.

app/src/main/java/org/oppia/android/app/databinding/ImageViewBindingAdapters.java
itsadityaaaaa: Yeah I also think the same.

PR #6082, opened 2026-01-26
Fix #5818: Robolectric edittext input action

### LINKED ISSUE #5818: [BUG]: Fix Robolectric append/replaceText handling in EditTextInputAction and add corresponding tests.
### Describe the bug

This issue tracks the bug fix and upcoming tests related to `EditTextInputAction` used in test infrastructure for EditText input.

### 1. The Issue

- [x]  There was a bug in the `EditTextInputAction` class where appending or replacing text didn’t behave correctly in Robolectric environments. Specifically: (addressed in [PR #5617 ](https://github.com/oppia/oppia-android/pull/5617)) 
- `appendText` does not work correctly as it replaces the text rather than appending it.

- Both actions incorrectly called `EditText.setText(text)`, meaning that even when `appendText()` was used, it replaced the entire content instead of appending to it.
- This issue was masked by passing transitive test cases, but broke the intended behavior of append vs. replace differentiation.
- The logic for handling `appendText()` and `replaceText()` needed to differentiate between Robolectric and Espresso environments.

---

### 2. Adding Corresponding Tests

- [ ] Tests still need to be added to ensure the behavior of `EditTextInputAction` remains correct across test environments.

To achieve this, we should:
- Add a dedicated test class (similar to the one for [TextInputAction](https://github.com/oppia/oppia-android/blob/d25fe1f0f29688f5ed485503bc0a9c4bfff2d73f/testing/src/main/java/org/oppia/android/testing/espresso/TextInputAction.kt))
- Write Robolectric-based tests verifying `appendText()` and `replaceText()` apply the correct text values
- Validate that the logic paths in the updated `updateText()` method execute correctly for both `isAppend=true` and `isAppend=false`

### Steps To Reproduce

1. Write a Robolectric test using `EditTextInputAction.appendText()` to input text into an `EditText`.
2. Set initial text in the `EditText` (e.g., "123").
3. Use `appendText("45")` and expect the final text to be "12345".
4. Run the test in Robolectric.
5. Observe that the text becomes "45" instead of "12345", indicating that `appendText()` is replacing rather than appending the text.

Concered file: [testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt](https://github.com/oppia/oppia-android/blob/d25fe1f0f29688f5ed485503bc0a9c4bfff2d73f/testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt#L22)

### Expected Behavior

When using `EditTextInputAction.appendText()` in Robolectric-based tests, the provided text should be appended to the existing contents of the `EditText`, preserving the original text.

Similarly, when using `replaceText()`, the original text in the `EditText` should be completely replaced with the new input.

These behaviors should match how Espresso handles `typeText()` and `replaceText()` on physical/emulated devices.

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

_No response_

Sandesh282:
Fix #5818: Roboelectric EditText input action

## Explanation

Fixes #5818: Refactors `EditTextInputAction` to remove its dependency on constructor injection of
`TestCoroutineDispatchers`, which caused Robolectric edit text input actions to be difficult
to invoke reliably in tests.

This change introduces `TestCoroutineDispatchersInjector`, which provides a static access
point for test coroutine dispatchers. `EditTextInputAction` is updated to retrieve
dispatchers through this injector instead of relying on injected fields.

As a result, `EditTextInputAction` methods can now be invoked directly from tests without
requiring `@Inject` setup, resolving the existing TODO in the file and simplifying test
usage across Robolectric-based tests.

### Key changes
- Added `TestCoroutineDispatchersInjector` to hold a static reference to
  `TestCoroutineDispatchers`.
- Updated `TestDispatcherModule` to initialize the injector.
- Refactored `EditTextInputAction` to access dispatchers via the static injector.
- Updated test usages to call `EditTextInputAction` methods directly.
- Removed `EditTextInputAction` injection from test classes.

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (or "Fix part of #bugnum: ..." where applicable).
- [x] Any changes to [`scripts/assets`](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.  
      *(Not applicable — no `scripts/assets` changes.)*

## For UI-specific PRs only

This PR does not introduce any UI-related changes. It only affects test infrastructure and
Robolectric-based test utilities. As a result, no UI verification, screenshots, or accessibility
checks are required.

### DIFF
--- a/app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
+++ b/app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
@@ -66,7 +66,6 @@ import org.oppia.android.domain.question.QuestionModule
 import org.oppia.android.domain.workmanager.WorkManagerConfigurationModule
 import org.oppia.android.testing.TestImageLoaderModule
 import org.oppia.android.testing.TestLogReportingModule
-import org.oppia.android.testing.espresso.EditTextInputAction
 import org.oppia.android.testing.firebase.TestAuthenticationModule
 import org.oppia.android.testing.platformparameter.TestPlatformParameterModule
 import org.oppia.android.testing.robolectric.RobolectricModule
@@ -100,7 +99,6 @@ import javax.inject.Singleton
 class TextInputLayoutBindingAdaptersTest {
   @Inject lateinit var context: Context
   @Inject lateinit var testCoroutineDispatchers: TestCoroutineDispatchers
-  @Inject lateinit var editTextInputAction: EditTextInputAction
 
   @Before
   fun setUp() {
--- a/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
+++ b/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
@@ -1,27 +1,19 @@
 package org.oppia.android.testing.threading
 
 /**
- * A test-only injector to retrieve [TestCoroutineDispatchers] statically.
- *
- * This is used to avoid injecting [TestCoroutineDispatchers] into utility classes that should be
- * usage-site agnostic.
+ * A static injector for [TestCoroutineDispatchers] to allow them to be accessed in some static
+ * contexts (like [EditTextInputAction]).
  */
 object TestCoroutineDispatchersInjector {
-  private var testCoroutineDispatchers: TestCoroutineDispatchers? = null
+  private lateinit var dispatchers: TestCoroutineDispatchers
 
-  /**
-   * Initializes the static [TestCoroutineDispatchers] instance.
-   *
-   * This should only be called by [TestDispatcherModule].
-   */
+  /** Initializes the injector with the provided [dispatchers]. */
   fun initialize(dispatchers: TestCoroutineDispatchers) {
-    testCoroutineDispatchers = dispatchers
+    this.dispatchers = dispatchers
   }
 
-  /** Returns the current [TestCoroutineDispatchers]. */
+  /** Returns the injected [TestCoroutineDispatchers]. */
   fun getDispatcher(): TestCoroutineDispatchers {
-    return checkNotNull(testCoroutineDispatchers) {
-      "TestCoroutineDispatchers not initialized. Ensure the test environment is set up correctly."
-    }
+    return dispatchers
   }
 }
--- a/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
+++ b/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
@@ -1,0 +1,27 @@
+package org.oppia.android.testing.threading
+
+/**
+ * A test-only injector to retrieve [TestCoroutineDispatchers] statically.
+ *
+ * This is used to avoid injecting [TestCoroutineDispatchers] into utility classes that should be
+ * usage-site agnostic.
+ */
+object TestCoroutineDispatchersInjector {
+  private var testCoroutineDispatchers: TestCoroutineDispatchers? = null
+
+  /**
+   * Initializes the static [TestCoroutineDispatchers] instance.
+   *
+   * This should only be called by [TestDispatcherModule].
+   */
+  fun initialize(dispatchers: TestCoroutineDispatchers) {
+    testCoroutineDispatchers = dispatchers
+  }
+
+  /** Returns the current [TestCoroutineDispatchers]. */
+  fun getDispatcher(): TestCoroutineDispatchers {
+    return checkNotNull(testCoroutineDispatchers) {
+      "TestCoroutineDispatchers not initialized. Ensure the test environment is set up correctly."
+    }
+  }
+}
--- a/testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
+++ b/testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
@@ -38,7 +38,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        appendText("45").perform(null, editText)
+        EditTextInputAction.Companion.appendText("45").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("12345")
       }
@@ -52,7 +52,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("")
 
-        appendText("hello").perform(null, editText)
+        EditTextInputAction.Companion.appendText("hello").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("hello")
       }
@@ -66,7 +66,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        appendText("").perform(null, editText)
+        EditTextInputAction.Companion.appendText("").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("123")
       }
@@ -80,7 +80,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("abc")
 
-        appendText("@#$").perform(null, editText)
+        EditTextInputAction.Companion.appendText("@#$").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("abc@#$")
       }
@@ -94,7 +94,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("Hello ")
 
-        appendText("\uD83C\uDF0D").perform(null, editText)
+        EditTextInputAction.Companion.appendText("\uD83C\uDF0D").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("Hello \uD83C\uDF0D")
       }
@@ -108,7 +108,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        replaceText("9").perform(null, editText)
+        EditTextInputAction.Companion.replaceText("9").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("9")
       }
@@ -122,7 +122,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("old text")
 
-        replaceText("").perform(null, editText)
+        EditTextInputAction.Companion.replaceText("").perform(null, editText)
 
         assertThat(editText.text.toString()).isEmpty()
       }
@@ -136,7 +136,8 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("old text")
 
-        replaceText("\u0645\u0631\u062D\u0628\u0627").perform(null, editText)
+        EditTextInputAction.Companion.replaceText("\u0645\u0631\u062D\u0628\u0627")
+          .perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("\u0645\u0631\u062D\u0628\u0627")
       }
--- a/testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
+++ b/testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
@@ -9,6 +9,8 @@ import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.oppia.android.testing.TextInputActionTestActivity
+import org.oppia.android.testing.espresso.EditTextInputAction.Companion.appendText
+import org.oppia.android.testing.espresso.EditTextInputAction.Companion.replaceText
 import org.oppia.android.testing.threading.TestCoroutineDispatchers
 import org.oppia.android.testing.threading.TestCoroutineDispatchersInjector
 import org.robolectric.annotation.Config
@@ -38,7 +40,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        EditTextInputAction.Companion.appendText("45").perform(null, editText)
+        appendText("45").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("12345")
       }
@@ -52,7 +54,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("")
 
-        EditTextInputAction.Companion.appendText("hello").perform(null, editText)
+        appendText("hello").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("hello")
       }
@@ -66,7 +68,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        EditTextInputAction.Companion.appendText("").perform(null, editText)
+        appendText("").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("123")
       }
@@ -80,7 +82,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("abc")
 
-        EditTextInputAction.Companion.appendText("@#$").perform(null, editText)
+        appendText("@#$").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("abc@#$")
       }
@@ -94,7 +96,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("Hello ")
 
-        EditTextInputAction.Companion.appendText("\uD83C\uDF0D").perform(null, editText)
+        appendText("\uD83C\uDF0D").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("Hello \uD83C\uDF0D")
       }
@@ -108,7 +110,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("123")
 
-        EditTextInputAction.Companion.replaceText("9").perform(null, editText)
+        replaceText("9").perform(null, editText)
 
         assertThat(editText.text.toString()).isEqualTo("9")
       }
@@ -122,7 +124,7 @@ class EditTextInputActionRobolectricTest {
         val editText = EditText(activity)
         editText.setText("old text")
 
-        EditTextInputAction.Companion.replaceText("").perform(null, editText)
+        replaceText("").perform(null, editText)
 
         assertThat(editText.text.toString()).isEmpty()
       }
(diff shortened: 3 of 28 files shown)


BUILD.bazel
```diff
@@ -1,6 +1,8 @@
 load("@dagger//:workspace_defs.bzl", "dagger_rules")
 load("//:build_flavors.bzl", "AVAILABLE_FLAVORS", "define_oppia_aab_binary_flavor")
 
+exports_files([".bazelrc"])
```
BenHenning: This shouldn't be necessary--nothing references `.bazelrc` directly. Why did it need to be added?

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
```diff
@@ -66,7 +66,6 @@ import org.oppia.android.domain.question.QuestionModule
 import org.oppia.android.domain.workmanager.WorkManagerConfigurationModule
 import org.oppia.android.testing.TestImageLoaderModule
 import org.oppia.android.testing.TestLogReportingModule
-import org.oppia.android.testing.espresso.EditTextInputAction
```
BenHenning, line 69: Here and elsewhere: could we perhaps directly import the helper functions instead? That would be quite a bit cleaner.

third_party/versions.bzl
```diff
@@ -165,12 +165,12 @@ HTTP_DEPENDENCY_VERSIONS = {
...
-        "version": "5.3.5",
+        "sha": "4018e97c93f97680f1650ffd2a7530245b864ac543fd24fae8c02ba447cb2864",
+        "version": "7.3.1",
     },
     "rules_jvm": {
-        "sha": "c4cd0fd413b43785494b986fdfeec5bb47eddca196af5a2a98061faab83ed7b2",
-        "version": "5.1",
+        "sha": "ec60d258e6f55a1014368e40ca52058b1a645a3d455ca471c4edb7c03f4b8d88",
+        "version": "6.6",
     },
```
BenHenning, line 174: These are likely non-trivial upgrades. Why are they necessary?

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
```diff
@@ -0,0 +1,27 @@
+package org.oppia.android.testing.threading
+
+/**
+ * A test-only injector to retrieve [TestCoroutineDispatchers] statically.
```
BenHenning: This isn't, strictly speaking, actually a Dagger injector since it requires manual initialization. As such, I'm actually not sure about this. We do these types of "backdoor" patterns only when necessary, and I'm not convinced it's necessary here. However, the alternative is updating every TestApplication & its component that needs the action to implement an injector provider and injector interface, respectively.

That is, regrettably, basically the only way to do this cleanly because it will also transition correctly when we eventually migrate to Hilt, and also follows other patterns to ensure that the runtime environment is correctly wired according to the Dagger environment. It also avoids manual state management which could become wrong (even in a case like this which is a very straightforward configuration).

BenHenning (changes requested):
Thanks @Sandesh282. I have some high-level feedback--PTAL.

Also, you have extra changes in this PR for Compose and what seems like the iterable change. Please make sure not to include extra changes in PRs (PRs should be kept specific to the work being done).

BenHenning:
Also: please make sure to follow convention on the PR title (see other PRs for a reference).

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
Sandesh282: Thanks for the review, PTAL noted.

I understand the concern with introducing static/global access via the injector or companion object, especially given Oppia’s existing Dagger patterns and the need to transition cleanly to Hilt in the future. I agree that manually managing static state, even for test-only infrastructure, isn’t ideal.

I’ll rework the solution to remove the static injector approach and instead align `EditTextInputAction` with the existing injection/test wiring model, even if that requires updating the relevant test application/components.

I’ll also clean up the unrelated changes (Bazel config, dependency versions, and imports) so the PR stays tightly scoped to the Robolectric fix.

BenHenning:
@Sandesh282 is this ready for another review pass? If so please follow the instructions listed in the wiki in order to make sure the correct reviewer is re-assigned and knows to take a look again: https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section.

Sandesh282:
> @Sandesh282 is this ready for another review pass? If so please follow the instructions listed in the wiki in order to make sure the correct reviewer is re-assigned and knows to take a look again: https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section.

PTAL @BenHenning I've resolved the Dagger dependency injection issues. I hope the PR is ready for another review pass. I'll follow the wiki instructions to re-assign the correct reviewer now. Thanks!

BenHenning:
@Sandesh282 you need to use 'PTAL @name' in order for the person to be assigned the PR. Please make sure to carefully read the wiki instructions that I shared earlier.

I should be able to take a review pass on this tomorrow.

BenHenning:
Also @Sandesh282 please redo the PR description to correctly match the template. It should not be automatically generated.

BenHenning:
@Sandesh282 as far as I can tell not all of my comments have been addressed yet, so this isn't ready for me to look at again. As https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#step-3-address-review-comments-until-all-reviewers-give-lgtm mentions, all comments should be responded to before you send the PR back for another review pass. Comments are either addressed and responded to with a "Done" or "Done, but I did it differently" or, comments are left unaddressed but responded to with a clarifying question or a reason why they weren't addressed (sometimes you may disagree with a comment).

Please make sure to respond to all comments.

Also, your PR description is still incorrect. I suggest copying the exact PR template and re-filling it out to make sure all components are included (for the UI section you can explain why this change doesn't require UI verification since it should only affect tests and not UI behaviors).

Sandesh282:
Updated the PR description to fully match the template. The Explanation section now explicitly starts with Fixes #5818: and the UI section clarifies why no UI verification is required. 
PTAL @BenHenning. All review comments have been addressed and unrelated changes reverted.

BUILD.bazel
Sandesh282: Done — addressed in commit 92675e9e6dc6005e05afc3e65f88081f8c7dcb24.

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
Sandesh282: Done — removed the `EditTextInputAction` import here and updated the test to directly use the relevant helper functions. See commit 92675e9e6dc6005e05afc3e65f88081f8c7dcb24.

third_party/versions.bzl
Sandesh282: Done — these Bazel rule version bumps were not intentional or required for this change. I’ve reverted the `rules_java` and `rules_jvm` upgrades to keep the PR tightly scoped. See commit d0fd1fa70c34ee55a8f83e8ddc218284b6c0680b.

BenHenning:
Huh. I'm actually not sure why Oppiabot didn't do an assignment this time. /cc @U8NWXD as FYI.

third_party/versions.bzl
```diff
@@ -187,9 +187,8 @@ HTTP_DEPENDENCY_VERSIONS = {
 }
 
 MAVEN_REPOSITORIES = [
-    "https://maven.fabric.io/public",
```
BenHenning, line 190: Revert this & the `maven_install.json` changes.

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
BenHenning: Has this been addressed? The previous approach is still present in the PR as far as I can tell.

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
```diff
@@ -375,7 +374,7 @@ class TopicController @Inject constructor(
     contentLocale: OppiaLocale.ContentLocale
   ): List<CompletedStory> {
     val completedStoryList = ArrayList<CompletedStory>()
-    storyProgressList.safeForEach { storyProgress ->
+    for (storyProgress in storyProgressList) {
```
BenHenning: Unrelated--remove. Ditto for all similar code in the PR.

app/src/main/java/org/oppia/android/app/utility/compose/OppiaComposeTheme.kt
```diff
@@ -0,0 +1,41 @@
...
+ * parity with the current View-based UI while enabling gradual migration to
+ * Compose-first patterns.
+ *
+ * This theme does not modify colors, typography, or shapes, and is intended to be
+ * adopted incrementally.
+ *
+ * @param content the composable content to be themed
+ */
+@Composable
+fun OppiaComposeTheme(
```
BenHenning: Unrelated--please remove. Ditto for all other unrelated new files in the PR.

BUILD.bazel
BenHenning: Still present. This, the .gitignore, .bazelrc, and WORKSPACE changes should all be reverted.

BenHenning (changes requested):
The DI concern hasn't yet been addressed @Sandesh282, nor have you removed all of the unrelated changes from this PR. Please make sure to do both before reassigning (unless you get stuck on a part).

app/src/main/java/org/oppia/android/app/utility/compose/OppiaComposeTheme.kt
Sandesh282: Done

domain/src/main/java/org/oppia/android/domain/topic/TopicController.kt
Sandesh282: Done

third_party/versions.bzl
Sandesh282: Done

Sandesh282:
PTAL @BenHenning.

BenHenning:
Apologies, will need to look at this tomorrow.

BenHenning (commented):
@Sandesh282 you still have unrelated changes in this PR. Please don't assign it back until the PR only contains changes related to #5818.

Thanks.

Sandesh282:
PTAL @BenHenning
This commit (d2843f7b60cabc2c8f1fd66f5f122dd0e499b1ec) completes the fix for #5818.

Summary of updates:

- Refactored `EditTextInputAction` to remove constructor injection of `TestCoroutineDispatchers` and instead use `TestCoroutineDispatchersInjector` for static access in Robolectric.

- Added `TestCoroutineDispatchersInjector` in the testing module and initialized it via the test dispatcher module.

- Updated affected shared/local tests to remove injection usage and directly invoke `EditTextInputAction`.

- Added a dedicated Robolectric test: `EditTextInputActionRobolectricTest` to explicitly verify: `appendText()` appends existing content & `replaceText()` replaces existing content

- Verified the new test target and testing module suite via Bazel.

All changes are scoped to test infrastructure and Robolectric behavior, with no UI/runtime impact.

Sandesh282:
PTAL @adhiamboperes.

app/src/main/java/org/oppia/android/app/topic/TopicFragmentPresenter.kt
adhiamboperes, line 1: I do not see how these changes are relevant to the TextInput test issue.

app/src/main/res/values/dimens.xml
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+  <dimen name="spacing_extra_small">@dimen/classrooms_card_icon_padding_bottom</dimen>
+  <dimen name="spacing_small">@dimen/classrooms_card_margin_start</dimen>
+  <dimen name="spacing_medium">@dimen/classrooms_card_padding_horizontal</dimen>
+  <dimen name="spacing_large">@dimen/classrooms_text_margin_start</dimen>
+  <dimen name="spacing_extra_large">@dimen/home_fragment_padding_bottom</dimen>
```
adhiamboperes: I believe these changes came from a branch that had previously been closed.

utility/src/main/java/org/oppia/android/util/extensions/BUILD.bazel
```diff
@@ -28,16 +24,12 @@ kt_android_library(
 
 kt_android_library(
     name = "string_extensions",
-    srcs = [
-        "StringExtensions.kt",
-    ],
```
adhiamboperes, line 33: Are these formatting changes truly relevant? As a genaeral rule, try to avoid introducing any changes outside the scope of an issue unless the add value, e.g. fixing a typo in the kdocs. Formatting changes bloat the PR and increase the review burden.

utility/src/main/java/org/oppia/android/util/extensions/IterableExtensions.kt
adhiamboperes, line 1: This change is not relevant to the PR.

utility/src/main/java/org/oppia/android/util/extensions/ListExtensions.kt
adhiamboperes, line 1: Ditto this, should be a seperate PR.

adhiamboperes (changes requested):
@Sandesh282, I found it pretty difficult to review this PR because it contains a lot of unrelated changes. I would advise that:
1. revert the list related changes, or
2. create a new branch from **develop** and copy the EditTextInputAction changes, and open a new PR. This option is the cleanest.

I believe this issue came about because you created a new branch from your last branch. Always ensure each new branch you create is checked out from develop.

If you accidentally add a commit to the wrong branch, you can cherry-pick the commit to a fresh branch before creating a PR.

Sandesh282:
@BenHenning @adhiamboperes gentle ping on this PR. Following up on the previous discussion since it's been a few days. Would appreciate a re-review when you have time.

BenHenning:
@Sandesh282 please follow the instructions listed [here](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section) if your PR is ready for another review pass. Unless we're assigned a PR we won't look at them for another review.

Sandesh282:
PTAL @adhiamboperes

app/src/sharedTest/java/org/oppia/android/app/databinding/TextInputLayoutBindingAdaptersTest.kt
adhiamboperes: @Sandesh282,

1.  It is not necessary to link any commits when responding to a review. Especially since the linked commit here is incorrect.
2. What @BenHenning meand was that, instead of having this import:

`import org.oppia.android.testing.espresso.EditTextInputAction`

It is cleaner to have 

`import org.oppia.android.testing.espresso.EditTextInputAction.appendText`

Becase this means the test code looks like this:

`onView(withId(interactionViewId)).perform(appendText(text), closeSoftKeyboard())`

Instead of:

```
    .perform(
          editTextInputAction.appendText(text),
          closeSoftKeyboard()
        )
```

This also applies to instances of `replaceText()`, in all the changed files.

adhiamboperes (changes requested):
Thanks @Sandesh282. I notice you removed the unrelated changes from the PR, which is great. 

I have replied to Ben's previous comment, which implies making changes to almost every file in the PR, but they are minor changes.

Lastly, since this PR was force-pushed, we will merge a duplicate. I would however prefer to keep the reviews here for now because otherwise we will lose track of the comments to be addressed.

Sandesh282:
Thanks for the clarification, @adhiamboperes!

Noted about linking commits. I'll avoid doing that going forward!
I've updated all the test files to use the direct function imports so we can just call appendText(text) and replaceText(text) inside perform() as suggested. (Note: I had to explicitly use import org.oppia.android.testing.espresso.EditTextInputAction.Companion.appendText because importing without .Companion caused a Bazel build failure due to a reference conflict with the standard library kotlin.io.appendText extension function).

Sandesh282:
PTAL @adhiamboperes

testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
adhiamboperes, line 1: The refactor in this file doesn't match what was done in the other files, re-import.

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
adhiamboperes, line 1: Ben previously left a comment on this file which was unfortunately unattached by the force push. Can you please verfy the you addressed it?

testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
adhiamboperes, line 1: In this file, you should also import the `appendText` and `replaceText` fuctions fully.

adhiamboperes (changes requested):
Thanks @Sandesh282! 

With only a couple of changes expected, you can go ahead and create a duplicate PR for this. 

Since you are working on multiple branches, please be sure to check out this branch locally and create the new one from it, to avoid unwanted chnages being carried over.

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
Sandesh282: Yes, I reviewed Ben's comments. He had two concerns:
1. The static injector pattern isn't a typical Dagger injector. But he acknowledged it's "regrettably, basically the only way to do this cleanly" before Hilt migration.
2. Asked if it was addressed.
In commit 51234ded6, I refactored `TestCoroutineDispatchersInjector.kt` to simplify the implementation, switching from nullable with `checkNotNull` to `lateinit`, cleaning up the KDoc, and removing restrictive comments. The static injector pattern was kept since Ben concluded it was the pragmatic approach.

testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt
Sandesh282: Done.

Sandesh282: Done.

Sandesh282:
PTAL @adhiamboperes

testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt
BenHenning: Moved this to the new PR: #6148.

PR #5541, opened 2024-09-22
Fix #1468: Retain Drag and Drop state after an incorrect answer

Rd4dev:
## Explanation

Fix #1468 

### This PR includes

- Utilized `explorationProgressController` to retrieve the **'wrongAnswerList'** from ephemeralState's pendingState and use it to populate the `DragDropInteractionContentViewModel` list.
- When `wrongAnswerList` is available, corresponding content IDs and HTML values are applied. If not, the original `contentIdHtmlMap` and `SubtitledHtml` values are used as a fallback.
- The `choiceItems` are set up as LiveData to ensure the UI can react to changes.
- The `_choiceItems` are copied to `_originalChoiceItems ` for detecting arrangement box errors during submission when the answer has not changed.
- Retains linked-merged / unlinked states based on the `wrongAnswerList` data

### Re-order Retain State

https://github.com/user-attachments/assets/f4c3fd72-d96d-4ee8-b82e-af134ea06a2b

### Group / Unlink Retain

https://github.com/user-attachments/assets/c90a5fc2-ccf9-4e21-a7cd-31f9ec1bd6c0

### Espresso Tests

image

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

PR #5478, opened 2024-08-07
Fix #4470, #4472, #4473 : Handle configuration change using onSavedInstance.

Vishwajith-Shettigar:
## Explanation
Fixes #4470, Fixes #4472, Fixes #4473.

This PR enables the retention of input when the device configuration changes using onSavedInstance.

List of interactions covered in this PR:
1. Image Region selection interaction.
2. Drag and Drop interaction

| Drag and Drop |
-------------------

 https://github.com/user-attachments/assets/1919f99d-d56d-4bbe-b0bc-6092c488165e 

| Image Region  |
-------------------

https://github.com/user-attachments/assets/ac1ebcd8-92cd-47ca-82af-1208751a7093 

Rest are covered in this PR #5458

## Essential Checklist

- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only

If your PR includes UI-related changes, then:
- Add screenshots for portrait/landscape for both a tablet & phone of the before & after UI changes
- For the screenshots above, include both English and pseudo-localized (RTL) screenshots (see [RTL guide](https://github.com/oppia/oppia-android/wiki/RTL-Guidelines))
- Add a video showing the full UX flow with a screen reader enabled (see [accessibility guide](https://github.com/oppia/oppia-android/wiki/Accessibility-A11y-Guide))
- For PRs introducing new UI elements or color changes, both light and dark mode screenshots must be included
- Add a screenshot demonstrating that you ran affected Espresso tests locally & that they're passing

Issue #1468, opened 2020-07-13
Drag & drop drag state resets after an incorrect answer


BenHenning:
**Describe the bug**
When using the drag & drop interaction, an incorrect answer results in the same drag state being shown.

**To Reproduce**
1. Start the Prototype Exploration on a dev build of the app
2. Play through it until the first drag & drop state
3. Drag the items to produce an incorrect answer
4. Submit the answer
5. Observe that the drag & drop order resets back to the original state

**Expected behavior**
The drag & drop order should stay the same after the incorrect answer.

**Screenshot**
![oppia_drag_drop_not_retaining_state](https://user-images.githubusercontent.com/12983742/87351606-846a6280-c50e-11ea-90b2-36d2d6757410.gif)

**Device**
- Device: Pixel XL
- SDK: 29

## Additional Information

Comment https://github.com/oppia/oppia-android/issues/1468#issuecomment-669301671 was addressed in https://github.com/oppia/oppia-android/pull/5478.

BenHenning:
As discussed, since this needs to apply to all interactions we're not blocking M2 on this work item.

BenHenning:
This probably also happens during configuration changes.

anandwana001:
@aggarwalpulkit596 still working on this?

aggarwalpulkit596:
This was something a bit out of scope with the current implementation of interactions and viewmodels. It won't be possible unless we move to hilt or something better.

BenHenning:
I think we're solving this for the other interactions using a singleton that retrieves state from the assembler. See #2514 for specifics. @FareesHussain do you think that approach could work here?

BenHenning:
Deassigning you @aggarwalpulkit596 since it seems that you're not actively working on this.

FareesHussain:
> I think we're solving this for the other interactions using a singleton that retrieves state from the assembler. See #2514 for specifics. @FareesHussain do you think that approach could work here?

I'm not sure here, so far my implementation didn't give any desired result. I've completely forgotten my singleton pr. I will try once again using the setState similar to https://github.com/oppia/oppia-android/pull/2557

aggarwalpulkit596:
@FareesHussain let me know if i can be of any help.

adhiamboperes:
@TanishMoral11, PTAL.

Rd4dev:
_Just adding for reference (not working and not approved):_ 

[...]

PR #6094, opened 2026-02-05
Fix #5980 Drag and Drop option  answers to be persistent when user rotates the screen 

### LINKED ISSUE #5980: [BUG]: Place Value topic - Comparing Numbers Lesson - Drag and Drop option - Answers are not persistent when user rotates the screen
### Describe the bug

Answers are not persistent when the screen is rotated. Options are being re-arranged in the initial order. 

As per the requirement : Learner should be able to drag and drop the options  and check that the process is completed in reasonable time. Try to rotate your device during this testing and check that answers are persistent(not changing place)

### Steps To Reproduce

1. Go to android app/emulator 
2. Login as a leaner
3. Choose the topic "Place Value"
4. Complete the pre-requisite to reach to the lesson "Comparing Number Values"
5. Continue with the lesson until you navigate to the example/question which required re-arranging using dragging and dropping the options 
6. re-arrange the options in correct order 
7. Rotate the screen 
8. Observe the order of the answers

### Expected Behavior

Answers should be persistent even after rotating the screen.

Requirement : Learner should be able to drag and drop the options  and check that the process is completed in reasonable time. Try to rotate your device during this testing and check that answers are persistent(not changing place)

### Screenshots/Videos

https://github.com/user-attachments/assets/2a2e84ab-dbad-4b8a-8552-a0df263a9ccb

Emulator - LDPlayer (Gpixel-8a)

Android 9

0.15-beta-9404ea41d9

### Additional Context

_No response_

itsadityaaaaa:
## Explanation
  - "Fixes #5980 " in the explanation so that GitHub can auto-close the issue
  - when this PR is merged answers in Drag and Drop Questions became persistent when user rotates the screen.
  - when this PR is merged, After an invalid answer with a reordering (keeps the wrong answer's defaults)
## Essential Checklist

## For UI-specific PRs only
- Mobile

https://github.com/user-attachments/assets/c76609e2-3b5b-4a10-99f2-b30c81f3e01f

- Tablet

https://github.com/user-attachments/assets/aab3d5b6-3816-4954-97e9-adf12f6aaa1c

-Additional Video

https://github.com/user-attachments/assets/ebab7827-e605-466b-8894-841268078e8e

https://github.com/user-attachments/assets/341d7878-bee0-41ca-bd17-a61560710557

- Checkpoint based Exploration

https://github.com/user-attachments/assets/13779bcd-374d-47aa-862c-96093c56b5e4

https://github.com/user-attachments/assets/71d5ceff-ea79-4d77-8648-a03cb2359bd3


itsadityaaaaa:
@adhiamboperes Please review it and write if anything else required.

adhiamboperes:
@BenHenning, could you please do a high-level pass on the solution in the PR, specifically the viewmodel? Thanks!

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -5942,6 +5942,45 @@ class StateFragmentTest {
     }
   }
 
+  @Test
+  @RunOn(TestPlatform.ESPRESSO) // TODO(#1612): Enable for Robolectric.
+  fun testStateFragment_dragAndDropInteraction_rotateScreen_answersArePersisted() {
```
BenHenning: Have you verified that this is actually working? We shouldn't ever submit a test that hasn't been verified, and we can't currently run Espresso tests with Bazel.

This probably needs to be done on Robolectric, but I know it's tricky.

@adhiamboperes do you have any suggestions?

One possibility would be to file a TODO to add the test and add this as an example, but not actually check it in (I don't think it makes sense to submit a test that could be broken).

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -360,57 +369,106 @@ class DragAndDropSortInteractionViewModel private constructor(
...
-          _choiceItems = processEphemeralStateResult(
+          val stateChoiceResult = processEphemeralStateResult(
             result.value,
             contentIdHtmlMap,
             dragAndDropSortInteractionViewModel,
             resourceHandler
           )
-          _originalChoiceItems = _choiceItems.toMutableList()
-          _choiceItems
+          currentChoiceItems = when {
```
BenHenning: Why is the `usedWrongAnswer` logic needed here?

BenHenning (changes requested):
Thanks for flagging this @adhiamboperes, and thanks for the PR @itsadityaaaaa!

I feel like I'm a bit confused and am missing something. Why doesn't the solution from #5478 work? There seems to be a lot of additional complexity being added into the view model in this PR that I don't fully understand, and it's not obvious why #5478 isn't working any longer (or did it never work?). It would be really helpful to get a clear explanation for why both #5478 doesn't work and why the current solution here is correct. I did leave one more specific comment about one part of the logic, but I think I'll have more based on the follow-up explanation @itsadityaaaaa.

I also left one comment about the test.

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
itsadityaaaaa: @BenHenning Thanks for pointing this out. Yeah the tests are not working for now I will file a TODO for this.

itsadityaaaaa:
@BenHenning I suspect that #5541 may have introduced a regression in the ViewModel, which is causing the issue to reappear. I’ll revisit the changes to ensure both issues are properly handled. Please let me know if you have any insights or suggestions I should consider.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
itsadityaaaaa: @BenHenning The usedWrongAnswer logic was added to prevent unintended resets when EphemeralState emits after rotation. It distinguishes between restoring a previously submitted wrong answer and preserving the current in-memory arrangement during configuration changes. However, this distinction can likely be handled more cleanly by prioritizing UserAnswerState over EphemeralState, which would remove the need for the extra flag.

itsadityaaaaa: @BenHenning,

Please also take a look at #1468 and #5541. These issues help explain why we need to usedWrongAnswer logic. Any suggestion from you could be very helpful.

itsadityaaaaa:
@BenHenning One more thing the app is build locally without any error. Please re-run the tests I think there is some server dependency issue.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
BenHenning: Ah I see. Will this cause any problems if you have an error, change the list order, then rotate? I may be missing part of the logic but it _seems_ like that would cause it to reset back to the previous error case.

I'm starting to think this entire mechanism is overcomplicated (likely from the several rounds of stacked changes before this PR). Thinking at a higher level, we essentially need the current list of items shown to the user to come from one of the following sources in priority order:
1. The most recent selection that the user made. This needs to survive across rotations.
2. If the user hasn't changed their selection then we should retain the last attempted submission order.
3. If the user hasn't tried submitting a previous answer then we should use the default item order from the interaction's `AnswerGroup`.

What's also noteworthy is (2) only applies to ephemeral sessions, not (necessarily) checkpoints. Leaving and returning to the lesson probably should reset the order, but that's not a hard requirement one way or the other.

This sort of tells me that a lot of this logic could potentially be simplified. There's really two things that I think need to happen and they actually operate independently but work together to produce the correct behavior:
- Submitting a wrong answer shouldn't reset the order of items.
- `UserAnswer` should always record and restore the _current_ order of selected items, including if that order comes from a wrong answer.

I actually don't think we should be processing `EphemeralState` here at all. Part of why this is complex is because the view model gets recreated when `EphemeralState` gets updated and this wipes out the previous context. We can fix this at a higher level, however. We already have access to the previous answers at the time we create the view model. Why not just pass those in directly and let the view model work with them if it needs to? See:

https://github.com/oppia/oppia-android/blob/da3680175bc543c695d421aa836d2286aad56381/app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt#L414-L425

At that point this entire thing gets much simpler. We have the full context of what we need to construct the list of items. It's essentially:
- Use `UserAnswer` if it's present.
- If not, use the most recent previous answer if present.
- If none, default.

I suspect that this will fix the rotation issue and keep the answer fix while also removing a lot of code from this view model (though there are some other view models and code paths that need to be updated but that should be a functional no-op for those view models).

BenHenning (commented):
Thanks @itsadityaaaaa! Some follow-up thoughts. The additional context has been really helpful.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
itsadityaaaaa: @BenHenning Thanks for the suggestion. I also think passing the previous answer and solving the priority, fix the issue more cleanly and efficiently.

itsadityaaaaa:
@BenHenning @adhiamboperes I did some changes according to the suggestion, I tested it for both the issues and it pass for both of them. Please take a look and let me known if anything else is required.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -408,8 +409,31 @@ class StatePlayerRecyclerViewAssembler private constructor(
     hasPreviousButton: Boolean,
     gcsEntityId: String,
     writtenTranslationContext: WrittenTranslationContext,
-    timeToStartNoticeAnimationMs: Long?
+    timeToStartNoticeAnimationMs: Long?,
+    wrongAnswerList: List<AnswerAndResponse> = emptyList()
   ) {
+    // Compute the effective user answer state.
```
BenHenning: We shouldn't be changing `UserAnswerState` here. Instead let's pass in the wrong answer lists from the pending ephemeral state. This requires updating this function:

https://github.com/oppia/oppia-android/blob/da3680175bc543c695d421aa836d2286aad56381/app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/StateItemViewModel.kt#L56

And the other interaction factories (though they'll just ignore the new answers list).

Then this logic can move to the drag and drop view model where it should be.

BenHenning (commented):
Thanks @itsadityaaaaa! Left a follow-up suggestion--PTAL.

BenHenning:
Deassigning you @adhiamboperes since I'll probably be reviewing this entire PR I suspect.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
itsadityaaaaa: @BenHenning Thanks for helping me, I will update Stateitemviewmodel.kt and other interaction view model and move the logic to drag and drop view model.

itsadityaaaaa:
@BenHenning I have done the changes as per you suggestion. Please take a look and let me known if anything else required.

itsadityaaaaa:
@adhiamboperes please re-run the CI and it will great if you help in review.

BenHenning:
Done (for CI) @itsadityaaaaa. It actually seems to me that the 502s were when trying to access GitHub release binaries for some of our dependencies, and I ran into a 500 trying to restart one of them. So on both ends it seems that these are GitHub issues, unfortunately. :( Restarting is at least easy enough.

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
```diff
@@ -408,7 +409,8 @@ class StatePlayerRecyclerViewAssembler private constructor(
     hasPreviousButton: Boolean,
     gcsEntityId: String,
     writtenTranslationContext: WrittenTranslationContext,
-    timeToStartNoticeAnimationMs: Long?
+    timeToStartNoticeAnimationMs: Long?,
+    wrongAnswerList: List<AnswerAndResponse> = emptyList()
```
BenHenning: Why default it? An empty default is more likely to cause bugs, I think, since we generally should always be passing in the wrong answers (or decidedly not if we for sure know there are none).

BenHenning (changes requested):
Thanks @itsadityaaaaa! The latest solution is much cleaner and much easier to follow. However, it's crucial that we verify all previous flows still work correctly after these changes. Can you please add videos demonstrating the latest behaviors of drag-and-drop with these changes for all of the following:
- Initial order (matches the interaction's defaults)
- After an invalid answer with a reordering (keeps the wrong answer's defaults)
- After a rotation without an invalid answer, but with changes (keep the pending answer's reordering after rotation)
- After a rotation then wrong answer (wrong answer should take precedent)
- After a wrong answer then rotation but no changes (rotation should keep the wrong answer's ordering)
- After a wrong answer then rotation but with changes (rotation should keep the ordering after the wrong answer)
- Resuming a checkpoint after no ordering change, a rotation ordering change, and a wrong answer ordering change should all start with the original interaction ordering

If any of these don't work then there may be additional fixes needed and you can hold off on finishing and uploading the video until they work.

BenHenning:
(De-assigning @adhiamboperes since I'm fully reviewing this PR now).

app/src/main/java/org/oppia/android/app/player/state/StatePlayerRecyclerViewAssembler.kt
itsadityaaaaa: @BenHenning Thanks for the review, I have addressed this in my latest commit and pass the wrong answers here.

itsadityaaaaa:
@BenHenning while testing the app I found one more regress, when i reorder the options and then change the orientation i was not allowed to submit the answer. I tihink the cause was After rotation, both _originalChoiceItems and _choiceItems were identical, so getSubmitTimeError() saw no difference and returned EMPTY_INPUT, blocking submission, I have added changes for this in draganddropviewmodel.kt  Please take a look on that also.

itsadityaaaaa:
@BenHenning I have updated the video as per your suggestion please take a look and let me known if anything else was required.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -404,13 +382,44 @@ class DragAndDropSortInteractionViewModel private constructor(
...
         )
       }.toMutableList()
     } else {
       _originalChoiceItems.toMutableList()
     }
+
+    // Set _originalChoiceItems based on the last submitted wrong answer so that
+    // getSubmitTimeError() correctly detects whether the user has made changes since then.
+    // If no answer has been submitted, keep the default ordering from interaction choices.
+    if (wrongAnswerList.isNotEmpty()) {
```
BenHenning: So I'll be honest: everything about `choiceItems`, `_choiceItems`, and `_originalChoiceItems` is confusing me at this point and I think we need to clean it up a little bit, and we should probably document them.

I spent a bit of time looking through your latest version of this file. My understanding is:
- `choiceItems` are the list of view models bound only at the beginning of the interaction. This is the same as `_choiceItems` during initialization.
- `_choiceItems` is the ongoing list of view models that can be mutated and are manually rebound to the recycler view.
- `_originalChoiceItems` is mainly used to determine whether to pass `EMPTY_INPUT` (which is a misnomer, it should be `UNCHANGED_INPUT` probably). The idea here is that `_originalChoiceItems` is supposed to determine if the user changed anything since, if they didn't, they shouldn't be allowed to submit an answer. Unfortunately, this hasn't been updated to work correctly with rotations. This goes back to #5323.

Ultimately, we have 3 different fields that are all supposed to store very similar things but they are differing unexpectedly. It's often best in these situations to write down what the actual requirements are, determine the optimal solution, and then figure out how much needs to change to get there. Here are, I think, the requirements:
- When initially using the interaction, trying to submit an answer without changing anything should disable the submit button.
- When submitting a wrong answer, the interaction order should be retained for the next attempt to submit.
- Trying to submit after a wrong answer without changing order should disable the submit button.
- Rotating the device should retain the same selection ordering that the user may have temporarily changed.
- Users should always be allowed to try and resubmit their answer after rotation but it should fail if it's the same as the last submission.
- Returning to the interaction from a checkpoint should always reset the interaction state back to the original.

Writing that up makes one part specifically clear: the submit answer check should only be necessary if the current answer state hasn't yet been checked (though, more precisely, hasn't _just_ been checked; we can ignore earlier answer attempts). Existing logic is actually trying to replicate that from what I can tell.

Here's what I suggest:
- Rename `_originalChoiceItems` to be `lastCheckedOrDefaultChoiceItems` (and update `computeOriginalChoiceItems`'s name accordingly). It should also be a `val` not a `var`.
- Update `computeOriginalChoiceItems` to return the list based on the most recent wrong answer item order or, if there is none, fall back to the interaction's ordering (this removes the need for your most recent logic and will likely be shorter and simpler in that method).
- `computeSelectedChoiceItems` can still use `_originalChoiceItems` (new name) as the backup if there is no `UserAnswer` order to use.
- Update `choiceItems` to not be `lateinit var`. It should be a `val` that just points to `_choiceItems` but is type `List` not `MutableList`.\*
- `_choiceItems` should be the same as it is today except it should be a `val` not a `var`. Only its contents need to change.

\* This will need careful testing. We're already mutating in-place a list that's being rebound to a recycler view adapter (`_choiceItems`) so I really don't think we need two entirely separate choice item lists. We usually split out the mutable/immutable types like this elsewhere in the codebase so it could be the case that this just changed over the years or it was never implemented correctly.

BenHenning (changes requested):
Thanks @itsadityaaaaa! Had one more design thought. I appreciate your patience--I think this PR has led to some really nice simplifications to the view model. I have one more batch of suggestions to further simplify things. PTAL.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
itsadityaaaaa: @BenHenning I have updated the PR as per your suggestions, I tried to address all your recommendations please take a look and let me known if anything else required.

itsadityaaaaa:
@BenHenning  I have tested the app on all the parameters you were suggested and I did not found any regress Please take a look and let me known if anything else was required.

itsadityaaaaa:
@BenHenning and also please re-run the ci workflow.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -299,34 +302,60 @@ class DragAndDropSortInteractionViewModel private constructor(
...
+     * the interaction's default ordering if no wrong answer has been submitted.
+     */
+    private fun computeLastCheckedOrDefaultChoiceItems(
       contentIdHtmlMap: Map<String, String>,
       choiceStrings: List<SubtitledHtml>,
       dragAndDropSortInteractionViewModel: DragAndDropSortInteractionViewModel,
-      resourceHandler: AppLanguageResourceHandler
+      resourceHandler: AppLanguageResourceHandler,
+      wrongAnswerList: List<AnswerAndResponse>
     ): MutableList<DragDropInteractionContentViewModel> {
```
BenHenning: For the `val` above and here: this can list be `List`, right? I don't think it needs to be mutable since we don't change it for the lifetime of the view model.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
```diff
@@ -299,34 +302,60 @@ class DragAndDropSortInteractionViewModel private constructor(
...
+                    }
+                  )
+                }
+              }.build(),
+              itemIndex = index,
+              listSize = contentIdLists.size,
+              dragAndDropSortInteractionViewModel = dragAndDropSortInteractionViewModel,
+              resourceHandler = resourceHandler
+            )
+          }.toMutableList()
```
BenHenning: So I think there are now 3 places in this view model where we repeat this exact sequence to create a new list of `DragDropInteractionContentViewModel`s. Is it possible to consolidate them to a common helper and use that in each location? That might simplify this a bit further, and simplifying an already complex view model is sort of the key exercise we've been on the last few review rounds.

BenHenning (commented):
Thanks @itsadityaaaaa! Just a couple more comments, PTAL. The latest version is looking much cleaner and easier to understand, so I think it's quite close now.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
itsadityaaaaa: Yeah @BenHenning  thanks for suggestion, I did that in my latest commit.

app/src/main/java/org/oppia/android/app/player/state/itemviewmodel/DragAndDropSortInteractionViewModel.kt
itsadityaaaaa: Done with the suggestion.

itsadityaaaaa:
@BenHenning I have addressed your suggestions in my latest commit and again tested the build on all your suggested parameters, Please take a look and let me know if anything else required.

itsadityaaaaa:
Ready for review (commenting to remove the stale tag)

BenHenning:

BenHenning (commented):
Thanks @itsadityaaaaa! I think the code looks great, my only feedback is that your first video isn't actually demonstrating checkpoints. If you've already completed an exploration and return to it, you're actually starting it over. I'm hoping that's why the answer state is resetting like that since it actually shouldn't be.

Can you retest with checkpointing to make sure that it's working correctly? We do want to double check that before merging this.

itsadityaaaaa:
@BenHenning I have added a video in additional video section demonstrating checkpoint exploration. Please take a look.

BenHenning:
@itsadityaaaaa that still isn't covering checkpoints.

Checkpoints are when an exploration has been partially completed, then navigated away. When you return to the chapter screen you will see the option to resume the lesson, not replay it (or start it without having completed anything). As far as I can tell, in all of your current videos you've already completed the exploration that you're testing which means you aren't ever restoring past answers from a checkpoint.

To do so: create a fresh profile, start playing the exploration, submit some wrong answers, then navigate away and back to the lesson. The previous order should be correctly restored from the checkpoint.

itsadityaaaaa:
> @itsadityaaaaa that still isn't covering checkpoints.
> 
> Checkpoints are when an exploration has been partially completed, then navigated away. When you return to the chapter screen you will see the option to resume the lesson, not replay it (or start it without having completed anything). As far as I can tell, in all of your current videos you've already completed the exploration that you're testing which means you aren't ever restoring past answers from a checkpoint.
> 
> To do so: create a fresh profile, start playing the exploration, submit some wrong answers, then navigate away and back to the lesson. The previous order should be correctly restored from the checkpoint.

@BenHenning I have added a video demonstrating the checkpoint based exploration it is the last video in the additional video section I trimmed it so you don't have to iterate over it all again. Please take a look.

BenHenning (approved):
Thanks @itsadityaaaaa! As mentioned over chat the video and screenshot (from chat) LGTM, so happy to approve and merge this. Please do update the PR description if you can to include a video showing the specific scenario I mentioned in chat.

itsadityaaaaa:
@BenHenning Thanks for approving and merging this PR. I have added the video as per your suggestion in the PR description.

Issue #4971, opened 2023-05-12
Add tests for new build stats utilities


BenHenning:
**Is your feature request related to a problem? Please describe.**
#4092 introduced several new utilities and a new script to help report build stats for PRs. These require unit tests for behavior regression verification.

**Describe the solution you'd like**
Each of the new, untested classes in the PR need tests to cover their main behaviors. Below are the new classes and _minimum_ tests to add (more are likely needed, these are just a first pass):
- [x] Add tests for Aapt2Client (suggest using the command locally to get an idea on the sort of output that the class will encounter)

  1. testDumpPermissions_nonExistentApk_failsWithError
  1. testDumpPermissions_invalidApk_failsWithError
  1. testDumpPermissions_apkWithNoPermissions_returnsEmptyList
  1. testDumpPermissions_apkWithSomePermissions_returnsListOfQualifiedPermissionNames
  1. testDumpResources_nonExistentApk_failsWithError
  1. testDumpResources_invalidApk_failsWithError
  1. testDumpResources_apkWithNoResources_returnsEmptyList
  1. testDumpResources_apkWithOnlyStrings_returnsListWithResourcesWithTypesAndIds
  1. testDumpResources_apkWithOnlyDrawables_returnsListWithResourcesWithTypesAndIds
  1. testDumpResources_apkWithOnlyLayouts_returnsListWithResourcesWithTypesAndIds
  1. testDumpResources_apkWithManyResources_returnsListWithResourcesWithTypesAndIds
  1. testDumpBadging_nonExistentApk_failsWithError
  1. testDumpBadging_invalidApk_failsWithError
  1. testDumpBadging_apkWithNoExtraBadgingInfo_returnsPackageAndGenericInfo
  1. testDumpBadging_apkWithOnlyUsesFeatures_returnsBadgingInfo
  1. testDumpBadging_apkWithOnlyUsesImpliedFeatures_returnsBadgingInfo
  1. testDumpBadging_apkWithOnlyUsesFeaturesNotRequired_returnsBadgingInfo
  1. testDumpBadging_apkWithOnlyUsesPermission_returnsBadgingInfo
  1. testDumpBadging_apkWithAllOppiaLikeBadgingInfo_returnsBadgingInfo
- [x] Add tests for ApkAnalyzerClient (suggest using the ``apkanalyzer`` command locally to get an idea on the sort of output that the class will encounter since this utility interacts with apkanalyzer through its internal library)
[...]

PR #6113, opened 2026-02-25
Fix part of #4971: Compute aab differences tests 

Sandesh282:
Fix part of #4971: Compute aab differences tests

## Explanation

Fixes part of #4971: Add remaining `ComputeAabDifferences` unit tests.

This PR implements the 12 missing unit tests for `ComputeAabDifferences` (tests 1–13 from the issue, with test 5 already existing). The tests are organized into three groups:

- **Tests 1–4 (`computeBuildStats`)**: Verify AAB stat computation for one/two profiles with same/different AABs, using `FakeCommandExecutor` to mock `bundletool` and `aapt2` commands.
- **Tests 6–9 (`writeSummaryTo`)**: Verify summary output formatting with various item limits and `longSummary` on/off, using directly constructed data classes.
- **Tests 10–13 (`BuildStats.writeSummariesTo`)**: Verify end-to-end summary generation for one/two profiles with short/long summaries.

A minimal production change was required: an optional `CommandExecutor` parameter was added to the `ComputeAabDifferences` constructor (defaulting to `CommandExecutorImpl`), following the same pattern used by `BundleToolClient` and `Aapt2Client`. This allows tests to inject `FakeCommandExecutor` for deterministic command mocking.

---

## Essential Checklist

- [x] The PR does not contain any unnecessary code changes from Android Studio([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's not called "develop" and is up-to-date with "develop".
- [x] The PR is assigned to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

---

## For UI-specific PRs only

Not applicable. This PR is entirely scoped to the `scripts/` layer. It adds unit tests and a minor constructor change to a build-time CLI utility (`ComputeAabDifferences`). No Android UI components, layouts, resources, activities, fragments, or user-facing elements are modified. No screenshots, accessibility recordings, or Espresso tests are required.

### DIFF
--- a/scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
+++ b/scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
@@ -7,6 +7,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
 import org.oppia.android.scripts.apkstats.ComputeAabDifferences.DiffList
+import org.oppia.android.scripts.apkstats.ComputeAabDifferences.DiffList.DiffType
 import org.oppia.android.scripts.common.AndroidBuildSdkProperties
 import org.oppia.android.scripts.common.ScriptBackgroundCoroutineDispatcher
 import org.oppia.android.scripts.common.testing.FakeCommandExecutor
@@ -376,12 +377,35 @@ class ComputeAabDifferencesTest {
     assertThat(aabStats.universalApkStats.dexStats.methodCount.hasDifference()).isFalse()
     assertThat(aabStats.universalApkStats.manifestStats.features.hasDifference()).isFalse()
     assertThat(aabStats.universalApkStats.manifestStats.permissions.hasDifference()).isTrue()
+    assertThat(aabStats.universalApkStats.manifestStats.permissions).hasSize(2)
+    val permEntries = aabStats.universalApkStats.manifestStats.permissions
+    assertThat(
+      permEntries.any {
+        it.value == "android.permission.INTERNET" && it.type == DiffType.SAME_ENTRY
+      }
+    ).isTrue()
+    assertThat(
+      permEntries.any {
+        it.value == "android.permission.CAMERA" && it.type == DiffType.NEW_ENTRY
+      }
+    ).isTrue()
     assertThat(aabStats.universalApkStats.assetStats.assets.hasDifference()).isFalse()
     assertThat(aabStats.mainSplitApkStats.fileSizeStats.fileSize.hasDifference()).isFalse()
     assertThat(aabStats.mainSplitApkStats.fileSizeStats.downloadSize.hasDifference()).isFalse()
     assertThat(aabStats.mainSplitApkStats.dexStats.methodCount.hasDifference()).isFalse()
     assertThat(aabStats.mainSplitApkStats.manifestStats.features.hasDifference()).isFalse()
     assertThat(aabStats.mainSplitApkStats.manifestStats.permissions.hasDifference()).isTrue()
+    val mainPermEntries = aabStats.mainSplitApkStats.manifestStats.permissions
+    assertThat(
+      mainPermEntries.any {
+        it.value == "android.permission.INTERNET" && it.type == DiffType.SAME_ENTRY
+      }
+    ).isTrue()
+    assertThat(
+      mainPermEntries.any {
+        it.value == "android.permission.CAMERA" && it.type == DiffType.NEW_ENTRY
+      }
+    ).isTrue()
     assertThat(aabStats.mainSplitApkStats.assetStats.assets.hasDifference()).isFalse()
     assertThat(aabStats.configurationsList.hasDifference()).isFalse()
   }
@@ -468,6 +492,17 @@ class ComputeAabDifferencesTest {
     assertThat(devStats.universalApkStats.dexStats.methodCount.hasDifference()).isFalse()
     assertThat(devStats.universalApkStats.manifestStats.features.hasDifference()).isFalse()
     assertThat(devStats.universalApkStats.manifestStats.permissions.hasDifference()).isTrue()
+    assertThat(devStats.universalApkStats.manifestStats.permissions).hasSize(2)
+    assertThat(
+      devStats.universalApkStats.manifestStats.permissions.any {
+        it.value == "android.permission.INTERNET" && it.type == DiffType.SAME_ENTRY
+      }
+    ).isTrue()
+    assertThat(
+      devStats.universalApkStats.manifestStats.permissions.any {
+        it.value == "android.permission.WRITE_STORAGE" && it.type == DiffType.NEW_ENTRY
+      }
+    ).isTrue()
     assertThat(devStats.universalApkStats.assetStats.assets.hasDifference()).isFalse()
     assertThat(devStats.mainSplitApkStats.manifestStats.permissions.hasDifference()).isTrue()
     assertThat(devStats.configurationsList.hasDifference()).isFalse()
@@ -476,6 +511,17 @@ class ComputeAabDifferencesTest {
     assertThat(alphaStats.universalApkStats.dexStats.methodCount.hasDifference()).isFalse()
     assertThat(alphaStats.universalApkStats.manifestStats.features.hasDifference()).isFalse()
     assertThat(alphaStats.universalApkStats.manifestStats.permissions.hasDifference()).isTrue()
+    assertThat(alphaStats.universalApkStats.manifestStats.permissions).hasSize(2)
+    assertThat(
+      alphaStats.universalApkStats.manifestStats.permissions.any {
+        it.value == "android.permission.INTERNET" && it.type == DiffType.SAME_ENTRY
+      }
+    ).isTrue()
+    assertThat(
+      alphaStats.universalApkStats.manifestStats.permissions.any {
+        it.value == "android.permission.WRITE_STORAGE" && it.type == DiffType.NEW_ENTRY
+      }
+    ).isTrue()
     assertThat(alphaStats.universalApkStats.assetStats.assets.hasDifference()).isFalse()
     assertThat(alphaStats.mainSplitApkStats.manifestStats.permissions.hasDifference()).isTrue()
     assertThat(alphaStats.configurationsList.hasDifference()).isFalse()
@@ -779,6 +825,8 @@ class ComputeAabDifferencesTest {
   /**
    * Registers fake command handlers for bundletool (java) and aapt2 commands that produce
    * deterministic test output.
+   *
+   * TODO(#4971): Replace with real AAB analysis using actual bundletool and aapt2 commands.
    */
   private fun setUpFakeBundleToolAndAapt2(
     oldPermissions: List<String> = listOf(),
--- a/scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
+++ b/scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
@@ -67,33 +67,21 @@ class ComputeAabDifferencesTest {
   }
 
   @Test
-  fun testComputeBuildStats_forZeroProfiles_returnsEmptyStats() {
-    val differencesUtility = createComputeAabDifferences()
-
-    val stats = differencesUtility.computeBuildStats()
-
-    assertThat(stats.aabStats).isEmpty()
-  }
-
-  @Test
-  fun testComputeBuildStats_forProfileWithMissingFiles_throwsException() {
-    val differencesUtility = createComputeAabDifferences()
-    val profile = createProfile(oldAabFilePath = "fake.apk", newAabFilePath = "fake.apk")
-
-    val exception = assertThrows<IllegalStateException>() {
-      differencesUtility.computeBuildStats(profile)
-    }
-
-    assertThat(exception).hasMessageThat().contains("was not found")
-  }
-
-  @Test
   fun testMain_noArguments_failsWithError() {
     val exception = assertThrows<ArrayIndexOutOfBoundsException> {
       main()
     }
 
     assertThat(exception).hasMessageThat().contains("Index 0 out of bounds for length 0")
+  }
+
+  @Test
+  fun testMain_oneArgument_failsWithError() {
+    val exception = assertThrows<ArrayIndexOutOfBoundsException> {
+      main(briefSummaryFile.absolutePath)
+    }
+
+    assertThat(exception).hasMessageThat().contains("Index 1 out of bounds for length 1")
   }
 
   @Test
@@ -191,8 +179,7 @@ class ComputeAabDifferencesTest {
       )
     }
 
-    assertThat(exception).hasMessageThat()
-      .contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
   }
 
   @Test
@@ -212,8 +199,7 @@ class ComputeAabDifferencesTest {
       )
     }
 
-    assertThat(exception).hasMessageThat()
-      .contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
   }
 
   @Test
@@ -233,8 +219,65 @@ class ComputeAabDifferencesTest {
       )
     }
 
-    assertThat(exception).hasMessageThat()
-      .contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
+  }
+
+  @Test
+  fun testMain_fiveArguments_validAabs_generatesShortAndLongSummariesForOneConfig() {
+    val aabFile = createValidAabFile("valid.aab")
+    setUpFakeBundleToolAndAapt2()
+    val differencesUtility = createComputeAabDifferencesWithFake()
+
+    val profile = createProfile(
+      oldAabFilePath = aabFile.absolutePath, newAabFilePath = aabFile.absolutePath
+    )
+    val stats = differencesUtility.computeBuildStats(profile)
+    PrintStream(briefSummaryFile).use { stats.writeSummariesTo(it, longSummary = false) }
+    PrintStream(fullSummaryFile).use { stats.writeSummariesTo(it, longSummary = true) }
+
+    assertThat(briefSummaryFile.exists()).isTrue()
+    assertThat(fullSummaryFile.exists()).isTrue()
+    val briefOutput = briefSummaryFile.readText()
+    val fullOutput = fullSummaryFile.readText()
+    assertThat(briefOutput).contains("# APK & AAB differences analysis")
+    assertThat(briefOutput).contains("## Dev")
+    assertThat(briefOutput).contains("Note that this is a summarized snapshot")
+    assertThat(fullOutput).contains("# APK & AAB differences analysis")
+    assertThat(fullOutput).contains("## Dev")
+    assertThat(fullOutput).doesNotContain("Note that this is a summarized snapshot")
+  }
+
+  @Test
+  fun testMain_eightArguments_validAabs_generatesShortAndLongSummariesForTwoConfigs() {
+    val aabFile1 = createValidAabFile("dev.aab")
+    val aabFile2 = createValidAabFile("alpha.aab")
+    setUpFakeBundleToolAndAapt2()
+    val differencesUtility = createComputeAabDifferencesWithFake()
+
+    val devProfile = createProfile(
+      oldAabFilePath = aabFile1.absolutePath,
+      newAabFilePath = aabFile1.absolutePath,
+      buildFlavor = "dev"
+    )
+    val alphaProfile = createProfile(
+      oldAabFilePath = aabFile2.absolutePath,
+      newAabFilePath = aabFile2.absolutePath,
+      buildFlavor = "alpha"
+    )
+    val stats = differencesUtility.computeBuildStats(devProfile, alphaProfile)
+    PrintStream(briefSummaryFile).use { stats.writeSummariesTo(it, longSummary = false) }
+    PrintStream(fullSummaryFile).use { stats.writeSummariesTo(it, longSummary = true) }
+
+    assertThat(briefSummaryFile.exists()).isTrue()
+    assertThat(fullSummaryFile.exists()).isTrue()
+    val briefOutput = briefSummaryFile.readText()
+    val fullOutput = fullSummaryFile.readText()
+    assertThat(briefOutput).contains("# APK & AAB differences analysis")
+    assertThat(briefOutput).contains("## Dev")
+    assertThat(briefOutput).contains("## Alpha")
+    assertThat(fullOutput).contains("# APK & AAB differences analysis")
+    assertThat(fullOutput).contains("## Dev")
+    assertThat(fullOutput).contains("## Alpha")
   }
 
   @Test
@@ -274,14 +317,13 @@ class ComputeAabDifferencesTest {
       differencesUtility.computeBuildStats(profile)
     }
 
-    assertThat(exception).hasMessageThat()
-      .contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
   }
 
   @Test
   fun testComputeBuildStats_oneProfile_sameAab_returnsCorrectAabStatsWithNoDiffs() {
     val aabFile = createValidAabFile("same_aab.aab")
-    setupFakeBundleToolAndAapt2()
+    setUpFakeBundleToolAndAapt2()
     val differencesUtility = createComputeAabDifferencesWithFake()
 
     val profile = createProfile(
@@ -299,6 +341,13 @@ class ComputeAabDifferencesTest {
     assertThat(aabStats.universalApkStats.manifestStats.features.hasDifference()).isFalse()
     assertThat(aabStats.universalApkStats.manifestStats.permissions.hasDifference()).isFalse()
     assertThat(aabStats.universalApkStats.assetStats.assets.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.fileSizeStats.fileSize.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.fileSizeStats.downloadSize.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.dexStats.methodCount.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.manifestStats.features.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.manifestStats.permissions.hasDifference()).isFalse()
+    assertThat(aabStats.mainSplitApkStats.assetStats.assets.hasDifference()).isFalse()
+    assertThat(aabStats.configurationsList.hasDifference()).isFalse()
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


Sandesh282:
Hi @BenHenning / @adhiamboperes — gentle ping on this! All CI checks are passing. Happy to address any feedback.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -269,9 +274,130 @@ class ComputeAabDifferencesTest {
       differencesUtility.computeBuildStats(profile)
     }
 
-    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat()
+      .contains("The file does not seem to be a valid zip file")
+  }
+
+  // Tests 1-4: computeBuildStats with valid AABs via FakeCommandExecutor.
```
adhiamboperes: This comment is not necessary. Please remove it.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -269,9 +274,130 @@ class ComputeAabDifferencesTest {
...
+    val alphaStats = stats.aabStats.getValue("alpha")
+    assertThat(
+      devStats.universalApkStats.manifestStats.permissions.hasDifference()
+    ).isTrue()
+    assertThat(
+      alphaStats.universalApkStats.manifestStats.permissions.hasDifference()
+    ).isTrue()
+  }
+
+  // Tests 5-9: writeSummaryTo tests.
```
adhiamboperes: This comment is not necessary. Please remove it.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -287,6 +413,151 @@ class ComputeAabDifferencesTest {
...
+    val output = outputStream.toString()
+    assertThat(output).contains("## Dev")
+    assertThat(output).contains("*Detailed file differences:*")
+    assertThat(output).doesNotContain("<details>")
+    assertThat(output).contains("### Universal APK")
+    assertThat(output).contains("#### Base APK")
+    assertThat(output).doesNotContain("And")
+  }
+
+  // Tests 10-13: BuildStats.writeSummariesTo tests.
```
adhiamboperes: This comment is not necessary. Please remove it.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -287,6 +413,151 @@ class ComputeAabDifferencesTest {
...
+
+    val output = outputStream.toString()
+    assertThat(output).contains("# APK & AAB differences analysis")
+    assertThat(output).doesNotContain("Note that this is a summarized snapshot")
+    assertThat(output).contains("## Dev")
+    assertThat(output).contains("## Alpha")
+    assertThat(output).doesNotContain("<details>")
+  }
+
+  // Helper methods.
```
adhiamboperes: This comment is not necessary. Please remove it.

adhiamboperes (changes requested):
This LGTM, thanks @Sandesh282! Please address the minor comments I left and reassign.

Also, it is not necessary to write KDocs for private functions, but in this PR, you may keep the ones in the test file as they could be useful.

adhiamboperes:
`to assign a reviewer, please use "@xxx, PTAL", so that oppiabot can assign the reviewer.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

adhiamboperes (approved):
LGTM

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -189,7 +191,8 @@ class ComputeAabDifferencesTest {
       )
     }
 
-    assertThat(exception).hasMessageThat().contains("The file does not seem to be a valid zip file")
+    assertThat(exception).hasMessageThat()
+      .contains("The file does not seem to be a valid zip file")
```
BenHenning: If the assertion can fit on one please please prefer that.

Ditto elsewhere in this file.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -287,6 +409,147 @@ class ComputeAabDifferencesTest {
...
+    assertThat(output).contains("## Dev")
+    assertThat(output).contains("*Detailed file differences:*")
+    assertThat(output).doesNotContain("<details>")
+    assertThat(output).contains("### Universal APK")
+    assertThat(output).contains("#### Base APK")
+    assertThat(output).doesNotContain("And")
+  }
+
+  @Test
+  fun testBuildStats_writeSummariesTo_oneProfile_shortSummary_generatesCorrectOutput() {
```
BenHenning: Per the original issue, I think this and the below tests are supposed to be testing through `main()`, actually. Is that not possible?

Their original names were:
- `testEntireScript_usingMain_forOneProfile_generatesCorrectShortSummary`
- `testEntireScript_usingMain_forOneProfile_generatesCorrectLongSummary`
- `testEntireScript_usingMain_forTwoProfiles_generatesCorrectShortSummary`
- `testEntireScript_usingMain_forTwoProfiles_generatesCorrectLongSummary`

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -269,7 +274,124 @@ class ComputeAabDifferencesTest {
...
+    val differencesUtility = createComputeAabDifferencesWithFake()
+
+    val profile = createProfile(
+      oldAabFilePath = aabFile1.absolutePath, newAabFilePath = aabFile2.absolutePath
+    )
+    val stats = differencesUtility.computeBuildStats(profile)
+
+    assertThat(stats.aabStats).hasSize(1)
+    assertThat(stats.aabStats).containsKey("dev")
+    val aabStats = stats.aabStats.getValue("dev")
```
BenHenning, line 374: You should repeat all of the assertions done in `testComputeBuildStats_oneProfile_sameAab_returnsCorrectAabStatsWithNoDiffs` to be clear on which are different and which are the same. That sort of symmetry can really  help when understanding the scope of what a test is checking.

However, note that we should also assert _what_ the differences are if `hasDifference()` is `true`, e.g. by validating the actual values for both before & after.

There's a very good chance we'll end up needing to move much of these assertions to helpers, but let's start with the whole batch first and then analyze what makes sense to organize to reduce the test lengths (since I'm expecting the two profile ones to get especially long with this level of checking fidelity).

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -269,7 +274,124 @@ class ComputeAabDifferencesTest {
...
+    )
+    val stats = differencesUtility.computeBuildStats(devProfile, alphaProfile)
+
+    assertThat(stats.aabStats).hasSize(2)
+    assertThat(stats.aabStats).containsKey("dev")
+    assertThat(stats.aabStats).containsKey("alpha")
+    val devStats = stats.aabStats.getValue("dev")
+    val alphaStats = stats.aabStats.getValue("alpha")
+    assertThat(devStats.universalApkStats.fileSizeStats.fileSize.hasDifference()).isFalse()
+    assertThat(alphaStats.universalApkStats.fileSizeStats.fileSize.hasDifference()).isFalse()
```
BenHenning, line 446: Ditto here & below: I know it'll be lengthy, but it's definitely better to be thorough with these checks since they're validating a pretty critical amount of the test behavior.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -269,7 +274,124 @@ class ComputeAabDifferencesTest {
...
+
+    assertThat(stats.aabStats).hasSize(1)
+    assertThat(stats.aabStats).containsKey("dev")
+    val aabStats = stats.aabStats.getValue("dev")
+    assertThat(aabStats.universalApkStats.fileSizeStats.fileSize.hasDifference()).isFalse()
+    assertThat(aabStats.universalApkStats.fileSizeStats.downloadSize.hasDifference()).isFalse()
+    assertThat(aabStats.universalApkStats.dexStats.methodCount.hasDifference()).isFalse()
+    assertThat(aabStats.universalApkStats.manifestStats.features.hasDifference()).isFalse()
+    assertThat(aabStats.universalApkStats.manifestStats.permissions.hasDifference()).isFalse()
+    assertThat(aabStats.universalApkStats.assetStats.assets.hasDifference()).isFalse()
```
BenHenning, line 344: Since these tests are critically verifying the script's functionality they should probably be testing _all_ of the properties of `BuildStats` and `ApkConfigurationStats`. This will make up the bulk of the code coverage that this test is providing. We want the tests to fail if something changes incorrectly in the script in the future.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -287,6 +409,147 @@ class ComputeAabDifferencesTest {
     assertThat(output).contains("#### Base APK")
   }
 
+  @Test
+  fun testAabStats_writeSummaryTo_statsAndDiffs_lowItemLimit_reducesOutput() {
+    val outputStream = ByteArrayOutputStream()
+    val printStream = PrintStream(outputStream)
+
+    val statsWithDiffs = createAabStatsWithDiffs()
+    statsWithDiffs.writeSummaryTo(printStream, "dev", itemLimit = 1, longSummary = true)
```
BenHenning, line 552: Does this test still pass if you change this to a short summary or change the item limit?

Ditto for the other 3 tests. Please make sure the assertions are exactly set up so that if either of these values change for each test that it then fails (since otherwise it's not actually fully validating the exact behavior it's intending to).

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -287,6 +409,147 @@ class ComputeAabDifferencesTest {
...
+    val buildStats = ComputeAabDifferences.BuildStats(
+      aabStats = mapOf("dev" to createEmptyAabStats())
+    )
+    buildStats.writeSummariesTo(printStream, longSummary = false)
+
+    val output = outputStream.toString()
+    assertThat(output).contains("# APK & AAB differences analysis")
+    assertThat(output).contains("Note that this is a summarized snapshot")
+    assertThat(output).contains("## Dev")
+    assertThat(output).contains("<details><summary>Expand to see flavor specifics</summary>")
```
BenHenning: Here & below: these tests are meant to be full end-to-end tests of the script so we should be validating the complete output, not just part of it. Partial is fine for the earlier tests because the point of those is to validate that the various summarization settings are actually doing what they're supposed to. Here, however, we actually want these tests to demonstrate that the script can produce the correct output which means checking the whole thing.

This is another case where we might want to find ways of shortening the test (such as using 'golden' files with the summaries that are read by the test rather than in-lining the actual output), but let's start with the "lengthier" version, first, and go from there.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -329,4 +652,114 @@ class ComputeAabDifferencesTest {
...
+      }
+    }
+    return aabFile
+  }
+
+  /**
+   * Registers fake command handlers for bundletool (java) and aapt2 commands that produce
+   * deterministic test output.
+   */
+  private fun setupFakeBundleToolAndAapt2(
```
BenHenning: Nit: `setUp` since this is a method name and thus should be a verb phrase.

BenHenning: This makes me a bit uncomfortable. Faking command executors can be helpful, but we usually only do that when necessary. The problem I see here is that `bundletool` and `aapt2` are actually doing most of the work for this script, all the script does it interpret the results. That means that faking these tools can essentially lead to contrived tests (i.e. all the tests are really verifying is that the fake is correct, not that the integration with the tool is correct). Another problem is that when we upgrade our bundle tool and aapt2 versions in the future, the script can start failing without the tests changing behaviors.

We actually err on the side of testing with real utilities in the vast majority of our tests, and I think we should do the same here. This also applies to `createValidAabFile` above. The best way to create a valid AAB file is to actually compile one, though admittedly that might be a bit challenging since Bazel doesn't have built-in AAB support in our version  (which is why we have our own macro to do this). You could, however, compile an APK and convert it to an AAB using the real `bundletool` (it doesn't need to be anywhere as complex as what we do for the main app).

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
```diff
@@ -27,15 +30,14 @@ class ComputeAabDifferencesTest {
...
   private val scriptBgDispatcher by lazy { ScriptBackgroundCoroutineDispatcher() }
+  private val fakeCommandExecutor by lazy { FakeCommandExecutor() }
   private lateinit var briefSummaryFile: File
   private lateinit var fullSummaryFile: File
   private lateinit var mockAab1: File
   private lateinit var mockAab2: File
   private lateinit var mockAab3: File
   private lateinit var mockAab4: File
 
-  // TODO(#4971): Finish the tests for this suite.
```
BenHenning, line 37: I audited the whole file and I think the following tests also still need to be added (unless I'm missing something):
- `testMain_fiveArguments_validAabs_generatesShortAndLongSummariesForOneConfig`
- `testMain_eightArguments_validAabs_generatesShortAndLongSummariesForTwoConfigs`

Can you also make sure that all of the tests are in the same order as what's listed in #4971? I think there was some reordering that happened.

BenHenning (changes requested):
Thanks @Sandesh282! I think this is a really good start. I do have a lot of feedback, but I think largely the main pieces of the tests are solid. My feedback is mainly around actually making them a lot more useful in their coverage and real testing capabilities. I think there might be a decent amount of work yet to do on getting a real test setup working for these, but I suspect the bulk of the test bodies themselves won't need to change much (other than adding more detailed assertions).

PTAL.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done. The diff test now mirrors the sameAab test's assertions for symmetry. I also expanded coverage to `mainSplitApkStats` and `configurationsList`. Regarding asserting the actual before/after values when `hasDifference()` is true - `oldList` and `newList` on DiffList are private, so I wasn't able to validate the actual values directly. Happy to look into making those accessible if you think it's worth it.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done. Both two-profile tests (sameAabs and diffAabs) now assert `mainSplitApkStats` and `configurationsList` for each profile as well.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Expanded to cover all AabStats properties: `universalApkStats`, `mainSplitApkStats`, and `configurationsList` are now asserted in every `computeBuildStats` test.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: I've tightened the assertions so each test now includes both presence and absence checks that are sensitive to `longSummary` and `itemLimit`. For example, the `longSummary=true` tests assert `doesNotContain("<details>")` and `contains("*Detailed file differences:*")`, while the `longSummary=false` tests assert the opposite. The low `itemLimit` test checks for `"And"` (truncation marker) while the high limit one asserts `doesNotContain("And")`.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Acknowledged — I've kept partial checks for now since with faked tools the exact output would change once we switch to real bundletool/aapt2 anyway. I think it makes more sense to add full golden-file validation after the real tools migration so we're not validating fake output. Let me know if you'd prefer I do it now regardless.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Fixed. Renamed to `setUpFakeBundleToolAndAapt2`

Sandesh282: I agree this is the right direction and would make the tests much more meaningful. I looked into it and the main challenge is generating valid AAB files at test time (bundletool needs proper BundleConfig.pb + module structure). I think the approach you suggested of compiling a minimal APK and converting to AAB using real bundletool makes sense. Would you be okay with tackling this as a follow-up, or would you prefer it as part of this PR?

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Added both missing tests (`testMain_fiveArguments_validAabs_generatesShortAndLongSummariesForOneConfig` and `testMain_eightArguments_validAabs_generatesShortAndLongSummariesForTwoConfigs`). Also added 
`testMain_oneArgument_failsWithError` which I noticed was missing from the file. Test ordering now matches #4971.

Sandesh282:
PTAL @BenHenning.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: I think you can access the values using `DiffList.size` and `DiffList.get()` and analyze the returned `DiffEntry`s. Would that work?

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: I don't think we can consider #4971 completed without that, so I actually defer to you. I'm okay with you checking in a fake implementation along with a new TODO and note on #4971 detailing the remaining work that's necessary, but I do want to be clear that without real implementations we can't be very confident that these tests can catch real problems (per my first comment on this thread above).

I would prefer a real implementation here, but it may depend on how hard that is to actually get working. If you want to tackle that then I suggest focusing on the simplest test to start (and disable all the others). Work on just getting that simple one working (probably one with a single AAB and no differences). Once that works, move on to one test with a small difference. Once that works you can expand to the remainder of the tests since it will hopefully be straightforward.

If you would like help on that feel free to commit draft changes to this PR with the other tests disabled and your WIP thoughts. I'm happy to comment on those and provide direction.

BenHenning:
Thanks @Sandesh282! I actually didn't take another review pass yet. Between my suggestion to further analyze the specific diffs and the possibility of the underlying implementation changing, there's a decent chance a lot of the test code ends up changing quite a bit so I will wait until you've addressed both comments (which, for the real vs. fake implementation comment, may actually be you choosing to not do that for this PR in which case I will provide direct feedback on the fake approach to try and 'tighten' that up a bit, as well).

Sandesh282:
Thanks @BenHenning. I think I'll keep the fake approach for this PR since the real tools migration involves setting up proper AAB generation at test time, which feels like a bigger scope change. Happy to tighten up the fakes based on your feedback. I've addressed the other comments. PTAL when you get a chance.

BenHenning:
@Sandesh282 got it. Four things then:
1. Please update the PR title & description to say that this is fixing part of #4971 rather than the whole issue (see other PRs for reference).
2. Add a TODO comment in the test pointing to #4971 and saying something along the lines of 'Replace with real AAB analysis'. This can go near the fake setup code.
3. Add a comment to #4971 detailing what work needs to be done yet after this PR is merged. Any specific suggestions that you can provide from your research would be really helpful to add, too.
4. Please address https://github.com/oppia/oppia-android/pull/6113#discussion_r2891831285 (this was the other thing I was referring to in my previous top-level comment that I wanted to see resolved before re-reviewing).

Once all of those are done, please assign back and I'll take another full pass on the test.

Sandesh282:
PTAL @BenHenning.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Alright, used DiffList.size and iterating over DiffEntrys to assert SAME_ENTRY for INTERNET and NEW_ENTRY for the added permission. Works great, thanks!

BenHenning:
Thanks. Will need to look in a day or two (apologies, this is a bit larger of a PR for me to look at since it needs a bit more careful analysis).

Sandesh282:
PTAL @BenHenning.

BenHenning:

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: So I think all that remains for the correct short/long summary tests is to actually valdiate the entire output, not just part of it. It is still valuable to verify it even with the fake implementation because:
1. It emphasizes what isn't included when diffing the two (i.e. properties that are the same).
2. It forces test updates once the implementation is swapped out (which is a good thing--we want tests to fail in those types of cases).

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: Can this be marked as resolved now @adhiamboperes?

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: Yup, I'm fine with it being deferred as also discussed in the other comment thread.

BenHenning (commented):
Thanks @Sandesh282. Just had one main comment left to address, I think. PTAL.

Sandesh282:
PTAL @BenHenning. Updated the summary tests to validate the entire output using isEqualTo() instead of partial contains() checks. This way any change to the output format will be caught, and it's clear which properties remain unchanged.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: We really ought to actually validate that this is diffing and outputting something since these are effectively "end-to-end" tests for the script. For here & the other 3 summary tests, let's make sure to not use any empty AAB stats AABs, we ought to actually have changes to diff.

BenHenning (commented):
Thanks @Sandesh282! Just missed one thing in my earlier reviews--we should be testing a bit more in the summary tests (please see my comment).

Otherwise the PR looks great, particularly the most recent changes.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: Done in the latest changes.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
BenHenning: As far as I can tell this is resolved. @adhiamboperes please reopen or leave a follow-up if more work is needed here, but resolving so that it unblocks the PR.

scripts/src/javatests/org/oppia/android/scripts/apkstats/ComputeAabDifferencesTest.kt
Sandesh282: Done.

Sandesh282:
PTAL @BenHenning

BenHenning (approved):
Thanks @Sandesh282! This LGTM.

BenHenning:
FWIW @Sandesh282 CI wasn't stalled, it's just backed up due to the Oppia organization frequently hitting out CI limits (mostly because of Oppia web's CI which is hit much harder than from Android's workflows). The last few weeks I've seen delays up to 3-4 hours, so you are best off just waiting unfortunately.

Edit: Also, I highly suggest relying on restarting workflows not the whole CI like you did just now since that will take far longer to run and has some change of flaking, anyway. If anything fails on this run we can just re-run that one thing which will unblock the PR from being merged.

Sandesh282:
@BenHenning, I couldn't find an option to re-run workflow
Screenshot 2026-03-20 at 14 01 38

BenHenning:
@Sandesh282 only team members with write access can restart workflows, but in those cases you can ask in the clam chat once you’ve been added to it.

PR #6120, opened 2026-02-28
 Fix #6086: Add transparent image detection and math tag validation

### LINKED ISSUE #6086: [Feature Request]: Add detection for transparency in lesson images & invalid math tags in download script

Images with transparency have a high chance of having night mode compatibility issues (see #5792).

We ought to add detection in the lesson download script to warn on:
- Any images that contain transparent pixels (that is, any pixels with a non-opaque alpha channel).
- Any math tags that do not contain LaTeX since that would force the app to use the math SVGs directly.

_No response_

### Additional context

_No response_

nikhilkumarpanigrahi:
## Explanation
Fixes #6086

This PR extends the existing asset download pipeline to detect:
1. **Transparent images** — pixels with alpha < 255 that cause poor visibility in dark mode
2. **Invalid math tags** — missing `raw_latex` content or using SVG fallback instead of LaTeX

## Changes
- **ImageRepairer.kt**: Added `hasTransparentPixels()` to scan image pixel data for transparency
- **StructureCompatibilityChecker.kt**: Added `checkHasValidMathTags()` with `MathTagMissingRawLatex` failure type, hooked into existing `checkHasValidHtml()` flow
- **DownloadLessons.kt**: Added post-download transparency scanning using `memoizedLoadedImageData` with count-based summary and per-image reporting
- **ImageRepairerTest.kt**: Added 7 tests for `hasTransparentPixels()`
- **StructureCompatibilityCheckerTest.kt**: Added 6 tests for `checkHasValidMathTags()`

## Essential Checklist
- [x] The PR title and explanation each start with "Fix #bugnum: " or "Fix part of #bugnum: ..."
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their calculations verified manually
- [x] The PR follows the [guide for making a code change](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR)
- [x] The PR does not decrease overall test coverage

### DIFF
--- a/scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
+++ b/scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
@@ -699,8 +699,6 @@ class LessonDownloader(
     )
     println("- ${translationIssues.size}/${issues.size} correspond to missing translations")
     println()
-    println("Note: Transparent image detection and math tag validation are performed separately.")
-    println()
     println("Images with invalid extensions:")
     imageInvalidExtIssues.groupBy { it.container }.forEach { (container, issues) ->
       println("- Within ${container.referenceString}:")
@@ -775,8 +773,11 @@ class LessonDownloader(
     val transparentImages = memoizedLoadedImageData.filter { (file, data) ->
       imageRepairer.hasTransparentPixels(data, file.extension)
     }.keys.toList()
+    println()
+    println(
+      "Transparent image check: ${transparentImages.size} image(s) found with transparency."
+    )
     if (transparentImages.isNotEmpty()) {
-      println()
       println("Images with transparent pixels (may cause dark mode visibility issues):")
       transparentImages.forEach { file ->
         println("- ${file.name}")
--- a/scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
+++ b/scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
@@ -699,6 +699,8 @@ class LessonDownloader(
     )
     println("- ${translationIssues.size}/${issues.size} correspond to missing translations")
     println()
+    println("Note: Transparent image detection and math tag validation are performed separately.")
+    println()
     println("Images with invalid extensions:")
     imageInvalidExtIssues.groupBy { it.container }.forEach { (container, issues) ->
       println("- Within ${container.referenceString}:")
@@ -770,6 +772,17 @@ class LessonDownloader(
       }
     }
 
+    val transparentImages = memoizedLoadedImageData.filter { (file, data) ->
+      imageRepairer.hasTransparentPixels(data, file.extension)
+    }.keys.toList()
+    if (transparentImages.isNotEmpty()) {
+      println()
+      println("Images with transparent pixels (may cause dark mode visibility issues):")
+      transparentImages.forEach { file ->
+        println("- ${file.name}")
+      }
+    }
+
     if (renamedImages.isNotEmpty() || convertedImages.isNotEmpty()) {
       println("WARNING: Images needed to be auto-fixed. Please verify that they are correct")
       println("(look at above output for specific images that require verification).")
@@ -777,7 +790,8 @@ class LessonDownloader(
 
     val hasAnyFailure = (
       issues.isNotEmpty() || imageDownloadFailures.isNotEmpty() ||
-        renamedImages.isNotEmpty() || convertedImages.isNotEmpty()
+        renamedImages.isNotEmpty() || convertedImages.isNotEmpty() ||
+        transparentImages.isNotEmpty()
       )
     if (hasAnyFailure && failOnError) {
       throw Exception(
--- a/scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
+++ b/scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
@@ -3,6 +3,8 @@ package org.oppia.android.scripts.gae.compat
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.AudioVoiceoverHasInvalidAudioFormat
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.HtmlInTitleOrDescription
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.HtmlUnexpectedlyInUnicodeContent
+import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.MathTagMissingRawLatex
+import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.MathTagUsingSvgFallback
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.MissingRequiredXlationLangForContentTranslation
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.MissingRequiredXlationLangForTitleOrDescFromWeb
 import org.oppia.android.scripts.gae.compat.StructureCompatibilityChecker.CompatibilityFailure.StateHasInvalidInteractionId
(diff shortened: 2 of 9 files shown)


BenHenning:
I'll need to look at this, but at a glance it seems like a strong start.

Note that this might be a bit tricky to merge because we don't have passing CI on this branch, and it may take me a bit before I can approve it because I'll need to get time to actually run the script fully. However that can happen after I take a full review pass.

nikhilkumarpanigrahi:
Thanks for the feedback, @BenHenning!
 I understand the CI situation on this branch. Happy to make any changes after your review — just let me know

scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
```diff
@@ -699,6 +699,8 @@ class LessonDownloader(
     )
     println("- ${translationIssues.size}/${issues.size} correspond to missing translations")
     println()
+    println("Note: Transparent image detection and math tag validation are performed separately.")
+    println()
```
BenHenning: This ought to be moved below and summarize the actual results (i.e. number of failing images and math tags). I also suggest treating them as two distinct checks: one is checking for transparent pixels, the other is checking for math tag correctness, so this should only cover the former here since math tag validation is done as part of the compatibility checker.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
```diff
@@ -54,7 +56,7 @@ import org.oppia.proto.v1.structure.LanguageType
 // TODO: Check image validity?
 // TODO: Check audio validity?
 // TODO: Check HTML parsability?
-// TODO: Check math exp parsability?
+// Math expression parsability is checked via checkHasValidMathTags().
```
BenHenning: It's not--math expressions are something entirely different from math tags. Please revert this line.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
```diff
@@ -608,6 +623,33 @@ class StructureCompatibilityChecker(
...
+        val rawLatex = RAW_LATEX_REGEX.find(mathContentJson)
+          ?.destructured?.let { (latex) -> latex }
+        if (rawLatex.isNullOrBlank()) {
+          failures += MathTagMissingRawLatex(contentId, origin)
+        }
+        val svgFilename = SVG_FILENAME_REGEX.find(mathContentJson)
+          ?.destructured?.let { (svg) -> svg }
+        if (!svgFilename.isNullOrBlank()) {
+          failures += MathTagUsingSvgFallback(contentId, svgFilename, origin)
+        }
```
BenHenning: This check probably isn't needed. All we really care about is that all math tags actually contain the LaTeX needed to render it.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
```diff
@@ -629,6 +671,13 @@ class StructureCompatibilityChecker(
     private val HTML_TAG_REGEX = "<\\s*([^\\s/>]+)[^>]*?>".toRegex()
     private val IMAGE_TAG_REGEX = "<\\s*oppia-noninteractive-image.+?>".toRegex()
     private val IMAGE_FILE_PATH_REGEX = "filepath-with-value\\s*=\\s*\"(.+?)\"".toRegex()
+    private val MATH_TAG_REGEX =
+      ("<\\s*oppia-noninteractive-math[^>]*?>" +
+        "[\\s\\S]*?</\\s*oppia-noninteractive-math\\s*>").toRegex()
+    private val MATH_CONTENT_VALUE_REGEX =
+      "math_content-with-value\\s*=\\s*\"(.+?)\"".toRegex()
+    private val RAW_LATEX_REGEX = "raw_latex[^:]*:\\s*\\\\?\"(.+?)\\\\\"".toRegex()
+    private val SVG_FILENAME_REGEX = "svg_filename[^:]*:\\s*\\\\?\"(.+?)\\\\\"".toRegex()
```
BenHenning: How did you derive these patterns?

scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
```diff
@@ -770,14 +772,26 @@ class LessonDownloader(
       }
     }
 
+    val transparentImages = memoizedLoadedImageData.filter { (file, data) ->
```
BenHenning, line 773: For this & the LaTeX check below, how have you tested them?

This is a WIP branch so there aren't any tests set up yet, but we might actually need to add some to validate these changes. You definitely don't need to add new tests for anything other than these changes, but I think it definitely makes sense to at least create the test suites and add the tests specific to your changes. This is fine to be limited to just `ImageRepairer` and `StructureCompatibilityChecker` (I definitely don't recommend trying to test `DownloadLessons`, that will be _much_ harder).

BenHenning (changes requested):
Thanks @nikhilkumarpanigrahi! Took a review pass--PTAL.

scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
nikhilkumarpanigrahi: Done. Removed the generic "Note" message and moved the summary below the transparent image scan results. It now prints the actual count: "Transparent image check: N image(s) found with transparency." followed by the per-image details if any are found. Math tag validation is no longer mentioned here since it's handled independently via the compatibility checker.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
nikhilkumarpanigrahi: Done. Reverted back to the original // TODO: Check math exp parsability?. I understand that math expressions are entirely different from math tags — apologies for the confusion.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
nikhilkumarpanigrahi: Done. Removed the MathTagUsingSvgFallback check entirely, including the data class, import, regex constant, and the check block. The validation now focuses only on ensuring math tags contain the LaTeX needed to render.

scripts/src/java/org/oppia/android/scripts/gae/compat/StructureCompatibilityChecker.kt
nikhilkumarpanigrahi: These regex patterns were derived by inspecting actual HTML content returned from the Oppia web API (via the GAE structure endpoints used by this script). Specifically:

MATH_TAG_REGEX: Matches <oppia-noninteractive-math>...</oppia-noninteractive-math> tags — the standard custom element Oppia uses for inline math in HTML content strings.
MATH_CONTENT_VALUE_REGEX: Extracts the math_content-with-value="..." HTML attribute, which contains JSON-encoded math content.
RAW_LATEX_REGEX: Parses the raw_latex field from within the decoded JSON content value to verify that LaTeX is actually present.
I also cross-referenced how LocalizationTracker (which has similar CUSTOM_MATH_TAG / CUSTOM_MATH_SVG_PATH_ATTRIBUTE constants) handles these tags to stay consistent. Removed SVG_FILENAME_REGEX per your feedback above.

scripts/src/java/org/oppia/android/scripts/assets/DownloadLessons.kt
nikhilkumarpanigrahi: Done. Added test suites for both:

ImageRepairerTest (7 tests): Covers opaque images, fully transparent, partially transparent, no-alpha-channel, SVG extension (case-insensitive), and mixed pixels.
StructureCompatibilityCheckerTest (6 tests): Covers HTML with no math tags, valid math tags with LaTeX, empty LaTeX, missing content, multiple tags with one invalid, and multiple valid tags.
Made checkHasValidMathTags internal (from private) to enable direct testing without requiring complex end-to-end setup through the full exploration pipeline. The checker is constructed with mock dependencies since the math tag validation only uses companion object regexes.

nikhilkumarpanigrahi:
Thanks for the detailed review @BenHenning! I've addressed all the feedback:

Moved the transparent image summary below the scan results with an actual count
Reverted the math expression TODO comment
Removed the SVG fallback check entirely (MathTagUsingSvgFallback data class, import, regex, and check block)
Added test suites for ImageRepairerTest (7 tests) and StructureCompatibilityCheckerTest (6 tests) with corresponding BUILD files
Explained the regex derivation
PTAL when you get a chance!

BenHenning:
@nikhilkumarpanigrahi please do not force push branches. This breaks reviewer history and is considered a bad practice when working on Oppia projects.

PR #6124, opened 2026-03-03
Fix: Resolve Robolectric dispatcher sync issues and remove reflection…

### LINKED ISSUE #6116: [BUG]: Fix dispatcher synchronization issues
### Describe the bug

`TestCoroutineDispatchersRobolectricImpl` has a bunch of fairly significant synchronization issues that are rather difficult to catch without specific checks in place but were discovered in #6062.

First, many tests are incorrectly trying to synchronize with `runCurrent` off of the main thread. This is not supported and can very easily cause a deadlock. Adding the following lines to `runCurrent()` and re-running the whole test suite will cause a bunch of tests to hang until Bazel terminates them:

```kotlin
    check(Looper.getMainLooper().isCurrentThread) {
      "Attempting to call runCurrent() off the main thread--this has a high chance to deadlock."
    }
```

Second, Robolectric has a substantial inconsistency between its `idle()` and `runOneTask()` behaviors. Both will only run tasks that haven't yet been scheduled however `runOneTask()` will ignore Android's internal task barrier which can wreak havoc on the Android choreographer and other internal mechanisms. There were attempts in #6062 to fix this problem by trying to bring the UI coordinator more in-line with the background and blocking test dispatchers but it's not obviously possible without using `idle()` (or some hefty reflection). It may be worth updating Robolectric to see if there's an alternative possible, but `idle()` _might_ have to be kept.

Third, not strictly an issue but the dispatcher implementations no longer need to use reflection since Gradle is no longer causing cross-dependency builds in that way anymore.

Commit 2ce2f31cc10426bf421b93c5545145a31302974c from #6062 removed these temporary changes that were done to try and address these problems. Reverting that commit into a local feature branch would be a good place to start when trying to tackle this issue.

### Steps To Reproduce

N/A

### Expected Behavior

N/A

### Screenshots/Videos

_No response_

_No response_

_No response_

_No response_

### Additional Context

_No response_

prathams-codes:
## Explanation
Fixes #6116: Resolve Robolectric dispatcher sync issues and remove reflection.

Since I do not have a local Bazel environment set up, I am opening this as a Draft PR to leverage the CI pipeline to identify which specific tests are violating the new `runCurrent()` main thread check.

**Changes made so far:**
* **Added Main Thread Check:** Inserted `check(Looper.getMainLooper().isCurrentThread)` inside `runCurrent()` to prevent deadlocks when called off the main thread.
* **Removed Reflection:** Stripped out the reflection logic (`Class.forName`, `.invoke()`) from `RobolectricUiTaskCoordinator` and replaced it with direct calls to Robolectric's `ShadowLooper`.
* **Fixed Dispatcher Bug:** Corrected a pre-existing copy-paste bug in `getNextFutureTaskTimeMillis` where `backgroundTestDispatcher` was being queried twice instead of `blockingTestDispatcher`.

## Essential Checklist
- [x] Any changes to scripts/assets files have their rationale included in the PR explanation.
- [x] The PR follows the style guide.
- [x] The PR does not contain any unnecessary code changes from Android Studio.
- [ ] The PR is **assigned** to the appropriate reviewers.

### DIFF
--- a/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersRobolectricImpl.kt
+++ b/testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersRobolectricImpl.kt
@@ -1,9 +1,10 @@
 package org.oppia.android.testing.threading
 
 import android.os.Build
+import android.os.Looper
 import androidx.annotation.RequiresApi
 import org.oppia.android.testing.time.FakeSystemClock
-import java.lang.reflect.Method
+import org.robolectric.shadows.ShadowLooper
 import java.time.Duration
 import java.util.TreeSet
 import javax.inject.Inject
@@ -31,6 +32,9 @@ class TestCoroutineDispatchersRobolectricImpl @Inject constructor(
   }
 
   override fun runCurrent() {
+    check(Looper.getMainLooper().isCurrentThread) {
+      "Attempting to call runCurrent() off the main thread--this has a high chance to deadlock."
+    }
     do {
       flushNextTasks()
     } while (hasPendingCompletableTasks())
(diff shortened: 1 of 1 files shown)


BenHenning:
@prathams-codes please don't open PRs for issues that you're not assigned. We use assignment to make sure that issues are only worked on by one person and that we know that person has a good place to start to work on the issue.

prathams-codes:
@BenHenning I apologize for opening the PR before being assigned to the issue. I wasn't aware of the assignment policy.

If the issue is still available, I would be happy to work on it after being assigned. Otherwise, I’ll look for another issue to contribute to.

Thanks for the clarification!

BenHenning:
@prathams-codes generally you should ask on the issue thread if it’s available and provide a brief explanation for how you would tackle it. Sometimes we ask clarifying questions or offer pointers before assigning the issue.

PR #6088, opened 2026-01-29
Fix #5935: Prevent submitted answer and feedback text from flashing previous card's content

Neer-rn:
<!-- READ ME FIRST: Please fill in the explanation section below and check off every point from the Essential Checklist! -->
## Explanation

Fixes #5935 
When submitting an answer, the submitted answer and feedback text briefly flashed the previous card's content before updating to the correct one. This happened because RecyclerView reuses old views and data binding updates aren't instant.

Fixed by calling `executePendingBindings()` to force immediate binding updates before HTML parsing starts. This clears the old recycled view content right away instead of showing it during the parsing delay.                          

Tested on device - no more flash. 

## Essential Checklist
<!-- Please tick the relevant boxes by putting an "x" in them. -->
- [x] The PR title and explanation each start with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] Any changes to [scripts/assets](https://github.com/oppia/oppia-android/tree/develop/scripts/assets) files have their rationale included in the PR explanation.
- [x] The PR follows the [style guide](https://github.com/oppia/oppia-android/wiki/Coding-style-guide).
- [x] The PR does not contain any unnecessary code changes from Android Studio ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#undo-unnecessary-changes)).
- [x] The PR is made from a branch that's **not** called "develop" and is up-to-date with "develop".
- [x] The PR is **assigned** to the appropriate reviewers ([reference](https://github.com/oppia/oppia-android/wiki/Guidance-on-submitting-a-PR#clarification-regarding-assignees-and-reviewers-section)).

## For UI-specific PRs only
  The video has been slowed down to 0.2x speed to make the issue and solution more visible.                                                                                                   
                                                                                                                                                                                              
  **Before**                                                                                                                                                                                  

https://github.com/user-attachments/assets/3a241518-7e70-4c39-b3ce-3cbb9d99ef8e

                                                                                                                                                                                           
  **After**                                                                                                                                                                                   

https://github.com/user-attachments/assets/e6c7d788-29ca-46af-9ed4-64ef43279303

PR #6126, opened 2026-03-04
Fix #5935: Prevent submitted answer and feedback text from flashing previous card's content

### LINKED ISSUE #5935: [BUG]: Submitted answer and Feedback text briefly flash incorrect response before updating to actual feedback

Neer-rn:
<!-- READ ME FIRST: Please fill in the explanation section below and check off every point from the Essential Checklist! -->
## Explanation

Replaces #6088 (previous PR had unrelated commits due to accidental merge instead of rebase).
Fixes #5935 
When submitting an answer, the submitted answer and feedback text briefly flashed the previous card's content before updating to the correct one. This happened because RecyclerView reuses old views and data binding updates aren't instant.

Fixed by calling `executePendingBindings()` to force immediate binding updates before HTML parsing starts. This clears the old recycled view content right away instead of showing it during the parsing delay.                          

Tested on device - no more flash. 

## Essential Checklist
<!-- Please tick the relevant boxes by putting an "x" in them. -->

## For UI-specific PRs only
  The video has been slowed down to 0.2x speed to make the issue and solution more visible.                                                                                                   
                                                                                                                                                                                              
  **Before**                                                                                                                                                                                  

https://github.com/user-attachments/assets/3a241518-7e70-4c39-b3ce-3cbb9d99ef8e

                                                                                                                                                                                           
  **After**                                                                                                                                                                                   

https://github.com/user-attachments/assets/e6c7d788-29ca-46af-9ed4-64ef43279303



  ## Test Verification Screenshots

  ### BindableAdapterTest
  **Without fix (FAILS):**
<img width="1920" height="1135" alt="Screenshot from 2026-03-11 14-48-30" src="https://github.com/user-attachments/assets/f527b9be-7071-4b9b-a1ee-67e51d0353bd" />


  **With fix (PASSES):**
<img width="1920" height="1135" alt="Screenshot from 2026-03-11 14-49-13" src="https://github.com/user-attachments/assets/4fe492aa-31f9-43ac-97eb-5956619d298b" />


  ### StateFragmentTest
  This test still passes without the fix because Robolectric handles data binding updates synchronously, so the stale content flash doesn't reproduce in the test environment.

### DIFF
--- a/app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
+++ b/app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
@@ -2,7 +2,6 @@ package org.oppia.android.app.player.state
 
 import android.content.Context
 import android.util.AttributeSet
-import android.view.LayoutInflater
 import android.view.MotionEvent
 import android.view.ViewConfiguration
 import androidx.core.view.isVisible
@@ -193,16 +192,14 @@ class DragDropSortInteractionView @JvmOverloads constructor(
     itemsInSamePositionAllowed: Boolean
   ): BindableAdapter<DragDropInteractionContentViewModel> {
     return singleTypeBuilderFactory.create<DragDropInteractionContentViewModel>()
-      .registerViewBinder(
-        inflateView = { parent ->
+      .registerViewDataBinderWithSameModelType(
+        inflateDataBinding = { inflater, parent, attachToParent ->
           viewBindingShim.provideDragDropSortInteractionInflatedView(
-            LayoutInflater.from(parent.context),
-            parent,
-            /* attachToParent= */ false
+            inflater, parent, attachToParent
           )
         },
-        bindView = { view, viewModel ->
-          viewBindingShim.setDragDropInteractionItemsBinding(view)
+        setViewModel = { binding, viewModel ->
+          viewBindingShim.setDragDropInteractionItemsBinding(binding)
           viewBindingShim.getDragDropInteractionItemsBindingRecyclerView().adapter =
             createNestedAdapter()
           adapter?.let { viewBindingShim.setDragDropInteractionItemsBindingAdapter(it) }
@@ -220,16 +217,14 @@ class DragDropSortInteractionView @JvmOverloads constructor(
 
   private fun createNestedAdapter(): BindableAdapter<String> {
     return singleTypeBuilderFactory.create<String>()
-      .registerViewBinder(
-        inflateView = { parent ->
+      .registerViewDataBinderWithSameModelType(
+        inflateDataBinding = { inflater, parent, attachToParent ->
           viewBindingShim.provideDragDropSingleItemInflatedView(
-            LayoutInflater.from(parent.context),
-            parent,
-            /* attachToParent= */ false
+            inflater, parent, attachToParent
           )
         },
-        bindView = { view, viewModel ->
-          viewBindingShim.setDragDropSingleItemBinding(view)
+        setViewModel = { binding, viewModel ->
+          viewBindingShim.setDragDropSingleItemBinding(binding)
           viewBindingShim.setDragDropSingleItemBindingHtmlContent(
             htmlParserFactory,
             resourceBucketName,
--- a/app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
+++ b/app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
@@ -193,11 +193,7 @@ class DragDropSortInteractionView @JvmOverloads constructor(
   ): BindableAdapter<DragDropInteractionContentViewModel> {
     return singleTypeBuilderFactory.create<DragDropInteractionContentViewModel>()
       .registerViewDataBinderWithSameModelType(
-        inflateDataBinding = { inflater, parent, attachToParent ->
-          viewBindingShim.provideDragDropSortInteractionInflatedView(
-            inflater, parent, attachToParent
-          )
-        },
+        inflateDataBinding = viewBindingShim::provideDragDropSortInteractionInflatedView,
         setViewModel = { binding, viewModel ->
           viewBindingShim.setDragDropInteractionItemsBinding(binding)
           viewBindingShim.getDragDropInteractionItemsBindingRecyclerView().adapter =
@@ -218,11 +214,7 @@ class DragDropSortInteractionView @JvmOverloads constructor(
   private fun createNestedAdapter(): BindableAdapter<String> {
     return singleTypeBuilderFactory.create<String>()
       .registerViewDataBinderWithSameModelType(
-        inflateDataBinding = { inflater, parent, attachToParent ->
-          viewBindingShim.provideDragDropSingleItemInflatedView(
-            inflater, parent, attachToParent
-          )
-        },
+        inflateDataBinding = viewBindingShim::provideDragDropSingleItemInflatedView,
         setViewModel = { binding, viewModel ->
           viewBindingShim.setDragDropSingleItemBinding(binding)
           viewBindingShim.setDragDropSingleItemBindingHtmlContent(
--- a/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
+++ b/app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
@@ -3,7 +3,6 @@ package org.oppia.android.app.recyclerview
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.databinding.DataBindingUtil
 import androidx.databinding.ViewDataBinding
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.LifecycleOwner
@@ -139,6 +138,10 @@ class BindableAdapter<T : Any> internal constructor(
     /**
      * Registers a [View] inflater and bind function for views in the recycler view.
      *
+     * Note that callers which use data binding internally (e.g. via [DataBindingUtil]) should use
+     * [registerViewDataBinderWithSameModelType] instead. Using this method with data binding will
+     * result in flickering due to deferred binding execution. See #5935.
+     *
      * The inflateView and bindView functions passed in here must not hold any references to UI
      * objects except those that own the RecyclerView.
      *
@@ -162,19 +165,16 @@ class BindableAdapter<T : Any> internal constructor(
         object : BindableViewHolder<T>(inflatedView) {
           override fun bind(data: T) {
             bindView(inflatedView, data)
-            // Force immediate binding execution after bindView completes. Some views registered
-            // via registerViewBinder use data binding internally (e.g. via DataBindingUtil) and
-            // set properties like htmlContent that are deferred by the binding framework. Without
-            // this call, RecyclerView may briefly display stale content from a recycled view
-            // before the binding framework flushes updates on the next frame. See: #5935.
-            DataBindingUtil.findBinding<ViewDataBinding>(inflatedView)?.executePendingBindings()
           }
         }
       }
       return this
     }
 
-    /** See [registerViewDataBinder]. */
+    /**
+     * See [registerViewDataBinder]. Binding changes are automatically synchronized after each
+     * rebind to prevent flickering from deferred data binding execution (see #5935 and #6093).
+     */
     fun <DB : ViewDataBinding> registerViewDataBinderWithSameModelType(
       inflateDataBinding: (LayoutInflater, ViewGroup, Boolean) -> DB,
       setViewModel: (DB, T) -> Unit
@@ -188,6 +188,9 @@ class BindableAdapter<T : Any> internal constructor(
     /**
      * Behaves in the same way as [registerViewBinder] except the inflate and bind methods
      * correspond to a [View] data-binding typed [DB].
+     *
+     * Binding changes are automatically synchronized after each rebind to prevent flickering from
+     * deferred data binding execution (see #5935 and #6093).
      *
      * @param inflateDataBinding a function that inflates the root view of a data-bound layout (e.g.
      *     MyDataBinding::inflate). This may also be a function that initializes the data-binding
@@ -218,6 +221,7 @@ class BindableAdapter<T : Any> internal constructor(
             // Attaching lifecycleOwner before view model initialization can sometimes cause a
             // NullPointerException because data might not be attached to the views yet.
             binding.lifecycleOwner = lifecycleOwner
+            synchronizeBindingWithRebind(binding)
           }
         }
       }
@@ -261,6 +265,11 @@ class BindableAdapter<T : Any> internal constructor(
      * specified here must be properly returned in the [ComputeViewType] function passed into
      * [Factory.create].
      *
+     * Note that callers which use data binding internally (e.g. via [DataBindingUtil]) should use
+     * [registerViewDataBinder] or [registerViewDataBinderWithSameModelType] instead. Using this
+     * method with data binding will result in flickering due to deferred binding execution. See
+     * #5935.
+     *
      * The inflateView and bindView functions passed in here must not hold any references to UI
      * objects except those that own the RecyclerView.
      *
@@ -287,12 +296,6 @@ class BindableAdapter<T : Any> internal constructor(
         object : BindableViewHolder<T>(inflatedView) {
           override fun bind(data: T) {
             bindView(inflatedView, data)
-            // Force immediate binding execution after bindView completes. Some views registered
-            // via registerViewBinder use data binding internally (e.g. via DataBindingUtil) and
-            // set properties like htmlContent that are deferred by the binding framework. Without
-            // this call, RecyclerView may briefly display stale content from a recycled view
-            // before the binding framework flushes updates on the next frame. See: #5935.
-            DataBindingUtil.findBinding<ViewDataBinding>(inflatedView)?.executePendingBindings()
           }
         }
       }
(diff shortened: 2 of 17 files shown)


Neer-rn:
@BenHenning --PTAL

Neer-rn:
@BenHenning  PTAL

app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
```diff
@@ -161,6 +162,12 @@ class BindableAdapter<T : Any> internal constructor(
         object : BindableViewHolder<T>(inflatedView) {
           override fun bind(data: T) {
             bindView(inflatedView, data)
+            // Force immediate binding execution after bindView completes. Some views registered
+            // via registerViewBinder use data binding internally (e.g. via DataBindingUtil) and
+            // set properties like htmlContent that are deferred by the binding framework. Without
+            // this call, RecyclerView may briefly display stale content from a recycled view
+            // before the binding framework flushes updates on the next frame. See: #5935.
+            DataBindingUtil.findBinding<ViewDataBinding>(inflatedView)?.executePendingBindings()
```
BenHenning: So I actually think this _does_ need to go in `registerViewDataBinder()`s, not here. It doesn't actually logically make sense here. I think the problem is that upstream callers are not actually using the correct register method. If databinding is needed then `registerViewDataBinder()` is supposed to be used. In all, I think the following changes will be needed:
- [x] A new helper should be introduced that has this comment context here and executes bindings for a `ViewDataBinding` (that way the context is kept in one place).
- [x] An aside: the comment here should be updated to point to both #5935 and #6093 (since the discussion actually has the more valuable context for reference).
- [x] Both `registerViewDataBinder()` methods should be updated to call this new helper. `registerViewBinder()`'s code shouldn't change.
- [x] The KDocs for both `registerViewDataBinder()` methods should be updated to mention that it will automatically synchronize binding changes with updates to the corresponding `RecyclerView`.
- [x] The KDocs for both `registerViewBinder()` methods should be updated to mention that if the caller is using databinding then they must use `registerViewDataBinder()` otherwise it's likely that flickering will occur.
- [x] All `registerViewBinder()` throughout the codebase need to be checked. If any (including those in `StatePlayerRecyclerViewAssembler`) use databinding then they need to be updated to use `registerViewDataBinder()` instead.
- [x] Verify that all direct tests corresponding to each change still works (for explorations run both `StateFragmentTest` and `StateFragmentLocalTest`).
- [x] Manually verify that all affected screens look correct.
- [x] Make sure to also add tests for this class per my suggestions in #6093 (though only the `registerViewDataBinder()` methods need to be verified as working correctly).

I made these a checkboxes--please check them off as you complete each one and respond when all are addressed or if you run into any problems.

BenHenning (changes requested):
Thanks @Neer-rn! Looking at this and your recent findings more, I think we need to actually do this a bit differently. The original proposal I made is probably the correct approach, but it may require changing other places in the app. Please take a look at my comment for more specific suggestions.

BenHenning:
Also: per my comment on your other PR please do not force push or rebase this PR's branch. I would really prefer not having the PR need to get reopened since my comment above is quite large and important for tracking.

app/src/main/java/org/oppia/android/app/recyclerview/BindableAdapter.kt
Neer-rn: Hey @BenHenning, I have pushed the changes you asked for. Here's what I did for each item:

  1. New helper function --> Added `synchronizeBindingWithRebind(`) as a private top-level function with KDoc referencing both #5935 and #6093.
  2. Updated comments --> References to #5935 and #6093 are included in the helper's KDoc and in the `registerViewDataBinder()` docs.
  3. Both `registerViewDataBinder()` methods updated --> Both SingleTypeBuilder and MultiTypeBuilder now call synchronizeBindingWithRebind(binding) after setting the lifecycleOwner.
  4. KDocs for `registerViewDataBinder()` --> Added note about automatic binding synchronization to prevent flickering.
  5. KDocs for `registerViewBinder()` --> Added warning that callers using data binding should use `registerViewDataBinder()` instead, otherwise flickering will happen.
  6. Audited all `registerViewBinder()` calls --> Found and migrated all callers that were using data binding internally:
    - StatePlayerRecyclerViewAssembler.kt -- 7 calls migrated
    - StoryFragmentPresenter.kt -- 1 call migrated
    - TopicLessonsFragmentPresenter.kt -- 1 call migrated
    - SurveyFragmentPresenter.kt -- 1 call migrated
    - ViewBindingShim.kt + ViewBindingShimImpl.kt -- Updated interface to pass `ViewDataBinding` instead of View
    - ComingSoonTopicsListView.kt, PromotedStoryListView.kt, SurveyMultipleChoiceOptionView.kt, SurveyNpsItemOptionView.kt, SelectionInteractionView.kt, DragDropSortInteractionView.kt -- All
  migrated to registerViewDataBinderWithSameModelType
  7. Tests pass --> `StateFragmentLocalTest` passes, `BindableAdapterTest` passes (12/12 including 2 new tests).
  8. Manual verification --> Verified in emulator as well as physical device.
  9. Added 2 new unit tests in BindableAdapterTest:
    - `testSingleTypeAdapter_withDataBinding_initialData_bindsViewSynchronously` --> verifies initial binding works
    - `testSingleTypeAdapter_withDataBinding_updateData_rebindsViewSynchronouslyWithoutRunCurrent` --> verifies that after updating data, the new text appears immediately without needing `runCurrent()`. This test would fail without `executePendingBindings()`.

  Also renamed the XML test layout ID from `text_view_for_string_no_data_binding` to `text_view_for_string_with_data_binding` since it was incorrectly named.

  No force push or rebase done, just added new commits on top. Let me know if anything needs changes!

Neer-rn:
> Also: per my comment on your other PR please do not force push or rebase this PR's branch. I would really prefer not having the PR need to get reopened since my comment above is quite large and important for tracking.

@BenHenning, yeah sorry about that, I completely understand now. I have been working on this issue for quite a while and spending hours debugging and figuring stuff out so everything was clear in my head, but I didn't really think about it from your side, like you probably deal with so many PRs and issues daily and it must be really hard to track everything if people keep closing and reopening PRs randomly. That was my bad, I will make sure to not do that again.

I haven't force pushed or rebased anything here, just added new commits on top like you asked.

  Also before merging I'll leave a detailed comment explaining the full journey, what approaches I tried, what mistakes I made along the way, and what actually fixed the issue in the end. That way even if someone visits this PR later they can understand the whole context without having to dig through multiple closed PRs and discussions.

  Thanks for the detailed review and for being patient with me throughout this, really appreciate it!

Neer-rn:
@BenHenning PTAL

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
```diff
@@ -193,16 +192,14 @@ class DragDropSortInteractionView @JvmOverloads constructor(
...
-        inflateView = { parent ->
+      .registerViewDataBinderWithSameModelType(
+        inflateDataBinding = { inflater, parent, attachToParent ->
           viewBindingShim.provideDragDropSortInteractionInflatedView(
-            LayoutInflater.from(parent.context),
-            parent,
-            /* attachToParent= */ false
+            inflater, parent, attachToParent
           )
         },
```
BenHenning:
```suggestion
        inflateDataBinding =
          viewBindingShim::provideDragDropSortInteractionInflatedView,
```

Can this be simplified to a reference? Ditto everywhere else in the PR.

app/src/main/res/layout/test_text_view_for_string_with_data_binding.xml
```diff
@@ -9,7 +9,7 @@
   </data>
 
   <TextView
-    android:id="@+id/text_view_for_string_no_data_binding"
+    android:id="@+id/text_view_for_string_with_data_binding"
```
BenHenning, line 12: Thanks for fixing this--I think I ran into it a while back but wasn't in a good PR to address it.

app/src/sharedTest/java/org/oppia/android/app/recyclerview/BindableAdapterTest.kt
```diff
@@ -568,6 +627,19 @@ class BindableAdapterTest {
...
+    measure(
+      View.MeasureSpec.makeMeasureSpec(1000, View.MeasureSpec.EXACTLY),
+      View.MeasureSpec.makeMeasureSpec(1000, View.MeasureSpec.EXACTLY)
+    )
+    layout(0, 0, 1000, 1000)
+  }
+
+  private fun RecyclerView.lookUpTextWithDataBindingAt(position: Int): String {
+    return (findViewHolderForAdapterPosition(position)?.itemView as? TextView)
+      ?.text?.toString() ?: ""
```
BenHenning: It seems fine to just return a `null` `String` here instead of defaulting (Truth's `assertThat()` will properly fail in such cases).

app/src/sharedTest/java/org/oppia/android/app/recyclerview/BindableAdapterTest.kt
```diff
@@ -433,6 +434,64 @@ class BindableAdapterTest {
...
+          getTestFragment(activity).view!!.findViewById(R.id.test_recycler_view)
+        recyclerView.measureAndLayout()
+        val displayedText = recyclerView.lookUpTextWithDataBindingAt(position = 0)
+        assertThat(displayedText).isEqualTo(STR_VALUE_0.boundStringValue)
+      }
+    }
+  }
+
+  @Test
+  fun testSingleTypeAdapter_withDataBinding_updateData_rebindsViewSynchronouslyWithoutRunCurrent() {
```
BenHenning, line 463: Please include a screenshot showing this test failing when it's run with the synchronization fix in `BindableAdapter` commented out (since the whole purpose of this test is validating that the synchronization is in place).

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
```diff
@@ -5942,6 +5942,60 @@ class StateFragmentTest {
...
+      scrollToViewType(FEEDBACK)
+      onView(withId(R.id.feedback_text_view)).check(matches(isDisplayed()))
+      clickContinueNavigationButton()
+
+      // State 3: RecyclerView recycles State 2's feedback view.
+      selectMultipleChoiceOption(optionPosition = 2, expectedOptionText = "Eagle")
+      clickSubmitAnswerButton()
+
+      // Regression test for #5935: feedback must be visible and correspond to State 3, not carry
+      // over stale content from State 2's recycled feedback view.
```
BenHenning, line 5993: For this & the previous test: will these actually fail correctly with the fix in `BindableAdapter` commented out? That's useful to also include in the PR description as a screenshot.

BenHenning (changes requested):
Thanks @Neer-rn. Just a few more comments, PTAL.

app/src/main/java/org/oppia/android/app/player/state/DragDropSortInteractionView.kt
Neer-rn: Done! Changed all the lambdas to method references wherever the lambda was just passing through the params directly.

app/src/sharedTest/java/org/oppia/android/app/recyclerview/BindableAdapterTest.kt
Neer-rn: Done, changed the return type to `String? `and removed the `?: ""` fallback. Makes sense, if the view isn't found, returning null gives a way clearer failure message than an empty string.

app/src/sharedTest/java/org/oppia/android/app/recyclerview/BindableAdapterTest.kt
Neer-rn: Added screenshots below! The test fails without the fix showing expected: Item 1 but was: Item 0 which means the RecyclerView still has stale data without executePendingBindings(). With the fix it passes correctly.

Small note: in the screenshot you might see `binding.toString()` instead of the line just being commented out when I first tried commenting out `binding.executePendingBindings()`, the build  failed because the binding parameter became unused and the project has `-Werror` so it treated the unused param warning as an error. So I used `binding.toString()` as a no-op to keep the parameter "used" while still removing the actual fix. That way the test could actually compile and run to show the failure properly.

  Without fix (FAILS):
Screenshot from 2026-03-11 14-47-15
Screenshot from 2026-03-11 14-48-30


  With fix (PASSES):
Screenshot from 2026-03-11 14-49-13

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
Neer-rn: Tried running this with `executePendingBindings()` commented out and the test still passes. The test is still useful as a regression test for making sure the correct feedback shows up in state 3, but yeah it can't specifically catch the timing issue. The `BindableAdapterTest` is the one that actually validates the fix isn't it?

BenHenning:
@Neer-rn are you still working on this? Please remember to use 'PTAL @<person>' if you want them to take another review pass.

This is sort of inadvertently blocking #6129 because that needs the core binding fix in order to validate the profile flicker fix and test.

Neer-rn:
Sorry, for the delayed response, I was travelling. All the review comments above have been addressed and it is ready for a another review. Sorry, I forgot to mention you.
PTAL @BenHenning

app/src/sharedTest/java/org/oppia/android/app/recyclerview/BindableAdapterTest.kt
BenHenning: Thanks @Neer-rn. This looks good!

app/src/sharedTest/java/org/oppia/android/app/player/state/StateFragmentTest.kt
BenHenning: It is, but I was curious if it would catch it or not. Thanks for checking!

BenHenning (approved):
Thanks @Neer-rn! This LGTM.

BenHenning:
Though @Neer-rn I think you have a failing test. I suggest merging with the latest `develop` changes and pushing to your branch to retrigger CI. Please fix any failures and assign back once CI is passing so that the PR can be merged.

BenHenning (changes requested):
(Updating to requesting changes since CI is failing).

Neer-rn:
The CI was failing because `executePendingBindings()` runs during `onBindViewHolder`, which calls `setPromotedStoryIndex()` before `onAttachedToWindow()` injects `resourceHandler`. This caused `UninitializedPropertyAccessException` in `HomeActivityTest`. Searching the root cause was very tricky for me so I took help of Gemini to find the root cause as I am not that great in searching for exact errors in log files but after I got the hint of the root cause the fix for it was very straightforward.

Fixed by deferring spotlight logic until injection is complete using `isInitialized` checks, same pattern used by `ContinueButtonView`, `LessonThumbnailImageView` etc. Also checked other custom views with same pattern, they already have guards or are safe.

The other 2 failures (`ProfileManagementControllerTest`, `TopicFragmentTest`) are flaky, both pass locally.

All the checks have passed. PTAL @BenHenning.

BenHenning (approved):
Thanks @Neer-rn. The PR LGTM!

> The CI was failing because `executePendingBindings()` runs during `onBindViewHolder`, which calls `setPromotedStoryIndex()` before `onAttachedToWindow()` injects `resourceHandler`. This caused `UninitializedPropertyAccessException` in `HomeActivityTest`. Searching the root cause was very tricky for me so I took help of Gemini to find the root cause as I am not that great in searching for exact errors in log files but after I got the hint of the root cause the fix for it was very straightforward.
> 
> Fixed by deferring spotlight logic until injection is complete using `isInitialized` checks, same pattern used by `ContinueButtonView`, `LessonThumbnailImageView` etc. Also checked other custom views with same pattern, they already have guards or are safe.
> 
> The other 2 failures (`ProfileManagementControllerTest`, `TopicFragmentTest`) are flaky, both pass locally.
> 
> All the checks have passed. PTAL @BenHenning.

That's interesting. This might mean we have other races that could cause problems which is really good to keep in mind. I suspect that your change actually makes things more, not less, deterministic though so hopefully this is the only such case.

PR #6136, opened 2026-03-09
Fix part of #5663: Use local thumbnails in dev builds

### LINKED ISSUE #5663: [Feature Request]: Remove JSON files & fix activity thumbnails

With #5652 and the removal of Gradle, we no longer need to support loading JSON lesson files.

Removal of all JSON lesson files and corresponding JSON loading code, along with removal of any gating code/logic across the codebase (i.e. ``LoadLessonProtosFromAssets`` can be completely removed).

Note that while the clean-up is straightforward, we can't actually move forward with this change without also fixing the fact that thumbnails do not load correctly with proto files. This needs to be fixed properly as well (ideally by introducing some sort of dev-specific override for loading certain thumbnails so that we can force them to be local, dev-only versions).

There's not a strong alternative for the JSON clean-up as proto is the predominant format the app will be using for lessons moving forward. However, the exact approach for fixing the thumbnail issue is unclear, and there are likely many alternative solutions possible.

### Additional context

_No response_

harshsomankar123-tech:
## Explanation
Fix #5663 Part 01: Fixes activity thumbnail loading in developer builds by introducing a debug-only ContentProvider.

**Context:**
Currently, when `thumbnailFilename` is set in the proto files, it constructs a GCS URL and tries to fetch the image from the internet. Because developer builds don't have access to these GCS assets, the thumbnail loading fails. (The JSON loading path previously avoided this by falling back to local `LessonThumbnailGraphic` drawables, but that path is being removed).

**Changes in this PR:**
* Introduces a developer/debug-only `ContentProvider` that intercepts thumbnail requests. For developer builds, it serves the local drawable assets instead of attempting to fetch GCS images.
* Adds a debug-only Dagger module to configure this specific behavior for developer builds, ensuring the current GCS pipeline remains completely unchanged in production.

*(Note: This PR strictly handles the thumbnail fix for dev builds. The JSON loading path removal is handled separately).*

## Essential Checklist


harshsomankar123-tech:
@BenHenning @adhiamboperes  PTAL

BenHenning:

app/src/main/java/org/oppia/android/app/customview/LessonThumbnailImageView.kt
```diff
@@ -101,7 +106,7 @@ class LessonThumbnailImageView @JvmOverloads constructor(
     } else {
       listOf()
     }
-    if (lessonThumbnail.thumbnailFilename.isNotEmpty()) {
+    if (lessonThumbnail.thumbnailFilename.isNotEmpty() && loadThumbnailsFromGcs) {
```
BenHenning: Have you tested this with proto lessons @harshsomankar123-tech? I wouldn't expect this to work, actually. I left more thoughts in https://github.com/oppia/oppia-android/issues/5663#issuecomment-4078001763. PTAL.

BenHenning (changes requested):
@harshsomankar123-tech I left one comment on the PR and another on the issue. I don't think that this is being approached quite the correct way, PTAL.

Also, your PR description says that you've solved this with a custom `ContentProvider` but no `ContentProvider` has actually been proposed in code here.

harshsomankar123-tech:
@BenHenning Thanks for the suggestions. I’ve made the changes mentioned in your comments.PTAL

app/src/main/java/org/oppia/android/app/application/dev/ThumbnailContentProvider.kt
```diff
@@ -0,0 +1,111 @@
...
+
+  override fun query(
+    uri: Uri,
+    projection: Array<String>?,
+    selection: String?,
+    selectionArgs: Array<String>?,
+    sortOrder: String?
+  ): Cursor? = null
+
+  override fun getType(uri: Uri): String = "image/png"
```
BenHenning: Is this correct? Perhaps these could be SVGs, instead?

app/src/main/java/org/oppia/android/app/utility/imageloading/ThumbnailContentProvider.kt
```diff
@@ -0,0 +1,111 @@
...
+     * textproto asset files.
+     */
+    private val THUMBNAIL_FILENAME_TO_DRAWABLE_MAP = mapOf(
+      "baker.img" to R.drawable.lesson_thumbnail_graphic_baker,
+      "child_with_book.img" to R.drawable.lesson_thumbnail_graphic_child_with_book,
+      "child_with_cupcakes.img" to R.drawable.lesson_thumbnail_graphic_child_with_cupcakes,
+      "child_with_fractions_homework.img" to
+        R.drawable.lesson_thumbnail_graphic_child_with_fractions_homework,
+      "duck_and_chicken.img" to R.drawable.lesson_thumbnail_graphic_duck_and_chicken,
+      "person_with_pie_chart.img" to R.drawable.lesson_thumbnail_graphic_person_with_pie_chart
```
BenHenning, line 48: Perhaps it would be cleaner to represent these as SVGs? We converted them to drawables originally because of how they were hooked up, but SVGs could be embedded as assets and returned directly (avoiding any temporarily file creation or loading needed).

app/src/main/java/org/oppia/android/app/application/dev/ThumbnailContentProvider.kt
```diff
@@ -0,0 +1,111 @@
+package org.oppia.android.app.application.dev
```
BenHenning: I think `org.oppia.android.app.utility.imageloading` is probably an okay place for now. This is a new pattern so we don't have an obvious place for it to go, but the current package is meant for top-level application setup which doesn't seem like the correct place.

The new package will need to be a proper Bazel package (i.e. with a `BUILD.bazel`) file, and you'll need to reference that new library target as a dependency for the top-level `//:oppia_dev` target as one of its dependencies.

domain/src/main/assets/GJ2rLXRKD5hw.textproto
```diff
@@ -7,6 +7,7 @@ subtopic_ids: 4
 is_published: true
 has_practice_questions: true
 topic_thumbnail {
+  thumbnail_filename: "child_with_fractions_homework.img"
```
BenHenning, line 10: Here & elsewhere: how did you decide on which thumbnails to use? We ought to match exactly the current JSON behavior to make it an easier before-and-after check.

app/src/main/AndroidManifest.xml
```diff
@@ -369,6 +369,10 @@
     <activity
       android:name=".app.testing.TextInputLayoutBindingAdaptersTestActivity"
       android:theme="@style/OppiaThemeWithoutActionBar" />
+    <provider
+      android:name=".app.application.dev.ThumbnailContentProvider"
+      android:authorities="org.oppia.android.provider"
+      android:exported="false" />
```
BenHenning: This needs to go into its own `AndroidManifest.xml` in the same package as the content provider. We will rely on the manifest merge to include this at the end. Adding it here means we need to ship the provider in production which we absolutely don't want to do.

utility/src/main/java/org/oppia/android/util/parser/image/DevImageParsingModule.kt
```diff
@@ -0,0 +1,38 @@
...
+ * thumbnail images to be served from local drawable resources in dev builds where GCS assets are
+ * inaccessible.
+ */
+@Module
+class DevImageParsingModule {
+  @Provides
+  @DefaultGcsPrefix
+  @Singleton
+  fun provideDefaultGcsPrefix(): String {
+    return "content://org.oppia.android.provider"
```
BenHenning: Let's use `content://org.oppia.android.provider/gcs` so as to not pollute future content provider potential.

utility/src/main/java/org/oppia/android/util/parser/image/BUILD.bazel
```diff
@@ -117,6 +117,21 @@ kt_android_library(
     ],
 )
 
+kt_android_library(
+    name = "dev_image_parsing_module",
+    srcs = [
+        "DevImageParsingModule.kt",
```
BenHenning: The convention in the codebase is to use 'debug' for developer dependencies, and to put it at the end of the name before module, e.g. `ImageParsingDebugModule`. Please make sure to also update the library target name.

app/src/main/java/org/oppia/android/app/application/dev/ThumbnailContentProvider.kt
```diff
@@ -0,0 +1,111 @@
...
+  companion object {
+    private const val AUTHORITY = "org.oppia.android.provider"
+    private const val THUMBNAIL_MATCH = 1
+    private val uriMatcher = UriMatcher(UriMatcher.NO_MATCH).apply {
+      // Match any path under the authority — the thumbnail filename is extracted from the
+      // last path segment of the URI.
+      addURI(AUTHORITY, "*/*/*/*/#", THUMBNAIL_MATCH)
+      addURI(AUTHORITY, "*/*/*/*", THUMBNAIL_MATCH)
+      addURI(AUTHORITY, "thumbnail/*", THUMBNAIL_MATCH)
+      addURI(AUTHORITY, "#", THUMBNAIL_MATCH)
```
BenHenning: Are all of these needed? It seems like we will subpath based on `<entity_type>/<entity_id>/assets/<image|thumbnail>/<file.name>` which seems like it only needs a single URI matcher, or am I missing something?

BenHenning (changes requested):
Thanks @harshsomankar123-tech! This is definitely on the right track. I have a bunch of comments, PTAL.

Also, after you address all of my comments could you add a video to the PR description showing loading with protos having correct thumbnails? Please make sure to correctly set up loading via `CachingModule` since protos aren't loaded by default in local dev builds currently.

adhiamboperes:
@harshsomankar123-tech, please feel free to share any issues you are facing with this PR here or in Github Discussions.

harshsomankar123-tech:
@adhiamboperes The CI is failing right now. I’ll work on fixing it and provide an update shortly.

adhiamboperes:
@harshsomankar123-tech, will you be finishing this PR, or prefer to close it for now?

harshsomankar123-tech:
Hi @adhiamboperes, I’m currently a bit busy, so I’ll close this PR for now. I can revisit it later if needed. Thanks!

Issue #209, opened 2019-10-07
Verify that number with units classification is correct


BenHenning:
Numbers with unit [classification](https://github.com/oppia/oppia/blob/37285a9b0e7866cbcdf5f8193a6ac2fab64458a7/extensions/interactions/NumberWithUnits/directives/number-with-units-rules.service.ts) involves converting the answer an input to a Mathjs-compatible string, then parsing it back into an actual object for comparison. It's not clear why this is necessary. We should figure out if it is, and adjust the number-with-units classification in the Android app as necessary.

BenHenning:
This may also include translating certain units for normalization (e.g. cm -> centimeters).

PR #6137, opened 2026-03-10
Fix part of #209: Introduce a tokenizer and parser for NumberWithUnits

theMr17:
## Explanation

Fixes part of #209
  
This PR introduces a `NumberWithUnitsTokenizer` that converts raw string input into a `Sequence<Token>`. It supports numeric literals, operators/symbols (-, /, ^, *, parentheses), currency symbols/prefixes/suffixes, SI prefixes, and a broad set of physical units.  The tokenizer is implemented with the existing `PeekableIterator`. This keeps behavior aligned with LL(1) parsing constraints and avoids unnecessary processing when parsing fails early.

Further the `Sequence<Token>` is passed to the `NumberWithUnitsParser`, which is responsible for enforcing the grammar rules as mentioned [here](https://docs.google.com/document/d/1PF1LdzBUwNO3tSqucCoMdYpPOWUySI3yS4m_knfLdtE/edit?tab=t.0#heading=h.bl4j28uatldq). Error handling of the parser is explicit and non-exception based via `NumberWithUnitsParsingResult.Success/Failure`, with parser failures mapped to domain-specific `NumberWithUnitsParsingError` values (e.g. missing exponent/denominator/unit, unbalanced parentheses, invalid tokens, trailing tokens)

A tokenizer Truth subject: `NumberWithUnitsTokenSubject` is added that provides typed assertion helpers for every relevant token shape (numeric tokens, symbols, units, invalid tokens). This reduces repetitive casting and improves readability in tokenizer tests. And a `NumberWithUnitsSubject` with helpers for asserting numeric representation (real vs fraction) and unit structure (count, names, name-based unit lookup, exponent assertions). Both custom subject files include TODOs to add dedicated tests for the subjects themselves (#6151). These helpers are straightforward wrappers around type checks and proto field assertions; follow-up dedicated subject tests are tracked separately to keep PR scope manageable.

## Essential Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for parsing and tokenizing numbers with attached units, including prefixes, suffixes, fractions, currency symbols, and unit exponents.
  * Introduced new test helpers for checking parsed values and token types in unit-related math expressions.

* **Bug Fixes**
  * Improved validation for malformed expressions, invalid symbols, missing units, unbalanced parentheses, and trailing content.
  * Expanded allowed test-file exemptions so the new math-related tests pass validation.


theMr17:
Hey @adhiamboperes, this PR is now ready for review. PTAL!

BenHenning:
Is it failing to let this go into review?

utility/src/main/java/org/oppia/android/util/math/BUILD.bazel
```diff
@@ -225,3 +225,43 @@ kt_android_library(
...
+    deps = [
+        "//model/src/main/proto:interaction_object_java_proto_lite",
+    ],
+)
+
+kt_android_library(
+    name = "number_with_units_tokenizer",
+    srcs = ["NumberWithUnitsTokenizer.kt"],
+    visibility = [
+        "//:oppia_api_visibility",
```
BenHenning, line 261: Does this need full API visibility? I feel like the tokenizer is an implementation detail so it doesn't quite make sense to broadly expose it.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1609 @@
+package org.oppia.android.util.math
+
+import org.oppia.android.util.math.PeekableIterator.Companion.toPeekableIterator
+
+/**
+ * A tokenizer for parsing mathematical expressions containing numbers and units.
+ *
+ *
```
BenHenning: Nit: extra unnecessary doc line.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1609 @@
...
+      ) : Token()
+
+      /** Represents a liter unit. */
+      class LiterUnit(
+        override val startIndex: Int,
+        override val endIndex: Int
+      ) : Token()
+
+      /** Represents a cubic centimeter unit (cc). */
+      class CcUnit(
```
BenHenning: `CubicCentimeterUnit` for consistency?

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1609 @@
...
+          integerEndIndex
+        )
+      }
+    }
+
+    /**
+     * Parses a sequence of digits into a string representation of an integer.
+     *
+     * This method consumes consecutive digit characters and allows whitespace between digits.
+     * It returns null if no digits are found at the current position.
```
BenHenning: This bit feels odd to me. This would allow the expression "1 2 cm" to be interpreted as 12 centimeters. Is that actually correct? Usually whitespace isn't allowed between consecutive numbers in math when they're meant to form a single value.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1609 @@
...
+    /**
+     * Tokenizes a unit or SI prefix starting from the current position.
+     *
+     * The method supports both singular and plural forms, abbreviated forms,
+     * and different case conventions where applicable.
+     *
+     * @param chars the [PeekableIterator] positioned at the start of a unit
+     * @return the appropriate [Token] representing the unit or an [Token.InvalidToken]
+     */
+    private fun tokenizeUnit(chars: PeekableIterator<Char>): Token {
```
BenHenning: Just to check, and apologies if we did discuss this in the document already, but why can't we just have a single `Token` type for 'Unit' and capture the full consecutive, non-whitespace values as the unit name (similar to what we do for variable/function names in math expressions)? This might actually simplify things a bit because then the parser can just switch via a `when` on all expected unit values when it encounters a `Unit` token to validate them and return the correct unit element.

For comparison, we need additional parsing at the tokenizing level for math expressions to deal with function names not being distinguishable from variables by default, and the distinction between a variable (single letter) and a function (more than one letter). Those constraints are different than here so it may be possible to simplify things.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1609 @@
...
+      ) : Token()
+
+      /** Represents a right parenthesis symbol, i.e. ')'. */
+      class RightParenthesisSymbol(
+        override val startIndex: Int,
+        override val endIndex: Int
+      ) : Token()
+
+      /** Represents a dollar currency prefix, e.g. '$'. */
+      class DollarPrefixUnit(
```
BenHenning: Here & below: maybe drop the prefix/suffix bits. That's true at the grammar level but from the tokenizer's perspective these are just symbols that need further interpretation to ensure they're in the correct place. Something like 'DollarSymbol' would probably work (and ditto for Rupee since if you can follow my suggestion above the 'Rs' will actually be interpreted as a different token and consolidated in the parser).

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,522 @@
...
+  private val rawExpression: String,
+  private val tokens: PeekableIterator<Token>
+) {
+  /**
+   * Grammar:
+   * ```
+   * number_with_units = prefix_formatted_value | suffix_formatted_value ;
+   * ```
+   */
+  private fun parseNumberWithUnits(): NumberWithUnitsParsingResult<NumberWithUnits> {
```
BenHenning: I think it's absolutely worthwhile to redefine `NumberWithUnits` to include the actual units we expect.

This may cause some friction with the importing code so my suggestion is to add new protos specific to the parser and we can convert them to `NumberWithUnits` upstream when performing classification. Then we can have a TODO issue tracking moving everything over to the new proto (which will also require changes in the lesson download script and a bunch of other places in the codebase, hence why I don't think we should figure all of that out now).

The nice thing about a new proto structure is it explicitly defines precisely what units are supported and is something we can actually enforce during lesson import time to ensure creators somehow don't use an unsupported unit.

As far as naming, I'm not quite sure what to call it to properly distinguish it from the existing class. You could consider renaming the existing proto to `NumberWithUnitsInteractionObject` (bit gross, but gets the point across) or find a better name for the new proto.

BenHenning (changes requested):
Thanks @theMr17! I took a first pass. I didn't review the test code yet, and I'm happy it's there and passing because I do have a number of refactoring suggestions that the tests can help verify. PTAL and let me know your thoughts.

Once the high-level structural suggestions are addressed I will take a deeper dive on the tokenizer and especially the parser, plus their tests if I don't have another round of high-level suggestions.

utility/src/main/java/org/oppia/android/util/math/BUILD.bazel
theMr17: The tokenizer (specifically the token classes) are imported in the `NumberWithUnitsTokenSubject` class as well (apart from the parser). So, I think this needs a full visibility.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: Removed.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: I just wanted to keep it as flexible as possible for the user. However, not allowing whitespace between digits should be more logical. Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: @BenHenning just to confirm my understanding: are you proposing to use a single `UnitToken` that captures the full consecutive non-whitespace string as the unit (stored as something like `parsedUnit`), instead of having separate token types for each unit?

From my perspective, this could introduce some challenges. Currently, the tokenizer handles normalization of different representations of the same unit. For example, a minute unit can appear as `min`, `mins`, `minute`, or `minutes`, all of which are mapped to a single `MinuteToken`. If we move to a single `UnitToken`, this normalization logic would likely need to shift to the parser, which might not necessarily simplify the overall design and could make the parser more complex.

Also, I'm not entirely clear on this part and would like some clarification:
> then the parser can just switch via a `when` on all expected unit values when it encounters a `Unit` token to validate them and return the correct unit element.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: As mentioned in https://github.com/oppia/oppia-android/pull/6137#discussion_r2978732563, if we pass `Rs` and `₹` as separate tokens, the parser would then need to handle mapping both of them into a common representation (e.g., a `RupeePrefixToken`). This effectively shifts normalization logic from the tokenizer to the parser, which could increase parser complexity rather than simplify the design.

Regarding the prefix/suffix distinction, I believe this separation is necessary and shouldn't be removed. Prefix and suffix forms of currency are not always interchangeable. For example, `$` (prefix) and `dollar` (suffix) follow different positional rules in expressions. Treating them uniformly at the tokenizer level could make it harder to enforce correct syntax later during parsing.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Thanks for the suggestion! I think this would be a nice refactor. Just wanted to confirm that you want to have something like an `enum` of all the supported unit in the new proto structure?

theMr17:
Thanks for the review @BenHenning! I have addressed some of your comments. However, I left some comments seeking clarificaton on the rest. PTAL.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
BenHenning: I was thinking this would be normalized in the parser, yeah. So your specific example might look like this:

```kt
private fun parseMinuteUnit(token: Token.Unit): Minute {
  return when (token.name) {
    "min", "mins", "minute", "minutes" -> ... build Minute ...
    else -> ... error ...
  }
}
```

(Obviously it would use the 'result' pattern like done broadly for parsing)

This also demonstrates the 'when' concept I was thinking. It does mean a larger parser but it also ensures all logic related to a specific unit is handled in one place (including for '$' vs 'dollars' vs 'usd', etc.).

Thoughts?

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
BenHenning: I replied to the other comment per the former point.

As to the latter point I think that comes down to convention and where is the best/easiest place to handle errors. I'm usually an advocate of keeping tokenizers simple because errors that happen at that level usually lack semantic context which means it's hard to tell the user what they did wrong. The parser, however, can tell much more easily: a '$' appeared later on when it wasn't expected. The parser actually knows what a prefix is, the tokenizer doesn't (it's technically just calling '$' a prefix even though it's not asserting that it's truly showing up at the correct location in the character stream to be a prefix).

I think that last bit is what really makes me think it doesn't make sense to use the term here: the tokenizer isn't actually guaranteeing 'prefix' state and assuming so could actually be detrimental to the parser (not that it is, but one could misinterpret it when looking at this in the future).

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
BenHenning: Either an `enum` or `oneof` depending on what makes the most sense when you actually figure out the proto. [Math expressions](https://github.com/oppia/oppia-android/blob/develop/model/src/main/proto/math.proto) used a combination of them. I suspect you'll probably be more in the `enum` camp because there aren't really property differences between the different units, but the `oneof` might come up for fraction vs. real and maybe parts of the actual expression structure (e.g. I could imagine you having three different variants to the number structure: a prefix/value, value/suffix, prefix/value/suffix variant).

BenHenning:
Thanks @theMr17. I only followed up to comments this round since the larger coding questions are still in-flight and will potentially lead to a lot more code churn that would be better to review once it's done. PTAL.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: I have introduced two new proto definitions: `NumberWithUnitsExpression` and `NumberUnitExpression`. The `NumberUnitExpression` includes three variants to handle different prefix/suffix combinations, and also defines a separate field for `SiPrefix`.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: That sounds like a better approach than the current implementation. I've updated the tokenizer to emit a single `Unit` token for all units, along with a separate `SiPrefix` token. The remaining handling is now delegated to the parser.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: I've implemented this approach in the parser. Could you please take a look and confirm if this aligns with your expectations? Happy to make further changes if needed.

theMr17:
Hey @BenHenning, I have addressed all the remaining comments, PTAL!

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,1350 @@
...
+    /**
+     * Tokenizes a unit or SI prefix starting from the current position.
+     *
+     * The method supports both singular and plural forms, abbreviated forms,
+     * and different case conventions where applicable.
+     *
+     * @param chars the [PeekableIterator] positioned at the start of a unit
+     * @return the appropriate [Token] representing the unit or an [Token.InvalidToken]
+     */
+    private fun tokenizeUnit(input: String, chars: PeekableIterator<Char>): Token {
```
BenHenning, line 151: This is quite the tokenization method!

How often do we see consecutive units without whitespace or other symbols between them? I'm wondering, for instance, if it's possible to simply consider all runs of letters to be a 'possible unit'. For example, we could tokenize the following expression "12N m/sqft" as:
- Integer: 12
- Possible unit: N
- Possible unit: m
- Divide
- Possible unit: sqft

The parser could then realize that:
- Units N and m next to each other indicates implicit multiplication.
- Unit sqft can be interpreted as ft^2 and be treated as such.

This does put further strain on the parser again but this is closer to what I had envisioned. If there's some impossibility for why we can't do that (e.g. there are other combined or sequential units that must be separated at this stage) then I can see the case for doing that, otherwise it does make sense to push more into the parser for better contextual error handling.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+   * `compound_unit`.
+   */
+  private fun parseNumber(
+    expressionBuilder: NumberWithUnitsExpression.Builder
+  ): NumberWithUnitsParsingResult<Unit> {
+    var isNegative = false
+    if (tokens.peek() is Token.MinusSymbol) {
+      tokens.next() // consume '-'
+      isNegative = true
+    }
```
BenHenning:
```suggestion
    val isNegative = if (tokens.peek() is Token.MinusSymbol) {
      true.also { tokens.next() } // consume '-'
    } else false
```

Debatable if this 'better' but avoiding 'var' is generally a good strategy since it reduces how much mutable state needs to be verified/contextualized as correct.

Alternatively you could follow the same two-line pattern you use below for parentheses.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+            -firstToken.parsedValue.toDouble()
+          } else {
+            firstToken.parsedValue.toDouble()
+          }
+          expressionBuilder.real = value
+          NumberWithUnitsParsingResult.Success(Unit)
+        }
+      }
+
+      else -> NumberWithUnitsParsingError.NumberExpectedError.toFailure()
```
BenHenning: Could just return `NumberExpectedAfterCurrencyPrefixError` here and avoid the extra `if` check above, I think.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+    val numberResult = parseNumber(expressionBuilder)
+    if (numberResult is NumberWithUnitsParsingResult.Failure) return numberResult
+
+    // compound_unit is required
+    if (!isAtSuffixUnit()) {
+      return NumberWithUnitsParsingError.UnitExpectedError.toFailure()
+    }
+
+    val compoundUnits = parseCompoundUnit()
+    if (compoundUnits is NumberWithUnitsParsingResult.Failure) return compoundUnits
```
BenHenning: Optional but you could pull over the `map`/`flatMap` and related functions that are in `MathExpressionParser` for its result structure and use them throughout this parser. They eliminate a bunch of the extra `if` checks because they can help short-circuit operations for failure results.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+   * ```
+   *
+   * Notes that [Token.PositiveRealNumber] is already tokenized as a single token, so decimal
+   * parsing does not happen here. `integer/integer` is always interpreted as a fractional numeric
+   * literal at this stage. Unit division (for example `m/s`) is parsed later as part of
+   * `compound_unit`.
+   */
+  private fun parseNumber(
+    expressionBuilder: NumberWithUnitsExpression.Builder
+  ): NumberWithUnitsParsingResult<Unit> {
```
BenHenning: Slight nit but this is a but of an inversion: most of the patterns of this parser are meant to be no side-effect other than for the tokenizer itself. I suggest keeping that strategy by returning the result rather than mutating an incoming builder like this.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+          // and then a unit or end-of-input. But we need to distinguish "1/2 kg" (fraction)
+          // from "5 m/s" (unit division). Since we are in parseNumber, the divide here always
+          // means a fractional number.
+          tokens.next() // consume '/'
+          val denomToken = tokens.peek()
+          if (denomToken is Token.PositiveInteger) {
+            tokens.next()
+            val fraction = Fraction.newBuilder().apply {
+              this.isNegative = isNegative
+              wholeNumber = 0
```
BenHenning, line 162: Are whole numbers ever valid for numbers with units?

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+          } else {
+            NumberWithUnitsParsingError.MissingDenominatorError.toFailure()
+          }
+        } else {
+          val value = if (isNegative) {
+            -firstToken.parsedValue.toDouble()
+          } else {
+            firstToken.parsedValue.toDouble()
+          }
+          expressionBuilder.real = value
```
BenHenning:
```suggestion
          expressionBuilder.real = if (isNegative) {
            -firstToken.parsedValue.toDouble()
          } else {
            firstToken.parsedValue.toDouble()
          }
```

Optional simplification.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+      if (tokens.peek() is Token.MinusSymbol) {
+        tokens.next() // consume '-'
+        negativeExponent = true
+      }
+
+      val expToken = tokens.peek()
+      if (expToken is Token.PositiveInteger) {
+        tokens.next()
+        val exponentMultiplier = if (negativeExponent)
+          -expToken.parsedValue
```
BenHenning, line 351: Multi-line `if` and `else` statements should always have curly braces, otherwise please try to keep them on one line.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+   */
+  private fun parseUnitWithExponent(): NumberWithUnitsParsingResult<NumberUnitExpression> {
+    val parsedUnitResult = parseSuffixUnit()
+    if (parsedUnitResult is NumberWithUnitsParsingResult.Failure) return parsedUnitResult
+    val parsedUnit = (parsedUnitResult as NumberWithUnitsParsingResult.Success).result
+    var finalExponent = parsedUnit.exponent
+    if (tokens.peek() is Token.ExponentiationSymbol) {
+      tokens.next() // consume '^'
+
+      var negativeExponent = false
```
BenHenning: Ditto here & anywhere else with a negative check.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+    val currencyUnit = when (rawUnit) {
+      "$" -> NumberUnitExpression.Unit.DOLLAR
+      "₹", "Rs" -> NumberUnitExpression.Unit.RUPEE
+      else -> return null
+    }
+    tokens.next()
+    return createUnitExpression(currencyUnit)
+  }
+
+  private fun parsePrefixableBaseUnit(token: Token?): NumberUnitExpression? {
```
BenHenning: Maybe use an extension function on `Unit` for checking this--might be a bit cleaner.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+      NumberUnitExpression.Unit.VOLT,
+      NumberUnitExpression.Unit.OHM -> parsed
+
+      else -> null
+    }
+  }
+
+  private fun parseAnyUnit(token: Token?): NumberUnitExpression? {
+    val raw = (token as? Token.Unit)?.unit ?: return null
+    return parseMeterUnit(raw)
```
BenHenning: As much as I generally encourage minimizing method sizes I think in this case it may actually be preferable to use a single `when` with sorting so that it's easier to see all of the different units that are supported for a raw `Token.Unit` (and to ensure no overlap).

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+      NumberUnitExpression.Unit.PASCAL,
+      NumberUnitExpression.Unit.AMPERE,
+      NumberUnitExpression.Unit.VOLT,
+      NumberUnitExpression.Unit.OHM -> parsed
+
+      else -> null
+    }
+  }
+
+  private fun parseAnyUnit(token: Token?): NumberUnitExpression? {
```
BenHenning: Maybe just return the `NumberWithUnitExpression.Unit` value? Seems perhaps a bit simpler since the whole expression is assembled piecewise.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+        }
+      }
+
+      else -> {
+        val parsedUnit = parseAnyUnit(tokens.peek())
+        if (parsedUnit != null) {
+          tokens.next()
+          NumberWithUnitsParsingResult.Success(parsedUnit)
+        } else {
+          NumberWithUnitsParsingError.UnitExpectedError.toFailure()
```
BenHenning: This is a bit misleading and makes me wonder if `parseAnyUnit` is returning the correct thing. The error should be a unit is expected if no unit follows but `parseAnyUnit` also returns `null` if an _invalid_ unit is encountered. We ought to handle both distinctly since they have different remediations.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+    "c" -> NumberUnitExpression.SiPrefix.CENTI
+    "m" -> NumberUnitExpression.SiPrefix.MILLI
+    "u" -> NumberUnitExpression.SiPrefix.MICRO
+    "n" -> NumberUnitExpression.SiPrefix.NANO
+    "p" -> NumberUnitExpression.SiPrefix.PICO
+    "f" -> NumberUnitExpression.SiPrefix.FEMTO
+    "a" -> NumberUnitExpression.SiPrefix.ATTO
+    "z" -> NumberUnitExpression.SiPrefix.ZEPTO
+    "y" -> NumberUnitExpression.SiPrefix.YOCTO
+    else -> null
```
BenHenning: Ditto here--should this perhaps indicate invalid rather than missing?

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,685 @@
...
+  private fun isCurrencyUnit(unit: NumberUnitExpression.Unit): Boolean =
+    unit == NumberUnitExpression.Unit.DOLLAR ||
+      unit == NumberUnitExpression.Unit.CENT ||
+      unit == NumberUnitExpression.Unit.RUPEE ||
+      unit == NumberUnitExpression.Unit.PAISA
+
+  private fun isCurrencyPrefixToken(token: Token?): Boolean {
+    val rawUnit = (token as? Token.Unit)?.unit ?: return false
+    return rawUnit == "$" || rawUnit == "₹" || rawUnit == "Rs"
+  }
```
BenHenning: Similarly these could also perhaps be extension functions.

BenHenning (changes requested):
Thanks @theMr17! Apologies for the delayed review, I should be able to be more responsive now.

I've taken a pass on the tokenizer again, and a quicker pass on the parser. Both are looking very solid but I did have a few suggestions to perhaps simplify things a bit and improve robustness. PTAL. I also found the proto approach to be much cleaner so I'm glad to see that!

Also if it keeps things simpler don't worry about getting the tests passing if you just want my feedback on approach changes (since I know some of my suggestions, particularly for the tokenizer, may warrant a bunch more testing changes).

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: No, whole numbers are never non-zero for number with units.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Updated according to your suggestion.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: I have added `map` and `flatmap` functions to the parser and have updated the entire parser to use those.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Introduced a new extension function:
```kotlin
private fun NumberUnitExpression.Unit.isPrefixable(): Boolean
```

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Added braces.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Thanks for pointing this out! Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Updated with the suggestion above.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: I've updated the tokenizer and parser to follow this approach. However, this introduces a limitation where units must be separated by a space or another delimiter; otherwise, they are treated as invalid. For example, `12 N m/sqft` is parsed correctly, whereas `12Nm/sqft` results in a parsing error.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: I have added a new error: `InvalidUnitError` which handles this case.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: I was initially planning to introduce a separate error type, `InvalidSiPrefixError`. However, with the current approach, that doesn't seem feasible.

To explain why, consider a case where the parser receives a unit like `InvalidMeter`, where `Invalid` could be interpreted as an invalid SI prefix. The parser first checks whether the unit starts with a valid SI prefix. If it finds one, it then attempts to parse the remaining part as a valid SI-prefixable base unit.

If no valid SI prefix is found, the parser instead treats the entire string as a unit and tries to parse it directly. If this also fails, it ultimately throws an `InvalidUnitError`.

Because of this flow, there isn't a clear point where we can confidently distinguish between an invalid SI prefix and an entirely invalid unit, making it difficult to introduce a dedicated `InvalidSiPrefixError`.

theMr17:
Thanks for the review, @BenHenning! It took longer than expected to address all the comments. The tokenizer tests are currently failing; I'll address those once the overall approach is confirmed.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,287 @@
...
+      val startIndex = chars.getRetrievalCount()
+      while (
+        chars.peek()?.let {
+          it in 'a'..'z' ||
+            it in 'A'..'Z' ||
+            it in '2'..'3' // Needed to tokenize units like m2 and m3
+        } == true
```
BenHenning: These ranges can be expanded out into a constant set and then we can just use `chars.peek() in VALID_UNIT_CHARS` or something similar to keep this code simpler.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
BenHenning: Why would spaces next to the digits matter? I'd expect `12m` and `12 m` to both produce `PositiveInteger` then `Unit` token stream.

As for ambiguously combined units like "Nm" we probably will need to handle that in the parser for the obvious cases, and fail for the others (which does seem like the correct approach, I think). This may mean the parser needs to be able to turn a single potential `Unit` (like "Nm") into multiple interpreted units.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,616 @@
...
+  private fun parseCompoundUnit(): NumberWithUnitsParsingResult<List<NumberUnitExpression>> {
+    return parseUnitsMultiplied().flatMap { numeratorUnits ->
+      val allUnits = numeratorUnits.toMutableList()
+
+      // Optional: division_operator , denominator_expression
+      if (tokens.peek() is Token.DivideSymbol) {
+        tokens.next() // consume '/'
+        parseDenominatorExp().map { denomUnits ->
+          allUnits.addAll(denomUnits)
+          allUnits
```
BenHenning:
```suggestion
          allUnits + denomUnits
```

Bit simpler, same memory overhead, and allows you to avoid declaring `allUnits` altogether since a mutable list isn't needed.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,616 @@
...
+    return createUnitExpression(currencyUnit)
+  }
+
+  private fun parsePrefixableBaseUnit(token: Token?): NumberUnitExpression.Unit? {
+    val unit = parseAnyUnit(token) ?: return null
+    return if (unit.isPrefixable()) unit else null
+  }
+
+  private fun parseAnyUnit(token: Token?): NumberUnitExpression.Unit? {
+    val raw = (token as? Token.Unit)?.unit ?: return null
```
BenHenning: You can probably just make the whole method non-nullable and perform the null check higher up. Could we pass in Token.Unit here?

Other helpers might be able to be similarly simplified.

BenHenning (commented):
Thanks @theMr17! I don't have many concerns. I left one thought on the combined unit case that might be worth considering, otherwise I think it's reasonable to work to finalize the PR and I'll then take a pass on the rest of it (including the tests).

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Done.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: I have updated the parser, it can now handle consecutive units without spaces.

theMr17:
Hi @BenHenning, I have addressed your comments. PTAL!

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,291 @@
...
+     */
+    fun tokenize(input: String): Sequence<Token> {
+      val chars = input.toCharArray().asSequence().toPeekableIterator()
+      return generateSequence {
+        chars.consumeWhitespace()
+
+        when (chars.peek()) {
+          in '0'..'9' -> tokenizeIntegerOrRealNumber(chars)
+          in 'a'..'z', in 'A'..'Z' -> tokenizeUnit(input, chars)
+          '-', '−', '–' -> tokenizeSymbol(chars) { startIndex, endIndex ->
```
BenHenning, line 35: For the third character, why en-dash?

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
```diff
@@ -0,0 +1,152 @@
...
+  private val actual: Token
+) : Subject(metadata, actual) {
+
+  /**
+   * Verifies that the token is a positive integer and returns an [IntegerSubject] for the value.
+   *
+   * @return an [IntegerSubject] to verify the integer value
+   * @throws AssertionError if the token is not a [Token.PositiveInteger]
+   */
+  fun isPositiveIntegerWhoseValue(): IntegerSubject {
```
BenHenning: Consider here & below to follow the "...WithValueThat()" pattern to be consistent, but I don't feel strongly about this if we use this sort of "...WhoseValue()" pattern in other subjects.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
```diff
@@ -0,0 +1,152 @@
...
+    actual.asVerifiedType<Token.InvalidToken>()
+  }
+
+  /**
+   * Verifies that the token is a [Token.Unit] with exactly the specified raw value.
+   *
+   * @param expectedValue the exact raw unit text expected from tokenization
+   * @throws AssertionError if the token is not a [Token.Unit] or has a different raw value
+   */
+  fun isUnitWithRawValue(expectedValue: String) {
```
BenHenning: Perhaps return a StringSubject for consistency?

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
```diff
@@ -0,0 +1,152 @@
...
+    val unit = actual.asVerifiedType<Token.Unit>()
+    Truth.assertThat(unit.unit).isEqualTo(expectedValue)
+  }
+
+  /**
+   * Verifies that the token is a [Token.Unit] whose raw value is in [expectedRawValues].
+   */
+  private fun assertUnitIn(vararg expectedRawValues: String) {
```
BenHenning: Not used anymore?

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,705 @@
...
+  }
+
+  /**
+   * Parses a single unit token, returning one or more [NumberUnitExpression]s.
+   *
+   * First attempts standard single-unit parsing (with optional exponent). If that fails due to
+   * an unrecognized unit, attempts to decompose the token into multiple known units.
+   */
+  private fun parseUnitsFromNextToken(): NumberWithUnitsParsingResult<List<NumberUnitExpression>> {
+    val singleResult = parseUnitWithExponent()
```
BenHenning, line 281: How will exponents work as part of decomposition? The order seems a bit off to me. Should exponent checking come after decomposition rather than replacing it?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
```diff
@@ -0,0 +1,1924 @@
...
+
+  @Test
+  @Iteration("₹ 10", "input=₹ 10", "expected=10")
+  @Iteration("Rs 10", "input=Rs 10", "expected=10")
+  fun testTokenize_rupeePrefixUnit_parsesCorrectly() {
+    val tokens = NumberWithUnitsTokenizer.tokenize(input).toList()
+
+    assertThat(tokens).hasSize(2)
+    assertThat(tokens[0]).isUnitWithRawValue(input.substringBefore(" "))
+    assertThat(tokens[1]).isPositiveIntegerWhoseValue().isEqualTo(expected.toInt())
```
BenHenning: Here & elsewhere in the test: we probably ought to not be validating based on the input.

Also, since the number is always 10, you could consider using expected to store the expected prefix and then just verify 10 directly.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
```diff
@@ -0,0 +1,1924 @@
...
+    assertThat(tokens[1]).isUnitWithRawValue("cuft")
+  }
+
+  @Test
+  fun testTokenize_incorrectCubicFootUnits_parsesInvalidToken() {
+    val tokens = NumberWithUnitsTokenizer.tokenize("10 Cuft").toList()
+
+    assertThat(tokens).hasSize(2)
+    assertThat(tokens[0]).isPositiveIntegerWhoseValue().isEqualTo(10)
+    assertThat(tokens[1]).isUnitWithRawValue("Cuft") // u
```
BenHenning: Extra trailing comment?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
```diff
@@ -0,0 +1,1924 @@
...
+    assertThat(tokens).hasSize(2)
+    assertThat(tokens[0]).isPositiveIntegerWhoseValue().isEqualTo(10)
+    assertThat(tokens[1]).isUnitWithRawValue(input.substringAfter(" "))
+  }
+
+  @Test
+  @Iteration("10 Mol", "input=10 Mol")
+  @Iteration("10 Mole", "input=10 Mole")
+  @Iteration("10 Moles", "input=10 Moles")
+  fun testTokenize_incorrectMoleUnits_parsesMegaPrefix() {
```
BenHenning: Is the result here (`_parsesMegaPrefix`) actually correct? Ditto above for `testTokenize_incorrectMinuteUnits_parsesMegaPrefix`.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
```diff
@@ -0,0 +1,1924 @@
...
+  fun testTokenize_correctNewtonUnits_parsesCorrectly() {
+    val tokens = NumberWithUnitsTokenizer.tokenize(input).toList()
+
+    assertThat(tokens).hasSize(2)
+    assertThat(tokens[0]).isPositiveIntegerWhoseValue().isEqualTo(10)
+    assertThat(tokens[1]).isUnitWithRawValue(input.substringAfter(" "))
+  }
+
+  @Test
+  fun testTokenize_incorrectNewtonUnits_parsesInvalidToken() {
```
BenHenning: Here & everywhere else: shouldn't it be `parsesInvalidUnit`? It's not actually producing a token indicating invalid so the test name doesn't seem to match the actual code which is potentially confusing.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5085 @@
...
+// FunctionName: test names are conventionally named with underscores.
+@Suppress("FunctionName")
+@RunWith(AndroidJUnit4::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+@Config
+class NumberWithUnitsParserTest {
+  @Test
+  fun testParser_emptyString_returnsEmptyExpressionError() {
+    val error = parseNumberWithUnitsExpectingFailure("")
+    assertThat(error).isInstanceOf(NumberWithUnitsParsingError.EmptyExpressionError::class.java)
```
BenHenning: I suggest direct importing the various `NumberWithUnitsParsingError`s if possible to help with readability.

This also might help you wrap some of the assertions so that they don't require 3 lines.

BenHenning (changes requested):
Apologies for the delayed review. I had this mostly done but then ran into GitHub's PR outage last week and couldn't finish it until now.

I've fully reviewed the whole PR except for the parser test in detail (largely due to its length and limited time). I took a quick glance for any high-level concerns and left just one comment for now. I plan to take a closer review of that file in the next review pass.

Overall the PR is looking quite good overall. Thanks @theMr17!

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
theMr17: Just wanted to cover everything that could potentially be used as a minus symbol by the user.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
theMr17: Removed.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
theMr17: Updated with the suggested pattern.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
theMr17: Done.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
theMr17: Removed.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Imported all the error classes directly.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
theMr17: Removed all assertions with the input.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
theMr17: Updated all the test names to `...parsesCorrectly`.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
theMr17: Updated both the test names.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Thanks for pointing this out! This behavior was actually intentional, since `Nm^2` is ambiguous--it could mean either `N * m^2` or `(N * m)^2`. Because of that ambiguity, the parser currently treats it as invalid. During decomposition, it produces `[N, m]`, leaving `^2` unconsumed, which then results in a `TrailingTokensError`. This behavior is currently being tested by [this test](https://github.com/theMr17/oppia-android/blob/f5b44e89b1234f0cc42614a39a9b49fb0b8d287b/utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt#L322-L328).

That said, there is an alternative approach we could take: interpret `Nm^2` as applying the exponent only to the meter unit (`N * m^2`), while requiring users to explicitly write `(Nm)^2` if they intend the exponent to apply to the entire compound unit.

Would you prefer that I implement the latter approach instead?

theMr17:
Hey @BenHenning, I have addressed your comments, PTAL!

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+        hasExponentThat().isEqualTo(1)
+      }
+    }
+  }
+
+  @Test
+  fun testParser_compoundUnit_withExponent_returnsTrailingTokensError() {
+    // "Nm^2" is ambiguous (N*m^2 vs (N*m)^2), so the parser should fail.
+    // Decomposition produces [N, m], then "^2" remains unconsumed.
+    val error = parseNumberWithUnitsExpectingFailure("5 Nm^2")
```
BenHenning: This case might lead to user confusion. How does mathjs handle it?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+      }
+    }
+  }
+
+  @Test
+  fun testParser_compoundUnit_spacedEquivalent_matchesCompound() {
+    // "N m" (space-separated) should produce the same result as "Nm" (compound).
+    val spacedResult = parseNumberWithUnitsExpectingSuccess("50 N m")
+    val compoundResult = parseNumberWithUnitsExpectingSuccess("50 Nm")
+    assertThat(compoundResult).apply {
```
BenHenning: You could also use proto lite subject's is equal to for these types of tests, I think, for confidence that the results are entirely and deeply equal.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+        hasExponentThat().isEqualTo(1)
+      }
+    }
+  }
+
+  @Test
+  fun testParser_compoundUnit_negativeNumber_parsesCorrectly() {
+    val result = parseNumberWithUnitsExpectingSuccess("-5 Nm")
+    assertThat(result).apply {
+      hasRealValueThat().isEqualTo(-5.0)
```
BenHenning: Everywhere here & in the tokenizer tests: should we consistently be using isWithin-esque assertions for real numbers to avoid skew?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+      }
+      hasSuffixWithIndexThat(index = 2).apply {
+        hasUnitThat().isEqualTo(Unit.METER)
+        hasExponentThat().isEqualTo(-1)
+      }
+    }
+  }
+
+  @Test
+  fun testParser_compoundUnit_wattSecond_spacedEquivalent_matchesCompound() {
```
BenHenning, line 860: Ditto here & for other equivalence checks.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+
+  @Test
+  fun testParser_fractionOnlyNoUnit_returnsUnitExpectedError() {
+    val error = parseNumberWithUnitsExpectingFailure("1/2")
+    assertThat(error).isInstanceOf(UnitExpectedError::class.java)
+  }
+
+  @Test
+  fun testParser_dollarPrefix_negativeFraction_parsesCorrectly() {
+    val result = parseNumberWithUnitsExpectingSuccess("$-1/4")
```
BenHenning: Does the other direction parse, too (negative before currency)?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+      hasUnitCountThat().isEqualTo(1)
+      hasSuffixWithIndexThat(index = 0).apply {
+        hasUnitThat().isEqualTo(Unit.PAISA)
+        hasExponentThat().isEqualTo(1)
+      }
+    }
+  }
+
+  @Test
+  fun testParser_paiseCapitalized_parsesCorrectly() {
```
BenHenning: Here & elsewhere: you have lots of groups of tests doing the same thing with just the value deviating (or can be made to just be the unit in most cases). Not a hard requirement, but it could be nice to leverage parameterized tests for these cases. It would pretty substantially reduce the length of the test file without compromising readability. However, I consider it optional since a test purist argument might be to never use parameterized tests, anyway, and it's extra work for minimal benefit (test maintenance and auditing, though I suspect we'll rarely do this since hopefully the parser rarely needs to be updated).

Edit: if you decide to act on this feedback, please do so in a separate commit after all the other changes are addressed so that I can review it separately since it will otherwise hide other test changes.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -0,0 +1,5036 @@
...
+
+  @Test
+  fun testParser_doubleDivision_returnsTrailingTokensError() {
+    // "10 m/s/kg" → after parsing m/s, the second '/' is a trailing token
+    val error = parseNumberWithUnitsExpectingFailure("10 m/s/kg")
+    assertThat(error).isInstanceOf(TrailingTokensError::class.java)
+  }
+
+  @Test
+  fun testParser_divisionWithNoNumeratorUnit_returnsUnitExpectedError() {
```
BenHenning: Test name doesn't match error tpye in outcome.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
BenHenning: How does mathjs handle this? I would imagine we should resolve to something predictable and be consistent about that since this doesn't seem like an actual error situation. However, I think I lean toward `Nm^2` being `N*m^2` since that seems a bit closer to cases like `m/Ns^2` where the `^2` attaches just to the second but both the Newton and the second are divided and thus get negative exponents (though I see the case here that this may be the same problem, anyway, so this may be circular reasoning).

I asked Gemini and it agrees with my preference but for a better and different reason: we should follow PEMDAS rules for operator precedence here with multiplications and left-hand associativity (or right-hand in the case expontentiation).

Edit: Here's my Gemini conversation. It suggests the grammar is correct but the parser may not have implemented it correctly, hence the discrepancy. See: https://share.gemini.google/zGXS1nGTUzy1.

BenHenning (changes requested):
Thanks @theMr17! Really sorry about the delay. I was hoping to get this wrapped up before starting traveling, but I didn't and traveling + a conference/work collaboration + GSoC catch-up has used up all of my time until now. I should be able to review subsequent changes quickly now, so hopefully we can get this wrapped up soon.

The PR looks fantastic. I just had a few comments, and there's one open discussion about precedence that needs to be addressed. I also will see if code rabbit can do a review pass since it would be nice to get an LLM to validate the grammar implementation more concretely, though I have no clue if it will do a good job.

BenHenning:
@coderabbitai review

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Done

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Updated these checks to allow a tolerance of up to 1e-5.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Done

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: No it doesn't parse the other way. Do we need to support that?

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Updated this and an another test which had a naming inconsistency.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
theMr17: Thanks for the detailed analysis and the Gemini conversation link! I've updated the parser to follow PEMDAS and SI conventions. Now, when the parser encounters something like `Nm^2`, it decomposes it and applies the exponent only to the immediately preceding unit (so it correctly parses as `N.(m²)` instead of throwing a `TrailingTokensError`). I've also added several test cases to verify this behavior, including negative exponents and divisions.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsSubject.kt
```diff
@@ -0,0 +1,181 @@
...
+      NumberWithUnitsExpression.ExpressionFormatCase.PREFIX_VALUE_SUFFIX_EXPRESSION ->
+        actual.prefixValueSuffixExpression.prefixUnitsList
+
+      else -> emptyList()
+    }
+    assertWithMessage("Expected to find at least one prefix unit, but found none")
+      .that(units)
+      .isNotEmpty()
+    return NumberUnitSubject.assertThat(units.first())
+  }
```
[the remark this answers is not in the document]
theMr17: The parser supports only a single prefix unit. There can never me multiple prefix units with the current setup.

testing/src/main/java/org/oppia/android/testing/math/NumberWithUnitsTokenSubject.kt
```diff
@@ -0,0 +1,144 @@
...
+
+// TODO(#6151): Add tests for this class.
+
+/**
+ * A Truth [Subject] for testing [Token] objects from [NumberWithUnitsTokenizer].
+ */
+class NumberWithUnitsTokenSubject(
+  metadata: FailureMetadata,
+  private val actual: Token
+) : Subject(metadata, actual) {
```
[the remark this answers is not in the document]
theMr17: Done

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,705 @@
...
+          finalExponent *= exponentMultiplier
+        } else {
+          return@flatMap NumberWithUnitsParsingError.MissingExponentError.toFailure()
+        }
+      }
+      NumberWithUnitsParsingResult.Success(
+        parsedUnit.toBuilder().setExponent(finalExponent).build()
+      )
+    }
+  }
```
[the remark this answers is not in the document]
theMr17: Good catch! I've updated the parser so that it only attempts compound unit decomposition if the initial single-unit parse didn't consume any tokens. This ensures that partially parsed units with missing exponents (like `kg^N`) correctly bubble up a `MissingExponentError` instead of falling back to decomposition and losing the original error context. I've also added tests to verify these edge cases.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsParser.kt
```diff
@@ -0,0 +1,705 @@
...
+          } else {
+            firstToken.parsedValue.toDouble()
+          }
+          NumberWithUnitsParsingResult.Success(expressionBuilder)
+        }
+      }
+
+      else -> NumberWithUnitsParsingError.NumberExpectedAfterCurrencyPrefixError.toFailure()
+    }
+  }
```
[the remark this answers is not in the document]
theMr17: I've updated `parseNumber(`) to accept a `hasCurrencyPrefix` context flag and updated it to throw the correct `NumberExpectedError` when the flag is false, otherwise it throws `NumberExpectedAfterCurrencyPrefixError`.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * A tokenizer for parsing mathematical expressions containing numbers and units.
+ *
+ * The tokenizer supports whitespace between tokens and can handle various unit formats
+ * including both singular and plural forms, abbreviated forms, and different naming conventions.
+ *
+ * See https://docs.google.com/document/d/1PF1LdzBUwNO3tSqucCoMdYpPOWUySI3yS4m_knfLdtE for the
+ * various units supported by this tokenizer.
+ */
```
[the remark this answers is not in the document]
theMr17: Updated.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,291 @@
...
+
+    /**
+     * Tokenizes a number (either integer or real) starting from the current position.
+     *
+     * This method handles both integers (e.g., "123") and real numbers with decimal points (e.g., "123.45").
+     * Whitespace is allowed between digits and around the decimal point.
+     *
+     * @param chars the peekable iterator positioned at the start of a number
+     * @return a [Token.PositiveInteger], [Token.PositiveRealNumber], or [Token.InvalidToken]
+     */
```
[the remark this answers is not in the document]
theMr17: Updated.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsTokenizerTest.kt
```diff
@@ -0,0 +1,1922 @@
...
+  @Test
+  @Iteration("10 °C", "input=10 °C")
+  @Iteration("10 °", "input=10 °")
+  @Iteration("10 Ω", "input=10 Ω")
+  fun testTokenize_unknownSymbols_parsesCorrectly() {
+    val tokens = NumberWithUnitsTokenizer.tokenize(input).toList()
+
+    assertThat(tokens[0]).isPositiveIntegerWithValueThat().isEqualTo(10)
+    assertThat(tokens[1]).isInvalidToken()
+  }
```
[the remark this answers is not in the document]
theMr17: This seems to be unnecessary.

utility/src/main/java/org/oppia/android/util/math/NumberWithUnitsTokenizer.kt
```diff
@@ -0,0 +1,291 @@
...
+     * @return true if all characters match exactly, false otherwise
+     */
+    private fun PeekableIterator<Char>.expectNextCharsForUnit(
+      chars: String
+    ): Boolean {
+      for (c in chars) {
+        expectNextValue { c } ?: return false
+      }
+      return true
+    }
```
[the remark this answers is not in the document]
theMr17: The unused function is now removed.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: The test is now updated, based on the other comment related to this.

theMr17: However, mathjs does not support units without spaces between them.
image

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Thanks for the suggestion! I agree it's worth it for the file size reduction. I've updated the file to use `OppiaParameterizedTestRunner` for the repetitive test groups (abbreviations, spelled-out forms, case variants, and SI prefixes).

This compressed 118 individual tests into 4 parameterized methods, saving almost 1400 lines of code. The total test execution count remains exactly the same.

For you reference: https://github.com/oppia/oppia-android/pull/6137/commits/dc26ae204da9858e454a285e3212e43977f75db1

theMr17:
Thanks for the review @BenHenning! I have addressed all the comments in your review as well as Rabbit's review. PTAL!

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
```diff
@@ -805,2114 +859,766 @@ class NumberWithUnitsParserTest {
...
+  @Iteration("integerWithCubicFoot", "input=5 cuft", "expectedUnit=CUBIC_FOOT")
+  @Iteration("integerWithCubicYard", "input=3 cuyd", "expectedUnit=CUBIC_YARD")
+  @Iteration("integerWithSquareMeter", "input=20 m2", "expectedUnit=SQUARE_METER")
+  @Iteration("integerWithCubicMeter", "input=8 m3", "expectedUnit=CUBIC_METER")
+  @Iteration("integerWithMinute", "input=30 min", "expectedUnit=MINUTE")
+  @Iteration("integerWithHour", "input=2 hr", "expectedUnit=HOUR")
+  fun testParser_abbreviatedUnit_parsesCorrectly() {
+    val result = parseNumberWithUnitsExpectingSuccess(input)
     assertThat(result).apply {
-      hasRealValueThat().isEqualTo(300.0)
```
BenHenning: We could probably keep this by updating the values to use the same number, or by adding an expected value to compare against.

Ditto for `testParser_spelledOutOrPluralUnit_parsesCorrectly` and the other parameterized tests below.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
BenHenning: Interesting...I suspect this will be fine since it means we're more, not less, permissive on Android (and we can always make it more restrictive though that could result in some awkwardness with checkpoints, but I think that's fine). I'm not inclined to suggest a major change for this.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
BenHenning: Similarly, let's match Math.js here (if it doesn't support it, then let's keep current behavior otherwise let's update to support it).

BenHenning (commented):
Thanks @theMr17! Sorry again for the delay...finding time to review has been challenging. I've closed out the two coderabbit conversations since I agree nothing more was needed on those. All other threads are closed except one existing one (regarding the negative sign relative to a prefix currency symbol) and one new one (to try and avoid the small test coverage regression that was introduced with parameterization).

Functionally, the PR looks good. Once these last two conversation points are addressed I will be happy to approve the PR.

Note that I will be unavailable for future reviews for the next two months so please send everything over to other Oppia Android reviewers (hopefully this will actually result in faster turnarounds since it won't be blocked on me).

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Mathjs does not support currency units at all. Keeping it as it is.

BenHenning: Perhaps a better question then is whether Oppia web supports this since it is configuring its Math.js integration to work with currency units.

BenHenning:
FYI I added another reply to the currency thread @theMr17. It would also be helpful to ensure CI is passing when you send this back for final review check-off since I'm assuming at least some of these failures are probably not flakes and will require actual changes.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: @BenHenning Just tried negative sign on the web, it seems like negative sign is not allowed at all with a currency unit.

image

image

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
theMr17: Done.

theMr17:
@BenHenning PTAL if you are available.

utility/src/test/java/org/oppia/android/util/math/NumberWithUnitsParserTest.kt
BenHenning: Got it. I suppose we should consider being more restrictive here but I don't have a strong feeling here, either. There's one argument for consistency on the negative sign but I think it's fine to handle that later, if at all. I think getting this PR merged and the feature progressing so that it can be tested is far more important.

BenHenning (approved):
Thanks @theMr17, this LGTM! Very well done!

PR #6148, opened 2026-03-12
Fix #5818: Robolectric edittext input action

### LINKED ISSUE #5818: [BUG]: Fix Robolectric append/replaceText handling in EditTextInputAction and add corresponding tests.

Sandesh282:
Fixes #5818

## Explanation

Fixes #5818: Refactors `EditTextInputAction` to remove its dependency on constructor injection of `TestCoroutineDispatchers`, which caused Robolectric edit text input actions to be difficult to invoke reliably in tests.

This change introduces `TestCoroutineDispatchersInjector`, which provides a static access point for test coroutine dispatchers. `EditTextInputAction` is updated to retrieve dispatchers through this injector instead of relying on injected fields.

As a result, `EditTextInputAction` methods can now be invoked directly from tests without requiring `@Inject` setup, resolving the existing TODO in the file and simplifying test usage across Robolectric-based tests.

### Key changes
- Added `TestCoroutineDispatchersInjector` to hold a static reference to
  `TestCoroutineDispatchers`.
- Updated `TestDispatcherModule` to initialize the injector.
- Refactored `EditTextInputAction` to access dispatchers via the static injector.
- Updated test usages to call `EditTextInputAction` methods directly.
- Removed `EditTextInputAction` injection from test classes.

This is a duplicate of https://github.com/oppia/oppia-android/pull/6082, which contains previous reviews but was closed due to force-pushing.

## Essential Checklist

- [x] The PR title starts with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The explanation section above starts with "Fixes #bugnum: " (If this PR fixes part of an issue, use instead: "Fixes part of #bugnum: ...".)

## For UI-specific PRs only

This PR does not introduce any UI-related changes. It only affects test infrastructure and
Robolectric-based test utilities. As a result, no UI verification, screenshots, or accessibility
checks are required.

### DIFF
--- a/scripts/assets/test_file_exemptions.textproto
+++ b/scripts/assets/test_file_exemptions.textproto
@@ -4276,10 +4276,6 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
-  exempted_file_path: "testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt"
-  test_file_not_required: true
-}
-test_file_exemption {
   exempted_file_path: "testing/src/main/java/org/oppia/android/testing/espresso/GenericViewMatchers.kt"
   test_file_not_required: true
 }
--- a/scripts/assets/test_file_exemptions.textproto
+++ b/scripts/assets/test_file_exemptions.textproto
@@ -4276,6 +4276,10 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
+  exempted_file_path: "testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
   exempted_file_path: "testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt"
   is_extra_test_file: true
 }
--- a/scripts/assets/test_file_exemptions.textproto
+++ b/scripts/assets/test_file_exemptions.textproto
@@ -4887,3 +4887,11 @@ test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/player/state/listener/ReturnToQuestionButtonListener.kt"
   test_file_not_required: true
 }
+test_file_exemption {
+  exempted_file_path: "testing/src/main/java/org/oppia/android/testing/threading/TestCoroutineDispatchersInjector.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
+  exempted_file_path: "testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt"
+  is_extra_test_file: true
+}
--- a/scripts/assets/test_file_exemptions.textproto
+++ b/scripts/assets/test_file_exemptions.textproto
@@ -15,10 +15,6 @@ test_file_exemption {
   override_min_coverage_percent_required: 0
 }
 test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/home/recentlyplayed/RecentlyPlayedActivity.kt"
-  source_file_is_incompatible_with_code_coverage: true
-}
-test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/activity/ActivityIntentFactoriesModule.kt"
   test_file_not_required: true
 }
@@ -279,6 +275,10 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/classroom/ThumbnailImage.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/classroom/classroomlist/AllClassroomsHeaderText.kt"
   test_file_not_required: true
 }
@@ -292,10 +292,6 @@ test_file_exemption {
 }
 test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/classroom/promotedlist/PromotedStoryList.kt"
-  test_file_not_required: true
-}
-test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/classroom/ThumbnailImage.kt"
   test_file_not_required: true
 }
 test_file_exemption {
@@ -383,6 +379,14 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragment.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragmentPresenter.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/AppRestartListener.kt"
   test_file_not_required: true
 }
@@ -427,23 +431,15 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersButtonClickListener.kt"
+  test_file_not_required: true
+}
+test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersDialogFragment.kt"
   test_file_not_required: true
 }
 test_file_exemption {
   exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersDialogFragmentPresenter.kt"
-  test_file_not_required: true
-}
-test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/ForceDownloadRemoteParametersButtonClickListener.kt"
-  test_file_not_required: true
-}
-test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragment.kt"
-  test_file_not_required: true
-}
-test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/AppRestartDialogFragmentPresenter.kt"
   test_file_not_required: true
 }
 test_file_exemption {
@@ -455,7 +451,7 @@ test_file_exemption {
   test_file_not_required: true
 }
 test_file_exemption {
-  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/SavePendingChangesDialogListener.kt"
+  exempted_file_path: "app/src/main/java/org/oppia/android/app/devoptions/RouteToFeatureFlagsListener.kt"
   test_file_not_required: true
 }
 test_file_exemption {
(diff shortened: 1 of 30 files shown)


Sandesh282:
@adhiamboperes This is a clean re-submission replacing #6082. All previous review feedback has been addressed:

- Fully imported `appendText` and `replaceText` instead of using `EditTextInputAction.Companion.` prefix (Ben's and your feedback)
- Verified Ben's comment on `TestCoroutineDispatchersInjector.kt`, the static injector was simplified but retained as Ben concluded it's the pragmatic approach before Hilt migration
- Removed all unrelated changes 
PTAL when you get a chance. Thanks!

scripts/assets/test_file_exemptions.textproto
adhiamboperes, line 1: This looks out of place @Sandesh282, PTAL.

adhiamboperes (changes requested):
Thanks @Sandesh282! All the test files look good to me now.

I am handing over the final reviews to @BenHenning to look at the TestCoroutineDispatchersInjector changes and the EditTextInputActionRobolectricTest.kt, specifically.

scripts/assets/test_file_exemptions.textproto
adhiamboperes: Always check the files changed tab of your PR to confirm that the files there are all relevant. This is required by the PR checklist.

Sandesh282: Done, reverted the unrelated changes. PTAL @adhiamboperes when you get a chance!

BenHenning:

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
```diff
@@ -19,50 +18,47 @@ import javax.inject.Inject
...
-  // the TestCoroutineDispatchers so that the outer class doesn't need to be injected.
-  /**
-   * Returns a [ViewAction] that appends the specified string into the view targeted by the
-   * [ViewAction].
-   */
-  fun appendText(text: String): ViewAction = updateText(
-    text, baseAction = typeText(text), isAppend = true
-  )
+class EditTextInputAction @Inject constructor() {
+  companion object {
```
BenHenning: This can just be a top-level object, no need for companion. Plus, the class is never injected so it doesn't make sense to have an `@Inject` constructor.

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
```diff
@@ -19,50 +18,47 @@ import javax.inject.Inject
...
+        override fun perform(uiController: UiController?, view: View?) {
+          // Appending text only works on Robolectric, whereas Espresso needs to use typeText().
+          if (Build.FINGERPRINT.contains("robolectric", ignoreCase = true)) {
+            val editText = view as? EditText
+            if (isAppend) {
+              editText?.append(text)
+            } else {
+              editText?.setText(text)
+            }
+            TestCoroutineDispatchersInjector.getDispatcher().runCurrent()
```
BenHenning: The correct way to do this is to introduce an injector interface and provider, then implement them in test applications throughout the codebase. I was hesitant to suggest this because it will require changing a lot of files, and since the rest of the PR is done, I am fine with you checking in the current solution provided:
1. You file an issue to migrate `TestCoroutineDispatchersInjector` to the correct pattern.
2. You add a code TODO pointing to that issue in `TestCoroutineDispatchersInjector` and mention that the class should not be used in any new locations.
3. You restrict Bazel access to the injector specifically to this class (which, since it isn't modularized, means you'll need to restrict it to `//testing/src/main/java/org/oppia/android/testing:__subpackages__` I think).

For a reference on a correctly set up injector provider, see: https://github.com/oppia/oppia-android/blob/develop/utility/src/main/java/org/oppia/android/util/system/OppiaClockInjector.kt and https://github.com/oppia/oppia-android/blob/develop/utility/src/main/java/org/oppia/android/util/system/OppiaClockInjectorProvider.kt.

With the injector pattern properly in place, you can get access to the underlying object using the following approach:

```kotlin
val context = ApplicationProvider.getApplicationContext<Application>()
val provider = context as TestCoroutineDispatchersInjectorProvider
val injector = provider.getTestCoroutineDispatchersInjector()
val testCoroutineDispatchers = injector.getTestCoroutinDispatchers()
```

(You could also use the `View`'s `Context` but it's a few extra steps and the above should work for both Robolectric & Espresso).

I will note that I was wrong in my earlier comment. I thought that this couldn't be done cleanly without Hilt because I assumed that the production application classes would need to implement the provider, but that isn't actually true since we almost always have a custom test application (and basically always should).

You may actually want to consider implementing this in this PR. It may not be as bad as it sounds because technically only the tests that use the action will need to implement the provider interfaces, and there aren't too many. I also suggest adding error checking to the sample code above (you can use optional casting via `as?` and then check that they cast correctly--failure error messages can inform developers that their test's test application class needs to implement the injector provider in order to use the action). The action's KDoc should also be updated to reflect this.

testing/src/main/java/org/oppia/android/testing/threading/BUILD.bazel
```diff
@@ -40,6 +40,7 @@ kt_android_library(
     testonly = True,
     srcs = [
         "TestCoroutineDispatchers.kt",
+        "TestCoroutineDispatchersInjector.kt",
```
BenHenning: This will need to be in its own package per my comment above.

BenHenning (changes requested):
Thanks @Sandesh282 and apologies for the delayed review. I made a mistake in my earlier review--there is a clean way to solve this. However, with the delays I think it's fine if you want to defer that to a different task (though you need to make sure it's captured correctly before we can merge this). However, I suspect that it might not actually be that much more work and is probably worth just going ahead and addressing directly in this PR.

PTAL at my comments and let me know if you have any questions.

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
Sandesh282: Alright. I have
- Removed `class @Inject constructor()` + `companion object`
- Now just `object EditTextInputAction { ... }`
- Removed `javax.inject.Inject` import

testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt
Sandesh282: Done.
- Added TODO: #6160
- Expanded KDoc explaining the correct approach (referencing `OppiaClockInjector/OppiaClockInjectorProvider`)
- Added note that "this class should not be used in any new locations"

testing/src/main/java/org/oppia/android/testing/threading/BUILD.bazel
Sandesh282: Done. Moved `TestCoroutineDispatchersInjector` into its own Bazel target (`test_coroutine_dispatchers_injector`) with restricted visibility to `//testing:__pkg__` and `//testing/src/main/java/org/oppia/android/testing:__subpackages__`. This ensures it can't be used from new locations outside testing infrastructure.

Sandesh282:
PTAL @BenHenning.

BenHenning (approved):
Thanks @Sandesh282! This LGTM!

BenHenning:
FYI @Sandesh282 I think you need to remove a line from the test file exemption (per the failing check). Otherwise @adhiamboperes will need to take a review pass yet.

testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionTest.kt
adhiamboperes, line 1: The reason why this test is causing issues with the coverage script is because it does not follow the expected naming pattern.

It should be `EditTextInputActionTest`.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -4279,6 +4279,10 @@ test_file_exemption {
   exempted_file_path: "testing/src/main/java/org/oppia/android/testing/espresso/EditTextInputAction.kt"
   test_file_not_required: true
 }
```
adhiamboperes: If you name the test file correctly you can remove this exemption.

scripts/assets/test_file_exemptions.textproto
```diff
@@ -4279,6 +4279,10 @@ test_file_exemption {
...
+test_file_exemption {
+  exempted_file_path: "testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionRobolectricTest.kt"
+  is_extra_test_file: true
+}
```
adhiamboperes: It is not correct that this is an extra test file, because no other test file exists for `EditTextInputAction`, hence its exemption above.

adhiamboperes (changes requested):
Thanks @Sandesh282! Just had some cleanup suggestions, PTAL.

testing/src/test/java/org/oppia/android/testing/espresso/EditTextInputActionTest.kt
Sandesh282: Done.

scripts/assets/test_file_exemptions.textproto
Sandesh282: Done.

scripts/assets/test_file_exemptions.textproto
Sandesh282: Done.

Sandesh282:
PTAL @adhiamboperes.

scripts/assets/test_file_exemptions.textproto
adhiamboperes: This exemption is still there, and likely why you have a failing static check. The newly renamed `EditTextInputActionTest` is the test file for `EditTextInputAction`, so the exemption is no longer valid. You should remove it.

Sandesh282: Done.

adhiamboperes (commented):
The changes LGTM, I am waiting for CI results before sending to merge queue.

adhiamboperes (approved):
Approving, thanks!

Issue #6149, opened 2026-03-12
[Feature Request]: Use official setup-bazel workflow


BenHenning:
### Is your feature request related to a problem? Please describe.

We're currently relying on an old, unmaintained workflow for setting up Bazel: https://github.com/abhinavsingh/setup-bazel/issues.

### Describe the solution you'd like

The Bazel team now maintains its own workflow: https://github.com/bazel-contrib/setup-bazel.

We should make sure that:
- This works correctly with Bazel 6.5.0.
- Doesn't require using bazelisk.
- Works correctly for all of our workflows.

Separately, this workflow supposedly may allow us to eliminate our custom caching that we're doing. This would be a really nice CI simplification but it needs to be tested and investigated to understand its limitations. Specifically:
- We want to reuse artifacts between workflows but we already shard them because of them being too large for GitHub artifacts (see #3757).
- We need to be careful about using up too much artifact space since it's shared with web and they need more of it than we do.

### Additional context

Note that the main work here is actually figuring out caching not just switching over the workflow.

ShankhanilSaha:
@BenHenning 

> Doesn't require using bazelisk.

Why is this needed? The repo has a `.bazelversion` and with `bazelisk-cache: true` we can avoid bazel being downloaded each time.

BenHenning:
We discussed this in the team chat earlier but to also ensure this is clarified here: we simply don't want to create a situation where:
- The main project requires Bazelisk.
- The CI workflows require Bazelisk and thus deviate from regular developer commands.

It's completely fine if Bazelisk is used under the surface (even if that is an alias) so long as Bazelisk can't introduce any specific failure points. Ultimately this is because we don't want the wrapper as it adds extra complexity for developers on top of an already complex build system.

ShankhanilSaha:
There are several workflows in the repo which are still using `abhinavsingh/setup-bazel`.

PR #6142, opened 2026-03-10
Fix #6141: Add workflows for issue assignment & force push checking

BenHenning:
## Explanation
Fixes #6141

This PR introduces two new GitHub Actions workflows:
1. A workflow to check for force pushes.
2. A workflow to verify issue assignments in PRs.

The force push workflow operates by checking the most recent commit on GitHub against the incoming commit and if the histories aren't compatible then it concludes there must be a force push involved (due to either dropped commits or a squash). In both cases it leaves a comment and closes the PR. Since we cannot reliably detect force pushes in retrospect, the workflow also checks if it previously commented with a closure comment and recloses the PR if it's reopened (to avoid easy circumvention). This does not stop users from force pushing before the PR is created, nor does it stop creating a new PR using the same branch (both of which are likely fine).

The issue assignment workflow:
- First checks for 'Fixes #<issue>' and 'Fixes part of #<issue>' in the PR description.
  - If these are incorrectly formatted (e.g. not on one line or uses 'Fix' rather than 'Fixes') then it will fail. We mandate this formatting requirement now.
  - If there are no issues being marked as fixed then it will also fail. This, too, is a new requirement: all PRs must be marked as fixing a specific issue or part of one.
- Then checks all of the issues being marked as fixed or partially fixed for assignment. If the PR author is not one of the issue assignees the check fails.
- If the check fails then it will move the PR into draft. This has some nice benefits:
  - It's a bit softer of a signal than outright closing the PR since the author can easily remedy the issue.
  - It enables creating drafts for not-yet-assigned issues which contributors sometimes do, and that seems fine. The workflow won't run for draft PRs.
  - It pairs nicely with the stale PR check. If a PR that shouldn't be created is created, it will move to draft. If the author then abandons the PR it will eventually be closed. This actually eliminates the full set of human steps involved in these PRs being automatically cleaned up.

For testing:
- The force push workflow was verified here: https://github.com/BenHenning/oppia-android/pull/32.
- The issue validation workflow was verified here: https://github.com/BenHenning/oppia-android/pull/29.

Both will only start running once this PR is merged since they use `pull_request_target` (which is necessary because they need to run even when contributors don't have permissions to run CI workflows as new contributors are the most likely to run into these problems).

Finally, I'm noting that a lot of this PR was generated by Gemini but then heavily refined, reduced, and simplified by me. I think I changed just about every line but Gemini did build the bulk of the workflows and probably saved me close to an hour of time by my guess.

## Essential Checklist
[...]

Issue #6141, opened 2026-03-10
[Feature Request]: Add workflows for catching common new contributor mistakes


BenHenning:
### Is your feature request related to a problem? Please describe.

We notice common recurring challenges for new contributors learning Oppia Android's approach:
- Force pushing
- Opening a PR for an issue they are not assigned

### Describe the solution you'd like

Two new workflows:
1. A workflow that automatically closes any PR that is force pushed.
2. A workflow that actually mandates a PR fixes an issue or part of an issue, and validates that the user is assigned to all such issues.

adhiamboperes:
Not sure why @github-actions reopened this. Could it have something to do with the new workflow?

BenHenning:
> Not sure why @github-actions reopened this. Could it have something to do with the new workflow?

@adhiamboperes oddly enough, this appears to have happened because the TODO bot failed to run correctly and went into an odd state of defaulting to reopening but had no comment to post. I will dig a bit on if we can make this more robust. See failure:

https://github.com/oppia/oppia-android/actions/runs/22993031591/job/66758266599

Edit: and a screenshot for posterity since the logs will disappear soon:

<img width="2335" height="1257" alt="Image" src="https://github.com/user-attachments/assets/a8a7b87c-b959-4309-8a9b-1af3d266f394" />

Issue #3926, opened 2021-10-14
Integrate with GitHub merge queues


BenHenning:
From core maintainers chat:

https://github.com/github/roadmap/issues/272 could help solve two different problems that are particularly useful for us:
1. It eliminates cases where two subsequent PRs together cause a CI breakage without conflicts preventing the second PR from being merged
2. (Hopefully) If GitHub adds an event for entering the merge queue, we might be able to split off expensive or long-running CI tasks to be blockers of the merge queue rather than for each commit. This could speed up development without needing to sacrifice test fidelity since the longer test runs only occur during merging. This could be coupled with more aggressive retry policies to reduce flakes blocking PR merges.

We saw (1) happen a few times this summer, and (2) could be *very* useful for future screenshot testing, end-to-end tests in CI, and Espresso tests in CI (/cc @vinitamurthi).

BenHenning:
Setting this up seems really straightforward: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/using-a-merge-queue. Unfortunately, it only supports standard merging & rebase/merge currently, not squash & merge (which we use). We'll need to wait until squash support is added before opting into queue support (plus, it's in beta so we might not qualify for early access, anyway).

BenHenning:
FWIW this has now been released to open beta (https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/) and squash & merge is available. I'll look into enabling this so that we can lift the "require branches to be up-to-date" requirement.

seanlip:
Per discussion with @BenHenning, we still want to do this, but it is unfortunately potentially not trivial. We use a custom system to determine what tests to run based on the changes in the PR. This needs to be updated to work with merge queues (where we run the whole suite during the merge queue pass), so we need to test it properly.

BenHenning:
Note that I did actually build out support for this in #6150 but it didn't work. See #6178 for follow-up items.

PR #6150, opened 2026-03-12
Fix part of #3926: Add support for GitHub merge queues plus a bunch of other CI improvements

BenHenning:
## Explanation
Fixes part of #3926

This PR introduces support for GitHub merge queues by:
- Enabling merge groups as a trigger for all required workflows (GitHub doesn't currently support configuring differrent required workflows for PRs vs. merge groups so the two must match).
- Of all the required workflows, those that depend on PR-specific properties (like title or base branch) have been updated to support merge group varieties.
  - For unit tests & code coverage this is far more complex because of the diffing behavior and how the synthetic PR branch differs in history from the linear merge group branch.
  - This is further complicated by '[RunAllTests]' needing to trigger a full CI run since it can be _any_ of the commits in the merge group.
  - For the check base branch CI check it should just be skipped.

Besides adding support for merge groups, this PR does a bunch of other things:
- It updates the setup Android workflow to read the SDK and build tools versions from `build_vars.bzl` so that we never need to update it again to keep things in sync (Bazel still requires an update).
- It updates the new workflows introduced in #6142 to specifically ignore the `translatewiki-prs` branch since it does not need to use the PR template nor is it prohibited from force pushes (this was missed in the original implementation).
- It refines the message for force push closes per @harshsomankar123-tech's suggestion at https://github.com/oppia/oppia-android/pull/6142#issuecomment-4042316908.
- It fixes a weird quirk with the script that reopens TODOs. If the script fails to build it will actually still continue running which leads to the issue being reopened (see https://github.com/oppia/oppia-android/issues/6141#issuecomment-4050631928).  This is being mitigated in two ways:
  - First by introducing a prior step to try and pre-build the script. If this fails then `todoIssueResolvedCheck` will be skipped which will lead to the other subsequent steps failing to run (because they rely on the workflow being `failure` status).
  - Building robustness in if there is a build failure. The issue is still reopened and a new message is left indicating that something probably went wrong. Hopefully we don't see this happen often but it's a nice catch to avoid issues being incorrectly closed when they haven't been fully resolved due to a CI flake.
  - One other change: 'reopened' was dropped because it was never actually used.
- It updates all of the third-party CI dependencies to their very latest (which actually made me realize we should be using something else for Bazel but the full audit of that is probably more work than this whole PR so I skipped it--see #6149).
- It fixes a problem that we noticed with the PR issue description check mysteriously auto-marking PRs as draft that seemed to correctly reference issues. We discovered this is because of an overly strict regex match on the template where it wasn't allowing trailing whitespace. Both regexes have been updated to allow all non-newline whitespaces (equivalent to the `\h` regex class which isn't supported in native JavaScript).

**Important**: I did not test these changes, and simply running through CI on this PR won't be sufficient. However I'm fairly sure most of these changes won't cause problems and should work without issue except maybe the merge groups (which we will see once merge queues are enabled after this PR is merged and another PR is sent to be merged). We may discover certain dependency upgrades causing problems. If that happens we can fix them forward since it shouldn't really be necessary to revert this PR.

For actual merge queues I am planning to set them up so that PRs are merged 3 times per day (every 8 hours) so try and maximize the chance of more than one PR being merged at once (though I think we average fewer than 1 PRs per day so this probably won't happen). This may be tweaked in the future to be much longer (I'm considering 24 hours) since merge queues actually overall increase the CI burden on the project (and this is causing some fairly substantial issues on Oppia web right now).

Finally, a fairly significant amount of this PR was either authored by Gemini (for the new logic portions) or otherwise helped in investigation (for merge queues and the TODO check bits). The version checks I did myself fully since there wasn't much advantage to an LLM there, and I basically rewrote everything that was suggested in one way or another. The tool was helpful in bouncing ideas mostly, though I will note it caught the potential deep fetch issue for the compute affected targets check for unit tests & code coverage--I likely would have missed that an would have needed to do a follow-up fix once we tried to use merge queues and saw it fail.

## Essential Checklist
- [x] The PR title starts with "Fix #bugnum: " (If this PR fixes part of an issue, prefix the title with "Fix part of #bugnum: ...".)
- [x] The explanation section above starts with "Fixes #bugnum: " (If this PR fixes part of an issue, use instead: "Fixes part of #bugnum: ...".)

## For UI-specific PRs only
N/A -- This is a CI infrastructure-only change that won't impact the end user experience in the app.

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Wait for unit tests to checks
-        uses: ArcticLampyrid/action-wait-for-workflow@v1.2.0
+        uses: ArcticLampyrid/action-wait-for-workflow@v1.3.0
         with:
           workflow: unit_tests.yml
           sha: auto
@@ -39,7 +39,7 @@ jobs:
     env:
       CACHE_DIRECTORY: ~/.bazel_cache
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
@@ -48,7 +48,7 @@ jobs:
         with:
           version: 6.5.0
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         id: scripts_cache
         with:
           path: ${{ env.CACHE_DIRECTORY }}
@@ -139,14 +139,14 @@ jobs:
     env:
       CACHE_DIRECTORY: ~/.bazel_cache
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: Set up Bazel
         uses: abhinavsingh/setup-bazel@v3
         with:
           version: 6.5.0
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         id: scripts_cache
         with:
           path: ${{ env.CACHE_DIRECTORY }}
@@ -195,7 +195,7 @@ jobs:
       # with Bazel since Bazel can share the most recent cache from an unrelated build and still
       # benefit from incremental build performance (assuming that actions/cache aggressively removes
       # older caches due to the 5GB cache limit size & Bazel's large cache size).
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         id: test_cache
         with:
           path: ${{ env.CACHE_DIRECTORY }}
@@ -265,14 +265,14 @@ jobs:
           bazel run //scripts:run_coverage -- $(pwd) $CHANGED_FILES --format=PROTO --processTimeout=15
 
       - name: Upload Coverage Report Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: coverage-report-${{ env.SHARD_NAME }} # Saving with unique names to avoid conflict
           path: coverage_reports
 
       - name: Upload bazel target file
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: code-coverage-bazel-targets-${{ env.SHARD_NAME }}
           path: bazel_targets.txt
@@ -287,10 +287,10 @@ jobs:
     env:
       CACHE_DIRECTORY: ~/.bazel_cache
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: Download Coverage Report Artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           path: coverage-report-artifact
           pattern: coverage-report-*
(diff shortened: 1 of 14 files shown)


.github/workflows/code_coverage.yml
```diff
@@ -86,15 +89,33 @@ jobs:
...
+
+      # Search all PR commit messages included in the merge group for any that require running all tests upon merging.
+      - name: Check merge group for [RunAllTests]
+        id: check_mg
+        if: ${{ github.event_name == 'merge_group' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MESSAGES=$(gh api repos/${{ github.repository }}/compare/${{ github.event.merge_group.base_sha }}...${{ github.sha }} --jq '.commits[].commit.message')
+          if echo "$MESSAGES" | grep -Fq "[RunAllTests]"; then echo "run_all=true" >> $GITHUB_OUTPUT; fi
```
BenHenning, line 106: Note to reviewer: this is the most "AI" part of the PR. I did not write this and can admit I only understand it at a high-level. From what I can tell it should be correct but I won't be certain until we test it.

Ditto for unit tests below (which does the same thing).

BenHenning (commented):
Quick self-review spot check.

BenHenning:
PTAL @adhiamboperes.

BenHenning:
(Turning off auto-merge because I need to fix the problem with the script preventing PRs from going into review).

BenHenning:
@adhiamboperes latest 2 commits should fix the other issue with PRs being moved to draft incorrectly. I'll update the PR description quickly to explain it, as well, and will re-enable auto-merge.

BenHenning (commented):
Quick self-review spot check.

.github/workflows/code_coverage.yml
adhiamboperes: Looks solid overall. The gh api compare - > jq flow is correct.

Minor edge case: the compare API caps at 250 commits, so anything beyond that would be missed (unlikely, but worth noting). E.g https://github.com/oppia/oppia-android/pull/5880 has 150 commits due to being part of a chain previously. Older PRs like this and the asset dowload one would be at risk.

Also worth double-checking that ${{ github.sha }} in merge_group events resolves to the merge group head SHA as expected.

adhiamboperes (approved):
LGTM overarll. I have replied to your comment, but do feel free to merge the PR.

.github/workflows/code_coverage.yml
BenHenning: I had the same concern with the `github.sha` but the easiest way to test is to actually try enabling merge groups.

The 250 commit limit is a good callout, but I don't think it will come through the way you're thinking here. Merge groups should be operating against squash commits which means hitting the 250 limit requires trying to assemble a merge group over 250 PRs at the same time, likely rare even for giant projects. Plus, if that happened then missing a single '[RunAllTest]' is probably fine because it seems extremely likely that a merge group that large would probably run everything, anyway, and this is best effort given that we always run all tests for the final merge into develop (i.e. after the merge group finishes).

BenHenning:
Thanks @adhiamboperes! I don't think there's additional work needed here--I responded to your comment and closed that thread. Restarted the cancelled build workflows and keeping auto-merge on.

